Issue #820, opened 2013-09-30
Inconsistency with handling null and empty string - (blank)


eximius313:
Google Refine has severe problems with handling nulls (https://github.com/OpenRefine/OpenRefine/issues/332, https://github.com/OpenRefine/OpenRefine/issues/252)

My problem was, that I cannot concatenate column values when one cell was null.
The solution is to "blank" cells not with null but with empty string. But there is no such option!

1)When you use option Edit cells -> Common transformations -> Blank out cells - Google Refine sets all cells to null.
2)When you do Edit cells -> Transform and you put "" - Google Refine sets all cells to null either!
3)Edit cells -> Transform and " ".trim() doesn't do anything
4)Luckily Edit cells -> Transform and " " and then Edit cells -> Common transformations -> Trim leading and trailing whitespace does the job right.

Conclusion
1)Bullets no. 2 and 3 above should perform exactly the same as no. 4
2)There should be two options: "Blank out cells" and "Null out cells"
3)Google refine should allow to concatenate nulls treating them as empty strings
4)"Text facet" should not treat nulls exactly the same as empty stings - currently both are displayed as (blank) and you cannot distinguish one from another

Kind regards

wetneb:
I agree with conclusions 1, 2 and 4 but it is not clear that 3 should be the desired behavior.

eximius313:
If there will be "coalesce" method, then I'm fine without 3

magdmartin:
@wetneb I documented the point 3 here: http://kb.refinepro.com/2011/07/merge-2-columns-that-have-both-blank.html

I think the idea is if a `column 1` has `blank` or `null` value and one does `cells["column 1"].value + " " + cells["column 2"].value`  then OpenRefine should only display `cells["column 2"].value` in the result.

wetneb:
@magdmartin yes I understand that this would be the desired result in some cases. But I think there should indeed be a "coalesce" method (that wolud translate all nulls to blanks) so that this is done explicitly.

For instance, consider a table where you would have email addresses split in two columns: one for the username (what is on the LHS of the at sign), one for the domain (RHS). If you want to recover email addresses from that with `cells["username"].value + "@" + cells["host"].value`, surely you don't want to have just `"@"` as value when both cells are null: you want that new cell to be null too, I think.

thadguidry:
[...]

PR #829, opened 2013-10-25
Text transform can set "" on null cells and vice versa.

PointyShinyBurning:
Previously if a transformation expression resulted in "" or null and the
cell's previous value was "" or null then nothing would be changed.

Partial fix for #820 ( 1 on his list) that opens smallest possible can
of worms re blank/null handling. May break backwards compatibility on
projects that relied on old behaviour, though I think it unlikely anyone
has such a thing.


wetneb:
Closing after years of inactivity and as no tests were provided. Feel free to resubmit once the comments on the PR are addressed.

ostephens:
This was eventually fixed by #1500

Issue #796, opened 2013-08-30
Columnize by key/value columns creates empty lines


hvmarck:
When using the following data set (in CSV format with ; separator):

ID;Cat;Val
1;a;1
1;b;3
2;b;4
2;c;5
3;a;2
3;b;5
3;d;3

and using the 'Columnize by key/value columns' with OpenRefine 2.6 beta 1 it creates extra blank lines (in CSV format again):

ID;a;b;c;d
;;;;
1;1;3;;
2;;4;5;
;;;;
3;2;5;;3

while it should have been (as done with version 2.5):

ID;a;b;c;d
1;1;3;;
2;;4;5;
3;2;5;;3

magdmartin:
@jackyq2015 can you check #849 and comment if it fix the issue.

jackyq2015:
This is the result I got after applied the patch made by @ultraklon 
		Id 	a	b	c	d
1		
2		1	1	3
3		2		4	5
4
5		3	2	5		3

But when I export the csv, It's fine(only 3 lines). Maybe it's because the empty line were removed when exporting.

So It's not a completely fix me.

thadguidry:
Jacky

Can you add those comments directly to #796 issue ?   Thanks.

BTW, Markdown is supported on comments in Github if you want to prettify
that alignment - https://guides.github.com/features/mastering-markdown/

Thad
+ThadGuidry <https://www.google.com/+ThadGuidry>

jackyq2015:
Tested again @ultraklon PR, It works. But as he pointed out, seems there's some discrepancy between the test result from the UI and from the Unit testing. 

Will be looking into it.

jackyq2015:
@ultraklon, for the PR #849 you created. To answer your concern for few null cell values, it's a side effect of the reuse row which aimed to address the performance. The post-operation column model is actually not consistent with the data itself because of the reuse. Ie, the cell[1], cell[2] as the old cell values should have been removed completely. Some cleanup has to be done for this operation on the column model and row model. Also releasing the null reference can free up some memory especially for big data set. 

Also for the UT, the expected cell number will be differ from row to row since not all the 5 cells are fully populated. so the expectation should also get changed.  

Will be working on the 2 changes mentioned above and hopefully can submit the PR this week.

jackyq2015:
PR created. It was based on the change @ultraklon made.

thadguidry:
@jackyq2015  still not fixed completely. See #1098

[...]

PR #849, opened 2014-02-02
proposed fix for #796

ultraklon:
Well, I removed currentRows and recordKey everywhere because I dind't find what are they meant to do.
Removing lines 164 to 169 is what fixes this problem.
I've tried it with the original error data and it works fine now.
@tfmorris:
I tried to run the test  https://github.com/tfmorris/OpenRefine/commit/2c0cb0d65700d48a68dbeec18103e1c4e71fbbe6
but it doesn't work, about this:
line 143 to 145: apparently project.rows.get(0).cells.get(1) and get(2) are allways null, I don't know why and also, when the last(s) cells of a column are empty, they are not stores, this leads to 143 to match with to be successful (it has cells 1 and 2 extra and 2 less at the end so, equilibrium) and 144 and 145 to complain saying that they have more cells than expected
line 147: "data1"? where does this come from, I think that should be "1" and for 148 and 149 have in mind last item

### DIFF
--- a/main/src/com/google/refine/operations/cell/KeyValueColumnizeOperation.java
+++ b/main/src/com/google/refine/operations/cell/KeyValueColumnizeOperation.java
@@ -130,13 +130,9 @@ public class KeyValueColumnizeOperation extends AbstractOperation {
         List<Row> newRows = new ArrayList<Row>();
         List<Row> oldRows = project.rows;
         Row reusableRow = null;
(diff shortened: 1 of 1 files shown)


magdmartin:
Qi have fix it with #1079 reusing some of your code. 
Thank you very much for your help and analysis on this issue, very much appreciated.

PR #854, opened 2014-02-11
ImportingJob.config thread-safety

FWennerdahl:
Made the remaining changes to ImportingJob and user classes to make the access to the config JSONObject thread-safe. Initial work for this was made in 1e5f89e84cb7c88cc157ae246c0bf7b9da3dc501.

Our reason for completing this was that we saw numerous Null pointer JSONExceptions thrown when calling the GetImportingJobStatusCommand during heavy load.

This is a large commit. We decided not to divide it since all changes are made to fix the same problem. Thread-safety can only be assured if all access to the config object is made in a thread safe manner.


wetneb:
Closing this stale pull request as the code base has diverged since

Issue #863, opened 2014-04-04
Form too large error when applying operations


hvmarck:
When trying to run a large (1130 steps) pipeline (using 'apply-operations' in OpenRefine 2.6-beta.1), I get an error

```
java.lang.IllegalStateException: Form too large1324593>1048576
```

As a workaround I have implemented submitting the operations sequenctialy in batches (I am also using/testing this function in an upcoming new version of my plugin):

```
function VIB_Bits_Apply_Operations(ops)
{
  var batchSize=800;
  var ofs=0;
  var next=function () {
    if (ops.length>ofs+batchSize) {
      console.log("Applying steps "+(ofs+1)+" - "+(ofs+batchSize));
      Refine.postCoreProcess(
    "apply-operations",
    {},
    { operations: JSON.stringify(ops.slice(ofs,ofs+batchSize)) },
    { everythingChanged: true },
    {   onDone: function(o) {
      if (o.code == "pending"){
        Refine.update({everythingChanged: true});
      }
      ofs+=batchSize;

      window.setTimeout(function() {
        next();
      }, 500); 
    }}
      );
    } else {
      console.log("Applying steps "+(ofs+1)+" - "+ops.length);
      Refine.postCoreProcess("apply-operations",
                 {},
                 {operations: JSON.stringify(ops.slice(ofs))},
                 {everythingChanged: true},
                 {onDone: function(o){
                   if (o.code == "pending"){
                 Refine.update({everythingChanged: true});
                   }}});
    }
  }
  next();
}
```

And pluging this in in HistoryPanel.prototype._showApplyOperationsDialog

```
HistoryPanel.prototype._showApplyOperationsDialog = function() {
  var self = this;
  var frame = $(DOM.loadHTML("core", "scripts/project/history-apply-dialog.html"));
  var elmts = DOM.bind(frame);

  elmts.dialogHeader.html($.i18n._('core-project')["apply-operation"]);
  elmts.or_proj_pasteJson.html($.i18n._('core-project')["paste-json"]);

  elmts.applyButton.html($.i18n._('core-buttons')["perform-op"]);
  elmts.cancelButton.html($.i18n._('core-buttons')["cancel"]);

  var fixJson = function(json) {
    json = json.trim();
    if (!json.startsWith("[")) {
      json = "[" + json;
    }
    if (!json.endsWith("]")) {
      json = json + "]";
    }

[...]

Issue #638, opened 2012-11-25
Can't "post" more than 1MB data to Merge Selected & Re-Cluster


bfeeny:
I am using Google Refine 2.5 on OSX 10.7.5.  I have increased my Java starting and max heap sizes to 2GB and 10GB respectively.

I am working with a document of 1011943 rows 51MB.  

I am selecting a column and making a text facet, then choosing to cluster.  There are 28,000 clusters.  I select all and then select Merge Selected & Re-Cluster and it complains I cannot post more than 1MB of data.  I have even tried to filter down to 4,000 or so clusters and I get the same.

I also find working in the Cluster & Edit View is almost unusable.  If I click on one of the filter sliders, I get a beachball (like hourglass in windows), and only after a long pause will it move.  I am using 12GB of memory, have over 3GB free, running nothing but Refine and Safari.  I am using a Mac Pro Quad Core 2.66Ghz.

I also tried to cut the file into half, import it and select just 50 clusters and its just sitting there at "Working..." and I am not sure if/when it will finish.

I am guessing something in Safari or Refine is limiting the maximum POST size to 1MB?  Apparently I need to send more data to refine when I Select All and then Merge Selected & Re-Cluster.

If there is anything I can try besides increasing the heap sizes for the application (which I have done by following the FAQ), please let me know.  Also in Activity Monitor I can clearly see the heap sizes I have set have taken effect, as well over 5GB is being used up between Refine and Safari.

I am simply wanting to take this one column of data, Cluster on it via Fingerprint, and then save it.

wetneb:
Similar to #2235

PR #876, opened 2014-05-16
Added system property refine.max_form_content_size

FWennerdahl:
To allow configuring the maximum allowed form content size in Jetty. We recieved

```
java.lang.IllegalStateException: Form too large
```

when uploading over 1MB of operation data to apply-operations. We wanted to be able to increase the hardcoded limit.

Being able to configure the allowed form content size can be useful in several cases, see https://github.com/OpenRefine/OpenRefine/issues/638 and https://github.com/OpenRefine/OpenRefine/issues/863.

### DIFF
--- a/server/src/com/google/refine/Refine.java
+++ b/server/src/com/google/refine/Refine.java
@@ -182,10 +182,11 @@ class RefineServer extends Server {
         }
 
         final String contextPath = Configurations.get("refine.context_path","/");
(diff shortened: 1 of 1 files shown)


baditaflorin:
How can i add the setting if i want to run directly the openrefine.exe file ?

cbdavis:
I was able to get this to work by adding `REFINE_MAX_FORM_CONTENT_SIZE=4194304` in the `refine.ini` file

PR #908, opened 2014-11-07
Added config item to refine.ini for system property refine.max_form_content_size

amsardesai:
When using OpenRefine with super large data sets and using the fingerprint clustering algorithm, I usually get an error (the same error as https://github.com/OpenRefine/OpenRefine/issues/638) when trying to merge and re-cluster a massive data set. I noticed that the max form content size has been added to the system properties in https://github.com/OpenRefine/OpenRefine/pull/876. This PR would make it a lot easier to modify the property by adding an entry into refine.ini, and it is commented out by default.

### DIFF
--- a/refine
+++ b/refine
@@ -917,6 +917,11 @@ if [ -z "$REFINE_MEMORY" ] ; then
 fi
 add_option "-Xms256M -Xmx$REFINE_MEMORY -Drefine.memory=$REFINE_MEMORY"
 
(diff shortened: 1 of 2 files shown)


kevinwong15:
I edited refine.ini, removed the hash and set REFINE_MAX_FORM_CONTENT_SIZE=2097152.
I am still getting the error form too large when I cluster my massive data. Any ideas?

jackyq2015:
@kevinwong15 , could you please explain in more detail what did you do and what exception stack you got from the back end if there's any?

PR #909, opened 2014-11-11
Allow longer expressions by sendig them as POST

fluf1024:
Expressions from expression-preview are send to server via HTTP GET. Because of this users aren't able to send expressions longer than a few hundreds chars (limit by url length). This fix sends expressions via POST and that solves the problem for longer scripts.


wetneb:
I think this PR looks pretty good. It is old but still looks relevant. And it works well for me. Should we merge it?

thadguidry:
checks pass and you manually verified it works via dialogs.   Merge is OK.  But I'm just wondering about client API's that might need a look over as well.  I don't think we have limits on the expression length elsewhere, but I could be wrong.  https://github.com/OpenRefine/OpenRefine/search?utf8=%E2%9C%93&q=expression&type=

PR #934, opened 2015-01-20
merge the jetty6 upgrade branch to trunk

jackyq2015:
Thank for the review by Manik, especially "use ExecutorThreadPool instead of QueuedThreadPool" for better performance in the Jetty 9.2.6


jackyq2015:
fixed the ant build

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: java
 jdk:
   - oraclejdk7
-  - openjdk6
```
tfmorris: Dropping support for an environment should be called out explicitly rather than hidden in the patch.  We'll probably replace this with openjdk7.

jackyq2015: Sure. We can replace with openjdk7. Actually I raised the issue in the dev
group yesterday. I can follow up the existing procedure to get this done.
Please advise

tfmorris:
Thanks for fixing the things that @maniksurtani pointed out.  I agree with all his comments.

The pull request also needs to be done from a feature branch, not from your forked master, because otherwise it's susceptible to getting polluted by unintended later changes (e.g. your removal of openjdk6 from the Travis config which might have been intended to be a purely local change).  Each feature branch/pull request should include a single coherent set of changes without any unrelated changes.

We won't be updating any dependencies until after the 2.6 release, but that should happen very soon.  At that point I'll be doing a major dependency update and probably another relatively quick release, so these code changes will come in handy then.

jackyq2015:
Agree. The feature branch make perfect sense.

wetneb:
@jackyq2015 what is the status of this PR?

jackyq2015:
Close it for now due to the file confliction. Separated PR needed to do the library upgrade.

PR #935, opened 2015-01-23
Change refine shell script to run with Bash

MattBlissett:
Line 215: `REVISION=${REVISION:4}` (and possibly others) use Bash syntax, but `/bin/sh` is not necessarily Bash — on Debian and Ubuntu it is usually Dash.  This should solve [this problem](https://groups.google.com/d/msg/openrefine-dev/FAJl-1XiAQM/NwvvIYgec7AJ).

I also added the "-e" option so script exits after an unexpected error.

*Simple, but not tested on OS X as I don't have access to a Mac.*


MattBlissett:
My `-e` has broken this, there's at least one error in the Bash script.  I'll look into it.

tfmorris:
Thanks for the fix.  It probably would be better to remove the bashisms from the script, but this is a simple solution for now.

PR #936, opened 2015-01-23
Convert config file line endings to LF or CRLF for Linux and Windows distributions.

MattBlissett:
A distribution built on the opposite OS has the wrong line endings for refine.ini, refine.bat and refine (which is important), and README.txt and the licenses.  This converts them appropriately during the build.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -329,18 +329,17 @@
             </fileset>
         </copy>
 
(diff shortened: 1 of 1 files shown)


tfmorris:
Thanks for the contribution.

PR #937, opened 2015-01-26
replace the openjdk6 to oraclejdk8

jackyq2015:


jackyq2015:
add the JDK8 support when removing the JDK6. Also fixed some Travis UT exception by using the LinkedHashMap instead of HashMap

PR #946, opened 2015-02-10
Fallback to English language to avoid need to maintain 'default' translation files.

MattBlissett:
I found

```
    // TODO: Switch this to just use English as the default language so we
    // so we don't have to maintain a separate redundant file.
```

so have done this.

The extension translations were identical, there was one difference between the core default and en files.


thadguidry:
:+1:

magdmartin:
+1

wetneb:
It looks like this PR is pretty uncontroversial but still has not been merged after two years… If no one complains I will merge this in the coming days.

jackyq2015:
I am fine to merge this.

PR #948, opened 2015-02-17
Change 'Add columns from Freebase' to allow using custom MQL sources.

MattBlissett:
As discussed on the [mailing list](https://groups.google.com/forum/#!topic/openrefine-dev/lRU-tleLYfM), I'd been working on an extension (based on the Freebase one) to allow OpenRefine to _read_ from MQL services other than Freebase.

This commit transfers that functionality to the Freebase extension, including a dialogue to add a new MQL service, and changes to record and display non-Freebase cells generated by the extension.

An Italian translation for a few new phrases is still needed.

It's probably not ideal that this is all one commit, but I did 90% of the work in my [extension](https://github.com/RBGKew/OpenRefine-Kew-Extension/commits/move-to-freebase-extension) before moving it across.

### DIFF
--- a/extensions/freebase/module/MOD-INF/controller.js
+++ b/extensions/freebase/module/MOD-INF/controller.js
@@ -91,6 +91,8 @@ function init() {
     [
       "scripts/extension.js",
 
+      "scripts/dialogs/extend-data-manager.js",
+
       "scripts/util/freebase.js",
 
       "scripts/dialogs/freebase-loading-dialog.js",
--- a/extensions/freebase/module/langs/translation-default.json
+++ b/extensions/freebase/module/langs/translation-default.json
@@ -88,9 +88,9 @@
 		"ok-button": "Yes, QA Data Load"
 	},
 	"fb-extend": {
-		"add-column": "Add Columns from Freebase Based on Column",
+		"add-column": "Add Columns using MQL Based on Column",
 		"warning-add-properties": "Please add some properties first.",
-		"querying-freebase": "Querying Freebase ...",
+		"querying-freebase": "Querying MQL service...",
 		"remove-column": "Remove this column",
 		"add-constraints": "Add constraints to this column",
 		"mql-constraints": "Enter MQL query constraints as JSON",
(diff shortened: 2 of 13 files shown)


thadguidry:
:+1: This looks like it is under the extension, so fine for merge I would say. We have other work to do later since the API will be deprecating as well later this year.

MattBlissett:
It is indeed all under the extension.

Does anyone speak Italian?  There are a few translations that need doing (or not, if this is going to need rewriting in a few months anyway).

fpompermaier:
I do speak italian if you need some translation. Just tell me what you need

tfmorris:
Hmm, and I was just getting ready to yank out all the Freebase stuff...  I suppose we could rename it and cut out all the Freebase-specific stuff related to Refinery (QA Data, Load, etc).

I'll do a more thorough review, but some notes to myself in the meantime:
- need to update French & Chinese translations (actually looks like they're missing)
- double check operation/protocol compatibility

tfmorris:
The Freebase extension that this depends on is currently out, but we need to decide if that's the right thing to do (the PR got merged without review). Yanking the whole thing is certainly the easiest approach, but it also removes useful functionality that is still available from Freebase (and, of course, prevents this proposed change).

jackyq2015:
We had discussion in the mailing list here:
https://groups.google.com/forum/#!topic/openrefine-dev/E8V1ogpzyqo

We may need to extract reusable classes from freebase. @MattBlissett do you have time to do that?

thadguidry:
@tfmorris We have source control, so if there are bits and pieces we need later from the extension work, we can always refactor them in for the Knowledge Graph API.  I have actually been using @MattBlissett extension against his Plant List API on a couple of fun projects.  So I would rather just see a broken Freebase menu option and I can deal with the emails regarding that it is a work-in-progress to remove the Freebase guts in a 2.6 minor release later.  Lesser of 2 evils.  KISS principle.

magdmartin:
in that case, we might

  * want to fix the pop up that say the freebase service is not available
  * add message in the application saying OpenRefine support and this
    function will be removed when freebase will be totally discontinued
    and some functionality might not work.

MattBlissett:
Hi all,

@thadguidry — I'm pleased to hear you were using the service on The Plant List!  Some researchers at @RBGKew and elsewhere were also finding it useful.

I have a new employer; I'm now working for the [Global Biodiversity Information Facility](http://www.gbif.org/) (@GBIF).  We're interested in work like this, but it will be several months before there's any time for me to work on OpenRefine or extensions.  By then, I guess the Knowledge Graph API will be released, and there won't be any point keeping MQL.  GBIF's new checklists should also be completed, which would be a good data source.

Thanks.

wetneb:
Closing this as it can no longer be merged after removal of the freebase extension.
A new version of that PR would be welcome, either with MQL or maybe with its more recent successor GraphQL.

Reconciliation services can also expose this functionality directly with https://github.com/OpenRefine/OpenRefine/wiki/Data-Extension-API (see #1179).

PR #993, opened 2015-05-01
prevent the multiple sorting

jackyq2015:
The dialog.js utility provided dismiss function and it's by level(ie if you dismiss the level 2 and the level 3 will be also dismissed). I added one function to allow it dismiss specified level so the "Working" dialog can survive after the main dialog is gone.

For the back end, changed two methods:
public synchronized FilteredRows getAllRows() 
public synchronized FilteredRecords getAllRecords() 
It will handle all the cases which processing the rows and records to prevent the reentry. Hopeful it's not overkill.

### DIFF
--- a/main/src/com/google/refine/browsing/Engine.java
+++ b/main/src/com/google/refine/browsing/Engine.java
@@ -91,7 +91,7 @@ public class Engine implements Jsonizable {
         _mode = mode;
     }
 
-    public FilteredRows getAllRows() {
+    public synchronized FilteredRows getAllRows() {
         return new FilteredRows() {
             @Override
             public void accept(Project project, RowVisitor visitor) {
@@ -134,7 +134,7 @@ public class Engine implements Jsonizable {
         throw new InternalError("Unknown mode.");
     }
 
-    public FilteredRecords getAllRecords() {
+    public synchronized FilteredRecords getAllRecords() {
         return new FilteredRecords() {
             @Override
             public void accept(Project project, RecordVisitor visitor) {
--- a/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
@@ -327,7 +327,7 @@ DataTableColumnHeaderUI.prototype._showSortingCriterion = function(criterion, ha
   elmts.blankErrorPositions.html(positionsHtml.join("")).sortable().disableSelection();
 
   var level = DialogSystem.showDialog(frame);
-  var dismiss = function() { DialogSystem.dismissUntil(level - 1); };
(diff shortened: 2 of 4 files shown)


main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
```diff
@@ -327,7 +327,7 @@ DataTableColumnHeaderUI.prototype._showSortingCriterion = function(criterion, ha
   elmts.blankErrorPositions.html(positionsHtml.join("")).sortable().disableSelection();
 
   var level = DialogSystem.showDialog(frame);
-  var dismiss = function() { DialogSystem.dismissUntil(level - 1); };
+  var dismiss = function() { DialogSystem.dismissLevel(level - 1); };
```
tfmorris: Why is this dismissing a single level partway down the stack instead of unwinding everything?

main/src/com/google/refine/browsing/Engine.java
```diff
@@ -91,7 +91,7 @@ public void setMode(Mode mode) {
         _mode = mode;
     }
 
-    public FilteredRows getAllRows() {
+    public synchronized FilteredRows getAllRows() {
```
tfmorris: Is this needed in addition to the front end change to show the busy dialog or is it redundant?

While it doesn't make sense to have multiple operations per client, there may be multiple clients, so we don't want this single threaded unless there are real synchronization issues to protect against.

main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
jackyq2015: Due to the order of the creation of the dialog, if use the dismissUntil the "waiting" window will get dismissed automatically when click the "sort" button. So dismiss the specific dialog to prevent the over kill.

main/src/com/google/refine/browsing/Engine.java
jackyq2015: If look at the RowVisitor interface, the getAllRows method is NOT read only which sounds like. so  it's also necessary to protect the methods since there is client such as python-client can access it in different ways

tfmorris: Sorry for the delay, but I wanted to make sure I wasn't missing something since synchronization is something that is non-trivial to get right.

In this case, I don't think this is going to gain you any advantage.  The *Engine* objects are created per HTTP request using the serialized facet description which is passed in with the request.  Since synchronized methods use the object's default lock, it's only protecting against multiple invocations from within a single HTTP request which would only be as the result of a particularly pathological bug in our code.

I can't ask you to revise the PR since @magdmartin already merged it, so I'll back out those changes by hand.

Issue #980, opened 2015-04-21
"on error"  isn't translated


magdmartin:
In the edit *cells* > *transform* windows expression, the string "on error" have been translated in Italian, French or Chinese but it isn't displayed. See here for example:

![screenshot from 2015-04-21 08 30 59](https://cloud.githubusercontent.com/assets/2589584/7252211/f4cdde62-e800-11e4-9f1a-7447da218fa6.png)

I trace it down to this line but I don't see anything wrong with it: https://github.com/OpenRefine/OpenRefine/blob/9b2a506caada4ea6d580d2140184fef8caa7566c/main/webapp/modules/core/scripts/views/data-table/menu-edit-column.js#L103

tfmorris:
Thanks for the bug report.  If the code is correct, the next place to check is the French translation.

magdmartin:
This affect all translation, not just the French one. So I think it comes
from the core, but I haven't been able to narrow the root cause.

tfmorris:
https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/scripts/views/data-table/text-transform-dialog.html

tfmorris:
Should I expect a pull request for this?

magdmartin:
I personally don't have the knowledge to fix that.

jackyq2015:
I will help with this to create a PR to fix this.

PR #998, opened 2015-05-02
to fix the ticket #980

jackyq2015:
Fix the "on error" label:
https://github.com/OpenRefine/OpenRefine/issues/980


tfmorris:
Thanks @jackyq2015 ! I probably won't be able to fully verify until Sunday, but merging on the assumption that it's good.

Issue #962, opened 2015-03-20
Wrong prompt on create project page


tfmorris:
The prompt for file selection on the Create Project tab says "Locate an existing Refine project file," which is wrong.  It sounds like it got mixed up with a prompt for Import Project.

![openrefine-issue-962](https://cloud.githubusercontent.com/assets/82178/6760543/fe2f1024-cf1f-11e4-9eef-dc87abac7cc6.PNG)

tfmorris:
This appears to be fixed in the current development version (and rc1).

magdmartin:
testing RC-1 on linux I still have *Locate an existing Refine project file (.tar or .tar.gz):* when creating a project.

tfmorris:
I just double checked the 2.6-rc1 Windows kit and it's definitely fixed.  It seems unlikely that it's a Linux specific bug.  

@magdmartin Can you make sure that you're not seeing cached artifacts by doing a force reload in your browser.  What browser are you using on Linux?

magdmartin:
Fixed in Chrome
I cleared the cache in Firefox and still see it.

tfmorris:
Very puzzling since the only browser specific code that we have is the old IE shim.

Hopefully someone else with a Linux system can help debug.

Just so we have as much information as possible - what distribution and version of Linux? What version of Firefox & Chrome.

jackyq2015:
under windows 7, 64 bit, 37.0.1 firefox, I have the same issue. chrome Version 42.0.2311.90 m works fine

jackyq2015:
It turned out that the label Id of the "import from local" and "import from project" is the same.  FF and firefox handle this situation differently.

I will fix it and create a PR

tfmorris:
No need.  I have a fix in hand already.  Thanks for investigating.

Tom

tfmorris:
Fixed by 75ec34ea93c6979b96d17e72ce17998c2178d940

PR #999, opened 2015-05-03
fix issue #962

jackyq2015:
verified under windows 7, chrome and firefox


tfmorris:
Thanks for looking into this.  I had already fixed it with 75ec34ea93c6979b96d17e72ce17998c2178d940 by the time I saw your PR.

jackyq2015:
It's ok as long as it got fixed. I deleted my fix branch already

Issue #983, opened 2015-04-21
proposal to integrate a new cluster library


lispc:
Hi I am a research student from [Database Group](http://dbgroup.cs.tsinghua.edu.cn/) , Tsinghua University and we have developed a java clustering library https://github.com/lispc/EditDistanceClusterer  which is much faster than the current simile-vicino used in OpenRefine.  I wonder whether it is possible to integrate the lib to OpenRefine. What features / tests / performance reports are needed ? (First time to do a open-source pull request, sorry for anything not corretly done)

baditaflorin:
I don`t know what is the procedure, but if you say that is faster, this is a good think, i would love to test it, when it will be intergrated into OpenRefine. 
I run Clustering on big datasets ( 15 M rows ) and usualy i have to wait around 40-50 seconds even on the simpliest algorithm 

Here is the code for the actual clustering methods https://github.com/OpenRefine/OpenRefine/tree/9b2a506caada4ea6d580d2140184fef8caa7566c/main/src/com/google/refine/clustering/binning

tfmorris:
Welcome - and thanks for the offer.  We would love to have an additional high performance clustering algorithm.  The existing simile-vicino algorithms are integrated here: https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/clustering/knn/kNNClusterer.java but that assumes that blocking is required.  Does the new algorithm use no blocking at all or just larger blocks? (I haven't had a chance to read the paper yet).

If you have a paper or blog post which describes the performance, that would be sufficient for the performance side of things.  For test coverage, we prefer to have test coverage for all new features (although we're not very good about enforcing that).

Next steps:
- fork the OpenRefine repo
- create a feature branch
- integrate your code and tests
- submit a pull request for us to review

If you have any questions, please ask on the openrefine-dev list.  We look forward to your contribution!

lispc:
@baditaflorin.  Thanks for the link. I have already read the code on clustering in OpenRefine / Simile-Vicino / SecondString all. My algorithm is particularly optimized for edit-distance clustering, so my target is not to replace most of the existed algorithms  but  only replace the edit-distance module. 
[...]

PR #1007, opened 2015-06-01
 Use new algorithm for levenshtein clustering

lispc:
This pull request is based on the discussion at https://github.com/OpenRefine/OpenRefine/issues/983 .
(1) The new algorithm is about 60 times faster than the current levenshtein-distance knn algorithm used in OpenRefine. (Detail : https://gist.github.com/lispc/d4ee8de81e7bfa6bc352)
(2) Even the new algorithm is fast, as a knn clustering algorithm, it is still much slower than binning method. 
(3) I also change some html and js of the cluster-dialog.


tfmorris:
Sorry for the delay in reviewing this.  Where is the source for EditDistanceJoiner.jar?  We don't accept binary contributions without source that we can recreate them from.  Also, what is the license associated with the code.

I'm going to revert this for the time being until we get these questions sorted out, but we're definitely interested in the contribution.

jackyq2015:
Agree, we should include the source code also in order to review and distribute

zyadtaha:
@wetneb I think this is the PR that includes optimizations for the distance clustering we discussed.

I suggest prioritizing:
- Designing the "view clusters" tab
- Dealing with the issue of limited space
- Implementing and merging the code

Afterward, in the next week or the week after, I can focus on integrating this interesting algorithm.

Issue #925, opened 2014-12-21
Update Apache POI to 3.10.1


tfmorris:
We are currently using 3.8 and there have been three releases since then, but 3.11 was just released today, so a safer strategy is probably to update to 3.10.1.

jackyq2015:
The latest version is 3.12
https://poi.apache.org/download.html#POI-3.12

Should we update to this version?

jackyq2015:
done. PR #1029 created for this.

PR #1029, opened 2015-07-11
upgrade POI from 3.8 to 3.12

jackyq2015:


magdmartin:
fix for #925

tfmorris:
The reason this was still pending is because it contained unverified binary files.  I'm not going to revert it, but all the files still need to be manually verified.

tfmorris:
I've replaced the binaries with those for 3.13 and deleted the POI 3.8 source jar.

PR #1047, opened 2015-07-30
Fixed #1046 Combine xls and xlsx formats by inspecting file header information in ExcelImporter

### LINKED ISSUE #1046: Combine xls and xlsx formats by inspecting file header information in ExcelImporter
Currently, Refine differentiates between xls and xlsx, and presents these as separate formats to the user.  Both formats use the ExcelImporter class, but depending on which format is picked a xmlBased flag is set.  This flag determines how the inputstream should get parsed by the apache.poi libraries.

Instead of performing the differentiation based on the format string, the inputstream's header can easily be inspected using the apache.poi library, as xlxs files have a zip header signature.  (Make sure to test if the input stream supports the mark and reset methods when checking the header and add the mark/reset ability to the stream if needed.)  When using this method, it is only necessary to present one format to the user instead of two.

lemmingapex:
fix for #1046

### DIFF
--- a/main/src/com/google/refine/importers/ExcelImporter.java
+++ b/main/src/com/google/refine/importers/ExcelImporter.java
@@ -37,12 +37,14 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.PushbackInputStream;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


magdmartin:
Thank you for the improvement. Tom will review and merge your PR
FYI, we are currently working on the 2.6 release so it might not be 
merge immediately.

Martin

tfmorris:
Thanks for the contribution!

Issue #1060, opened 2015-09-04
Inconsistent selection behavior with "Cluster & Edit"


joewiz:
In the "Cluster & Edit" window, if I select a checkbox (to indicate that I want to merge entries) and select merge, the merge works as expected. Or if instead I click on one of the entries (to change the value to be used for merging) and then select merge, the merge works as expected. But if I select a checkbox AND click on one of the entries, I get an error dialog, stating: "The page at localhost:3333 says: You must check some Edit? checkboxes for your edits to be applied." Only by carefully *un*checking the checkbox will the merge proceed as intended. Alternatively, re-clicking on one of the entries will allow the merge to proceed. Clicking on an entry should only re-affirm the checkbox's checked status; it should not un-select the checkbox or cause a hidden selection toggle.

(I'm using 2.6RC1 with current versions of Mac OS X and Google Chrome.)

magdmartin:
not sure if it is a duplicate or related to ##1056

joewiz:
Yes, this is the same issue, though I provided some more detail than #1056.  The most concerning aspect of this is that the behavior can result in merges even without the checkbox checked.  In other words, this can lead to merges even when the user has intentionally said "no" by unchecking a box.  Less concerning but equally a bug is when the user intends the merge, but the merge isn't made - even though the checkbox is checked..

magdmartin:
Going back few version this bug has been introduce between the 2.5 and 2.6-beta release.

I am also closing #1056 as this issue provide more details.

jackyq2015:
can't reproduce it under windows7 + chrome(Version 44.0.2403.130 m). Can anyone confirm this only happens on the Mac?

magdmartin:
I have the issue with Linux. You need to click on the merge button to get
the error message.

joewiz:
Many thanks again, @jackyq2015 and all!

PR #1065, opened 2015-09-11
fix issue #1060 to skip the checkbox deselect when clicking on the cl…

jackyq2015:
…ustering text candidate

### DIFF
--- a/main/webapp/modules/core/scripts/dialogs/clustering-dialog.js
+++ b/main/webapp/modules/core/scripts/dialogs/clustering-dialog.js
@@ -175,7 +175,9 @@ ClusteringDialog.prototype._renderTable = function(clusters) {
               cluster.value = value;
               
               parent.find("input[type='text']").val(value);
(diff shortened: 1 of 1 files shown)


joewiz:
Wonderful! Thank you, @jackyq2015!

magdmartin:
fix for #1060

main/webapp/modules/core/scripts/dialogs/clustering-dialog.js
```diff
@@ -175,7 +175,9 @@ ClusteringDialog.prototype._renderTable = function(clusters) {
               cluster.value = value;
               
               parent.find("input[type='text']").val(value);
-              parent.find("input[type='checkbox']").attr('checked', true).change();
+              var checkbox = parent.find("input[type='checkbox']");
+              if (!checkbox.attr('checked'))
```
tfmorris: Conditionalizing this doesn't fix the underlying issue which is that the internal state doesn't track the state of the checkbox.  Instead the fix should be applied at line 222.

Issue #995, opened 2015-05-01
Change Jython to include the word Python to make it easier to find


tfmorris:
Not all users who are familiar with Python know that Jython is the name for the version that runs in the JVM.  We should change the name in the UI to make it easier for people to identify.

tfmorris:
I'm going to use the wording from July 29 version that I created in b296c29937b20bf3cf7f12b25fc94284615a11d1as you can see in the history above.  I don't see the point of leaving Google Refine references lying around.

PR #1070, opened 2015-09-22
fix issue #995

jackyq2015:
fix issue #995


tfmorris:
I'm going to use the wording from July 29 version that I created in b296c29937b20bf3cf7f12b25fc94284615a11d1. I don't see the point of leaving Google Refine references lying around.

PR #1071, opened 2015-09-22
Concurrent access issue to jython evaluables.

mgalushka:
Description: any jython function used in the project has implicit name __\__temp\____ which creates issues when there are multiple projects which transforms are applied concurrently.

Fix replaces default name to include hashcode of function code itself and making this safer in concurrent environments.

Created corresponding test which was failing before and now passes.

We are using openrefine on production to apply transformations to datasets and found that during concurrent access sometimes we are getting corrupted data.

Patch will also help to all forks which are trying to make batch processing transformations engine based on openrefine (what we actually did for internal usage)


tfmorris:
Thanks for the contribution -- and for including a test!

In the future, please review the diff before you submit it and make sure it's clean of changes unrelated to its purpose (e.g. the import reorg that your IDE apparently decided was necessary).  Otherwise we just get noise injected from dueling editor/IDE setups.

mgalushka:
(y)

PR #1075, opened 2015-09-27
change in format guesser to enable a difference of more than 0,1

stundzig:
Hi, 

i'm working with the attached CSV, which is ; separated. But the SeparatorBasedImporter returns *null* because of the test in line 245.

*separator.stddev / separator.averagePerLine* for  this file results in 0.10101010, which is more than the 0.1 in the if clause. What is the reason for this test and also for this 0.1? If i remove the if statement its working great.

Regards
   Steffen...

### DIFF
--- a/main/src/com/google/refine/importers/SeparatorBasedImporter.java
+++ b/main/src/com/google/refine/importers/SeparatorBasedImporter.java
@@ -242,9 +242,9 @@ public class SeparatorBasedImporter extends TabularImportingParserBase {
                     });
                     
                     Separator separator = separators.get(0);
-                    if (separator.stddev / separator.averagePerLine < 0.1) {
+                    //if (separator.stddev / separator.averagePerLine < 0.1) {
                         return separator;
-                    }
+                    //}
                    
                 }
             } finally {


magdmartin:
Thank you for the pull request, however I can't find the attached filed. I think Github now support file attachment, can you retry to upload it. 

On a separate note, I am curious to learn more about the integration you are doing with polymap4

Thank you.

stundzig:
Hi,

you are right, i couldn't upload the file. Hmm. It's the CSV file from https://www.govdata.de/suchen/-/details/stala-sn-service-1679024015 a german open data website.

stundzig:
With polymap4 we create a geo location plattform. There you could e.g. upload CSV files. And i use OpenRefine for the parsing of the CSV content. Therefore i've written a server side implementation class, which encapsulates all the refine methods, see https://github.com/stundzig/polymap4-core/blob/develop-rap2.3/plugins/org.polymap.core.data.refine/src/org/polymap/core/data/refine/impl/RefineServiceImpl.java

tfmorris:
The purpose of that check is to make sure that there's a good likelihood that the character being returned is actually a separate as opposed to just frequently occurring random character from the data stream.

We're not going to just remove the check, so I'm closing this PR.  Feel free to create an issue which describes what behavior you think is wrong.  I'd need to dig into your example more, but I'm guessing that there is some other underlying issue which is triggering this behavior since, at a glance, the file appears to have exactly 12 semicolons (;) per line which should easily pass the check.

As an aside, any time you have the data available in both Excel and CSV formats, you should choose the Excel format because it contains more information.  In this case it correctly distinguishes the strings which look like numbers from actual numbers.

tfmorris:
I had a closer look at the file and the behavior is being triggered by the first line which contains no semicolons.  This makes the standard deviation just high enough to exceed the threshold.  The file is somewhat unusual in that, in addition to the first unformatted line, it also contains 5 lines of sheet header description and 5 lines of column headings.

Some possible solutions:
- always ignore the first line
- increase the threshold allowed for the standard deviation of separator counts
- redo the separator guessing when the user changes their import parameters (since the user is presumably going to ignore the first 5 lines anyway)

Feel free to transfer this discussion to the issue you create.

stundzig:
Hi Tom, thanks for your thoughts. I think, increasing the threshold would have only a minimal influence to the other code. And in my file i have a value of 0.1010101, but the check was only <0.1.

PR #1076, opened 2015-09-28
ES translation updated

nestorjal:
Line 516 fixed.


tfmorris:
Thanks for the update.  I'm not sure how familiar you are with git, but if you can create separate branches for your pull requests, it minimizes the chances that two unrelated changes on your `master` branch will interfere with each other.

Thanks again for all your help with translating!

nestorjal:
Actually I'm learning how to use git, thanks for your advice! count me in if any update is necessary or new strings are created.

PR #1081, opened 2015-10-16
Updated build.xml to include a RUN target

thadguidry:


tfmorris:
Thanks for the PR.  Currently our build/run setup is a weird mix of command shell and Ant.  The run stuff is all currently in the refine / refine.bat scripts.  Is there a specific need to move this piece from the shell scripts to the Ant file?  Note also, that the shell script does a bunch of additional processing.

While I'm not opposed to merging this, I'd prefer to make a single concerted cleanup pass through everything (which might also mean switching to Maven instead of Ant).

As an aside, it's preferable to have the PR on it's own branch (e.g. thadguidry-patch-1) as opposed to master, so that it doesn't accidentally get polluted with later changes that are unrelated.

thadguidry:
The specific reason was getting the Netbeans environment playing nicely.  I have some other PRs coming for that later.  Good point, I'll create my own branch and then ask for a PR there. Thanks Tom! and welcome back ! :)

thadguidry:
@tfmorris @jackyq2015 @magdmartin @MattBlissett 
OK, I am now working on setting up our new Gradle build.  (and with the Gradle Wrapper, so you won't even have to install Gradle beforehand !)  I'm about halfway done with the new organizational changes, but from a high level it will be treated as a [Gradle multi-project build](https://docs.gradle.org/current/userguide/multi_project_builds.html) - A multi-project build in gradle consists of one root project, and one or more subprojects that may also have subprojects.:

* /openrefine
  * /client
    * build.gradle
  * /server
    * build.gradle
  * /shared
* build.gradle
* settings.gradle

I will maintain it.

This NEW build way will be pushed into my repo later for eveyone to review once I am done, Then later we can clone it under openrefine/gradle_build at some point in the future once everyone is happy. (this way I don't slow anyone down,and we get an easier build and even easier distributions !! in the future)

(@tfmorris - And, we will not even have to lose the Eclipse .project settings etc. [apply plugin: 'eclipse'](https://docs.gradle.org/current/userguide/eclipse_plugin.html)

I looked at the current state of affairs with Maven and SBT and I still was not pleased enough - most of the major players are now going Gradle and for good reason, its also defacto standard added now in Eclipse Mars !! - no plugin needed.  We also use it at Ericsson)

thadguidry:
I don't need this anymore.

Issue #1086, opened 2015-10-30
single quotes as quote characters


stundzig:
hi folks,

i've a csv file with ' as quote characters. As i see in the SeparatorBasedImporter, i can change the quote character in calling the OpenCSV CSV Parser. But currently refines options, doesn't have a parameter for that. So i created a fork and a pull request which adds the quote character on server side.

PR #1087, opened 2015-10-30
#1086 add support for quote character

stundzig:


thadguidry:
Looks like we don't really have an option that "keeps" the quote characters if a user wants.

    "this is somethng","this is something else"

Currently, on the preview, we remove the double quote characters, regardless of the checkbox for "Quotation marks are used to enclose cells..."

I see 3 further enhancements also needed.
1. A click option to keep the quotes but still separate on a separator char. (supporting single or double quotes)
2. Changing the phrase "Quotation marks are used to enclose cells containing column separators" to the actual action that it will perform, similar to our other import options that use an actionable label.
3. Give the user a clickable option to allow or not on single quotes.

@stundzig your pull request does not address those, but actually hides to the user the data manipulation that OpenRefine is doing during import.  We try to avoid hiding to the user what is happening to their data and instead like to give them checkboxes, etc to apply actions and visually see the changes during preview.  Can you work on those further enhancements and the above and request again ?  We would absolutely love that !

wetneb:
Closing, see #1452.

PR #1091, opened 2015-11-17
Fix typo on encodeURIComponent JS call

vitorbaptista:
I found this while reading the code. I haven't tested it, but as this is the only instance of the `encodeURICompnent` string in the codebase, I suppose it's a typo.


tfmorris:
Thanks!

Issue #871, opened 2014-05-05
Close button not displayed properly when selecting Facet Choices as Tab Separated Values


magdmartin:
Using OpenRefine 2.6 beta (with chrome on ubunut machine) the close button in the window Facet Choices as Tab Separated Values is not displayed properly.

![screenshot from 2014-05-05 11 12 00](https://cloud.githubusercontent.com/assets/2589584/2879265/c82b4d02-d467-11e3-97c1-932f124a5a0a.png)

tfmorris:
Fixed in 2.6-rc1

Issue #344, opened 2012-10-15
Feature request - Flag cluster of selected records


tfmorris:
_Original author: sachdev...@gmail.com (March 02, 2011 09:28:20)_

<b>What steps will reproduce the problem?</b>
1. On column, edit cells and then Cluster &amp; Edit

<b>What is the expected output? What do you see instead?</b>
Along with 'Merge &amp; Recluster' button, there should be one button 'Flag Selected Records'

<b>What version of Google Refine are you using?</b>
2.0

<b>What operating system and browser are you using?</b>
Ubuntu 10.10

<b>Is this problem specific to the type of browser you're using or it happens</b>
<b>in all the browsers you tried?</b>

<b>Please provide any additional information below.</b>
Feature Request

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=344_

PR #1092, opened 2015-11-23
Add export button to "Cluster and Edit"

corajr:
This PR adds an export button to the "Cluster and Edit..." dialog, which saves the proposed clusters to a JSON file. (See attached screenshot.)
Export button

We'd be happy to make any needed changes-- for example putting the button in a different position or changing the way the export is generated.


magdmartin:
Thanks you for your contribution. I am sure a lot of user will appreciate
the cluster export.

Tom will review and merge the PR.

tfmorris:
Thanks for the contribution.  Can you explain a little bit more about the motivation/use case?  Does this match up with an enhancement request in the issue tracker?  Some use case of your own?

corajr:
Certainly! @kgrons might be able to say more, as she was hoping to use this at her day job, but I'll try to explain.

We wanted to address a situation in which merging clusters and deciding on the canonical representation of a datum requires input from stakeholders beyond just the person using OpenRefine. Higher-ups who would like to be involved in the process would be unlikely to install OpenRefine on their own machines and learn how to use it just for this purpose.

Our use case actually requires two components: the export functionality in OpenRefine, and a separate tool that could consume this JSON and display it in a separate interface. To this end, we also built [a proof-of-concept tool](https://github.com/amiaopensource/TIDY) that can work with the data. (Such a tool is certainly out of scope for the main OpenRefine project, but it can't work at all without some way to get the clusters out of OpenRefine.)

kgrons:
Thanks for the explanation @corajr - as mentioned, this is our primary use case. The fundamental use case underlying a cluster export feature is the ability to output clustered data with relationships outside of OpenRefine in a portable format, so that people can build whatever they need - for us, a prototype vocabulary selection tool and some basic data viz with freely available tools. 

The high value of the clusters as decision-making tools is echoed in [Enhancement Request 344](https://github.com/OpenRefine/OpenRefine/issues/344), which requests a flag on a cluster so that users can "export the clustered records out of refine to perform some manipulation on a external database".

We believe the cluster export is within the current scope of OpenRefine as a new option that enables more data normalization by not limiting decision-making activities to the tool itself.

joewiz:
Just spotted this - nice work! Is there a facility here for re-importing the results of a decision set, or does the PR just export the clusters from OR? (I'm not a reviewer - just an interested user.)

vladan-me:
This is quite useful, I have similar use case as already mentioned - export clusters, use them in a separate program for human verification before merging.

kgrons:
re: @joewiz - this PR is only to add the cluster export function/button.

We hope adding this feature encourages additional PRs to import & merge data from external processes. cc @corajr

thadguidry:
@kgrons The Export Cluster button barely shows up for me on latest Firefox and Chrome.  Can you fix that and request pull again ?  Then I'll be happy to merge if my functional tests pass.

almostinvisibleexportclusterbutton

magdmartin:
@thadguidry this is similar to #871 and looks like a missing translation or text.

thadguidry:
@kgrons The download does work and is good Json representation.  Button text just needs to be fixed and then I can merge.

corajr:
Thanks for the movement on this! I've tried to update all the translations to ensure that none are blank. Can you please let me know if the text still isn't showing up for you?

thadguidry:
@corajr Works perfect. Thanks ! Merged.

joewiz:
@corajr @kgrons Congrats!

Issue #486, opened 2012-10-15
Character encoding not being set in new projects


tfmorris:
_Original author: tfmorris (November 18, 2011 19:42:20)_

The problems that Thad reported with reintepret here:

http://code.google.com/p/google-refine/issues/detail?id=237#c6

I think are due to the character encoding being set to &quot;&quot; in newly created projects.

It looks to me like if the user doesn't explicitly select a character encoding we are using &quot;&quot;.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=486_

drangons:
Hi,
I am the issue with importing text written in German language. When csv file is imported to refine, the column values are displayed as ? for ö ü characters. I am using 2.5-r2407 . 
for example für is displayed as fÃ¼r.

Can you suggested me what can be done to get it to work.
Thanks.

Update: I understand that I have to change the character encoding. Can it be configured in file ? .

jackyq2015:
Is this still a problem for current code base? seems Tom already fixed it by putting a default encoding when it's empty.

I tried the testing file under win 7. it works fine. If someone can provide the testing file and procedures to reproduce it, that will be helpful.

jackyq2015:
@tfmorris I tried the attached testing file with GREL reinterpret(value,"utf-8") and reinterpret(value,"big5"). Neither of them has the error as mentioned in original post. 

Could you please elaborate if you still think it's problem?

magdmartin:
I still have the error in 2.5 and 2.6rc-1 with both chrome and firefox on a linux machine. Here the project with the error: https://drive.google.com/file/d/0B7tW6wTrzLf7X3dkQUg2QUVzbG8/view?usp=sharing

jackyq2015:
I downloaded the project and confirmed that the encode was wrongly set to us-ascii. In windows the same file when importing it is set to utf-8. Will come up a solution for this issue under Linux.

jackyq2015:
@magdmartin can you please run command locale at command line and paste the result?

magdmartin:
martin@martin-Lenovo-G585:~$ locale
LANG=en_CA.UTF-8
LANGUAGE=en_CA:en
LC_CTYPE="en_CA.UTF-8"
LC_NUMERIC="en_CA.UTF-8"
LC_TIME="en_CA.UTF-8"
LC_COLLATE="en_CA.UTF-8"
LC_MONETARY="en_CA.UTF-8"
[...]

PR #1094, opened 2015-11-27
#486 add encoding guesser based on icu4j

stundzig:
Hi folks,

i've added a encoding guesser based on icu4j, which checks during _load-raw-data_ what is the most likely character encoding. Also currently i set the first found encoding, as initial ui-option.

I would also propose to update the icu4j version to the latest 56.1 -> http://icu-project.org/apiref/icu4j/

Regards
   Steffen...

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -84,5 +84,6 @@
 	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/poi-3.13-20150929.jar"/>
 	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/poi-ooxml-3.13-20150929.jar"/>
 	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/poi-ooxml-schemas-3.13-20150929.jar"/>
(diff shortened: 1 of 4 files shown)


tfmorris:
Thanks for the contribution.  I'm not sure we've got the headroom to include ICU4J and still make it under Github's 100MB file size limit for the Mac kit, but I'll have a look.

wetneb:
Looking at the latest releases, it seems that the aforementioned 100MB file size limit is no longer true (122MB for Mac). Should we merge this? I think it would be good to have a test or two to demonstrate the cases where that is useful. Maybe something based on #1197?

thadguidry:
@wetneb I'll defer to @jackyq2015 as to how we should handle this.  @jackyq2015 thoughts ?

jackyq2015:
Added the inline comments. But I agree that encoding detector such as ICU4J should be added into OR

wetneb:
Closing because of failing build - feel free to rebase and resubmit, ideally with tests.

Issue #1093, opened 2015-11-26
Problem with Columnize by key/value columns


psychemedia:
Using OpenRefine v2.6-rc2 Release Candidate 2 (linux) I'm getting a problem with columnize by key / value, eg in the below trying to columnize on the Indicator / Value columns I get the following pop-up:

![image](https://cloud.githubusercontent.com/assets/82988/11423541/2fa519b8-943d-11e5-98e8-882f6f3884b8.png)

felixlohmeier:
Same problem here with 2.6rc2

> 14:24:43.800 [ refine] POST /command/core/key-value-columnize (67718ms)
> 14:24:44.375 [ command] Exception caught (575ms)
> java.lang.IndexOutOfBoundsException: Index: 1, Size: 0
> 	at java.util.ArrayList.rangeCheck(ArrayList.java:653)
> 	at java.util.ArrayList.remove(ArrayList.java:492)
> 	at com.google.refine.operations.cell.KeyValueColumnizeOperation.createHistoryEntry(KeyValueColumnizeOperation.java:262)
> 	at com.google.refine.model.AbstractOperation$1.createHistoryEntry(AbstractOperation.java:52)
> 	at com.google.refine.process.QuickHistoryEntryProcess.performImmediate(QuickHistoryEntryProcess.java:73)
> 	at com.google.refine.process.ProcessManager.queueProcess(ProcessManager.java:85)
> 	at com.google.refine.commands.Command.performProcessAndRespond(Command.java:248)
> 	at com.google.refine.commands.cell.KeyValueColumnizeCommand.doPost(KeyValueColumnizeCommand.java:66)
> 	at com.google.refine.RefineServlet.service(RefineServlet.java:177)
> 	at javax.servlet.http.HttpServlet.service(HttpServlet.java:820)

jackyq2015:
duplicate of #1138 
PR already created

PR #1098, opened 2015-11-29
Issue 1093 Problem with Columnize by key/value columns

jackyq2015:
fix Issue #1093 Problem with Columnize by key/value columns

Added the boundary protection and enhanced the Unit Testing.


thadguidry:
@jackyq2015 @tfmorris  I tested using #796 use case given by user and using clipboard import, however, @jackyq2015 only partially fixes the bug introduced, and still cannot perform a complete UNDO after the Transpose change because the Cat and Val columns are blown away.

jackyq2015:
not fully fixed. Also have issue with undo/redo since the original table structure is altered after transpose.

PR #1137, opened 2016-06-08
translated language-package into Japanese

ymnder:
Hello :) I translated language-package into Japanese.


magdmartin:
Thank you a lot. We usually merge translation related PR quickly, but I want to understand what the AppVeyor build failed first.

ymnder:
Thank you for your reply. Maybe failed reason is that there is no solution file or project in the directory. So if not contain those files, it is necessary to turn off build settings on project settings page. https://www.appveyor.com/docs/build-phase

tfmorris:
Thanks for the contribution. It's always great to have a new translation!

(I've disabled the AppVeyor build until the setup is complete).

PR #1155, opened 2016-09-19
Add labels for checkboxes in text facet, making them clickable

joewiz:
Checkboxes are small targets. Making the labels clickable for toggling checkboxes improves usability.


joewiz:
Thanks @jackyq2015!

PR #1160, opened 2016-11-15
fix not working clustering functions when it select japanese lang.

tomokane:
Hello. I fixed bug.

Select Japanese lang and, when change keying functions on cluster dialog, it does not working.

So I changed for sending value to server from option tag text to option attribute value.

Thanks.


thadguidry:
Thanks for this fix !  Tested against English to ensure no breakage of functionality of Clustering methods.

PR #1166, opened 2017-02-07
Change default reconciliation service to Wikidata

wetneb:
This PR proposes to replace the default (and defunct) Freebase reconciliation service by the Wikidata version.

OpenRefine's language setting is propagated to the Wikidata reconciliation service, autocompletion and previews by means of the "wd-recon-lang" string in the translation files, that designates the language parameter to use for Wikidata (normally, identical to the i18n one).

As my Wikidata interface is compliant with the standard reconciliation panel, there is no need for ReconFreebaseQueryPanel. I think we could delete it: I don't see the point of adapting it for Wikidata as the generic one works fine.


thadguidry:
@wetneb Let me try this out locally and review tonight. And thanks for working on those issues I brought up about the Wikidata recon service interface.

wetneb:
@thadguidry I've made a simpler version, where the settings of the endpoint are not hardcoded but fetched if no other interface is present. This seems to be more in line with the previous state of the code.

The only downside is that people upgrading to the new OpenRefine will not see the new interface if they have already added other reconciliation interfaces. And also, when changing OpenRefine's language, one needs to delete the existing service to see the new one with the updated language.

thadguidry:
@wetneb ok, let me try the simpler version.

thadguidry:
This works well Antonin.  Good Job. Minor issues for the fuzzy matching scoring on the service itself that I will address on [your project](https://github.com/wetneb/openrefine-wikidata)

thadguidry:
@wetneb Do you want or need the full bounty ?  Should it be split among others ? Magnus ?  Thoughts ?

wetneb:
Magnus does deserve a lot of bounties for his amazing work, I've reached out to him about that.

kontactone:
Unsubscribe

joewiz:
Excellent work, @wetneb - looking forward to taking this for a spin. I would be grateful for any tutorial or screencast illustrating how the new feature works.

To @thadguidry and other core developers, I'm just curious, since this plugs a big hole left by Freebase, would this open up an opportunity to do a new RC and move toward a formal release of 2.6?

thadguidry:
@joewiz Already asked Tom Morris and Jacky on our OpenRefine developers mailing list.  I'd like to see a formal release of 2.6 with Wikidata Recon.

joewiz:
@thadguidry @tfmorris @jackyq2015 Very exciting!

PR #1167, opened 2017-02-08
Freebase cleanup

wetneb:
This PR removes various Freebase-specific parts of the code and interface:
* The ReconFreebaseQueryPanel
* The dialog to add a namespaced service, as this notion was specific to Freebase
* Various parts of the standard reconciliation code that checked for Freebase ids to handle them in a specific way.

Grepping for freebase still yields some results but I'm not entirely confident about how to deal with them. Feel free to have a look. I don't think it's critical anyway…


thadguidry:
@wetneb Looks good.  And tested OK.  But just to be sure... please check on this issue https://github.com/wetneb/openrefine-wikidata/issues/6 before we go further with a merge.  I just want to make sure.

PR #1168, opened 2017-02-10
fix PR #1092 to work with Firefox

corajr:
@thadguidry Thanks for the merge! Unfortunately just afterward I noticed a problem with the cluster export functionality in Firefox. The generated link has to be added to the document body in order for the download to work -- now it works on both Firefox and Chrome.


thadguidry:
Thanks ! Merged.

PR #1178, opened 2017-03-07
Add caching in URL fetching

wetneb:
This PR introduces a cache while fetching URLs to create a new column. This cache is a simple HashMap which lives during the operation and is cleared afterwards.

Notable changes to discuss:
- [ ] Errors are also cached: if an URL errors once, it will not be retried and the same error will be stored in the table (if errors are stored). I think that's fine but we can change it if you prefer to retry.
- [ ] This changes the semantics of the throttle delay input by the user. Until now, the actual sleep time was the delay input by the user minus the time spent doing the previous request. As a consequence, if all requests take less time than the delay input by the user, the request rate was exactly determined by the delay.
With this PR, the sleep time is exactly the delay input by the user, without subtracting the time spent doing the request. This is a more responsible rate-limiting strategy because it lets servers reduce the load by taking more time to respond. The querying rate of the client adapts to the load of the server. The benefit might sound dubious but it does make a difference on servers, and it is important to spare the resources of the non-profits running free to use APIs.

This operation does not seem to be currently covered by a test case. If it were, I would have been happy to add a few regarding this new feature. I am not comfortable enough with the testing environment to build one from scratch, though.

As I wrote on the ML this dataset can be used to check the feature:

    fruit,city
    apple,milano
    pear,milano
    plum,trento
    strawberry,milano
    banana,nairobi
    kiwi,trento

Add a column by fetching URLs from the second column, with this URL expression:

`"https://www.random.org/integers/?num=1&min=1&max=100&col=1&base=10&format=plain&rnd=new&city="+value`

The resulting column should have equal values when the values of the second column are equal.

Ah, and also, sorry for removing the trailing whitespace, I hope that's not an issue.

### DIFF
--- a/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
+++ b/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
@@ -23,8 +23,8 @@ LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
 A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
 OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
 SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
-LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,           
-DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY           
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
 THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
@@ -43,6 +43,8 @@ import java.net.URLConnection;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
+import java.util.Map;
+import java.util.HashMap;
 
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -76,14 +78,14 @@ public class ColumnAdditionByFetchingURLsOperation extends EngineDependentOperat
     final protected String     _baseColumnName;
     final protected String     _urlExpression;
     final protected OnError    _onError;
-    
+
     final protected String     _newColumnName;
     final protected int        _columnInsertIndex;
     final protected int        _delay;
 
     static public AbstractOperation reconstruct(Project project, JSONObject obj) throws Exception {
         JSONObject engineConfig = obj.getJSONObject("engineConfig");
-        
+
         return new ColumnAdditionByFetchingURLsOperation(
             engineConfig,
             obj.getString("baseColumnName"),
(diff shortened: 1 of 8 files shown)


wetneb:
Actually it wasn't that hard to test so you get that too!

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
```diff
@@ -213,66 +217,89 @@ public void run() {
...
-                    }
-                }
+
+                if (_canceled) {
+                    break;
+	        }
             }
-            
+
+	    _urlCache.clear();
```
jackyq2015: clear() won't shrink the HashMap.
[WeakHashMap](https://docs.oracle.com/javase/8/docs/api/java/util/WeakHashMap.html) maybe better in this case. 
An entry in a WeakHashMap will automatically be removed when its key is no longer in ordinary use.

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
```diff
@@ -213,66 +217,89 @@ public void run() {
...
+		url = new URL(urlString);
+	    } catch (MalformedURLException e) {
+		return null;
+	    }
+
+	    Serializable cellResult = _urlCache.get(url);
+	    if (cellResult == null) {
+		cellResult = fetch(url);
+		if (cellResult != null) {
+		    _urlCache.put(url, cellResult);
```
jackyq2015: why not use the urlString as the key?

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
```diff
@@ -213,66 +217,89 @@ public void run() {
...
-                url = new URL(urlString);
-            } catch (MalformedURLException e) {
-                return null;
-            }
-            
+
+	CellAtRow cachedFetch(CellAtRow urlData) {
+	    String urlString = urlData.cell.value.toString();
+	    URL url = null;
+	    try {
```
jackyq2015: Better to keep this try block in the fetch() method. More efficient and clear.

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
wetneb: Why would it help to remove entries when the key is no longer in use? Here the keys are strings or urls that are constructed just before the request, and are no longer referenced anywhere after we continue to the next row… So if the GC is a bit aggressive, this could effectively keep the cache empty, I think.
I am not an expert but using WeakHashMap here sounds a lot like the misunderstanding described here: https://stackoverflow.com/questions/1802809/javas-weakhashmap-and-caching-why-is-it-referencing-the-keys-not-the-values#1802879

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
wetneb: I am happy with that too.

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
wetneb: Done in the next commit.

wetneb:
@jackyq2015 I've moved to Guava's cache features: https://github.com/google/guava/wiki/CachesExplained
This does not add any dependency and looks like a good compromise between simplicity and scalability.
The maximum cache size and invalidation delay are currently hard-coded, is there any sensible place where they should be defined?

thadguidry:
This is looking better, however, there will need to be a "Use Cache" checkbox option on the Fetch URLs dialog.

I'm sorry, but this is a hard requirement because many of our folks use OpenRefine for querying (polling) different kinds of networks, devices, services, sensors, etc. for research and fun and sometimes expect to get different data returned upon each successive request to the same URL.  Things like IoT sensor measurement reading, etc.

Please add a "Use Cache" checkbox option with some hover text that briefly describes what it will/won't do. 

Thanks Antonin, coming along great !

wetneb:
Yeah I guess we don't want to loose ground in the random.org fan club either! What's the point of having a power tool to work with messy data if there's no entropy to play with…

wetneb:
I've added the checkbox to the dialog to create the column, because there was some empty space where it fits nicely, I think. Whether the cache is enabled or not is stored in the JSON of the operation.
The code is retro-compatible with JSON payloads generated by previous versions: if the parameter is not present, we assume that there should not be any caching (because that's how it was before). Users can upgrade their scripts by adding "cacheResponses": true to their payloads.

thadguidry:
@wetneb Looks good and works well I think. Just tested the functionality against your random.org test urls.

@jackyq2015 any more thoughts ? good to merge ?

thadguidry:
@jackyq2015 @wetneb I think we still have an issue lurking somewhere... on repeated builds on Travis we get a failing test because of this error https://travis-ci.org/OpenRefine/OpenRefine/jobs/308507072#L1314

wetneb:
@thadguidry yeah the timeout for my test should be changed to a larger value - or make the test more clever. It can usually be solved by re-running the build. I'll fix that asap.

thadguidry:
@wetneb awesomeness. Thank you !

PR #1180, opened 2017-03-15
Fix bug: load wrong new value for RowFlagChange

xiaomj:


xiaomj:
@thadguidry hi, I come across this issue when reverting Flag or Starred rows back to normal rows and it does not work.

thadguidry:
@xiaomj Ah, your right !  Undo Redo of Flags and Stars does not work well at all !  Thanks for this fix !

Issue #1183, opened 2017-03-23
GREL OR boolean is not allowing more than two conditions


BobHarper1:
The [Wiki](https://github.com/OpenRefine/OpenRefine/wiki/GREL-Boolean-Functions) states:

> ### or(boolean b1, boolean b2, ...)
> 
> Logically OR two or more booleans to yield a boolean. For example, `or(1 < 3, 1 > 7)` returns `true` because at least one of the conditions (the first one) is true.

However, in my GREL expression where I want to use more than two (regular expression) booleans
`if(or(isNotNull(cells['column1'].value.match(/A/)),isNotNull(cells['column1'].value.match(/B/)),isNotNull(cells['column1'].value.match(/C/))),"This matches","This does not match")`

I get 

> Error: or expects two booleans

Granted that there's probably a better way to write my expression - I would have expected that `or()` would accept more than two conditions.

If I were to reduce that to only two conditions, vis:

`if(or(isNotNull(cells['column1'].value.match(/A/)),isNotNull(cells['column1'].value.match(/B/))),"This matches","This does not match")`

it does work as expected.

Either the boolean operator is not working as it should, or the documentation is wrong... or my reading of it is wrong.

BobHarper1:
Is this relevant line of code?
[com.google.refine.expr.functions.booleans.Or.java](https://github.com/OpenRefine/OpenRefine/blob/d5fb07384242d07241b0ce103c47d930046f1135/main/src/com/google/refine/expr/functions/booleans/Or.java#L49)

Not a java guy, but should it be `if (args.length >= 2 ...`??

jackyq2015:
Though or only accept 2 arguments, but you can use multiple boolean operator to achieve what you need. something like this:
`or(or(isNotNull(cells['column1'].value.match(/A/)),isNotNull(cells['column1'].value.match(/B/))),isNotNull(cells['column1'].value.match(/C/)))`

BobHarper1:
Ok, so the documentation needs to change then.

jackyq2015:
Wiki mentioned above fixed.

PR #1189, opened 2017-04-26
allow more than 2 AND and OR conditions

BobHarper1:
In reference to my issue #1183 (which was resolved by editing the Wiki to reflect that `AND`/`OR` operators accept *only* two conditions).

I believe it is possible to allow both `AND`/`OR` allow *two or more* conditions, as the changes in my request establish. 

This is an improvement - there is no need to nest `AND`/`OR` functions within each other, and makes testing more than two conditions simpler and more intuitive. It also matches the functionality of the comparable functions in Excel which allow 2+ arguments.


jackyq2015:
please fix the CI errors

BobHarper1:
Also added multiple params to `xor` to be consistent (see http://stackoverflow.com/questions/6222162/in-java-xor-with-three-true-inputs-returns-true-why). I've [updated Wiki](https://github.com/OpenRefine/OpenRefine/wiki/GREL-Boolean-Functions) in anticipation!

PR #1191, opened 2017-05-05
Fixes #1181

### LINKED ISSUE #1181: "Store file source" doesn't work for multiple sheets in Excel file
**System:**

OpenRefine Version 2.6-rc.2 [TRUNK]
Google Chrome Version 56.0.2924.87 (64-bit)
Mac OSX 10.11.6

**Issue:**

I have a .xlsx file with about 8 sheets, each of which has a few hundred rows. I want to import all the sheets to a single project, including the file/sheet source as a column. When I check "store file source", only the source of the first sheet whose box I check will be stored - all of the others will have a "source" column set to blank.

screen shot 2017-03-16 at 10 01 26 am

**Steps:**

1. Click "create project" and choose .xlsx source file with multiple sheets. Click "next".
2. Uncheck "parse column headers", as my sheets have none.
3. Check "store file source in each row".
4. Begin checking boxes for each sheet I want to include.

**Current results:**

The "store file source" option works fine for the first sheet (it shows `my_file.xlsx#sheet_n`) but no information shows in the preview for any of the other sheets (the source column is blank). Updating the preview does nothing, and when I create the project the results are the same as the preview. If I uncheck all of the sheets, then begin checking them again, *only the sheet that I check first* - regardless of which sheet it is - will have its source stored. All of the rest will be blank.

**Expected results:**

Each sheet would have an entry in the source column similar to `my_file.xslx#sheet_1`, so that I could extract the `sheet_1` information and use it.

adieyal:
The check for a duplicate column was previously being done with an assert which is switched off by default. As a result, tests were always passing.

### DIFF
--- a/main/src/com/google/refine/importers/ImportingParserBase.java
+++ b/main/src/com/google/refine/importers/ImportingParserBase.java
@@ -164,16 +164,20 @@ abstract public class ImportingParserBase implements ImportingParser {
     
     protected static int addFilenameColumn(Project project) {
         String fileNameColumnName = "File";
-        assert project.columnModel.getColumnByName(fileNameColumnName) == null;
-        try {
-            project.columnModel.addColumn(
-                0, new Column(project.columnModel.allocateNewCellIndex(), fileNameColumnName), false);
+        if (project.columnModel.getColumnByName(fileNameColumnName) == null) {
+            try {
+                project.columnModel.addColumn(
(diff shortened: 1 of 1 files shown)


jackyq2015:
I mean for #1181, can you import a multi-sheet xls and follow the procedure to verify that your fix works?

adieyal:
Yep, works fine now. The problem before was that the check for a duplicate name was done using an assert which obviously needs to be explicitly enabled at runtime. I replaced it with and if statement

PR #1208, opened 2017-07-14
Added Start and Stop Script For Linux

shubhra02:
The script to start OpenRefine is "start.sh". It starts OpenRefine as a background process and logs are created respectively in "open-refine.log" file.
The script to stop OpenRefine is "stop.sh". It kills the OpenRefine process and displays a message like "successfully stopped open-refine".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ broker/core/test-output/
 tmp/
 /test-output
 /bin
+open-refine.log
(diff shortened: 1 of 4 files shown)


charmygarg:
We have updated the script now the user will also have the choice to add their own IP address otherwise default [127.0.0.1] will be used

wetneb:
It might be worth adding open-refine.log to .gitignore to prevent this :)

charmygarg:
We have removed open-refine.log and added entry for the same in .gitignore. Let me know if anything is needed to be changed.
Thanks

wetneb:
@charmygarg thanks a lot! it looks good to me.

shubhra02:
Thanks a lot @Thad

Issue #1179, opened 2017-03-11
Add column from Wikidata


wetneb:
The Freebase extension for OpenRefine had a nice feature: once you had reconciled a column to Freebase, you could fetch data from Freebase using the properties associated to the items.

https://youtu.be/5tsyz3ibYzk

It is currently possible to do that for Wikidata, using the "Add column by fetching URLs" feature. The reconciliation endpoint provides an (undocumented) API to help you do that.

https://tools.wmflabs.org/openrefine-wikidata/en/fetch_values?item=Q1377&prop=P856

With the following parameters:
- item=Q1377 gives the item to fetch the value from
- prop=P856 gives the property storing the value
- flat=true returns the plain value instead of a JSON payload
- label=false can be used when the property points to an item and we want to retrieve the identifier instead of the label

This works well, but is clearly not as user-friendly as the Freebase interface! So I see two options:

- either we migrate the freebase extension to Wikidata

- or we recognize that fetching data associated with ids is a fairly generic use case, so we augment the reconciliation API with an additional endpoint to do that. This would enable other reconciliation endpoints to implement the feature and have it nicely integrated in OpenRefine.

What do you think?

P.S: bounties welcome!

magdmartin:
I vote for option 2 make it a generic use case. There is a lot of
reconciliation endpoint out there and I think the community will greatly
appreciate it.



2017-03-10 19:14 GMT-05:00 Antonin Delpeuch <notifications@github.com>:

thadguidry:
@wetneb yes, option 2.  Although, I'm not sure if some of the older recon services, like VIAF, still provide the service or not (but it shouldn't stop our wanted position to extend a users data with additional data.  https://github.com/OpenRefine/OpenRefine/wiki/Reconciliation-Service-API#examples

Just to make sure you looking at the right places also Antonin...

2.6-rc1 was the last time we had freebase extension fully included here: https://github.com/OpenRefine/OpenRefine/tree/v2.6-rc1/extensions
and the operation was called ExtendDataOperation.java btw

I snipped the following from our wiki as a bit of provenance as well... https://github.com/OpenRefine/OpenRefine/wiki/Reconciliation-Service-API

-- snip --
Because the reconciliation service only returns a single ID, getting additional column values could be done one of two ways:

1. The Freebase way using a separate command a la "Add column by fetching from Freebase" (perhaps just using "Add column by fetching URL" against a local REST service)

[...]

PR #1210, opened 2017-07-16
Add data extension capabilities to the reconciliation API

wetneb:
This PR adds support for a data extension protocol, integrated in the reconciliation API and described at 
https://github.com/OpenRefine/OpenRefine/wiki/Data-Extension-API . It is largely based on the Freebase extension.

This PR is meant to foster discussion around the protocol specification draft, so that we can have something concrete to look at. I anticipate some aspects of the API will change, so providers and consumers should not rush to implement compliance with it.

The Wikidata endpoint supports the API as it is currently drafted. To try the feature, you need to delete your current Wikidata reconciliation endpoint in your list of reconciliation services, and add it again manually with `https://tools.wmflabs.org/openrefine-wikidata/en/api`. This is because the service metadata has changed (and is still likely to change as we improve the API). Changes to this service metadata will not affect users of stable OpenRefine versions.

Related to issue #1179 . Pinging potentially interested contributors: @magdmartin @codeforkjeff @thadguidry @jackyq2015 @nickynicolson.

Screencast


thadguidry:
Looks good enough for now and to let folks build and play and test more.  I'll merge this in knowing that the API can change and you'll have more PR's :)  Great Job again @wetneb

codeforkjeff:
Thanks for the ping, I haven't forgotten about this, just trying to find time to digest this extension and get a working test implementation in conciliator. Soon, I hope.

PR #1233, opened 2017-08-13
Small fixes to the data extension interface

wetneb:
- removed mention to a missing .less file
- fixed the position of the Reset button in the preview dialog
- fixed the error for empty property selections


wetneb:
@codacy-bot I don't think we are going to be friends for a long time…

Issue #1106, opened 2016-01-16
Undo/Redo sorting order


Opennat:
In 2.6-rc.2 (probably in 2.6-rc.1 too), default Undo / Redo list sorting was changed, but it was more usable in previous versions (newest changes on top of the list).
1. Now you should scroll down to the end of the list (it could be very big)
2. Last action in some cases (i.e. when last action was transform with long regexp string) does not shown correctly (It hides behind browsers' bottom line), and you should spend some time for check to be sure that is really last action you trying to revert.

I'll make screenshot later to describe (2) visually.

wetneb:
@OpenRefine/core I agree with @Opennat, I have run into the same problem myself. I think displaying the last operation at the top makes more sense. Should we revert that before the next release?

ostephens:
This is marked as closed but I can't see the original issue (sort order of operation history) is addressed? But also can't see that the order of the history panel has ever changed. Am I missing something?

wetneb:
@ostephens indeed, only point 2. was solved.
I remember looking for the change of order and I have not found it either. I think it would be a sensible change, feel free to reopen.

PR #1251, opened 2017-09-20
Fix overflow of history items outside the window.

wetneb:
This fixes the auto-resize code which ensures that history items don't overflow beyond the end of the window. The order of the history items is still preserved (oldest first). According to @Opennat in #1106 it used to be in the reverse order, but I can't find the commit that changed that. I think the current order is less of a problem now that this bug is fixed.


wetneb:
Everything is green, let's merge!

thadguidry:
@wetneb lol, crazy guy... yeah this looks good.  Thanks for the fix man, this bug has also upset me for a while.

Issue #499, opened 2012-10-15
Create architecture for several metadata layers


tfmorris:
_Original author: dfhu...@gmail.com (November 29, 2011 20:05:23)_

Currently there are (conceptually) 2 metadata layers, one for reconciliation metadata that is stored within the Cell objects, and one for storing row dependencies within records, modeled in the RecordModel. These layers should be generalized.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=499_

wetneb:
One example of possible generalization of the reconciliation metadata would be to add provenance information to cells (references). This would be useful for #1243.

PR #1254, opened 2017-09-23
Add support for references in Wikitable importer.

### LINKED ISSUE #1243: Wikitext importer: support for references
It would be awesome if the wikitext importer could somehow extract the references in each cell. The problem is that the format of the references vary greatly, but I think a good start would just be to extract the first URL contained in the `<ref></ref>`. This would miss some references (for instance references which don't have any URLs but have a DOI) but would still have a decent coverage.

Concerning storage in OpenRefine's model, the ideal solution would be to have a generic way to add this information to cells in a separate field (much like the reconciliation storage) but that sounds quite hard. One simpler way would be to add them in separate colums (with the appropriate column groups in the records model).

wetneb:
This lets users import the individual references of each cell. As OpenRefine does not support adding arbitrary metadata to cells yet (see #499), references are added alongside the values they support, in their own column.

Closes #1243.

### DIFF
--- a/main/src/com/google/refine/importers/WikitextImporter.java
+++ b/main/src/com/google/refine/importers/WikitextImporter.java
@@ -3,7 +3,11 @@ package com.google.refine.importers;
 import java.io.IOException;
 import java.io.Reader;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


wetneb:
thank you @codacy-bot for being cooperative tonight.

Issue #1258, opened 2017-09-27
Notify user if a more recent release is available


ostephens:
As per https://groups.google.com/forum/#!topic/openrefine-dev/AoLLeS7j8UE

@thadguidry said:

> What is much more important is within OpenRefine letting users know about a new update.  One way we talked about would be to just provide an update notifier link "Get newest OpenRefine 2.7.1 here" or somesuch near our logo in OpenRefine that provides a link to the releases page on our Github.  To implement perhaps that could be just a bit of javascript to check our https://github.com/OpenRefine/OpenRefine/releases/latest Github page and compare to the version running and make the update link visible ?

ostephens:
I've been having a look at this, and there is some old code in /main/webapp/modules/core/scripts/index.js that does something very similar, but checking Google Code site. If it finds a more recent release it puts a notification at the top of the page.

I've been playing around with modifying it so uses https://api.github.com/repos/openrefine/openrefine/releases/latest to check the latest version and that's straightforward.

I'm assuming we use the 'tag_name' from the Github JSON to get the latest version number - is that a safe assumption (that the tag_name will reflect the version number)?

In terms of comparing versions, should I assume semver for versioning?

wetneb:
According to GitHub

>  It’s common practice to prefix your version names with the letter v. Some good tag names might be v1.0 or v2.3.4.

So I would recommend we follow this practice… (this lets us define tags which are not releases).

ostephens:
OK - that sounds reasonable.
(n.b. you can have tags that aren't releases anyway - a tag isn't a release by default. The Github /releases/latest api call only returns full releases - not draft releases or prereleases).

ostephens:
I've made an attempt at fixing this. It assumes Semantic Version numbering format of form X.Y.Z where X, Y, and Z are non-negative integer. It ignores any trailing letters e.g. 2.7.1-rc.1 will be interpreted as equivalent to 2.7.1

Comments on PR welcome if this could be approached in a better/more robust way

thadguidry:
@wetneb FYI, Most of the open source world has aligned to GAVC convention borrowed from Maven.

Group:Artifact,:Version:Classifier.
org.openrefine:OpenRefine:2.7.1:SNAPSHOT

and filenames look like
openrefine-2.7.1-SNAPSHOT.zip
openrefine-2.7.1-rc1.zip

Typically, tag names are just the VC parts ... version and classifier

2.7.1-SNAPSHOT
[...]

PR #1259, opened 2017-09-27
Alert user if more recent version available from github

ostephens:
Addressing #1258

### DIFF
--- a/main/webapp/modules/core/scripts/index.js
+++ b/main/webapp/modules/core/scripts/index.js
@@ -79,13 +79,14 @@ $(function() {
     while(thisVerParts.length < 3)
       thisVerParts.push(0);
     while(latestVerParts.length < 3)
-      latestVersParts.push(0);
+      latestVerParts.push(0);
 
     for(var i=0; i<3; i++) {
-      console.log(parseInt(latestVerParts[i],10));
-      if(parseInt(thisVerParts[i],10) == parseInt(latestVerParts[i],10)) {
+      var thisVerPartInt = parseInt(thisVerParts[i],10);
+      var latestVerPartInt = parseInt(latestVerParts[i],10);
+      if(thisVerPartInt == latestVerPartInt) {
         continue;
-      } else if (parseInt(thisVerParts[i],10) > parseInt(latestVerParts[i],10)) {
+      } else if (thisVerPartInt > latestVerPartInt) {
         return false;
       } else {
         return true;
@@ -104,7 +105,7 @@ $(function() {
           $("#openrefine-version").text($.i18n._('core-index')["version"]+" " + OpenRefineVersion.full_version);
           
 
-            $.getJSON( "https://api.github.com/repos/openrefine/openrefine/releases/latest",
+            $.getJSON("https://api.github.com/repos/openrefine/openrefine/releases/latest",
              function( data ) {
               var latestVersion = data.tag_name;
               var latestVersionName = data.name;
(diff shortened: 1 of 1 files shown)


main/webapp/modules/core/scripts/index.js
```diff
@@ -68,28 +68,30 @@ Refine.selectActionArea = function(id) {
...
-    return latestRev > thisRev;
+    var thisVerParts = thisVer.split(".");
+    var latestVerParts = latestVer.split(".");
+    while(thisVerParts.length < 3)
+      thisVerParts.push(0);
+    while(latestVerParts.length < 3)
+      latestVersParts.push(0);
+
+    for(var i=0; i<3; i++) {
+      console.log(parseInt(latestVerParts[i],10));
```
wetneb: We should remove this `console.log` before merging.

main/webapp/modules/core/scripts/index.js
```diff
@@ -68,28 +68,30 @@ Refine.selectActionArea = function(id) {
...
+      if(parseInt(thisVerParts[i],10) == parseInt(latestVerParts[i],10)) {
```
wetneb: Why not store `parseInt(thisVerParts[i],10)` in a variable and do the comparisons on that variable? Same for `parseInt(latestVerParts[i],10)`, it would look a bit cleaner to parse the string only once.

main/webapp/modules/core/scripts/index.js
```diff
@@ -68,28 +68,30 @@ Refine.selectActionArea = function(id) {
...
-    // Parse version from dot separated string into array of integers
-    
-    // compare left to right, including suffix
-    return latestRev > thisRev;
+    var thisVerParts = thisVer.split(".");
+    var latestVerParts = latestVer.split(".");
+    while(thisVerParts.length < 3)
+      thisVerParts.push(0);
+    while(latestVerParts.length < 3)
+      latestVersParts.push(0);
```
wetneb: This should be `latestVerParts` (no `s`). Have you tested your code?

ostephens: Odd - I did test. Thanks for the catch

main/webapp/modules/core/scripts/index.js
ostephens: doh!

main/webapp/modules/core/scripts/index.js
ostephens: OK

main/webapp/modules/core/scripts/index.js
```diff
@@ -68,28 +68,31 @@ Refine.selectActionArea = function(id) {
...
+    for(var i=0; i<3; i++) {
+      var thisVerPartInt = parseInt(thisVerParts[i],10);
+      var latestVerPartInt = parseInt(latestVerParts[i],10);
+      if(thisVerPartInt == latestVerPartInt) {
+        continue;
+      } else if (thisVerPartInt > latestVerPartInt) {
+        return false;
+      } else {
+        return true;
+      }
```
wetneb: Things of the form `if foo then return false else return true` can be rewritten as `return not foo`.

wetneb:
Also, I wonder if the download link should open in a new tab. I think I was expecting that when I clicked the link. Any thoughts?

main/webapp/modules/core/scripts/index.js
ostephens: Thanks - will re-write that.
In terms of opening link in new window, I didn't change the notification code, so it is just following its previous behaviour. I think opening in a new window makes sense - I'll amend

thadguidry:
Good Job Guys !  :)

Issue #1247, opened 2017-09-13
Range facet to/from is incorrect for Reconciling "Best candidate's score"


thadguidry:
From Pull #1246

When adding a "Best candidate's score" facet to a column with equal reconciliation scores, I get the following facet JSON:

 {
          "to": 1,
          "expression": "cell.recon.best.score",
          "selectError": false,
          "selectNumeric": true,
          "selectBlank": true,
          "name": "occupation: best candidate's score",
          "columnName": "occupation",
          "selectNonNumeric": true,
          "from": 100,
          "type": "range"
        }
I think the problem is rather in the UI: the facet selection UI looks buggy in that case. (In the JSON above, the "to": 1 is wrong).

Anyway, I think it would be quite dangerous to change the semantics of the range facets: this can potentially break a lot of user scripts. We just need to make sure that OpenRefine generates the right facet (with min=score and max=score+1) by default.

wetneb:
Closed by #1263.

PR #1263, opened 2017-10-09
Ensure _max is never less than _min for numeric range facet

ostephens:
If there is only a single numeric value for a numeric range facet, the 'max' value for the facet is set to '1' ( actually _step, but _step is hard coded to be '1'). This leads to the behaviour reported in #1247 (for any situation where all numeric values are equal, not just in the "Best candidate's score" scenario described)

This change ensures that the max value is min+step in this scenario - i.e. one more than 'min' in the case where 'step' is hard coded to '1'


wetneb:
@ostephens thanks for looking into that! Your fix looks perfect to me.

Issue #1279, opened 2017-10-23
Use best practices in i18n


remram44:
I applaud your attempt at internationalizing OpenRefine, however you did not read the manual. As you know other languages have their own grammar and using the English grammar on all languages is obviously not going to work:

https://github.com/OpenRefine/OpenRefine/blob/e5fdf48680a6f810c84706671344dc93d9a4fb75/main/webapp/modules/core/scripts/dialogs/clustering-dialog.js#L245

The jquery.i18n library has a lot of awesome features inherited from gettext such as string interpolation (and even custom plurals). This is not the only location in the code where messages are made up from pieces using concatenation, please use interpolation instead so that languages that use a different word order don't look broken.

Other example:

https://github.com/OpenRefine/OpenRefine/blob/e5fdf48680a6f810c84706671344dc93d9a4fb75/main/webapp/modules/core/scripts/preferences.js#L86

Please just use a construct like:

```javascript
if (window.confirm($.i18n('confirm-delete-key', key))) { 
```

It also seems that you added sections in your translation JSON, placing JSON objects where messages should be, which is probably going to break all the machinery (and is responsible for the awkward `i18n(...)[...]`). Please read library's manuals or at least [their repository's README before proceeding](https://github.com/wikimedia/jquery.i18n#message-format).

wetneb:
@remram44 I completely agree! Pull requests are more than welcome.

remram44:
Fixing this would mean all translation files have to be adapted (though I suspect the program is not very usable in non-English locales anyway), so a commitment from the project leads is necessary before work can happen here.

wetneb:
@remram44 The project leads (if there is such a thing) do not master all the languages that we currently support, so we cannot fix these translations ourselves… So I don't really see what you mean! What "commitment from project leads" do you expect? The people who wrote the code that you quoted are not active in OpenRefine anymore.
Feel free to invalidate existing translations, people will go to Weblate to fix them if they use the translation.

remram44:
> Feel free to invalidate existing translations, people will go to Weblate to fix them if they use the translation.

Sounds good! I just don't want to do work to then hear that "that would break translations" and have the patch rejected. I will see if I can do that but I'm short on time at the moment, so if somebody else wants to grab this go ahead.

remram44:
[...]

PR #1280, opened 2017-10-23
Fix some French translations

remram44:
I noticed some typos so I thought I'd go over the translation. I gave up on going over the rest of this because of #1279 (I have to grep every bit and pieces, and there is no way to make some of those make sense when formatted in the English order).

However here are the obvious ones.


wetneb:
@remram44 Feel free to use Weblate for that (so that you will not even have to file a PR).
https://hosted.weblate.org/projects/openrefine/

PR #1282, opened 2017-10-23
Add copyright header for i18n library

remram44:
MIT is one of the most permissive licenses, it has only one requirement: the copyright header has to be kept on source reproductions of the work.

Let's add this copyright.

Amends 817c6cc8d4468c5bf8e776cbe2f384385e74bf69, cc @Blakko


wetneb:
Thanks!

PR #1283, opened 2017-10-23
Remove jQuery 1.9.1, unused

remram44:
You have both 1.9.1 and 1.11.1 in there, but only the latter is loaded.


wetneb:
I confirm this was unused. Thanks!

PR #1285, opened 2017-10-23
WIP: Move to jquery.i18n (from wikipedia)

### LINKED ISSUE #1279: Use best practices in i18n

remram44:
This attempts to move to [wikipedia's jquery.i18n](https://github.com/wikimedia/jquery.i18n) so we can use modern features such as interpolation and plurals.

Currently WIP, don't merge.

Fixes #1279


remram44:
It works, but I haven't gone over the extensions. In particular, do we want some kind of namespacing for the string IDs? Such as `core-discard`, `gdata-discard`, etc.

remram44:
Can someone configure @codacy-bot to ignore external libs?

thadguidry:
@remram44 yeah, on it now :)  5 mins :)

thadguidry:
@remram44 Ignored all under /main/webapp/modules/core/externals/

wetneb:
@remram44 I think prefixing the string ids with the module name would make sense indeed!

remram44:
I will do that then: `core-msg`, `gdata-msg`. Thanks!

ostephens:
@remram44 Is there anything I can do to help progress this work?

remram44:
I've been really busy but I'll come back to this soon! Sorry about the delay.

ostephens:
@remram44 no problem at all! V grateful for your work - just shout if there is anything I can do to help

thadguidry:
@remram44 Hey Remi ! Just checking in on your progress.

remram44:
I can update this patch for current OpenRefine. Note that once this is merged, translation files will have to be updated.

wetneb:
@remram44 that would be fantastic! But I was thinking, maybe it would be better to avoid collisions of translation strings, by adding the namespacing mentioned above?

`$.i18n._('core-dialogs')["content"]` -> `$.i18n._('core-dialogs_content')`

instead of

`$.i18n._('core-dialogs')["content"]` -> `$.i18n._('content')` which looks more risky (especially because we have extensions)!

Also given that we have had quite a lot of changes since this PR was started, it might be easier to restart it from the current master branch.

I can with any of this - you already made a great contribution by pointing us to the right library and good practices :)

wetneb:
@remram44 As this PR is quite old now, I'm taking it over and restart it from the current master. I'll try to reuse your commits where possible to preserve attribution.

remram44:
Thanks @wetneb! Sorry I couldn't find time to finish this!

PR #1290, opened 2017-10-26
Extend cross() function to support multiple-value-cell-input

claussni:
Resolves #1289

Issue #1289, opened 2017-10-26
Extend cross() function to support multiple-value-cell-input


felixlohmeier:
From a user perspective, I would like to call the cross function with a separator as 4th argument, so that multiple values (separated by a custom separator, e.g. ␟) in an input cell get "crossed".

### project "Christmas Gifts"

| gift | recipient |
| ------------- | ------------- |
| lamp  | mary  |
| clock  | john  |
| sweets  | mary,john  |
| wine  | anne,mary  |

### project "My Address Book"

| friend | address |
| ------------- | ------------- |
| john  | 120 Main St.  |
| mary  | 50 Broadway Ave.  |
| anne  | 17 Morning Crescent |

### Transformation

add values from column "address" from project "My Address Book" based on keys (with multiple values separated by ",") in column "recipient" from project "Christmas Gifts"

`forEach(value.cross("My Address Book", "friend", ","),r,r.cells["address"].value).join(",")`

### Output

| gift | recipient | address |
| ------------- | ------------- | ------------- |
| lamp  | mary  | 50 Broadway Ave. |
| clock  | john  | 120 Main St. |
| sweets  | mary,john  | 50 Broadway Ave.,120 Main St.
| wine  | anne,mary  | 17 Morning Crescent, 50 Broadway Ave.

## History

see last few comments in #1204

felixlohmeier:
[example-data-1289.zip](https://github.com/OpenRefine/OpenRefine/files/1419178/example-data-1289.zip)

ostephens:
An issue here is that you cannot tell from the results which value resulted in the listed outcome. To take the simplest example:
### project "Christmas Gifts"

| gift | recipient |
| ------------- | ------------- |
| lamp  | mary  |
| clock  | john  |
| sweets  | mary,john  |
| wine  | anne,mary  |

### project "My Address Book"

| friend | address |
| ------------- | ------------- |
| john  | 120 Main St.  |
| mary  | 50 Broadway Ave.  |

### Transformation

add values from column "address" from project "My Address Book" based on keys (with multiple values separated by ",") in column "recipient" from project "Christmas Gifts"

`forEach(value.cross("My Address Book", "friend", ","),r,r.cells["address"].value).join(",")`

### Output

| gift | recipient | address |
| ------------- | ------------- | ------------- |
| lamp  | mary  | 50 Broadway Ave. |
| clock  | john  | 120 Main St. |
[...]

PR #1292, opened 2017-10-27
Extend cross function

wetneb:
This is a new version of #1290.

### DIFF
--- a/main/src/com/google/refine/InterProjectModel.java
+++ b/main/src/com/google/refine/InterProjectModel.java
@@ -69,7 +69,7 @@ public class InterProjectModel {
             this.toProjectColumnName = toProjectColumnName;
         }
 
-        public HasFieldsListImpl getRows(final Object rowKey) {
+        public HasFieldsListImpl getRows(final Object rowKey, String separatorRegexp) {
             Project toProject = ProjectManager.singleton.getProject(toProjectID);
             if (toProject == null) {
                 return null;
@@ -79,8 +79,8 @@ public class InterProjectModel {
 
             if (ExpressionUtils.isNonBlankData(rowKey)) {
                 Object[] rowKeys;
-                if (rowKey instanceof String) {
-                    rowKeys = ((String) rowKey).split(",");
+                if (separatorRegexp != null && !separatorRegexp.isEmpty() && rowKey instanceof String) {
+                    rowKeys = ((String) rowKey).split(separatorRegexp);
                 } else {
                     rowKeys = new Object[]{rowKey};
                 }
@@ -111,9 +111,10 @@ public class InterProjectModel {
      * @param fromColumn
      * @param toProject
      * @param toColumn
+     * @param separatorRegexp
      * @return
      */
-    public ProjectJoin getJoin(String fromProject, String fromColumn, String toProject, String toColumn) {
+    public ProjectJoin getJoin(String fromProject, String fromColumn, String toProject, String toColumn, String separatorRegexp) {
         String key = fromProject + ";" + fromColumn + ";" + toProject + ";" + toColumn;
         if (!_joins.containsKey(key)) {
             ProjectJoin join = new ProjectJoin(
@@ -122,8 +123,8 @@ public class InterProjectModel {
                 ProjectManager.singleton.getProjectID(toProject), 
                 toColumn
             );
-            
-            computeJoin(join);
+
+            computeJoin(join, separatorRegexp);
             
             synchronized (_joins) {
                 _joins.put(key, join);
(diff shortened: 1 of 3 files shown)


main/src/com/google/refine/InterProjectModel.java
```diff
@@ -158,11 +175,21 @@ protected void computeJoin(ProjectJoin join) {
...
+
         for (Row fromRow : fromProject.rows) {
-            Object value = fromRow.getCellValue(fromColumn.getCellIndex());
-            if (ExpressionUtils.isNonBlankData(value) && !join.valueToRowIndices.containsKey(value)) {
-                join.valueToRowIndices.put(value, new ArrayList<Integer>());
+            Object fromRowKey = fromRow.getCellValue(fromColumn.getCellIndex());
+            if (ExpressionUtils.isNonBlankData(fromRowKey)) {
+                Object[] fromRowKeys;
+                if (separatorRegexp != null && !separatorRegexp.isEmpty() && fromRowKey instanceof String) {
+                    fromRowKeys = ((String) fromRowKey).split(separatorRegexp);
```
wetneb: I think it would be more efficient to compile this regexp to a `Pattern` only once, outside the loop, and re-use that inside the loop.
https://stackoverflow.com/questions/1720191/java-util-regex-importance-of-pattern-compile#1721778

main/src/com/google/refine/expr/functions/Cross.java
```diff
@@ -47,33 +47,39 @@
...
-    
+
+    public static final String EVAL_ERROR_MESSAGE =
+            " expects a string or cell, a project name to join with, and a column name in that project. " +
+                    "Optional accepts a regular expression separator for source values.";
+
     @Override
     public Object call(Properties bindings, Object[] args) {
-        if (args.length == 3) {
+        if (args.length >= 3) {
```
wetneb: Calling `cross` with more than 4 arguments should fail with an error, it looks like it currently does not.

main/tests/server/src/com/google/refine/tests/expr/functions/CrossFunctionTests.java
```diff
@@ -9,6 +9,7 @@
 import java.util.List;
 import java.util.Properties;
```
wetneb: Thanks for updating the existing tests! It would be great if we could also have one or two tests to demonstrate the functionality that you have added.

wetneb:
@claussni I realize you don't have push access to this repository so you won't be able to fix the remaining issues in this PR… Sorry for this mess! It looks like we need to create yet another PR from your own fork. Let me know if you need git help to set that up.

claussni:
@wetneb I started a new fork and branch, gradually fixing the issues you raised: https://github.com/claussni/OpenRefine/tree/cross-func-split-2

I'm currently trying to improve unit test coverage for the new use case - which appears to be more challenging than the actual fix. @felixlohmeier provided more sample data for testing.

A new PR is coming up and we can continue the discussion there, if that is ok?

wetneb:
@claussni yes let's do that, thanks a lot!

ostephens:
See my latest comment on #1289 - not sure just amending the cross function is going to be sufficient

main/src/com/google/refine/expr/functions/Cross.java
thadguidry: I saw that also after the revert and agree, let's fail with error on greater than 4 arguments as well as having a test for when there are missing arguments... in other words, when there are arguments 1,2,null,4 it should throw error as we currently do.

wetneb:
@msaby this PR has actually never been merged as such - see the discussions above.

msaby:
thanks @wetneb 
I see it is a complex issue...

felixlohmeier:
A fork with the extended cross function is still available here: https://github.com/opencultureconsulting/OpenRefine.

There is a release based on a development snapshot at October 28, 2017 (i.e. something between OpenRefine 2.7 and 2.8): https://github.com/opencultureconsulting/OpenRefine/releases/tag/2017-10-28-with-pr1294. I am currently using this version intensively in an ongoing project for the German Literature Archive (slides: https://doi.org/10.5281/zenodo.2678112).

It should work with OpenRefine 3.0, but I haven't tested it yet. Comparing changes based on OpenRefine release tag 3.0: https://github.com/OpenRefine/OpenRefine/compare/3.0...opencultureconsulting:master

Because there have been changes to the files for the cross function (Cross.java) and the interproject model (InterProjectModel.java) in the meantime, automatic merging with OpenRefine >3.0 is no longer possible.

felixlohmeier:
I forgot to mention that there is also a [docker container](https://hub.docker.com/r/felixlohmeier/openrefine) available:
```
docker run -p 3333:3333 felixlohmeier/openrefine:2017-10-28-with-pr1294
```

PR #1293, opened 2017-10-27
Fix regex filter error

### LINKED ISSUE #1203: Filter by regex - grey window spinning wheel on version 2.7
Hi,
I'm using the 2.7 version, with linux (Ubuntu 17.04), Firefox.
My dataset contains +/- 1200 rows and 20 columns.

If I ask a column to be filtered by a regex, and once refine starts to interpret the regex, The page turns grey and I cannot interact with my dataset without having to reload the page.  

Can you repeat this bug on your machines? Or is it just me?

ostephens:
To fix #1203 
- Changes get-rows call to request JSON rather than JSONP
- Catches situations where JSON returned for get-rows or compute-facets includes an 'error' element
  - Where there is an error with get-rows, the existing row set is used for the data-table
  - Where there is an error with compute-facets, the error message from the JSON is displayed at the top of the facets/filter sidebar (some minor modifications to css and DOM manipulation were needed to achieve this alongside the 'refreshing facets' notification)

Comments welcome

### DIFF
--- a/main/webapp/modules/core/scripts/project/browsing-engine.js
+++ b/main/webapp/modules/core/scripts/project/browsing-engine.js
@@ -33,7 +33,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 function BrowsingEngine(div, facetConfigs) {
   this._div = div;
-  this._mode = theProject.recordModel.hasRecords ? 'record-based' : 'row-based';
+  this._mode = theProject.recordModel.hasRecords ? "record-based" : "row-based";
 
   this._facets = [];
   this._initializeUI();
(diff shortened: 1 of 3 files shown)


main/webapp/modules/core/scripts/project/browsing-engine.js
```diff
@@ -240,19 +241,31 @@ BrowsingEngine.prototype.update = function(onDone) {
...
+        var clearErr = $('#err-text').remove();
+        var err = $('<div id="err-text">')
+                  .text(data.message)
+                  .appendTo(self._elmts.errors);
+         self._elmts.errors.css("display", "block");
+        if (onDone) {
+          onDone();
+        }
+        return;
+      }
```
wetneb: In principle I'd prefer an `if () { ... } else { ... }` rather than `if () { ... ; return; } ...` but it's clearly not a big deal…

thadguidry:
@ostephens same look and feel in Firefox and Chrome ?

thadguidry:
@ostephens Just a minor training point.  In Java and Javascript, there are standard equality operands and STRICT equality operands.  One does the type conversion AND THEN does the equality check, and the other does not do type conversion, I.E. it is STRICT.  If your comparing identity, you want to ensure your using STRICT operands (also known as the identity compare operands) (===, !==) Always use the bible reference with a handy chart at the bottom https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators

In Java, its common practice for String comparisons to use .equals() rather than === , but its a preference and the later only available in later JDK's.

Your doing GREAT however, Owen, and really like these constant commits from you !  Keep up the learning !

jackyq2015:
looks good for me. will merge.

ostephens:
Thanks all. @thadguidry yes have checked it in Safari, FF & Chrome. 

Thanks to you all for guidance & encouragement :)

PR #1300, opened 2017-11-02
Add Java 9 to Travis

wetneb:
Following the changes made by @jackyq2015 to adapt the script for Java 9 I have tried to enable Java 9 in Travis. It looks like one tests fails with this Java version, due to a different behavior in date parsing. I'm just submitting this PR so that people are aware of the issue. It should of course only be merged once we find a fix for that, and we should not claim that OR runs with Java 9 before that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: java
 jdk:
   - oraclejdk8
+  - oraclejdk9
   - openjdk7
+  - openjdk8
 
 env:
   # encrypted Codacy key, see https://docs.travis-ci.com/user/encryption-keys/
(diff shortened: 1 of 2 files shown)


jackyq2015:
we need to fix those failed test cases.

olleolleolle:
In order to allow the reflective access, some Java 9-specific settings need to be applied as JAVA_OPTS.

```
   [testng] WARNING: An illegal reflective access operation has occurred
   [testng] WARNING: Illegal reflective access by com.google.refine.expr.util.CalendarParser (file:/home/travis/build/OpenRefine/OpenRefine/main/webapp/WEB-INF/classes/) to method sun.util.calendar.ZoneInfo.getAvailableIDs()
```

To hint better at what I mean, here's one `--add-opens` example for some other module names:

```
export JAVA_OPTS='--add-opens java.base/java.util.zip=ALL-UNNAMED --add-opens java.base/java.security.cert=ALL-UNNAMED --add-opens java.base/java.security=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED'
```

jackyq2015:
The UrlFetchingTests.java has been fixed. it is not about how long to sleep. The web site block the appvevor. we have to skip it when it happens. I will close this one. Maybe you can submit another one to add the options in the script

wetneb:
@jackyq2015 this PR was not about this test case, it is about the date parsing issue. I have reopened it at #1427.

PR #1301, opened 2017-11-03
metadata for project

jackyq2015:
The columnar metadata will be in another PR. Hopefully can be released with 2.7 RC2


main/src/com/google/refine/ProjectMetadata.java
```diff
@@ -263,4 +329,110 @@ public void setCustomMetadata(String key, Serializable value) {
...
+    
+    public void setDescription(String description) {
+        this._description = description;
+        updateModified();
+    }
+
+    
+    public int getRowNumber() {
+        return _rowNumber;
+    }
```
wetneb: What is this `rowNumber` metadata? If that counts the number of rows in the project, shouldn't it rather be `rowCount`? `rowNumber` reads like a row index…

main/src/com/google/refine/ProjectMetadata.java
```diff
@@ -179,10 +227,28 @@ static public ProjectMetadata loadFromJSON(JSONObject obj) {
                     }
                 }
             } catch (JSONException e) {
-                // ignore
+                logger.error(ExceptionUtils.getFullStackTrace(e));
             }
         }
-
+        
+        if (obj.has("importOptionMetaData") && !obj.isNull("importOptionMetaData")) {
```
wetneb: I would spell this `Metadata` rather than `MetaData` to keep it consistent with the rest (and common usage).

main/webapp/modules/core/scripts/index/edit-metadata-dialog.js
```diff
@@ -0,0 +1,98 @@
+
+
+function EditMetadataDialog(metaData, targetRowElem) {
+  this._metaDataUIs = [];
+  this._metaData = metaData;
+  
+  this._MetaDataUI = function(tr, key, value, project) {
```
wetneb: Same here (`Metadata` instead of `MetaData`)

main/src/com/google/refine/ProjectMetadata.java
jackyq2015: Agreed. just changed

.gitignore
```diff
@@ -1,28 +0,0 @@
...
-extensions/rdf-exporter/module/MOD-INF/classes/
-broker/appengine/module/MOD-INF/classes/
-broker/core/module/MOD-INF/classes/
-broker/core/WEB-INF/lib/
-broker/core/data/
-broker/core/test-output/
-tmp/
-/test-output
-/bin
-open-refine.log
```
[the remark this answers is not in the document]
jackyq2015: yes. will add some test for the new command. The deletion was an incident. thanks for catch-up. will put it back

PR #1324, opened 2017-11-13
remove the rename link

jackyq2015:


jackyq2015:
what did you do to get this error message? that "name" is from the translation json file. Did you try to switch the language?

jackyq2015:
There is an issue with translation on master branch. Issue #1325  created

Issue #952, opened 2015-02-26
Feature request: text filter exclude


simon1tan:
Provide a text box to for the text filter widget to exclude data instead of include data. I know the facet has an inverse but a little different.

ettorerizza:
It would be a very simple yet usefull enhancement.

thadguidry:
@ettorerizza Something like this would work that says "exclude" where the "undefined" is ?

![screenshot-2017-10-31 clipboard - openrefine](https://user-images.githubusercontent.com/986438/32225316-b95156e4-be12-11e7-8a69-42b0807d094a.png)

For whomever wants to hack and learn... @ostephens !!! :) the code is here:
https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/browsing/facets/TextSearchFacet.java
https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/scripts/facets/text-search-facet.js

```
  elmts.caseSensitiveCheckbox.bind("change", function() {
    self._config.caseSensitive = this.checked;
    if (self._query !== null && self._query.length > 0) {
      self._scheduleUpdate();
    }
  });
  elmts.regexCheckbox.bind("change", function() {
    self._config.mode = this.checked ? "regex" : "text";
    if (self._query !== null && self._query.length > 0) {
      self._scheduleUpdate();
    }
  });
```
Just extend with something like this...and also wire it up properly in other places, just use the regexCheckbox as an example of what to add in other areas.

```
 elmts.excludeCheckbox.bind("change", function() {
    self._config.mode = this.checked ? "exclude" : "text";
    if (self._query !== null && self._query.length > 0) {
      self._scheduleUpdate();
    }
  });
```

wetneb:
@thadguidry you seem to be pretty close to a working prototype of that feature, why don't you propose a PR yourself? I'm sure you can do it!

ostephens:
I'm happy to have a look, but probably not this week now :)

I see this working the same as 'invert' on the facets - you retrieve the inverse set of rows/records to those that would be retrieved usually - i.e. all those that don't match the text in the filter.

What about implementing as an 'Invert' link to keep consistency with the Facet panel UI and terminology?

jackyq2015:
MR merged

PR #1227, opened 2017-08-03
Add test coverage and code analysis reports with Codacy

wetneb:
This adds Jacoco test coverage to the build file.

Developers can use the HTML reports are generated in `build/report/html` by `./refine server_test`. XML reports are submitted by Travis to Codacy and can be viewed [on their interface](https://www.codacy.com/app/OpenRefine/OpenRefine/dashboard).

I have changed the Travis script to run `./refine server_test` instead of `./refine test` because the browser tests are broken (they rely on Windmill, which is abandoned: see #733). This saves a few seconds of build time.

I have added a Codacy badge to the README.md : [Codacy Badge](https://www.codacy.com/app/OpenRefine/OpenRefine).
There is another badge that displays coverage ([Codacy Badge](https://www.codacy.com/app/OpenRefine/OpenRefine)) but we need to merge this into master before adding it.

I have also enabled Codacy integration on GitHub, so this PR will be reviewed by Codacy (and I hope it will be happy with it! I suspect it will not find any issue because I did not touch the Java code at all).

@thadguidry I have seen your Gradle file, but it seems easier to me to add Jacoco in the ant build - this way, it works both with your new Gradle setup and with the current build system.

PR #1327, opened 2017-11-14
Invert text search facet/text filter

ostephens:
Implementing an 'Invert' option for the text search facet (a.k.a filter) as per #952 

* New 'Invert' and 'Reset' options at the top left of the Text filter facet panel
* Click the 'Invert' link to invert the filter - i.e. find all rows which don't match the filter text or regular expression
* When 'Invert' is applied, panel gets an orange header to indicate the filter is inverted. The same device also implemented on text facets. In both cases the 'Invert' link appears struck-through (this latter device is the current behaviour for the Text Facet)
* 'Reset' clears all the options (invert, text in the filter box, case-sensitive, regular expression) of the filter

Comments on the UI welcome (@ettorerizza, @thadguidry, @wetneb)

I want to implement some tests for this functionality, but I can't work out how to best do this - @jackyq2015 @wetneb @thadguidry - any hints on the right approach to writing a test for these? I'm not quite sure how to apply the filter with options to a test set of data to see what the results are (I know what I expect the results to be - I'm just not sure on the mechanics to get the results). Pointers v welcome.

### DIFF
--- a/main/src/com/google/refine/browsing/filters/ExpressionStringComparisonRowFilter.java
+++ b/main/src/com/google/refine/browsing/filters/ExpressionStringComparisonRowFilter.java
@@ -52,11 +52,13 @@ import com.google.refine.model.Row;
  */
 abstract public class ExpressionStringComparisonRowFilter implements RowFilter {
     final protected Evaluable _evaluable;
+    final protected Boolean   _invert;
     final protected String    _columnName;
     final protected int       _cellIndex;
     
-    public ExpressionStringComparisonRowFilter(Evaluable evaluable, String columnName, int cellIndex) {
+    public ExpressionStringComparisonRowFilter(Evaluable evaluable, Boolean invert, String columnName, int cellIndex) {
         _evaluable = evaluable;
+        _invert = invert;
         _columnName = columnName;
         _cellIndex = cellIndex;
     }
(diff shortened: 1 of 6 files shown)


wetneb:
@ostephens I would write a test like this:
* create a sample project with a CSV importer
* add a facet that uses your inversion (this is done via the "Engine")
* create a RowVisitor and run it on the engine - the visitor will only be called for the rows that match the facet - you can do assertions in the visitor, or based on the visitor result.

main/src/com/google/refine/browsing/filters/ExpressionStringComparisonRowFilter.java
```diff
@@ -67,44 +69,44 @@ public boolean filterRow(Project project, int rowIndex, Row row) {
...
-        
+        Boolean invert = _invert;
         Object value = _evaluable.evaluate(bindings);
         if (value != null) {
             if (value.getClass().isArray()) {
                 Object[] a = (Object[]) value;
                 for (Object v : a) {
                     if (checkValue(v instanceof String ? ((String) v) : v.toString())) {
-                        return true;
+                        return (invert) ? false : true;
```
wetneb: I'd replace that by just `!invert`.

main/src/com/google/refine/browsing/filters/ExpressionStringComparisonRowFilter.java
```diff
@@ -67,44 +69,44 @@ public boolean filterRow(Project project, int rowIndex, Row row) {
...
                         if (checkValue(a.get(i).toString())) {
-                            return true;
+                            return (invert) ? false : true;
                         }
                     } catch (JSONException e) {
                         // ignore
                     }
                 }
-                return false;
+                return (invert) ? true : false;
```
wetneb: And that by `invert`.

ostephens:
Thanks @wetneb - is there an example of a test already that uses a RowVisitor with assertions?

wetneb:
@ostephens actually forget about that, I think there is a simpler way to test a facet: just run it on a row and check the result. So:
* instanciate your `Facet`
* use its `getRowFilter` to get a `RowFilter` out of it
* run it on a few rows from your project and check the result
Thanks for your effort, it's brave to write tests for a part of the code that is currently very poorly tested…  Once we have a few, it will be easier!

jackyq2015:
@ostephens You can take a look of CacheTests.java. For the Javascript, I am not quite sure. There are some popular framework such as  [Karma](https://karma-runner.github.io) seems quite good. But I don't know how OR can fit into it.

thadguidry:
Being test framework agnostic is a "good thing".
Owen, it probably would be wise to go headless with PhantomJS.   I like
that Karma can support it.

I myself have used Robot framework and also Sahi in the past.

+1 Karma seems to take a much easier route from my quick inspection.

-Thad
+ThadGuidry <https://plus.google.com/+ThadGuidry>

ostephens:
Thanks for that hint @wetneb - now added tests

ostephens:
@jackyq2015 @thadguidry not thinking of adding tests for the javascript/UI at the moment. I think there was an earlier attempt to add some tests for the JS using Windmill (https://github.com/windmill) but these have been disabled in the Travis build by #1227 because they were broken and Windmill discontinued

ostephens:
Thanks all!

PR #1329, opened 2017-11-14
add tests for SetProjectMetadataCommand

jackyq2015:


main/tests/server/src/com/google/refine/tests/commands/project/SetProjectMetadataCommandTests.java
```diff
@@ -0,0 +1,212 @@
...
+    @BeforeTest
+    public void init() {
+        logger = LoggerFactory.getLogger(this.getClass());
+    }
+    
+    // System Under Test
+    SetProjectMetadataCommand SUT = null;
+
+    // variables
+    long PROJECT_ID_LONG = 1234;
```
thadguidry: I would prefer this be representative of a real LONG.  I've often seen other projects test code use a max java long
9,223,372,036,854,775,807
to help uncover data type errors or conversions if needed or introduced in the future.  This will also help bulletproof ProjectManager a bit more as well in the hashmap.

main/tests/server/src/com/google/refine/tests/commands/project/SetProjectMetadataCommandTests.java
```diff
@@ -0,0 +1,212 @@
...
+        // verify
+        verify(request, times(2)).getParameter("project");      
+        verify(projMan, times(1)).getProject(PROJECT_ID_LONG);
+
+        verify(response, times(1))
+                .setHeader("Content-Type", "application/json");
+        verify(proj, times(1)).getMetadata();
+        try {
+            verify(response, times(1)).getWriter();
+        } catch (IOException e) {
```
thadguidry: Could any of these not also be a Servlet based error potentially as well ?  catch (IOException|ServletException e)
Just trying to harden our tests.

main/tests/server/src/com/google/refine/tests/commands/project/SetProjectMetadataCommandTests.java
```diff
@@ -0,0 +1,212 @@
...
+            Assert.fail();
+        }
+        verify(pw, times(1)).write("{ \"code\" : \"ok\" }");
+        
+        JSONObject obj = (JSONObject) proj.getMetadata().getUserMetadata().get(0);
```
thadguidry: hmm, wondering about maximums here for UserMetadata ... like if they use a client API for filling it instead ?
Coming from telecom, I'm always worried probably too much about maximums and seg faulting :)

main/tests/server/src/com/google/refine/tests/commands/project/SetProjectMetadataCommandTests.java
jackyq2015: We are using mockup object, so no ServletException. :)

main/tests/server/src/com/google/refine/tests/commands/project/SetProjectMetadataCommandTests.java
jackyq2015: I believe there is max size of JSONObject. need to dig more on this.

main/tests/server/src/com/google/refine/tests/commands/project/SetProjectMetadataCommandTests.java
thadguidry: 👍

main/tests/server/src/com/google/refine/tests/commands/project/SetProjectMetadataCommandTests.java
thadguidry: 👍

main/tests/server/src/com/google/refine/tests/commands/project/SetProjectMetadataCommandTests.java
wetneb: @thadguidry for a test like this it really does not matter at all: that's not what it is meant to test

thadguidry: It does not, I agree.  I just come from a background where tests are fairly extreme and hit system limits.  OpenRefine is not mil-spec :)

PR #1333, opened 2017-11-16
check if the project has the userMetadata

jackyq2015:
@thadguidry This is to fix the bug you found today that the index page will be blank after creating the userMetadata from the preference page.

### DIFF
--- a/main/webapp/modules/core/scripts/preferences.js
+++ b/main/webapp/modules/core/scripts/preferences.js
@@ -50,6 +50,13 @@ dictionary = data;
 $.i18n.setDictionary(dictionary);
 //End internationalization
 
+function deDupUserMetaData(arrObj)  {
+    result = _.uniq(JSON.parse(arrObj), function(x){
+        return x.name;
+    });
+    
+    return JSON.stringify(result).replace(/"/g, '\"');
+}
 
 function PreferenceUI(tr, key, value) {
   var self = this;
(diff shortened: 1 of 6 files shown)


jackyq2015:
Will put a filter when user define the userMetadata

wetneb:
@thadguidry again I think you should wait for a programmer to review the code before merging! Your reviews are very useful for the user standpoint, but we also need to watch for code quality. We really need to stick to this rules, otherwise we are just going to ruin the codebase. I am not saying I should have the last word on every PR: any coder with some familiarity with the code can do it (e.g. @ostephens).

thadguidry:
@wetneb Jacky and I would like to talk with you offline in a chat sometime.  Do you have time this weekend ?

wetneb:
@thadguidry sure! preferably in your morning so that it's not in the middle of my night :)

main/webapp/modules/core/scripts/preferences.js
```diff
@@ -50,6 +50,13 @@ dictionary = data;
 $.i18n.setDictionary(dictionary);
 //End internationalization
 
+function deDupUserMetaData(arrObj)  {
+    var result = _.uniq(JSON.parse(arrObj), function(x){
+        return x.name;
+    });
+    
+    return JSON.stringify(result).replace(/"/g, '\"');
```
wetneb: Why this `.replace(/"/g, '\"')`?

Issue #1328, opened 2017-11-14
Importing multiple Excel workbooks, each with multiple sheets


ColinVanAlstine:
**Refine**: Version 2.7
**OS**: Windows 7 Enterprise, Service Pack 1
**Browser**: Chrome - Version 61.0.3163.100 (Official Build) (64-bit)
**Google group thread**: [https://groups.google.com/d/msg/openrefine/EZ24zNWbzqo/mMNlPoLBAAAJ](url)

Opening a project with multiple Excel workbooks, each with a variable number of sheets, is causing some wonky behavior.  Users are only allowed to select the sheets to include from the first workbook.  If subsequent workbooks have more sheets than the first, only the number of initially selected sheets are pulled into the project.  If a workbook has less sheets than are initially selected, the project will hang during creation.  I've attached some files to test this behavior.  An option for 'select all sheets' would be helpful, in cases where there are a great number of sheets and selecting each one individually would be tedious.

> I've attached 4 of the workbooks that I am processing as an example.  If you create a new project and select the four files from your computer, when you get to Configure Parsing Option, you only are presented with the names of the sheet from the first workbook.  In my case, those are the sheets in Al-Ahram, of which there are only four.  

> I'm not given an option to select more sheets, even though Alamphon and Al-chark both have eight sheets each.  I put Concert Record (with only two sheets) as an example, because when you try to create the project, having selected the four sheets that are present in Al-Ahram, the project won't be created - it will just hang.  Try it with all four first to see the hanging, then try to create a project with just Al-Ahram, Alamphon and Al-chark.  For the three files, my goal is to see the data from all 20 sheets in Refine.  I'm currently only able to create a project with 16 sheets.

[Al-Ahram.xlsx](https://github.com/OpenRefine/OpenRefine/files/1471317/Al-Ahram.xlsx)
[Alamphon.xlsx](https://github.com/OpenRefine/OpenRefine/files/1471318/Alamphon.xlsx)
[Al-chark.xlsx](https://github.com/OpenRefine/OpenRefine/files/1471319/Al-chark.xlsx)
[Concert Record.xlsx](https://github.com/OpenRefine/OpenRefine/files/1471320/Concert.Record.xlsx)

ettorerizza:
By the way, it would be a good idea to consider adding a simple "select all the sheets" checkbox for Excel files. It is not uncommon for OCRs to produce Excel files containing several tens of sheets (one per page).

jackyq2015:
[...]

PR #1334, opened 2017-11-16
fix import multiple excel with mulitple sheets issue #1328

jackyq2015:


jackyq2015:
also added the open office for the same

Issue #1113, opened 2016-01-17
Separator in "Join multi-valued cells"


Opennat:
Edit cells -> Join multi-valued cells have `,` (comma & whitespace) separator as default value. If you change it to `,` (just comma) you will anyway get result with comma & whitespace. In history you could see just comma: "separator": ",")

ultraklon:
I've just tested this with a file with the following content
```
a,b
c,d
,e
f,g
,h
,i
```
and it looks like it works, when you leave the default ", " or you manually introduce that, you get
```
c|d, e
f|g, h, i
```
when the value "," is used, you get
```
c|d,e
f|g,h,i
```
BUT, when I tested it with ",&nbsp;&nbsp;&nbsp;&nbsp;" (comma and four spaces), I got the same results as with ", ". IDK if this is expected or not
With ",&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;@" (comma, five spaces and @), you get
```
d|@e
g|@h, @i
```
so this is looks like collapsing multiple spaces into one

ostephens:
So I've been having a look at this - it looks to me like the issue is with display rather than with the join. In the display, the spaces are collapsed (standard behaviour for html/web browsers), but if you actually access the value in the field, the spaces are present. See screenshot recreating scenario above with cell shown next to edit box for the cell - both contain the same information, but display differently.

<img width="720" alt="screen shot 2017-10-24 at 18 15 39" src="https://user-images.githubusercontent.com/576174/31957747-a205d65c-b8e7-11e7-99b3-af7b1279a63d.png">

ostephens:
Which leads me to ask - is there any reason we don't style cells with css `white-space: pre-wrap` ? This would be a simple and I think useful change

ettorerizza:
Arf, you're right @ostephens. This is particularly misleading, since these extra spaces are also invisible in the transformation window.

![screenshot-localhost-3333-2017-10-24-19-36-25-495](https://user-images.githubusercontent.com/6286410/31958687-d5a92972-b8f2-11e7-9348-e03729d77515.jpeg)

thadguidry:
[...]

PR #1336, opened 2017-11-16
Add tests for JoinMultiValuedCells

ostephens:
Increasing test coverage and also ensuring that this function works as expected (see #1113)


ostephens:
:) I wrote the tests anticipating I was going to need a test if I made changes because of #1113 - but then found it was actually a display issue - so more bad planning than heroic!

Issue #1332, opened 2017-11-16
transformer dialog  is displayed all along the screen


jackyq2015:
As mentioned in https://github.com/OpenRefine/OpenRefine/issues/1221#issuecomment-344499767, the dialog should keep the consistence with others.

thadguidry:
Thanks for the temp fix on this Jacky.  Hoping we can get much better dialog handling in later milestones.  We'll close this now.

tfmorris:
The highest priority for consistency is between the three transform expression dialogs, which are different after this change, and, only then, with other similar dialogs.

We also don't have support for resizable dialogs (as mentioned in #421), so they can't be sized by percentages.

PR #1339, opened 2017-11-17
to adjust the width of transform dialog. issue #1332

jackyq2015:
change to 80% from 100%

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/text-transform-dialog.html
+++ b/main/webapp/modules/core/scripts/views/data-table/text-transform-dialog.html
@@ -1,4 +1,4 @@
-<div class="dialog-frame" >
+<div class="dialog-frame" style="width: 80%;" >
   <div class="dialog-border">
     <div class="dialog-header" bind="dialogHeader"></div>
     <div class="dialog-body" bind="dialogBody">
--- a/refine
+++ b/refine
@@ -942,7 +942,7 @@ elif [ "$OS" = "linux" ] ; then
 fi
 
 echo You have "$freeRam"M of free memory. 
-echo Your current configuration will allow to use $REFINE_MEMORY of memory.
+echo Your current configuration is set to use $REFINE_MEMORY of memory.
 echo OpenRefine can run better when given more memory. Read our FAQ on how to allocate more memory here:
 echo https://github.com/OpenRefine/OpenRefine/wiki/FAQ:-Allocate-More-Memory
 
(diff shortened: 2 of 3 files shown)


main/webapp/modules/core/scripts/views/data-table/text-transform-dialog.html
```diff
@@ -1,4 +1,4 @@
-<div class="dialog-frame" >
+<div class="dialog-frame" style="width: 80%;" >
```
wetneb: Wouldn't it be cleaner to store that in CSS? It's generally recommended to avoid inline styling.

ostephens: The approach @jackyq2015 has taken here is consistent with all the other 'dialog-frame' divs in the app at the moment. IMO ideally all these should be done instead by adding a class to the div and styling via css. BUT, I'd suggest consistency is better than doing this one instance 'correctly' - then we can tidy them all together (I'd be happy to take this on once this PR is merged)

jackyq2015: agree

refine
```diff
@@ -942,7 +942,7 @@ elif [ "$OS" = "linux" ] ; then
 fi
 
 echo You have "$freeRam"M of free memory. 
-echo Your current configuration will allow to use $REFINE_MEMORY of memory.
+echo Your current configuration is set to use $REFINE_MEMORY of memory.
```
wetneb: This PR contains two unrelated changes… It's generally better to use separate branches and PRs for that.

main/webapp/modules/core/scripts/views/data-table/text-transform-dialog.html
thadguidry: Team, we know its best practice, but let's not focus too much on the frontend incidentals.  Our plan is for a UI overhaul with Google's help.  Honestly, we just need to work on the backend and just "fast quick repair" on the frontend to make things "work good enough for now", knowing that our UI re-work can happen with Google's assistance in a bit.  I already talked to Jacky about this during my hang out with him.  Simon@Google is working on getting a designer or two for helping us out.

Speaking of hangouts, we need to come to some regularly scheduled hangouts for the 4 of us.  Oh @ostephens if you didn't know, we kinda voted you as a core contributor in the past day :) and I'll send out an official email shortly, announcing.

refine
thadguidry: @wetneb We know.  Jacky is just fixing some incidentals that I bring up in our hangouts.  There just minor typo fixes. I know, I know, it makes the changes look dirty. I know.  But for typos, I am OK with dirty changes of typos within feature pulls.  I know your probably not, but I and Jacky are OK with them.  We're moving fast, because we have an agenda to get this release done, moved out...so that we can actually begin to plan better and focus on "the 1st quarter fun".

Issue #1353, opened 2017-11-23
Allow to delete all projects


xseris:
maybe an option that allows to delete all projects can be useful:
![](https://i.imgur.com/WnEnsWR.png)

ettorerizza:
Most dangerous option ever ! Especially since it can be done easily by opening the workspace directory, CTRL + A, DEL. 

@xseris Your tag system is pretty cool. Could you share the recipe? :)

jackyq2015:
should just go to work space and delete all files easier? but delete all is really dangerous option to provide

xseris:
@ettorerizza I know that could be dangerous, but in my personal implemetation is a "tag-based remove all". The upper bar contains all defined project tags, and clicking on one of them returns and shows a filtered list of projects that contain that specific tag. Then can be all removed in block (Of course with a warning message).  Answering your question: Yes, I can share this tag-system implementation, but i need some time to isolate the feature's code from other junk personal implemented stuff.

ettorerizza:
In this case, it's different. The option seems quite justifiable.

xseris:
@ettorerizza I have isolate the tag system code, the following is a fast screencast of the feature...

![](https://i.imgur.com/wSVMexx.gif)

I created a PR ( #1357 ) if you want to test it more deeply

ettorerizza:
@xseris Many thanks, I'll try this !

jackyq2015:
PR was merged

PR #1357, opened 2017-11-24
Tag system

xseris:
According to discussion in #1353 , I propose my personal project tag system. At project's creation time, the user can generate new tags or choose existing ones for the project. This tags works as "filters" in the open project page.


xseris:
I implemented the feature quite long time ago when the metadata structure introduced in OpenRefine 2.8 did not exist yet. That's why tags are treated as metadata in the code. I am not expert but if it is logically correct and/or it is better for future developments, we can keep tags separated from metadata. For the visualization part of the discussion: you focused pros and cons, I cannot say which solution is better.

ettorerizza:
"By showing them, it provides a good access point to edit the tag, but at the same time it take too much space. "

@jackyq2015 Looks like the system is able to handle a relative large number of tags without adversely affecting the display.

screenshot-localhost-3333-2017-11-25-01-11-07-077

jackyq2015:
@xseris I see. I will test further. I am worry that the editing the metadata and the user metadata might not work properly because the tag takes a column but the current UI does not aware of this. Besides that, you may try to polish the codacy issues as you can(some are irrelevant which you can ignore). Also some basic Unit Test will help also.

I think we can merge it if other developers has no issue. 

@ettorerizza It looks good. I really like the navigation bar at the top!

fpompermaier:
We usually use those tags to (improperly) separate projects coming from
different customers..this because IMHO Refine lacks the cocept of workspace
(i.e. a group of projects).
I think that addition would be VERY useful

ettorerizza:
I wonder if it would not be a good idea to make the dark blue tags clickable.

thadguidry:
@wetneb I disagree with @ettorerizza and purely because of semantics. :)  I'd rather NOT replace the "subject" field.  Here's why:

In Schema.org, DCMI, and even Wikidata itself these are different properties.
Tags can be of the form "lamborghini time trials" or even #michaeljacksonconcerts.
Which is different than the subjects of Automobile Racing and Pop Star.
http://schema.org/keywords
http://pending.schema.org/about
Let's keep these concepts separate, even scientists treat them separate at times.

jackyq2015:
Agree. I think subject is formal and tag system is more free style. Maybe it can replace the "user defined metadata" instead?

ettorerizza:
You may be right @thadguidry . I do not have a strong opinion on the question. In Dublin Core, the "Subject" element can be expressed as "keywords, key phrases or classification codes that describe a topic of the resource". In this case, using tags instead of a single subject has the advantage of the polyhierarchy. But I know little Schema.org.

wetneb:
The problem I have with the "subject" field is that it is just free text: that's not very useful to organize your project library. The notion of subject is ill-defined, no importers fill that field because no serious file format will bother providing a field for that… And most users will just not bother filling up that field either, so we'll have a sad empty column taking a lot of space. So yeah, this field is just useless for me, that's why I'd be happy to see it replaced by the tags.

@jackyq2015 I don't understand why you say that "subject is formal and tag system is more free style"… In fact I was about to write the exact opposite!

ettorerizza:
""subject is formal and tag system is more free style"… In fact I was about to write the exact opposite!"

@wetneb I agree for this particular case. I really like the possibility of adding tags when creating the project. It can become more than just folksonomy. As long as these tags are stored in a separate Json file, this would give users the ability to create and use their own controlled vocabulary.

jackyq2015:
@wetneb quoted from @ettorerizza :
 In Dublin Core, the "Subject" element can be expressed as "keywords, key phrases or classification codes that describe a topic of the resource. Recommended best practice is to select a value from a **controlled vocabulary or formal classification scheme**".

But the tags is user defined which means you can choose what ever you like and they just have meaning for that user. It is NOT a mandate that they have to chosen from a controller vocabulary. In other words, the tag system could be local and does not have to to published. 

Currently since OR's metadata is not standardised, you can argue that subject means something else.  But that's not the point.

thadguidry:
@wetneb Careful.  Just because "subject" is not useful to you, doesn't mean that @ostephens and others in OpenRefine's bibliographic world will not appreciate it.  :)

There's no harm in keeping "subject" and it provides value by NOT squashing 2 different concepts, A Tag, and A Subject into one thing.  It is best to keep these 2 concepts separate, especially in regards to any metadata discussion in OpenRefine.  This also helps when having to integrate data with Knowledge Based System (KBS) tools outside of OpenRefine used in libraries and research centers.  Our JDBC export feature will need an option to publish the metadata as well eventually, but can work on that later.  Many Database Schemas have 2 columns for this, one for Subject Matter and one for Tags.
(specifically, for instance in MODS metadata http://www.loc.gov/standards/mods/mods.xsd the
```<xs:element ref="subject"/>``` 
 versus the
```<xs:element ref="relatedItem"/>```)

wetneb:
@jackyq2015 if the "Subject" field in OR is meant to be "dc:subject", I think this should be made clear in the UI… The screenshots I have seen did not use it in this way ;-) . But anyway, you can also decide to use a controlled vocabulary for the tags (and you should!). Both of them can be used with a controlled or uncontrolled vocabulary, but the tagging system seems a lot easier to use for that purpose. If you like "dc:subject", then we can decide that each tag is meant to be a value for this property! :)

@thadguidry yeah if people prefer free text over structured lists, fine! I don't actually care about that, I'm just saying that this looks pretty useless to me (and a waste of space) in the current form. But it's clearly not a hill I want to die on…

ostephens:
Just chipping in on the subject vs tag question. I think the key thing to differentiate is 'metadata to help navigate OR projects on your local install' vs 'metadata that may be imported/exported with the project'. I'm in favour of the tag system for local project management, but it may not be the right information to (e.g.) export as 'keywords' if we support metadata export for things like data packages in the future.

I don't have very strong feelings about it, but my instinct is with @thadguidry that 'subject' should be kept for more general purposes and tags should go into a field of their own (which could be 'local-subject' rather than 'tags'?). But it's only an instinct - no strong reasoning to back this up

xseris:
@jackyq2015 I solved the null pointer exception in tests, but now I get another error: "org.mockito.exceptions.verification.NoInteractionsWanted" that i don't know how to fix it. Can you help me?

ettorerizza:
"this looks pretty useless to me (and a waste of space) in the current form. "

@wetneb @jackyq2015 Same problem, and even worse, with the metadata "creator". I did not bothered once to fill it. In 99.9% of cases, I am the creator. Should not it be automatically filled in with the username of the session, leaving that field editable?

jackyq2015:
@xseris Those verifyNoMoreInteractions(project) issue means that your code introduced new invocation against the project other than ALL the verify statements before it. you can find out which invocation you missed from the testng log and add tha to do a verify. It is not that straightforward. But once you does one, rest of them most likely are the same. 

@ettorerizza The problem is that OR don't know who you are thus can not fill that for you. we can put a setting as a workaround though

jackyq2015:
@xseris Great work! I will checkout locally and do some UI testing. If there is no issue found, I will approve this PR. We can have more discussion regarding which field can be replaced with the "tag system" before or even after the merge. It is not I don't like the change. It just take a while to have everyone at the same page. Agree?

thadguidry:
@jackyq2015 yeap, I agree

wetneb:
Thank you so much for this work! Here are my thoughts after the changes:
* Why are the tags aligned to the right? In a left-to-right language I would expect to see them aligned to the left.
* There is no consensus to replace the subject field by the tags. What if we kept the subject field, but without displaying it in the table? (It would only be accessible via the about link. I would do the same for the author and description fields actually) Or, if there is still no consensus for that: changing the position of the tags column so that it is just after the "Name" column.
* Tags should be clickable (that was pointed out by others)
* The tag input is not as easy to use as it could be - could we use a jQuery plugin for that? Here is a list: http://jquery-plugins.net/tag/tag-input (but that's just a nice plus to have - I'd be happy to merge without that)

jackyq2015:
@xseris I found some issue with the position of the "tag". we support the "user defined metadata" and I just write down the procedure to do so:
https://github.com/OpenRefine/OpenRefine/wiki/Metadata

Currently the "tag" column always take the last column and it will mess up the "user defined metadata" if there is any ( the tag will take the space of the last user metadata). But if you can move the tag right after the "name" column and that will solve the problem automatically without doing anything else. We can  hide "subject" and "creator" later on if everyone agree. but that can happen after the merge.

So from my side, I only expect that tag column get moved after "name" column(which @wetneb  already pointed out)

jackyq2015:
@xseris Do you mind I merge your change into my repo from where I am doing some other meta-data feature? All the history can be kept. 

I need to add a "Data Package" link besides the "About". I would like to merge your change to avoid more complex  3-way merge in future. Let me know if you are ok with that?  I will move the tag column right after the name column as we discussed.

xseris:
@jackyq2015 I tried locally but just moving tag column right after the name one does not solve the issue that you raised. This because the setAnyField method seems not to be able to interpret the tag string into an array (tags are actually stored as string array). 
I already solved it (and also moved the column right to name) with a workaround. Also the filter bar needed to be refreshed and the tags into the specific cell rebuild. I still have to solve a little GUI problem, but it will not take too much time (max 1 day).
Then you are free to merge all modifications into your repo.

xseris:
@jackyq2015 You can merge into your repo now

jackyq2015:
@xseris I merged to my repo:
https://github.com/jackyq2015/OpenRefine/commits/feature/metadata

Looks great! I noticed that there is some sample tags like "big data test1 test2". where those comes from? I have not start to set the tag yet.

xseris:
@jackyq2015 This is strange, I don't see those tags. There are not even tags injected in the code...

jackyq2015:
My bad. I think i tested it before.

wetneb:
@thadguidry your review is still marked as "requested changes" - do you still have any concerns?

thadguidry:
@jackyq2015 feel free to merge when ready.  You can tackle remaining updates or bug fixes with another pull. Let's merge this as you say to avoid a 3 way merge.  Thanks team !

jackyq2015:
@xseris Seems when moving mouse point to the "tag" column, there is not menu displayed to add tags any more in the current code base. I tested on your repo and it's the same. could you please take a look and create another PR to fix this issue?

PR #1362, opened 2017-11-27
Update WeLoveOurBackers.md

thadguidry:
Removing Bountysource, since we unanimously decided to go away from this support model.
Adding Google Inc. News Lab.


wetneb:
Are we sure that we want to put Google on an equal footing as individual contributors? Their contribution is of a different nature…

ostephens:
I think all the backers contribute in different ways - whether they are individuals, groups or organisations. I'm OK with having a single list including all the different types of backer

Issue #1297, opened 2017-11-01
Text filter causes exception when regex pattern is not yet complete


thadguidry:
Using OpenRefine release 2.7.0
This can be reproduced with any text cell value, for instance

1. Add text filter on column.
2. Type into the text filter column ...

```^tha```

3. now type 1 more character, a backslash character \

```^tha\```

4. Notice that you get a waiting spinner and the following exception caught in the console:

```
16:43:41.814 [                  command] Exception caught (1ms)
org.json.JSONException: Unexpected internal error near index 5
^tha\
     ^
        at com.google.refine.browsing.facets.TextSearchFacet.initializeFromJSON(
TextSearchFacet.java:107)
        at com.google.refine.browsing.Engine.initializeFromJSON(Engine.java:202)

        at com.google.refine.commands.Command.getEngine(Command.java:154)
        at com.google.refine.commands.row.GetRowsCommand.internalRespond(GetRows
Command.java:97)
        at com.google.refine.commands.row.GetRowsCommand.doPost(GetRowsCommand.j
ava:70)
        at com.google.refine.RefineServlet.service(RefineServlet.java:177)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:820)
        at org.mortbay.jetty.servlet.ServletHolder.handle(ServletHolder.java:511
)
        at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(Servlet
Handler.java:1166)
        at org.mortbay.servlet.UserAgentFilter.doFilter(UserAgentFilter.java:81)

        at org.mortbay.servlet.GzipFilter.doFilter(GzipFilter.java:132)
        at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(Servlet
Handler.java:1157)
        at org.mortbay.jetty.servlet.ServletHandler.handle(ServletHandler.java:3
88)
        at org.mortbay.jetty.security.SecurityHandler.handle(SecurityHandler.jav
a:216)
        at org.mortbay.jetty.servlet.SessionHandler.handle(SessionHandler.java:1
82)
        at org.mortbay.jetty.handler.ContextHandler.handle(ContextHandler.java:7
65)
        at org.mortbay.jetty.webapp.WebAppContext.handle(WebAppContext.java:418)

        at org.mortbay.jetty.handler.HandlerWrapper.handle(HandlerWrapper.java:1
52)
        at org.mortbay.jetty.Server.handle(Server.java:326)
[...]

PR #1365, opened 2017-11-28
User-friendly regular expression errors

ostephens:
An attempt to provide users with user-friendly messages for common issues with regular expressions as per #1297 

What I've done here is to parse some of the error messages from java.util.regex.PatternSyntaxException and return a user-friendly equivalent. It does the job, but I'm unsure this is really the best approach.

Feedback welcome including suggestions for alternative approaches.

Currently only implemented for the Text Filter w Reg Expressions, but it could be implemented to do the same for any GREL operations accepting Reg ex by using it in the appropriate part of com.google.refine.grel.Parser

### DIFF
--- a/main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
+++ b/main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
@@ -86,9 +86,6 @@ public class PatternSyntaxExceptionParser {
         switch(desc)
         {
             case "Unclosed character class":
-            //Need these errors to be more human readable
-            //Possibly include html for formatting
-            //Update tests first with user friendly errors
                 sb.append("The regular expression is missing a closing ']' character, or has an empty pair of square brackets '[]'.");
                 break;
             case "Unmatched closing ')'":
--- a/main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
+++ b/main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
@@ -1,6 +1,6 @@
 /*
 
-Copyright 2010, Google Inc.
+Copyright 2017, Owen Stephens.
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -13,7 +13,7 @@ notice, this list of conditions and the following disclaimer.
 copyright notice, this list of conditions and the following disclaimer
 in the documentation and/or other materials provided with the
 distribution.
-    * Neither the name of Google Inc. nor the names of its
+    * Neither the name of the copyright holder nor the names of its
 contributors may be used to endorse or promote products derived from
 this software without specific prior written permission.
 
--- a/main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
+++ b/main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
@@ -1,0 +1,107 @@
+/*
+
+Copyright 2010, Google Inc.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+    * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+    * Neither the name of Google Inc. nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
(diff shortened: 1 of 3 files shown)


main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
```diff
@@ -0,0 +1,127 @@
+/*
+
+Copyright 2010, Google Inc.
```
thadguidry: Why are you copyrighting as Google instead of yourself ?

main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
```diff
@@ -0,0 +1,127 @@
...
+            "Unclosed character family"
+            "Empty character family"
+            "Unknown Unicode property {name=<" + name + ">, "+ "value=<" + value + ">}"
+            "Unknown character script name {" + name + "}"
+            "Unknown character block name {" + name + "}"
+            "Unknown character property name {" + name + "}"
+            "named capturing group has 0 length name"
+            "named capturing group is missing trailing '>'"
+            "Named capturing group <" + name + "> is already defined"
+            "Look-behind group does not have " + "an obvious maximum length"
```
thadguidry: Yup, this seems to output nicely, regardless.

`Look-behind group does not have an obvious maximum length near index 169 \n(?<kiloCharacters>[^\n]{1000})(?<=(?<newLine>\n)(?<pairsAndText>[^'\n]{0,1001}|[^\n']{0,1001}'[^\n']{0,1001}'[^\n']{0,1001}){0,1001}(?<oddComa>')(?<text>[^\n']{0,1001}))(?(?<=')(?!')) ^`

main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
```diff
@@ -0,0 +1,127 @@
...
+
+*/
+
+package com.google.refine.util;
+
+import java.util.regex.PatternSyntaxException;
+
+public class PatternSyntaxExceptionParser {
+    /*  Class to translate PatternSyntaxExceptions into more user friendly error messages
+        Currently translates the following error messages from java.util.regex.Pattern
```
thadguidry: What was your basis on translating these into friendly messages ?  Did you look to something online to assist with this set of PatternSyntaxExceptions ?  Are we complete on showing "something useful" for all kinds of PatternSyntaxExceptions in your mind ? Not missing any ?

jackyq2015:
can this take of all the exceptions or just the ones listed in the code?

main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
ostephens: Lazy just copying existing files as a template - will amend!

main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
ostephens: I used the list from Regexr as a starting point, but basically it was just the ones that seemed the most likely (to me) to be used and cause confusion

ostephens:
@jackyq2015 yes - it takes a PatternSyntaxException as an argument and looks for specific patterns in the error message. If it doesn't find any of the defined patterns it falls through to just returning the message which PatternSyntaxException returns from its getMessage() method

main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
ostephens: However, no problem in extending the ones it catches - could do them all if we feel it is worthwhile

thadguidry: Sure. OK, this is good enough for now.

thadguidry:
@ostephens However Codacy doesn't seem to detect that falling through.  Click on the yellow banner on this for details  https://www.codacy.com/app/jackyq2015/OpenRefine/file/12515874477/issues/source?fileBranchId=5930252#l@result.result.startLine

ostephens:
@thadguidry I think that's caused by the lack of a break statement in the default case - I can update that

main/src/com/google/refine/util/PatternSyntaxExceptionParser.java
```diff
@@ -121,6 +121,7 @@ public String getUserMessage() {
                 // If no special handling in place fall back on error msg
                 // created by java.util.regex.PatternSyntaxException
                 sb.append(exception.getMessage());
```
jackyq2015: Not sure the StringBuffer is necessary here. It is one time assign. String should be more efficient and straightforward. You can also consider a HashMap.

Issue #1380, opened 2017-12-13
Update "Last modified" date only when project is actually modified and not when it is just opened


DavidFichtmueller:
In the project list under "Open Project", each time a project is opened, the "last modified" date of the project is reset, even when no edits were made. The time stamp however should only be updated when actual changes are made to the project (i.e. steps are added to the history). 

Steps to reproduce: 

- Start OpenRefine
- Go to "Open Project" and select an old project (preferably one that is not on top of the list)
- Open the project and do nothing
- Go back to the OpenRefine start page
- Refresh the page by pressing F5
- Go to "Open Project"
- The viewed project is now on top of the list with the current time as "last modified"

Alternatively a second column for "Last viewed" could be added for this kind of behavior but the "Last modified" column should remain stable until the project is actually modified. 

System Info
- OpenRefine 2.8 (this behavior was not present in 2.7)
- Windows 7
- Tested with Firefox 57 and Chrome 63.

GlassGruber:
Sorry to bump but I'm experiencing the very same behavior with latest release 3.6.2 with included OpenJDK on windows. 

On the console I can see this (I'm pasting a screenshot because the console windows does not allow copy)

![image](https://user-images.githubusercontent.com/1172428/195799207-45bbdd29-3f41-4fbd-8a66-bd71225eb937.png)

Is it possible to take more precise info for this? I've checked the [docs ](https://docs.openrefine.org/manual/installing#logs) but I guess no other logs are available?

Thank you!

thadguidry:
@wetneb I think it makes more sense now to have the label changed to "last opened" versus "last modified" to actually reflect what our code is doing (and I think it was always "last opened" in reality in very early versions, but let me doublecheck).  Also, we might visit the related #3047 which seems a low impact direct fix that could be done (without squabbling about metadata of projects)
What do you think?

DavidFichtmueller:
> I think it makes more sense now to have the label changed to "last opened"

While this would actually better describe what is happening, it is not the kind of fix that would be useful. That is like changing the sign of a broken escalator to "stairs". Technically correct but neither what the people want or need nor what it was originally build for. 

I can not really help much in that regard, but I wanted to add my two cents, because I feel that changing the label would cement the current state.

PR #1382, opened 2017-12-15
don't refresh row number each time. issue #1380

jackyq2015:


.settings/org.eclipse.wst.jsdt.core.prefs
```diff
@@ -186,6 +186,7 @@ org.eclipse.wst.jsdt.core.formatter.insert_space_after_comma_in_superinterfaces=
 org.eclipse.wst.jsdt.core.formatter.insert_space_after_comma_in_type_arguments=insert
 org.eclipse.wst.jsdt.core.formatter.insert_space_after_comma_in_type_parameters=insert
 org.eclipse.wst.jsdt.core.formatter.insert_space_after_ellipsis=insert
+org.eclipse.wst.jsdt.core.formatter.insert_space_after_function_keyword=do not insert
```
wetneb: Is that related to the bug?

main/src/com/google/refine/commands/row/GetRowsCommand.java
```diff
@@ -180,7 +180,8 @@ protected void internalRespond(HttpServletRequest request, HttpServletResponse r
             }
             
             // metadata refresh for row mode and record mode
-	    if (project.getMetadata() != null) {
+	    if (project.getMetadata() != null
+	            && project.getMetadata().getRowCount() == 0) {
```
wetneb: What if the row count changes later on (for instance if the user removes some rows, or uses a transpose operation)? Is the row count going to be updated somewhere else?

I would check if the row count is different from what is stored in the metadata, rather than comparing to zero.

jackyq2015: The row number should be recounted after operation. Thanks for checking. just make sure.

.settings/org.eclipse.wst.jsdt.core.prefs
jackyq2015: no. for some reason, my eclipse automatically changed the setting while importing project

wetneb: then I guess it's cleaner if we don't include that in the PR :)

Issue #1386, opened 2017-12-18
Projects with timezone in modified time metadata are not displayed in Open Project


ostephens:
This is on the current Master branch with new Tags functionality implemented. None of my projects created on a pre-tags version of OR display in the Open Project screen. However, projects created with the Tags version of OR do display even if no tags assigned.

The 'All' option should display all projects, no matter whether any Tags metadata exists or not.

jackyq2015:
Not sure if it is caused by merge or not. let me check and update.

jackyq2015:
@ostephens Did you see any exception at back end? could you please paste the console message while startup?

ostephens:
No exceptions or issues flagged in the console. However, checking I can see that get-all-project-metadata is coming back with an empty list:

`{"projects":{}}`

ostephens:
OK, think I have tracked down the issue to the parsing of the "modified" date from the project metadata.json. This happens in ParsingUtilities.stringToLocalDate(String s)

```
static public LocalDateTime stringToLocalDate(String s) {
      System.err.println(LocalDateTime.parse(s, DateTimeFormatter.ISO_LOCAL_DATE_TIME));
      return LocalDateTime.parse(s, DateTimeFormatter.ISO_LOCAL_DATE_TIME);
    }

```
The string is going in, but nothing is being returned (which I think leads the creation of the projectmetadata to fail in ProjectMetadata.loadFromJson"

```
ProjectMetadata pm = new ProjectMetadata(JSONUtilities.getLocalDate(obj, "modified", LocalDateTime.now()));
```

The issue seems to be that the existing metadata.json records time with strings like:
`  "modified": "2014-01-15T21:46:25Z",`

But `LocalDateTime.parse(s, DateTimeFormatter.ISO_LOCAL_DATE_TIME);` requires it to have no timezone information.

If I manually edit the metadata.json to read:
`  "modified": "2014-01-15T21:46:25",`
The project displays OK in the UI

jackyq2015:
I see. It is a regression introduced by local time stamp. I had a UT for testing the new API can parse the old project. but still missed something. will figure out and close it

ostephens:
@jackyq2015 I have projects created over a long period - I don't know which version of Java or OR I was using with them I'm afraid

jackyq2015:
that is fine. no worry, Owen!

ettorerizza:
[...]

PR #1388, opened 2017-12-19
Fix Issue #1386

jackyq2015:
parse the time end with Z

### DIFF
--- a/main/src/com/google/refine/util/ParsingUtilities.java
+++ b/main/src/com/google/refine/util/ParsingUtilities.java
@@ -38,6 +38,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.UnsupportedEncodingException;
+import java.text.SimpleDateFormat;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
--- a/main/src/com/google/refine/util/ParsingUtilities.java
+++ b/main/src/com/google/refine/util/ParsingUtilities.java
@@ -38,9 +38,10 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.UnsupportedEncodingException;
-import java.text.SimpleDateFormat;
+import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
+import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.Properties;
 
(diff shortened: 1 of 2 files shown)


main/src/com/google/refine/util/ParsingUtilities.java
```diff
@@ -193,6 +195,10 @@ static public OffsetDateTime stringToDate(String s) {
     }
     
     static public LocalDateTime stringToLocalDate(String s) {
-      return LocalDateTime.parse(s, DateTimeFormatter.ISO_LOCAL_DATE_TIME);
+        if (s.endsWith("Z")) {          // UTC time
+            Instant instant = Instant.parse(s);
+            return LocalDateTime.ofInstant(instant, ZoneId.systemDefault());
+        } else
```
ostephens: Missing brace

ostephens:
Once that syntax issue fixed then this PR solves the immediate problem. Just wondering - are we sure that the only dates out there in older projects are of the format yyyy-MM-ddThh:MM:ssZ ? Could they ever be expressed with other timezone data - e.g. the Java ISO_DATE_TIME format which allows datetimes like:

'2011-12-03T10:15:30+01:00[Europe/Paris]'

If we are sure it is only those ending in Z that are problematic then that's OK, just wondering if we need a more comprehensive solution.

In relation I think it would be good to catch and deal with situations where the string->datetime conversion fails silently (which is essentially what happened in this case) so that the project isn't hidden from view. Potentially have a check for a valid datetime before trying to create the ProjectMetdata object, and fallback to using current date if the one parsed from string is invalid/null/zero length?

jackyq2015:
yes, I am sure that is the only format. there is a formatter defined for iso8601 in the parsing utility class.

main/tests/server/src/com/google/refine/tests/util/ParsingUtilitiesTests.java
```diff
@@ -107,4 +110,25 @@ public void parseProjectBeforeJDK8() {
...
+    public void strSubstitutorTest() {
+        Map<String, String> data = new HashMap<String, String>(6);
+        
+        data.put("value", "1234");
+        data.put("field_format", "String");
+        
+        StrSubstitutor sub = new StrSubstitutor(data);
+        String message = "The value ${value} in row ${row_number} and column ${column_number} is not type ${field_type} and format ${field_format}";
+        String result = sub.replace(message);
+        System.out.println("XXXXXXXX:" + result);
```
wetneb: I wouldn't print anything like this in a unit test.

PR #1392, opened 2017-12-22
Support log for Windows folks running refine.bat

thadguidry:
Too often we need to ask multiple questions from folks running OpenRefine on Windows.  Let's make this easier for us from a support perspective and just ask the users to paste the contents of the support.log file to us.  We can expand on any support.log enhancements later as we come across the need to add more.

### DIFF
--- a/refine.bat
+++ b/refine.bat
@@ -68,6 +68,19 @@ for /f "tokens=1,* delims==" %%a in (refine.ini) do (
     set %%a=%%b
 )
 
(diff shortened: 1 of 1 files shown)


jackyq2015:
As discussed, this batch will not be called when user run openrefine.exe. But with this PR, we still can have user to run refine.bat if we need extra info to do the troubleshooting. It is not ideal. We can have some sort of file logger to collection all the general info for troubleshooting in future.

Issue #1269, opened 2017-10-18
Export to MediaWiki's Tabular JSON and upload to Wikimedia Commons


wetneb:
MediaWiki supports a tabular data format that can hold some metadata. It would make sense to ease the export to that format and upload to Wikimedia Commons.

https://www.mediawiki.org/wiki/Help:Tabular_Data

- [ ] Write a Tabular JSON exporter
- [ ] Optional: Add upload wizard for Wikimedia Commons as part of the Wikidata extension

jackyq2015:
Is exporting metadata to tabular data format  a general approach or it is just for wikidata? why we need to upload to wikimedia Commons

wetneb:
Hi @jackyq2015 - the tabular capabilities of Wikimedia Commons are (in principle) independent from Wikidata. However, they are of high interest to the Wikidata community because they provide an alternative storage method for some types of datasets that are too big to be represented as Wikidata claims (e.g.: if you have an hour-by-hour history of some exchange rate between two currencies over the past 10 years, you don't want to add one claim per line!).

Here is an example of what it looks like:
https://commons.wikimedia.org/wiki/Data:Taipei_Population.tab

thadguidry:
@wetneb Is there export functionality on Wikimedia Commons for the .tab to CSV or supporting [CSV on the Web](https://www.w3.org/TR/tabular-data-primer/) ?

jackyq2015:
My understanding of metadata is "data of data". seems wikimedia also call the data itself "metadata".

wetneb:
Here is an example of the sort of metadata that these files can contain and that OpenRefine does not currently store:

![metadata](https://user-images.githubusercontent.com/309908/32014991-a4997cfe-b9b7-11e7-9bfe-12a56ac32824.png)

@thadguidry I think the specs are a subset of OKFN's Data Package (#778) but the features are quite limited for now.

So, to implement this exporter, we would need to have support for more project-level metadata (description, source) and column-level metadata (type, label, and maybe more?).

thadguidry:
@jackyq2015 Right understood.

wetneb:
This feature is still experimental and few properties use it in Wikidata - so I'm removing the high priority.

tfmorris:
[...]

Issue #1096, opened 2015-11-28
Tabular dataset metadata - http://www.w3.org/TR/tabular-data-model/


psychemedia:
If you're collecting enhancement requests, would it be useful to adopt the W3C Proposed [Metadata Vocabulary for Tabular Data](http://www.w3.org/TR/2015/PR-tabular-metadata-20151117/) as part of import and export functionality?

(I'm not sure if things like http://dataprotocols.org/json-table-schema/ and http://okfnlabs.org/projects/goodtables/ are conformant with it?)

This could add metadata to a package file that identifies the columnar data types as well as providing to the option to export to, and import from, a CSV file packaged with a tabular metadata file.

danbri:
Yes please :) /cc @rtroncy who also suggested this recently

tfmorris:
Enhancement requests are always welcome.

OKFN (cc @rgrp) has suggested we support the Data Package standard (http://data.okfn.org/doc/data-package), which, as I understand it, the W3C work was based on before they diverged.  They have funding from some foundation to help support this.

I know the W3C likes to make things all standardy and "improved" with their professional standards makers, which sometimes puts them in conflict with more pragmatic folks (vis WHATG), but in this particular case I haven't dug into it to see whether OpenRefine should be picking one to support, support both, or waiting until the pissing contest is over and there's a clear winner.

I don't know the whole backstory, but when I read minutes of a meeting where the original author isn't in attendance and everyone else votes to dismiss all his suggestions, I wonder what's really going on behind the scenes.  http://www.w3.org/2015/09/23-csvw-minutes.html#res

@psychemedia Have you seen anyone else consider supporting the W3C standard if it gets approved?

thadguidry:
@danbri Hasn't Gregg already put lots of work into it and all the tests ? to cover the uses cases that sort of originally inspired it all ? http://www.w3.org/2013/csvw/wiki/Use_Cases

danbri:
The Data Package spec is fine enough for what it is, but it has no interest/capability for mapping from CSVs into entity/relationship graphs (RDF-style, Freebase/KG, schema.org etc etc.). My advice is that you'll find a bigger win with the W3C spec (but I was co-chair, so I have a natural preference). 

[...]

Issue #778, opened 2013-08-09
Support for Simple Data Format Data Packages


rufuspollock:
This is a discussion issue atm. Also a place to ask questions about how best to patch OpenRefine to support this.

[Introduction and overview to Simple Data Format](http://data.okfn.org/standards/simple-data-format) ([full spec](http://www.dataprotocols.org/en/latest/simple-data-format.html))

wetneb:
@rufuspollock We have a first working version of this, just merged. @jackyq2015 has implemented import and export for data packages distributed as Zip files, and also data packages hosted on the web as a JSON file.

The various properties of columns (types, constraints, descriptions, …) are not currently exposed in the UI - this will require more work and coordination.
I also plan to work on import/export for data packages with embedded tabular data. This might require PRs to the java library for data packages.

By the way, thanks for setting up https://github.com/frictionlessdata/test-data, this is very useful.

rufuspollock:
@wetneb awesome 👍 👏 /cc @pwalsh @vitorbaptista

pwalsh:
Amazing @wetneb . Happy to support with reviews and discussion on the [Data Package Java Library](https://github.com/frictionlessdata/datapackage-java).

thadguidry:
@wetneb So what's left to deliver the features on this issue ?  I'd like to see a Task breakdown please with linked issue numbers on each task remaining.

wetneb:
I don't think we have issues for these yet:
* expose column metadata in the UI and let the user edit it
* design how type validation and other constraint validation (uniqueness, format) should be handled in OpenRefine ; implement it
* add support for inline data packages (one JSON file storing both metadata and tabular data)

lwinfree:
Hi! I've noticed that the datapackage import & export functionality is no longer present in the latest versions (3.2 and 3.3) but it was there in 3.0. Are there any plans to re-implement this functionality? If so, is there anything you need help with to get it working?

Thanks!

wetneb:
Hi @lwinfree,

Yes indeed, we removed it because it relied on a non-free library, see https://github.com/frictionlessdata/datapackage-java/issues/26.

It would be great to have this back though! We do not have short-term plans to work on this but would surely welcome PRs in that direction.

[...]

PR #1398, opened 2017-12-24
data package metadata 

jackyq2015:
- The import, export  for data package is done, the json editor works well and there is validation from the frond end.
- Data schema alignment base on the "table schema". for validation, there is ValidateOperationTests.java. you can run it with JUnit, not testng. 
- Also there is a command for validation to return a json response with all the problem. 
- csvw can be added easily with current infrastructure. 
**Next step:**
- Will create the "data validation" UI 
- Add CVSW metadata


jackyq2015:
@wetneb Thank for the review. For the 5 pointed you pointed out:
1. I can add different way to import a data package.
2. The "path" specified in that json is not exist. so it will generate a 404 error. will handle this error more friendly
3. the inline data is not supported from the official data package repo. I will raise an issue and try to create a PR to merge into the official data package.
4. The project can be created with a "data package" or without. The json editor to add or update the data package information before they can export it as a data package. That's the rationally behind the editor. It is not just for display.
5. The design is the ProjectMetadata will be still the primary metadata for openrefine.  The data package and the other forms are extra metadata based on the need of the user. The data package will also NOT replace the current data model for schema. But I agree that we should loop the current column model into the data package. That can be deferred to the stage when user want to use the data package ( for example to export as data package or do the data validation). That will go back to item 4. Still the user  need a UI interface to define the detail data schema for the constrain etc ( the application can only populate basic skeleton of the schema from the current model).

wetneb:
Thanks for the changes. Concerning point 4. and 5., I really think this is the wrong approach.

Adding support for some standard in a piece of software does not mean "add a slot to display and edit the JSON representation of the format described by the specs". It means integrating the specs of the format into the data model of the software to enable a seamless use.

Let's make an analogy to illustrate my point. Suppose you are asked to add FTP support in a web browser that only supports HTTP so far. Your task is **not** to add a window where users can type FTP commands. Your task is to make it possible to click on a "ftp://" link and get the exact same browsing experience that you have with HTTP.

Just look at examples of presentation of data packages, like https://datahub.io/core/s-and-p-500-companies
The JSON is not displayed at all, and for a good reason: that is not what users care about! Users care about their data, the types of their columns, their constraints, and so on. They do not care about the specifics of the Data Package JSON format. They should not have to read the specs to use Data Packages in OpenRefine. It's called "frictionless data" for a reason!

So, yes, a proper integration would take more effort than what you have done. And of course it is fine not to have a perfect implementation in one PR. But the issue with your implementation is that it is not even a step towards a proper integration - it's just not in the right direction at all. If we merge and release this, we will have to undo a lot of things later on and migrating projects to a better implementation will be a nightmare.

I'm sorry to say that at this stage of your work, but I did voice my concerns in #778 originally.

jackyq2015:
The link from datahub is for an existing data package. But in order to generate a data package, the creator have to deal with the detail metadata definition one way or another.  And there is no system can automatically do it for you. Currently the json editor is an interface to allow user to do this from one entry point.  It has a template which can be validated to guide user to create the proper metadata based on the data package schema. It is not just a random json file. One area I can think of is to enhance the way user define the schema though. For other field, it is quite straightforward. For example, user have to input the  "license"  if he want it be part of his metadata.

I am not sure you are saying we should swap out the current Project Metadata with the "data package" metadata. I had disucssion with Thad, and we agreed that the current Project Metadata will be preserved because it's already fully integrated with the OpenRefine and it is easier to extend since it does not tie with any external spec. we can get rid of it and it just not worth to do it. Any external metadata will be built as an extra with the import/export. Also other necessary functionality which can be enabled and leveraged by that particular metadata spec(for data package, that's "data validation"). I already posted a brief design for your review before I started the implementation and haven't got any response yet:
https://docs.google.com/document/d/1RdnpZ1Nr0NWXG4RfNQ7tyDbmkSPSltFMcCI6hxl5484/edit#
  
In #1096 there are too many topics got discussed and most of them is about "data package" vs CSVW. In terms of how it should be integrated into the existing system, my understanding it we should always use the existing resources. in this case two repos were directly introduced and integrated with current internal metadata. As the UT I mentioned above, that also enabled the back end of data validation  using the data package spec. But for the data package data itself, user still need to input somehow.

wetneb:
> The link from datahub is for an existing data package. But in order to generate a data package, the creator have to deal with the detail metadata definition one way or another.

No. The data package can be generated by a tool (for instance some implementations of the specs infer types automatically). So the user does not need to know anything about the way Data Packages are represented in JSON.

> For example, user have to input the "license" if he want it be part of his metadata.

This only works for very simple, unstructured parts of the metadata, but that is completely impractical for something like the tabular schema.

> I am not sure if you are saying we should swap out the current Project Metadata with the "data package" metadata. I had disucssion with Thad, and we agreed that the current Project Metadata will be preserved because it's already fully integrated with the OpenRefine and it is easier to extend since it does not tie with any external spec.

Well, I think this is a big mistake. This should have been discussed publicly. One one hand you have introduced these metadata fields in OR 2.8 (supposedly following Dublin Core), and now you are introducing duplicates of these fields - I don't see why this would be practical at all for users. This is going to be confusing (for instance, the metadata that is displayed in the project list will not be exported to a data package… as a user I would find that very frustrating if I go the extra mile to fill these fields).

> I already posted a brief design for your review before I started the implementation and haven't got any response yet:
> https://docs.google.com/document/d/1RdnpZ1Nr0NWXG4RfNQ7tyDbmkSPSltFMcCI6hxl5484/edit#

Oh, I did not see that! Where did you post it? It should really have been on GitHub so that people who got involved in #1096 could comment. If I had seen that earlier I could have warned in advance that this was the wrong approach.

> In terms of how it should be integrated into the existing system, my understanding it we should always use the existing resources.

Yes exactly! And guess what, OpenRefine already had a good basis for a tabular schema! We have the Column class, that already holds some limited metadata about the column, such as its name and its reconciliation statistics (which service type was it reconciled against, importantly). The tabular schema should be baked into that, by adding new fields to these classes. That makes the integration much tighter, much more transparent for the user, and much easier to integrate with the UI and the operations and changes.

> But for the data package data itself, user still need to input somehow

No! The user could simply click on a column and have some operation there to set the datatype there, for instance. (Or infer it automatically). The user should simply not have to interact with the JSON at all.

> Regarding the workflow you mentioned, I think one thing we are missing is to incorporate the current "column model" and the "data package schema" then we can get the final schema.

This should really have been the starting point of the integration. That is what I tried to explain in #1096. (but clearly I failed…)

Just step back and think about why and how standards like CSVW and Data Packages were created. The goal of these standards is to enhance data files (mostly tabular ones) with rich metadata. But shifting the open data ecosystem from CSV to something richer takes a lot of effort: you need to adapt the tools, educate the data providers, the data transformers, and the users. If these users are required to read the specifications to use the format, this is never going to work! Many users do not even know what JSON is, and they should not need to. OpenRefine is used by these people, and OpenRefine is (so far) really good at hiding a lot of the complexity of data formats (typically, with its JSON and XML importers). So let's keep it that way, and provide a clean integration of Data Packages in the model, where both project metadata and the schema can be edited where users expect it, and not in an obscure JSON editor. People do not need OpenRefine to edit some JSON, they can already do it very well with any decent text editor.

jackyq2015:
I know the schema can be inferred from the data itself. The current  official "data package" already implemented this. But that cannot fit well in the OpenRefine where we already have the schema(it can be used to help combine the current column model into final schema though) Also the infer is NOT designed to generate the "data package". It is an educated guess based on the data set. For example, infer cannot know your constrain, not even mention other metadata fields other than schema.

For the metadata 2.8 version, I confirmed with @ettorerizza, @magdmartin and @thadguidry  on the github issue publicly. It will be sorry again that you did not see it. we cannot take advice from many users and also remember that there is no authority to define the internal metadata. what I did is totally based on the user input, field by field. The point is that I did not decide what metadata should be included. There is duplication but we discussed in "tag system" PR, we will clean it up gradually after reach the consensus. we cannot get rid of current metadata for now. One is because it was tight with other model, another reason is still consensus is hard to reach. Like the ongoing debate between "data package" and CSVW. Every one has different opinion and bias. Also think about if CSVW is introduced in the future, there is for sure lots of duplicate between metadata models. Even we can get rid of current project metadata. we will still come to a point to decide which spec is better/more useful if we want to only keep only 1 metadata and thus remove the duplicate. 

The importer function is straightforward and user can use it to pull the data and import into OR without knowing the detail of data package. For exporter, if the user want to create a "data package" he has to fill in the information especially when the project is not created from existing data package . Schema can be taken from the current data model but he still need to fill in the constraints and other other simple fields.

wetneb:
> Also think about if CSVW is introduced in the future, there is for sure lots of duplicate between metadata models. Even we can get rid of current project metadata. we will still come to a point to decide which spec is better/more useful if we want to only keep only 1 metadata and thus remove the duplicate.

That is exactly my point! This duplication should be removed right now, there is no point in introducing it in the first place.

> For the metadata 2.8 version, I confirmed with @ettorerizza, @magdmartin and @thadguidry on the github issue publicly.

That is a totally different issue - this did not include the schema at all.

> The point is that I did not decide what metadata should be included. There is duplication but we discussed in "tag system" PR, we will clean it up gradually after reach the consensus. we cannot get rid of current metadata for now.

The problem is that by leaving these metadata fields outside OpenRefine's core, you are making it very hard to build any meaningful integration with the rest of the software. This is way too loosely coupled and this is going to hit us back as soon as we try to implement features that will actually be useful to users.

main/src/com/google/refine/model/Project.java
```diff
@@ -270,6 +280,80 @@ public void update() {
...
+                         columnModel);
+                 dp.getPackage().addResource(resource);
+                 schema = new Schema();
+             }
+        } catch (Exception e) {
+            logger.error(ExceptionUtils.getStackTrace(e));
+        }
+        
+        String changeClazzName = change.getClass().getSimpleName();
+        switch (changeClazzName) {
```
wetneb: Sorry, but this is really not acceptable: it is not extensible at all. What if an extension declares a new operation that changes the columns? It will not be able to register the operation in your switch.

I am afraid it is going to be very hard to recover your implementation - the whole architecture needs to be started again from scratch. Again, the clean solution here is to store the schema in the `Column` objects themselves (or in a class that is linked directly from them).

main/src/com/google/refine/model/medadata/MetadataFactory.java
```diff
@@ -11,4 +33,46 @@ public static IMetadata buildMetadata(MetadataFormat format) {
...
+    /**
+     * build an empty Data Package Metadata
+     * @return
+     */
+    public static DataPackageMetadata buildDataPackageMetadata() {
+        return (DataPackageMetadata) buildMetadata(MetadataFormat.DATAPACKAGE_METADATA);
+    }
+    
+    /**
+     * Build an empty data package metadata, then populate the fields from the Project Metadata
```
wetneb: Sorry but again, that is not a solution - what if the metadata changes later on? For instance:
* `buildDataPackageMetadata` is called
* the user edits the schema
* the user edits project metadata
=> the data package metadata is out of sync from the project metadata

At this stage, calling `buildDataPackageMetadata` again would discard the changes made to the schema.

Again, the clean solution for this is just not to duplicate this metadata in the first place.

wetneb:
@jackyq2015 @thadguidry @ostephens @magdmartin I think it would be good to have a skype meeting to talk about all this in person and try to find a solution. I feel bad about obstructing Jacky's work because he has clearly put a lot of effort into this but at the same time I am very concerned by the shape of this PR.

thadguidry:
@wetneb we can now if you want

wetneb:
@thadguidry I'm available too! Just sent you an invite.

wetneb:
Example workflow that should (ideally!) be supported:
- create a project from a bare CSV with no metadata
- rename and reorder the columns
- add types on some columns
- enhance the project metadata (from the project list, for instance)
- reconcile a column to a reconciliation service, with some type
- export the project to a data package

The data package should contain:
- the updated project metadata
- the tabular schema induced by the new datatypes and the reconciliation type

main/src/com/google/refine/model/Column.java
```diff
@@ -47,13 +48,25 @@
...
     final private int       _cellIndex;
     final private String    _originalName;
     private String          _name;
     private ReconConfig     _reconConfig;
     private ReconStats      _reconStats;
     
+    // from data package metadata Field.java:
+    private String type = "";
+    private String format = Field.FIELD_FORMAT_DEFAULT;
+    private String title = "";
```
wetneb: What is the difference between the `title` attribute and the `_name` that we already have?
Also, shouldn't these attributes follow the naming convention of the previous ones (start with `_`)?

main/src/com/google/refine/model/Column.java
```diff
@@ -47,13 +48,25 @@
...
+
 public class Column implements Jsonizable {
     final private int       _cellIndex;
     final private String    _originalName;
     private String          _name;
     private ReconConfig     _reconConfig;
     private ReconStats      _reconStats;
     
+    // from data package metadata Field.java:
+    private String type = "";
```
wetneb: Do we want to allow any type or restrict them to an enumeration?

main/src/com/google/refine/model/Column.java
```diff
@@ -168,4 +231,23 @@ static public Column load(String s) throws Exception {
...
+    
+    public <Any> Any castValue(String value)
+            throws InvalidCastException, ConstraintsException {
+        if (this.type.isEmpty()) {
+            throw new InvalidCastException();
+        } else {
+            try {
+                // Using reflection to invoke appropriate type casting method from the
+                // TypeInferrer class
+                String castMethodName = "cast" + (this.type.substring(0, 1).toUpperCase() + this.type.substring(1));
```
wetneb: It seems quite dangerous to do this if we allow types to be any string.

main/src/com/google/refine/model/medadata/validator/checks/MaximumConstraint.java
```diff
@@ -26,9 +26,9 @@ public boolean checkCell(Cell cell) {
         boolean valid = true;
         
         try {
-            Comparable value = field.castValue(cell.value.toString(), false);
+            Comparable value = column.castValue(cell.value.toString());
             // return this - threashold
-            if (value.compareTo(field.castValue(threashold, false)) > 0)
+            if (value.compareTo(column.castValue(threashold)) > 0)
```
wetneb: there is a typo here: `threashold` should be `threshold` (and at other places)

main/src/com/google/refine/model/Column.java
thadguidry: @wetneb He's just following the spec... Title is also typical of an extra field seen from news organizations as well.
https://frictionlessdata.io/specs/table-schema/

"name": "name of field (e.g. column name)",
      "title": "A nicer human readable label or title for the field",

main/src/com/google/refine/model/Column.java
thadguidry: @wetneb I'd rather not impose restriction within our column model for now against Types ( I know I know, folks will whine and complain, but I'm against this directly in our modeling)  Later we can bring about rules and sets (and extensions can even impose them as well).  That was the long term plan back in the day and is the better way to handle enumerations or restrictions as well as casting.

main/src/com/google/refine/model/Column.java
thadguidry: why are Trys dangerous in Java ?  Sorry, I thought they avoid danger ? :)

thadguidry:
@wetneb The Type's are that of more Semantic types and less of rigid types.  For instance, we want folks to be able to type a column against the concept of http://schema.org/funder or for that matter a http://schema.org/CreativeWork  In the real world, sometimes Types are uplifted as "property names" with expected rules / restrictions against domains and hard Types like http://pending.schema.org/maximumAttendeeCapacity having an expected type of Integer.  In our model, we want to allow BOTH, and let a Semantic overlay or extension or further data package managers apply further Semantic operations for Type casting, refinement, reconciling, etc.  By not constraining the Types in our model, we allow the most flexibility for others to expand on our core.

main/src/com/google/refine/model/medadata/validator/checks/MaximumConstraint.java
jackyq2015: fixed

main/src/com/google/refine/model/Column.java
jackyq2015: Right, that's the spec. it is kind of weird when I first saw the title for the column. But after 2nd thought, it make sense.

wetneb: oh yeah, okay, thanks!

main/src/com/google/refine/model/Column.java
wetneb: it's not the trys, it's just that the introspection features should be used only when needed, and this does not seem to be a case where they should be used. I would go for an enum for the type and a switch for the cast.

wetneb:
@thadguidry ah now I understand why you want to keep types as strings! That's because you are mixing up two things: the data type and the RDF type.

The data type says whether the column is made of integers, strings, booleans and so on. The Data Package format does specify the acceptable values for that, see https://frictionlessdata.io/specs/table-schema/#types-and-formats. So we should just take do the same: use an enumeration (which might not initially contain all the types they support).

The RDF type (called "Rich Type") is a *different* sort of type that can be provided, which can accept things like "http://schema.org/funder".

main/src/com/google/refine/model/Column.java
wetneb: @thadguidry concretely, what is it going to look like as a user? if I want to set a type for a column, will I see a text field where I can input any string that will represent the type? it seems sloppy to me.

jackyq2015:
All the types are defined in the Field.java from official repo which follows the spec. According to the spec, the rdfType is another properly parallel  with the type. But it is just a definition. Also the officail repo does not have this included( it can be included, but will have no reference to it)

main/src/com/google/refine/model/Column.java
thadguidry: @wetneb I had thought about just starting with that for now, yes ?  And then expand or let your recon or extensions begin to change its form, say for instance to a drop down of enumerated values to pick from.  Or a changing enumerated list of values depending on the context switch in a particular extension or domain need.

thadguidry: @wetneb this could even be driven by an expression, no ?  Imagine providing some rules via a Python or GREL expression that controls the enumerated values available.  I feel that is very powerful and rather than having Data Types being baked in to our core, we allow for expanding domain usage.  This is reminiscent of Wikidata and Schema.org's evolving Types / Properties as well.  In fact, I am on the working group for IoT.Schema.org and there is MUCH debate on mapping between vocabularies regarding various Actions and Controls of Sensors and Systems.

wetneb: @thadguidry you are mixing up two things (see my comment below). Data types should not be free-form strings or even URIs.

thadguidry:
@wetneb OK, so when you said Type you were saying Data Type in a programming language like our Java.  And Yes, I was talking about BOTH.  Yeah, the Data Package standard has a field called Type (which is really Data Type ) and it could be enumerated and controlled.  The problem that I have is that of ... where's the fucking controlled list ? https://frictionlessdata.io/specs/table-schema/#types-and-formats  that doesn't really spell it out and says its based on json-schema but with modifications and then where are those?  What does the spec say is the final controlled list of allowed Types ?  I couldn't find it or see it through the weeds.  Did you ?

wetneb:
@thadguidry yes the controlled list is just there in the specs - "string", "number", "integer", "boolean" and so on.

@jackyq2015 okay I had not seen that this library uses Strings to encode these types too. All I am saying is that we should validate this String against the list of supported types.

jackyq2015:
comment removed

build.xml
```diff
@@ -30,6 +30,9 @@
 
     <property name="webapp.dir" value="${main.dir}/webapp" />
     <property name="webapp.src.dir" value="${main.dir}/src" />
+    <property name="table-schema.src.dir" value="${basedir}/tableschema-java/src/main/java/" />
```
thadguidry: For some reason when I try to build on Windows it cannot find this path and complains.

jackyq2015: Did you do git clone --recursive https://github.com/OpenRefine/OpenRefine

wetneb: Again we should really use jars to avoid that!

thadguidry: @jackyq2015 change that... I don't like this setup... jars preferred.  So what if its a bit of maintenance.  This is more pain than its worth.

thadguidry:
Cannot launch OpenRefine on Windows with Jacky's branch (although Jetty runs but the app doesn't start and gives 503 Service Unavailable.  He will debug later today or tomorrow when he has time.  Sent him email with some details.

main/src/com/google/refine/commands/Command.java
```diff
@@ -52,11 +52,12 @@
 
 import com.google.refine.Jsonizable;
 import com.google.refine.ProjectManager;
-import com.google.refine.ProjectMetadata;
 import com.google.refine.RefineServlet;
 import com.google.refine.browsing.Engine;
 import com.google.refine.history.HistoryEntry;
 import com.google.refine.model.Project;
+import com.google.refine.model.medadata.IMetadata;
```
thadguidry: this isn't needed.

main/src/com/google/refine/exporters/TemplatingExporter.java
```diff
@@ -42,6 +42,7 @@
 import org.json.JSONObject;
 import org.json.JSONWriter;
 
+import com.google.refine.ProjectManager;
```
thadguidry: wait, why is this needed here ?

main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -255,107 +281,26 @@ public void update(long bytesRead, long contentLength, int itemCount) {
                     
                 } else if (name.equals("download")) {
                     String urlString = Streams.asString(stream);
-                    URL url = new URL(urlString);
-                    
-                    JSONObject fileRecord = new JSONObject();
-                    JSONUtilities.safePut(fileRecord, "origin", "download");
-                    JSONUtilities.safePut(fileRecord, "url", urlString);
+                    download(rawDataDir, retrievalRecord, progress, fileRecords, update, urlString);
+                } else if (name.equals("data-package")) {
```
thadguidry: we could flip this around to avoid an NPE and it will just return false ... but maybe throwing that NPE is better for debugging for now ? https://www.codacy.com/app/OpenRefine/OpenRefine/file/13416576553/issues/source?bid=6129680&fileBranchId=6129680#l285

main/src/com/google/refine/model/medadata/AbstractMetadata.java
```diff
@@ -0,0 +1,77 @@
...
+
+import org.apache.commons.beanutils.PropertyUtils;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.json.JSONWriter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class AbstractMetadata implements IMetadata {
+    private final static Logger logger = LoggerFactory.getLogger("AbstractMetadata");
```
thadguidry: I don't see this used.

main/src/com/google/refine/model/medadata/validator/checks/AbstractValidator.java
```diff
@@ -0,0 +1,122 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.refine.model.Cell;
+import com.google.refine.model.Column;
+import com.google.refine.model.Project;
+import com.google.refine.model.Row;
+import com.google.refine.model.medadata.validator.ValidatorSpec;
+
+import io.frictionlessdata.tableschema.Field;
```
thadguidry: where is this used ?

main/src/com/google/refine/importing/ImportingUtilities.java
jackyq2015: I think we should put name ahead if we are sure the name supposed to have a value flip it around silently hide the issue and redirect the execution to last default case.

main/src/com/google/refine/commands/Command.java
jackyq2015: removed

main/src/com/google/refine/model/medadata/AbstractMetadata.java
jackyq2015: removed

main/src/com/google/refine/model/medadata/validator/checks/AbstractValidator.java
jackyq2015: removed

main/src/com/google/refine/importing/ImportingUtilities.java
thadguidry: yeah, my thoughts also, better to not hide in this case.  Also... isn't the name without hypen per the spec ?  
A Data Package descriptor MUST be a valid JSON object. (JSON is defined in RFC 4627). When available as a file it MUST be named **datapackage.json** and it MUST be placed in the top-level directory (relative to any other resources provided as part of the data package).

thadguidry:
@jackyq2015 Cleanup on all of these "blue colored" issues (skip the yellow), since now you have reverted the code mostly.  See https://www.codacy.com/app/OpenRefine/OpenRefine/pullRequest?prid=1186996

thadguidry:
@jackyq2015 Can we change to using randomX or https://github.com/coddec/random-required or something else instead of the random.org , in order to fix the constantly failing testUrlCaching ? https://ci.appveyor.com/project/openrefine/openrefine/build/1.0.136#L345  

a few others are listed here: https://alternativeto.net/software/randomorg/?license=free

thadguidry:
@jackyq2015 BUG:  Autosave is failing ... 
```
15:09:24.188 [            refine_server] Starting Server bound to '127.0.0.1:3333' (0ms)
15:09:24.190 [            refine_server] refine.memory size: 1400M JVM Max heap: 1407188992 (2ms)
15:09:24.198 [            refine_server] Initializing context: '/' from 'C:\Users\THAD\OpenRefine\main\webapp' (8ms)
15:09:24.433 [            refine_server] Failed to use jdatapath to detect user data path: resorting to environment variables (235ms)
15:09:24.433 [            refine_server] Failed to use jdatapath to detect user data path: resorting to environment variables (0ms)
15:09:24.442 [                   refine] Starting OpenRefine 2.8 [TRUNK]... (9ms)
15:09:28.932 [                   refine] POST /command/core/load-language (4490ms)
15:09:28.951 [                   refine] GET /command/core/get-preference (19ms)
15:09:28.954 [                   refine] POST /command/core/load-language (3ms)
15:09:29.027 [                   refine] POST /command/core/get-importing-configuration (73ms)
15:09:29.108 [                   refine] GET /command/core/get-all-project-tags (81ms)
15:09:29.118 [                   refine] GET /command/core/get-all-project-metadata (10ms)
15:09:29.198 [                   refine] GET /command/core/get-languages (80ms)
15:09:29.238 [                   refine] GET /command/core/get-version (40ms)
java.lang.NullPointerException
        at com.google.refine.model.medadata.AbstractMetadata.isDirty(AbstractMetadata.java:38)
        at com.google.refine.io.FileProjectManager.saveToFile(FileProjectManager.java:313)
        at com.google.refine.io.FileProjectManager.saveWorkspace(FileProjectManager.java:271)
        at com.google.refine.ProjectManager.save(ProjectManager.java:232)
        at com.google.refine.RefineServlet$AutoSaveTimerTask.run(RefineServlet.java:92)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
15:14:24.564 [       FileProjectManager] Failed to save workspace (295326ms)
```

jackyq2015:
Please pull again

thadguidry:
@jackyq2015 On export of the Data Package... if we could, lets sort the Object list to something sensible the first time... like Name, Title, Description, Version, License, etc..  if its not too much trouble...otherwise just use that default Sort of Ascending in that Object grid control the UI has on the Object level.

jackyq2015:
will take a look about the order request. The JSON is essentially unordered. but will do some research and get back.

All the validation you mentioned is based on the data package schema. but it has the non-restrict mode, means you can disobey and save it, though it is not recommended. It may cause issue for other application follow strictly to the spec.

For now, only JSON URL is supported. will add the zip file support later on.

thadguidry:
@jackyq2015

- Hmm, then don't worry about the JSON ordering.
- Then don't worry about if Data Package spec wants to be strict about "name", it is fine (but I disagree with their validation on some of the properties)  So let's leave as is.
- Great that zip file Data Packages will be able to be imported later.

main/src/com/google/refine/importing/ImportingUtilities.java
jackyq2015: The name is just a key for internal exchange(from the javascript), not the file name itself

wetneb:
@jackyq2015 let me know when the PR is ready for a new code review :)

jackyq2015:
It is ready. :)

wetneb:
So, I have found time to give it a go. I haven't reviewed the code yet, just trying it as a user for now:

* In the import UI, the data package importer should be integrated just like the other importers: it should be possible to import a datapackage both from a file, a URL, and the clipboard. So the "Data Package (JSON URL)" option should be removed.

* It does not look like zipped data packages are supported but I thought you intended to support that - if it is supported, how?

* The import code looks a bit brittle. Importing http://pintoch.ulminfo.fr/38972ea28/datapackage.json gave me the following error in the logs:
```

09:10:48.444 [                   refine] POST /command/core/importing-controller (5015ms)
09:10:48.506 [..ata.DataPackageMetadata] Data Package metadata loaded (62ms)
Exception in thread "Thread-7" org.json.JSONException: JSONObject["keywords"] not found.
	at org.json.JSONObject.get(JSONObject.java:471)
	at org.json.JSONObject.getJSONArray(JSONObject.java:618)
	at com.google.refine.importing.ImportingUtilities.populateDataPackageMetadata(ImportingUtilities.java:1163)
	at com.google.refine.importing.ImportingUtilities.createProjectSynchronously(ImportingUtilities.java:1116)
	at com.google.refine.importing.ImportingUtilities.access$200(ImportingUtilities.java:114)
	at com.google.refine.importing.ImportingUtilities$7.run(ImportingUtilities.java:1070)
09:10:49.466 [                   refine] POST /command/core/get-importing-job-status (960ms)
```
This error stalls the UI without anything being reported to the user.
Same issue with http://pintoch.ulminfo.fr/347584eb7/datapackage.json .

Can you point me to any data package that is imported correctly? It would be useful to review the rest.

jackyq2015:
The zip is supported from the local file import. The reason a dedicated menu is given is for data package url is that user has to indicate it's a data package instead of json file.  I will add an entry for importing from paste. 

The keywords is mandatory according to the spec. The json you provided is not qualified json. But I can change it and let it go without it.

wetneb:
Thanks for the example data package that works. It's great to see that everything from the Data Package seems to be imported correctly and to see the description show up by itself on the project list. I noticed a small issue though: any tags entered while creating the project are overwritten by the tags contained in the Data Package. Wouldn't it be better to merge the two lists?

About the mandatory keywords: it's your call, you can choose to require keywords - you just need to report the error to the user somehow. You cannot expect people to check the logs for that, it needs to be properly exposed in the web UI. That is true for all required fields. Your code should just not throw any exception when used normally. Never trust user input.

Another example: when importing https://pkgstore.datahub.io/core/cofog/datapackage_zip/data/ed438b8b276d0169131523969f4e3582/datapackage_zip.zip , I get this exception:
```
09:27:58.272 [                   refine] POST /command/core/importing-controller (8439ms)
Exception in thread "Thread-7" java.lang.NullPointerException
	at io.frictionlessdata.tableschema.TypeInferrer.infer(TypeInferrer.java:161)
	at com.google.refine.importing.ImportingUtilities.createProjectSynchronously(ImportingUtilities.java:1132)
	at com.google.refine.importing.ImportingUtilities.access$200(ImportingUtilities.java:114)
	at com.google.refine.importing.ImportingUtilities$7.run(ImportingUtilities.java:1070)
09:27:59.297 [                   refine] POST /command/core/get-importing-job-status (1025ms)

```
Again this should be reported to the user in a clean way. If the underlying Data package library is buggy then fix it or just wrap your calls to it with try/catch blocks.

About the way the import is presented: there already are cases in which a file can be handled by multiple importers - I would just tweak format detection so that the Data Package importer is selected for JSON payloads that look like Data Packages, and fall back on the standard JSON importer otherwise. It is going to be much more usable.

For the zip import: okay, I think I got it working now. But it's a bit curious because at first it looks like the Zip is being treated like any other Zip file. It's not clear which files in the zip should be selected, for instance. Only the JSON files were selected by default in my case - I had to tick the main CSV one and manually set that it was a CSV, not a TSV. It's not a huge deal but it means people really have to understand how a Data Package is made to use the feature.

jackyq2015:
fixed the issue for zip:
https://pkgstore.datahub.io/core/cofog/datapackage_zip/data/ed438b8b276d0169131523969f4e3582/datapackage_zip.zip 

For clipboard integration, it is hard to for program to detect it's a data package unless user say so. Those tags can be optional and can be there or cannot be there. For the spec, the ground to tell it's a data package is based on its name datapackage.json. obviously it can't apply to the clipboard.

jackyq2015:
Merged the tags from user input. 

@wetneb For the entry for "JSON URL", I would like to make it explicitly to the user that it deal with "data package" json metadata. User HAS to know what is "data package" but not necessarily the detail of it. Hiding those  import logic does not bring benefit, instead, it is confuse user sometimes. Also data package is not a data format(if it is, then it make sense to handle all the logic for the user as you mentioned), it is just how the data and metadata is organised. The same idea can apply to the clipboard. I would not think it's a normal user case you paste it from somewhere. there only 2 cases you have the existing datapackage.json or you don't have it:
- If you have it(could be from URL, or it is zipped in the zip file locally), you can use the current importing options to "import" it. 
- If you don't have the datapackage.json,  you either import the data as old ways then export to the data package if you need a data package and you will get a datapackage.json automatically. Or you need to create your own  datapackage.json by yourself(means you manually create it and all on yourself. :)

wetneb:
@jackyq2015 thanks for merging the tags.

For the local and clipboard import: if you don't see how to do it that's fine, I can take that on once this PR is merged. The clipboard import is still useful in some cases, for instance the ones with inline resources (typically from Wikimedia commons). In this case the Data Package is indeed a data format in itself. So let's keep it like this for now and I will migrate the importer to a standard one when I do #1269 (but that should be done before the next release otherwise users will be confused).

I will now try to review the code. It would really help if you could avoid merging master in your branch because it makes code review a lot harder. If you want to keep your branch up to date with master, it would be nicer if you could just rebase your branch instead. That would keep the history a lot cleaner. And also, please try not put unrelated changes in the PR (such as updating the dependencies or removing the "creator" field from the project list). Doing separate PRs and rebasing once they get merged works really well.

thadguidry:
@wetneb I've added notes about those merge points and PR expectations to our Developer Wiki as well.. https://github.com/OpenRefine/OpenRefine/blob/master/CONTRIBUTING.md#how-to-submit-prs-pull-requests-patches-and-bug-fixes

extensions/sample/.classpath
```diff
@@ -1,6 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<classpath>
-	<classpathentry kind="src" path="src"/>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.6"/>
-	<classpathentry kind="output" path="module/MOD-INF/classes"/>
-</classpath>
```
wetneb: Same comment.

main/src/TableSchemaValidator.json
```diff
@@ -0,0 +1,213 @@
...
+          "type": "schema",
+          "context": "body",
+          "weight": 7,
+          "message": "The value {value} in row {row_number} and column {column_number} does not conform to the maximum length constraint of {constraint}",
+          "description": "A lenght of this field value should be less or equal than schema constraint value.\n\n How it could be resolved:\n - If this value is not correct, update the value.\n - If value is correct, then remove or refine the `maximumLength` constraint in the schema.\n - If this error should be ignored disable `maximum-length-constraint` check in {validator}."
+        }
+
+
+    }
+}
```
wetneb: I don't think this file is in an appropriate location - this folder is for Java source files.

main/src/com/google/refine/ProjectManager.java
```diff
@@ -307,13 +308,10 @@ public int compare(SaveRecord o1, SaveRecord o2) {
...
-
-            for (int i = 0;
-                 i < records.size() &&
-                    (allModified || (LocalDateTime.now().atZone(ZoneId.systemDefault()).toInstant().toEpochMilli() - 
-                            startTimeOfSave.atZone(ZoneId.systemDefault()).toInstant().toEpochMilli() < QUICK_SAVE_MAX_TIME));
+             
+            for (int i = 0;i < records.size() &&
+                    (allModified || (ChronoUnit.MILLIS.between(startTimeOfSave, LocalDateTime.now()) < QUICK_SAVE_MAX_TIME));
                  i++) {
-
```
wetneb: So again this file is full of changes that do not seem related to the feature - if I remember correctly they are meant to address issue #1380 so it looks fine to me, but please use a separate PR next time.

main/src/com/google/refine/commands/GetAllPreferencesCommand.java
```diff
@@ -52,9 +52,9 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
         Project project = request.getParameter("project") != null ? getProject(request) : null;
-        PreferenceStore ps = project != null ? 
-                project.getMetadata().getPreferenceStore() : 
-                ProjectManager.singleton.getPreferenceStore();
+        
+        // project level has no specific preference
+        PreferenceStore ps = ProjectManager.singleton.getPreferenceStore();
                 
         try {
             response.setCharacterEncoding("UTF-8");
```
wetneb: I'm not sure I understand this change - do we used to have project-level preferences that have been removed by this PR?

main/src/com/google/refine/commands/project/SetMetadataCommand.java
```diff
@@ -0,0 +1,80 @@
...
+            if (StringUtils.isBlank(jsonContent)) {
+                respond(response, "error", "json cannot be blank");
+                return;
+            }
+            
+            IMetadata metadata = MetadataFactory.buildMetadata(format);
+            InputStream in = IOUtils.toInputStream(jsonContent, "UTF-8");
+            metadata.loadFromStream(in);
+            metadata.validate();
+
```
wetneb: It's not clear to me how this metadata is saved to the project - it looks like it is just validated and then it is not used - can you explain what is happening here?

main/src/com/google/refine/importing/UrlRewriter.java
```diff
@@ -33,12 +33,32 @@
...
+            this.download = download;
+            this.metaDataFormat = metaDataFormat;
+        }
     }
     
-    public Result rewrite(String url);
+    public List<Result> rewrite(String url) throws MalformedURLException, IOException;
+    
+    public boolean filter(String url);
 }
```
wetneb: It looks like this interface is used to translate relative paths to full URLs - maybe it would be worth putting a few comments explaining what it does (for instance, why does it return a List of Results instead of just a Result)?

thadguidry:
@wetneb A lot of the changes are IDE changes for Eclipse (to improve tooling support for others pulling our project into Eclipse).  But I'm wondering if some of this is really necessary in our repo ?  Typically .settings , .project , and others are not pushed into Git repos, but we have historically added config and preferences for Eclipse and other IDE's under https://github.com/OpenRefine/OpenRefine/tree/master/IDEs/eclipse 

@jackyq2015 Can the .settings and .classpath and others be moved under that path to avoid the mixing of source versus IDE config and preferences ?  Did some of these get pulled into your branch from Tony's DB work ?

wetneb:
@thadguidry yes I can see some of these changes are related to eclipse but I am not sure why they are necessary. As I said I would rather have liked not to have them at all in this PR at all because I think we already have enough things to discuss here…

thadguidry:
@wetneb Right, and I agree. I actually want you guys to work on this and add .settings and .classpath and even .project to our gitignore file to prevent these things from happening.

jackyq2015:
For the Eclipse setting file, I think it make sense to keep it in the repo since OR is a single project and you really want to programmer can setup easily without too much effort. For the project you given above, they have the gradle to help them handle the setup and it is easier to maintain just one build system instead of 2(but gradlle can generate the eclipse project).

wetneb:
@jackyq2015 okay, then could you put some infos about that on the wiki - how to recover these sub-projects with gradle?

build.xml
```diff
@@ -140,6 +141,10 @@
             <classpath refid="webapp.class.path" />
         </javac>
         <copy file="${webapp.src.dir}/log4j.properties" tofile="${webapp.classes.dir}/log4j.properties"/>
+        <copy file="${webapp.src.dir}/datapackage-template.json" tofile="${webapp.classes.dir}/datapackage-template.json"/>
+        <copy file="${webapp.src.dir}/TableSchemaValidator.json" tofile="${webapp.classes.dir}/TableSchemaValidator.json"/>
```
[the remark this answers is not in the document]
jackyq2015: Files moved

extensions/gdata/.project
```diff
@@ -1,31 +0,0 @@
...
-			<arguments>
-			</arguments>
-		</buildCommand>
-	</buildSpec>
-	<natures>
-		<nature>org.eclipse.jdt.core.javanature</nature>
-		<nature>org.eclipse.wst.common.project.facet.core.nature</nature>
-		<nature>org.eclipse.wst.jsdt.core.jsNature</nature>
-	</natures>
-</projectDescription>
```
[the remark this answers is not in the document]
jackyq2015: added javascript facet to indicate having javascrip and eclipse can handle the check, highlight etc

extensions/sample/.classpath
jackyq2015: sample project is not necessary

main/src/TableSchemaValidator.json
jackyq2015: moved

main/src/com/google/refine/commands/GetAllPreferencesCommand.java
jackyq2015: No. nothing is changed. but just adding some comments to make it clear

main/src/com/google/refine/importing/UrlRewriter.java
jackyq2015: comment added

main/src/com/google/refine/commands/project/SetMetadataCommand.java
jackyq2015: /**
 * We are noting save other metadata file for now, only the ProjectMetaData is saved.
 * But we keep this command in case we want to save the medadata which requires to be saved for different reasons. 
 * For example we cannot restore the original metadata if don't save it
 *
 */

jackyq2015:
@wetneb We need to have the gradle build firstly then eclipse project can be generated from that.
https://docs.gradle.org/current/userguide/eclipse_plugin.html

thadguidry:
@jackyq2015 I gave you a headstart on it over 1 year ago !  Its in this branch https://github.com/OpenRefine/OpenRefine/tree/java8-refactor

/gradle/wrapper
build.gradle
gradlew
gradlew.bat

Starting usage docs I put inside the https://github.com/OpenRefine/OpenRefine/blob/java8-refactor/build.gradle

But yeah, there's more that can be done to fully support Eclipse config within the build.gradle file. I just haven't done much with it myself, but have at it !

wetneb:
@jackyq2015 okay, so if it needs more work, let's keep all the eclipse/gradle-related stuff for a different PR?

jackyq2015:
@wetneb Yes. Gradle migration is not minor change. we have to factor in the change both for the build script and the bootstrap script refine/refine.bat etc. Also we can discuss we should go gradle or maven?

wetneb:
@jackyq2015 great. So can you remove the eclipse-related changes in this PR?
Yes, we can open a different issue for the gradle vs. maven discussion.

thadguidry:
My preference is to go Gradle only.  Its an easier setup for developers. (once I do the work of getting our Ant converted over to a nicer build.gradle file)

wetneb:
@jackyq2015 I have left comments on the PR, flagging many of these files - can you see them?
Here are some instances of such files: `.gitattributes`, `.settings/.jsdtscope`, `.settings/org.eclipse.wst.jsdt.core.prefs`, `.settings/org.eclipse.wst.validation.prefs`, ` broker/appengine/.project`, `broker/appengine/.classpath`, `extensions/gdata/.classpath`… I understand some of them need to be updated to include the new dependencies, but I really don't see why so many of them should be deleted entirely?

The idea is pretty simple: you can view the difference between this PR and master here:
https://github.com/OpenRefine/OpenRefine/pull/1398/files
All of the changes that you see there should be required to implement data packages. For simplicity, it would be great if you could reply to the comments I have made there. For instance, I have added a comment on the `.gitattributes` that you deleted:

diff

broker/appengine/.classpath
```diff
@@ -1,17 +0,0 @@
...
-	<classpathentry kind="lib" path="/grefine/webapp/WEB-INF/lib/json-20100208.jar" sourcepath="/grefine/webapp/WEB-INF/lib-src/json-20100208-sources.jar"/>
-	<classpathentry kind="lib" path="/grefine/webapp/WEB-INF/lib/httpclient-4.0.1.jar" sourcepath="/grefine/webapp/WEB-INF/lib-src/httpclient-4.0.1-sources.jar"/>
-	<classpathentry kind="lib" path="/grefine/webapp/WEB-INF/lib/httpcore-4.0.1.jar" sourcepath="/grefine/webapp/WEB-INF/lib-src/httpcore-4.0.1-sources.jar"/>
-	<classpathentry kind="lib" path="/grefine/webapp/WEB-INF/lib/slf4j-api-1.5.6.jar" sourcepath="/grefine/webapp/WEB-INF/lib-src/slf4j-api-1.5.6-sources.jar"/>
-	<classpathentry combineaccessrules="false" kind="src" path="/grefine-broker"/>
-	<classpathentry kind="lib" path="/grefine/webapp/WEB-INF/lib/rhino-1.7R2.jar"/>
-	<classpathentry kind="lib" path="/grefine/webapp/WEB-INF/lib/velocity-1.5.jar"/>
-	<classpathentry kind="lib" path="/grefine/webapp/WEB-INF/lib/commons-collections-3.2.1.jar"/>
-	<classpathentry kind="output" path="module/MOD-INF/classes"/>
-</classpath>
```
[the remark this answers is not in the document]
jackyq2015: The broker is not being used any more.

extensions/gdata/.classpath
```diff
@@ -1,32 +0,0 @@
...
-	<classpathentry kind="lib" path="module/MOD-INF/lib/google-api-client-1.20.0.jar"/>
-	<classpathentry kind="lib" path="module/MOD-INF/lib/google-api-client-servlet-1.20.0.jar"/>
-	<classpathentry kind="lib" path="module/MOD-INF/lib/google-api-services-drive-v2-rev168-1.20.0.jar"/>
-	<classpathentry kind="lib" path="module/MOD-INF/lib/google-api-services-fusiontables-v2-rev3-1.20.0.jar"/>
-	<classpathentry kind="lib" path="module/MOD-INF/lib/google-http-client-1.20.0.jar"/>
-	<classpathentry kind="lib" path="module/MOD-INF/lib/google-http-client-jackson2-1.20.0.jar"/>
-	<classpathentry kind="lib" path="module/MOD-INF/lib/google-oauth-client-1.20.0.jar"/>
-	<classpathentry kind="lib" path="module/MOD-INF/lib/google-oauth-client-servlet-1.20.0.jar"/>
-	<classpathentry kind="output" path="module/MOD-INF/classes"/>
-</classpath>
```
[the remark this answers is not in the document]
jackyq2015: This extension was including in the openrefine. No need to keep it and maintain it

main/src/com/google/refine/model/Column.java
jackyq2015: That's part of the table scheme java library

.gitattributes
```diff
@@ -1,22 +0,0 @@
...
-*.doc	 diff=astextplain
-*.DOC	 diff=astextplain
-*.docx diff=astextplain
-*.DOCX diff=astextplain
-*.dot  diff=astextplain
-*.DOT  diff=astextplain
-*.pdf  diff=astextplain
-*.PDF	 diff=astextplain
-*.rtf	 diff=astextplain
-*.RTF	 diff=astextplain
```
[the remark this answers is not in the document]
jackyq2015: added back

extensions/gdata/.classpath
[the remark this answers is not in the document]
wetneb: hmm, I don't see why? Yes we do want to maintain the gdata extension - actually one of our goals is to update this extension so that it works again with the new Google Sheets API. In any case this change is unrelated to data packages, so let's keep this file.

extensions/sample/.classpath
wetneb: I think the sample extension is definitely useful - I don't see why it would be a good thing to delete its `.classpath`. Again this is unrelated to data packages so let's not delete this file in this PR.

main/src/com/google/refine/commands/project/SetMetadataCommand.java
wetneb: Is this command currently called anywhere? If so, it should be renamed to something like `ValidateMetadataCommand` otherwise it will cause confusion. If not, it should be deleted.

main/src/com/google/refine/commands/GetAllPreferencesCommand.java
wetneb: well you did change the code too - `project.getMetadata().getPreferenceStore()` is not called anymore, so it's not just about adding some comments.

main/src/com/google/refine/model/medadata/validator/ValidatorConfig.java
```diff
@@ -0,0 +1,6 @@
+package com.google.refine.model.medadata.validator;
+
+
+public class ValidatorConfig {
+
+}
```
wetneb: Is this class needed? It does not seem to do much…

main/src/validator-resource-bundle.properties
```diff
@@ -0,0 +1,13 @@
...
+required-constraint=Column ${column_number} is a required field, but row ${row_number} has no value
+pattern-constraint=The value ${value} in row ${row_number} and column ${column_number} does not conform to the pattern constraint of ${constraint}
+unique-constraint=Rows ${row_numbers} has unique constraint violation in column ${column_number}
+enumerable-constraint=The value ${value} in row ${row_number} and column ${column_number} does not conform to the given enumeration: ${constraint}
+minimum-constraint=The value ${value} in row ${row_number} and column ${column_number} does not conform to the minimum constraint of ${constraint}
+maximum-constraint=The value ${value} in row ${row_number} and column ${column_number} does not conform to the maximum constraint of ${constraint}
+minimum-length-constraint=The value ${value} in row ${row_number} and column ${column_number} does not conform to the minimum length constraint of ${constraint}
+maximum-length-constraint=The value ${value} in row ${row_number} and column ${column_number} does not conform to the maximum length constraint of ${constraint}
```
wetneb: How can we localize this file?

main/src/com/google/refine/model/medadata/validator/checks/Validator.java
```diff
@@ -0,0 +1,36 @@
+package com.google.refine.model.medadata.validator.checks;
```
wetneb: These validators look great to me, but I cannot figure out how to access them as a user. Is there any way to see them in the UI?

server/.classpath
```diff
@@ -1,13 +0,0 @@
...
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.6"/>
-	<classpathentry kind="lib" path="lib/servlet-api-2.5.jar" sourcepath="lib-src/servlet-api-2.5-sources.jar"/>
-	<classpathentry kind="lib" path="lib/jetty-6.1.22.jar" sourcepath="lib-src/jetty-6.1.22-sources.jar"/>
-	<classpathentry kind="lib" path="lib/jetty-util-6.1.22.jar" sourcepath="lib-src/jetty-util-6.1.22-sources.jar"/>
-	<classpathentry kind="lib" path="lib/jdatapath-alpha2.jar" sourcepath="lib-src/jdatapath-alpha2-sources.jar"/>
-	<classpathentry kind="lib" path="lib/slf4j-api-1.5.6.jar" />
-	<classpathentry kind="lib" path="lib/log4j-1.2.15.jar" />
-	<classpathentry kind="lib" path="lib/slf4j-log4j12-1.5.6.jar" />
-	<classpathentry kind="output" path="classes"/>
-</classpath>
```
wetneb: I think this file should be kept too

server/.project
```diff
@@ -1,17 +0,0 @@
...
-		<buildCommand>
-			<name>org.eclipse.jdt.core.javabuilder</name>
-			<arguments>
-			</arguments>
-		</buildCommand>
-	</buildSpec>
-	<natures>
-		<nature>org.eclipse.jdt.core.javanature</nature>
-	</natures>
-</projectDescription>
```
wetneb: same here

main/src/TableSchemaValidator.json
wetneb: thanks. Any idea how we could localize this? Could this be done UI-side?

jackyq2015: No. this is server-side resource. we don't have the server-side i18n support yet. Can be added as a infrastructure and then do the localization.

main/src/com/google/refine/commands/GetAllPreferencesCommand.java
jackyq2015: it's remove and added backed as above

main/src/com/google/refine/commands/project/SetMetadataCommand.java
jackyq2015: As I mentioned above, it can be used for future metadata format

main/src/com/google/refine/model/medadata/PackageExtension.java
```diff
@@ -0,0 +1,88 @@
...
+    
+    /**
+     * To build a Package object from a template file contains empty metadata
+     *  
+     * @param templateFile
+     */
+    public static Package buildPackageFromTemplate() {
+        try {
+            ClassLoader classLoader = PackageExtension.class.getClassLoader();
+            File file = new File(classLoader.getResource(DATAPAAKCAGE_TEMPLATE_FILE).getFile());
```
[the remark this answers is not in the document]
jackyq2015: FIXED

main/src/validator-resource-bundle.properties
jackyq2015: Once we have i18n support on the server side

main/src/TableSchemaValidator.json
wetneb: or it should also be possible to use this file directly from the UI… but okay, let's keep this for now.

main/src/com/google/refine/model/medadata/validator/ValidatorConfig.java
jackyq2015: For now, it's a placeholder for future use

main/src/com/google/refine/model/medadata/validator/checks/Validator.java
wetneb: okay you wrote in the PR that it's not implemented yet. fine.

jackyq2015: There is UT ValidateOperationTests.java to demonstrate the basic use of it. but we don't have the UI to trigger  this. It's non-trivial task. we can add it in future on top of current UI infra or defer until we separate the front end and have a modern UI framework.

main/src/com/google/refine/commands/project/SetMetadataCommand.java
wetneb: well let's remove it for now, because it is not needed and redundant with the command to set the project metadata. I don't see how it could be useful in the future given the new way we store metadata. We should rather have commands to change column metadata (types, constraints) and this should be exposed in the menu for each column.

server/.classpath
jackyq2015: Current OpenRefine is a single project and single classpath. It is not worth keep it if we don't use it and have to maintain it and confuse programmer. Once we move to other build system such as gradle or maven, those setting will be all gone. Since we don't use it, deleting them can pave the road while we do the gradle/maven migration.

server/.project
jackyq2015: same above

extensions/gdata/.classpath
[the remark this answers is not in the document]
jackyq2015: The gdata src already covered by OpenRefine project

main/src/com/google/refine/commands/GetAllPreferencesCommand.java
wetneb: no @jackyq2015 it is not "added backed as above", there is a difference in the code and I would like you to explain it to me. I accept to turn a blind eye on some dirty corners of this PR but it would be nice if you could cooperate a bit more… I am trying to help you improve your PR.

main/src/com/google/refine/model/medadata/validator/ValidatorConfig.java
wetneb: it should not be too hard to add it back once it is needed, so let's remove it

.settings/org.eclipse.wst.jsdt.core.prefs
```diff
@@ -318,4 +319,5 @@ org.eclipse.wst.jsdt.core.formatter.tabulation.char=space
 org.eclipse.wst.jsdt.core.formatter.tabulation.size=4
 org.eclipse.wst.jsdt.core.formatter.use_tabs_only_for_leading_indentations=false
 org.eclipse.wst.jsdt.core.formatter.wrap_before_binary_operator=true
-semanticValidation=enabled
+semanticValidation=disabled
+strictOnKeywordUsage=disabled
```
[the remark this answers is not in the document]
jackyq2015: The way Butterfly handle the javascript can not fully compliant with the ECMA. Disable the semantic validation to stop the annoying message

.classpath
```diff
@@ -96,5 +102,20 @@
...
+	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/commons-csv-1.5.jar"/>
+	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/junit-4.12.jar"/>
+	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/bsh-2.0b4.jar"/>
+	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/byte-buddy-1.6.14.jar"/>
+	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/byte-buddy-agent-1.6.14.jar"/>
+	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/cglib-nodep-2.2.2.jar"/>
+	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/hamcrest-core-1.3.jar"/>
+	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/javassist-3.21.0-GA.jar"/>
+	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/jcommander-1.48.jar"/>
+	<classpathentry kind="lib" path="main/webapp/WEB-INF/lib/mockito-core-2.8.9.jar"/>
```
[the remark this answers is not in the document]
jackyq2015: It is required by table schema

.settings/.jsdtscope
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry excluding="build/**|main/webapp/modules/core/MOD-INF/controller.js|main/webapp/modules/core/externals/|test-output/" kind="src" path=""/>
+	<classpathentry kind="src" path=""/>
```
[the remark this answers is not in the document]
jackyq2015: The original intention is to suppress the warning messages from the js validation. But it can not exclude all the JS files it meant to.

.settings/org.eclipse.wst.validation.prefs
```diff
@@ -0,0 +1,8 @@
+DELEGATES_PREFERENCE=delegateValidatorList
+USER_BUILD_PREFERENCE=enabledBuildValidatorList
+USER_MANUAL_PREFERENCE=enabledManualValidatorList
+USER_PREFERENCE=overrideGlobalPreferencestruedisableAllValidationtrueversion1.2.700.v201508251749
+eclipse.preferences.version=1
+override=true
+suspend=true
+vf.version=3
```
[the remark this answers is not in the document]
jackyq2015: removed.

.settings/org.eclipse.wst.jsdt.core.prefs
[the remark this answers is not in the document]
wetneb: okay, thanks for the explanation!

extensions/sample/.classpath
jackyq2015: Again, this already covered by OpenRefine project. No need redundancy to setup the project

.settings/.jsdtscope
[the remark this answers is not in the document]
wetneb: okay, great

server/.classpath
wetneb: okay, fine for this time

main/src/com/google/refine/commands/GetAllPreferencesCommand.java
jackyq2015: Let me paste the diff and explain:
-                ProjectManager.singleton.getPreferenceStore();      // 1 line was removed 
+        
+        // project level has no specific preference                       // 1 comment added
+        PreferenceStore ps = ProjectManager.singleton.getPreferenceStore();  // 1 line was added back

main/src/com/google/refine/commands/project/SetMetadataCommand.java
jackyq2015: remvoed

main/src/com/google/refine/commands/GetAllPreferencesCommand.java
wetneb: Well then we are not seeing the same diffs!

This is what I see:

Three lines removed:
```
-        PreferenceStore ps = project != null ? 
-                project.getMetadata().getPreferenceStore() : 
-                ProjectManager.singleton.getPreferenceStore();
```

Three lines added:
```
+        
+        // project level has no specific preference
+        PreferenceStore ps = ProjectManager.singleton.getPreferenceStore();
```

So do we agree that there is a change in the code? You removed a condition. It does not seem harmful to me, but that is what I wanted to discuss… Makes sense?

wetneb: It's actually a change worth being discussed, because it means that the project passed as parameter to the command is now completely ignored - so that means the preceding line can be removed too:
`Project project = request.getParameter("project") != null ? getProject(request) : null;`

But that is suspicious - why was there any need to pass this argument before? Why would the preference store be different depending on the project? Do we rely on that anywhere? If so, there is a risk that this PR introduces a bug. I am not saying it does, I am just saying this needs to be investigated.

jackyq2015: The diff given by github didn't provided 3 "-" for some reason. it only shows 1 "-".

Reverted the change

wetneb: Great, thanks a lot! (But that's worrying - it means we can't rely on GitHub to review PRs?! That sounds like a serious bug to me!)

jackyq2015:
yeah. that's very terrible bug of GitHub. Not sure if that is there any
good alternative to it? Atom might be a one. but not sure if it is behave
the same. I know @thad use it, @thad  can you please confirm

thadguidry:
@jackyq2015 bugs in displaying added and removed lines in code ?  dunno.  But shows correctly in IntelliJ and gitkraken

jackyq2015:
seems have to stick to the github.com for PR review which other git clients don't provide.

wetneb:
@jackyq2015 okay so I think we have covered the whole PR now - unless people have other issues in mind I will merge in a few days.

jackyq2015:
@wetneb Thanks a lot! I just fixed last 2 conflicts of the master.

PR #1399, opened 2017-12-26
Update translation-jp.json for 2.8

isaomatsunami:
Some items are added for 2.8. Some are changed not to break the original layout.
The following 2 pages are hard-coded in English?
1) import from Google Data
2) parse type for PC-AXIS text files


wetneb:
@isaomatsunami thank you very much! I cannot read Japanese but it looks good to me. Did you know you can also use Weblate for that? https://hosted.weblate.org/projects/openrefine/translations/

PR #1407, opened 2017-12-28
Revised translation-jp.json to be consistent with Google Docs Extension translation

isaomatsunami:
I am sorry to bother you.
At last I understand there are 2 files to translate. I should have read your advices on Github.


isaomatsunami:
please ignore the weblate version, which I made it to 100% just before I find teh second file.

PR #1412, opened 2018-01-03
Jython tests

wetneb:
This updates the ant build scripts to execute extension tests directly from the main directory, with `./refine extensions_test`. I have added that to the Travis script.

I have only enabled the Jython test for now, but the setup can easily be reused to test other extensions.


build.xml
```diff
@@ -206,6 +206,10 @@
         </jacoco:report>
     </target>
 
+    <target name="extensions_test" depends="build">
```
thadguidry: Add a comment about the What.  For instance, something like what you said in the pull request "This will execute extension tests directly from the main directory, with ./refine extensions_test ...blah"

extensions/jython/build.xml
```diff
@@ -48,7 +58,23 @@
             <include name="**/*.jar" />
         </fileset>
         <pathelement path="${refine.classes.dir}"/>
+        <pathelement path="${classes.dir}" />
     </path>
+
+    <path id="extension.tests.class.path">
+        <path refid="class.path"/>
+        <pathelement location="${tests.classes.dir}"/>
+        <pathelement location="${refine.tests.lib.dir}/testng-6.8.jar"/>
```
thadguidry: why not the latest ?  6.10 ?

extensions/jython/build.xml
```diff
@@ -60,7 +86,46 @@
...
+        <mkdir dir="${tests.classes.dir}" />
+        <javac source="${java_version}" target="${java_version}" encoding="utf-8" srcdir="${tests.src.dir}" destdir="${tests.classes.dir}" debug="true" includeAntRuntime="no">
+            <classpath refid="extension.tests.class.path" />
+        </javac>
+        <copy file="${tests.src.dir}/tests.log4j.properties" tofile="${tests.classes.dir}/tests.log4j.properties"/>
+    </target>
+ 
+    <mkdir dir="${tests.build.dir}" />
+    <target name="test" depends="build_tests">
+	<jacoco:coverage destfile="${tests.report.dir}/jacoco.exec">
```
thadguidry: I'm glad we used Jacoco.  Cobertura SUCKS ASS.

extensions/jython/tests/conf/tests.xml
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
+ 
+<suite name="Jython extension">
+  <test name="tests">
+    <groups>
+      <run>
+        <exclude name="broken" />
```
thadguidry: What does this do exactly ?

extensions/jython/build.xml
wetneb: That's what we use in OpenRefine so far - we can update it, but it's a different topic

extensions/jython/tests/conf/tests.xml
wetneb: Yep, that just lets you skip tests that are known to fail for some reason.

extensions/jython/build.xml
thadguidry: OK

extensions/jython/tests/conf/tests.xml
thadguidry: how does it know they have failed before ? it keeps some history or xml in a folder , or ?

build.xml
wetneb: sure, done (I've left out the bit about ./refine because people can also use ant directly)

extensions/jython/tests/conf/tests.xml
wetneb: it's something you add manually, when you are fed up by one particular test that fails because of something outside your control

thadguidry: ah k

thadguidry:
Merge this whenever your ready Antonin.  Looks like your still working out the Appveyor failures, that's fine.

PR #1413, opened 2018-01-03
Fix appveyor

wetneb:
This tweaks the URL caching test so that it passes on Appveyor, where random.org returns 503 errors.

### DIFF
--- a/main/tests/server/src/com/google/refine/tests/model/UrlFetchingTests.java
+++ b/main/tests/server/src/com/google/refine/tests/model/UrlFetchingTests.java
@@ -169,13 +169,13 @@ public class UrlFetchingTests extends RefineTest {
 
         // Inspect rows
         String ref_val = (String)project.rows.get(0).getCellValue(1).toString();
-	if (ref_val.contains("HTTP error 403") || ref_val.contains("HTTP error 503"))
+	if (ref_val.contains("HTTP error 403"))
             return;
         Assert.assertTrue(ref_val != "apple"); // just to make sure I picked the right column
         for (int i = 1; i < 4; i++) {
             System.out.println("value:" + project.rows.get(i).getCellValue(1));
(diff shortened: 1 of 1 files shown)


main/tests/server/src/com/google/refine/tests/model/UrlFetchingTests.java
```diff
@@ -169,13 +169,13 @@ public void testUrlCaching() throws Exception {
 
         // Inspect rows
         String ref_val = (String)project.rows.get(0).getCellValue(1).toString();
-	if (ref_val.startsWith("HTTP error 403"))
+	if (ref_val.contains("HTTP error 403"))
             return;
```
jackyq2015: to skip the 500 error etc, I think:
if (ref_val.startsWith("HTTP error"))
would be better.

wetneb: sounds good (it does not actually work anyway but oh well)

thadguidry:
@wetneb @jackyq2015 Why can't we just drop random.org and instead just mock this cache test instead with Mockito ?  I see tons of examples on the web such as this one https://reliablesoftwareblog.wordpress.com/2016/04/19/testing-a-cacheable-method-with-mockito/

wetneb:
@thadguidry it seems that your link is about something else - but I agree in theory we could try to mock that. It looks quite painful to me - I suggest we just merge this so that appveyor goes back to green, and we can consider separately whether we should mock HTTP requests in the tests, in general (which could speed them up).

thadguidry:
@wetneb my point is that this specific test could use behavior verification rather than state verification. And mocking is a great way to get there.  This is from 2007, but explains the concepts that we have learned and often  https://martinfowler.com/articles/mocksArentStubs.html

Issue #1377, opened 2017-12-12
Jython: not all variables are available


wetneb:
Not sure if this is intended, but I find it annoying that the fields of some variables available in GREL do not seem to be available in Python. For instance, the famous `cell.recon.match.id` GREL expression works fine, but `return cell.recon.match.id` does not work in Jython. Wouldn't it be better to make this more uniform?

ettorerizza:
@wetneb It can be accessed, but as a value in a dictionary ( return cell['recon']['match']['id'] ). the dot notation does not work in Python.

![screenshot-127 0 0 1-3333-2017-12-12-09-30-44-368](https://user-images.githubusercontent.com/6286410/33874586-974133a2-df1f-11e7-9758-c0bb5735f432.png)

wetneb:
@ettorerizza ah, thanks a lot! Do you think it make sense to also allow the dot notation too?

ettorerizza:
Of course, it would be much more convenient ! To get the list of values in a record, the formula is `row['record']['cell']['column name']['value']`

This is very boring to write without autocompletion ...

PR #1414, opened 2018-01-03
Expose fields as attributes in Jython

wetneb:
This implements #1377 : fields in Jython are available both as items and as attributes, just like GREL.
In other words, both `return cell.recon.match.id` and `return cell['recon']['match']['id']` are possible in Jython.


wetneb:
@ettorerizza sure, it was just for the user perspective! :)

ettorerizza:
Everything looks fine for me!

wetneb:
@jackyq2015 if you are happy with this PR, can you approve it so that we can merge?

PR #1419, opened 2018-01-05
Update from Weblate.

weblate:


wetneb:
Aaah it's so good to see all these green checks :)

PR #1426, opened 2018-01-07
Nobel prize winners CSV data typos

olleolleolle:
This PR fixes spelling details in the CSV data.


ostephens:
Not that it matters I don't think, but the use of the spelling "heterogenous" rather than "heterogenous" is deliberate because it directly quotes the Nobel citation which uses the first spelling rather than the second (both are valid spellings of the word)

PR #1427, opened 2018-01-10
Enable openjdk8 and oraclejdk9 in Travis

wetneb:
New version of #1300.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: java
 jdk:
   - oraclejdk8
+  - oraclejdk9
+  - openjdk8
 
 env:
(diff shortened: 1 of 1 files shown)


thadguidry:
@wetneb Yeah, this will help us test Tony's DB work.  Also, Appveyor has the services for various DB's that can be configured in it's yml file so that we can test against Windows as well.

jackyq2015:
Did you test jdk9?  I would hold the jdk9 until we have it tested. It may bring more trouble than benefit even it past the UT

wetneb:
The point of this PR is to test jdk9.

PR #1394, opened 2017-12-23
OpenRefine Database Import Extension

tcbuzor:

PR #1430, opened 2018-01-13
Revert "OpenRefine Database Import Extension"

jackyq2015:
Reverts OpenRefine/OpenRefine#1394

Have to revert this one.

The src and test need to be added into project and classpath. also there is some jar issues need to addressing.


tcbuzor:
Please  provide more description to the problem. src and test is already in the .classpath file.

jackyq2015:
Your project has no issue standalone(ie, when pull it and install it as an
extension to current master). The problem is that the library setting is
not in the current master .classpath UNDER THE ROOT. The current way of
setup in order to include the extension is to add the src and test folder
directly under the build path instead of adding the project.

The problem you need to fix is on the OpenRefine project setting, NOT your
own project setting. If you open the project setting of the OpenRefine and
check how gdata is setup. you will understand what I mean.

tcbuzor:
Thanks Jacky.
I made the changes and generated another pull request.

-Tony

PR #1431, opened 2018-01-14
OpenRefine db extension with fix for #1430

tcbuzor:


tcbuzor:
That is something we can explore later. One thing to note is that the default login information is different between travis and appveyor so test setup will have to determine the target environment.

PR #1434, opened 2018-01-16
Set http req headers

ostephens:
Support setting http headers when using "Add column by fetching URL"

### DIFF
--- a/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
+++ b/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
@@ -350,7 +350,6 @@ public class ColumnAdditionByFetchingURLsOperation extends EngineDependentOperat
                         urlConnection.setRequestProperty(info.header, httpHeaders.get(headerLabel));
                     }
                 }
-//                    urlConnection.setRequestProperty(_headerKey, _headerValue);
 
                 try {
                     InputStream is = urlConnection.getInputStream();
--- a/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
+++ b/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
@@ -42,12 +42,15 @@ import java.net.URL;
 import java.net.URLConnection;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.ExecutionException;
 
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.json.JSONArray;
 import org.json.JSONWriter;
 
 import com.google.refine.browsing.Engine;
@@ -66,6 +69,8 @@ import com.google.refine.model.Project;
 import com.google.refine.model.Row;
 import com.google.refine.model.changes.CellAtRow;
 import com.google.refine.model.changes.ColumnAdditionChange;
+import com.google.refine.commands.HttpHeadersSupport;
+import com.google.refine.commands.HttpHeadersSupport.HttpHeaderInfo;
 import com.google.refine.operations.EngineDependentOperation;
 import com.google.refine.operations.OnError;
 import com.google.refine.operations.OperationRegistry;
@@ -77,6 +82,7 @@ import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.LoadingCache;
 import com.google.common.cache.CacheLoader;
 
+
 public class ColumnAdditionByFetchingURLsOperation extends EngineDependentOperation {
     final protected String     _baseColumnName;
     final protected String     _urlExpression;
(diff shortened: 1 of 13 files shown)


main/tests/server/src/com/google/refine/tests/model/UrlFetchingTests.java
```diff
@@ -221,4 +227,59 @@ public void testInvalidUrl() throws Exception {
...
+           "Host": "headers.jsontest.com",
+           "User-Agent": "OpenRefine",
+           "Accept": "*"
+        }
+        */
+
+        project.rows.add(row0);
+
+        String userAgentValue =  "OpenRefine";
+        String authorizationValue = "Basic";
```
thadguidry: Only Basic Auth for now, right ?

thadguidry: hmm, wondering about Proxy-Authorization as well ? https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication perhaps a later pull if folks want it ?

ostephens: Basic auth is prob going to be the most common, but there are other methods that work through the Authorization request header (e.g. http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html)

Would be easy to add the Proxy-Authorization request header as well

main/webapp/modules/core/scripts/views/data-table/add-column-by-fetching-urls-dialog.html
```diff
@@ -21,6 +21,11 @@
 	     <input type="checkbox" name="dialog-cache-responses" id="$add-column-cache-responses" checked="checked" />
 		<label for="$add-column-cache-responses" bind="or_views_cacheResponses"></label></td>
         </tr>
+        <tr><td colspan="4"><h3 class="inline-header"><span bind="or_views_httpHeaders"></span></h3>
```
wetneb: This &lt;h3&gt; renders as bold text for me. I think this should be normal text just like the other options, so that "Formulate the URLs to fetch:" remains the single bold label. Otherwise I feel like I am prompted to set custom headers before defining the URLs.

wetneb:
(Travis seems to have problems at the moment - we should wait for it to go green before merging this.)

ostephens:
I'm having issues with some tests locally at the moment - I don't think it is anything in this PR... is Master testing OK currently?

main/webapp/modules/core/scripts/views/data-table/add-column-by-fetching-urls-dialog.html
ostephens: Agreed - will take this out

ostephens:
@ettorerizza would be good to get any feedback you have on the functionality if you have any chance to look at it

wetneb:
@ostephens your tests sometimes fail here. It must be something with the timeout being too low. I had the same issue with the URL caching test before. We could look into mocking the HTTP requests in the URL fetching tests to make this more reliable.

ostephens:
Ah yes - I saw that discussion previously. I think there was an issue with access to the URI being used in the testing being blocked as well?

I think the 'solution' was to allow Travis to skip these tests at the moment until we can mock the requests?

wetneb:
The test works fine on Travis but we had issues on Appveyor.

ostephens:
Ah - thanks. Looks like its built successfully now :)

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
```diff
@@ -140,6 +150,7 @@ public void write(JSONWriter writer, Properties options)
         writer.key("onError"); writer.value(TextTransformOperation.onErrorToString(_onError));
         writer.key("delay"); writer.value(_delay);
         writer.key("cacheResponses"); writer.value(_cacheResponses);
```
jackyq2015: Not sure if this should under the model level? maybe should a preference?

ostephens: Not sure I understand (sorry). Do you mean the headers should be set at a preference level? I did consider this, but each request might need different headers, and if you haven't recorded the headers used when making a specific request you can't undo/redo in a predictable way

ostephens:
Unless there are any further comments, I'm going to go ahead and merge this ...

ostephens:
Thanks @wetneb !

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
thadguidry: @ostephens but we could have PER PROJECT Preferences as well as GLOBAL Preferences.
@jackyq2015 Didn't your model/metadata re-work actually afford that now for us ? or would there be more work needed ?

ostephens: At the moment these are per operation. I'm not clear that having these at a project or global level is that useful? Even within a single project different operations might require different headers? What's the demand for having these per project or global?

That said I don't have any particular objection, but it would need a new issue

PR #1442, opened 2018-01-25
Simplify project creation in unit tests

wetneb:
When writing unit tests for various parts of the code, it is often required to create a new project, which is a fairly heavy operation. This PR provides a helper in RefineTest that makes it easier to create a test project from a CSV in just one call. More tests could be migrated to use this helper - I just request your feedback before doing it.

I think a lot more sharing between unit tests should be done. For instance, some of this refactoring could also be done for tests which fill the rows of the project programmatically (such as https://github.com/OpenRefine/OpenRefine/blob/refactor-test-initialization/main/tests/server/src/com/google/refine/tests/expr/functions/FunctionTests.java).

In general I think we should really try to make it as easy as possible to write tests - one should not have to copy and paste dozen of boilerplate lines from another test class as a preamble to any new test.


jackyq2015:
Looks good. But should we refactor other UT to use the helper method

wetneb:
@jackyq2015 thanks! I'm going to fix a few other things soon, let's wait a bit before merging.

wetneb:
Bonus track: no more
`[       FileProjectManager] Failed to load workspace from any attempted alternatives.`
in the test logs!

Now the tests are run in the same test workspace, which makes it a bit easier to inspect the test projects if need be.

wetneb:
@jackyq2015 this time it's ready

PR #1444, opened 2018-01-27
Various fixes on the database tests

wetneb:
This makes a few changes to the way database tests are run:
- First, they are disabled by default (because we don't expect our developers to run the three database servers locally). Anybody can enable tests for a specific database vendor by changing `extensions/database/test/conf/int_tests.xml` (where they can also configure the connection to the database).
- All three databases are enabled on Travis. MySQL and Postgres are enabled on appveyor.
- I have removed all the `try` blocks from the database tests. Such blocks defeat the purpose of tests: if an exception is raised, we do want the test to fail, otherwise the failure will not be reported.

@tcbuzor it looks like there are two configuration files for the database tests (int_tests.xml and unit_tests.xml). Only the first one seems to be actually used - I propose we rename it to just "tests.xml" and remove the other.

By the way I want to thank Tony again for the amazing work done on these tests (and the rest of the extension) - the database extension is by far the part of the code that is the most thoroughly tested! It's very impressive.


extensions/database/test/conf/int_tests.xml
```diff
@@ -1,41 +1,56 @@
...
-		    <class name="com.google.refine.extension.database.pgsql.PgSQLConnectionManagerTest"/>
-			<class name="com.google.refine.extension.database.pgsql.PgSQLDatabaseServiceTest" />
-		</classes>
-	</test>
-</suite>
\ No newline at end of file
+    <test name="DatabaseExtensionUnitTest">
+        <groups>
+            <run>
+                <exclude name="requiresMySQL"/>
```
jackyq2015: why exclude all db-specific testing? only do it on travis?

wetneb: I am not sure what you mean - we already have three files for local, travis and appveyor (the fourth one should be deleted - I am just waiting for @tcbuzor's confirmation). It's not clear to me how to avoid the copy but feel free to commit on this branch with the setup you are thinking about. Note that currently no copy is needed when you test locally - the default configuration file already excludes all tests which require database access. Copying is only required on Travis and Appveyor (but that does not seem to be a big issue to me).

tcbuzor:
Hi Antonin,
Your changes look good. I was working on the Appveyor test but no longer required as you have it working nicely. 
Its okay to rename  int_test.xml and delete unit_tests.xml. 
Good Job!

thadguidry:
@tcbuzor @wetneb @jackyq2015 I guess we are only testing MySQL on Appveyor from the looks of it...also on Appveyor, it seems there's a Postgres setup error here, why do we use the same travis-SQL file for Appveyor, wouldn't it be better to have one just for Appveyor also ?  https://ci.appveyor.com/project/openrefine/openrefine/build/1.0.300#L29

wetneb:
Why are you saying that we are only testing MySQL? I can see both MySQL and PgSQL:

```
   [testng] PASSED: testGetMySQLDBService
   [testng] PASSED: testGetPgSQLDBService
```

I don't know about the PgSQL error for Appveyor - feel free to investigate, I am not hugely bothered as it seems that the tests pass fine with it.

Intuitively I would expect that the SQL files are identical for Windows and Linux (they should only depend on the SQL vendor) but again I know nothing about Windows (I actually had to lookup the Windows-equivalent of `cp` to write this PR…)

So feel free to look into that if you think that is important - you know much more about SQL on Windows than I do. As long as Appveyor is green, I am happy :)

PR #1452, opened 2018-02-03
Add support for quote character

### LINKED ISSUE #1086: single quotes as quote characters

wetneb:
This finishes the work of #1087 and closes #1086.


jackyq2015:
There is compile errors

thadguidry:
@wetneb I can't even checkout his branch properly. WTF Github !?!?

THAD@DESKTOP-TJUMQT0 MINGW64 ~/OpenRefine (master)
$ git checkout -b stundzig-develop/1086-quotecharacter origin/stundzig-develop/1086-quotecharacter
fatal: 'origin/stundzig-develop/1086-quotecharacter' is not a commit and a branch 'stundzig-develop/1086-quotecharacter' cannot be created from it

It looks like the stupid Github web editor for merge conflict resolve grabbed the wrong line I told it for prepareOptions.  I of course WANTED to handle this in IDE, but the Git branch error above is preventing me !

I attempted to resolve the merge conflict, its resolved now(reversed via master into his branch), but we still have to fix the missing quoteCharacter field on some in TsvCsvImporterTests.java

Can you checkout his branch or do you get a similar error as me ?

wetneb:
@thadguidry it's just that you removed the method when you merged. I added it back in the commit above.
For your git issues, I think you can just do
```
git pull
git checkout stundzig-develop/1086-quotecharacter
```
and that should work, I think.

thadguidry:
@wetneb found my issue... I forgot I had forced origin to Jacky's branch during the holidays for easier review !
remote.origin.url=git://github.com/jackyq2015/OpenRefine

main/src/com/google/refine/importers/SeparatorBasedImporter.java
```diff
@@ -98,9 +101,15 @@ public void parseOneFile(
         boolean processQuotes = JSONUtilities.getBoolean(options, "processQuotes", true);
         boolean strictQuotes = JSONUtilities.getBoolean(options, "strictQuotes", false);
         
+        Character quote = CSVParser.DEFAULT_QUOTE_CHARACTER;
+        String quoteCharacter = JSONUtilities.getString(options, "quoteCharacter", null);
+        if (!StringUtils.isBlank(quoteCharacter)) {
+            quote = quoteCharacter.charAt(0);
```
jackyq2015: should this be handled at front end to only allow 1 char(or possible restriction on what char cannot be used)? If use input [blank]'[blank] then it will cause problem.

thadguidry:
@wetneb @jackyq2015 Here's a clipboard text that I used for testing... looks like it works...where line 4 in preview has the column 3, or not depending on the "Use character" checkbox.

`"this string has double quotes around it","this string also has double quotes around it<NEWLINE>"
this string has no quotes, "but this string does!", 'and this string has single quotes around it<NEWLINE>'
'"and this string is wrapped first with double quotes and then single quotes <NEWLINE>"'
"this string has double quotes around it and single quotes around this: 'TERM'", "and this string has double quotes around it and a comma is here, <-- COMMA<NEWLINE>"`

We should probably also allow the quoteCharacter to be changed by the user ?

main/src/com/google/refine/importers/SeparatorBasedImporter.java
wetneb: The front end already enforces that the maximum length of this string is 1, but sure, we can make that more robust for people who do not use the web frontend.

wetneb:
> We should probably also allow the quoteCharacter to be changed by the user ?

Isn't that already the case? (Note: you might be confused by the fact that OR does not pick the CSV/TSV importer by default for some inputs. Make sure you select the CSV/TSV importer, and you should see the input to change the character.)

thadguidry:
Works great !  Thanks to @wetneb and @stundzig !

Issue #1475, opened 2018-02-12
Fix OpenRefine functions that break upon returning array


joanneong:
When reporting a bug please provide the following information to help reproduce the bug:

**Version of OpenRefine used:** OpenRefine 2.8
**Operating Systems and version**: MacOS High Sierra V10.13.3
**Browser + version used**: Google Chrome Version 64.0.3282.140 (Official Build) (64-bit)
**Steps followed to create the issue**
1. Using any project, go to a column and try to transform the column
2. Use any function that can possibly return an array, e.g. `slice` and try to get an array
3. See that the transformation is not applied (original values are kept), or that `String[] is not storable` (if user opts to see error message)

![issue with array](https://user-images.githubusercontent.com/24493463/36090797-a44ca940-101c-11e8-8ee9-bd20aeecff3d.gif)

This is terrible because as a user I fully expect the array to be displayed. This is particularly because the documentation does give such [an example of using the `slice` function](https://github.com/OpenRefine/OpenRefine/wiki/GREL-Array-Functions), which makes even less sense.

Although I see that arrays not being supported is a known [issue](https://github.com/OpenRefine/OpenRefine/issues/1088), I am not sure if the functions are supposed to work like this... :/

ettorerizza:
Hi @joanneong .  The documentation is right to say that `slice()` returns an array. The problem is that the arrays do not display in the columns. Open Refine columns only display strings, numbers, booleans, or dates. To display all the elements of an array in the column, you have to concatenate them with `join()`:

`value = [1,2,3,4]
value.join (',')`

I'm sure all of this is hidden somewhere in the doucmentation, but it might need to be mentioned more explicitly in the [Array functions](https://github.com/OpenRefine/OpenRefine/wiki/GREL-Array-Functions) section.

joanneong:
Thanks for the response @ettorerizza ! This is because I am working on a new `range` function (which should return an array) so I was looking more into this issue. Hmm seems like I have to return a string for now then, thanks!

ostephens:
I was sure we had an issue (or maybe a discussion on the Google group?) about doing an enhancement so that if there was an attempt to store an array in a cell, it would get serialised to a string (probably a JSON representation of the array) before it was stored - to prevent this odd behaviour of the preview looking fine but the outcome of the operation being unexpected.

Can anyone (@thadguidry @magdmartin ?) remember this discussion and point to where it happened so we can cross reference to here?

thadguidry:
[...]

Issue #1088, opened 2015-10-30
Give the user an Error when they try to populate a cell with an array


thadguidry:
"You can't populate a cell with an array, and when you try this, you just get a blank cell (it's a shame there isn't an error message instead which would tell you what had happened,)" - from mailing list by Owen Stephens

2 things are needed for this Enhancement Request:
1. Provide an error.
2. Also provide an alternative such as outputting to a string with join(), etc.

thadguidry:
#443

ostephens:
There is more extensive discussion of this concept in a thread on the Google Group https://groups.google.com/d/msg/openrefine/aDxquAI4MqY/P-KD7mghFAAJ

ostephens:
In that discussion it was proposed that rather than giving an error, then the array could be stored as a string which is a JSON array, that could then be parsed in future transforms with parseJson().

@thadguidry did a proof of concept implementation at the time saying

> For this, I made changes only to ExpressionUtils.java
> 
> ```
> 
> 
>     static public boolean isStorable(Object v) {
>         return v == null ||
>             v instanceof Number ||
>             v instanceof String ||
>             v instanceof Boolean ||
>             v instanceof Date ||
>             v instanceof Calendar ||
>             v instanceof EvalError ||
>             v instanceof Arrays ||      // NEW
>             v instanceof Collection ||      // NEW
>             v instanceof List;      // NEW
>     }
> 
>     static public Serializable wrapStorable(Object v) {
>         if (v instanceof JSONArray) {
>             return ((JSONArray) v).toString();
>         } else if (v instanceof JSONObject) {
>             return ((JSONObject) v).toString();
> 
> // NEW
>         } else if (v instanceof List) {
>             return ((List<?>) v).toString();
>         } else if (v.getClass().isArray()) {
>             return (Arrays.deepToString((Object[]) v));  // This is kinda cool and uses http://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#deepToString(java.lang.Object[])
>         } else if (v instanceof Collection<?>) {
>             return ((Collection<?>) v).toArray().toString();
> // NEW
> 
> 
>         } else {
>             return isStorable(v) ?
>                 (Serializable) v :
>                 new EvalError(v.getClass().getSimpleName() + " value not storable");
>         }
>     }
> ```

ostephens:
[...]

PR #1480, opened 2018-02-12
Add a new range() function to GREL

### LINKED ISSUE #1319: Add a new range() function to GREL
Use case taken from mailing list 
https://groups.google.com/d/topic/openrefine/SE2F7tn3lH0/discussion

Hello all,

I've been trying to figure this out, but need some help. I have a column full of ranges of numbers, e.g.:

row 1: 2-4
row 2: 7-12

I would like to be able to convert the cell so that it displays

row 1: 2,3,4
row 2: 7,8,9,10,11,12

My eventual goal being to split the cell with each number on a separate row in the record with Split multivalued cells. Is there a way to do this in OpenRefine?

Thank you!!
Olivia

joanneong:
Fixes #1319 

The original specification that @thadguidry suggested for the `range` function was:

> It will work like this...
>
>can take accept a String form:
>range("2-4") --> [2,3,4]
>
>or a String form in a value
>range(value) --> [2,3,4]
>
>or 2 numbers
>range(2, 4) --> [2,3,4]
>
>and doing join(",") will convert the Array to a String
>range(2,4).join(","). --> 2,3,4

However, because of issues such as [cells being unable to display arrays](https://github.com/OpenRefine/OpenRefine/issues/1088) and the [potential confusion it can cause](https://github.com/OpenRefine/OpenRefine/issues/1475), I have taken the liberty to display the results of `range` as a string with comma-separated values instead.

Hence, my implementation works like this:

**(a) String arguments**
1. `range("2-4") --> 2, 3, 4`
2. `range("4-2") --> 4, 3, 2` 
3. `range("2, 4") --> 2, 3, 4` 
4. `range("4, 2") --> 4, 3, 2`
5. `range(value) --> 2, 3, 4 //given that value is written as 2-4 (text) or 2, 4 (text) in the cell`

**(b) Integer arguments**
1. `range(2, 4) --> 2, 3, 4`
2. `range(4, 2) --> 4, 3, 2`

*`range(value)` does not apply to integer arguments since `2, 4` is stored as text in a cell.

**Additional notes**
1. Given valid arguments, either an increasing sequence or a decreasing sequence is returned depending on whether the `rangeStart` value is larger than the `rangeEnd` value.
2. Integer arguments cannot be negative; otherwise, an `EvalError` is returned.
3. Numeric arguments can only be integers, and decimal values will cause an `EvalError` to be returned.
4. It is okay to have whitespace(s) within the string (e.g. `range("2   -   4  ")` will work as well).
 
**Questions to discuss**
1. Should the implementation use integers? What about `long` or `BigInteger`? Should we worry about when users try to input very large values (e.g. 2980572305402385707520357208) to the `range` function?
2. Should the result returned be an array as @thadguidry has suggested? In this case, the result will directly be displayed in the cells after the transformation without having to use `join`, but is this consistent with other functions already provided by OpenRefine?
3. Should negative integers, or even decimal numbers be supported by this function?

Lastly, I have also added some basic tests for the new function (although more comprehensive testing can be done with real examples). I am willing to help to update the documentation on GitHub pages for this function after it is approved :)

Thanks for reading this lengthy pull request message, and I look forward to any incoming feedback :)

### DIFF
--- a/main/src/com/google/refine/expr/functions/strings/Range.java
+++ b/main/src/com/google/refine/expr/functions/strings/Range.java
@@ -30,7 +30,7 @@ public class Range implements Function {
     private static final int DEFAULT_START = 0;
     private static final int DEFAULT_STEP = 1;
 
-    private static final String[] EMPTY_STRING_ARRAY = new String[0];
+    private static final Integer[] EMPTY_ARRAY = new Integer[0];
 
     @Override
     public Object call(Properties bindings, Object[] args) {
@@ -280,15 +280,15 @@ public class Range implements Function {
      */
     private static Object createRange(int start, int stop, int step) {        
         if ((start > stop && step > 0) || (start < stop && step < 0) || step == 0) {
-            return EMPTY_STRING_ARRAY;
+            return EMPTY_ARRAY;
         }
 
         int rangeSize = (int) (Math.ceil(((double) Math.abs(start - stop))/ Math.abs(step)));
 
-        String[] generatedRange = new String[rangeSize];
+        Integer[] generatedRange = new Integer[rangeSize];
 
         for (int i = 0; i < rangeSize; i++) {
-           generatedRange[i] = Integer.toString(start + step * i);
+           generatedRange[i] = start + step * i;
         }
 
         return generatedRange;   
--- a/main/src/com/google/refine/expr/functions/strings/Range.java
+++ b/main/src/com/google/refine/expr/functions/strings/Range.java
@@ -1,5 +1,7 @@
 package com.google.refine.expr.functions.strings;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.Properties;
 
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


thadguidry:
@joanneong good work so far !  My idea was to handle the 80% of the use cases that folks would use the GREL range() function for.  The design should mimic what Jython/Python does and it's range() https://docs.python.org/3/library/stdtypes.html?highlight=range#range

1. Only Integers supported, I would say, just like Python's range()
2. Yes, return it as an array.  We have other issues regarding enhancing our displays (via coloring, syntax, etc) that will let folks know that their output is not of a String type that can be written to a cell.  So for range() just handle as we do with other functions that output an Array.
3. Yes, negative numbers should be supported, just as they are with Python's range()

One more argument that I think would be valuable would be a STEP, just as Python's range() has.
`range(start, stop[, step])`

Also testing for equality should mimic as Python does and treat the range objects as sequences. See notes on the range() class in Python's docs url above.

ettorerizza:
I also think that this function should produce an array, not a string. It would be on the one hand more coherent intellectually and on the other more practical in many cases. After all, a range function is most often used as iterable.

The optional step argument is a good idea. In this way, we could make the `forRange` function obsolete  (who uses it?) and use only `forEach` to iterate arrays:

`forEach (range (number from, number to, number step), variable v, expression e)`

joanneong:
Got it, thanks for the prompt responses @thadguidry and @ettorerizza ! I will update this pull request soon :P

ettorerizza:
Technical detail: must the stop be included in the array? The range function of Python does not include it:

`list(range(1,10)) --> [1,2,3,4,5,6,7,8,9]`

The current function `forRange` either:

`forRange(1,10,1,e, e) --> [1,2,3,4,5,6,7,8,9]`

Personally, I would support the idea of including it, as 90% of humans that would answer the question "give me the numbers from 1 to 10".

thadguidry:
@ettorerizza the Step is optional Ettore

thadguidry:
@ettorerizza Yes, the Stop needs to be output.

wetneb:
I would be in favour of not including the bound, just to be consistent with `forRange`, Python and a good deal of other programming languages.

`range(10)` should return the array of the first 10 integers.

But it is clearly not a hill I want to die on.

ettorerizza:
It's true that it'll introduce inconsistency, since all OR array functions use a zero-based indexing.

```
value = [1,2,3]
value.slice(1,2) -> [2] #(and not [1,2] as it would be the case with this range() function)
``` 
All this seems very clear to any programmer (except perhaps those who only use R), but I remember that it disturbed me a lot when I started learning OpenRefine...

jackyq2015:
+1 for returning a array instead of string. It will provide a easier way to loop through it also leave the decision to the end user how to decorate(split, padding etc) the elements

joanneong:
**UPDATE SUMMARY**
1. The `range` function now returns an array and requires users to use `.join(",")` if they want to see it in the cells.

2. The `range` function now has a `step` parameter (optional).

3. The `range` function can accept 1 to 3 string/integer arguments:

   (a) 1 argument: 0 becomes the default range start, and the given argument is the range end
      e.g. `range(5)` --> `["0", "1", "2", "3", "4", "5"]`
              `range("5")` --> `["0", "1", "2", "3", "4", "5"]`

   (b) 2 arguments: the first argument is the range start, and the second argument is the range end. If the range start is smaller than the range end, the sequence generated is an increasing sequence (increments by 1 each time). Otherwise, it is a decreasing sequence (decreases by 1 each time).
      e.g. `range(0, 5)` --> `["0", "1", "2", "3", "4", "5"]`
             `range(5, 0)` --> `["5", "4", "3", "2", "1", "0"]`

   (c) 3 arguments: the first argument is the range start, the second argument the range end, and the third argument the step. Steps are all positive numbers, and the sequence generated is a decreasing one only if the range start is larger than the range end. Step 0 will produce an empty array. Steps that are bigger than the range will produce only the range start.
      e.g. `range(1, 5, 2)` --> `["1", "3", "5"]`
             `range(5, 1, 2)` --> `["5", "3", "1"]`
             `range(1, 5, 0)` --> `[""]`
             `range(1, 5, 100)` --> `["1"]`

4. Any permutation of strings and integers is possible when using > 1 arguments for the `range` function (e.g. `range("1", 2) --> ["1", "2"]), except when the first argument is an integer and the last two arguments are strings (e.g. `range(1, "2, 3")` --> `EvalError`) (can be fixed if wanted).

5. Whitespaces are still ignored as long as the values (string/ integer) are valid.

6. Both given range start and range end are included in the generated sequence.

Let me know your thoughts about this!

P/S Since arrays are now returned instead, the tests became a little unwieldy since I had to use the java `String`'s `join` method to check against the expected output :/ Not sure if there's a better way to go about doing this...? Particularly since I notice that OpenRefine's testing seems to be quite lacking at the moment :P

thadguidry:
@joanneong I revoke my previous comment that STOP should also be output.  The range STOP and the output is inconsistent once I played with your changes.  I strongly agree with @wetneb that we keep our 0-based indexing that we have throughout OpenRefine.  So a 5-element range call looks like

`range(5) --> ["0", "1", "2", "3", "4"]`

and not

`range(5) --> ["0", "1", "2", "3", "4", "5"]`

joanneong:
@thadguidry Just to clarify, does that mean that for something like `range(2, 4)` you should get `["2", "3"]` instead of `["2", "3", "4"]`?

ettorerizza:
@joanneong The current behaviour is : 

GREL
`forRange(2,2,1, e, e) --> []`

Python/Jython
`return range(2,2) --> []`

thadguidry:
@joanneong the expectation is that it works like Python 3.3+ range() does

joanneong:
**UPDATE SUMMARY**

The `range` function has been updated to follow Python's `range` function closely.

1. The range function returns an array and requires users to use `.join(",")` if they want to see it in the cells.

2. The range function can accept 1 to 3 string/integer arguments. Sequences generated include any given range start and exclude any given range end.

   (a) 1 argument: 0 becomes the default range start, and the given argument is the range end
e.g. `range(5)` --> `["0", "1", "2", "3", "4"]`
`range("5")` --> `["0", "1", "2", "3", "4"]`
`range(-1)` --> `[""]`

   (b) 2 arguments: the first argument is the range start, and the second argument is the range end. If the range start is smaller than the range end, the sequence generated is an increasing sequence (increments by 1 each time). Otherwise, an empty array is returned.
e.g. `range(-1, 5)` --> `["-1", "0", "1", "2", "3", "4"]`
`range(5, -1)` --> `[""]`

   (c) 3 arguments: the first argument is the range start, the second argument the range end, and the third argument the step. Step 0 will produce an empty array. Steps that are bigger than the range will produce only the range start.
e.g. `range(1, 5, 2)` --> `["1", "3"]`
`range(5, 1, -2)` --> `["5", "3"]`
`range(1, 5, 0)` --> `[""]`
`range(1, 5, 100) --> ["1"]`

3. Any permutation of strings and integers is possible when using > 1 arguments for the range function.
e.g. `range("1", 5, "2")` --> `["1", "3"]`
`range(1, "5, 2")` --> `["1", "3"]`
`range("1", "5, 2")` --> `["1", "3"]`
etc.
 
4. Whitespaces are ignored as long as the string value(s) is/are valid.

5. Only the comma separator is supported to reduce code complexity. Dash operators can be replaced with commas by using a simple replace transformation prior to using the range function.

Tests have been updated to check for these functions. Do let me know if there is still something else that I have missed out!

ettorerizza:
> can you remind me why range returns an array of strings and not integers?

I thought first it was to facilitate concatenations, but I forgot that the `join()` function is a good girl and it automatically cast into strings.

joanneong:
@wetneb @ettorerizza 

I am aware that I could have returned either a string array or an integer array, so I just picked string because I felt that most of the inputs are likely to be strings (i.e. most likely from the data/cells rather than directly inputting integers in the transform box) so I might as well keep the output as a string array...

Ah but in that case `Assert.assertArrayEquals` seems to be more convenient for comparing integer arrays :P  

What do you mean by an empty list? The reason why the array is empty but no error is thrown is because all the arguments are valid integers (not decimal etc.) but the step is negative even though the range start is smaller than the range end. The equivalent example in my unit test is `range(1, 5, -1)` -> `[""]`. Also, Python's `range` function exhibits this behaviour too (or at least when I tested [here](https://repl.it/repls/BrilliantRundownPackage))

joanneong:
@wetneb 

I went to think about it again, and I think another possible reason for using string arrays might be to improve performance. If I were to use an integer array, I have to use an `Integer` array rather than an `int` array because `Integer` is an `Object` but `int` is a primitive. This will require auto-boxing and auto-unboxing operations which can *possibly* make things slower, but I'll have to test this to be sure.

In any case, I do agree about the consistency, so I have updated the `range` function accordingly. So yes, now `range(1, 5, -2)` will return `[ ]`. Always ready for more suggestions!

thadguidry:
@joanneong @wetneb What if we just added a Range Type as well ?  We could look into using existing libraries to gain benefits for future expansion and also add a Range Type so we can do map-reducing Range overlap/compare/etc once we implement Beam support later this year ?
1. https://github.com/google/guava/wiki/RangesExplained
2. https://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/Range.html

I think (in a new issue someone can create) we probably want to additionally have users tell us from a new drop-down menu option 

Edit cells -> Common transforms -> To range 

that a column is full of Range type values.  Because I'm seeing this kind of data about Ranges in the wild in scientific data sets both old and new... So to make @joanneong life even harder on this issue .... :)

```

1-9      <-- The OP's original format and why we have this issue 
[1-9]
1..9
(1..9)
[1..9]    <-- Not seeing this one in any datasets however
```

thadguidry:
@wetneb Yes of course, but what we have historically tried to do with GREL has always been giving easy power commands to help our users.  I feel strongly that having a Range type will benefit our users and allowing them to work with Range types and specifically in scientific data sets, [Interval Notation](https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals), a bit easier.  By they way, I did find a generator, http://commons.apache.org/proper/commons-math/apidocs/org/apache/commons/math4/util/IntegerSequence.Range.html

PR #1495, opened 2018-02-16
Imported base classes required by a batch processing

fpompermaier:
I imported the minimal set of core classes required in order to corretcly deserialize history objects.
In the test now I'm able to correctly deserialize the history entry to ColumnAdditionOperation.

Unfortunately Refine operations doesn't provide any getters..should we add them? What do you think?

### DIFF
--- a/openrefine-beam/pom.xml
+++ b/openrefine-beam/pom.xml
@@ -252,7 +252,20 @@
 	</profiles>
 
 	<dependencies>
-		<!-- Adds a dependency on the Beam SDK. -->
+		<!-- https://mvnrepository.com/artifact/org.apache.commons/commons-lang3 -->
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-lang3</artifactId>
+			<version>3.7</version>
+		</dependency>
+		<!-- https://mvnrepository.com/artifact/org.clojure/clojure -->
+		<dependency>
+			<groupId>org.clojure</groupId>
+			<artifactId>clojure</artifactId>
+			<version>1.9.0</version>
+		</dependency>
+
+
 		<dependency>
 			<groupId>org.apache.beam</groupId>
 			<artifactId>beam-sdks-java-core</artifactId>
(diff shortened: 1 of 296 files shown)


wetneb:
@fpompermaier hmm it's not clear to me why you are duplicating most of OpenRefine's source code in a new folder? didn't you want to break down the existing model in reusable parts?

The advantage of working on the existing files means that your commits are more portable: for instance, if you realize that getters are needed somewhere in the existing classes, there are chances that we might want these changes to end up in master. If you work on a separate copy (as done in this PR), we simply cannot cherry-pick the commit…

thadguidry:
@fpompermaier I agree with @wetneb and also thought you were going to just be leveraging against our existing model and adapting and improving in this new branch.  I then wanted to look at this new branch under the eyes of Structure101 to begin to see the new forms and where we can refactor, and also ask David Huynh or Google Dataflow team their thoughts or pitch them questions in email as we go along.

fpompermaier:
Actually my work is not a normal PR, it's more like an experimental branch related to a very first attempt of integation between OR and Apache Beam. I replicated almost all core classes because, CURRENTLY, it is bot possible to import them via maven. I just wanted to see which were the minimal set of classes required to support grel expressions...The goal of this repo is basically to have a playground where to prematurely highlight what will be the problems that we will encounter in the future. For example: should we support current history replication and grel expressions? Should we use the current row model or should we use something like Arrow?
I don't know if I made myself clear...

wetneb:
@fpompermaier yes I can see this is not a normal PR - that's why I don't want you to do any PR at all! Just push to your branch, do your experiments, and that's fine. There is nothing to review here, so it's better if we can focus on PRs where review is required.

If you want to raise issues about the structure of OpenRefine, feel free to use the GitHub issues for that: we can discuss there.

thadguidry:
@fpompermaier In that case, just push this again directly to the branch OpenRefine/OpenRefine:beamusupscotty where I have already given you push rights.  I'll close this PR since we don't need Review, and the expectation is you will do your work directly in beamusupscotty where you have full control and ownership of that branch under our Repo. (You are now officially a collaborator for that branch)  If you need me to send the collaborator email invite again because you didn't receive it for some reason, then just let me know.

Issue #1491, opened 2018-02-15
Ensure consistent behaviour on populating a cell with an empty string


ostephens:
From #820 

If you try to populate a cell in OpenRefine with an empty string, there is different behaviour in different circumstances. Specifically if the cell already contains null, then doing either of:

* Edit cells -> Transform and put ""
* Edit cells -> Transform and " ".trim()

Results in the cell still containing null.

However, if the cell contains a non-null value, then the two operations above result in the cell being populated with an empty string.

Both transformations should always result in the cell containing an empty string and this behaviour should be consistent whether the cell initially contains a null or non-null value **<-- THAD HAS ISSUE WITH THIS STATEMENT...User knows best, we should let user decide, avoid saying "always" which usually implies MAGIC for us behind the scenes and we traditionally try to avoid that.**

ostephens:
It looks like it is 'sameValue' in main/src/com/google/refine/expr/ExpressionUtils.java that leads to the current behaviour. This treats a null value and a zero length string value as being the same.

It's easy to change this, but before doing this, I'm trying to think of any implications of ceasing treating these as equivalent here.

@thadguidry do you have any background info on why these are currently treated as equivalent when comparing values?

thadguidry:
@ostephens Yeap, the background for treating them the same was to make simplifying the GREL handling characteristics.  We assumed that most users of OpenRefine did not care about NULL and just wanted Blanks or ZERO LENGTH STRING.  However, we now know, 8 years later, that these should actually be handled distinctly, since that has risen to be the expectation from most journalist, data scientists, etc.  Basically, we assumed too early that the world didn't care much about NULL as much as we did.  We were wrong.  They care :)  So we can now go back and change this assumption so that NULL and ZERO LENGTH STRING will be treated distinctly in OpenRefine.  Its a minor change to how OpenRefine treats the semantic meaning of NOTHING.  For purists and data scientists, its NULL.  For journalists and the vast majority of our users its ZERO LENGTH STRING.  This will take care of Export value issues automatically, since users can now explicitly convert NULL references to BLANK or whatever String value they want (like what I do sometimes <NULL>.

Issue #1489, opened 2018-02-14
Relabel 'Blank out cells' in Common Transformations menu to 'Blank out cells (null)' 


ostephens:
From #820 

In OpenRefine when term 'Blank' is used in relation to cells it is generally used to mean "either null or containing an empty string". This makes sense in the context of 'Facet by Blank' or 'isBlank'. However, it does not make sense when referring to setting the value of a cell - as you are either setting it to null or to an empty string - we should be explicit about what is happening here.

Issue #1488, opened 2018-02-14
Implement 'Set cells to empty string' command and add to Common Transformations menu


ostephens:
From #820

A "Set cells to empty string" function would replace the content of the cells in all current rows to an empty string "". This would work in a similar way to the "Blank out cells" command in the Common Transformations menu but would set the value of the cells to an empty string, rather than null

ostephens:
May require #1491 to be done first, as simply using the TextTransform operation here (as with the current 'Blank cells' function) results in the inconsistent behaviour described in #1491 (null is left as null rather than replaced with an empty string)

Issue #1487, opened 2018-02-14
Implement "Facet by empty strings" Command and add to Customised Facets menu


ostephens:
From #820 

A "Facet by empty string" function would create a facet with the values 'true' (the cell value contains an empty string) and 'false' (the cell value does not contain an empty string). It would work similarly to the existing 'Facet by Blank' option, but would only include cells containing a value consisting of an empty string, not those containing null

PR #1500, opened 2018-02-18
Blank out

ostephens:
Implements 
#1488 
#1489 
#1491 

Non-english language translation files will need new "blank-out-empty" string adding/translating
Where possible existing labels for "blank-out" have had " (null)" add to the end, but not for non-latin alphabets


ostephens:
Thanks @thadguidry I've done what I can with the translations - but some I've not been able to update/add.

thadguidry:
@ostephens Hmm, OK, now I'm remembering WHY the conglomeration of Blank and Null a bit more.... lolol.   What is the plan for the Text Facet, Numeric Facet, etc however ?  I think we should additionally support showing a label for **(null)** as well, since we show (blank).  If its not in your plans on any of these issues you created, then can you please add 1 more issue for that ? Thanks :)

capture

The other thing to note is that in our other Common transforms -> "to somethings" , we are converting (blank) to (null) , although I haven't checked our code.   I just tried this with having a column of numbers, blanks, and nulls and using "to number" on 2.8 release and it converted my (blank) cells to (null) explicitly.

ostephens:
@thadguidry I think #1486 and #1487 cover the first point - I was going to look at these after this PR was done.

This PR doesn't change the current behaviour of 'toDate', 'toNumber' or 'toString', but now I've looked at them toDate and toNumber are written in a slightly odd way, and with toNumber I think lead to unexpected results (converting an empty string toNumber results in null). I'll create issues for these separately.

Issue #1507, opened 2018-02-21
toDate returns null if run with zero arguments


ostephens:
When reporting a bug please provide the following information to help reproduce the bug:

Version of OpenRefine used (Google Refine 2.6, OpenRefine2.8, an other distribution?):
-----
2.8

Operating Systems and version:
-----
All

Browser + version used - Please note that OpenRefine doesn't support Internet Explorer but works OK in most cases:
-----
All

Steps followed to create the issue:
-----
In the transformation expression window type `toDate()`

Current Results:
-----
`null`

Expected Results:
-----
This should give the error indicating that toDate() requires at least one argument which is an object (usually a string?)

Issue #1505, opened 2018-02-21
toNumber handling of empty string


ostephens:
Currently using toNumber function with an empty string `"".toNumber()` results in `null`. This is  incorrect. Rather the function should return the error "Cannot parse to number" as it would with any other string that cannot be parsed to a Long or Double

thadguidry:
@ostephens right, and not convert the users data silently, as its currently doing with this bug.

PR #1515, opened 2018-02-21
Conversions

### LINKED ISSUE #1506: toNumber returns null if run with zero arguments

Version of OpenRefine used (Google Refine 2.6, OpenRefine2.8, an other distribution?):
-----
2.8

Operating Systems and version:
-----
All

Steps followed to create the issue:
-----
In the transformation expression window type `toNumber()`

Current Results:
-----
The outcome is `null`

Expected Results:
-----
This should give the error indicating that toNumber() requires a single argument which is an object of some type (usually a string?)

ostephens:
Correct behaviour of toNumber and toDate so returns errors rather than nulls if wrong number of arguments provided or attempt to parse null or empty strings.

Also minor change to toString function to look up function name when returning error

Fixes #1506 #1507 #1505

### DIFF
--- a/main/src/com/google/refine/expr/functions/ToString.java
+++ b/main/src/com/google/refine/expr/functions/ToString.java
@@ -43,6 +43,7 @@ import org.json.JSONException;
 import org.json.JSONWriter;
 
 import com.google.refine.expr.EvalError;
+import com.google.refine.grel.ControlFunctionRegistry;
 import com.google.refine.grel.Function;
 import com.google.refine.util.StringUtils;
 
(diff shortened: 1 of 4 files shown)


thadguidry:
Screenshot of test
capture

main/src/com/google/refine/expr/functions/ToString.java
```diff
@@ -68,7 +69,7 @@ public Object call(Properties bindings, Object[] args) {
                 } 
             }
         }
-        return new EvalError("ToString accepts an object and an optional second argument containing a date format string");
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " accepts an object and an optional second argument containing a date format string");
```
thadguidry: Ah, now I see, just the empty function call without args...but can we make these translatable ?

capture3

wetneb: Unfortunately not, we would need to work on backend localization and that is a non-trivial task.

Issue #1509, opened 2018-02-21
NullPointerException when importing an XML file due to the Data Package integration


wetneb:
When importing an XML file with OpenRefine (master branch, https://github.com/OpenRefine/OpenRefine/commit/ad06d4c7a3b2ea53d8b302f4c758fac1d991146c), I get the following exception:

```
10:18:55.608 [                   refine] POST /command/core/importing-controller (18765ms)
Exception in thread "Thread-6" java.lang.NullPointerException
	at io.frictionlessdata.tableschema.TypeInferrer.infer(TypeInferrer.java:161)
	at com.google.refine.importing.ImportingUtilities.createProjectSynchronously(ImportingUtilities.java:1137)
	at com.google.refine.importing.ImportingUtilities.access$1(ImportingUtilities.java:1083)
	at com.google.refine.importing.ImportingUtilities$7.run(ImportingUtilities.java:1075)
```

This is a critical bug because Data Package integration breaks an unrelated existing importer.

The following file can be used to reproduce the bug:
http://pintoch.ulminfo.fr/ebae3e8b22/jorf.xml

jackyq2015:
@wetneb Thanks for the sample. will look into this.

Issue #1517, opened 2018-02-23
JSON deserialization exception due to the upgrade of org.json library in data package PR


wetneb:
Yet another bug introduced by the data package PR:

OpenRefine is now unable to deserialize its own StandardReconConfig from JSON:
```
16:14:54.821 [             recon-config] Reconstruct failed (84ms)
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.google.refine.model.recon.ReconConfig.reconstruct(ReconConfig.java:100)
	at com.google.refine.operations.recon.ReconOperation.reconstruct(ReconOperation.java:82)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.google.refine.operations.OperationRegistry.reconstruct(OperationRegistry.java:83)
	at com.google.refine.history.HistoryEntry.load(HistoryEntry.java:165)
	at com.google.refine.history.History.load(History.java:314)
	at com.google.refine.model.Project.loadFromReader(Project.java:210)
	at com.google.refine.model.Project.loadFromInputStream(Project.java:180)
	at com.google.refine.io.ProjectUtilities.loadFromFile(ProjectUtilities.java:187)
	at com.google.refine.io.ProjectUtilities.loadDataFile(ProjectUtilities.java:123)
	at com.google.refine.io.ProjectUtilities.load(ProjectUtilities.java:135)
	at com.google.refine.io.FileProjectManager.loadProject(FileProjectManager.java:258)
	at com.google.refine.ProjectManager.getProject(ProjectManager.java:507)
	at com.google.refine.commands.Command.getProject(Command.java:181)
	at com.google.refine.commands.project.GetProjectMetadataCommand.doGet(GetProjectMetadataCommand.java:56)
[...]

Issue #1508, opened 2018-02-21
Cannot open projects created with current Master branch build with OR 2.8


ostephens:
Version of OpenRefine used (Google Refine 2.6, OpenRefine2.8, an other distribution?):
-----
Current build from Master branch

Operating Systems and version:
-----
MacOS High Sierra

Browser + version used - Please note that OpenRefine doesn't support Internet Explorer but works OK in most cases:
-----
All

Steps followed to create the issue:
-----
Build OR from current Master branch. Run, and create a project.
Close down OR and run OR 2.8
Try to open the project you've just created

If you are allowed and are OK with making your data public, it would be awesome if you can include the data causing the issue or a URL pointing to where the data is (if your concerned about keeping your data private, ping us on our mailing list):
-----

Current Results:
-----
Trying to open projects created with the current Master branch build with OR2.8 results in a null pointer exception in the console:

```
09:47:41.807 [                   refine] GET /command/core/get-all-project-metadata (2199ms)
09:47:41.810 [                   refine] GET /command/core/get-version (3ms)
09:47:41.813 [                   refine] GET /command/core/get-preference (3ms)
09:47:41.832 [                   refine] GET /command/core/get-languages (19ms)
09:47:41.922 [                   refine] POST /command/core/load-language (90ms)
09:47:41.930 [                   refine] POST /command/core/load-language (8ms)
09:47:41.974 [                   refine] POST /command/core/get-importing-configuration (44ms)
09:47:58.295 [                   refine] GET /command/core/get-project-metadata (16321ms)
09:47:58.297 [                   refine] GET /command/core/get-models (2ms)
09:47:58.303 [                   refine] GET /command/core/get-history (6ms)
09:47:58.303 [          org.mortbay.log] /command/core/get-history (0ms)
java.lang.NullPointerException
09:47:58.309 [                   refine] GET /command/core/get-preference (6ms)
09:47:58.440 [                   refine] POST /command/core/load-language (131ms)
09:47:58.450 [                   refine] POST /command/core/load-language (10ms)
09:47:58.495 [                   refine] GET /command/core/get-history (45ms)
09:47:58.496 [          org.mortbay.log] /command/core/get-history (1ms)
java.lang.NullPointerException
09:47:58.496 [                   refine] POST /command/core/get-rows (0ms)
09:48:31.303 [           ProjectManager] Saving some modified projects ... (32807ms)
java.lang.NullPointerException
java.lang.NullPointerException
[...]

PR #1518, opened 2018-02-23
Fix issue #1509

jackyq2015:


thadguidry:
@jackyq2015 this will fix for any importer ?

jackyq2015:
This is not to do with the actual memory the data set consumed. It will only take 100 rows as the sample to infer the type. After that it will be gced. 

@thadguidry It will apply to all importers.

wetneb:
Oh okay - I had not noticed the copying indeed. So I think this should solve this particular issue - but I am still worried about the overall instability brought by the Data Package PR. What is now clear is that it has introduced bugs that are unrelated to Data Packages.

The risk is high that some of the bugs it has introduced have still not been discovered and could make it into the next release. I tried my best to scrutinize the PR but it was very hard given its poor state. The fact that we have serialization issues is serious - potentially, that could lead to corrupting users' projects and losing data. This is a high risk to pay for a feature that will go mostly unnoticed given the lack of integration in the UI.

thadguidry:
@wetneb Besides reverting the PR... what are you proposing to improve the PR ? More tests ?  An alternate form of testing this somehow or seeing the impacts ?  How can we bring confidence to the importers and Data Package / Metadata enhancements ?

wetneb:
* Each update of dependency should be made in its own PR, explaining why the update is needed and assessing how the library is currently used in the code. It does not need to be extremely detailed, of course: it's just about assessing what are the sort of changes made in the library (any breaking ones) and see what sort of risk it implies for us. For instance, when we updated Jython it was fairly clear that we were touching a fairly well isolated part of the code base, and we saw that their new release included mostly bug fixes, so that's good. It's just an analysis that needs to be done.
* No dead code should be introduced - currently there are loads of constraint checkers that are not exposed at all in the UI: it's much safer just not to include them at all.
* iIf the Data Package library we rely on is not quite ripe yet, fix it. Yes, that takes time, but it's necessary to avoid ugly hacks like the use of reflection to handle datatypes.
* More unit testing. A lot more.
* Commits should reflect logical changes and have meaningful commit messages. That makes reviewing much easier: you can just look at each commit in order, understand its logic and assess it independently of the rest. It's impossible to do that if the PR is crippled with unrelated commits and merges: you have to review the final diff and it's much harder to spot problems.

Generally speaking, just pretending to do some quality software engineering…

jackyq2015:
@wetneb It is interesting to see that you say it is not related to the data package.  It looks like you are the only guy who can do the quality software work in the world.  I am speechless at your attitude. that's all I can say.

wetneb:
Well, it's related because it was introduced by your PR, but take the standpoint of a user: they just want to import an XML file and they don't care at all about Data Packages. These are two unrelated things for users. What I mean is that it is a lot more serious than any bug in the Data Package exporter for instance (which are not going to disturb any existing user workflow).

Sorry if it sounds like I am trying to teach you a lesson - I was just trying to give concrete points to @thadguidry. I am just trying to point out some elementary principles that could have helped avoid the critical bugs that you have introduced in OpenRefine. Bugs happen, it's a normal thing - but these ones are really serious and you are getting paid for this work. There is just a minimum quality standard that should be met.

I clearly have a lot to learn myself about good practices in software engineering - and I will soon submit a PR for the Wikidata package where your criticism will be valued. I am happy to skype today if you want to discuss things directly.

thadguidry:
@jackyq2015 I don't think anyone is displaying an attitude.  Antonin is trying to help improve things and had legitimate concerns as did I, but most of those concerns are removed.  Let's not be too harsh, after all, its only the few of us that I can count on 1 hand. :) right ?  And thanks to both of you for even caring about this little project called OpenRefine that we all work so hard on !

OK, so, let's get to work as the awesome team we are ... we know we need to improve the tests here, around Project Creation and Importers... @jackyq2015 any other areas that you feel would help to assure quality in future PR's around Data Package and Metadata that you can think of ?  Also, Antonin brought up a good point about the Data Package java library...should we invest some time to help fix that up, anything within that concerned you while working with it Jacky ?

jackyq2015:
The first thing I want to address is that this change is integral part of the data package change. It is NOT unrelated change. Since we need to apply the types infer defined by data package and that has to be done at the import stage. From the QA aspect, I agree we need more UT and that won't solve all the problem. That part of code  is covered by UT and not that that similar test data which can break it. For this particular case, I will write a UT to cover it.

thadguidry:
@jackyq2015 Awesome, thanks Jacky. that will help.

wetneb:
@jackyq2015 sorry yes, I should have made clear that I am actually more worried about the other issues #1517 and #1508 than this one…

And just to be clear: the tests that were already in place in OpenRefine are really sparse (the test coverage is so low!), so it really does not help us in our work… typically, this bug should have been caught by an existing test for the XML importer (one that you should not have needed to write)!

jackyq2015:
@thadguidry The infer only impact the column model, the row model is impacted.

PR #1524, opened 2018-03-01
support project backward compatibility to old version of OpenRefine

jackyq2015:


ostephens:
This fixes #1524. There is still some slight oddness with date/time when switching between versions, but it is no longer preventing the project loading in older versions of the s/w

ostephens:
@tfmorris fair point - I've added an issue to make sure this is tackled

PR #1525, opened 2018-03-02
Google api

jackyq2015:
Part of OpenRefine
(Phase 1 enhancements)
https://docs.google.com/document/d/1UwoT1nFk9zwqSlH8rmqKPmiLS2Liw7-KM5HTKD2VVi8/edit


thadguidry:
@jackyq2015 awesomeness !  keep those fingers going !

extensions/gdata/module/langs/translation-en.json
```diff
@@ -44,6 +44,7 @@
 	"gdata-exporter": {
 		"uploading": "Uploading...",
 		"upload-error": "Upload error: ",
+  		"upload-success": "Project was exported successfully to google drive with Id ",
```
thadguidry: @jackyq2015 Perhaps this should say "uploaded" instead of "exported" ?

extensions/gdata/src/com/google/refine/extension/gdata/GDataImportingController.java
```diff
@@ -128,13 +93,11 @@ private void doListDocuments(HttpServletRequest request, HttpServletResponse res
...
-                listSpreadsheets(GDataExtension.getDocsService(token), writer);
+                listSpreadsheets(GoogleAPIExtension.getDriveService(token), writer);
                 listFusionTables(FusionTableHandler.getFusionTablesService(token), writer);
-            } catch (AuthenticationException e) {
-                TokenCookie.deleteToken(request, response);
-            } catch (ServiceException e) {
-                e.printStackTrace();
-            } finally {
+            }  catch (Exception e) {
+                logger.error("doListDocuments exception:" + ExceptionUtils.getStackTrace(e));
```
thadguidry: @jackyq2015 Nice, so we know which method exactly failed.

extensions/gdata/src/com/google/refine/extension/gdata/GDataImportingController.java
```diff
@@ -197,95 +168,56 @@ private void listFusionTables(Fusiontables service, JSONWriter writer)
...
-        try {
-            JSONObject result = new JSONObject();
-            JSONObject options = new JSONObject();
-            JSONUtilities.safePut(result, "status", "ok");
-            JSONUtilities.safePut(result, "options", options);
-            
-            JSONUtilities.safePut(options, "skipDataLines", 0); // number of initial data lines to skip
-            JSONUtilities.safePut(options, "storeBlankRows", true);
-            JSONUtilities.safePut(options, "storeBlankCellsAsNulls", true);
+        JSONUtilities.safePut(result, "status", "ok");
```
thadguidry: @jackyq2015 nothing about encoding options ? needed / supported ?

extensions/gdata/src/com/google/refine/extension/gdata/GoogleAPIExtension.java
```diff
@@ -0,0 +1,205 @@
...
+import com.google.api.services.sheets.v4.Sheets;
+import com.google.api.services.sheets.v4.SheetsScopes;
+
+import com.google.refine.util.ParsingUtilities;
+
+import edu.mit.simile.butterfly.ButterflyModule;
+
+abstract public class GoogleAPIExtension {
+    protected static final String SERVICE_APP_NAME = "OpenRefine-Google-Service";
+    private static final String CLIENT_ID = "455686949425-d237cmorii0ge8if7it5r1qijce6caf0.apps.googleusercontent.com";
```
thadguidry: @jackyq2015 hmm, where how are these tokens maintained ? oh wait, are these even tokens , or registered ids with Google Apps somewhere ? I'm just wondering if Tom maintained an account for this stuff or not , and if we should know about this more from him ?

extensions/gdata/src/com/google/refine/extension/gdata/GoogleAPIExtension.java
```diff
@@ -0,0 +1,205 @@
...
+      }
+    
+    static public Drive getDriveService(String token) {
+        GoogleCredential credential = new GoogleCredential().setAccessToken(token);
+        
+        return new Drive.Builder(HTTP_TRANSPORT, JSON_FACTORY, credential).setHttpRequestInitializer(new HttpRequestInitializer() {
+            @Override
+            public void initialize(HttpRequest httpRequest) throws IOException {
+                credential.initialize(httpRequest);
+                httpRequest.setConnectTimeout(300 * 60000);  // 300 minutes connect timeout
```
thadguidry: @jackyq2015 Is this timeout value still what Google recommends ?  I know that Apache HTTP has a internal default of 3 minutes timeout that can be overridden in Apache's newer package version.

jackyq2015:
Will fix the conflicts after review

extensions/gdata/src/com/google/refine/extension/gdata/GoogleAPIExtension.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.regex.Pattern;
+
+import javax.servlet.http.HttpServletRequest;
+
+import com.google.api.client.auth.oauth2.AuthorizationCodeResponseUrl;
+import com.google.api.client.googleapis.auth.oauth2.GoogleAuthorizationCodeRequestUrl;
+import com.google.api.client.googleapis.auth.oauth2.GoogleAuthorizationCodeTokenRequest;
+import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
+import com.google.api.client.googleapis.auth.oauth2.GoogleTokenResponse;
+import com.google.api.client.http.HttpRequest;
```
thadguidry: @jackyq2015 Hmm, should we care about thread safety here ?
Their https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/http/apache/package-summary
is thread safe based on Apache V4,
 versus
https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/http/HttpRequest

extensions/gdata/src/com/google/refine/extension/gdata/SpreadsheetSerializer.java
```diff
@@ -0,0 +1,121 @@
...
+        acr.setFields("*");
+        acr.setSheetId(0);
+        acr.setRows(rows);
+        batchRequest.setAppendCells(acr);
+        
+        requests.add(batchRequest);
+    }
+    
+    public String getUrl() {
+        return "https://docs.google.com/spreadsheets/d/" + spreadsheetId + "/edit#gid=0";
```
thadguidry: @jackyq2015 best to wrap things like this with URL encode as a best practice.

extensions/gdata/src/com/google/refine/extension/gdata/UploadCommand.java
```diff
@@ -139,16 +104,70 @@ static private String upload(
...
+            
+            fileMetadata.setName(name)
+                .setDescription(METADATA_DESCRIPTION)
+                .setContentHints(contentHints);
+            FileContent projectContent = new FileContent("application/zip", filePath);
+            File file = GoogleAPIExtension.getDriveService(token)
+                    .files().create(fileMetadata, projectContent)
+                .setFields("id")
+                .execute();
+            logger.info("File ID: " + file.getId());
```
thadguidry: @jackyq2015 let's mention the method also while logging for easier debugging, like you did before.

main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -88,10 +88,13 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.beust.jcommander.Strings;
```
thadguidry: @jackyq2015 Why JCommander and not just use [Apache StringUtils methods](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isNoneBlank-java.lang.CharSequence...-) for null conversion where you need it ?

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -272,7 +272,10 @@
         "char-enc": "Character encoding",
         "line-sep": "Line separator",
         "upload-to": "Upload to",
-        "json-text": "The following JSON text encodes the options you have set in the other tabs. You can copy it out and save it for later, and paste it back in and click Apply to re-use the same options."
+        "json-text": "The following JSON text encodes the options you have set in the other tabs. You can copy it out and save it for later, and paste it back in and click Apply to re-use the same options.",
+        "choose-export-destination": "Please choose the destination for project export",
+        "export-to-local": "Export to local",
+  		"export-to-google-drive": "Export to google drive"
```
thadguidry: @jackyq2015 Always capitalize proper nouns and registered trademarks, as a best practice. "Google Drive"

main/webapp/modules/core/scripts/project/exporters.js
```diff
@@ -169,23 +169,85 @@ ExporterManager.prepareExportRowsForm = function(format, includeEngine, ext) {
...
+  
+  _elmts.cancelButton.click(function() { DialogSystem.dismissAll(); });
+  
+  DialogSystem.showDialog(dialog);
+  
+  // save to google drive
+  var doExportToGoogleDrive = function() {
+      var name = window.prompt(prompt, theProject.metadata.name);
+      if (name) {
+        var dismiss = DialogSystem.showBusy($.i18n._('gdata-exporter')["uploading"]);
```
thadguidry: @jackyq2015 maybe add some extra dots, "uploading..." ?

extensions/gdata/module/langs/translation-en.json
thadguidry: @jackyq2015 also capitalize "Google Drive" since it's a registered trademark.

jackyq2015: updated the json

extensions/gdata/src/com/google/refine/extension/gdata/GDataImportingController.java
jackyq2015: There is no encoding options to get a spreadsheet:
https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets/get

I think it make sense because all the data can be stored and unified to utf-8 by google service/google API.

extensions/gdata/src/com/google/refine/extension/gdata/GoogleAPIExtension.java
jackyq2015: It is a registered ids but with new set of APIs enabled. I created the Id

main/webapp/modules/core/langs/translation-en.json
jackyq2015: done

main/src/com/google/refine/importing/ImportingUtilities.java
jackyq2015: we do not need this. removed.

extensions/gdata/src/com/google/refine/extension/gdata/UploadCommand.java
jackyq2015: This is a INFO level logger. should we need that?

extensions/gdata/src/com/google/refine/extension/gdata/GoogleAPIExtension.java
jackyq2015: well, it is not recommendation from Google. 300 mins seems too much, Let me find out what value is better.

main/webapp/modules/core/scripts/project/exporters.js
jackyq2015: It is already there in json file.

extensions/gdata/src/com/google/refine/extension/gdata/GoogleAPIExtension.java
jackyq2015: My understanding is HttpRequest uses the HttpTransport:
https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/http/HttpTransport.html

HttpTransport is underline layer. so HttpRequest should also thread safe

extensions/gdata/src/com/google/refine/extension/gdata/SpreadsheetSerializer.java
jackyq2015: done

extensions/gdata/src/com/google/refine/extension/gdata/GoogleAPIExtension.java
jackyq2015: set to 3 mins. it make more sense.

extensions/gdata/src/com/google/refine/extension/gdata/GoogleAPIExtension.java
wetneb: Is this tied to your own Google account somehow? And is this token really meant to be public? (Could other applications maliciously reuse this token?) (Not that I can see any way to avoid that…)

extensions/gdata/src/com/google/refine/extension/gdata/SpreadsheetSerializer.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+        
+        BatchUpdateSpreadsheetRequest requestBody = new BatchUpdateSpreadsheetRequest();
+        requestBody.setIncludeSpreadsheetInResponse(false);
+        requestBody.setRequests(requests);
+
+        Sheets.Spreadsheets.BatchUpdate request;
+        try {
+            logger.info("spreadsheetId: " + spreadsheetId);
+            logger.info("requestBody:" + requestBody.toString());
```
wetneb: I assume we should remove this logging because the request body (and response below) could be quite verbose.

extensions/gdata/src/com/google/refine/extension/gdata/SpreadsheetSerializer.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    @Override
+    public void addRow(List<CellData> cells, boolean isHeader) {
+        if (batchRequest == null) {
+            batchRequest = new Request();
+            rows = new ArrayList<RowData>(BATCH_SIZE);
+        }
+        List<com.google.api.services.sheets.v4.model.CellData> cellDatas = new ArrayList<com.google.api.services.sheets.v4.model.CellData>();
```
wetneb: Is there any reason why the full import path is specified here, rather than just `List<CellData>`?
Also, since we have migrated to Java 8, the right-hand side can be simplified to just `new ArrayList<>();`

main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -1132,7 +1133,17 @@ static private void createProjectSynchronously(
...
+                    // convert the null object to prevent the NPE
+                    for (Row row : rows) {
+                        for (int i = 0; i < row.cells.size(); i++) {
+                            Cell cell = row.cells.get(i);
+                            if (cell == null) {
+                                row.cells.set(i, new Cell(StringUtils.EMPTY, null));
+                            }
+                        }
+                        listCells.add(row.cells.toArray());
+                    }
```
wetneb: I'm confused - wasn't this change already merged?

wetneb:
Here is an NPE I got:
```
14:38:51.813 [                   refine] GET /command/core/get-languages (1174ms)
14:38:52.037 [ GDataImportingController] doListDocuments exception:java.lang.NullPointerException
	at com.google.refine.extension.gdata.GDataImportingController.listFusionTables(GDataImportingController.java:152)
	at com.google.refine.extension.gdata.GDataImportingController.doListDocuments(GDataImportingController.java:97)
	at com.google.refine.extension.gdata.GDataImportingController.doPost(GDataImportingController.java:67)
	at com.google.refine.commands.importing.ImportingControllerCommand.doPost(ImportingControllerCommand.java:62)
	at com.google.refine.RefineServlet.service(RefineServlet.java:178)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:820)
	at org.mortbay.jetty.servlet.ServletHolder.handle(ServletHolder.java:511)
	at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1166)
	at org.mortbay.servlet.UserAgentFilter.doFilter(UserAgentFilter.java:81)
	at org.mortbay.servlet.GzipFilter.doFilter(GzipFilter.java:132)
	at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1157)
	at org.mortbay.jetty.servlet.ServletHandler.handle(ServletHandler.java:388)
	at org.mortbay.jetty.security.SecurityHandler.handle(SecurityHandler.java:216)
	at org.mortbay.jetty.servlet.SessionHandler.handle(SessionHandler.java:182)
	at org.mortbay.jetty.handler.ContextHandler.handle(ContextHandler.java:765)
	at org.mortbay.jetty.webapp.WebAppContext.handle(WebAppContext.java:418)
	at org.mortbay.jetty.handler.HandlerWrapper.handle(HandlerWrapper.java:152)
	at org.mortbay.jetty.Server.handle(Server.java:326)
	at org.mortbay.jetty.HttpConnection.handleRequest(HttpConnection.java:542)
	at org.mortbay.jetty.HttpConnection$RequestHandler.headerComplete(HttpConnection.java:923)
	at org.mortbay.jetty.HttpParser.parseNext(HttpParser.java:547)
	at org.mortbay.jetty.HttpParser.parseAvailable(HttpParser.java:212)
	at org.mortbay.jetty.HttpConnection.handle(HttpConnection.java:404)
	at org.mortbay.jetty.bio.SocketConnector$Connection.run(SocketConnector.java:228)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
 (224ms)
14:38:52.756 [                   refine] GET /command/core/get-version (719ms)

```

extensions/gdata/src/com/google/refine/extension/gdata/SpreadsheetSerializer.java
thadguidry: I wouldn't remove it, it useful for debugging.  Instead just log this as DEBUG rather than INFO on slf4j

jackyq2015: will change to DEBUG. @wetneb I sent the info though the email to dev core. The token is to identify the application. Further more, before can use the API you need to use your own email to authorise.

extensions/gdata/src/com/google/refine/extension/gdata/SpreadsheetSerializer.java
jackyq2015: The reason is there is anther CellData in:
com.google.refine.exporters.TabularSerializer interface.

But yes we can skip the things between <> when new

main/src/com/google/refine/importing/ImportingUtilities.java
jackyq2015: This is part of result of the merge. It won't do anything after merge back to master.

wetneb:
@jackyq2015 I propose that we merge this PR as is - the OAuth credentials can be used for dev. For releases, it would be good to generate fresh ones, so that invalidation of these credentials do not affect users. What do you think about that?

jackyq2015:
Yes, I can merge it. Checked with google java client github, seems they don't have solution for this either. But I don't think we need to generate the fresh one for release. Firstly credentials cannot be hidden. Secondly and also most importantly, the blocked Id will be the malicious Ids who abuse it, not the OpenRefine application Id.

wetneb:
Well in OAuth it is always possible for a provider to revoke the application credentials (the ones that you have put in the source code) if these credentials are misused - typically if they are reused in another application for malicious purposes. I know you have not found anything in the docs about that, but that is just how OAuth works in general. That's why I would be in favour of changing the tokens before the release - the tokens you put in the packaged versions will not be as visible as the ones in the source code.

ettorerizza:
Is this related ? https://www.quora.com/Why-is-it-OK-to-expose-the-OAuth-client-ID-and-use-it-as-an-API-key

jackyq2015:
@ettorerizza I think it's related. The point is that though the OAuth, you can not really cause damage to others'(document in this case). The worst case is you use it too much, but again the Google will put a cap on the user but not the application itself(the cap for the application is unlimited according to the Google Console). 

Also, @wetneb I don't quite get when you say "the tokens you put in the packaged versions will not be as visible as the ones in the source code". you mean we get a new token to replace the old one. what's the difference if it is open sourced.

wetneb:
I just don't see any solution where OpenRefine will fully comply with Google's terms and I am therefore proposing a way to mitigate the risk caused by this situation.

jackyq2015:
@thadguidry I am worry about @wetneb worry too much. :)

jackyq2015:
No problem, shall I go ahead to merge it then?

PR #1527, opened 2018-03-02
pass the limit as int from javascript

jackyq2015:


wetneb:
@thadguidry hmm… did you just merge this PR without discussion? I have reverted it.

@jackyq2015 I think it would be better that the server accepts JSON where the limit is either an int or a string. That would be more robust, for instance if people reuse existing JSON edit histories or use OR as a server for batch processing.

thadguidry:
@wetneb I just checked and looks like older versions did actually perform a stringify and set the limit param as a String value and not as an Int.
capture

wetneb:
@thadguidry yes, that's why it's really important to make sure we are maintaining compatibility with previous versions.

thadguidry:
@wetneb thanks for keeping us honest. :)  our old memories are fading, but your young blood keeps the team FRESH :) :)

PR #1529, opened 2018-03-04
Make server accept the empty string for the limit in a reconciliation config

wetneb:
This just adds an extra safety check to make sure we still accept the same commands as the UI used to generate, on top of @jackyq2015's fix for the UI.

### DIFF
--- a/main/src/com/google/refine/model/recon/StandardReconConfig.java
+++ b/main/src/com/google/refine/model/recon/StandardReconConfig.java
@@ -111,8 +111,13 @@ public class StandardReconConfig extends ReconConfig {
         }
         
         JSONObject t = obj.has("type") && !obj.isNull("type") ? obj.getJSONObject("type") : null;
-        
-        int limit = obj.has("limit") && !obj.isNull("limit") ? obj.getInt("limit") : 0;
+
+        int limit;
+        try {
+            limit = obj.has("limit") && !obj.isNull("limit") ? obj.getInt("limit") : 0;
+        } catch(JSONException e) {
+            limit = 0;
+        }
         
         return new StandardReconConfig(
             obj.getString("service"),
--- a/main/src/com/google/refine/model/recon/StandardReconConfig.java
+++ b/main/src/com/google/refine/model/recon/StandardReconConfig.java
@@ -111,13 +111,8 @@ public class StandardReconConfig extends ReconConfig {
         }
         
         JSONObject t = obj.has("type") && !obj.isNull("type") ? obj.getJSONObject("type") : null;
-
-        int limit;
-        try {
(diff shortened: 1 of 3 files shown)


jackyq2015:
@wetneb As far as I can think of, the only scenario of client is history replay through the JSON interface which is keep untouched before and after change change. If you can name a client as an example which still send the string, then we should consider to support it. It is important we can keep the interface clean and unambiguous as possible (of course without service interruption)

wetneb:
@jackyq2015 it just does not cost anything to add this check… it is just about backward compatibility with the UI. No I am not aware of a client that still does this, but who knows? It's cheap for us to keep things compatible, so why would we break them on purpose?

thadguidry:
@wetneb No one breaks things on purpose.  Let's please be patient with everyone here and just try to tweak the defensive attitude a bit, just some personal advice from me to you. :) Both Jacky and I approve of your passion for OpenRefine and of course all of us appreciate everyone's efforts.  Both of you are extremely talented well beyond my capacity and would probably buy each other a beer if you met up in a bar! :)  Jacky's not retorting actually to this PR, his mannerism is different than yours and mine.  He's more factual in communication than you and I, and we both should appreciate his style as it balances with the rest of our team.  If you re-read, he's just thinking out loud and even acknowledging your effort to keep the UI healthy and avoid service interruption and also acknowledging to support clients that sends a string.  He's agreeing, in his own style of communication.

wetneb:
@thadguidry sure! it's clearly not a hill I want to die on. I completely agree with you that it's important to improve our communication. It's important that things get discussed. Be wordy! PRs are meant to be an occasion to discuss things and help each other (or at least that's what I would like to see) but that does not work if we merely use them just because we cannot push directly to master.

jackyq2015:
There are 2 interactions involved here: one is UI to the back end and another is when application load the json from the file. I think we all agree that there is no impact for the latter interaction(ie, there is no service interruption if user want to replay their history because there is nothing changed on the persistent file). The only question is that should we support 2 ways to send the "limit" from the front end to back end? My answer is no. The reasons are:

- This is internal communication. Even later we finish the frond/back end separation, the current front end has to be changed. Backward compatibility is only required if the UI and back end already separated and we want to connect the old version of UI to the back end.  It is not worthy to carry on the legacy debt here without benefit
- The origin purpose it to fix a mistake made before. No need to keep it if there is no impact for historical record or the future record
- Keep the interface clean and unambiguous by only allow the int for limit, which is nature for the programmer. Having the extra lines to accept different types is cheap, but it cause confusion.

wetneb:
Ah, glad to see such a nice long answer! You are making good points, I will remove my commit.

Issue #117, opened 2012-10-15
"Search for match" ignores the reconciliation service you've used and only searches Freebase


tfmorris:
_Original author: rob.02...@gmail.com (August 19, 2010 09:24:23)_

With the development build:
1. Choose a column, select Reconcile -&gt; Start reconciling
2. Choose to reconcile against a non-Freebase reconciliation service
3. For a cell that has multiple matches, click the &quot;search for match&quot; link

My expectation is that this &quot;search for match&quot; would search the reconciliation service I'd chosen, and show me matching results.

Instead &quot;search for match&quot; only searches Freebase.

I'm using the latest project source code as at Aug 19 2010.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=117_
##

thadguidry:
@wetneb Do you know if this functionality is working against Wikidata or any standard reconcile service, since Freebase is now deprecated and you removed some legacy Freebase stuff from our code ?

wetneb:
@thadguidry I don't see why this bug would still be present, because clearly for Wikidata the search works and does not fall back on Freebase… And there is nothing specific about Wikidata in the code (except that the service is configured by default). I think we can close this.

thadguidry:
@wetneb awesome.  Thanks for your input. Closing it.

PR #1530, opened 2018-03-04
Wikidata extension

### LINKED ISSUE #1213: Schema alignment for Wikidata
A schema alignment for Wikidata would enable OpenRefine users to prepare their datasets for inclusion in Wikidata. It should be implemented in a Wikidata extension (inside this repository or not).

* [x] create an OverlayModel that matches Wikibase's model (with native support for qualifiers, references and ranks)
* [x] add a dialog, operation and change to let the user edit this OverlayModel
* [x] enable export to QuickStatements format
- [x] add direct editing capabilities without going through QuickStatements
- [x] add warnings (no references, unreconciled cells…)
<s>- [ ] add delete statements</s>
<s>- [ ] enable configuration of statement mapping (for instance, adding or ignoring qualifier statement)</s>

The last points will be done later as it requires significant work in the underlying library (Wikidata-Toolkit).

wetneb:
This adds an overlay model to OpenRefine which let users translate their projects to Wikidata edits. It is inspired by the RDF and Freebase extensions. It works like this:
* Users can create a schema, pretty much in the same way they would do manual edits on Wikidata, except that they can drag and drop OpenRefine columns in place of the values of their statements, qualifiers, references and terms. These columns will act as variables: for each row, they will be replaced by their value, generating statements and terms.
* They get feedback about the issues their edits may have (before the edits are performed). This largely relies on Wikidata's constraint system, but not all constraint violations will be reported. Some of the issues that are reported do not correspond to any constraint either (for instance, issues with the labels, descriptions and aliases).
* They can export their project to the QuickStatements format (https://www.wikidata.org/wiki/Help:QuickStatements) and perform their edits with that tool.
* They can also perform the edits directly from OpenRefine. This counts as an operation in OpenRefine, so using the JSON history of operations it is effectively possible to write simple bots as OpenRefine scripts, with no knowledge of programming.

In both export modes, new items are supported (using the existing "reconcile to new" option). New items are not created at reconcilation time, only once the rest of the edits are performed.

**Important note regarding testing this extension**

> As this extension introduces new operations, switching back and forth between versions of OpenRefine which contain the extension and those which do not can cause loss of the content of these operations (which makes it impossible to extract JSON from these operations, for instance). Make sure you **backup your workspace** before testing the extension on any important project.

Reviewing this PR on GitHub will probably be quite tedious because of the large number of added files - I guess the simplest way is just to check out the branch directly. I am looking for seasoned Wikidata editors to give it a spin. Documentation also needs to get written somewhere.

Test coverage can be found here: http://pintoch.ulminfo.fr/wikidata-extension-test-coverage/ (and generated with `./refine test` as always).

This closes #1213.


thadguidry:
@wetneb Perhaps this label could be improved with "type item or drag reconciled column here"
capture

thadguidry:
@wetneb Moving the Alignment dialog around, seems like containment isn't made for the Hover Tips ?
capture2

extensions/wikidata/src/org/openrefine/wikidata/exporters/QSValuePrinter.java
```diff
@@ -0,0 +1,106 @@
...
+ * 
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ ******************************************************************************/
+package org.openrefine.wikidata.exporters;
```
jackyq2015: Should the package be "com.google.refine.*" or there is reason for that?

extensions/wikidata/src/org/openrefine/wikidata/exporters/QSValuePrinter.java
```diff
@@ -0,0 +1,106 @@
...
+ * with "LAST". It is fine to do this assumption because we are working on edit
+ * batches previously scheduled by {@link QuickStatementsUpdateScheduler}.
+ * 
+ * @author Antonin Delpeuch
+ *
+ */
+public class QSValuePrinter implements ValueVisitor<String> {
+
+    @Override
+    public String visit(DatatypeIdValue value) {
```
jackyq2015: A unsupported exception should be thrown instead of return null silently

main/src/com/google/refine/model/Recon.java
```diff
@@ -51,8 +51,14 @@
  
 public class Recon implements HasFields, Jsonizable {
     
+    /**
+     * Freebase schema URLs kept for compatibility with legacy reconciliation results
+     */
     private static final String FREEBASE_SCHEMA_SPACE = "http://rdf.freebase.com/ns/type.object.id";
```
jackyq2015: freebase still available?

extensions/wikidata/src/org/openrefine/wikidata/exporters/QSValuePrinter.java
wetneb: Well I thought there wasn't any point in continuing to use "com.google.refine" given that the project has changed name. I don't think we should necessarily migrate the rest of the software to a new package, but for new code I thought it would be nice to feel a bit more at home :) This choice was made before we had funding from the Google News Lab. But I honestly don't care ^^

main/src/com/google/refine/model/Recon.java
wetneb: No, but it can still be referred to: Freebase ids are still in use, for instance on Wikidata. There are still a number of places in the code (and possibly in other extensions / reconciliation services) where this is used. Are you proposing to delete this, and if so why?

extensions/wikidata/src/org/openrefine/wikidata/exporters/QSValuePrinter.java
thadguidry: I'd like to see all the packages renamed eventually.

main/src/com/google/refine/model/Recon.java
thadguidry: The Freebase namespace is still referenced out in the wild.  keep this.

extensions/wikidata/src/org/openrefine/wikidata/editing/ConnectionManager.java
```diff
@@ -0,0 +1,148 @@
...
+ */
+
+public class ConnectionManager {
+
+    public static final String PREFERENCE_STORE_KEY = "wikidata_credentials";
+
+    private PreferenceStore prefStore;
+    private ApiConnection connection;
+
+    private static class ConnectionManagerHolder {
```
jackyq2015: why we need the class to hold the instance?

wetneb:
@jackyq2015 @thadguidry thanks for the comments so far! I forgot to mention that it can good to backup your workspace to test this (see the notice added in the PR description above).

extensions/wikidata/src/org/openrefine/wikidata/editing/ConnectionManager.java
```diff
@@ -0,0 +1,148 @@
...
+    public void login(String username, String password, boolean rememberCredentials) {
+        if (rememberCredentials) {
+            try {
+                JSONArray array = new JSONArray();
+                JSONObject obj = new JSONObject();
+                obj.put("username", username);
+                obj.put("password", password);
+                array.put(obj);
+                prefStore.put(PREFERENCE_STORE_KEY, array);
+                // TODO save preferences (find out how)
```
jackyq2015: No saved? Maybe a cookie is an option?

wetneb: This comment is actually out of date - OR saves preferences itself. Thanks for picking that up though.

extensions/wikidata/src/org/openrefine/wikidata/commands/LoginCommand.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+        response.setCharacterEncoding("UTF-8");
+        response.setHeader("Content-Type", "application/json");
+
+        StringWriter sb = new StringWriter(2048);
+        JSONWriter writer = new JSONWriter(sb);
+
+        try {
+            writer.object();
+            writer.key("logged_in");
```
jackyq2015: Always respond "logged_in" no matter if it is really login in at the back end?

extensions/wikidata/src/org/openrefine/wikidata/editing/ConnectionManager.java
wetneb: Not sure why I introduced this - I'll see if it still works as a plain static member (I don't see why it wouldn't…)

extensions/wikidata/src/org/openrefine/wikidata/commands/LoginCommand.java
```diff
@@ -0,0 +1,77 @@
...
+            writer.value(manager.isLoggedIn());
+            writer.key("username");
+            writer.value(manager.getUsername());
+            writer.endObject();
+        } catch (JSONException e) {
+            e.printStackTrace();
```
jackyq2015: As a best practice, consider to log it instead of printStackTrace.

extensions/wikidata/src/org/openrefine/wikidata/commands/LoginCommand.java
wetneb: yes, this field holds a boolean value that says whether we are logged in or not.

extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
```diff
@@ -0,0 +1,210 @@
...
+            return;
+        }
+        if (batchCursor == currentBatch.size()) {
+            prepareNewBatch();
+        }
+        ItemUpdate update = currentBatch.get(batchCursor);
+
+        // Rewrite mentions to new items
+        ReconEntityRewriter rewriter = new ReconEntityRewriter(library, update.getItemId());
+        update = rewriter.rewrite(update);
```
jackyq2015: A little bit weird to pass "update" and return "update". Can skip the return?

wetneb: what do you mean?

extensions/wikidata/src/org/openrefine/wikidata/editing/ReconEntityRewriter.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public ItemUpdate rewrite(ItemUpdate update) {
+        Set<MonolingualTextValue> labels = update.getLabels().stream().map(l -> copy(l)).collect(Collectors.toSet());
+        Set<MonolingualTextValue> descriptions = update.getDescriptions().stream().map(l -> copy(l))
+                .collect(Collectors.toSet());
+        Set<MonolingualTextValue> aliases = update.getAliases().stream().map(l -> copy(l)).collect(Collectors.toSet());
+        List<Statement> addedStatements = update.getAddedStatements().stream().map(l -> copy(l))
+                .collect(Collectors.toList());
+        Set<Statement> deletedStatements = update.getDeletedStatements().stream().map(l -> copy(l))
```
jackyq2015: should be List here?

extensions/wikidata/src/org/openrefine/wikidata/editing/ReconEntityRewriter.java
```diff
@@ -0,0 +1,102 @@
...
+                            "Trying to rewrite an update where a new item was not created yet.");
+                }
+                return Datamodel.makeItemIdValue(newId, recon.getRecon().identifierSpace);
+            }
+        }
+        return super.copy(value);
+    }
+
+    public ItemUpdate rewrite(ItemUpdate update) {
+        Set<MonolingualTextValue> labels = update.getLabels().stream().map(l -> copy(l)).collect(Collectors.toSet());
```
jackyq2015: For performance, is that possible to process the "update" in one streaming. Also parallel can be used

extensions/wikidata/src/org/openrefine/wikidata/editing/ReconEntityRewriter.java
wetneb: yes, for added statements the order matters.

extensions/wikidata/src/org/openrefine/wikidata/editing/ReconEntityRewriter.java
wetneb: It's not clear to me how to do just one streaming. Can you elaborate?
The overhead of `parallel()` is way too high for the data sizes that we are dealing with here - typically these sets will contain just a bunch of elements (a dozen, say).
https://stackoverflow.com/questions/20375176/should-i-always-use-a-parallel-stream-when-possible#20375622

wetneb:
@thadguidry about the behaviour of the reconciliation suggest service when moving the dialog, this is a generic problem that we also have at other places in the UI (for instance the reconciliation dialog or the "search for match" dialog). I agree it would be nice to fix it but I propose we do that in a separate PR (with the rest of the improvements on reconciliation, like adding descriptions…)

thadguidry:
@wetneb of course.  I was only commenting, I didn't block you. :)

extensions/wikidata/src/org/openrefine/wikidata/editing/ReconEntityRewriter.java
wetneb: sorry I didn't realize your comment was about `deletedStatements`. For deleted statements the order does not matter (because the document will be the same no matter in which order you deleted the statements).

extensions/wikidata/src/org/openrefine/wikidata/commands/LoginCommand.java
jackyq2015: Make sense. I misread this

extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
jackyq2015: I mean:
rewriter.rewrite(update);

wetneb: @jackyq2015 it was like this before, but I have changed EntityUpdate to be immutable (just like the other classes in Wikidata-Toolkit). It's something we don't do much in OpenRefine (everything is public and mutable) but it brings a lot of good things (thread safety, easier testing)…

extensions/wikidata/src/org/openrefine/wikidata/editing/ReconEntityRewriter.java
jackyq2015: It is fine if there is not so many elements.

extensions/wikidata/src/org/openrefine/wikidata/commands/PreviewWikibaseSchemaCommand.java
```diff
@@ -0,0 +1,132 @@
...
+
+            // Evaluate project
+            Engine engine = getEngine(request, project);
+            List<ItemUpdate> editBatch = schema.evaluate(project, engine, warningStore);
+
+            StringWriter sb = new StringWriter(2048);
+            JSONWriter writer = new JSONWriter(sb);
+            writer.object();
+
+            {
```
jackyq2015: No need the {}. or  you want to extract into method?

extensions/wikidata/src/org/openrefine/wikidata/commands/PreviewWikibaseSchemaCommand.java
```diff
@@ -0,0 +1,132 @@
...
+            response.setCharacterEncoding("UTF-8");
+            response.setHeader("Content-Type", "application/json");
+
+            String jsonString = request.getParameter("schema");
+
+            WikibaseSchema schema = null;
+            if (jsonString != null) {
+                try {
+                    JSONObject json = ParsingUtilities.evaluateJsonStringToObject(jsonString);
+                    schema = WikibaseSchema.reconstruct(json);
```
jackyq2015: why not just create a method reconstruct(String jsonString) to avoid string->JSONObject->String->Jackson ObjectMapper->Object conversion?

extensions/wikidata/src/org/openrefine/wikidata/utils/JacksonJsonizable.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public static <T> T fromJSONClass(JSONObject obj, Class<T> klass)
+            throws JSONException {
+        ObjectMapper mapper = new ObjectMapper();
+        String json = obj.toString();
+        try {
+            return mapper.readValue(json, klass);
+        } catch (JsonParseException e) {
+            throw new JSONException(e.toString());
```
jackyq2015: Using another advanced JSON library is fine. But if it is possible we should avoid converting back and force. It is hard to read and debug. Also itself create a hurdle to migrate off the JSONObject. If some places JSONObject is unavoidable(for example as part of the interface signature), maybe use a POJO is easier make more sense.

extensions/wikidata/module/MOD-INF/controller.js
```diff
@@ -0,0 +1,76 @@
...
+            "org.openrefine.wikidata.operations.PerformWikibaseEditsOperation$PerformWikibaseEditsChange",
+            "org.openrefine.wikidata.operations.PerformWikibaseEditsOperation$PerformWikibaseEditsChange");
+    
+    RefineServlet.cacheClass(Packages.org.openrefine.wikidata.operations.SaveWikibaseSchemaOperation$WikibaseSchemaChange);
+    RefineServlet.cacheClass(Packages.org.openrefine.wikidata.operations.PerformWikibaseEditsOperation$PerformWikibaseEditsChange);
+
+    /*
+     *  Attach a Wikibase schema to each project.
+     */
+    Packages.com.google.refine.model.Project.registerOverlayModel(
```
jackyq2015: To EACH project? should overlay model always optional?

extensions/wikidata/src/org/openrefine/wikidata/operations/PerformWikibaseEditsOperation.java
```diff
@@ -0,0 +1,236 @@
...
+import com.google.refine.history.HistoryEntry;
+import com.google.refine.model.AbstractOperation;
+import com.google.refine.model.Project;
+import com.google.refine.operations.EngineDependentOperation;
+import com.google.refine.operations.OperationRegistry;
+import com.google.refine.process.LongRunningProcess;
+import com.google.refine.process.Process;
+import com.google.refine.util.Pool;
+
+public class PerformWikibaseEditsOperation extends EngineDependentOperation {
```
jackyq2015: Maybe I am not quite understand the wikibase. Seems the menu "Edit wikidata schema" is to publish, "Perform the Edit" is actual edit. Am I right?

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ItemUpdateScrutinizer.java
```diff
@@ -0,0 +1,57 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ ******************************************************************************/
+package org.openrefine.wikidata.qa.scrutinizers;
+
+import java.util.List;
+
+import org.openrefine.wikidata.updates.ItemUpdate;
+
+public abstract class ItemUpdateScrutinizer extends EditScrutinizer {
```
jackyq2015: In general, the Scrutinizer's extension chain  is a little bit long. For example this class could consider be flatten. Providing a method to processing a list does not bring too much value. seems not worth another layer. Some abstract class(for example EditScrutinizer.java) can be replaced with interface(java 8 support default method):
https://www.journaldev.com/2752/java-8-interface-changes-static-method-default-method

extensions/wikidata/src/org/openrefine/wikidata/commands/PreviewWikibaseSchemaCommand.java
wetneb: it's just meant to provide some structure to match the JSON's own structure (but I can remove it if you prefer, it is cosmetic)

extensions/wikidata/src/org/openrefine/wikidata/utils/JacksonJsonizable.java
wetneb: We can avoid the conversion at deserialization, but I don't see how to avoid it at serialization. I did try to find a way to convert a JsonNode to a JSONObject but there does not seem to be any simple solution for that. Going through a String isn't great but the overhead isn't much and I don't see any other solution…

extensions/wikidata/module/MOD-INF/controller.js
wetneb: It just means we add an empty one by default, which does not cost anything… I believe that's how it's done in the Freebase and RDF extensions. I don't see how you would "select" which projects to enable?

extensions/wikidata/src/org/openrefine/wikidata/operations/PerformWikibaseEditsOperation.java
wetneb: "Edit Wikidata schema" creates the schema (and saves it as an overlay model with SaveWikibaseSchema), "perform edits" pushes the edits to Wikidata (with this operation).

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ItemUpdateScrutinizer.java
wetneb: Yeah I'm not that happy with that structure, I think I see a different way which should reduce iteration costs.

wetneb:
@thadguidry which ones specifically? It is flooded by cosmetic javascript issues.

thadguidry:
@wetneb scroll just past halfway down that page...then you will begin to see the Java files and issues.

wetneb:
@thadguidry yeah - but typically the issue you point out is debatable (some prefer not to have visibility modifiers there). So which Java issue are you worried about?

thadguidry:
@wetneb thanks for reviewing. your done then.  yes sometimes the issues are debatable. Jacky has his "quality" level. and I have mine. its fine, your done.

wetneb:
@thadguidry yeah, sometimes Codacy is useful… and sometimes it isn't. I guess if we wanted to use it better we could spend time configuring it better so that it only shows the issues we care about… but last time I checked their interface was half broken, I got errors when I tried to do that :-/

thadguidry:
@wetneb and its not always the Yellow ones... quality is subjective and I hate seeing unused imports and bloat... like this one you have... https://app.codacy.com/app/jackyq2015/OpenRefine/file/15226927129/issues/source?fileBranchId=6695671#l@result.result.startLine

Just basically see that the Codacy review is not so good.. click on Details link...scan through briefly to see where issues are really at that can improve our code quality.  Its a subjective science, and the team pretty much trusts that we at least do a quick visual scan through Codacy review and it only takes usually 1-2 minutes.  At my work, we use Sonarqube and have it highly tweaked and try to maintain an 75% test coverage rate and 80% custom quality rate.

wetneb:
@jackyq2015 @thadguidry thanks for your detailed feedback - I think I have covered it all now. Do you see anything else or should we merge this?

thadguidry:
@wetneb LGTM.  Can you get them to update the API org.wikidata.wdtk.wikibaseapi.ApiConnection ???

```
09:26:18.799 [                   refine] POST /command/wikidata/login (12151ms)
09:26:20.314 [..kibaseapi.ApiConnection] Login Error: Failed (1515ms)
09:26:27.894 [                   refine] POST /command/wikidata/login (7580ms)
09:26:29.206 [..kibaseapi.ApiConnection] Login Error: Failed (1312ms)
09:26:34.532 [                   refine] POST /command/wikidata/login (5326ms)
09:26:35.970 [..kibaseapi.ApiConnection] API warning [main]: Subscribe to the mediawiki-api-announce mailing list at <https://lists.wikimedia.org/mailman/listinfo/mediawiki-api-announce> for notice of API deprecations and breaking changes. Use [[Special:ApiFeatureUsage]] to see usage of deprecated features by your application. (1438ms)
09:26:35.970 [..kibaseapi.ApiConnection] API warning [login]: Main-account login via "action=login" is deprecated and may stop working without warning. To continue login with "action=login", see [[Special:BotPasswords]]. To safely continue using main-account login, see "action=clientlogin". (0ms)
```

tfmorris:
This PR doesn't appear to include any help or documentation for the users. That would both help review it and help the eventual users. Is it located somewhere else?

tfmorris:
There's a whole pile of modules which, without reviewing them in detail, sound like they're generic Wikidata things (e.g. org.openrefine.wikidata.schema, etc.)

Is there a reason that a generic Wikidata library couldn't be used instead? e.g. https://github.com/Wikidata/Wikidata-Toolkit

wetneb:
@thadguidry yes that message is annoying. The good news is that I now have commit rights in this library (Wikidata-Toolkit) so it's something I should be able to fix soon. (A decent chunk of the work for this PR was actually to refactor the datamodel of that library… something I had not really planned to do).

@tfmorris about the documentation: absolutely, there is a lot of work to do on this. I don't think the documentation should be built into the software itself (that makes it harder to update), I was thinking about writing it in OpenRefine's wiki or even directly over at Wikidata.

Concerning Wikidata-Toolkit: yes that's exactly what I am using (and contributing to). The datamodel there cannot be reused directly to model expressions though.

wetneb:
@tfmorris this schema module is the equivalent of the "protograph" or skeleton in Freebase terms: it's an expression that can be evaluated into the Wikibase data model. So you cannot reuse classes for the data model itself: you need to be able to represent variables (columns), and serialize them as such (to store the schema as an overlay model).

That being said, I do think some of the classes introduced in this extension could eventually make it into Wikidata-Toolkit (the maintainer is keen).

wetneb:
@thadguidry I value your feedback very much, I think you always make very good points as a user. As I have said before, I think you should leave the programming side of PR reviews to programmers (and not merge any PR that involves a code change without the approval of someone who contributes to the code). Codacy is nice, but you need to understand what is going on in the code to take advantage of its warnings. The ones you point out are spurious, and some of them are even wrong apparently: for instance, https://app.codacy.com/app/OpenRefine/OpenRefine/file/15383025410/issues/source?bid=6695960&fileBranchId=6695960#l615 gets the scoping wrong: I do want to declare `var endpoint` independently in each branch.

There is a lot of valuable reviewing work that you can do as a user: just pointing out the things that do not work as you expect, or those where you cannot even figure out how to use them (in which case, that's my own fault). Things about the appearance of the UI, the wording of the labels (as you did), and so on. That sort of feedback is absolutely critical - because that is what eventual users care about.

thadguidry:
@wetneb They are not spurious.  You are deliberately confusing the JS runtime engine.  For instance, Why do you need to add "toolbar" in L218 again , my preference would be to rename with "grouptoolbar" ?

Your also deliberately saying you want to make it harder for others coming in by introducing a variable hoisting pattern.  I would rather avoid those.

wetneb:
@thadguidry it looks like I offended you - that was not my intention. I am available to talk any time today.

I am happy to rename that `toolbar` to `grouptoolbar` if that floats your boat, as a token of good will :)

thadguidry:
@wetneb lol, no you'll never offend me.  I'm a very happy go lucky kinda guy.  I'm just asking you to improve the quality a bit, in areas that I know have caused grief for JS programmers in the past.  Runtime errors are the bane of all JS programmers and those 2 anti-patterns that I pointed out cause 90% of runtime errors in JS.

Your correct in thinking that this possibly cannot happen.  I am taking the stance that IT DOES HAPPEN, WHEN WE LEAST EXPECT IT. :)  So let's try to avoid it when we can.

I'm also not taking a hardline here.  You can leave things as is Antonin.  But I just want to help you improve your coding skills in these particular areas, since I've had to chase down runtime errors in the past with UI teams more often than not.

wetneb:
@ostephens @tfmorris @thadguidry I have made a quick tutorial to explain how the extension works. A lot more documentation needs to be written but it should already give a good understanding of the basic workflow: https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing

thadguidry:
@wetneb Thanks for the tutorial.  My thorough testing will come this weekend when I finally have dedicated time.

wetneb:
@tfmorris Detailed docs are now available at https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing#Conclusion if you want to take a look.

ettorerizza:
Thanks for the tutorial, @wetneb ! I was a little worried about testing the extension with fake data, which I could inject by mistake into Wikidata. But now I've a real dataset containing useful information about Belgian municipalities, I will use it for the test.

wetneb:
@ettorerizza awesome, let me know how it goes! If you are not happy about the result it is easy to revert the changes in Wikidata.

wetneb:
And now two more tutorials ready to be tested - don't be shy, give it a try!
* https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing/Tutorials/Inverse_Listeria
* https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing/Tutorials/Working_with_APIs

Also, let's try not publicize this extension too much before the release, so that we get a clean occasion to announce all the new features and acknowledge our funder.

thadguidry:
@wetneb I always test with proper merge master into the PR and then test functionality.  This PR https://github.com/OpenRefine/OpenRefine/pull/1527 got reverted so getting the "not an int" error and cannot test.

wetneb:
@thadguidry sure, that's because #1529 has not been merged - I will do that now.

thadguidry:
@wetneb Is a full column of data sent to determine the Column Type ?
http://127.0.0.1:3333/command/core/guess-types-of-column?project=2242467183748&columnName=recordrow+4&service=https://tools.wmflabs.org/openrefine-wikidata/en/api
This would not be aligned with how Freebase did things.  As we worked through a dataset with Facets... I could filter on a particular subset, Column A, i.e., Country=France in my Facet, and then only rows from my Column B of wine regions in France (selected in Facet) would be sent as JSON, not all of the rows from Column B which included wine regions from other parts of the world, but I'm only focusing on France for now, yet, I get returned other region types...
image

thadguidry:
@wetneb Any idea what's going on with the reference ?

image

wetneb:
@thadguidry hmmm, I'm not sure about your first issue - but as far as I understand it, it is something specific to the way standard reconciliation services are treated (which has not changed for years now). This PR does not change anything in terms of how reconciliation services work (or at least it should not!). Shall we move this to a different issue? (If you can confirm that you observe this problem in master too).

For the reference - it looks like your browser does not like this unicode character. That's very good to know, I will replace it by an actual image.

According to your first screenshot, your schema is incomplete: you have not entered any property in your second statement. The UI should make that clearer indeed, I will change that.

thadguidry:
@wetneb your right, I just checked BLAME and that post config for Recon has not changed.  My memory faded, but I now recall, the expectation was that a column should be constructed for Recon against a single Type.  So ignore that as an issue.

Yeah, if you can make the UI clearer that would be nice.  "Statement is missing property".
To help even more... show "Statement X" on the left side (plenty of space) of each one.  The coloring is not enough.

wetneb:
@thadguidry what do you mean by "Statement X" ?

thadguidry:
Can I screenshare with you for 10 mins , and show few things ?  Google Hangout ?

wetneb:
absolutely!!

thadguidry:
@wetneb hop on to your gmail for Hangouts...just tried calling ya.

thadguidry:
@wetneb I think just adding a trashcan on the empty Property input boxes would work nicely. I noticed that even qualifiers need a way to remove when the input box is empty.  So having a trashcan icon for empty input boxes makes sense.

wetneb:
@thadguidry sure. I am actually thinking about making bigger changes: instead of having a small dialog (where it feels a bit too narrow for all this) I was thinking about making it a new tab in the main UI, so that users could switch from the main view (where they see the cells). There would be three new tabs: one for the schema, one for the issues and one for the statements preview. They would not show up by default (as we do not want to annoy the users who do not use the wikidata interface), but only when first triggered via the extension menu. Also, I will drop the quickstatements preview (which is only useful if you know quickstatements, and people should not have to know about QS to use the interface) and replace it by a preview of the actual statements, with the same look&feel as the schema. I think this is going to make the interface a lot clearer.

thadguidry:
@wetneb Awesome !  Then in that case, your free to use whatever UI javascript and libraries you want or need.  Get Creative !

wetneb:
@thadguidry most of the UI improvements we discussed are implemented (the rest needs changes in the backend which I am going to do later)
image

wetneb:
@ostephens @ettorerizza in general if there is anything that you find unclear in the interface or the documentation, it would massively help if you could report it…  I need to fix it, because nobody else will understand if you don't! We had a very productive hangouts with Thad yesterday, I'm happy to do that one-to-one with others to collect more feedback if you prefer this format.

Otherwise this is what is going to happen at the next release: http://thehipperelement.com/post/93404286580/how-it-people-see-users-using-their-app-for-the

ettorerizza:
@wetneb For my part, my like on your last comment does not mean "hooray! Antonin will change everything!", But rather "it's a good idea in general to avoid these bad editing windows". I tested the PR as soon as you published it, and my first impression reminded me of the RDF extension. That is, something that requires a bit of learning before being used. In addition of the fear of making a mistake by editing Wikidata. In cases like this, my #procrastination takes over. But after your three tutorials, I have no excuse: go! (tomorrow morning at the first hour)

ettorerizza:
@wetneb  Sorry for coming in late. First, congratulations for this tool! 

Note: I used the extension whitout first reading the tutorials (as would any user ...) and I'm not a big editor of Wikidata.

For the moment, everything seems very clear and my only remarks concern details :

- It would be nice if the first drag and drop automatically triggers an "add item" (or if an item is present by default). 

- A "save and close" button in the "Align with Wikipedia" window would be useful.

- And why not a button "quick edit Wikidata", which would apply immediately the schema ?

- The operation "perform edits on Wikidata" seems a bit slow (30 edits/min), but I guess it does not depend on your extension.

- Perhaps it would be usefull to store somewhere the history of the editions made, with a link to their "Edit group" ([https://tools.wmflabs.org/editgroups/b/OR/d9f4bef/](https://tools.wmflabs.org/editgroups/b/OR/d9f4bef/)).

I'm going to continue my tests, and I feel that I will finally be able to thank Wikidata for all the bandwidth I use with my huge reconciliations. :)

wetneb:
Awesome! Thanks a lot for the feedback.
* the drag and drop to the empty schema makes a lot of sense
* save and close would make sense too… maybe we just need a "save" button that closes, and a "cancel" button that leaves without closing (I don't think the "reset" button is really needed)
* the low edit rate is intentional (high edit rates hurt the servers) but it can be adjusted, I will discuss that with the Wikidata dev team
* for now we don't store the edits locally indeed - I am not sure what it would look like?

Finally, I have left a message here: https://www.wikidata.org/wiki/Wikidata:Edit_groups/OR/27b276a - did you get a notification for that? (This sort of message is quite experimental too).

ettorerizza:
Thanks also for your feedback on my poor edits (note : I didn't receive any notification about it). This is precisely what I meant by "storing somewhere a link to the "Edit groups" page : we should have in Open Refine a link to cancel the edits.

I will start again, but reading the tutorials this time.

ettorerizza:
> I am actually thinking about making bigger changes: instead of having a small dialog (where it feels a bit too narrow for all this) I was thinking about making it a new tab in the main UI, so that users could switch from the main view (where they see the cells).

In this case, if as Thad says you can use more modern libraries, an interactive tutorial might be a good idea.

wetneb:
that would make sense - or at least point to a video tutorial, just like the default text we have when there are no facets.

ostephens:
Hi @wetneb - sorry not to be more forthcoming with feedback - I have tried it out a few days ago but hadn't managed to feedback.

I'm probably not a very typical user at this point so finding it hard to approach this as a novice. However a couple of things I managed to do wrong:

1. I added a 'label' to the schema, but didn't populate it - leading me to overwrite the existing labels - I think a warning about empty labels/descriptions/properties is worthwhile? (although I guess we have to allow it if that's what the user wants to do)

2. I had originally filtered my data set to only do one edit - but at some point I removed the filter and then went ahead with the wikidata edit. I think an extra pop-up to warn "You are about to edit XX wikidata items" OK/Cancel might avoid others making the same mistake

wetneb:
@ostephens thanks a lot! both points make a lot of sense and I will address them.

wetneb:
New preview tab (work in progress):
image

jackyq2015:
@wetneb I recommend you do a merge from master for this branch. This branch use the different format of time with the master for the project metadata. This issue was fixed on master and should be merged here otherwise it will cause issue with metadata. At least it is related to issue #1542

extensions/wikidata/module/scripts/dialogs/schema-alignment-dialog.js
```diff
@@ -818,16 +820,22 @@ SchemaAlignmentDialog._initField = function(inputContainer, mode, initialValue,
     });
     if (mode === "amount") {
         input.attr("placeholder", $.i18n._('wikidata-schema')["amount"]);
+        SchemaAlignmentDialog.setupStringInputValidation(input, /^[\-+]?\d+(\.\d*)?(E[\-+]\d+)?$/);
     } else if (mode === "url") {
         input.attr("placeholder", $.i18n._('wikidata-schema')["full-url"]);
+        SchemaAlignmentDialog.setupStringInputValidation(input, /^https?:\/\/.+$/);
```
thadguidry: Is this always https ?  or can it fall back to http sometimes ?

wetneb: that is a regular expression which validates URLs - as you can see there is a `?` after the `s`, which means that `s` can be omitted - so yes, both http and https are accepted in this input.

wetneb:
@jackyq2015 sure, I'll just rebase to keep the history clean

wetneb:
New UI, more tightly integrated with the rest of the UI. The main advantage is that you can use facets when inspecting the edits and issues. And there is spaaaaace to breathe.

To get the statements previews and issues to be updated when the facets change, I have added an extension point in the core JS (rather than monkey-patching that code as GOKb had to do).

screencast-facets

ostephens:
@wetneb sorry to have been quiet - but I've been using the extension and really appreciating it - saving me lots of time! I've not had a chance to try the latest version yet but one thing I wanted to be able to do and couldn't find a way was setting a property to 'no value' or 'unknown value'. Have I missed something, or is this not possible?

Thanks again for this fantastic work so far :)

wetneb:
@ostephens Cool! I haven't seen your edits show up in [EditGroups](https://tools.wmflabs.org/editgroups/), did you push them via QuickStatements maybe?

For novalue and somevalue claims, you are correct that they are not supported yet. It's not clear to me how that should work:
* either we only let users add claims that are always novalue and somevalue in the schema (by replicating Wikidata's interface)
* or we make it possible to generate novalue or somevalue for variables too. I'm reluctant to translate nulls or empty strings to these values because I think users rarely mean that, so we would probably need to define a special string that we never expect to represent an actual value, like `#Wikibase:NoValue` or something ugly along these lines…

ostephens:
Agreed that using null or blank strings is probably a bad idea as we know these are ambiguous! I don't have a better idea than some special value that could be used I'm afraid

ostephens:
Links for the items from the Preview tab go to Freebase rather than Wikidata - highlighted in this screenshot

screen shot 2018-04-13 at 16 14 11

wetneb:
Woot!? For me it does link to Wikidata. That's very puzzling. Do you have more info about in which context that has happened? Is it just for the subject of the item or also the values?

ettorerizza:
Could it be related to #117 ?

ostephens:
Yes I was creating new items.

I think the relevant JSON in the history is:

{
    "op": "core/recon-judge-similar-cells",
    "description": "Mark to create one single new item for all cells containing \"Junta Departamental de Concepción\" in column Name of House or Chamber ES",
    "engineConfig": {
      "mode": "record-based",
      "facets": []
    },
    "columnName": "Name of House or Chamber ES",
    "similarValue": "Junta Departamental de Concepción",
    "judgment": "new",
    "shareNewTopics": true
  },
  {
    "op": "core/recon-judge-similar-cells",
    "description": "Mark to create one single new item for all cells containing \"Junta Departamental de San Pedro\" in column Name of House or Chamber ES",
    "engineConfig": {
      "mode": "record-based",
      "facets": []
    },
    "columnName": "Name of House or Chamber ES",
    "similarValue": "Junta Departamental de San Pedro",
    "judgment": "new",
    "shareNewTopics": true
  },

After the new items were created I also then did notice some oddities in terms of the content of the cell - not linking to the Wikidata item as I'd expect - I ended up re-reconciling the cells in the column to fix it

wetneb:
@ostephens yes, the problem about refreshing column variables should have been fixed by the commits I pushed yesterday. There are other issues with this new UI for the schema (for instance, if you undo some operations that changed the schema, the schema editor is not updated to reflect that… but it's tricky because we might not want to override any unsaved changes there). I am thinking about adding buttons to save the schema (as we had before) and to clear unsaved changes.

ostephens:
@wetneb I think options to save the schema / clear unsaved changes would be good

wetneb:
@ostephens I have added the buttons to save and clear changes (plus a bunch of other UI tweaks).

ostephens:
Just noting https://www.wikidata.org/wiki/Wikidata:Project_chat#Limit_page_creation_and_edit_rate - does/should the extension have rate limiting to avoid falling foul of limits on Wikidata?

wetneb:
@ostephens yes, the library used to edit (Wikidata-Toolkit) has built-in throttling features. The edit rate is 30 edits/sec on the long run (but the first few edits get through more quickly). This can be observed at https://tools.wmflabs.org/editgroups/?tool=OR .

wetneb:
I have added the number of edits made both to the preview tab and the upload dialog as suggested by Owen.

wetneb:

Issue #1410, opened 2018-01-01
Fetch URLs should follow redirections


wetneb:
Apparently, HTTP redirections are not followed by the URL fetcher: they probably should.

ostephens:
I'm already looking at the URL fetching as part of #1217 so I can look at this as well

ettorerizza:
[Related question on the mailing list.](https://groups.google.com/forum/#!topic/openrefine/TKdZPToV3hY)

ostephens:
OK - the issue seems to be that the URLConnection library does not follow redirects when they use different protocols to the original request. In particular this means that if you request an http URI, and there is a redirect to an https URI, this will not be followed. Redirects which are to the same protocol (http->http or https->https) work as expected in OpenRefine.

This behaviour is deliberate, as it stops redirects taking the user from a secure protocol to an unsecured one (https->http). The second answer in this StackOverflow post gives a good explanation of why this is a bad idea https://stackoverflow.com/questions/1884230/urlconnection-doesnt-follow-redirect

What is less clear to me is whether there are any issues in supporting http->https redirects (which will be the far more common scenario) - this feels like it is increasingly common and we could support it without any security concerns. 

Any other views?

wetneb:
I agree there should not be any security concern with HTTPS -> HTTP. By the way, if URLConnection does this sort of nonsense, it might be worth migrating to a more modern library (https://stackoverflow.com/questions/1322335/what-is-the-best-java-library-to-use-for-http-post-get-etc). Ideally one that we already have in our dependencies, and in a dream world something that can be easily mocked for tests.

ostephens:
#748 looks like already implements a work around for data import

PR #1534, opened 2018-03-08
make OpenRefine more robust to misconfigured charset strings,

DavidFichtmueller:
there are a lot of services out there that are not programmed/configured correctly and that deliver character encoding strings with additional spaces or an semicolon at the end. Regular browsers are robust and don't care, but OpenRefine fails in those cases. This change removes additional spaces, quotation marks and semicolon at the beginning/end of the charset string from the Content-Type HTTP header. Calls to external services that are wrongly configured now return the desired content instead of "HTTP error 200 : OK | ".

### DIFF
--- a/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
+++ b/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
@@ -365,6 +365,7 @@ public class ColumnAdditionByFetchingURLsOperation extends EngineDependentOperat
                                 }
                             }
                         }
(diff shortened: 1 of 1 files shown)


thadguidry:
@DavidFichtmueller Thanks for the PR !  So, you have seen some of these incorrect header strings on public endpoints ?  Curious which ones ?

wetneb:
@DavidFichtmueller thanks for the PR! Honestly we are not happy with the way HTTP requests are currently made - there are a number of other issues related to that (like #1410).

I think we should just migrate to a modern HTTP client library that does all these things for us - all these issues should be handled by that sort of library, not OR itself.

Issue #1496, opened 2018-02-16
Add new "Coalesce" function to GREL


ostephens:
From #820 

Add a new GREL function of 'coalesce'. The function should operate as the COALESCE operator in the ANSI SQL-92 standard (https://docs.faircom.com/doc/sqlref/33405.htm). It should take as input one or more expressions or values, and return the first non-null value.

If there are no non-null values supplied, coalesce should return null.

For example:

coalesce(null, "string") -> "string"
coalesce(null, null, "string") -> "string"
coalesce(null, 1) -> 1
coalesce(null, [array]) -> [array]
coalesce(null) -> null

thadguidry:
@ostephens as stated on your pull request... I'm backtracking now and thinking that the last one
```
coalesce(null) -> null
```
might or should perhaps be
```
coalesce(null) -> <blank>
```
This need wider discussion and just feeling around with it I think from the community. I should not be the sole decider on this one.

ostephens:
I'm inclined to keep `coalesce(null) -> null` because that keeps us in line with the SQL function and treating null and blank in a more relaxed way is partly what has led to the need for the coalesce function in the first place.

Very happy to get other input on it @ettorerizza @eximius313

thadguidry:
@ostephens good point.

ettorerizza:
@ostephens I'm trying to understand how to use this function.  If you have a column containing given names and another with the names, the safe concatenation can be done with `coalesce(cells.given_name.value, cells.name.value)`, that's it ?

ostephens:
Hi @ettorerizza all `coalesce` does is take the first non-null value, it doesn't do any concatenation. You can use this to fall back on an alternative string value or empty string when the value in the cell is null. So it's a helper function to avoid trying to concatenate null values, but doesn't reduce it to a single step.

So for example if you try, from the 'name' column the GREL
```
cells["given_name"].value + " " + value.coalesce("")
```
OR
```
cells["given_name"].value + " " + coalesce(value,"")
```
You should see
<img width="400" alt="screen shot 2018-03-15 at 17 44 05" src="https://user-images.githubusercontent.com/576174/37481076-bcf2935a-2878-11e8-873e-9a2360f4c03e.png">

[...]

PR #1536, opened 2018-03-14
Add coalesce function

ostephens:
Adds coalesce function to GREL as specified in #1496 .

### DIFF
--- a/main/src/com/google/refine/expr/functions/Coalesce.java
+++ b/main/src/com/google/refine/expr/functions/Coalesce.java
@@ -1,0 +1,72 @@
+/*
+
+Copyright 2018, Owen Stephens
+All rights reserved.
+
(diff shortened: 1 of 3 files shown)


thadguidry:
should we also support purity of our GREL null string ?
```
coalesce(null)  ->  <blank cell>
```
but then that might be more of an expansion of string handling for GREL, and I think David avoided that previously, but this seems logical if not useful ... since the purpose of this function was to provide an easy GREL way to replace nulls with blanks ? Thoughts ?

ostephens:
Discussion on #1496 highlighted that coalesce should have at least 2 arguments - I will update this PR to make the function require at least 2 arguments

ostephens:
Now updated as per #1496 discussion

Issue #1286, opened 2017-10-23
Add a DISPLAY UNPRINTABLE CHARS toggle mode


thadguidry:
As mentioned in #1268 ... because I'm a purist and because I work with UTF-8 quite a bit and unprintable control characters that are significant to my work...

I want to see in OpenRefine a toggle mode for **DISPLAY UNPRINTABLE CHARS** for our users (like the ACK and CR/LF below)...so that they show up and are highlighted with special coloring/font.  This toggle mode should work perhaps on both datagrids & facets, and perhaps other dialogs or extension mechanisms, yet to be determined.

This is work to be done in the future in part of our [UI Enhancement project(s)](https://github.com/OpenRefine/OpenRefine/projects/2). 

![capture](https://user-images.githubusercontent.com/986438/31906485-cfd28f2c-b7f6-11e7-9195-e9bfff812270.PNG)

thadguidry:
@ostephens @ettorerizza @wetneb Do you guys also like that orange colored example ?  Does it have a good fit and finish for cell styling in OpenRefine ?  Of course we'll work with a Google Designer for some of this later on, but for now I wanted to just show a better example of how the look could be.  Both Jacky and I prefer the coloring and style using graphic symbols, like example 2 shows, rather than the text based control character symbols as shown in example 1 and typically used in other programs such as "SP", "ACK", "CR", "DEL", "TAB", etc. 

VOTE: Agree that example 2 is more towards our liking and preference for the future UI toggle of "display all chars" ?

ettorerizza:
The colored symbols also seem to me both clearer and more aesthetic.

ostephens:
To be contrary, I have to admit I prefer the text based control character symbols. But happy to go with the majority view!

thadguidry:
@ostephens Why do you like them more ?  because they stand out more ?  they are a bit larger ?

![capture](https://user-images.githubusercontent.com/986438/32663183-a6769e5c-c5f2-11e7-9307-401e4a281eaa.PNG)
![capture](https://user-images.githubusercontent.com/986438/32663255-e794dafc-c5f2-11e7-812b-d93c65766751.PNG)

ostephens:
I prefer them both because they are easier to read (larger text) and I feel they are more explicit and easier to learn - I don't have to learn what each of the icons/graphics stand for.

On the latter point - easier to know the difference between SP and NBSP in comparison to learning which one is represented by a filled in square, and which is an empty square.

[...]

PR #1539, opened 2018-03-16
Implement "Facet by null" and "Facet by empty string" and add to customized facets menu

### LINKED ISSUE #1486: Implement "Facet by null" Command and add to Customised Facets menu
From #820 

A "Facet by null" function would create a facet with the values 'true' (the cell value is null) and 'false' (the cell value is not null). It would work similarly to the existing 'Facet by Blank' option, but would only include cells containing a value of 'null', not those containing empty strings

### LINKED ISSUE #1487: Implement "Facet by empty strings" Command and add to Customised Facets menu

joanneong:
Fixes #1486 and fixes #1487 .

Normally I would open a pull request for each issue, but in this case the issues are very similar in nature so I worked on them together. 

In the meantime, do let me know if the main logic and UI changes that I have implemented thus far work for Open Refine.


thadguidry:
@joanneong tested just now... works great !

@jackyq2015 Wondering why my local merge of Joanne's branch causes a temp file missing?

```
C:\Users\THAD\OpenRefine>"C:\Program Files\Java\jdk1.8.0_151\bin\java.exe" -cp "server\classes;server\lib\*"   -Xms1400M -Xmx1400M -Drefine.memory=1400M -Drefine.max_form_content_size=1048576 -Drefine.port=3333 -Drefine.host=127.0.0.1 -Drefine.webapp=main\webapp -Djava.library.path=server\lib/native/windows com.google.refine.Refine
10:53:58.968 [            refine_server] Starting Server bound to '127.0.0.1:3333' (0ms)
10:53:58.984 [            refine_server] refine.memory size: 1400M JVM Max heap: 1407188992 (16ms)
10:53:58.984 [            refine_server] Initializing context: '/' from 'C:\Users\THAD\OpenRefine\main\webapp' (0ms)
10:53:59.285 [            refine_server] Failed to use jdatapath to detect user data path: resorting to environment variables (301ms)
10:53:59.285 [            refine_server] Failed to use jdatapath to detect user data path: resorting to environment variables (0ms)
10:53:59.285 [                   refine] Starting OpenRefine 2.8 [TRUNK]... (0ms)
10:53:59.285 [                   refine] initializing FileProjectManager with dir (0ms)
10:53:59.285 [                   refine] C:\Users\THAD\AppData\Roaming\OpenRefine (0ms)
10:53:59.379 [         project_metadata] java.io.FileNotFoundException: File 'C:\Users\THAD\AppData\Roaming\OpenRefine\1825969120998.project\metadata.temp.json' does not exist
        at org.apache.commons.io.FileUtils.openInputStream(FileUtils.java:137)
        at com.google.refine.model.medadata.ProjectMetadata.loadFromFile(ProjectMetadata.java:551)
        at com.google.refine.io.ProjectMetadataUtilities.loadFromFile(ProjectMetadataUtilities.java:153)
        at com.google.refine.io.ProjectMetadataUtilities.load(ProjectMetadataUtilities.java:98)
        at com.google.refine.io.FileProjectManager.loadFromFile(FileProjectManager.java:396)
        at com.google.refine.io.FileProjectManager.load(FileProjectManager.java:361)
        at com.google.refine.io.FileProjectManager.<init>(FileProjectManager.java:94)
        at com.google.refine.io.FileProjectManager.initialize(FileProjectManager.java:81)
        at com.google.refine.RefineServlet.init(RefineServlet.java:130)
        at javax.servlet.GenericServlet.init(GenericServlet.java:241)
        at edu.mit.simile.butterfly.Butterfly.init(Butterfly.java:180)
        at org.mortbay.jetty.servlet.ServletHolder.initServlet(ServletHolder.java:440)
        at org.mortbay.jetty.servlet.ServletHolder.doStart(ServletHolder.java:263)
        at com.google.refine.RefineServer.configure(Refine.java:296)
        at com.google.refine.RefineServer.init(Refine.java:208)
        at com.google.refine.Refine.init(Refine.java:114)
        at com.google.refine.Refine.main(Refine.java:108)
 (94ms)
10:53:59.394 [         project_metadata] java.io.FileNotFoundException: File 'C:\Users\THAD\AppData\Roaming\OpenRefine\1791016523934.project\metadata.temp.json' does not exist
        at org.apache.commons.io.FileUtils.openInputStream(FileUtils.java:137)
        at com.google.refine.model.medadata.ProjectMetadata.loadFromFile(ProjectMetadata.java:551)
        at com.google.refine.io.ProjectMetadataUtilities.loadFromFile(ProjectMetadataUtilities.java:153)
        at com.google.refine.io.ProjectMetadataUtilities.load(ProjectMetadataUtilities.java:98)
        at com.google.refine.io.FileProjectManager.loadFromFile(FileProjectManager.java:396)
        at com.google.refine.io.FileProjectManager.load(FileProjectManager.java:361)
        at com.google.refine.io.FileProjectManager.<init>(FileProjectManager.java:94)
        at com.google.refine.io.FileProjectManager.initialize(FileProjectManager.java:81)
        at com.google.refine.RefineServlet.init(RefineServlet.java:130)
        at javax.servlet.GenericServlet.init(GenericServlet.java:241)
        at edu.mit.simile.butterfly.Butterfly.init(Butterfly.java:180)
        at org.mortbay.jetty.servlet.ServletHolder.initServlet(ServletHolder.java:440)
        at org.mortbay.jetty.servlet.ServletHolder.doStart(ServletHolder.java:263)
        at com.google.refine.RefineServer.configure(Refine.java:296)
        at com.google.refine.RefineServer.init(Refine.java:208)
        at com.google.refine.Refine.init(Refine.java:114)
        at com.google.refine.Refine.main(Refine.java:108)
 (15ms)
10:53:59.879 [       database-extension] Initializing OpenRefine Database... (485ms)
10:53:59.879 [       database-extension] Database Extension Mount point /extension/database/ [*] (0ms)
```

ettorerizza:
> Remove "Facet by blank" - related code to clean up the codebase

Maybe I did not understand the changes in progress. I thought it was adding a facet by null and by empty string, but keeping "facet by blank" (null *and* empty). 

This function is probably the most used. Lambda users need to know every ten minutes which lines contain a blank, either the null value or the empty string. This is why I regret that it is so difficult to find it in the drop-down menu tree.

joanneong:
A few hours ago I thought I received some email notification on keeping the "Facet by blank" command. Can I just check whether this facet should or should not be kept? Personally I think having all three facets (blank, null, empty string) can be confusing for new users, but I'm not sure about the perspective of seasoned users so...

wetneb:
It's a good question, I am not so sure.

We also need to keep in mind that the current location of the "Facet by blank" button is hidden deep down the column menu. I would be in favor of putting the new facets somewhere more accessible (in the main Facet menu).

Maybe one solution would be to put the new facets in the main Facet menu, and keep the "Facet by blank" burried where it currently is? This way, seasoned users will still be able to dig for that, and new users will first discover the new versions, without being immediately confused by the "Facet by blank" (because you really need to be looking for it to find it…)

ettorerizza:
Hi @joanneong , Yes, I wrote:

> "Maybe I did not understand the changes in progress. I thought it was adding a facet by null and by empty string, but keeping facet by blank (null and empty). 
> 
> This function is probably the most used. Users need to know which lines contain a blank, either the null value or the empty string. This is why I regret that it is so difficult to find in the tree drop-down menu."

But seeing your code and Thad's comment here (https://github.com/OpenRefine/OpenRefine/issues/820#issuecomment-325765811), I thought I must have misunderstood, so I deleted my comment.

From my point of view, it's the null-empty string distinction that brings confusion. That's why software like Rstudio offers the possibility to treat nulls and "" as NA when importing. But we must not forget that in Open Refine, the blanks have the particular use of allowing to create records. Finding empty cells quickly is therefore even more important than in R. Without Facet by blank, we need two facets to identify empty lines!

ostephens:
Definitely in favour of keeping Facet by Blank. For backwards compatibility (applying historical transformations that use a ‘blank’ facet as part of criteria) and also in many situations users won’t care about the difference between empty string and null because they just want to identify missing data etc.

I would keep the options together where they currently are for this PR - moving things around in the menus means lots if past documentation becomes inaccurate or needs updating- we should only do this where we are v clear on the benefits. 

I’m not against re-thinking how these menu options are arranged but I’d prefer to see an overall approach as part of a major release

wetneb:
@ostephens sure… but when introducing new buttons, wouldn't it make sense to put them at a more convenient place directly? That would not move any existing option around and I think it would help reduce the confusion. But it's just my intuition.

ostephens:
Ah @wetneb I misread your comment - sorry. Agree that the new options could go elsewhere in the menu without breaking anything. 

Personally I’m in favour of keeping the three options together as I think this is easier for instruction etc and also gives context when explaining the function (blank == empty OR null).

joanneong:
Right now the simple solution is to put them like this:

image

But I think what everyone has said so far makes sense because this is indeed quite nested. Would shifting the blank, null, and empty string facets to a section in the same level as the scatterplot facet, but below that section work?

ettorerizza:
> Maybe one solution would be to put the new facets in the main Facet menu, and keep the "Facet by blank" burried where it currently is? 

@wetneb I thought precisely the opposite: put facet by blank in the foreground, since it is the most generic function, and hide the other two (since being able to distinguish nulls vs empty string is something more specialized).

I mean, the Open Refine file is often exported to CSV, where the null-empty string distinction does not make sense.

ostephens:
I’m in favour of the arrangement in the screenshot posted by @joanneong

I’m against putting the new options higher in the menu hierarchy thank ‘by blank’ for the same reasons as @ettorerizza

ettorerizza:
@joanneong @ostephens If everyone agrees on this version, can I just suggest leaving "facet by blank" at the bottom, so seasoned users aren't disoriented ?

screenshot 12h 57m 37s

ostephens:
Can we use “null or empty” rather than “null & empty”?

ettorerizza:
Problem a little different, but related: the creation of records is not well documented on the Wiki. One might think that any empty cell below a non-empty cell in the first column belongs to the same record, whereas it works only for null cells.

It's confusing, even for seasoned users. Will the current work on displaying invisible characters allow to distinguish visually nulls and empty strings?

screencast

ostephens:
@ettorerizza - suggest we need a separate issue to display null in cells - could you add one? I see this as distinct from #1286 which is about the display of invisible characters

jackyq2015:
@thadguidry  It happens some time when the metadata file get corrupted. It should nothing to do with this PR though. You can file an issue if you still can see this exception.

thadguidry:
@joanneong Yes, Joanne, that last design is intuitive enough.  Just change to "or" instead of "&" as @ostephens suggests.  Great job again !

joanneong:
Update: 

The English UI now looks like this:

screen shot 2018-03-17 at 10 32 33 pm

I have also updated the other translations using a mix of Google Translate and my own judgement as per whatever is already available.

I noticed a minor UI problem prevalent across several languages. Here is an example:

screen shot 2018-03-17 at 10 23 28 pm

Basically, words protrude out of the menu because they are too long.

But the most severe issues come in for the Deutsch language, with the UI looking like this:

screen shot 2018-03-17 at 10 26 25 pm

Since I am aware that [this is a work in progress](https://github.com/OpenRefine/OpenRefine/pull/1285), I will not touch it. Just raising a flag again anyway as a reminder :P 

I will be updating the tests for these facets soon.

ettorerizza:
Yep, with just 13.2 % of translated terms, it is hard to say that there is a German version...

https://hosted.weblate.org/projects/openrefine/translations/

main/webapp/modules/core/langs/translation-zh.json
```diff
@@ -490,7 +490,9 @@
         "log-length-facet": "文本长度的对数值归类",
         "unicode-facet": "Unicode字符归类",
         "facet-error": "按错误归类",
-        "facet-blank": "按空白归类",
+        "facet-blank": "按空白归类 (空值或空字符串)",
+        "facet-null": "按空值归类",
```
jackyq2015: I think maybe keep the null as-is is better to understand

joanneong: Can I double check what you mean by this?

jackyq2015: I mean from the translation aspect, it is not so clear to me what's the difference between those 3 facets.  按空值归类 is a little bit confusing even for the programmer, not mention the non-programmer end user. keeping null untranslated(as is) may not be best option, but it is better than confusion.

wetneb: The language fallback does not work so if you leave this string untranslated the button will show as blank in the UI.

joanneong: @jackyq2015 I can understand Chinese so this did not occur to me. @wetneb What do you make of @jackyq2015 's suggestion to use "按null归类" instead? That is fine with me if it helps users.

jackyq2015: @wetneb Not the whole string not get translated, just part of it(null). 
@joanneong I can tell you can understand Chinese from your name. I am from China but from the translation itself alone, it does not get cross the meaning of "facet by null". In a other word, there is no good Chinese translation for null. Instead of trying to translate it, I would rather keep the null as is. Like the "Unicode" new lines above

ostephens: The same thought had occurred to me with some other translations of `null` as it has quite a specific meaning in this context. I can't (sadly) comment on the specific of the translation, but I think it is OK (and sensible) to the term with specific meaning where there is no direct translation

ettorerizza: The Chinese version of Wikipedia does not try to translate it: https://zh.wikipedia.org/wiki/Null_(%E7%B7%A8%E7%A8%8B). It's interesting, I did not know that so many technical words were not translated at all.

joanneong: Ah I see! One more question, do you think we should just keep the word "null" across all the different languages, for consistency and in case such a scenario applies to other languages as well?

ettorerizza: On Wikipedia, it looks like null is never translated when it means `null` (and not nothing). It's a kind of constant, a bit like Pi.

ostephens: When I looked at this for #1500 I found that most translations used `null` but Spanish and Brazillian Portuguese translated it as "nulo". I copied this pattern when I updated the translation files for that PR on the basis that the people who did that translation were in a better position to make that call than me :)

I've been doing some looking round this morning and found the Microsoft terminology search on their language portal https://www.microsoft.com/en-us/language?sString=lock&langID=fr-fr. Based on this it seems that Microsoft translates in some languages and not others (and for Brazillian Portuguese it agrees with the use of `nulo`, while for Chinese it agrees with using `null`)

I suspect we could do worse than follow Microsoft practice here.

Note that the use of `null` in the transformation window as illustrated by @ettorerizza could equally be because this comes from the OR backend where we don't yet support any translations.

ostephens: But I also don't want to prolong this discussion - I'm happy with the decision either way - lets get this PR done! :)

jackyq2015: I can only speak for Chinese language. For this particular term, I believe keep it as-is is better. There might be good fit for null in other languages. But the principle we should follow is that if the translation cannot get across  the real meaning and we should NOT try to translate it to prevent further confusion.

joanneong:
My apologies for the delayed update. I was looking into the appropriate location to add tests for these facets, and so far I have found `facets.js` which seems to be testing a very small subset of the facets using some predefined data. I also realised that testing seems to be done using [Windmill](https://github.com/windmill/windmill/wiki/JavaScript-Tests) even though the page says that it is no longer being actively maintained and that Selenium should be used instead. 

In light of this, should I (a) write tests following the format of the old tests using Windmill, (b) write tests using Selenium (but I am not sure if OpenRefine has taken up Selenium yet), or (c) not touch tests for the time being. What do you think?

jackyq2015:
@joanneong That is a good question. Windmill was unused for quite long time. If you can take a look to see how much effort requires to recovery it, that will be good. That's the only automation testing tool we had used. It will be useful when we do the front end and back end split. But if it's too much to do, we may consider drop it for now. There are many good testing framework now a days can do better.

wetneb:
@joanneong For now the UI is completely untested indeed. I would not go down the Windmill route given that is completely abandoned. If you want to try to set up Selenium that would be fantastic - but it will probably take some effort. Sauce labs can be used in conjunction with Travis to run browser tests in builds. But I think this PR can be accepted without UI tests.

If you are keen to work on testing, the backend itself is mostly untested (the test coverage is really poor and tons of bugs routinely go through) so that might be an easier way to improve the tests.

joanneong:
@wetneb Got it. Give me some time to look into testing for Open Refine, and I will open a new issue to initiate some further discussion when I am ready :P

PR #1540, opened 2018-03-17
Update from Weblate.

weblate:


wetneb:
@tdombos thanks a lot for this translation! keep up the good work!

Issue #1541, opened 2018-03-17
Visually distinguish null and empty strings


ettorerizza:
The ongoing discussions on the `null ` vs. empty string `""` distinction (#820), and the new facets related to it (#1539), reminded me that the creation of records is not well documented on the Wiki. One might think that any empty cell below a non-empty cell in the first column belongs to the same record, whereas it works only for `null ` cells.

![screencast](https://user-images.githubusercontent.com/6286410/37555909-533c568a-29ef-11e8-92b8-5f62b0aecef7.gif)

It's confusing, even for seasoned users. It would be necessary to visually distinguish the two, a bit like this.

![screenshot 14h 28m 51s](https://user-images.githubusercontent.com/6286410/37555936-ace28916-29ef-11e8-892c-0709991543a1.jpg)

According to Owen, this is a separate issue from #1286, which concerns only non-printable chars.

EDIT : Looks like I'm wrong, it's possible to produce a record with empty strings. The problem is in the manual edit, not in the difference between null and empty string.

![screenshot-localhost-3333-2018 03 17-18-04-11](https://user-images.githubusercontent.com/6286410/37558076-b535d5d6-2a0e-11e8-97b5-62dce4ca6596.png)

ostephens:
This is what I'm proposing: `null` will display (as it does in the Expression preview window). Empty string - by definition there is nothing to display. I suggest we just leave this blank.

Note that we will still need a resolution to differentiate between an empty string and a string of only invisible/non-printing characters

<img width="503" alt="screen shot 2018-03-17 at 16 43 02" src="https://user-images.githubusercontent.com/576174/37557837-862f591c-2a02-11e8-84d3-a06e108c1adc.png">

Row 2 = empty string
Row 3 = null

Adding this display of null is trivial and I can do a PR now if @ettorerizza is happy with this approach

ettorerizza:
@ostephens This looks like a small improvement, but it helps to understand why your example certainly produces three records rather than one. In addition, users will better understand the utility of `facet by null ` and `facet by empty strings` (which may seem very abstract).

ettorerizza:
The more I think about it and the more I think that the simplest solution would be to not authorize empty strings, to automatically turn it into `null`... What would be the drawbacks ?

ostephens:
[...]

PR #1544, opened 2018-03-17
Added display of `null` in data table view

ostephens:
As per #1541

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
@@ -60,7 +60,7 @@ DataTableCellUI.prototype._render = function() {
   .mouseleave(function() { editLink.css("visibility", "hidden"); });
 
   if (!cell || ("v" in cell && cell.v === null)) {
-    $('<span>').html("&nbsp;").appendTo(divContent);
+    $('<span>').addClass("data-table-null").html('null').appendTo(divContent);
   } else if ("e" in cell) {
     $('<span>').addClass("data-table-error").text(cell.e).appendTo(divContent);
   } else if (!("r" in cell) || !cell.r) {
--- a/main/webapp/modules/core/styles/views/data-table-view.less
+++ b/main/webapp/modules/core/styles/views/data-table-view.less
@@ -201,6 +201,10 @@ div.data-table-cell-content-numeric > a.data-table-cell-edit {
   color: #282;
   }
 
+.data-table-null {
+  color: #aaa;
+  }
+
 .data-table-error {
   color: red;
   }


ettorerizza:
The result looks very good! Perhaps we should now use a slightly bolder font for non-zero values, but it's a detail.

screenshot-127 0 0 1-3333-2018 03 17-18-21-34

ostephens:
@jackyq2015 styling invisible characters is separate issue #1286. null is not an invisible character in that sense - it is a cell value not a character

The character issue is more complex and better to keep separate IMO

ostephens:
V happy to look #1546 but prefer to keep tgat separate to this issue & PR

wetneb:
argh - no option to deactivate this? are we 100% sure about that?

ettorerizza:
Rather than an option to disable the display of `null`, why not an option to use empty strings instead of `null ` (basically, a shortcut to Transform all -> `if(isNull(value), "", value)` ? I feel that it is good practice that the users always know when they're working with columns containing null -  that cannot be concatenated directly to one another, unlike columns containing empty strings.

thadguidry:
@ostephens whoa, wait a sec... this was supposed to be an option, as @wetneb said.  We don't want to make this default...its visually too busy for regular users.  Put more work into this and make it an option in our preferences.vt

ostephens:
OK - Apologies for jumping the gun @thadguidry @wetneb. Is the best approach to revert this PR merge and then do some further work for a new PR?

wetneb:
@ostephens I don't think you need to revert this, I think a toggle could just be built on top of that with little effort. It can be quite minimal: just a button that, when clicked, toggles a class on a div that contains the grid, and a CSS rule that puts the nulls in `display: none` when the parent div has that class.

I would be in favour of making this feature opt-in for the coming release (so, disabled by default). Based on user feedback we could consider enabling it by default in the following release.

Basically I'm all in for a better treatment of nulls and empty strings. This is actually something that would be highly beneficial to the Wikidata extension. But I fear that some users might be used to the way things currently are - we don't want to "brutalize" them by forcing on them this distinction that they not might need in their own workflows.

@ettorerizza I'd be careful with such a transformation - it would break the records mode pretty badly, no?

ostephens:
OK - no problem. Will have a look at it hopefully this weekend

ettorerizza:
> I'd be careful with such a transformation - it would break the records mode pretty badly, no?

@wetneb That's what I thought too, but after a few tests, it seems that records can be created with any blanks, nulls or empty strings. 

screencast

In fact, it's not clear to me how exactly records are created and how they can be broken. I mean, I know the big picture of course, but some details are confusing. And there is no documentation on the Wiki. The only mention of the row-records difference I found is on a [Martin's blog post](http://kb.refinepro.com/2012/03/difference-between-record-and-row.html) (@magdmartin : do you think it could be a good idea to copy-paste it on the Wiki?). We will probably have to dive into the source code to flesh out the doc.

jackyq2015:
I took a look of record Modle when I fixed one bug. Here is a simple one:
when column can group together, we have the keyedGroup. Based on the keyedGroup, rowDependency is generated
by searching  the last non-blank cell at following rows. The record model is build upon the rowDependency.
In another word, if there is no rowDependency, record model is meaningless. That is exactly the case when 
importing a FLAT csv file. record model is more suitable for a tree based data structure.

wetneb:
@ettorerizza I agree that the documentation of the records mode is very poor (which is sad because it's quite a fundamental feature) - thankfully there's @magdmartin post but that would clearly deserve to be treated in the wiki. I agree with your analysis of the code - I thought I had read in the past that it relied specifically on nulls but that's apparently wrong. I think that if empty and nulls are conflated in such a crucial part of the code, it's yet another reason not to precipitate displaying the distinction to users…

ettorerizza:
@jackyq2015 The part I don't understand and that makes me feel dumb is : how can the `ExpressionUtils.isNonBlankData() ` method be used to identify both nulls and empty strings (eg in the GREL function [`isBlank()`](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/grel/controls/IsBlank.java)) while its definition seems to exclude empty strings? (isNonBlankData = non null and not an error and (not a string but possibly a string greater than 0)

PR #1432, opened 2018-01-15
add find function. issue 1296

jackyq2015:

PR #1558, opened 2018-04-05
Contains function supports now regex as second argument 

### LINKED ISSUE #1554: Enhancement : a contains() that support regex
Issue #1550 shows that a [`contains()`](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/expr/functions/strings/Contains.java) function that accepts Regex might be useful. Now that @jackyq2015 has coded the [`find()`](https://github.com/OpenRefine/OpenRefine/blob/9ab897cb72111de58956721e46fb28f26f121f3b/main/src/com/google/refine/expr/functions/strings/Find.java) function, looks like even a complete beginner would be able to draw on his code to enrich `contains()`. Could it make a "good first issue" ?

gobertm:
Closes #1554 

Implementation based on Partition function.

### DIFF
--- a/main/src/com/google/refine/expr/functions/strings/Contains.java
+++ b/main/src/com/google/refine/expr/functions/strings/Contains.java
@@ -34,6 +34,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package com.google.refine.expr.functions.strings;
 
 import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.json.JSONException;
 import org.json.JSONWriter;
@@ -49,6 +51,14 @@ public class Contains implements Function {
             Object s2 = args[1];
             if (s1 != null && s2 != null && s1 instanceof String && s2 instanceof String) {
                 return ((String) s1).indexOf((String) s2) > -1;
+            } else if (s1 != null && s2 != null && s1 instanceof String && s2 instanceof Pattern){
+                String s = (String) s1;
+                Pattern pattern = (Pattern) s2;
+                Matcher matcher = pattern.matcher(s);
+                if (matcher.find()) {
+                    return true;
+                }else
+                    return false;
             }
         }
         return null;
--- a/main/tests/server/src/com/google/refine/tests/expr/functions/ContainsFunctionTests.java
+++ b/main/tests/server/src/com/google/refine/tests/expr/functions/ContainsFunctionTests.java
@@ -1,0 +1,75 @@
+
+package com.google.refine.tests.expr.functions;
+
+import com.google.refine.RefineServlet;
+import com.google.refine.grel.ControlFunctionRegistry;
+import com.google.refine.grel.Function;
+import com.google.refine.tests.RefineServletStub;
+import com.google.refine.tests.RefineTest;
+import org.slf4j.LoggerFactory;
+import org.testng.Assert;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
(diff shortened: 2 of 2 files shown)


ettorerizza:
Looks like special characters such as `. ^ $` are always matched, even in string mode. 

screenshot-127 0 0 1-3333-2018 04 05-12-23-43

Could you try with : 

```
value.contains("[a]")
value.contains("$")
value.contains(".")
value.contains("^")
```

main/src/com/google/refine/expr/functions/strings/Contains.java
```diff
@@ -47,8 +49,14 @@ public Object call(Properties bindings, Object[] args) {
         if (args.length == 2) {
             Object s1 = args[0];
             Object s2 = args[1];
-            if (s1 != null && s2 != null && s1 instanceof String && s2 instanceof String) {
-                return ((String) s1).indexOf((String) s2) > -1;
+            if (s1 != null && s2 != null && s1 instanceof String && (s2 instanceof String || s2 instanceof Pattern)) {
```
ettorerizza: Do a better distinction between String and Pattern, as in the [replace() method](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/expr/functions/strings/Replace.java) ?

gobertm:
@ettorerizza It depends what is considered string mode and regexp mode. 
In OpenRefine do you accept regular expression inside double quotes or regular expression must only be in / /  ? 
Because it seems that currently the behavior is not consistent throughout all functions : 
- `find`  (as you also mentioned [here](https://github.com/OpenRefine/OpenRefine/pull/1432#issuecomment-378906943 ). 
but also `match`
 **interprets regular expression in double quotes**. 

image

- `replace` and `partition`  **only consider regexp when in ` / / `**

My original commit of `contains` this morning worked like replace and partition. That's why I couldn't write tests easily (regexp can not be passed as String).

I guess you have to choose between the two interpretation.

gobertm:
@thadguidry 
I just checked all [String functions](https://github.com/OpenRefine/OpenRefine/wiki/GREL-String-Functions) mentionning regular expression.
Those that accepts regex in `" "` are : find and match . Others are ok.

@ettorerizza They indeed act like that currently. If I only keep the initial commit of this PR the contains function works like partition.

ettorerizza:
`Those that accepts regex in " " are : find and match . Others are ok.`

Good news, match() can only be used with regexp anyway. Just leave it like that, it will avoid retrocompatibility problems.

List of functions that use regexp:

```
find()
contains()
match()

partition()
rpartition()
split()
replace()
```

ettorerizza:
@gobertm I'm really not a specialist in Git or Github, but I think it's better to do PRs from a new branch of your master named eg `issue/<issueenumber>`.

ettorerizza:
I think it's okay now.

```
value = "rose is a rose"

value.contains("\s+")--> false
value.contains("\\s+") -->  false
value.contains(/\s+/)--> true
value.contains("$")  --> false
value.contains(/$/)  --> true
value.contains("rose") --> true
value.contains("r.se") --> false

```

gobertm:
Hi,
Is there something else to do in order to have this one completed?

Cheers,
Maxime

jackyq2015:
@gobertm can you please add a unit test?

gobertm:
Hi,
I added the tests based on @ettorerizza tests. 

But sorry I'm messing with rebase and so on. I try to clean this up. :/

gobertm:
Hi,
I added tests based on @ettorerizza examples.
String input , regular expression pattern injected as string are interpreted as string and for the tests of regular expression I tested by directly innjecting a Pattern.

PR #1563, opened 2018-04-09
migrate from JRDF to JENA library

jackyq2015:
To fix the #1560


jackyq2015:
@thadguidry According to [this](https://jena.apache.org/documentation/io/rdf-input.html):
Much of the functionality is accessed via the Jena Model API; direct calling of the RIOT subsystem isn't needed. 

I tried to remove few of jars and all lead to exception when loading

thadguidry:
@jackyq2015 is this an exposed option or not ?

To aid in checking for errors in UTF8-encoded files, there is a utility which reads a file of bytes as UTF8 and checks the encoding.

    utf8 -- read bytes as UTF8

jackyq2015:
@thadguidry I cannot see the RIOT is a exposed option.

For the utf8 part, how is related to this change?

thadguidry:
@jackyq2015 forget it.  we can deal with that option later.

PR #1568, opened 2018-04-14
Fix identifier space for cells reconciled to New

wetneb:
There is a bug in the operation that marks cells as reconciled to new (discovered by @ostephens): when sharing topics, the identifier space used for the reconciliation is ignored and the default one is used instead. This fixes it (by using the identifier space associated with the column).


wetneb:

Issue #1564, opened 2018-04-10
Date parsing fails on formats that were previously supported


wetneb:
When reporting a bug please provide the following information to help reproduce the bug:

#### Steps followed to create the issue:

open a project created with an earlier version of OpenRefine

#### Current Results:

```
11:42:35.336 [..ject_metadata_utilities] load metadata failed: /home/antonin/.local/share/openrefine/2329149621165.project/metadata.old.json (1ms)
11:42:35.336 [..ject_metadata_utilities] java.time.format.DateTimeParseException: Text '2018-03-21T11:49:28.992' could not be parsed at index 23
	at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1949)
	at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851)
	at java.time.Instant.parse(Instant.java:395)
	at com.google.refine.util.ParsingUtilities.stringToLocalDate(ParsingUtilities.java:196)
	at com.google.refine.model.medadata.ProjectMetadata.extractModifiedLocalTime(ProjectMetadata.java:296)
	at com.google.refine.model.medadata.ProjectMetadata.loadFromJSON(ProjectMetadata.java:216)
	at com.google.refine.model.medadata.ProjectMetadata.loadFromStream(ProjectMetadata.java:583)
	at com.google.refine.model.medadata.ProjectMetadata.loadFromFile(ProjectMetadata.java:555)
	at com.google.refine.io.ProjectMetadataUtilities.loadFromFile(ProjectMetadataUtilities.java:162)
	at com.google.refine.io.ProjectMetadataUtilities.loadMetaDataIfExist(ProjectMetadataUtilities.java:109)
	at com.google.refine.io.ProjectMetadataUtilities.load(ProjectMetadataUtilities.java:98)
	at com.google.refine.io.FileProjectManager.loadProjectMetadata(FileProjectManager.java:122)
	at com.google.refine.io.FileProjectManager.recover(FileProjectManager.java:452)
	at com.google.refine.io.FileProjectManager.initialize(FileProjectManager.java:83)
	at com.google.refine.RefineServlet.init(RefineServlet.java:130)
	at javax.servlet.GenericServlet.init(GenericServlet.java:241)
	at edu.mit.simile.butterfly.Butterfly.init(Butterfly.java:180)
	at org.mortbay.jetty.servlet.ServletHolder.initServlet(ServletHolder.java:440)
[...]

PR #1570, opened 2018-04-19
parse the date time string with offset

jackyq2015:
To fix issue #1564

### DIFF
--- a/main/src/com/google/refine/expr/functions/ToDate.java
+++ b/main/src/com/google/refine/expr/functions/ToDate.java
@@ -85,7 +85,7 @@ public class ToDate implements Function {
             try {
                 return CalendarParser.parse( o1, (month_first) ? CalendarParser.MM_DD_YY : CalendarParser.DD_MM_YY);
             } catch (CalendarParserException e) {
-                OffsetDateTime d = ParsingUtilities.stringToDate(o1);
+                OffsetDateTime d = OffsetDateTime.parse(o1);
                 if (d != null) {
                     return d;
                 } else {
--- a/main/src/com/google/refine/expr/functions/ToDate.java
+++ b/main/src/com/google/refine/expr/functions/ToDate.java
@@ -51,7 +51,6 @@ import com.google.refine.expr.util.CalendarParser;
 import com.google.refine.expr.util.CalendarParserException;
 import com.google.refine.grel.Function;
 import com.google.refine.grel.ControlFunctionRegistry;
-import com.google.refine.util.ParsingUtilities;
 
 public class ToDate implements Function {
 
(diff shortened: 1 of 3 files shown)


jackyq2015:
Though both return OffsetDateTime, the ParsingUtilities.stringToDate always return the Zulu time, but the toDate will keep the offset as is.

wetneb:
That does not solve the bug I reported. The behaviour of stringToDate was changed and this breaks the data extension operation, among others. Please make sure that stringToDate supports all the formats it supported before.

wetneb:
Hi @jackyq2015, have you seen my report in #1564? Just reconcile a column (for instance containing the names of famous people) and fetch dates from that (for instance their date of birth, P569) and you will observe the error.

So, here is the situation. In commit 23ec54b78 you have changed the behaviour of `ParsingUtilities.stringToDate`. Many parts of the code rely on this function to parse dates. Therefore, this potentially introduces bugs in all these areas. I have shown you two instances of such bugs, but of course we should not fix just them: they are just the symptoms and we should fix the root cause. The root cause is that `ParsingUtilities.stringToDate` previously supported ISO dates with offsets, and now it fails on them. So, it would be great if you could fix that.

If it helps, here is a test case that should pass (it passes before 23ec54b78 and breaks after):
```
    @Test
    public void parseOffsetDate() {
        String offsetDate = "1958-08-29T00:00:00+00:00";
        OffsetDateTime parsed = ParsingUtilities.stringToDate(offsetDate);
        OffsetDateTime expected = OffsetDateTime.parse(offsetDate);
        Assert.assertEquals(expected, parsed); 
    }
```

If anything is still unclear, do let me know (I am happy to discuss that directly over hangouts for instance). If you want me to take this over, I can also do that.

jackyq2015:
@wetneb I cannot reproduce this. I reconciled name "George Washington" and use "Add column from reconciled data"  to get the "date of birth", I can get the 1732-02-22T00:00:00Z without issue. And I can reload it after I bounce the OR. You need have the master code and a refresh project in order to work. If you can get a detail step to reproduce the issue I will give another try. 

Please see my comment [here](https://github.com/OpenRefine/OpenRefine/issues/1564#issuecomment-381005850). The commit 23ec54b is actual to restore the original behaviour, not to change it.

But still you can go ahead to improve it whatever you think it's appropriate.  Maybe I am mistaken

wetneb:
Hmm, it's curious that you cannot reproduce the issue… Which JDK are you using? I can observe the issue with OpenJDK8.

I'll go ahead and propose another version that should hopefully work for everyone.

jackyq2015:
$ java -version
openjdk version "1.8.0_162"
OpenJDK Runtime Environment (build 1.8.0_162-8u162-b12-0ubuntu0.17.10.2-b12)
OpenJDK 64-Bit Server VM (build 25.162-b12, mixed mode)

PR #1571, opened 2018-04-19
Added option to toggle show/hide null values in cells in data-table

ostephens:
As discussed in #1544 this PR adds a toggle option for displaying 'null' in cells containing a null value (as opposed to an empty string). Adds option under the All->View menu. Option is persistent while viewing a project (unless the user forces a refresh of the screen via the browser).

Default is to not display the null values to keep the screen from becoming too busy

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
+++ b/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
@@ -39,6 +39,7 @@ function DataTableView(div) {
   this._collapsedColumnNames = {};
   this._sorting = { criteria: [] };
   this._columnHeaderUIs = [];
+  this._shownulls = false;
 
   this._showRows(0);
 }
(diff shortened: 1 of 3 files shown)


main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -741,6 +745,13 @@ DataTableView.prototype._createMenuForAllColumns = function(elmt) {
...
             self.render();
           }
+        },
+        {
+          label: $.i18n._('core-views')["display-null"],
+          id: "core/display-null",
+          click: function() {
+            $(".data-table-null").toggle();
+            self._shownulls = !(self._shownulls);
         }
```
wetneb: I get the following JS error here:
`Uncaught SyntaxError: Unexpected token ]`
Both with Firefox and Chromium

ostephens: Doh! I caught that error when I tested and managed not to fix it in PR. I’ll fix when I’m back at my computer. Sorry!

PR #1581, opened 2018-04-23
Migrate "Match all filtered cells" to use standard reconciliation services

### LINKED ISSUE #1376: Reconcile->"Match All Filtered Cells to..." option
The "Match All Filtered Cells to..." action (under the Reconcile/Actions section of the column drop down menu) is designed to look up a value on Freebase and allow the user to match all cells (in the current filtered view) to that Freebase topic.

Since Freebase no longer exists, this option no longer works as expected.

On the basis we shouldn't have options that don't work, I suggest that we:

* Remove this option for now
* Consider implementing a new, similar, option, but one which (a) uses the Suggest API and (b) allows the user to choose from any Reconciliation target implementing the Suggest API which the install of OR has in its list of registered reconciliation services

Comments? @thadguidry @ettorerizza @wetneb

wetneb:
Closes #1376.

Some translations will need to be updated - I think Weblate will flag them as such (by detecting that the English version has changed).

### DIFF
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -463,7 +463,7 @@
         "valid-values": "Valid values",
         "blanks": "Blanks",
         "errors": "Errors",
-        "search-fb-topic": "Search Freebase for a topic to match all filtered cells:",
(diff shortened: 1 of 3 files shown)


wetneb:

PR #1584, opened 2018-04-25
Update httpclient and related libs

### LINKED ISSUE #1583: Incompatibility between Jena and Httpclient
The Jena libraries added in #1563 do not work with our outdated version of Apache httpclient (4.2.5) when querying SPARQL endpoints. I propose to update httpclient to a newer version. It looks like we are not using this that much inside OR itself, just when downloading files at import time.

```
./build/log4j.properties:log4j.logger.org.apache.http.client=WARN
./broker/core/src/com/google/refine/broker/RefineBrokerImpl.java:import org.apache.http.client.HttpClient;
./broker/core/src/com/google/refine/broker/RefineBroker.java:import org.apache.http.client.HttpClient;
./broker/core/src/com/google/refine/broker/RefineBroker.java:import org.apache.http.client.ResponseHandler;
./broker/core/src/com/google/refine/broker/RefineBroker.java:import org.apache.http.client.entity.UrlEncodedFormEntity;
./broker/core/src/com/google/refine/broker/RefineBroker.java:import org.apache.http.client.methods.HttpPost;
./broker/appengine/src/com/google/refine/broker/AppEngineRefineBrokerImpl.java:import org.apache.http.client.HttpClient;
./main/src/com/google/refine/importing/ImportingUtilities.java:import org.apache.http.client.ClientProtocolException;
./main/src/com/google/refine/importing/ImportingUtilities.java:import org.apache.http.client.methods.HttpGet;
./main/src/log4j.properties:log4j.logger.org.apache.http.client=WARN
./main/webapp/WEB-INF/classes/log4j.properties:log4j.logger.org.apache.http.client=WARN
```

wetneb:
Closes #1583.


jackyq2015:
According to this:
https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient/4.5.5
httpclient 4.5.5 better go with httpcore  4.4.9. Most likely your version will be working, but better stick to the compile dependency in case of surprise

wetneb:
Agreed, let's do that.

PR #1591, opened 2018-05-02
Make sure toDate always returns Calendar instances. Closes #1588

### LINKED ISSUE #1588: Date retrieved from Wikidata registering as wrong type

Comparing master+wikidata extesion with OR 2.8

Another date issue I'm afraid.  Not sure if this is related to general date handling or something about the wikidata extension - not had a chance to look yet.

Using the code from master branch + wikidata extension
Adding a column using "Add columns from reconciled values" with a date type (e.g. 'date of birth'). 

#### Current Results:

The retrieved date is not identified as a 'date' type in OR but rather as a 'Java.time.OffsetDateTime'

screen shot 2018-04-28 at 12 56 38

however, then opening the same project (no changes) in OR2.8 the same cell shows correctly as 'date' type

screen shot 2018-04-28 at 13 05 12

In addition this seems to stop date based functions working on the value in master branch:

screen shot 2018-04-28 at 13 04 31

But the same function works fine in 2.8

screen shot 2018-04-28 at 13 05 24

#### Expected Results:

Expect the data to register as a 'date' type, and for date functions to work as in 2.8

wetneb:
This is a fix for the issue reported, but the root issue is the partial migration to JDK8 date classes which is likely to cause more trouble in the future, I think.

### DIFF
--- a/main/src/com/google/refine/commands/recon/PreviewExtendDataCommand.java
+++ b/main/src/com/google/refine/commands/recon/PreviewExtendDataCommand.java
@@ -35,6 +35,7 @@ package com.google.refine.commands.recon;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Calendar;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


jackyq2015:
@wetneb can you please  write a UT to cover this case?

wetneb:
This will be redundant with your migration, so I am closing this one.

PR #1592, opened 2018-05-03
Issue/1287 data part for nano

jackyq2015:


thadguidry:
I already stated my design decision to get past the differing opinions.
That is to go forward...take the extra effort and time to go to Java 8
classes and using java.time as necessary. We will not use Java 7 classes.

wetneb:
@thadguidry dear "Lead Designer", let's have a call maybe? I value your feedback as a user but this is about a technical decision.

thadguidry:
Can't. On vacation. Do not rush. We have time to do this right is what I am
saying.

wetneb:
I was hoping to get a release before 20th May, so that it's ready for the Wikimedia Hackathon 2018, but that's not going to happen then. I guess I'll just release a snapshot with the fixes and the extension then.

jackyq2015:
@webnet No worry. you can cut a tag or branch to hold the version for the Wikimedia Hackathon 2018. At the same time, i will have a branch to fully migrate to java 8 time API. It will take a while but it is not impossible to finish before 20th. But if we cannot make it, you can still have the branch to show. :) what do you think?

wetneb:
@jackyq2015 thanks! But this is a big change that needs thorough reviewing and testing, so even if we manage to do it quickly we should not rush to release that…

PR #1594, opened 2018-05-06
Give better feedback when values with invalid formats are provided.

wetneb:
Raises explicit warnings linking to the docs when Wikibase datavalues are skipped due to format errors.


wetneb:
Merging this in a few days unless anybody has any concerns

Issue #1036, opened 2015-07-20
Using cells["col name"].value on null gives error


ostephens:
If a cell contains 'null' when you access the cell value using 'value' you get 'null':

<img width="549" alt="screen shot 2015-07-20 at 10 08 47" src="https://cloud.githubusercontent.com/assets/576174/8772847/53c30f3e-2ec7-11e5-9b36-180014f48adf.png">

However, if you access the same cell through 'cells["col name"].value you get an error instead:

<img width="565" alt="screen shot 2015-07-20 at 10 08 40" src="https://cloud.githubusercontent.com/assets/576174/8772852/5d5a3b8a-2ec7-11e5-984b-65a91785aa74.png">

This means that you get inconsistent behaviour when using GREL expressions that check for null - e.g. isNull:

<img width="435" alt="screen shot 2015-07-20 at 10 04 38" src="https://cloud.githubusercontent.com/assets/576174/8772893/ad1c3f60-2ec7-11e5-9a3a-745217d4dc80.png">

vs

<img width="472" alt="screen shot 2015-07-20 at 10 10 18" src="https://cloud.githubusercontent.com/assets/576174/8772884/9737b134-2ec7-11e5-91f4-cde75f82bbeb.png">

The behaviour here should be consistent - preferably you should get 'null' whether you use the 'value' syntax or the 'cells["Col name"].value' syntax

ostephens:
Tested on OpenRefine 2.6 beta, and OpenRefine 2.6 RC1

jackyq2015:
Very sharp eyes and very through explanation!

tfmorris:
There are a number of inconsistencies related to null handling.  There are places where nulls silently get promoted to the empty string ("") as well as this dichotomy.  I think you'll also sometimes see different results from value.operation() vs operation(value) when value is null.

Right now we're halfway between a strict, consistent implementation and a totally permissive implementation.

ostephens:
This issue has now come up under #820 (and related issues) as it causes problems when trying ignore null values when concatenating values from across columns.

I've been investigating and just trying to document before I lose the thread although probably not all relevant to the issue in hand...

If we have a null value/cell `cells["col name"].value` fails with an error. The error comes from:
https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/grel/ast/FieldAccessorExpr.java#L66

so is part of the general field accessor code rather than something specific to the `cell` model. 

[...]

PR #1598, opened 2018-05-10
No field null

ostephens:
Attempts to get a non-existent field will return null. Applies also to attempt to get a field from null
For #820 #1036

### DIFF
--- a/main/tests/server/src/com/google/refine/tests/grel/GrelTests.java
+++ b/main/tests/server/src/com/google/refine/tests/grel/GrelTests.java
@@ -99,7 +99,7 @@ public class GrelTests extends RefineTest {
     public void testEvalError() {
         String tests[] = {
 //                "1=1", // TODO: Throws NullPointerException
-                "a.value",
+                "value.datePart()",
                 };
         for (String test : tests) {
             try {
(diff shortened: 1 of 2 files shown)


main/tests/server/src/com/google/refine/tests/grel/GrelTests.java
```diff
@@ -166,6 +166,26 @@ public void testString() throws ParsingException {
...
+    @Test
+    public void testGetJsonFieldAbsent() throws ParsingException {
+        String test =  "\"[{\\\"one\\\": \\\"1\\\"}]\".parseJson()[0].two";
+        Evaluable eval = MetaParser.parse("grel:" + test);
+        Assert.assertNull(eval.evaluate(bindings));
+    }
+
+    @Test
+    public void testGetFieldFromNull() throws ParsingException {
+        String test =  "null.value";
```
tfmorris, line 185: This should be an error. There is no way to fetch a field from a null.

ostephens: Yes - as noted by @wetneb above. However, this handling for GREL is deliberate following discussion on #820 and #1036

Issue #294, opened 2012-10-15
Exporting date type column to TSV/CSV shows java debugging information instead of value


tfmorris:
_Original author: geoffh...@gmail.com (December 22, 2010 23:50:39)_

<b>What steps will reproduce the problem?</b>
1. Transform a column containing date data represented as strings (in this case, it was different formats including ones like &quot;10/30/2010&quot; as well as ones like &quot;07 January 2003&quot;) into a date type column using Edit cells -&gt; Common transforms -&gt; To date
2. Export the data as Tab-separated value or Comma-separated value

<b>What is the expected output? What do you see instead?</b>

I expect to see a similar date representation as what I see in Refine, for instance, &quot;2010-10-30T00:00:00Z&quot;

Instead, I see &quot;java.util.GregorianCalendar[time=1288414800000,areFieldsSet=true,areAllFieldsSet=false,lenient=true,zone=sun.util.calendar.ZoneInfo[id=&quot;&quot;America/
Chicago&quot;&quot;,offset=-21600000,dstSavings=3600000,useDaylight=true,transitions=235,lastRule=java.util.SimpleTimeZone[id=America/
Chicago,offset=-21600000,dstSavings=3600000,useDaylight=true,startYear=0,startMode=3,startMonth=2,startDay=8,startDayOfWeek=1,startTime=7200000,startTimeMode=0,endMode=3,endMonth=10,endDay=1,endDayOfWeek=1,endTime=7200000,endTimeMode=0]],firstDayOfWeek=1,minimalDaysInFirstWeek=1,ERA=?,YEAR=2010,MONTH=9,WEEK_OF_YEAR=?,WEEK_OF_MONTH=?,DAY_OF_MONTH=30,DAY_OF_YEAR=?,DAY_OF_WEEK=?,DAY_OF_WEEK_IN_MONTH=?,AM_PM=?,HOUR=?,HOUR_OF_DAY=?,MINUTE=?,SECOND=?,MILLISECOND=?,ZONE_OFFSET=?,DST_OFFSET=?]&quot; 

<b>What version of the product are you using? On what operating system?</b>

Google Refine 2.0 on Ubuntu Linux 10.04

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=294_

PR #1602, opened 2018-05-13
Codified [date].toString() behaviour from 2.8 in tests

ostephens:
This is my attempt to add tests for the 2.8 [date].toString() behaviour so we can replicate with the new DateTime API.

The equivalent tests in 2.8 would be written as follows, and these all pass in 2.8:

```
      String inputDate = "2013-06-01";
      String inputDateTime = "2013-06-01 13:12:11";
      Assert.assertEquals(invoke("toString", CalendarParser.parse(inputDate)), "01-Jun-2013");
      Assert.assertEquals(invoke("toString", CalendarParser.parse(inputDate), "yyyy-MM-dd"), "2013-06-01");
      Assert.assertEquals(invoke("toString", CalendarParser.parse(inputDate), "yyyy/dd/MM"), "2013/01/06");
      Assert.assertEquals(invoke("toString", CalendarParser.parse(inputDate), "yyyy-MMM"), "2013-Jun");
      Assert.assertEquals(invoke("toString", CalendarParser.parse(inputDate), "yyyy-MM-dd hh:mm:ss"), "2013-06-01 12:00:00");
      Assert.assertEquals(invoke("toString", CalendarParser.parse(inputDateTime)), "01-Jun-2013");
      Assert.assertEquals(invoke("toString", CalendarParser.parse(inputDateTime), "yyyy-MM-dd"), "2013-06-01");
      Assert.assertEquals(invoke("toString", CalendarParser.parse(inputDateTime), "yyyy-MM-dd hh:mm:ss"),"2013-06-01 01:12:11");
```

There is definitely some slightly odd behaviour here (specifically the rendering of the time which I believe to be incorrect) and I'm happy to update the tests to reflect what we think the behaviour *should* be. However, this PR is to give us a starting point for compatibility with 2.8

### DIFF
--- a/main/tests/server/src/com/google/refine/tests/expr/functions/strings/ToFromConversionTests.java
+++ b/main/tests/server/src/com/google/refine/tests/expr/functions/strings/ToFromConversionTests.java
@@ -117,12 +117,32 @@ public class ToFromConversionTests extends RefineTest {
       Assert.assertEquals(invoke("toString", Double.valueOf(100.0)),"100.0");
       Assert.assertEquals(invoke("toString", Double.valueOf(100.0),"%.0f"),"100");
       
-      String intputDate = "2013-06-01";
-      String expectedDate = "2013-06-01";
-      Assert.assertEquals(invoke("toString", CalenderParser.parseAsOffsetDateTime(intputDate)), 
-              expectedDate);
-      Assert.assertEquals(invoke("toString", CalenderParser.parseAsOffsetDateTime(intputDate), "yyyy-MM-dd"),
-              expectedDate);
+      String inputDate = "2013-06-01";
+      String inputDateTime = "2013-06-01 13:12:11";
+      Assert.assertEquals(invoke("toString", CalenderParser.parseAsOffsetDateTime(inputDate)
+              ), 
+              "01-Jun-2013");
+      Assert.assertEquals(invoke("toString", CalenderParser.parseAsOffsetDateTime(inputDate), 
+              "yyyy-MM-dd"),
+              "2013-06-01");
+      Assert.assertEquals(invoke("toString", CalenderParser.parseAsOffsetDateTime(inputDate), 
+              "yyyy/dd/MM"),
+              "2013/01/06");
+      Assert.assertEquals(invoke("toString", CalenderParser.parseAsOffsetDateTime(inputDate), 
+              "yyyy-MMM"),
+              "2013-Jun");
+      Assert.assertEquals(invoke("toString", CalenderParser.parseAsOffsetDateTime(inputDate), 
+              "yyyy-MM-dd hh:mm:ss"),
+              "2013-06-01 12:00:00");
+      Assert.assertEquals(invoke("toString", CalenderParser.parseAsOffsetDateTime(inputDateTime)
+              ),
+              "01-Jun-2013");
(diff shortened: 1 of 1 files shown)


ostephens:
@thadguidry @ettorerizza I'm inclined to change the tests 
`Assert.assertEquals(invoke("toString", CalendarParser.parse(inputDate), "yyyy-MM-dd hh:mm:ss"), "2013-06-01 12:00:00");`
`Assert.assertEquals(invoke("toString", CalendarParser.parse(inputDateTime), "yyyy-MM-dd hh:mm:ss"),"2013-06-01 01:12:11");`

As these current convert to midday and 1AM, rather than midnight and 1PM which is what I'd really expect. This is because the current behaviour in 2.8 does this as well.

Do you agree we should change this behaviour rather than keeping backward compatibility in this case?

thadguidry:
@ostephens That's 24 hour time format as the default.  I don't think we should change this.  Our tests can check for various time formats, certainly.  But 24 hour time is the default time format for nearly everything in tech and programming.  And besides... we already know that and documented as such in the code comment already for CalendarParser https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/expr/util/CalendarParser.java#L82

ostephens:
But shouldn't:

[date 2013-06-01 13:12:11].toString("yyyy-MM-dd hh:mm:ss") -> "2013-06-01 13:12:11"

? 

In 2.8 it goes to 2013-06-01 01:12:11

That doesn't seem right to me

thadguidry:
@ostephens We have always conformed to ISO8601 since Google Refine 2.1 days and we added localization of dates during 2.6 also.... In fact, I've tested this many times before, and see Tom's mention here: https://github.com/OpenRefine/OpenRefine/issues/294#issuecomment-9432067

Your doing it wrong.
What your doing is your asking for 12 hour format....instead of 24 hour format.

`value.toDate().toString("yyyy-MM-dd HH:mm:ss")`

notice my capital HH's

jackyq2015:
@ostephens Thanks for the UT. The PR branch has removed the Calendar to represent a date type. All the date cell will be represented as OffsetDateTime. so in my UT, I only pass the OffsetDateTime type as the 1st argument. The V2.8 use default Calendar constructor which use the local date. This does not sound right when dealing with the date whose data source could be come from anywhere. The test cases passed on 2.8 but won't pass in this branch. But as I mentioned, since in real life a Calendar won't be passed as argument due to all date will be using OffsetDateTime, probably it is not that relevant any more.

I extracted your UT but use the OffsetDateTime for the date type.

thadguidry:
@ostephens @jackyq2015 Guys... please ensure we have testing for 12/24 Hour formats as well !!!!  And converting back and forth between them with

`toString("yyyy-MM-dd HH:mm:ss")`

 and

`toString("yyyy-MM-dd hh:mm:ss")`

thadguidry:
@jackyq2015 and just to remind you, recall we added support for locale with toDate() in 2.6 and various internationalization bits https://github.com/OpenRefine/OpenRefine/releases/tag/2.6-alpha.2  Here is Tom's commit for the toDate() changes https://github.com/OpenRefine/OpenRefine/commit/fa072df85c27ce601f867a7f39c5a5d1bf288dec

ostephens:
@thadguidry doh! Thank you :)

ostephens:
Thanks @jackyq2015 - closing this now

Issue #1604, opened 2018-05-16
Support for Java versions > 9


martinjrobins:
Hi, it doesn't seem like OpenRefine2.8 currently supports Java versions greater than 9, can this be fixed to allow newer Java versions? 

I notice that https://github.com/OpenRefine/OpenRefine/blob/master/refine#L843 only allows versions 6-9, but not greater than this. Adding a 10 onto this list worked OK for me

#### Version of OpenRefine used (Google Refine 2.6, OpenRefine2.8, an other distribution?):

OpenRefine2.8

#### Operating Systems and version:
Ubuntu 18.04

#### Browser + version used - Please note that OpenRefine doesn't support Internet Explorer but works OK in most cases:

Google Chrome version 66.0.3359.139 

#### Steps followed to create the issue:

```
$ ./refine
```

#### Current Results:

Error: OpenRefine requires Java version 6 or later. If you have multiple versions of Java installed, please set the environment variable JAVA_HOME to the correct version.

#### Expected Results:

OpenRefine to start-up

wetneb:
Indeed this constraints needs updating as we no longer support Java 7 or older. Would you consider making a pull request for that?

martinjrobins:
Can do, created a pull request here #1605

PR #1605, opened 2018-05-16
#1604 upgrade java version check to accept all versions greater than 6

martinjrobins:
I've removed support for version 6, and allowed all java versions greater than or equal to 7. Is this the range of java version supported by OpenRefine?

### DIFF
--- a/refine
+++ b/refine
@@ -839,10 +839,8 @@ checkJavaMajorVersion() {
     # Java 9+ starts with x using semver versioning
     major=`echo ${java_ver} | sed -E 's/([0-9]+)(-ea|(\.[0-9]+)*)/\1/g'`
   fi
-  
-  JAVA_VERSION=$(echo ${major} | egrep '^(6|7|8|9)')
-  if [ -z "$JAVA_VERSION" ] ; then
-    error "OpenRefine requires Java version 6 or later. If you have multiple versions of Java installed, please set the environment variable JAVA_HOME to the correct version."
+  if (( ${major} < 7 )); then
+    error "OpenRefine requires Java version 7 or later. If you have multiple versions of Java installed, please set the environment variable JAVA_HOME to the correct version."
   fi
 }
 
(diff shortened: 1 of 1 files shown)


thadguidry:
@martinjrobins Our current testing platform on Travis and Appveyor do not test against Java 9 or 10 or 11.  We traditionally setup our testing frameworks with the latest release Java and run tests against those and fix issues that come up.  I am not against this PR, but it might expose bugs/issues for users that we have not yet discovered.  Then again, those users NOT using the latest Java are probably very advanced users to begin with. :) 

We could certainly also improve our testing framework scripts, so that we test on current Java and the next version upcoming.  Feel free to submit a pull request against our Travis or Appveyor files as well for that.

jackyq2015:
@martinjrobins Have you tried OR on 10? I never have one so I am not sure. for 9, we did some testing but  there was some minor issue at that time. It maybe has been fixed.  

I would prefer to have the version 10 tested before we put into script otherwise will give people false hope. :)

martinjrobins:
At the moment it looks like you are testing against Oracle JDK 8 on travis. it would be simple enough to add Oracle JDK 9 and Open JDK 6-8, as travis seems to support these (https://docs.travis-ci.com/user/reference/trusty/#JavaScript-and-Node.js-images). Travis does not currently support version 10, although this looks to be in the works (https://github.com/travis-ci/travis-ci/issues/9368)

The reason why I ran into this issue is that Ubuntu 18.04 uses open JDK 10.0.1 as the default version, so  I would suggest that blocking OpenRefine use on this platform is a bad idea. OR worked fine for me using Open JDK 10, just playing around applying filters to a single dataset, although this was certainly not thorough testing. Perhaps the error could be changed to a warning for newer, untested versions of java? Then once travis supports Open/Oracle JDK 10 start testing and update the warning?

martinjrobins:
Added a warning for Java versions > 9, see what you think.

PR #1611, opened 2018-05-18
Fix inverse constraint test for Wikidata extension.

wetneb:
The inverse constraint on the part of property was removed.


wetneb:
Exceptionally merging without review because it is necessary to review other PRs where the test fails.

PR #1617, opened 2018-05-22
Remove DataExtensions from change serialization

wetneb:
It turns out that the part of DataExtension changes that was not serialized consistently is not actually needed to apply and unapply the change. So I have just removed this deserialization and serialization. The code accepts changes made by OR 2.8 and the changes created on master (future OR 3.0) are also retro-compatible with OR 2.8.

### DIFF
--- a/main/src/com/google/refine/model/changes/DataExtensionChange.java
+++ b/main/src/com/google/refine/model/changes/DataExtensionChange.java
@@ -1,6 +1,6 @@
 /*
 
-àCopyright 2010, Google Inc.
+Copyright 2010, Google Inc.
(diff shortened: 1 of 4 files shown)


wetneb:
The failing tests are due to Wikidata changing the way they store some constraints… Grrrr.

jackyq2015:
why the "apply and unapply the change" is not needed for DataExtensions?

wetneb:
@jackyq2015 because the former rows and the new rows are also serialized - so the DataExtension objects themselves are actually redundant with that. See the `unapply` code for instance: no reference is made to the DataExtensions.

PR #1638, opened 2018-06-04
Do not discard the recon space when marking cells as new items.

### LINKED ISSUE #1637: Reconciliation

v 3.0 beta

MacOS 10.13.4

Chrome 66.0.3359.181

- Reconcile Finnish street names with Wikidata (P31 --> street, P131 --> Helsinki)
- For some the Finnish label is missing -->
- Reconcile against English although English labels don't exist (yields right results)
- Manually mark the remaining items
a) Choose new match --> search for match --> Don't reconcile cell or
b) Create new item
- The green bar does not show 100% coverage although all items have been marked something. Even after removing the items marked "don't reconcile" some remain not marked (not green).
- Schema page does not display a reconciled column (green underline), although it may suddenly appear perhaps depending on the chosen subset of items.
- I was able to set up the data for Wikidata import once, but aborted it because of a mistake in the data. This challenge has been persistent since. I suspect the reconciliation data is corrupt. I have restarted OpenRefine, cleared reconciliation data etc. but the problem remains.

#### If you are allowed and are OK with making your data public, it would be awesome if you can include the data causing the issue or a URL pointing to where the data is (if your concerned about keeping your data private, ping us on our mailing list):

[Katuja-Helsingiss.openrefine.tar.gz](https://github.com/OpenRefine/OpenRefine/files/2066037/Katuja-Helsingiss.openrefine.tar.gz)

#### Current Results:
Not able to upload to Wikidata

#### Expected Results:
Able to upload to Wikidata

wetneb:
I have made the same fix as before, for the other two (?!) operations that mark cells as new items. The same caveat about judging cells in unreconciled columns applies.

Fixes #1637.

### DIFF
--- a/main/src/com/google/refine/commands/recon/ReconJudgeOneCellCommand.java
+++ b/main/src/com/google/refine/commands/recon/ReconJudgeOneCellCommand.java
@@ -167,10 +167,23 @@ public class ReconJudgeOneCellCommand extends Command {
             }
 
             Judgment oldJudgment = cell.recon == null ? Judgment.None : cell.recon.judgment;
+            
+            Recon newRecon = null;
+            if (cell.recon != null) {
+                newRecon = cell.recon.dup(historyEntryID);
+            } else if (identifierSpace != null && schemaSpace != null) {
+                newRecon = new Recon(historyEntryID, identifierSpace, schemaSpace);
+            } else if (column.getReconConfig() != null) {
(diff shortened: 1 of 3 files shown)


wetneb:
Actually, to fix the issue, we could also ensure that if we already have reconciliation data for the cell, but the identifierSpace is null, then infer it from the ReconConfig…

PR #1641, opened 2018-06-08
Mass edit fix for blanks, numbers and booleans

ostephens:
This fix is for #1631 and #180

Issue #1631, opened 2018-05-30
Using "Apply to All Identical Cells" gives error when cell contains Number or Boolean


ostephens:
When reporting a bug please provide the following information to help reproduce the bug:

#### Version of OpenRefine used (Google Refine 2.6, OpenRefine2.8, an other distribution?):
3.0 Beta

#### Steps followed to create the issue:
Edit two cells to contain the same Number or Boolean
Click edit in one of the cell, change the value, and click "Apply to All Identical Cells"

#### Current Results:
Error in pop-up:
JSONArray[0] not a string.

#### Expected Results:
The identical cells should all update to the edited value

See also #332

ostephens:
From console:

```
23:56:25.124 [                  command] Exception caught (1ms)
org.json.JSONException: JSONArray[0] not a string.
	at org.json.JSONArray.getString(JSONArray.java:405)
	at com.google.refine.operations.cell.MassEditOperation.reconstructEdits(MassEditOperation.java:124)
	at com.google.refine.commands.cell.MassEditCommand.createOperation(MassEditCommand.java:59)
	at com.google.refine.commands.EngineDependentCommand.doPost(EngineDependentCommand.java:78)
	at com.google.refine.RefineServlet.service(RefineServlet.java:178)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:820)
	at org.mortbay.jetty.servlet.ServletHolder.handle(ServletHolder.java:511)
	at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1166)
	at org.mortbay.servlet.UserAgentFilter.doFilter(UserAgentFilter.java:81)
	at org.mortbay.servlet.GzipFilter.doFilter(GzipFilter.java:132)
	at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1157)
	at org.mortbay.jetty.servlet.ServletHandler.handle(ServletHandler.java:388)
	at org.mortbay.jetty.security.SecurityHandler.handle(SecurityHandler.java:216)
	at org.mortbay.jetty.servlet.SessionHandler.handle(SessionHandler.java:182)
	at org.mortbay.jetty.handler.ContextHandler.handle(ContextHandler.java:765)
	at org.mortbay.jetty.webapp.WebAppContext.handle(WebAppContext.java:418)
[...]

Issue #180, opened 2012-10-15
Editing single blank cell and applying to identical cells does not work affect other blank cells


tfmorris:
_Original author: dfhu...@google.com (November 05, 2010 17:15:12)_

Click edit on a blank cell, enter something, and click Apply to identical cells. That does not affect other blank cells.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=180_

ostephens:
Just an update - I'm working on this and work in progress is at https://github.com/ostephens/OpenRefine/tree/mass-edit-fix

tl/dr fix #1631 is relatively straightforward, but I'm trying to make tests and fix #1632, #180 and #332 at the same time which is proving more complicated and taking more time

ostephens:
Fixed by #1642

PR #1642, opened 2018-06-08
Mass edit fix for blanks, numbers and booleans

ostephens:
This fix is for #1631 and #180
(clean version of previous PR #1641)


ostephens:
The fix for #180 treats empty strings and nulls as the same. I think that's OK for now (and is also how the mass edit from a text facet across the same values would work).

We may want to revisit this later to at least offer the option to handle empty string and null separately, but that's beyond the scope of this fix

ostephens:
I'll merge this in a day if there are no further comments

PR #1643, opened 2018-06-08
Increasing to 10 second sleep because of Travis

thadguidry:
In Travis these tests are occasionally failing https://travis-ci.org/OpenRefine/OpenRefine/jobs/3896227 which is probably due to the CPU demand spike placed with createProcess.


thadguidry:
@wetneb However not sure what to do about the occasional Wikibaseapi issues.

wetneb:
@thadguidry I will remove some of them so that it reduces these bugs a bit

PR #1647, opened 2018-06-13
Change default precision of Wikibase dates to match the UI's defaults.

wetneb:
The wikibase data model lets users store dates with precisions. The Wikidata-Toolkit library and the Wikibase implementation disagreed on the default values for these precisions. This leads to conflicting values being added to the items (see https://www.wikidata.org/w/index.php?title=Q391056&action=history for instance).

This was fixed in Wikidata-Toolkit (https://github.com/Wikidata/Wikidata-Toolkit/pull/375) and this PR brings the fix to OpenRefine.

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
@@ -119,7 +119,7 @@ public class WbDateConstant implements WbExpression<TimeValue> {
             calendar.setTime(date);
             return Datamodel.makeTimeValue(calendar.get(Calendar.YEAR), (byte) (calendar.get(Calendar.MONTH) + 1), 
                     (byte) calendar.get(Calendar.DAY_OF_MONTH), (byte) calendar.get(Calendar.HOUR_OF_DAY),
-                    (byte) calendar.get(Calendar.MINUTE), (byte) calendar.get(Calendar.SECOND), (byte) precision, 0, 1,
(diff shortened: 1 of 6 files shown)


extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
```diff
@@ -119,7 +119,7 @@ public static TimeValue parse(String datestamp)
             calendar.setTime(date);
             return Datamodel.makeTimeValue(calendar.get(Calendar.YEAR), (byte) (calendar.get(Calendar.MONTH) + 1), 
                     (byte) calendar.get(Calendar.DAY_OF_MONTH), (byte) calendar.get(Calendar.HOUR_OF_DAY),
-                    (byte) calendar.get(Calendar.MINUTE), (byte) calendar.get(Calendar.SECOND), (byte) precision, 0, 1,
```
jackyq2015: what's the implication of this change? Maybe add some comments since Datamodel is from wikidata.

wetneb: the implication is that now OpenRefine and the Wikidata UI will agree on the default value for this "after" parameter. I can elaborate here but I am not sure how useful it would be to add more comments in the code itself, given that it's just one of the default values (with the "before" and "timezone" parameters) - having a default value of 0 is fairly intuitive, no?

wetneb:
merging this in a few days unless anybody disagrees

Issue #1635, opened 2018-06-02
Transform value.toString() converting (blank) values in a cell to a string with 'null' as the value  


dbutlerdb:
When reporting a bug please provide the following information to help reproduce the bug:
When reviewing open issues this #332 seems to be the closest I could get to, but what is being described is not the same thing I am running into. 

#### Version of OpenRefine used (Google Refine 2.6, OpenRefine2.8, an other distribution?):
2.8 and 3.0-beta

#### Operating Systems and version:
Windows 10 Version 1803

#### Browser + version used - Please note that OpenRefine doesn't support Internet Explorer but works OK in most cases:
Chrome Version 66.0.3359.181 (Official Build) (64-bit), using the API via python 2.7 and 3.6 (only http request)

#### Steps followed to create the issue:
Started new project with the attached file. 
[louisiana-elected-officials.zip](https://github.com/OpenRefine/OpenRefine/files/2065279/louisiana-elected-officials.zip)
In the Zip Code column clicking transform. I use the following code
`value.toString()[0,5]`
What happens is it converts the previously (blank) rows into 'null' strings

#### Current Results:
value.toString() on nulls is turning them into strings with the value 'null'

#### Expected Results:
null / blank rows  kept as null / blank values in the row

thadguidry:
Hi !  We just released 3.0 Beta of OpenRefine.  Please try it out, as we have add several new features to help you with NULL values, such as new Facets and a new Toggle mode for displaying NULL cells.  You can read about the new features and download the newest version on our releases page https://github.com/OpenRefine/OpenRefine/releases

We would appreciate any feedback you can give us on how we can make things even easier for your needs !

thadguidry:
Sorry, WHAT are you trying to perform ?  So you have some blank Zip Codes on 891 rows... they basically look like this ""  ... so what do you want to do with that empty string (also called a blank) ?

ostephens:
@dbutlerdb 
The short explanation is that the behaviour of 'toString' on a null cell is deliberately set to result in the string `null`.

The short solution:
Use the 'Facet on blank' option on the column to remove all rows with blank and/or null values before applying the 'toString' transformation on the column

More information: 
probably worth giving some background on this first - apologies if some/all of this you may know or is obvious

Each cell in an OpenRefine project contains a value of a specific 'type'. The types supported are:

- String
- Number
- Date
- Boolean
- error
- null

[...]

PR #1649, opened 2018-06-14
Null to string

ostephens:
Changed behaviour of using `null`.toString() as per #1635
This makes the behaviour constistent with using "string" + `null` and more intuitive/safer for users


ostephens:
This has some extraneous commits in it - going to close and try a tidier PR

PR #1655, opened 2018-06-15
Migrate to new scope constraint format

wetneb:
Wikidata is changing the way they are representing scope constraints. This PR adapts the code to the new format.
Discussion: https://www.wikidata.org/wiki/Wikidata:Property_proposal/property_scope


wetneb:
merging this in a few days unless anybody disagrees

Issue #1445, opened 2018-01-27
Installer on package for Linux


pabloab:
Would be great to offer/build Linux installer on a package form: Classic DEB, RPM or new AppImage/Snap/Flatpak. Either way also would be great to have it available on official repos of main Linux distros.

wetneb:
@pabloab That's very true! This mostly requires familiarity with these packaging systems rather than OpenRefine itself, so it's a good task for anyone who wants to help without diving in the internals too much.

magdmartin:
@pachamaltese can you send a PR so our core team can review and provide feedback and merge it?

thadguidry:
@pachamaltese Hi!  We would like to see this issue get some love.  Do you think you could volunteer as a maintainer for our Linux packaging?  What do you need from us to help, let us know.

pachamaltese:
@thadguidry HI! Of course, let me retake this. A lot of things have changed since my last commit to this, so now I can dedicate time to this.

pachamaltese:
@thadguidry I shall follow package creation in a very similar way to RStudio, I; ll start with DEB files

pachamaltese:
@thadguidry here is an updated deb package https://github.com/pachamaltese/openrefine-deb/releases/download/3.2/openrefine-3.2.deb
now I shall find a way to create an rpm package for fedora/fedora-based

thadguidry:
@pachamaltese How's the progress going on this?  By the way, we are about to push out Github Actions to create nightly releases (and later general releases, packages) so this might be an opportunity to coordinate.

@wetneb @magdmartin any other thoughts on this issue and Github Actions?

pachamaltese:
hi!!

debian/ubuntu: ok and ready

fedora/centos: I need to ask a friend, it's a different world

—————

*Mauricio Vargas Sepúlveda 帕夏*
Do you like Data Science? visit pacha.hk
你爱科学数据专吗？你走pacha.hk

pachamaltese:
@thadguidry Hi. I shall need help to translate the Ubuntu/Debian work (ready) to Fedora/CentOS/RedHat. It's a totally different world. The unified solution would be to create a snap package, but I'd avoid it as much as I can.

wetneb:
Feel free to submit your DEB packaging on its own - it's already something that would be very welcome even without an accompanying RPM.

pabloab:
[...]

PR #1660, opened 2018-06-19
Deb package

pachamaltese:
@magdmartin here's the PR for now
it should be tested so let's work on a PPA later


wetneb:
Hi @pachamaltese - thanks for this PR! However, there is a problem: you deleted most files from our repository. For OpenRefine to work, we need these files to stay.
Can you try making a new pull request without deleting any files? Thanks.

magdmartin:
it was first PR for #1445 (linking the two tickets)

magdmartin:
@pachamaltese  any update on the new PR?

pachamaltese:
@magdmartin are you ok with a PPA or just a package that works for any Debian based distro?

magdmartin:
@thadguidry @wetneb @jackyq2015  any preferences?

thadguidry:
prefer the widest distro packaging...if that is .deb or whatever nowadays,  ppa seems limiting?

wetneb:
Ideally a PPA would help I think - it is more useful than just a .deb file. I am not familiar with Debian's policies but in an ideal world we could try getting in the official repository. But that is probably a lengthy process.

PR #1661, opened 2018-06-22
Mass edit dates

### LINKED ISSUE #1632: Using "Apply to All Identical Cells" does nothing when cell contains Date

3.0 Beta

Edit two cells to contain the same Date
Click edit in one of the cell, change the value, and click "Apply to All Identical Cells"

#### Current Results:
No cells are updated (including the one that you started to edit).
Message on screen states "Mass edit 0 cells in column <column name>"

#### Expected Results:
The identical cells should all update to the edited value

See also #332 and #1631

ostephens:
This PR fixes #1632.
This is a surprisingly complex area, as it relates to how dates are transformed to strings. This can happen in different ways in different places in OR. When converting a date to a string results in a different value to the required one, it can result in unexpected behaviour.

This PR:

- Adds tests for how Mass Edits are reconstructed for dates to ensure OR does this correctly. This test fails in OR 3.0 beta
- Changes the behaviour of StringUtils.toString() to convert OffsetDateTime to a string using ParsingUtilities.dateToString -  this helps to standardise behaviour in converting dates to strings in OpenRefine, including for Mass Edits. This makes the test introduced a passing test
- Changes the expected behaviour being tested for in ToFromConversionTests.java to expect new format when using StringUtils.toString()

These changes had a knock on impact on the behaviour of using 'Edit' from a List facet. The representation of non-string objects in a List/Text facet has been broken for some time, but the mass edit function had worked for dates, so this PR restores that working functionality by:

- Adding tests for the facet grouper to make sure that for date values the facet 'value' converted to a string and facet 'label' are the same
- Changes the behaviour of DecoratedValue to pass the test (by always converting dates to Strings before populating the facet value)

The behaviour of text or list facets in relation to non-string values needs much more thought and work,  but the existing behaviour (2.8 and earlier) is also incorrect, and there is no attempt to fix further than ensure mass edit works.

Finally the implication of changing the behaviour of StringUtils.toString() is that the default format obtained when using 'toString()' in GREL is to convert to a string with the format:
`yyyy-MM-ddThh:mm:ddZ`
instead of
`dd MMM yyyy`

which was the previous default.
This change in behaviour seems relatively minor and ensures that using 'toString' results in the same value in the cell visually whether the cell is left as a Date type or converted to a String type

### DIFF
--- a/main/src/com/google/refine/browsing/DecoratedValue.java
+++ b/main/src/com/google/refine/browsing/DecoratedValue.java
@@ -33,12 +33,14 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package com.google.refine.browsing;
 
+import java.time.OffsetDateTime;
 import java.util.Properties;
 
 import org.json.JSONException;
 import org.json.JSONWriter;
 
 import com.google.refine.Jsonizable;
+import com.google.refine.util.StringUtils;
 
 /**
  * Store a value and its text label, in case the value is not a string itself.
@@ -52,8 +54,12 @@ public class DecoratedValue implements Jsonizable {
     final public String label;
     
     public DecoratedValue(Object value, String label) {
-        this.value = value;
-        this.label = label;
+      if (value instanceof OffsetDateTime) {
+          this.value = StringUtils.toString(value);
+      } else {
+          this.value = value;
+      }
+      this.label = label;
     }
     
     @Override
--- a/main/src/com/google/refine/operations/cell/MassEditOperation.java
+++ b/main/src/com/google/refine/operations/cell/MassEditOperation.java
@@ -59,6 +59,7 @@ import com.google.refine.model.changes.CellChange;
 import com.google.refine.operations.EngineDependentMassCellOperation;
 import com.google.refine.operations.OperationRegistry;
 import com.google.refine.util.ParsingUtilities;
+import com.google.refine.util.StringUtils;
 
 public class MassEditOperation extends EngineDependentMassCellOperation {
     final protected String         _expression;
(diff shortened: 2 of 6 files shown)


thadguidry:
@ostephens Right...and CONSISTENCY MATTERS.  So, I'm glad that your making it so.

But...ummm, why is Fetch URLs not working ???  Have a few test rows with values like
 `http://www.wikidata.org/entity/Q457689`
`http://www.wikidata.org/entity/Q458573`

ostephens:
See also #1662 for discussion of list facet behaviour with non-strings

ostephens:
Following discussion no #1662 there may be some further changes to make before this is ready to merge. 

Comments on this PR and on #1662 would be welcome

jackyq2015:
what's the connection between the fix of #1632 and changing the default format of StringUtil's toString(date)? I think the previous format is more concise and human-readable.

ostephens:
The issue is that the mass edit of dates relies on a string representation of the date. If different string representation of the dates are used in different places then the mass edit doesn't work.

The way List facets are built also relies on a string representation of the date, which means that making a change to the date format and the resulting string representations of those dates causes knock on problems. In particular both the 'v' and 'l' values are used creating the List facet and list facet counts, which means you can't just change the 'v' and leave the 'l' (label) alone.

My argument for changing the default format of StringUtils toString is so we consistently do the same thing when we transform a date to a string. I don't have strong opinions on the conciseness or human readability of the date format - what I've done is bring this one outlier in line with everything else.

However, it should be noted that following discussion at #1662 this may become irrelevant as I'm now working on a change to this PR which means that non-strings only appear as a grouped set in List facets. This would remove the issue of the default transform of a date to a string completely from the Facet, and so in that scenario the previous default behaviour in StringUtils could be changed back. I'd be against that because it seems inconsistent with the rest of the date behaviour in OpenRefine (in particular the display of dates in both the transformation preview and the data grid is different to what you actually get if you do date.toString() - that seems inconsistent to me)

jackyq2015:
Making the conversion consistent make sense. 

Few things I observed: 
- After pull your PR on local, I still can see the issue of #1632. Ie, when selecting a date facet, it gives me nothing instead of the 2 date rows
- This is an issue on version 2.7 also
- The "facet by timeline" works weird. The right boundary controller is gone

ostephens:
@jackyq2015 thanks for this - this PR doesn't address the issue of the selection of a date in a text facet - see new PR #1666 and related issue #1662.

I think these issues have been in OR for some time to a greater or lesser extent - see @thadguidry comments on #1662

Issue #1662, opened 2018-06-23
Behaviour of date, number and boolean values in a test/list facet is inconsistent


ostephens:
#### Version of OpenRefine used (Google Refine 2.6, OpenRefine2.8, an other distribution?):
Tested on OpenRefine 2.8 & 3.0 beta, but suspect the same behaviour is in earlier versions of OR as well

#### Current behaviour
There are a variety of scenarios - some described here:
Create a 'text facet' on a column that contains date values (i.e. as dates, not strings)
Try selecting rows by clicking the date value in the facet
See that no rows are selected (and in 2.8 at least an additional value appears in the facet)

![screen shot 2018-06-23 at 09 40 48](https://user-images.githubusercontent.com/576174/41807640-95def00c-76c9-11e8-9dfe-3009911b41dd.png)
![screen shot 2018-06-23 at 09 40 57](https://user-images.githubusercontent.com/576174/41807641-95f79e9a-76c9-11e8-98d2-46b9e424d4f7.png)

Create a column containing non-string values and strings that have the same visible value - e.g.:
```
true -> boolean
"true" -> string
```
Look at the facet and see the "count" is the number you would expect if you  treated the string and non-string values as being equivalent.

Try selecting the value in the facet - see only the rows containing either the non-string, or only the string values are selected (which set are selected depends on the order of the cells in the project)

![screen shot 2018-06-23 at 09 35 31](https://user-images.githubusercontent.com/576174/41807629-622302b2-76c9-11e8-8b26-60695684d411.png)
![screen shot 2018-06-23 at 09 35 41](https://user-images.githubusercontent.com/576174/41807630-62434acc-76c9-11e8-8258-f272ca067e76.png)

This latter behaviour also effects numbers and dates
#### Expected Behaviour:
There is a fundamental question here of how date, number and boolean values should be treated in text/list facets. They are currently counted as if they are equivalent to strings, and also if you try to use 'mass edit' they are treated as if they are the same as the equivalent string (e.g. in the example above do an edit from the facet and the boolean `true` and the string "true" would both be changed in the project.

I think trying to treat objects as equivalent to some strings in this situation is probably a bad idea. I can see two options

1) Similarly to how timeline/number facets treat other values OR could simply not include the non-string values in the facet and have a checkbox as to whether they are included in the filter or not - see
[...]

PR #1666, opened 2018-06-28
List facet changes

ostephens:
As per discussion #1662 this PR changes the behaviour of List facets.

With this PR, List facets will no longer convert non-string values of Dates, Numbers or Booleans to a string for display in the facet, but will instead group together these non-string types in the facet:

screen shot 2018-06-28 at 07 24 54

There is no 'edit' function next to these groupings.

Note, the 'mass edit' behaviour has not changed - which means that editing string values such as 'true' will also update cells containing the Boolean value of `true` - this is the same behaviour as 2.8 and earlier.


thadguidry:
@ostephens Nice.  The functionality of it is fine for me.  Like I said before in issue, this can be enhanced further later on once we get a new UI, but for now its functions as expected.  Where later you might want to catch when they are about to mass edit a 'true' or 'false' and throw up a pop-up warning "This Edit will also update rows of boolean values True or False", since it is a bit unexpected when they see 2 groupings and select 1 to mass edit and lo-and-behold the other grouping also changes even when they didn't select it, but that's exactly what we have been doing since 2.0 days.

So up to you on warning dialogs, now or later.

wetneb:
@ostephens I tried to fix the merge conflicts earlier but the tests did not pass after that. Could you have a look? It would be great to have this PR merged before 3.0.

ostephens:
Just noticed that this PR breaks "Facet by blank" functionality. I think it's an easy fix, but I need to do this before this PR is merged.

thadguidry:
One small question:  Are we OK with having "undefined" showing for null cells when using a Custom Facet with grel:value.type()  ?  Or should those cells show up in the facet as "null" ? hmmm...

capture

wetneb:
The backwards incompatibility with previous workflows spotted by @ostephens worries me a bit. (Very good catch by the way!) Maybe we should only merge this after the next release, and make sure the changelog explains how to migrate existing workflows by adding the extra `.toString()` to the facet expressions.

thadguidry:
No need to worry.  Folks that want to continue to have surprises in their data and hiding away those surprises can continue to use 2.8 and previous.  If they want less surprises and better Facets that show data inconsistencies then they can move forward with OpenRefine 3.0+  Those that move forward will have minimal changes to adapt their workflows.

This is one of those features where it will actually help our users, rather than surprise them.  Since we will no longer hide their data inconsistency, but expose that inconsistency fully with these Facet changes...so that they can deal with them and clean it up.

Merge it.

ostephens:
@wetneb yes. As noted it potentially breaks reapplying transforms from previous versions of refine but @thadguiry has argued quite strongly to go ahead and I'm convinced

PR #1670, opened 2018-07-02
Simplify parsing of HTTP headers in fetch-url operation

### LINKED ISSUE #1669: Fetching from URL: Crossref reports "HTTP error 401" (no authorization required)
**Describe the bug**
When fetching a URL from Crossref the API returns "HTTP error 401 : Unauthorized (...)". The Crossref API does not require an authorization (cf. https://www.crossref.org/services/metadata-delivery/rest-api-users/) though.
API call worked fine in OR 2.8 + 3.0 beta until June 21st 2018. Error first occured on June 25th. Contacted Crossref Support June 26th – they cannot identify error on their side. They told us to look into this further, status unclear at the moment

**To Reproduce**
1. Create Column 'DOI' (example value: `10.1016/j.csndt.2016.09.002`)
2. Edit Column -> Add column by fetching URLs
3. Leave Default Language (GREL) + Add 'Expression': `"http://api.crossref.org/works/" + value`

**Current Results**
`HTTP error 401 : Unauthorized | <html><body><h1>401 Unauthorized</h1> You need a valid user and password to access this content. </body></html>`

**Expected behavior**
return Crossref Metadata in JSON format

tested on different systems with both OR 2.8 and 3.0 – latest systems were:
1. Windows
 - Windows 10
 - Chrome 67

2. Mac
 - OSX 10.10.5
 - Chrome 65.0...

- OpenRefine 3.0 Beta

**Datasets**
some example DOIs:
10.1103/physrevb.96.041124
10.1103/physreve.95.052419
10.1155/2017/6820482

**Additional context**
Workaround for this problem: API call in Jython returns the expected data in JSON format

Steps to workaround this behavior:
1. Create Column 'DOI' (example value: 10.1016/j.csndt.2016.09.002)
2. Edit Column -> Add column based on this column
3. Change Language: Python / Jython
3. Add 'Expression':
```
import urllib2
url = 'http://api.crossref.org/works/' + value
response = urllib2.urlopen(url)
json = response.read()
json = str(json)
return json
```
4. See JSON Data from Crossref :)

wetneb:
Closes #1669 and makes it possible to specify other headers via the JSON representation of the operation.

@michaelavoigt could you try this and confirm that it solves the issue on your side? Let me know if you need help to run the development version on this branch.

@ostephens what do you think of this simpler version? It lets users specify arbitrary HTTP headers in the JSON representation of the operation - I guess that could be useful for some APIs using some non-standard headers.

### DIFF
--- a/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
+++ b/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
@@ -338,16 +338,12 @@ public class ColumnAdditionByFetchingURLsOperation extends EngineDependentOperat
             try {
                 URLConnection urlConnection = url.openConnection();
                 if (_httpHeadersJson != null) {
-                    Map<String, String> httpHeaders = new HashMap<>();
-                    for (int i = 0; i < _httpHeadersJson.length(); i++) {            
+                    for (int i = 0; i < _httpHeadersJson.length(); i++) {
                         String headerLabel = _httpHeadersJson.getJSONObject(i).getString("name");
                         String headerValue = _httpHeadersJson.getJSONObject(i).getString("value");
-                        httpHeaders.put(headerLabel, headerValue);
-                    }
-                    for (String headerLabel : HttpHeadersSupport.getHttpHeaderLabels()) {
-                        HttpHeaderInfo info = HttpHeadersSupport.getHttpHeaderInfo(headerLabel);
-
-                        urlConnection.setRequestProperty(info.header, httpHeaders.get(headerLabel));
+                        if (headerValue != null && !headerValue.isEmpty()) {
+                            urlConnection.setRequestProperty(headerLabel, headerValue);
+                        }
                     }
                 }
 


thadguidry:
@wetneb just tried.  Has the same error.
capture

Interestingly, in by Git Bash if I do `curl -v http://api.crossref.org/works` it seems to retrieve.

wetneb:
@thadguidry hmmm, curious… are you 100% sure that you recompiled on this branch? Can you show your JSON operations?

ostephens:
Hi @wetneb I need to think about this. I deliberately wrote it that way to make it easily extensible by extensions and to stop arbitrary headers being pushed. I think it also deals with some issues in having different naming in the JSON to the header key (from memory I think there was an issue with valid characters around this).

I need to take a look at the code again and try to remember all the reasons I did it that way!

michaelavoigt:
dear @wetneb wow, you were quick! Thanks for looking into it. To be honest, I have never installed patches for OR before and would need some help. But since somebody confirmed that your fix is currently working - it's maybe not necessary for me to try? Again, thank you...

wetneb:
@michaelavoigt sure, no problem!

@ostephens yes I think I understand the spirit of your original version. My reasoning for this new version is as follows:
* Letting users use arbitrary headers is a feature - it accommodates for more situations and does not pose a security risk - the user controls these headers fully, they do not depend on the project data.
* KISS principle
but I'm happy with anything else that solves the issue, really.

ostephens:
So I think the problem for #1669 is that the current code (before this PR) submits the Authorisation header with an empty string. Any fix which stops this should resolve #1669 

Having reviewed the code, the original code was intended to stop the passing of arbitrary headers to end services - basically just sanitising end-user input. On the basis that this doesn't present much of a security risk (if at all), happy to simplify as @wetneb has here.

wetneb:
@jackyq2015 good catch - I'll remove that.

ostephens:
I think that still needs including to register the headers for use? This needs to happen somewhere

ostephens:
HttpHeadersSupport is also referenced in /main/src/com/google/refine/commands/project/GetModelsCommand.java

jackyq2015:
@ostephens I am little bit confused about the intention of the HttpHeadersSupport class. It seems was designed as an immutable class to inject few http headers. In that case, how can a user to setup something like:
` Authorization: Basic YWxhZGRpbjpvcGVuc2VzYW1l `

ostephens:
@jackyq2015 the HttpHeadersSupport class was originally meant to control which HTTP Headers were supported. It provides a registerHttpHeader method so that extensions can also register headers.

It uses the registerHttpHeader internally to register the ones we currently support and set default values.

The registered headers and default values are passed to the client via GetModelsCommand.java

In the client these are displayed in the Add Column from URL dialogue, and the user can change the values (e.g. set Authorization to "Basic YWxhZGRpbjpvcGVuc2VzYW1l") in that dialogue.

It is the values passed from the client that are used when making the http request which is used to put values into the column being added.

Hope that helps

PR #1672, opened 2018-07-04
translation-en.json: add colon to "custom" label

waldyrious:
The styling of the text field that appears after this label makes it easy to miss as an input field.
Having the colon after the label makes it clearer that the user can type into the text box.

### DIFF
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -151,7 +151,7 @@
         "col-separated-by": "Columns are separated by",
         "commas": "commas (CSV)",
         "tabs": "tabs (TSV)",
-        "custom": "custom",
+        "custom": "custom:",
         "escape": "Escape special characters with \\",
         "use-quote": "Use character",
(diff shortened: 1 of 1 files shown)


jackyq2015:
Should this be part of the translation or part of html/style? I would prefer this change on the html/style to mix up those two. Also, the change only make the English looks better and that should be nothing to do with the languages.

waldyrious:
Good point, @jackyq2015. Should the change be made instead on this line? https://github.com/OpenRefine/OpenRefine/blob/f1416ae50c4bd170f3225936e206ca79af885484/main/webapp/modules/core/scripts/index/parser-interfaces/separator-based-parser-ui.js#L137

I'm not really sure how to do it there, any hints welcome.

wetneb:
I am not sure it is better to put this in the HTML. The typographic rules for these things depend on languages. For instance, in French you would need to put a space before the ":", but not in English. It would be weird to change the French translation to have a space at the end (the translator would need to know that it is going to be concatenated with a ":" afterwards).

PR #1705, opened 2018-08-10
improve log descriptions for stale import job removals

thadguidry:


main/src/com/google/refine/importing/ImportingManager.java
```diff
@@ -307,12 +307,13 @@ static private void cleanUpStaleJobs() {
         synchronized(jobs) {
             keys = new ArrayList<Long>(jobs.keySet());
         }
+        logger.info("Cleaning Stale Import Jobs ...");
```
wetneb: I would rather not put logging here, as it is going to be displayed every time this is run (every 10 minutes) even if there is no job to clean:

```
18:10:08.083 [                importing] Cleaning Stale Import Jobs ... (410217ms)
18:20:08.083 [                importing] Cleaning Stale Import Jobs ... (600000ms)
18:30:08.083 [                importing] Cleaning Stale Import Jobs ... (523849ms)
```

PR #1710, opened 2018-08-15
Improve facet expression for type match.

### LINKED ISSUE #1667: Reconciled cells showing as (unreconciled) in facet
**Describe the bug**
After reconciling cells, a "best candidate's types match" facet shows them as (unreconciled)

**To Reproduce**
I'm not completely sure what behaviour led to these particular cells showing as "unreconciled" - it doesn't happen to all cells. 

I think, but not 100%, that this has occurred where after the initial reconciliation step there are no matches at all, and then I've used the "Search for match" option to find the correct match

**Current Results**
screen shot 2018-06-29 at 11 09 46

**Expected behavior**
I would expect all the cells shown here as '(unreconciled)' in the facet to be showing as either 'true' or 'false'

 - OS: MacOS High Sierra
 - Browser: Safari

 - Version: OpenRefine 3.0 Beta

wetneb:
Closes #1667.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
+++ b/main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
@@ -272,7 +272,7 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
                 {
                   "name" : column.name + ": "+$.i18n._('core-views')["best-cand-type-match"],
                   "columnName" : column.name, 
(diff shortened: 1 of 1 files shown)


wetneb:
Merging this in a few days unless anybody sees any issue.

thadguidry:
@wetneb The recon null check is what we wanted, so LGTM for merge

wetneb:
Thanks!

PR #1721, opened 2018-09-01
fix Windows typo and clarify Apache Ant

thadguidry:


README.md
```diff
@@ -16,8 +16,8 @@ Run from source
 ------------------
 If you have cloned this repository to your computer, you can run OpenRefine with:
 * `./refine` on Mac OS and Linux
-* `refine.bat` on Window
-This requires JDK 8 and Ant.
+* `refine.bat` on Windows
+* This requires JDK 8 and Apache Ant.
```
wetneb: Actually, Ant should not be required at all, because the `refine.bat` script should download it by itself. This would be worth checking though.

Issue #1652, opened 2018-06-15
Replace ORG.JSON library


magdmartin:
Because of the [extra line in the JSON.org license](https://github.com/OpenRefine/OpenRefine/blob/dc6573d17b8fa35ca592f8441e4aebb73fe3b675/licenses/json.LICENSE.txt#L13), it is considered as non-free license by GNU/FSF, Debian, Fedora, Red Hat and Google and is causing issue to join the SFC. 

Alternative free license exist to avoid this issue. See for example: https://wiki.debian.org/qa.debian.org/jsonevil

@thadguidry  is also looking at Jackson library

thadguidry:
@magdmartin No, I ALREADY looked at Jackson and it's usable, but quite a bit of work for @jackyq2015 or whomever, like... A LOT of work actually :-)

wetneb:
Migrating out of this library is a lot of work that is going to cause major incompatibility with most extensions. It's something that we want to do eventually, but it's most likely going to happen gradually. It could be possible to migrate gradually, by isolating the parts that use Jackson / JSON: for instance, the Wikidata extension uses Jackson everywhere except where it interfaces with the core.

magdmartin:
I am adding this ticket to the front/end back end separation so we can take it into account when we define the technical roadmap.

ostephens:
I guess we need to discuss exactly how we are going to approach this? Can it be broken down into smaller tasks that less experienced developers (like me!) can tackle?

thadguidry:
@ostephens I would probably begin looking at this through an Interface lens.  So that TDD (Test Driven Development) is approached from the very beginning.  We actually have an Interface: 

com.google.refine.Jsonizable
who's consumers implement it, and would be the first test points :

com.google.refine.model.AbstractOperation
com.google.refine.model.Cell
com.google.refine.clustering.Clusterer
com.google.refine.model.Column
com.google.refine.model.ColumnGroup
com.google.refine.model.ColumnModel
com.google.refine.commands.Command
com.google.refine.commands.browsing.ComputeClustersCommand
com.google.refine.commands.browsing.ComputeFacetsCommand
com.google.refine.grel.Control
com.google.refine.browsing.DecoratedValue
com.google.refine.browsing.Engine
com.google.refine.expr.EvalError
com.google.refine.browsing.facets.Facet
com.google.refine.grel.Function
com.google.refine.commands.history.GetHistoryCommand
com.google.refine.commands.history.GetProcessesCommand
[...]

PR #1722, opened 2018-09-01
Add serialization tests for most Jsonizable classes

wetneb:
This adds tests to pin the serialization format of most classes that implement `Jsonizable`. A few of them are not covered, for instance the clustering classes, because I want to do some refactoring on them before (separating the configuration data from the actual operation). This will ease the migration to Jackson.

See roadmap at #1652.


wetneb:
Whoops, my rebasing introduced some build errors, I'll fix that in the coming days.

wetneb:
Merging this in a few days unless anybody sees any issue.

PR #1729, opened 2018-09-10
Store engine configuration in dedicated class

wetneb:
The configuration of the "engine", which determines which rows or records are visible through facets, is currently stored in a `JSONObject` instance. This is bad practice, as it delays parsing checks and mixes them with application logic. This PR introduces an `EngineConfig` class that represents the same information. It is parsed from JSON and can be serialized back.

This refactoring will make it easier to migrate to Jackson (#1652), by reducing the explicit dependencies on the underlying JSON library.


thadguidry:
WOW NICE ONE @wetneb  !  You get a virtual bonus cookie today ! 🥇

wetneb:
Merging this now that 3.0 is released.

PR #1730, opened 2018-09-10
Refactor clustering serialization in JSON

wetneb:
This PR is similar to #1729 in that it refactors serialization with the migration to Jackson in mind (#1652).

New classes are introduced to store the configuration of each clusterer. These classes will be improved and simplified once we migrate to Jackson.


wetneb:
Merging this now that 3.0 is released.

PR #1731, opened 2018-09-12
Custom column names for separator based importer

xseris:
Sometimes there is the need to change the column names of an importing separator based file or to add them because they are missing (without renaming individually once the file have been imported).

This option allows to define a comma separated string that will be parsed as the column names for the project. The option is only available when the "Parse next" checkbox is not selected. Selecting the "Parse next" checkbox will disable and clear the column names input box:

[demo](https://ibb.co/gxk1aU)

### DIFF
--- a/main/src/com/google/refine/importers/SeparatorBasedImporter.java
+++ b/main/src/com/google/refine/importers/SeparatorBasedImporter.java
@@ -101,6 +101,28 @@ public class SeparatorBasedImporter extends TabularImportingParserBase {
         boolean processQuotes = JSONUtilities.getBoolean(options, "processQuotes", true);
         boolean strictQuotes = JSONUtilities.getBoolean(options, "strictQuotes", false);
         
+        
+        List<Object> retrievedColumnNames = null;
+        if (options.has("columnNames")) {
+          String[] strings = JSONUtilities.getStringArray(options, "columnNames");
+          if (strings.length > 0) {
+            retrievedColumnNames = new ArrayList<Object>();
+            for (String s : strings) {
+              s = s.trim();
+              if (!s.isEmpty()) {
+                retrievedColumnNames.add(s);
+              }
+            }
+
+            if (retrievedColumnNames.size() > 0) {
+              JSONUtilities.safePut(options, "headerLines", 1);
+            } else {
+              retrievedColumnNames = null;
+            }
+          }
+        }
+        final List<Object> columnNames = retrievedColumnNames;
+        
         Character quote = CSVParser.DEFAULT_QUOTE_CHARACTER;
         String quoteCharacter = JSONUtilities.getString(options, "quoteCharacter", null);
         if (quoteCharacter != null && quoteCharacter.trim().length() == 1) {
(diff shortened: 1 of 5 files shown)


thadguidry:
@xseris This doesn't work for me...

1. Use clipboard to define data
2. with the following data:
```
a,b,c
d,e,f
```
3. Choose CSV/TSV importer
4. Use the column names:
```
1st,2nd,3rd
```
5. Preview does not update with new column names.
6. Click CREATE button and also no column names are renamed from defaults.

xseris:
@thadguidry I think the problem is that there is no auto-update on form change. If you click outside the  form, or the update preview button, column names should change. Can you confirm this?

xseris:
@wetneb you see it disabled because the "Parse next" checkbox is checked. The logic is either you parse next n lines as header or specify a custom one.

fpompermaier:
Hi @wetneb , I don't think this is a fair review..we use it A LOT and it's super useful..also because there's no an easy way to rename all columns at once! We just need to fix the option button labels (that 
in our version are close to the circle) but for the rest I think this simple feature is very useful

wetneb:
@fpompermaier @xseris then I would be more in favour of a command to rename all columns at once. This would have the advantage of being usable from all data sources, not just CSV/TSV, and appear as an operation in the JSON workflow. I understand this is useful for your own workflows but we should also keep the importer UI simple and principled.

Also I don't find it very clean to expose a simple text field to input a list, by expecting the user to separate the items with a designated, non-configurable separator.

That's just my intuitive reaction, it's not a hill I will die on!

fpompermaier:
I will list my thoughts on it:
1. The fixed length importer suffer of the same identical problem (you have to provide a comma separated list of field names)
2. Also if you rename all columns at once you have to provide the comma separated list of fields
3. I don't think that producing an header as a separated list of column names is a very big problem (at least for Refine users) also because it remains optional
4. The new option doesn't occupy "a lot of space" at all...in the current Refine version that space is just empty (there are check boxes on the right side that go far below the new option)

wetneb:
"you rename all columns at once you have to provide the comma separated list of fields" -> why is that? I don't think we have an operation for that yet, so are you saying that any UI for such an operation would use this input method? Surely we could do something nicer.

xseris:
@wetneb I think one feature does not exclude the other. We can have a custom column name on different importers on project creation time, and an option to bulk change column names in the project page itself, if necessary.

Agree that changing column names can be done in a prettier way, but usually char-separated strings are the faster way to bulk define or change column names of datasets with lots of fields. And, in my opinion, a char-separated string is also more reusable when datasets are similar.

wetneb:
Okay, why not then. I think the UI is still not quite ready though. Here are some comments on the layout:

image

Although I do understand your logic to disable the field, it will probably confuse other users. How about adding a checkbox in front of the option (just like the other options) - checking this box would disable the one for "parse next n lines" and conversely?

Also, we would need server-side unit tests.

thadguidry:
I am OK with having the feature and understand it from a users perspective.
I also agree with @wetneb that this could be done better in the UI... BUT for now, this can work, and in the future, we can have a dialog to make this prettier.

TODO:
1. Shrink the length of the "Column names" input box 50% at least.
2. remove the right side label of "comma separated" and instead just relabel above the input box with "Column names (comma separated)".

xseris:
Hi @thadguidry. Following @wetneb guidelines I've already moved all stuff on one single line and enabled/disabled options properly. This last commit produced the single "Column names (comma separated)" label. Check the box size if it is good for you.

thadguidry:
@xseris move the input box BELOW the label (instead of on the right side currently)

xseris:
@thadguidry now input area is below the label. @wetneb said to keep it on the same line, it would be good if you decide :smile:

wetneb:
@thadguidry let's avoid giving orders in CAPS to our kind contributors, especially when they are contradictory with previous suggestions :)

thadguidry:
@xseris does Update Preview not show the renamed columns ?  Could it ?

xseris:
@thadguidry What do you mean exactly? When the column names input changes, the preview automatically updates. there is no need to click the "Update Preview" button.

PR #1733, opened 2018-09-13
Fix mixup of dialog titles

### LINKED ISSUE #1657: The column selection window of "Transform all" incorrectly labeled "SQL exporter"

Version 3 beta

Windows 10

All

Click on "All", then "Transform", then "OK"

#### If you are allowed and are OK with making your data public, it would be awesome if you can include the data causing the issue or a URL pointing to where the data is (if your concerned about keeping your data private, ping us on our mailing list):
Any dataset

#### Current Results:
screenshot-127 0 0 1-3333-2018 06 19-12-04-13

#### Expected Results:
A column selection window labeld "Select the columns to transform", something like that.

wetneb:
Closes #1657.


ostephens:
I think https://github.com/OpenRefine/OpenRefine/blob/9f7d5b878628c40d68a445b751fd50140d922e7d/main/webapp/modules/core/scripts/dialogs/expression-column-dialog.js#L34 needs changing as well and another new lang string needs creating for that to full fix #1657 ?

wetneb:
@ostephens makes sense!

PR #1744, opened 2018-09-23
Add menu for searching and replacing

msaby:
Add a new menu "Replace" in "Edit Cells" > "Common Transforms" and a new interaction box for entering parameters.
Under the hood this is a wrapper around GREL replace() function.
The behavior is the same as replace() except the automatic backslah escaping of "Replace with" parameter.
The 2 main parameters in the box are 2 textaras : "Find" and "Replace with"
2 cumulative options are available for "Find" parameter : "Case insensitive" and "Regular Expression"
1 option is available for "Replace with" parameter : "Do not escape backslash automatically"
A validity control of the pattern is made when "Regular expression" is checked.


msaby:
The UI must look like that :
capture d ecran 2018-09-23 a 20 35 15
[clipboard.csv.zip](https://github.com/OpenRefine/OpenRefine/files/2409110/clipboard.csv.zip)

To test :
Try to search & replace strings and patterns and play with the options
 I join a csv file with some weird strings, but you can craft your owns.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -674,7 +675,17 @@
...
         "records": "records",
         "show": "Show",
-        "hide": "Hide"
+        "hide": "Hide",
+        "text-to-find": "Find:",
+        "replacement-text": "Replace with:",
+        "case-insensitive": "case insensitive",
+        "finding-info1": "Leave blank to add the remplacement string after each character.",
+        "finding-info2": "Check \"regular expression\" to find special characters (new lines, tabulations...) or complex patterns.",
+        "finding-options": "Finding options:",
```
thadguidry: I'd remove the "finding-options" and "replacement-options" to save space.  It's self evident already they are options.

thadguidry: Also, I would put this under the "Cluster and edit..." on 1st menu.

msaby:
ok for me. I updated my branch https://github.com/msaby/OpenRefine-devms/commits/replaceGUI
Can you tell me if I need to update the pull request? or is it automatic?

msaby:
thanks @ostephens . So you can test with the new commit and see if it works as planned.

msaby:
Managing the escaping of newlines and tabs in textareas gave me a headache. So Ok for a single line input.
I propose to keep the escaping I made for both fields.
I the user need to search for a tab or a newline, he must check "regular expression", and use \n and \t
I must try to find something less obscure than "do not escape backlash automatically". Maybe "interpret \n as newline, \t as tab, \\ as \"

msaby:
Maybe something like the options in the search & replace box of Notepad++ would be nice ?
image

In Notepadd++ you can choose to find whole words only (could be implemented by using regexp under the hood and adding \b before and after the word).
You can also use escaped characters like \n \t but without using regexp (list http://docs.notepad-plus-plus.org/index.php/Searching_And_Replacing ). If you don't check the box, "\n" is interpreted as 2 characters : \ and n

msaby:
I added a followup commit with the requested changes, and I added a new option for matching whole words (like in Notepad++ and Atom), and changed the behavior of the escaping option for replacement text.
Could you test it please?
It seems a conflict happened with translation-en.json file. Should I do something ? I don't know how to manage it...

Issue #1717, opened 2018-08-27
Nothing is on the Web Page


wentianq:
**Describe the bug**
When OpenRefine is running, my browser at http://127.0.0.1:3333/. However, nothing is on the web page except an icon

**Screenshots**
![image](https://user-images.githubusercontent.com/42724736/44634755-e8f89480-a96b-11e8-8a7d-b677d3220e81.png)
![image](https://user-images.githubusercontent.com/42724736/44634756-eeee7580-a96b-11e8-9c1e-15411736265a.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome

**OpenRefine (please complete the following information):**
 - Version: OpenRefine 3.0

wetneb:
Hi! Thanks for the report. Can you open the javascript console in your browser and check if there are any errors there?

wentianq:
Thx for your reply. Here is the console. Can you please help me to fix these error?
![image](https://user-images.githubusercontent.com/42724736/44856857-7e31ac80-ac3c-11e8-927e-c56e7b9cba12.png)

wetneb:
Can you click on the `index-bundle.js:63726` link and post another screenshot of what you see?

wetneb:
can you scroll down to line 63726 and see what is there?

wetneb:
@wentianq which version of Chrome are you using and have you tried updating it?

wentianq:
It is version 68.0.3440.106.

wetneb:
Ok then I really don't know - have you added any extension or is it just the default OpenRefine package?

wentianq:
It is the default package.

ostephens:
Do you have any extensions installed in Chrome? If so can you try turning these off and seeing if that helps? Or trying in another browser?

wentianq:
After I forbade the extensions in Chrome, it still didn't work. I also tried IE, but I got the same result.

jackyq2015:
could you please provide:
- the exact version you downloaded
- windows version
- Java version

Another thing you can try is to switch to another user which has no space in between

wentianq:
- Version Downloaded: openrefine-win-3.0-rc.1
- Windows Version: Microsoft Windows 10 Home
- Java Version: 10.0.2

jackyq2015:
[...]

PR #1745, opened 2018-09-25
force the encoding to UTF8

jackyq2015:
Fix issue #1717. Solution is to always read the file in UTF8. Otherwise it will cause problem on windows 10 and updated chrome.


thadguidry:
@jackyq2015 interesting...how did you know WHAT to fix based on what was detailed in #1717 ?  We had never had issues before and placing this on controller.js seems wrongly placed...but perhaps I am missing what your Root Cause Analysis has discovered... can you share those details ?

thadguidry:
Is it only with Chrome or any browser on Windows 10 ?

jackyq2015:
For any browsers under 10

PR #1754, opened 2018-10-02
Directly use column as reconciliation results

### LINKED ISSUE #1596: Directly use column as reconciliation results
I have not been able to find a way to directly use a column as reconciliation results. In the case of Wikidata, reconciling a column of item IDs works (I assume it’s just due to the fact that `wbsearchentities` supports searching for the item ID), but it takes an unnecessarily long time.

This would be very useful for a kind of workflow where I don’t want to work with the dataset in OpenRefine all the time as one monolithic project, but rather prefer to use plain old text files and associated Unix tools (`cut`, `join`, `awk`, …) as much as possible, and exchange parts of the data with OpenRefine as needed – see https://github.com/lucaswerkmeister/bundestag/ for example (most of the details are in the commit messages).

wetneb:
This adds a new operation to reconcile a column of identifiers without hitting the reconciliation service. The operation can be found in the `Reconcile...` menu of each column.

This closes #1596.

### DIFF
--- a/main/src/com/google/refine/commands/recon/ReconUseValuesAsIdentifiersCommand.java
+++ b/main/src/com/google/refine/commands/recon/ReconUseValuesAsIdentifiersCommand.java
@@ -1,0 +1,31 @@
+package com.google.refine.commands.recon;
+
+import javax.servlet.http.HttpServletRequest;
+
+import org.json.JSONObject;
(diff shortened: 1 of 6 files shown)


wetneb:
Merging this in a few days.

PR #1755, opened 2018-10-02
Jackson serialization

wetneb:
This PR is the second step of the migration out of org.json (#1652). It adds Jackson configuration to many classes so that they are serialized correctly with Jackson.

<s>Deserialization is unchanged for now.</s> Edit: doing the deserialization in the same PR.

This PR is a work in progress. **Make sure you test this on a disposable workspace - there is A LOT of potential for data corruption in this PR.**

See https://github.com/OpenRefine/OpenRefine/wiki/Migration-guide-for-extension-and-fork-maintainers#migrating-from-orgjson-to-jackson for migrations instructions for extensions and forks.


wetneb:
This PR should now be ready to be merged. This just one part of the migration, with one first milestone reached: we do not import `JSONWriter` anymore (except in the broker, but we will delete it as it is no longer used).

This move to Jackson relied heavily on all the tests introduced in #1730, #1729 and #1722 and introduced many other tests: these tests were validated against the org.json implementation and were used in turn to validate the Jackson implementation.

That being said, a lot of the changes in this PR are not covered by tests: these were mostly translations of JSON production in commands. This is simply because these translations were fairly straightforward compared to the rest, less critical (because if they introduced some issue it is less likely to cause data corruption). Writing tests for all these would basically amount to writing tests for the entire code base from scratch, which is not doable within our time budget.

The next step of this migration will be the conversion of deserialization methods. This is also a very significant step and should take about the same amount of time.

jackyq2015:
@wetneb Seems there are some compilation error for testing. could you please take a look?

wetneb:
@jackyq2015 indeed, this was introduced by my rebase…

**By the way, make sure you test this on a disposable workspace - there is A LOT of potential for data corruption in this PR.**

PR #1760, opened 2018-10-13
fix potential POI resource leak not closing Excel workbooks

thadguidry:


wetneb:
@thadguidry I am a bit puzzled by this… Why would we close a resource just after it is created? It looks like you are trying to solve a warning from some tool, maybe? Or did you run into issues yourself when using the exporter?

thadguidry:
@wetneb sorry, yes, I assumed to much.  Closing.

Issue #1674, opened 2018-07-07
Allow smartSplit() to handle String of chars, not just a single char


thadguidry:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
[smartSplit()](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/expr/functions/strings/SmartSplit.java) does not currently handle more than 1 separator character to split on.  

**Describe the solution you'd like**
allow [smartSplit()](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/expr/functions/strings/SmartSplit.java) to take a full string of characters to split on. And have it work like the other split functions and allow sep to be a String and not just charAt(0) , and keeping its "smarts" with special handling of quotes and guessing tabs or commas if sep String is not supplied.

**Describe alternatives you've considered**
none

**Additional context**
 Comment from issue #1586 
@ostephens Its a smartSplit() issue.  Looks like David Huynh never went in and added the multichar support (I checked the code and history in link below) and so it is only looking at a single char for the sep String instead of any number of characters for sep.  This was added as an enhancement for split() and also on the menu for "Split into several columns..." and the reason I know this is because I asked @dfhuynh to change that functionality back in 2010, which he did.  BUT...doesn't look like we did that for smartSplit() .... so yeah... [smartSplit() only takes the 1st character in the sep argument supplied and ignores any other characters in the argument.](https://github.com/OpenRefine/OpenRefine/blob/b1bfcb586cb601d6666eb1bae727a6a362c2f5d5/main/src/com/google/refine/expr/functions/strings/SmartSplit.java#L79)

Notice here it splits on everything with "r" ... but not my supplied "rd"
I'd say that's a enhancement from 2010 that I asked for that never happened in smartSplit()  😀 and not a classic bug 😄 
![image](https://user-images.githubusercontent.com/986438/39341070-034bd778-4997-11e8-9f10-d961952dd1b2.png)

So your Store Array functionality is fine.  And works great from my testing on all sides, so far.

Issue is our inaccurate documentation and description of smartSplit() saying it works with a sep String, when it fact it only works with a sep Char ... and make a change to have it work like the other split functions and allow sep to be a String and not just charAt(0) , and keeping its "smarts" with special handling of quotes and guessing tabs or commas if sep String is not supplied.

@ostephens @ettorerizza @wetneb If you agree on above enhancement with smartSplit() then create new issue with above info copied in.

[...]

PR #1761, opened 2018-10-14
Fix: Allow smartSplit() to handle String of chars not just a single char

omkarnix:
This commit adds the support of String separator in function smartSplit. Previously only char separator is supported. Now smartSplit can support String as well as single char as separator. Changed to existing implementation of overloaded constructor of CSVParser which takes string separator.

### DIFF
--- a/main/tests/server/src/com/google/refine/tests/expr/functions/strings/SmartSplitTests.java
+++ b/main/tests/server/src/com/google/refine/tests/expr/functions/strings/SmartSplitTests.java
@@ -1,15 +1,100 @@
 package com.google.refine.tests.expr.functions.strings;
 
+import static org.junit.Assert.assertArrayEquals;
+import static org.testng.Assert.assertTrue;
+import java.util.Properties;
+import org.testng.annotations.AfterSuite;
+import org.testng.annotations.BeforeSuite;
 import org.testng.annotations.Test;
 
+import com.google.refine.expr.EvalError;
 import com.google.refine.expr.functions.strings.SmartSplit;
+import com.google.refine.grel.ControlFunctionRegistry;
+import com.google.refine.grel.Function;
 import com.google.refine.tests.util.TestUtils;
 
 public class SmartSplitTests {
+
+    private static Properties bindings;
+    private static String FUNCTION_NAME = "smartSplit";
+
+    @BeforeSuite
+    public void setUp() {
+        bindings = new Properties();
+    }
+
+    @AfterSuite
+    public void tearDown() {
+        bindings = null;
+    }
+
+    @Test
+    public void testSmartSplitInvalidParams() {
+        assertTrue(invoke(FUNCTION_NAME) instanceof EvalError);
+        assertTrue(invoke(FUNCTION_NAME, "teststring1", 1, "teststring2", 2) instanceof EvalError);
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


wetneb:
Hi @omkarnix, thanks a lot for this pull request! As discussed in the issue it would be great to have a few tests for this (as you have noticed the test coverage is really poor at the moment so we are trying to improve that for all new changes). Let me know if you need any help.

omkarnix:
@wetneb, Sure Thanks for reviewing. Actually, I realized that low test coverage after creating PR and started the discussion on https://github.com/OpenRefine/OpenRefine/issues/1674. Will amend my PR with tests tomorrow and update! Thank you for quick responses and encouragement!

main/tests/server/src/com/google/refine/tests/expr/functions/strings/SmartSplitTests.java
```diff
@@ -1,15 +1,100 @@
...
+        Function function = ControlFunctionRegistry.getFunction(name);
+        if (function == null) {
+            throw new IllegalArgumentException("Unknown function " + name);
+        }
+        if (args == null) {
+            return function.call(bindings, new Object[0]);
+        } else {
+            return function.call(bindings, args);
+        }
+    }
```
wetneb: Fantastic, thanks a lot! As a small note, I think `invoke` should not need to take the function name as parameter, as it is always `FUNCTION_NAME` in this class (and the method is private). But that is a cosmetic concern.

omkarnix: Yeah. Actually I thought we should move the invoke method to some test util class or to probably RefineTest. Actually sorry I forgot to extend SmartSplitTest class by RefineTest. I think I should probably add that code right away.

wetneb: @omkarnix moving to a test util class would make a lot of sense too.


The `RefineTest` class is used to create a test workspace and initialize various other things that are required when dealing with projects. For function tests it is not needed to extend it, as function registration is done statically as you have rightly noted.

omkarnix: Ok. Thanks. I would certainly like to refactor it and move the invoke method to some util class. 
I noticed that travis CI build failed due to assertion failure in serializeProcessManager whereas appveyor build succeeded. Could it be a glitch while building on travis-ci? Can we trigger the build manually?

omkarnix: > Fantastic, thanks a lot! As a small note, I think `invoke` should not need to take the function name as parameter, as it is always `FUNCTION_NAME` in this class (and the method is private). But that is a cosmetic concern.

@wetneb I agree. It was really not required to create constant for the local static value. So made the change suggested by yo .Also added one more misc change. Can you review it?

wetneb: yes don't worry about unrelated test failures (although this one is curious). I can restart the build when it happens.

omkarnix:
@wetneb Thank you very much for your support and encouragement on contribution. The community is very supportive. I would certainly try to contribute as much as possible!

Issue #1757, opened 2018-10-08
options are not getting applied in create-project-from-upload API


pranavjadhao:
I am trying to use "POST /command/core/create-project-from-upload" API from java client.
For this, I want to specify "options" as a form data rather than using the default options.
I am passing the "options json" as form data but It is not getting applied and I am still getting the output  with the default options.
It is happening for both csv and json data. Due to this issue I am not able to create project with json data as I am not able to define recordData.

wetneb:
@pranavjadhao our API is poorly documented and not very consistent - apologies for that!
Are you calling the API over HTTP from a Java client or using OR as a library that you call in Java directly?

Have you tried inspecting the shape of the requests made in the browser and replicating these manually?

pranavjadhao:
![options_not_applied](https://user-images.githubusercontent.com/11486559/46658103-8849b200-cbcf-11e8-84f1-3a55f6811f4d.jpg)

@wetneb Please refer the above image. Here I have imported a csv file with default options, but the output was not as expected so I have changed the options and the preview is showing correct results.

After that I have tried same by calling the API through "postman" with the required form data (with modified options) as per the documentation. But still I am getting the table as if the default options were applied.

Please let me know if I am doing something wrong here or is it the problem with API?

wetneb:
@pranavjadhao it seems that you have correctly copied the POST data, but what about the GET parameters?

pranavjadhao:
@wetneb Sorry I couldn't get it. Which GET parameters you are talking about?

wetneb:
@pranavjadhao the HTTP request that you are trying to replicate uses not only POST data but also GET data. You should make sure you provide both.

pranavjadhao:
![image](https://user-images.githubusercontent.com/11486559/46664783-9e5f6e80-cbdf-11e8-9c62-bfd8ea3782eb.png)

@wetneb I am unable to find the request similar to "create-project-from-upload" in above log. can you please point me to correct request which should be referred. or at least the GET parameters which should be passed.

ancore:
The options parameter is taken from the URL GET parameters and ignored in form fields.

CreateProjectCommand Line 139:
`String optionsString = request.getParameter("options");`

I think the line should rather be:
`String optionsString = parameter.getproperty("options");`

wetneb:
@ancore ah great - do you want to open a PR for that?

ancore:
Sure. I think I need permissions to create a branch for this.

[...]

PR #1764, opened 2018-10-16
Using the parsed form parameter 'options' instead of looking for an U…

ancore:
This fix solves [Issue #1757](https://github.com/OpenRefine/OpenRefine/issues/1757).

### DIFF
--- a/main/src/com/google/refine/commands/project/CreateProjectCommand.java
+++ b/main/src/com/google/refine/commands/project/CreateProjectCommand.java
@@ -106,7 +106,7 @@ public class CreateProjectCommand extends Command {
             }
             
             JSONObject optionObj = null;
-            String optionsString = request.getParameter("options");
+            String optionsString = parameters.getProperty("options");
             if (optionsString != null && !optionsString.isEmpty()) {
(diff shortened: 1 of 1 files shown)


wetneb:
@ancore great, thanks a lot! Sorry for the delay in reviewing. Ideally it would be useful to have a unit test demonstrating that the options are recognized in the exporter.

nsiebler-dm:
Any possibility to finalize this? Would be much appreciated.

ostephens:
@ancore @nsiebler-dm we'd really prefer to have a unit test which checks for the correct behaviour alongside this change - we always try to add tests when we change code because we have had a history of poor test coverage.

Is there any chance tests could be added to this PR?

wetneb:
@ostephens makes a lot of sense! Merging this then.

thadguidry:
@ostephens Correct, more correctly, Stefano added it for Windmill-based UI testing.  Command used on L259 within...
https://github.com/OpenRefine/OpenRefine/blame/95d0cf546d37dc81c493c9ef09e946a7547cd1ba/refine#L256
Original commit here: https://github.com/OpenRefine/OpenRefine/commit/6dbe794658feae7acda9e29a45315b42ece0fa47

PR #1803, opened 2018-11-03
Fix cell index computation in filldown/blankdown

wetneb:
So the fill down and blank down incorrectly assumed that the first column had cell index 0, which gave the weird bugs observed in #1573. This should fix it.

@msaby can you confirm this fixes the issue on your side?

### DIFF
--- a/main/src/com/google/refine/operations/cell/BlankDownOperation.java
+++ b/main/src/com/google/refine/operations/cell/BlankDownOperation.java
@@ -103,6 +103,7 @@ public class BlankDownOperation extends EngineDependentMassCellOperation {
         
         return new RowVisitor() {
             int                 cellIndex;
+            int 			    keyCellIndex;
             List<CellChange>    cellChanges;
             Cell                previousCell;
             Mode                engineMode;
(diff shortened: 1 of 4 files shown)


wetneb:
@msaby hmmm that is annoying! Have you tried doing `./refine clean` first?

msaby:
Thanks, it worked!

So, I suppose it's ok :
From this project with 14 lines and 3 columns, with "x" in the 3rd column to force OR to keep the record number at 3
image

I can get 3 records, and apply the blanking down on column 2. In row mode it works like before

image

But in record mode the operation is applied on each record
image

Now I try the filling down, after editing manually some cells

image

It works well : 
image


So the only disturbing point is the behavior of record mode when a row contains only null values: a record is created for that row. But I believe this behavior is not new, even if I discover it today... So this has to link with your patch. Can you confirm it?

image

wetneb:
@msaby yes we would need to change the way empty rows are treated in records (which I have not touched neither in this PR nor in #1573). It's definitely something we can consider, but I maybe not do this in a rush before 3.1 beta.

wetneb:
@thadguidry yes I agree it is annoying, I will look into that.

PR #1806, opened 2018-11-05
Streamline Eclipse setup with Maven

### LINKED ISSUE #1805: Add maven goal to run OR
The Maven configuration files used by Eclipse do not contain any goal to actually run the software, which means that the developer must configure Eclipse manually to run OR with the right arguments. We should add a Maven goal to run OR.

We might not want `./refine` and `refine.bat` to use the Maven goal to run OR as it would require shipping Maven in the packaged versions.

wetneb:
This tweaks the Maven configuration files so that they are better accepted by Eclipse.
To run OpenRefine from Eclipse, add a new maven execution goal profile with "exec:java" as goal.
It is also possible to run OpenRefine from command line with `mvn compile test-compile exec:java`.

This closes #1805.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ If you have cloned this repository to your computer, you can run OpenRefine with
 * `./refine` on Mac OS and Linux
 * `refine.bat` on Windows
 
-This requires JDK 8 and Apache Ant.
+This requires JDK 8 and Apache Maven.
 
 Documentation and Videos
 -------------------------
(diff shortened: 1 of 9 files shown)


wetneb:
I have started to update the docs (help welcome!):
https://github.com/OpenRefine/OpenRefine/wiki/Building-OpenRefine-From-Source

wetneb:
Merging per confirmation on the mailing list that it works.

Issue #1812, opened 2018-11-06
Longer dialogs like "Select Encoding" get cutoff on bottom initially


thadguidry:
Longer dialogs get cutoff on smaller screens.  I saw this caused issues for users sometimes, like she had during the GLAM Wikidata 2018 at this time point https://youtu.be/VOO8IS73Cq0?t=20597

To help a bit, we should initially set to 10% from Top, instead of our current vertical centering via element height approach in `dialog.js`
```javascript
//  container.css("top", Math.round((overlay.height() - elmt.height()) / 2) + "px");
  container.css("top", "10%");
```

PR #1813, opened 2018-11-06
Make dialog containers start from 10% from top

thadguidry:
...instead of our current vertical centering approach, which sometimes cuts off longer dialogs like "Select Encoding" on users with smaller screen sizes that sometimes occur during projecting for training.

Fixes issue #1812


wetneb:
Thanks, well spotted! It's massively useful to look out for bugs like this one in the wild.

In general, vertical centering (if done correctly) is the right thing to do. This particular dialog is just not vertically centered at the moment, because of the fiddling with the visibility of the second tab in this dialog. I think we should just fix that fiddling and restore the two-tab interface that it was originally designed for.

thadguidry:
Fixed now by merged PR #1814

PR #1814, opened 2018-11-06
Restore tab behaviour in encoding dialog.

### LINKED ISSUE #1812: Longer dialogs like "Select Encoding" get cutoff on bottom initially

wetneb:
Alternate way to fix #1812.


thadguidry:
...testing it now...

Issue #608, opened 2012-10-15
Support of conversion of millisecond Unix time to Refine datetime


tfmorris:
_Original author: to...@benmoshe.com (September 15, 2012 11:52:15)_

It would be nice to have a function such as FORMAT_UTC_USEC to convert Unix time ( which is a number) to a Datatime format, and to extract only year/month/day

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=609_

thadguidry:
For whomever works on this... Here's a good primer, since OpenRefine now uses java.time.offsetdatetime https://stackoverflow.com/questions/21242110/convert-java-util-date-to-java-time-localdate

Our code for the GREL toDate() function is here: https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/expr/functions/ToDate.java

ostephens:
So I'm currently looking at toDate as per #1759 and @thadguidry reminded me of this (thanks Thad!).

At the moment I can think of two options for handling this:

Extend toDate to have a GREL specific 'format' option which is (something like) 'epochms' (epoch milliseconds) which tells toDate to use a long number (or a string which can be converted to a long number) as millisecond Unix time and use that to generate a date/time

The alternative I can see is to have a special new function such as 'epochToDate' which does this specifically.

The second approach avoids making the toDate function even more complex than it already is, but has the downside of introducing a new function rather than a single 'toDate' function handling all conversions to a date.

@thadguidry do you have thoughts on this?

thadguidry:
Summary:
I think I would choose the format option to make it painless for our users (we do the hard things, to make it easier for them). 

Errata:
While it is painful and complex within our toDate() function, this is necessary to keep GREL syntax nice and tidy. This is also how other programming languages handle this convention, like Javascript (Moment.js), Python, Pandas, Ruby.  Our toDate() already has an optional format argument.  Where I could see this leading to is possibly extending to handle a 3rd argument for timezone or utc.

Moment.js does things so beautifully and should be our model https://momentjs.com/docs/#/parsing/unix-timestamp-milliseconds/
Ruby uses time.at  http://ruby-doc.org/core-1.9.3/Time.html#method-c-at
Pandas uses Pandas.Timestamp  https://pandas.pydata.org/pandas-docs/version/0.23.4/generated/pandas.Timestamp.html#

[...]

PR #1826, opened 2018-11-15
Fix toDate

### LINKED ISSUE #1759: Bug in version 3.0 that means toDate() with format parameter returns error
Using toDate on a string like '13/04/2008" with value.toDate("d/M/y") leads to error. On previous versions including 2.8 this correctly converted the string to date. value.toDate() with no parameters appears to work, but in the case above where day is first this is not sufficient.

ostephens:
I've fixed #1759 and refactored code in toDate function. I think the way I've reorganised the code makes it easier to read and understand, but I'd welcome feedback on the way I've approached this.

### DIFF
--- a/main/tests/server/src/com/google/refine/tests/expr/functions/strings/ToFromConversionTests.java
+++ b/main/tests/server/src/com/google/refine/tests/expr/functions/strings/ToFromConversionTests.java
@@ -138,7 +138,7 @@ public class ToFromConversionTests extends RefineTest {
       Assert.assertTrue(invoke("toDate", (Object) null) instanceof EvalError);
       Assert.assertTrue(invoke("toDate", "") instanceof EvalError);
       Assert.assertTrue(invoke("toDate", 1.0) instanceof EvalError);
-      Assert.assertTrue(invoke("toDate", "2012-03-01","xxx") instanceof EvalError); // bad format string
+      //Assert.assertTrue(invoke("toDate", "2012-03-01","xxx") instanceof EvalError); // bad format string
       Assert.assertTrue(invoke("toDate", "2012-03-01") instanceof OffsetDateTime);
       Assert.assertEquals(invoke("toDate", "2012-03-01"),CalendarParser.parseAsOffsetDateTime("2012-03-01"));
       //parse as 'month first' date with and without explicit 'true' parameter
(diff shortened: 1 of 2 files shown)


thadguidry:
@ostephens Shall I test ??

thadguidry:
@ostephens You only have 4 digit year tests.  You should have least this one, since year last is a default for the library, yes/no ?

`"02-02-01".toDate()`  - > `[date 2001-02-02T00:00:00Z]`

ostephens:
Thanks for testing @thadguidry 

Agree we should test:
"02-02-01".toDate() - > [date 2001-02-02T00:00:00Z]
I will add that in.

I don't understand:
"0002-0222-021".toDate() -> [date 0222-02-21T00:00:00Z]
It does work with the current code, but I don't understand why that string be parsed to a valid date?

I think dealing with pre-gregorian calendar dates and dates using AD/BC/BCE are new feature requests and so need new issues describing expected behaviour (btw once we've got this PR done I can look at adding the epoch format as discussed in #608)

ostephens:
@wetneb thanks for the review

thadguidry:
@ostephens sorry I just blast everything in 1 comment.  super busy guys.

main/tests/server/src/com/google/refine/tests/expr/functions/strings/ToFromConversionTests.java
```diff
@@ -138,7 +138,7 @@ public void testToDate() throws CalendarParserException {
       Assert.assertTrue(invoke("toDate", (Object) null) instanceof EvalError);
       Assert.assertTrue(invoke("toDate", "") instanceof EvalError);
       Assert.assertTrue(invoke("toDate", 1.0) instanceof EvalError);
-      Assert.assertTrue(invoke("toDate", "2012-03-01","xxx") instanceof EvalError); // bad format string
+      //Assert.assertTrue(invoke("toDate", "2012-03-01","xxx") instanceof EvalError); // bad format string
```
tfmorris, line 141: Disabling tests or changing expected behavior shouldn't be done as part of a refactoring. The whole point of refactoring is to keep the same behavior.

I've reenabled this test and corrected the code to return an error here.

thadguidry: @tfmorris That's very true.  But to @ostephens credit, he was doing his best here and was leaning on senior developers to help spot things for him.  He has admitted in the past that even now he's still learning things.
Again, we are so grateful to have you around again!

PR #1842, opened 2018-11-21
Flatten translation files

wetneb:
As part of the migration to the jquery i18n plugin by Wikimedia, this flattens the translation files.


wetneb:
In the interest of staying in sync with Weblate and to avoid conflict with other changes I am merging this PR without review - the change has already been discussed in #1285.

PR #1846, opened 2018-11-21
Migrate to Wikimedia's i18n jQuery library

### LINKED ISSUE #1279: Use best practices in i18n

wetneb:
Based on #1285, closes #1279.

This still does not do language fallback yet, because English strings are not loaded by the UI when another language is selected (although the library supports language fallbacks). This can be fixed by explicitly loading English strings but I am still not sure what is the cleanest way to do that. It can be done in another PR.


wetneb:
I have added migration instructions to https://github.com/OpenRefine/OpenRefine/wiki/Migration-guide-for-extension-and-fork-maintainers

wetneb:
Thanks for reviewing! Yes more work will be needed to take advantage of the plurals and interpolation too. Merging this now to avoid further merge conflicts as this impacts most of the javascript files in the code base.

Issue #1816, opened 2018-11-08
Support all date precisions in Wikidata schema


wetneb:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
It is currently impossible to add dates with century precision.

**Describe the solution you'd like**
Add a format (similar to those listed in https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing/Schema_alignment#Dates) to support all precisions allowed by Wikibase.

PR #1860, opened 2018-11-24
Add support for sub-year precisions in Wikidata schema

wetneb:
This adds new date formats for millenia, centuries and decades, so that users can input these in statement values. For #1816.


thadguidry:
@wetneb Since we have `java.time` now ... What about Era and Period ?  Is that important for Wikidata ?

https://docs.oracle.com/javase/8/docs/api/java/time/chrono/Era.html
https://docs.oracle.com/javase/8/docs/api/java/time/Period.html

wetneb:
@thadguidry the Wikibase data model still bases itself on dates and just adds a quite rudimentary notion of precision on top of that:
https://www.mediawiki.org/wiki/Wikibase/DataModel#Dates_and_times

thadguidry:
@wetneb ah ok, good to know.

jackyq2015:
@wetneb Do you know if Wikibase has any plan to migrate to new API

wetneb:
@jackyq2015 do you mean changing their representations of era to match that of Java? I think it would be quite unlikely.

PR #1877, opened 2018-12-02
Fix reconciliation result deserialization

### LINKED ISSUE #1876: Guess reconciliation type command does not work since Jackson migration
This is probably due to invalid queries being made or incorrect parsing of the results.

wetneb:
Closes #1876.

### DIFF
--- a/main/tests/server/src/com/google/refine/tests/model/recon/StandardReconConfigTests.java
+++ b/main/tests/server/src/com/google/refine/tests/model/recon/StandardReconConfigTests.java
@@ -20,6 +20,7 @@ import com.google.refine.model.recon.ReconConfig;
 import com.google.refine.model.recon.ReconJob;
 import com.google.refine.model.recon.StandardReconConfig;
 import com.google.refine.model.recon.StandardReconConfig.ColumnDetail;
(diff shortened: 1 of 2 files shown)


main/tests/server/src/com/google/refine/tests/model/recon/StandardReconConfigTests.java
```diff
@@ -172,4 +173,17 @@ public void deserializeColumnDetail() throws JsonParseException, JsonMappingExce
     	ColumnDetail cd = ParsingUtilities.mapper.readValue(uiJson, ColumnDetail.class);
     	TestUtils.isSerializedTo(cd, backendJson);
     }
+    
+    @Test
+    public void deserializeReconResult() throws JsonParseException, JsonMappingException, IOException {
+    	String json = "{\"score\":100.0,"
+    			+ "\"match\":false,"
+    			+ "\"type\":["
```
jackyq2015: should be "types"?

wetneb: No, the JSON format implement with `org.json` used `type`, not `types`. We need to keep this format so that reconciliation services still work.

PR #1880, opened 2018-12-02
Avoid localizing time for dates with unspecified time.

### LINKED ISSUE #1874: Tests do not pass if the local time is not GMT
**Describe the bug**
The date conversion tests do not pass when run from a timezone other than GMT.

**To Reproduce**
1. Set your timezone to a location outside GMT
2. Run the tests with `./refine test`

**Current Results**
`[ERROR]   ToFromConversionTests.testToDate:164 expected [2012-03-01T00:00Z] but found [2012-02-29T15:00Z]`

wetneb:
Closes #1874.

This effectively disables localization for dates. I think this is the right thing to do, otherwise users might have curious behaviors, such as `datePart(toDate("2012-03-01"), "month")` evaluating to `2` instead of `3`.

### DIFF
--- a/main/src/com/google/refine/util/ParsingUtilities.java
+++ b/main/src/com/google/refine/util/ParsingUtilities.java
@@ -229,7 +229,8 @@ public class ParsingUtilities {
         if (parsed == null) {
             return null;
         }
-        return parsed.toLocalDateTime();
+        return parsed.withOffsetSameInstant(OffsetDateTime.now().getOffset())
+        		.toLocalDateTime();
     }
     
     static public String instantToString(Instant instant) {
--- a/main/tests/server/src/com/google/refine/tests/io/ProjectMetadataTests.java
+++ b/main/tests/server/src/com/google/refine/tests/io/ProjectMetadataTests.java
@@ -2,24 +2,46 @@ package com.google.refine.tests.io;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.TimeZone;
 
 import org.apache.commons.io.IOUtils;
+import org.testng.annotations.BeforeSuite;
 import org.testng.annotations.Test;
 import com.google.refine.tests.util.TestUtils;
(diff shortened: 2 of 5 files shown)


main/src/com/google/refine/util/ParsingUtilities.java
```diff
@@ -229,7 +229,8 @@ static public LocalDateTime stringToLocalDate(String s) {
         if (parsed == null) {
             return null;
         }
-        return parsed.toLocalDateTime();
+        return parsed.withOffsetSameInstant(OffsetDateTime.now().getOffset())
```
jackyq2015: Is it necessary to return another copy of the offsetDateTime with the same instant value? If so, could you please add a UT to prove the previous way does NOT work?

wetneb: Sure, there you go.

jackyq2015: :+1:

main/tests/server/src/com/google/refine/tests/io/ProjectMetadataTests.java
```diff
@@ -2,24 +2,46 @@
...
+	
+	@Test
+	public void serializeProjectMetadataInDifferentTimezone() throws JsonParseException, JsonMappingException, IOException {
+    	TimeZone.setDefault(TimeZone.getTimeZone("JST"));
+    	try {
+	        ProjectMetadata metadata = ParsingUtilities.mapper.readValue(jsonSaveMode, ProjectMetadata.class);
+	        TestUtils.isSerializedTo(metadata, jsonNonSaveMode);
+	        TestUtils.isSerializedTo(metadata, jsonSaveMode, true);
+    	} finally {
+    		TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
```
tfmorris, line 44: This should restore the timezone default to its original value, not set it to an arbitrary fixed value.

Issue #1827, opened 2018-11-16
Filter overrides facet in 3.1 beta


ostephens:
**Describe the bug**
Adding a text filter to a column when you already have a facet selected on a different column leads to the facet selection being reset.

**To Reproduce**
Example data:
Col 1,Col 2
a,b
c,d
e,f

1. Create a facet on Col 1
2. Select 'a' in that facet -> one row shows
3. Creata a text filter on Col 2
4. Type 'd' into the text filter
5. See that the facet is reset

**Current Results**
The filter essentially overrides the fact

**Expected behavior**
That the facet should remain with the same value selected, the filter should add to this, not override it

**Screenshots**
In 3.0 this works (see 'a' is still selected in facet)
<img width="673" alt="screenshot 2018-11-16 at 09 19 36" src="https://user-images.githubusercontent.com/576174/48612486-0a708780-e981-11e8-8cad-84dffdc7f807.png">

In 3.1 the facet is reset - see that 'a' is no longer selected in the facet:
<img width="666" alt="screenshot 2018-11-16 at 09 22 13" src="https://user-images.githubusercontent.com/576174/48612518-1f4d1b00-e981-11e8-97cc-0dd3f4faf9f0.png">

**Desktop (please complete the following information):**
 - OS: macOS Mojave
 - Browser Version: Safari Version 12.0.1 (14606.2.104.1.1)
 - java version "1.8.0_72"

**OpenRefine (please complete the following information):**
 - OpenRefine 3.1 Beta

ostephens:
Haven't had time to check thoroughly, but my guess is that the change in behaviour is related to https://github.com/OpenRefine/OpenRefine/commit/f327b09a596d8723a2915fd53a4d38c1bfc36c40

PR #1889, opened 2018-12-06
Update Butterfly to 1.0.2.

### LINKED ISSUE #1882: Jackson annotations ignored for extensions
Serialization and deserialization of classes provided by extensions currently fail, even if the classes were properly migrated to Jackson and tested. The problem comes from the fact that they are loaded dynamically by Butterfly, by a classloader that does not have access to the definition of the Jackson annotations. The annotations are therefore ignored, so Jackson attempts to deserialize and serialize them without any annotations.

wetneb:
So, the problem did come from Butterfly. Long story short, Butterfly implements its own ClassLoader called `ButterflyClassLoader`, with some custom logic to reload classes on the fly when they are modified.

To do so, the method `loadClass` from the parent class `URLClassLoader` was overridden, with a home-made interpretation of [the delegation model](https://docs.oracle.com/javase/7/docs/api/java/lang/ClassLoader.html) which was incorrect: the problem was that it tried to load a class first by itself before trying the parent loader (whereas the parent loader should be tried first).

This lead to Jackson annotations being marked as loaded by `ButterflyClassLoader` instead of `WebappClassLoader`, so the hashes of the classes would differ and the Jackson introspection code would not find them when generating serializers and deserializers for POJOs coming from extensions.

I have therefore updated https://github.com/OpenRefine/simile-butterfly with a fix on the `openrefine` branch and released a new version of butterfly `1.0.2`. This PR updates OpenRefine to use that new version of Butterfly.

Closes #1882.


jackyq2015:
can not see the butterfly change. where it is?

thadguidry:
> 
> 
> can not see the butterfly change. where it is?

He commited it here: https://github.com/OpenRefine/simile-butterfly/commits/openrefine
Click the hashes to see the changes.
Jacky, you gotta learn Github more :)

capture

capture2

thadguidry:
@wetneb Thoroughly impressed with your detective work !  And you didn't even have to ping Stefano !  Well done mate !

jackyq2015:
I think the purpose of overriding the "loadClass" is to have all the extension has its own set of classes instead of influence each other. Though it fixes the jackson issue, we may need to consider the impact overall. Not sure how many extensions are still in use. My expression is that most of them won't work after our major change recently.

wetneb:
@jackyq2015 ah okay, that was probably the reason why this was overridden, indeed!
Yes I also think most extensions will need upgrading anyway. But if you can think of any other way to fix this I am more than happy to consider anything else!

tfmorris:
Butterfly isn't our project, so I'd recommend that the version numbering make it clear that it's a private fork, in case they ever do a 1.0.2 (unlikely, but still not a good practice to make a possibility -- and yes, looking back at the commit history, I realize I did it wrong 6-7 years ago).

The internal implementation leakage to extensions is a **much** bigger issue, but if most/all of them will be broken by this change, it'd be a good time to put a more formal interface in place so that doesn't happen in the future. The Butterfly class loader is one piece of that, but clearly it's not sufficient.

If an extension gets broken by the JSON serialization change, I'd argue that that indicates that it's not well enough isolated and the solution isn't to make the abstraction even more leaky, but, rather, to force the extensions to provide their own JSON serialization, unless it's a facility provided by OpenRefine through an opaque API.

There's not a lot of context on #1882. Is there an email thread or something else which has more detail?

wetneb:
@tfmorris thanks for jumping in… here are a few answers:

> Butterfly isn't our project, so I'd recommend that the version numbering make it clear that it's a private fork, in case they ever do a 1.0.2 (unlikely, but still not a good practice to make a possibility -- and yes, looking back at the commit history, I realize I did it wrong 6-7 years ago).

Butterfly is as dead as a project can be and we are the only users in town. OpenRefine was already using a `1.0.1` version which actually included some custom patches… It really means that in effect, we are the maintainers of this library. Of course, the sooner we can get rid of Butterfly, the better.

> If an extension gets broken by the JSON serialization change, I'd argue that that indicates that it's not well enough isolated and the solution isn't to make the abstraction even more leaky, but, rather, to force the extensions to provide their own JSON serialization, unless it's a facility provided by OpenRefine through an opaque API.

Well, I would be very curious to hear how you would have done this migration then! Just have a look at the project before the migration: most of the extension points (operations, overlay models, GREL functions) are defined by interfaces (most notably `Jsonizable`) which explicitly refer to the `org.json` package in their signature. How can you get rid of this dependency without changing the interfaces?

So: yes I agree that this is a very bad breaking change (as explained in #1652 and over email), but I don't see any other way the migration can be done. I did try to make that clear to the Conservancy (this move is very bad for the OpenRefine community), but their policy is not negotiable. We could have chosen not to join the Conservancy for this reason, but we would have hit the same road block later on.

> There's not a lot of context on #1882. Is there an email thread or something else which has more detail?

No, but feel free to ask any questions if you need more detail. The bottom line is: Jackson requires its `ObjectMapper` and the annotations on the POJOs it deals with to be loaded with the same ClassLoader (https://github.com/FasterXML/jackson-databind/issues/542) and Butterfly 1.0.1 did not enforce that.

thadguidry:
@wetneb It might be beneficial to ask the Jackson community if there is any potential workaround for the ObjectMapper, etc. requirements ?  I just pulling back and thinking what other communities might have done in this case...perhaps they might have further answers...dunno.

Regarding Tom's comments, probably based around the concept of [ISP](https://en.wikipedia.org/wiki/Interface_segregation_principle), and Antonin's concern's around EngineConfig ... I strongly agree that OpenRefine be the entity that enforces the contract within interfaces, and that means a standard serialization library that we dictate. (and Spring gets it right here, fwiw, imho)

wetneb:
@tfmorris sorry for the length of this reply (I don't want to sound rude and I am genuinely interested in your opinion on all this - I have just spent a lot of time thinking about this over the past few months).

Here is one other thing to consider: at the moment JSON handling is still highly entangled with the business logic, sadly. For instance, the importers' configuration was stored in `JSONObject` ([now `ObjectNode`](https://github.com/OpenRefine/OpenRefine/blob/5639f1b2f17303b03026629d763dcb6fef98550b/main/src/com/google/refine/importing/ImportingParser.java#L54)), so classes from the JSON library are still mentioned by the interfaces between OR and its extensions. So there is no hope that we can easily decouple JSON serialization for core and extensions (even if we agreed that it would be a good move in the first place).

I did try to reduce this, for instance for operations in #1729 or for clustering in #1730, but it's a big task to do this cleanup everywhere else. (And it's prone to introducing bugs, such as #1827 introduced by #1729).

wetneb:
Thanks for the reply!

> Was any investigation done into a clean room implementation of the critical pieces of the org.json APIs? I don't know how much work it'd be, but that'd be one way to finesse the problem if compatibility is paramount. It would mean hijacking their namespace and would still require a recompile, but might be worth considering depending on the number of extensions, how hard they are to update, etc.

Yes, I have considered that. Beyond hijacking their namespace, that would mean introducing a breaking change with little maintainability benefit - we would keep all the issues that org.json have, and add yet another poorly maintained library to our dependencies. Also I am not a lawyer but I would find it questionable to copy `org.json`'s interface without acknowledging their license, especially because for a library like this, the interface pretty much determines the implementation.

Anyway this choice was made months ago, I don't think it would be wise to reverse it now. I have [made my plans clear](https://github.com/OpenRefine/OpenRefine/issues/1652#issuecomment-412343762) in August and then was the time to discuss other approaches.

> Breaking the API will make extensions incompatible, whether it's a big break or a little break, so it makes sense to take the time to get the new API right and do all the redesign at once, in my opinion. Making a series of smaller breaking changes, just makes life more difficult for the extension developers.

I still think it was sensible not to do the Maven, i18n and Jackson migrations in the same release, as these changes are unrelated. But I agree that it is worth putting effort into improving the extension mechanism before the first 3.2 release. Ideally we could migrate out of Butterfly in the same go.

> My recommendations, which I, unfortunately, don't have time to test out right now are:
>  *  evaluate an API compatible org.json workalike if the surface area of the used API is small enough
>  *  if that's not feasible, make a clean break and don't expose Jackson through the API

No for point 1 - we have already made the discussion to migrate to Jackson.
For point 2, I don't see how this is possible. Concretely, how do you propose to expose serialization through the extension interface? Via JSON payloads represented in `String`?

jackyq2015:
For point 2, I would say String might be an option to find the common media between the core and the extension. It is a shame that Java don't have the [light weigh json ](https://openjdk.java.net/jeps/198) yet even other languages already embrace it as the 1st citizen class. Developer still need to reply on 3rd party library to convert the json back and force.

 It will break anyway. so would be better to make it right and use the native representation

wetneb:
This approach would only make some sense if we are to stick with Butterfly on the long term (because migrating to another web Framework such as Spring will impose other serialization choices anyway). And even so I still think it would be a big design mistake, but hey, I respect your different architectural tastes.

So it looks like there are basically two approaches here:
* @tfmorris and @jackyq2015 seem to be keen to stick with Butterfly, its current classloading behaviour and change the interfaces so that they do not refer to any JSON library;
* @thadguidry and I (and Wes on the mailing list who seems to be enthusiastic about Spring) are thinking about migrating to a modern web framework and let the JSON library be part of the interface between extensions and the core.

In my opinion, Butterfly is dead, so I think it would be a mistake to let it guide long term design choices. Migrating out of it is a natural first step before cleaning up the web API, documenting it properly and getting a more modern frontend.

The background dilemma is to determine what is to determine what we are aiming for with the current development. Is the goal to just keep the lights on, maintaining OpenRefine as it is now, with the legacy we inherited (legacy build system, legacy web framework, legacy data storage, legacy web UI) and keeping compatibility with extensions? Or do we want to modernize it, breaking compatibility with extensions on the way?

I think both plans are definitely worth pursuing - there is definitely value in maintaining things as they are. That does not require much effort but will continue to give value to our existing user base. But in that case we should not have migrated any of the build system, localization library and JSON library - because none of it was needed to maintain existing functionality. Joining SFC is really not needed if all we want to do is to fix bugs, implement some new GREL functions, tweak the clustering methods and other things like that.

If on the contrary we think the project can be updated and revamped, then we need to do it completely and not stop halfway through, so that we get it right the first time and minimize the trouble for any extension maintainer. My opinion is therefore that we should get rid of Butterfly in the same go.

A few months ago I clearly was not keen to go down that route, because of all the incompatibilities that would generate. But the SFC application process has forced the decision. And I don't regret it at all, we already have had positive feedback about these migrations, the RDF extension is being updated, we have opened up the possibility of being packaged on Debian & Ubuntu. 

The good thing is that this is an open source project and we are running out of funding anyway, so if we cannot find an agreement we can simply fork and part ways :)

ostephens:
I don't feel I have enough knowledge to have a view on the technical aspects of these decisions, but I think it is clear that Butterfly is not under active development and is unknown to the majority of developers. On this basis I would be in favour of migrating to a more widely used and actively maintained framework sooner rather than later.

thadguidry:
In regards to breaking some compatibility with some extensions.  We had already discussed this as a team and voted to break compatibility and then help extension authors to align (even at one point Google mentioned that might help with funding for this).

In regards to migrating to a newer framework (maybe Spring), we also discussed this before in fact and even in this issue we see these votes:
 - Antonin 👍 
 - Jacky 👍  (and he's mentioned this to me before and was OK with Spring, he just thinks maybe easier route to lift up Butterfly to help with compatibility, but forgets we already agree to break some and that he agreed was OK.)
 - Thad 👍  (who feels Spring brings even more contributors to us through easier understanding of our project, but strongly feels we use it wisely where it makes sense, and avoid abstraction complexity where it does not make sense, keeping our codebase WELL COMMENTED as we go forward)
 - Tom 👎 
 - Owen ~

So its clear to me we already have enough consensus to move forward on this issue.
Also, as meritocracy means those with merit (the current doers) cast the greater vote, its also very clear we have contributors willing to help with a migration.  Although I note that those with previous merit (Tom) are worried about breaking things, we have already decided on this long ago as OK.

Let's move forward now.

tfmorris:
@thadguidry Can you provide a pointer to where this consensus was built/documented? The [link](https://github.com/OpenRefine/OpenRefine/issues/1652#issuecomment-412343762) that Antonyn provided basically said "I'm doing this" without any discussion, agreement, or disagreement.

I'll try and find time Thursday to send a note to the dev list where things are more visible. I'd hope that any casual extension developer could search for "breaking extension incompatibility" or something similar and find the relevant proposals, discussion, and decision.

re "breaking **some** compatibility with **some** extensions" [emphasis added] - what extensions would **not** be broken? Are there any?

re "previous merit" - I'm not familiar with that phrase. Definition please?

wetneb:
@thadguidry thanks, but in any case moving to a different web framework is a big move. What I can do in the remaining time is just to come up with architectural proposals: evaluating how we can translate OpenRefine's extension mechanism to other web frameworks. And then we can do an informed choice about the successor of Butterfly.

@tfmorris yes indeed all extensions will be broken by the move. Most extensions have also been broken by the move to Maven and the Wikimedia jQuery i18n plugin. About not having any "any discussion, agreement, or disagreement", well, I would have very much welcomed any!

I did not realize that this migration discussion would be more visible on the mailing list for some developers than on GitHub - the people who got involved on openrefine-dev recently were all active on GitHub. I will make sure I post these sort of things there in the future if that is helpful for you.

In any case, your informed feedback will be very welcome in the future!

thadguidry:
@wetneb Tom is still a watcher for OpenRefine, https://github.com/OpenRefine/OpenRefine/watchers, and gets all notifications to his email already, he just decided for personal reasons not to participate much in last 3 years.

wetneb:
Let's take the heat off this! It's fine for contributors to be busy with other things and Tom is right that we should have better communicated about these plans.

thadguidry:
@wetneb ??? There's no heat.  We're discussing openly, I'm just answering Tom's questions politely about my reasoning that he asked about, nothing more.  Tom and I have worked together on OpenRefine since the beginning, we're online friends, I've even paid him for valuable work.  He's just asking about my rationale, and I'm just responding about our current project status and where he can see the proposal doc and letting him know I'd like him to participate with us more.  I think your getting the wrong impression that there is some squabble when there is none.

@tfmorris Let me know if there's something we need to change in Github to allow you to watch the project status better, but from my looking at things, you should be all set to get emails already for any issues/commits.  Perhaps you have an email filter on that might be mis-categorizing, dunno.  Let me know.

wetneb:
@thadguidry great! Just so you know the fact that someone watches a repository on Github does not necessarily mean that they are receiving any emails about that repository (on-site notifications do not need to be sent by email).

wetneb:
Unless anyone else has a concrete plan forward, I will merge this soon - this is blocking for the migration of the RDF extension to Jackson.

thadguidry:
A little progress...is still some progress. :) Thanks Antonin for moving us forward.

Issue #1620, opened 2018-05-24
Ubuntu Linux 16.04, OpenJDK 1.9.0: java.lang.Error: class java.lang.IllegalAccessException


richiebful:
#### Version of OpenRefine used (Google Refine 2.6, OpenRefine2.8, an other distribution?):
- OpenRefine 2.8
- **OpenJDK 1.9.0 (Java 9)**

#### Operating Systems and version:
Ubuntu Linux 16.04
OpenJDK

#### Browser + version used - Please note that OpenRefine doesn't support Internet Explorer but works OK in most cases:
Firefox 60.0.1, but this error isn't browser-related

#### Steps followed to create the issue:
1. Run ./refine from the command line
2. Browser window will open with only the OpenRefine logo
3. Crashes with results below

#### Current Results:
```
11:27:44.374 [                   refine] Starting OpenRefine 2.8 [TRUNK]... (6107ms)
11:28:06.756 [          org.mortbay.log] Error for /index-bundle.js (22382ms)
java.lang.Error: class java.lang.IllegalAccessException: java.lang.IllegalAccessException: class org.mozilla.javascript.MemberBox cannot access class sun.net.www.protocol.file.FileURLConnection (in module java.base) because module java.base does not export sun.net.www.protocol.file to unnamed module @4543a47b
	at org.mozilla.javascript.Context.throwAsScriptRuntimeEx(Context.java:1777)
	at org.mozilla.javascript.MemberBox.invoke(MemberBox.java:169)
	at org.mozilla.javascript.NativeJavaMethod.call(NativeJavaMethod.java:247)
	at org.mozilla.javascript.optimizer.OptRuntime.callProp0(OptRuntime.java:119)
	at org.mozilla.javascript.gen.c2._c5(file:/home/rfulop/workspace/scratch/openrefine/OpenRefine-2.8/main/webapp/modules/core/MOD-INF/controller.js:567)
	at org.mozilla.javascript.gen.c2.call(file:/home/rfulop/workspace/scratch/openrefine/OpenRefine-2.8/main/webapp/modules/core/MOD-INF/controller.js)
	at org.mozilla.javascript.ContextFactory.doTopCall(ContextFactory.java:398)
	at org.mozilla.javascript.ScriptRuntime.doTopCall(ScriptRuntime.java:3065)
[...]

Issue #1890, opened 2018-12-06
 support the Java 9


jackyq2015:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
To fully support the Java 9

**Describe the solution you'd like**
Fix the run time issue and few problem with time API

**Additional context**
Will keep the Java 8 and Java 9 at the same time. Later can consider remove the Java 8

Issue #1616, opened 2018-05-19
javax.xml.bind.DatatypeConverter.parseDateTime is not included in Java 9 by default


jackyq2015:
When reporting a bug please provide the following information to help reproduce the bug:

#### Version of OpenRefine used (Google Refine 2.6, OpenRefine2.8, an other distribution?):
OpenRefine2.8

#### Operating Systems and version:
Ubuntu 18.0.4

#### Browser + version used - Please note that OpenRefine doesn't support Internet Explorer but works OK in most cases:
Chrome

#### Steps followed to create the issue:
Run UT ToFromConversionTests

#### Current Results:
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.refine.expr.util.CalendarParser (file:/media/cui/deve/src/jackyq/openrefine-community/main/webapp/WEB-INF/classes/) to method sun.util.calendar.ZoneInfo.getAvailableIDs()
WARNING: Please consider reporting this to the maintainers of com.google.refine.expr.util.CalendarParser
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
PASSED: testEscape
PASSED: testToNumber
PASSED: testToString
FAILED: testToDate
java.lang.Error: Unresolved compilation problem: 
	javax.xml.bind cannot be resolved

	at com.google.refine.expr.functions.ToDate.call(ToDate.java:92)
	at com.google.refine.tests.expr.functions.strings.ToFromConversionTests.invoke(ToFromConversionTests.java:93)
	at com.google.refine.tests.expr.functions.strings.ToFromConversionTests.testToDate(ToFromConversionTests.java:138)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:86)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:643)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:820)
[...]

PR #1898, opened 2018-12-08
support both jdk 8, oracle jdk9 and openjdk11

jackyq2015:


jackyq2015:
To close issues #1620, #1616, #1890, #1899

Issue #1676, opened 2018-07-11
Replace smart quotes transformation


ostephens:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
Sometimes data includes a mixture of smart quotes and straight quotes. Sometimes this is not wanted and it can cause confusion and problems when you want to compare data (e.g. via 'cross') or act on all quoted data

**Describe the solution you'd like**
Add on option on the Common transformations menu to "Replace smart quotes"

**Describe alternatives you've considered**
This can be achieved by the `replace` GREL, but this requires multiple replacing and finding all variations of single and double smart open and close quotes

**Additional context**
n/a

ostephens:
This originated on the discussion list - see https://groups.google.com/d/msg/openrefine/gsmP9QOKuhQ/LGPpvuovBAAJ

thadguidry:
Pages and Pages of existing code and libraries to help already with this:  https://github.com/search?q=smart+quotes

PR #1903, opened 2018-12-09
Add Common transform which replaces smart quotes with ascii equivalent

ostephens:
Adding new 'Common transform' in response to #1676. This is a trivial change, my only concern is that the list of "common transforms" that people might want to have from this menu is essentially endless. I feel that adding this one is fine, but feel there is some need here to have something easily extensible that doesn't require adding code to add things to this list (maybe hooked into 'starred' transforms?)

@thadguidry @ettorerizza comments welcome

### DIFF
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -604,6 +604,7 @@
     "core-views/trim-all": "Trim leading and trailing whitespace",
     "core-views/collapse-white": "Collapse consecutive whitespace",
     "core-views/unescape-html": "Unescape HTML entities",
+    "core-views/replace-smartquotes": "Replace Smart quotes with ascii",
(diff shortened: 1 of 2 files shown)


ostephens:
nb this replaces the following quotes:

“ &ldquo; U+201C (left curly quote) -> "
‘ &lsquo; U+2018 (left single curly quote) -> '
” &rdquo; U+201D (right curly quote) -> "
’ &rsquo; U+2019 (right single curly quote) -> '
« &laquo; U+00AB (left angle) -> "
» &raquo; U+00BB (right angle) -> "
‹ &lsaquo; U+2039 (left single angle) -> '
› &rsaquo; U+203A (right single angle) -> '
„ &bdquo; U+201E (bottom quote) -> "
‚ &sbquo; U+201A (single bottom quote) -> '

PR #1906, opened 2018-12-11
Make clustering extensible

### LINKED ISSUE #1893: Make clustering functions extensible to support other languages and scripts
Feature request from @isaomatsunami at the OpenRefine meetup today in Tokyo.

Clustering functions are quite european-centric (for instance the phonetic models were designed for English, the edit distance makes little sense for the Chinese alphabet). Implementing clustering and fingerprint functions for all languages is not easily doable but it would be good to let extensions define their own clustering methods.

Make clustering extensible: extensions should be able to register new fingerprint and distance functions.

wetneb:
This makes it possible for extensions to register custom fingerprint functions and clustering distances.

By doing so it exposes in the UI all such distances and fingerprints supported by the backend. This needs careful reviewing - why were these features not exposed in the UI before? For instance, are the various metaphone versions useful to have or should we only keep Metaphone3? It looks like the Soundex keyer was disabled for a good reason (not robust enough):
```
11:54:20.732 [                   refine] POST /command/core/compute-clusters (1507ms)
11:54:20.733 [                  command] Exception caught (1ms)
java.lang.IllegalArgumentException: The character is not mapped: À
	at org.apache.commons.codec.language.Soundex.map(Soundex.java:227)
	at org.apache.commons.codec.language.Soundex.getMappingCode(Soundex.java:181)
	at org.apache.commons.codec.language.Soundex.soundex(Soundex.java:265)
	at com.google.refine.clustering.binning.SoundexKeyer.key(SoundexKeyer.java:48)
	at com.google.refine.clustering.binning.BinningClusterer$BinningRowVisitor.visit(BinningClusterer.java:165)
```

So we need to review these clustering methods and unregister the unwanted ones from the backend. Maybe @tfmorris has some insights to share about this?

Closes #1893.


thadguidry:
@wetneb They weren't exposed before because Stefano made the decision that other algorithms were not much better for finding "clusters".  He mentioned this before [on mailing list here](https://groups.google.com/forum/#!msg/openrefine/bXM54ssq6Aw/Do07pPGYUaYJ)

wetneb:
I have disabled in the backend the keyers and distances that were not exposed in the frontend. We can re-activate them on an individual basis if users request them.

PR #1921, opened 2018-12-31
Sample extension to add some clustering methods

wetneb:
This adds a `phonetic` extension which adds two phonetic clustering methods (Beider-Morse and Daitch-Mokotoff). These encoders are available from Apache commons-codec so this does not require any new dependency (after upgrading Apache POI, which depends on commons-codec).

We could also implement this directly in main, but the purpose is to show how anyone can add their own clustering methods without changing OpenRefine itself.

@isaomatsunami if you are interested in implementing custom clustering distances or keyers for particular scripts, you could perhaps do it in a similar way. This makes it possible to add dependencies to heavy Java libraries if needed.

### DIFF
--- a/extensions/phonetic/module/MOD-INF/controller.js
+++ b/extensions/phonetic/module/MOD-INF/controller.js
@@ -1,0 +1,50 @@
+/*
+
+Copyright 2010, Google Inc.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+ * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+ * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+ * Neither the name of Google Inc. nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,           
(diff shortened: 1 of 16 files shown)


thadguidry:
@wetneb Nice one !  and it doesn't look like it took much effort !  Are you also thinking long term about supporting the need for different forms of UI for clustering (params, scopes, etc), as we discussed in other issues ?

extensions/phonetic/module/MOD-INF/controller.js
```diff
@@ -0,0 +1,60 @@
...
+var html = "text/html";
+var encoding = "UTF-8";
+var ClientSideResourceManager = Packages.com.google.refine.ClientSideResourceManager;
+
+/*
+ * Function invoked to initialize the extension.
+ */
+function init() {
+  // Register new keyers
+  Packages.com.google.refine.clustering.binning.KeyerFactory.put("daitch-mokotoff", new Packages.org.openrefine.phonetic.keyers.DaitchMokotoffKeyer());
```
thadguidry: I am thinking we should explain KeyerFactory here... help introduce the high level concepts and what they are used for.  Folks just stepping into this are looking for understanding as this will be our working example clustering extension and if they are not completely familiar with the whole part of our clustering ways of working, having comments will be useful.  Also perhaps have the link for our wiki, where Stefano explained the whole binning, etc.

wetneb:
@thadguidry no, currently it is not possible to pass parameters to custom clustering methods, as the UI is not extensible.

extensions/phonetic/src/org/openrefine/phonetic/keyers/DaitchMokotoffKeyer.java
```diff
@@ -0,0 +1,15 @@
+package org.openrefine.phonetic.keyers;
+
+import com.google.refine.clustering.binning.Keyer;
+import org.apache.commons.codec.language.DaitchMokotoffSoundex;
+
+public class DaitchMokotoffKeyer extends Keyer {
```
thadguidry: Similarly, we should introduce with a comment the concept of Keyer and what it is and why its used.  The tests help somewhat, but useful to really explain here about binning keys.

thadguidry:
> 
> 
> @thadguidry no, currently it is not possible to pass parameters to custom clustering methods, as the UI is not extensible.

Right, I know, but that's what we want to support long term.  Regardless, I'm confident in that you understand the vision here so that's good, even if we can't do much currently with the UI for extensions yet without more design mocks and discussions.

extensions/phonetic/module/MOD-INF/controller.js
wetneb: @thadguidry sure, feel free to add that!

extensions/phonetic/src/org/openrefine/phonetic/keyers/DaitchMokotoffKeyer.java
wetneb: @thadguidry I think that should rather go on the Keyer interface itself

tfmorris:
> These encoders are available from Apache commons-codec so this does not require any new dependency (after upgrading Apache POI, which depends on commons-codec).

Extensions should declare their dependencies, even if the main OpenRefine has the same dependencies, because OpenRefine's dependencies can (and do) change over time.

PR #1931, opened 2019-01-10
Timeline facet - show max date #1537

### LINKED ISSUE #1537: Timeline facet - show max date
In the current timeline facet when we move the selector on the left (minimum date) OpenRefine displays both the date and the time where the facet starts. 
When we move the selector on the right (max date), OpenRefine shows only the time. We should also display the date.

In order to make the reading easier we should also group visually the start date and time. 

facetbydate

browncow5:
Fixes #1537 

This is my first contribution to OpenRefine, please let me know if I missed anything in the contribution guidelines.

This change was to add a maximum date on the timerange facet shown here:
37438857-62f7f2d0-27cb-11e8-8a56-92cee5114dfa

This is what the change looks like:
image

### DIFF
--- a/main/webapp/modules/core/scripts/facets/timerange-facet.js
+++ b/main/webapp/modules/core/scripts/facets/timerange-facet.js
@@ -290,7 +290,7 @@ TimeRangeFacet.prototype._setRangeIndicators = function() {
   } else {
     var timeOfDayformat = "HH:mm:ss";
     var dayOfYearFormat = "yyyy-MM-dd";
-    this._elmts.statusDiv.html("<b>" + fromDate.toString(dayOfYearFormat) + "</b>  " + fromDate.toString(timeOfDayformat) + " &mdash; " + toDate.toString(timeOfDayformat) + "<b>  " + toDate.toString(dayOfYearFormat) + "</b>" );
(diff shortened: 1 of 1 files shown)


browncow5:
> @browncow5 Thanks, Ryan! BTW, how did you find out about OpenRefine and choosing this issue?

@thadguidry 
I was looking for a project that I could contribute to, so that I could keep up to date on my skills. This project just happened to meet my criteria. Which were:
-Implemented in Java
-Data oriented project
-Most importantly it was extremely well documented and easy to jump into, and had a set of issues that were mark as "Good First Issue"

browncow5:
@wetneb
I agree with what you are saying, it would be more logical if the date time was in the same order on both sides. Do you have any opinion on the date being **bold**? I was torn over whether or not I thought it was needed but in general I think it helps with readability.

**2010-07-11** 12:21:20 -- **2010-07-11** 12:21:20

thadguidry:
@browncow5 Awesome, "Good First Issue" is working!  Yeah, we added some, as suggested by Github.  So I am glad that new feature they added is working!  Thanks for helping and so glad you found us! ;-)

wetneb:
@browncow5 so it looks like there is a strong consensus for changing the order of end date and end time. Could you change that? (You can do this with another commit on the same branch, no need to make a new PR.)

Issue #1623, opened 2018-05-25
Error when trying to change workspace via -d terminal command


DFoltzMorrison:
#### Browser: Chromium 66.0.3359.139

I am trying to switch between different workspaces [as listed in the FAQ](https://github.com/OpenRefine/OpenRefine/wiki/FAQ).

I've made a copy of the exact contents of my current workspace into a folder ('/workspace a') on another drive, as well as copies of two other workspaces from different machines into folders ('/workspace b' and '/workspace c').

Using the terminal commands `./refine -d '/path/to/workspace a/'` and 
`./refine -p 3333 -i 0.0.0.0 -m 14000M -d '/path/to/workspace a/'`

results in the following error:

```
Starting OpenRefine at 'http://127.0.0.1:3333/'

Error: Could not find or load main class 00.openrefine

**Comment by jackyq2015** [comment]
you need to remove the space in the path.

**Comment by DFoltzMorrison** [comment]
I can rename the workspace directory to remove the space, but the drive name has a space in it:
 `/media/USER/HDD 00/openrefine_workspaces/`.

Removing the space in the drive name allows openrefine to launch in my browser and displays the following:

```
HTTP ERROR 503

Problem accessing /. Reason:

    java.lang.NullPointerException

Caused by:

javax.servlet.UnavailableException: java.lang.NullPointerException
	at org.mortbay.jetty.servlet.ServletHolder.makeUnavailable(ServletHolder.java:415)
	at org.mortbay.jetty.servlet.ServletHolder.initServlet(ServletHolder.java:458)
	at org.mortbay.jetty.servlet.ServletHolder.doStart(ServletHolder.java:263)
	at com.google.refine.RefineServer.configure(Refine.java:296)
	at com.google.refine.RefineServer.init(Refine.java:208)
	at com.google.refine.Refine.init(Refine.java:114)
	at com.google.refine.Refine.main(Refine.java:108)

Powered by Jetty://
```

DFoltzMorrison:
I renamed the workspace directory to remove the space, and had to change the parent drive's label in a partition editor to remove another space.

I can successfully switch between workspaces, but now have the job of fixing a lot of broken file paths.

Perhaps a future version of openrefine can support file paths with spaces wrapped with quotes?

jackyq2015:
@DFoltzMorrison  Glad it works eventually. sorry for the inconvenience. Will add into next milestone.

PR #1932, opened 2019-01-11
enable spaces in data directory

Rots:
* pass options as proper bash list items so that even if directories
contain spaces, they will be properly passed on to java - #1623
* ignore empty options

### DIFF
--- a/refine
+++ b/refine
@@ -93,7 +93,9 @@ EOF
 }
                 
 add_option() {
-    OPTS="$OPTS $1"
+    if [ ! -z "$*" ] ; then
+        OPTS+=("$@")
+    fi
 }
                 
 load_configs() {
@@ -497,7 +499,7 @@ ui_test() {
     echo "Killing OpenRefine"
     /bin/kill -9 $REFINE_PID
     echo "Cleaning up"
-    rm -rf $REFINE_DATA_DIR
+    rm -rf "$REFINE_DATA_DIR"
 }
 
 server_test() {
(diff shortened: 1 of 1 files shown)


thadguidry:
hmm, this breaks it on Windows now.

wetneb:
@thadguidry hang on, you are running the `refine` script on windows? Shouldn't you use `refine.bat` instead?

thadguidry:
I am using `refine.bat` on Windows within a Windows Command Prompt.  Using this as a test...
`C:\Users\THAD\OpenRefine>refine /d "C:\Users\THAD\test Workspace A"`

but my old workspace still shows up in Refine.

Rots:
Does `.bat` file need to be adjusted as well? I don't have Windows to test, sorry.

wetneb:
@thadguidry right, so this PR did not touch `refine.bat` at all, so the problem should not be new (we can reopen the corresponding issue if this is still a problem on windows).

thadguidry:
@wetneb it is not a feature of the `refine.bat` that I personally use.  I was just testing it out.  Sure, you guys can open an issue for the `refine.bat` to fix the /d option not working as it does on Mac/Linux.  Up to you guys, but it doesn't affect me as I said since I personally don't use that `/d` option on Windows, but others might want to see it fixed.

wetneb:
If anyone knows batch files and wants to work on that great. This does not seem like a huge priority as I expect most Windows users use our `.exe` releases (whereas the `refine` script is the default for Linux users). But of course it would always be good to have.

PR #1959, opened 2019-02-19
 Show reconciliation previews on hover of reconciliation candidates 

### LINKED ISSUE #1943: Item preview in reconciliation should show up when hovering a candidate or match
From a user:

> I cannot see the Wikidata hovercards. I use OpenRefine 3.1 and tested it with Firefox and Chrome. The browser doesn't make any requests when holding my mouse over an reconciled entry. The preview url (https://tools.wmflabs.org/openrefine-wikidata/en/preview?id=Q42) does work though. Is there something you have to turn on in order to see the hovercards?

This seems to be a fairly common misunderstanding for people who see screenshots or videos and expect that hovering links will display the preview. We should make sure previews are displayed when hovering both matched cells and candidates in unmatched cells.

wetneb:
When hovering unmatched reconciliation candidates and matched cells, the reconciliation preview from the service shows up (if provided).

Clicking on unmatched reconciliation candidates now has the same behaviour as clicking on matched cells: it opens the page of the entity in a new tab.

As argued in #1943 I think this is a much more natural behaviour and I have heard of multiple users who were confused about this. I think it should not be too hard for experienced users to adapt.

Tested in Firefox and Chromium: please test extensively on as many environments as possible and report if anything does not feel natural.

One possible downside of this is that people could feel like the previews pop up too often, when the user inadvertently hovers a reconciliation link without any intent to inspect it. If you think this is a big issue, maybe it's worth removing the hover on matched cells, but my gut feeling is that it is nice to preview matched cells too (especially when using the auto-match feature in the reconciliation operation, which is enabled by default).

Closes #1943.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
@@ -108,7 +108,7 @@ DataTableCellUI.prototype._render = function() {
         a.attr("href", encodeURI(service.view.url.replace("{{id}}", match.id)));
       }
 
-      self._previewOnHover(service, match, a, false);
+      self._previewOnHover(service, match, a, a, false);
 
       $('<span> </span>').appendTo(divContent);
       $('<a href="javascript:{}"></a>')
@@ -153,7 +153,7 @@ DataTableCellUI.prototype._render = function() {
               a.attr("href", encodeURI(service.view.url.replace("{{id}}", candidate.id)));
             }
 
-            self._previewOnHover(service, candidate, liSpan, true);
+            self._previewOnHover(service, candidate, liSpan.parent(), liSpan, true);
 
             var score;
             if (candidate.score < 1) {
@@ -475,7 +475,7 @@ DataTableCellUI.prototype._previewCandidateTopic = function(candidate, elmt, pre
 /**
  * Sets up a preview widget to appear when hovering the given DOM element.
  */
-DataTableCellUI.prototype._previewOnHover = function(service, candidate, element, showActions) {
+DataTableCellUI.prototype._previewOnHover = function(service, candidate, hoverElement, coreElement, showActions) {
     var self = this;
     var preview = null;
     if ((service) && (service.preview)) {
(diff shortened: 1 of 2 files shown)


thadguidry:
@wetneb I was hoping that we had a Toggle Preference for "hover.reconcilePreview"="true"?  Other than that, I agree it the new tab feels natural, but popup reconcile preview feels unnecessary and necessary at different times of the data cleaning/reconciling phases.  So let's see if we can make this a preference to turn on or off as needed.

wetneb:
@thadguidry a toggle could be useful, but where would we put it?
* I don't think it would make sense to put it in the reconciliation dialog - that contains the configuration options for an operation, not for the UI.
* Putting it in the global preferences would be pretty useless - users are not likely to discover it and it is going to be inconvenient to go back and forth between the project and the preferences to switch that when required. Moreover the preferences are global, not per project, so it's inconvenient if you have one project where you need the previews and not the other.
* Putting it next to the records/rows mode and the number of rows to display could work, but IMHO it's a very niche setting that does not deserve to occupy highly-visible space there. And it's going to make this bar longer, which is likely to cause problems on small screens.

It's easy to add a toggle in the global preferences but my gut feeling is that it's not going to be used…

thadguidry:
We could just add it to the existing "Project metadata" dialog.
`//*[@id="metadata-body"]`
But instead of it being a Key/Value... but the new toggle option right above the Key/Value table.
Label the toggle option "Hover Recon Preview" and default it on and give it a tooltip to explain it.
Store the setting as part of the Project metadata for now if we can to simplify.
We can always partition it later with
"Project metadata"
"Project settings"
(or if not too much work, you can add a 2nd storage layer to Projects called "Settings" in addition to "Metadata"? but I think that's more work, and quick and dirty is fine for now)

wetneb:
Still not convinced, it's hacky. It really does not belong to project metadata: UI settings should not go there. I will add a toggle in the global preferences for users who get annoyed by the popups and that will do.

wetneb:
My commit message was incorrect, it's `cell-ui.previewMatchedCells=false` that you need to add to disable previews on matched cells. This requires reloading the project page too.

Issue #1957, opened 2019-02-19
New behaviour of list facet breaks predefined facets


wetneb:
The change in #1666 requires users to add `.toString()` at the end of the expression for their facets.
Some predefined facets have not been adapted for this.

This is a bad breaking change and should be reverted.

If we want to introduce a different behaviour for a facet, here is a better way to do it:
* introduce in the backend a new facet with the new behaviour. The facet will have a different identifier and the previous facet will still be available.
* update the UI to migrate some or all of the predefined facets to this new facet identifier with new behaviour.
* if the old facet is no longer exposed in the UI, and after a long period of depreciation (so that people can adapt their workflows), the old facet may be dropped from the backend.

thadguidry:
@wetneb We talked about this in #1666 here: https://github.com/OpenRefine/OpenRefine/pull/1666#issuecomment-421719191

Can it be reverted and still show users their data inconsistency as I mentioned in my comment?  I don't think so, but open to discuss.  My worry and Owen's was that the previous facet being available does not give users a good view on their messy data.  That's very important to us, that we not hide data issues and I mentioned this in my comment.

But I guess your take is that the existing facets may hide data issues, but we'll provide new facets with new behavior.  My worry there is that we stay in the same rut with folks missing data issues by using the old facets and not new facets.  Anything you can do to make the new facets be named the same as old facets (from a display perspective to users) and in the same menu structure would certainly help there I guess.  They get the new facets, but for compatibility reasons, we still have the old facets in some submenu with a different label.

wetneb:
@thadguidry yes I understand the motivation of the changes. I think the new facet is great but it needs to be introduced in a different way to avoid the bugs that this change has created. This has confused a lot of users.

So yes, if someone wants to work on making this new behaviour work for everyone, I would welcome a PR which would follow the steps above. In the meantime, the first step is to revert the change.

thadguidry:
@wetneb OK thanks for confirming. Agree to revert and work towards a better steps as you identified above.

ostephens:
I've commented against #1969 that I'm not against reverting the current changes - and I'm still happy if you want to go ahead with this. However, re-reading #1662 I'm not completely convinced that reverting the changes is really the right thing to do - #1662 is clear that there was some really funky behaviour in the previous situation (like selecting a value from the facet and not seeing all the rows from that facet).

[...]

PR #1969, opened 2019-02-22
Revert "Fix Reconcile facets affected by change to list facet boolean handling"

wetneb:
@ostephens  this  proposes to revert OpenRefine/OpenRefine#1888 and #1666, for #1957. As I explained there I agree the feature is useful to have but it is important that we introduce it in a non-breaking way. Is it okay on your side?


ostephens:
Not against this - but a couple of points:

* assume the intention is to revert all the changes in this PR and you've just opened it with the single change as a starter?
* These reverts you've made here don't actually need reverting if #1888 is reverted - they should work exactly the same whether 'toString()' is appended or not once #1888 is reverted (I don't mind reverting them, but just noting it isn't necessary)

wetneb:
@ostephens sorry I had not realized that I did not push the commit that reverts the main change #1666, indeed. This is now done.

I agree that the problem can be resolved without reverting #1888. I think it is simplest to do it anyway for clarity as this was introduced as a fix which is not needed once we revert #1666. Once a new facet with the new behaviour is introduced, we can then add the `.toString()` where the new facet is used.

Issue #1977, opened 2019-03-06
add menu option "All / Facet / Facet by non-blank"


felixlohmeier:
I would like to suggest to add the custom text facet recipe [Count non-blank values in each column](https://github.com/OpenRefine/OpenRefine/wiki/Recipes#count-non-blank-values-in-each-column)
`filter(row.columnNames,cn,isNonBlank(cells[cn].value))` to the "All" menu, e.g.
* "All > Facet > Facet by non-blank"
* or "All > Facet > Count non-blank values"

I guess this recipe is a common first step to get a quick overview of your dataset, so it should be more easily available for new users.

wetneb:
@felixlohmeier sounds like a sensible thing to do! Do you want to try submit a PR for that? It should not be too hard, you can get some inspiration from the existing facets in the menus.

PR #1979, opened 2019-03-08
Add custom text facet recipes from the wiki to "All" menu

felixlohmeier:
I'd like to suggest to add three custom text facet recipes from the wiki to the "All" menu to make them easily available for new users.

* [count-non-blank-values](https://github.com/OpenRefine/OpenRefine/wiki/Recipes#count-non-blank-values-in-each-column)
  * rows: All / Facet / Count non-blank values/column #1977
  * records: All / Facet / Count non-blank records/column
* [Rows with blank cells in all columns (true/false facet)](https://github.com/OpenRefine/OpenRefine/wiki/Recipes#facet-for-rows-with-a-certain-number-of-blank-cells)

openrefine-all-menu-new-facets_1

openrefine-all-menu-new-facets_2

test data: [openrefine-all-menu-new-facets.openrefine.tar.gz](https://github.com/OpenRefine/OpenRefine/files/2947200/openrefine-all-menu-new-facets.openrefine.tar.gz)


felixlohmeier:
Oops, looking at the screenshots again, I've just realized that I mxied up true/false in the Facet "Blank rows". Fixed with commit 9c55c7e

openrefine-all-menu-new-facets_3

wetneb:
@felixlohmeier it looks great to me, thanks a lot! This is going to be very useful, I think. I'll have a closer look and merge this soon, it should make it into 3.2.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -581,6 +581,11 @@
     "core-views/starred-rows": "Starred Rows",
     "core-views/facet-flag": "Facet by flag",
     "core-views/flagged-rows": "Flagged Rows",
+    "core-views/blank-rows": "Blank Rows",
+    "core-views/count-non-blank-values": "Count non-blank values/column",
```
wetneb: Would it be cleaner (and more consistent with the facet name) to write "values per column" instead of "values/column"?

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -581,6 +581,11 @@
...
+    "core-views/non-blank-values": "Non-blank values per column",
+    "core-views/count-non-blank-records": "Count non-blank records/column",
```
wetneb: Same here.

main/webapp/modules/core/langs/translation-en.json
felixlohmeier: Yes, but the width of the menu (2nd level) seems to be fixed. The (longer) string "Count non-blank values per column" would cross the border line.

thadguidry: You can drop the word "Count".  It's implied as other facets do.  Just say "Non-Blank values per column", etc.

felixlohmeier: Alright, fixed with 2c266d3

felixlohmeier:
Screenshot from 2019-03-11 20-31-24

Issue #1976, opened 2019-03-06
Improve diagnostic information available to users of the cross function


ostephens:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
The `cross` function can sometimes fail for reasons unknown. Troubleshooting the failure is difficult because there is a lack of reporting either in the console or through `errors` that can be stored in OpenRefine

**Describe the solution you'd like**
When the `cross` fails because of a specific issue (e.g. that the project or column referenced does not exist) then the error should be available to be stored in a cell

When the cross function fails to find a value we should consider the option of returning an error ("value not found in project X column Y") to be stored (instead of 'null') if the user requests it

See also #1950

thadguidry:
I'd also like to see a snippet "cross lookup failed" or whatever in the console logs when `cross("my_project", "my_column")` fails. (this will help test/debug later on with new UI)

wetneb:
I don't think we should log this, this is likely to generate lots of redundant logging messages as the cross function is typically applied on all rows. Using the existing GREL error mechanism seems more natural.

thadguidry:
I would aggregate >0 then throw error, but that's fine Antonin.  No log then.

Also, seems [Project renaming](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/commands/project/RenameProjectCommand.java) affects this as well...I renamed my project and then now `cross()` lookups fail even after restarting Refine.  David, why did you hardcoded the tests so much and not use ProjectMetadata?  He probably was rushing to get this to me at the time, lolol.  I really needed cross() to add extra data on a large Freebase upload.

ostephens:
I agree with @wetneb that console logging is going to be too verbose here. I can see an argument for having a 'verbose' mode for the console log which enable us to push information for debugging when necessary - but that's much bigger scope than this particular issue

ostephens:
Just putting down some notes about this:

Having had a look at `cross` and `ProjectJoin.getJoin` my first thought is that the process of translating a Project Name string to a project ID should be separated out from `ProjectJoin.getJoin`. If the project ID lookup is done directly from `cross` this gives us better opportunity to report on any errors in the process. If `ProjectJoin.getJoin` accepts project IDs rather than Project Name strings, I think this will also resolve one of the issues noted by @wetneb in #1950 :

[...]

PR #1985, opened 2019-03-17
Cross error reporting

ostephens:
For #1976 
Refactor of cross (and related code) to:
* Fail with clear error to user when attempt to cross with non-existent project name
* Fail with clear error to user when attempt to cross using a non-unique project name
* No longer require the name of source project (from which the cross is being done) is unique

### DIFF
--- a/main/src/com/google/refine/InterProjectModel.java
+++ b/main/src/com/google/refine/InterProjectModel.java
@@ -46,6 +46,7 @@ import com.google.refine.expr.WrappedRow;
 import com.google.refine.model.Column;
 import com.google.refine.model.Project;
 import com.google.refine.model.Row;
+import com.google.refine.util.JoinException;
 
 public class InterProjectModel {
     static public class ProjectJoin {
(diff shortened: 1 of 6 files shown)


wetneb:
@ostephens thank you very much for working on this! It looks very sensible to me. Have you considered adding a few tests to demonstrate?

ostephens:
Happy to add some tests when I get some time (I'll try to find some time soon!)

thadguidry:
So @ostephens there's going to be another PR to address the "column name not found" scenario also since you mentioned both project and column in #1976 ?
Currently its sorta silent if the column name is not found in a project that is found.... it returns `null` instead of a nice error to the user like "NonExistingTestColumn cannot be found in project clipboardCrossTest".
```
cell.cross("clipboardCrossTest", "NonExistingTestColumn")
```
returns `null`

Thanks so much Owen for this so far and looking forward to additional PR's to address the missing column name scenario and more tests!

ostephens:
Added checks for non-existent columns and tests

Issue #1999, opened 2019-03-28
Cell types being set back to String 


ostephens:
**Describe the bug**
Cells which have been converted to non-string types of "number" or "boolean" are being set back to "string" on restarting OpenRefine (or possible on project save?)

**To Reproduce**
1. Create a project
2. Insert number or boolean values into a cell
3. Check these have been correctly set by using GREL "value.type()"
4. Restart OpenRefine
5. Check the cell types by using GREL "value.type()"

**Current Results**
Before restarting the values show as the expected types
After restarting the values show as Strings

**Expected behavior**
Values should retain type as set

**Screenshots**
Before restart:
<img width="888" alt="Screenshot 2019-03-28 at 09 32 56" src="https://user-images.githubusercontent.com/576174/55146448-830bff80-513c-11e9-985d-13e59bb427ca.png">

After restart:
<img width="917" alt="Screenshot 2019-03-28 at 09 34 08" src="https://user-images.githubusercontent.com/576174/55146562-ba7aac00-513c-11e9-8a32-c01652f47741.png">

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Browser Version: Safari Version 12.0.3 (14606.4.5)
 - JRE or JDK Version: JRE 1.8.0_72

**OpenRefine (please complete the following information):**
 - 3.2 beta

**Additional context**
Dates do not seem affected by this issue

ostephens:
Possible this is related to #1996?

wetneb:
I agree this is important, it's probably linked to the Jackson migration.

ostephens:
@wetneb I've done some further investigation into this and #1996. It looks to me like the problem is with com.google.refine.modelCell.getValueasString https://github.com/OpenRefine/OpenRefine/blob/ebaad96dfc33f44b1a0904800cd8bdac7c655220/main/src/com/google/refine/model/Cell.java#L105 which leads "v" to always serialise to a string

In the case of booleans or numbers (of any type that can be represented as a number in JSON) then these should be serialized to their native types in JSON rather than as a string. I could probably fix this, but I'm not very confident.

I also feel that possibly this whole aspect of Cell.java might benefit from re-factoring using Jackson annotations (e.g. could the date processing done here be equally well handled by a @JsonFormat annotation? ) - but I don't have the knowledge to be sure about the best way to approach this.

Any views? I'm happy to try a PR which you can comment on if you want? Or do you already have this in hand?

[...]

Issue #1996, opened 2019-03-25
Numbers and Booleans don't display in green in cells (3.2 beta)


ostephens:
**Describe the bug**
When a cell is populated with a number or boolean, it doesn't not display as green text in the UI

**To Reproduce**
1. Open a project
2. Edit a cell to a number or boolean
3. View project

**Current Results**
Cell text appears in the same colour as a string

**Expected behavior**
Cell text for numbers and booleans should display in green
(n.b. this behaviour works for Dates)

**Desktop (please complete the following information):**
 - OS: macOS Mojave
 - Browser Version: Safari 12.0.3

**OpenRefine (please complete the following information):**
 - 3.2 Beta

ostephens:
I think this might be part of a more serious bug where types aren’t being applied correctly on the backend. More investigation needed

ostephens:
JSON being returned on a get-rows request isn't include a "t" property for numbers/booleans

```
{
    "mode": "row-based",
    "rows": [
        {
            "starred": false,
            "flagged": false,
            "cells": [
                {
                    "v": "2019-01-01T00:00:00Z",
                    "t": "date"
                }
            ],
            "i": 0
        },
        {
            "starred": false,
            "flagged": false,
            "cells": [
                {
                    "v": "1.0"
                }
            ],
            "i": 1
        },
        {
            "starred": false,
            "flagged": false,
            "cells": [
                {
                    "v": "2.0"
                }
            ],
            "i": 2
        },
        {
            "starred": false,
            "flagged": false,
            "cells": [
                {
                    "v": "true"
                }
            ],
            "i": 3
        }
    ],
    "filtered": 4,
    "total": 4,
    "start": 0,
    "limit": 4,
    "pool": {
        "recons": {}
    }
}
```

The second row should have `"t": "number"`

Might be related to Jackson migration and serialisation of the JSON?

wetneb:
Probably! Thanks for the debugging, I'll look into that.

ostephens:
OK - I was wrong about that - the issue is not the lack of a "t" value but that in the JSON the number and boolean is being included as a string not a JSON number/boolean. So instead of:
```
{
            "starred": false,
            "flagged": false,
            "cells": [
                {
                    "v": "true"
                }
            ],
            "i": 3
        }
```
it should be
```
{
            "starred": false,
            "flagged": false,
            "cells": [
                {
                    "v": true
                }
            ],
            "i": 3
        }
```

[...]

PR #2008, opened 2019-04-04
Cell serialization fixes

ostephens:
This PR adds more tests for cell serialization for various types of cell data, and ensures that number, date and boolean cells are serialized in the correct way.

Also modifies wikidata extension to only try to compute recon features such as name match, levenshtein distance etc when the cell contains a string


ostephens:
@wetneb I think this fixes the serialization issues reported in #1999 & #1996, and adds tests to try to catch any future issues in this area (I think we can probably simplify the getValue further, but I want to be confident of the tests first).

I'm not so sure I've done the right thing in terms of reconciliation values - what do you think?

Issue #1909, opened 2018-12-15
Generic JDBC Support


kidehen:
OpenRefine needs a Generic JDBC Connector for binding to any JDBC Driver. This isn't the case right now and the [closed status of the original issue relating to this matter](https://github.com/OpenRefine/OpenRefine/issues/1277) obscures the current state of affairs. 

I am reopening this as a new feature request. 

Naturally, opening this case doesn't mean that I won't consider engaging the services of @OpenLinkSoftware regarding issue resolution contribution. 

/cc @thadguidry

kidehen:
Little note to self, re connecting these JDBC threads.
{
<http://linkeddata.uriburner.com/about/id/entity/https/github.com/OpenRefine/OpenRefine/issues/1277>
skos:related <http://linkeddata.uriburner.com/about/id/entity/https/github.com/OpenRefine/OpenRefine/issues/1909> .
}

thadguidry:
@kidehen 
OK, so, my vote, SURE, let's have a way to hook in a Generic JDBC connector.
On other Enterprise applications (Pentaho, Talend, etc), I have seen this handled fairly easily with forms that let's you specify the .jar file path of the driver you intend to use, along with configuring Options and Advanced settings.

Example info & pics here: https://help.pentaho.com/Documentation/6.1/0P0/0U0/010/000

But we'd need a sponsor to pickup the bill for doing the work, or would welcome contributions !

wetneb:
@kidehen I think it would make sense to have a generic JDBC support as you hinted at.
Do you have an idea of what this would look like as a user? How would I supply a JDBC driver to OpenRefine?

Also, in case your fork is public, it would be interesting to see how you implemented the feature there, so that we can take inspiration from it.

kidehen:
We haven't done anything bar add support for our Virtuoso platform which has built-in bridging to ODBC or JDBC data sources. 

JDBC integration simply boils down to a JAR location and a JDBC Connection URL. 

SeeAlso: 
https://github.com/OpenRefine/OpenRefine/issues/1909#issuecomment-447607838  -- earlier comment from @thadguidry .

PR #2036, opened 2019-04-30
Add Virtuoso connector for database extension

wetneb:
This adds a Virtuoso connector to the database extension. This was developed by @smalinin in the OpenLinkSoftware fork of OpenRefine (released under the same license, so we can pull it back upstream).

Things to consider:
- [ ] can we have tests for this? (Does Virtuoso run on Travis and/or Appveyor)?
- [ ] can we avoid duplicating code from other connectors?

Any other thoughts?


wetneb:
Closing due to the lack of interest.

tfmorris:
I've rebased this at https://github.com/tfmorris/OpenRefine/tree/virtuoso-db, but in looking at it, I think we need to refactor the DB drivers and eliminate all the cut&paste duplication, perhaps as a precursor to, or part of, #1909.

PR #2038, opened 2019-05-01
Allow GREL to uniquely hash non-string types including arrays.

### LINKED ISSUE #2037: Allow GREL to uniquely hash non-string types including arrays
GREL should be able to uniquely hash non-string types including arrays

**Background**  
GREL provides two hash functions, md5 and sha1.

These allow you to hash values as shown in the following examples:  
`md5('my_cool_string')` -> `8aa410c1edaf7f783f84cf5fc3f102b8`  
`sha1('my_cool_string')` -> `6a90523cab73e50bfff96efb745650b2f009e9a8`  
`md5('row number: ' + rowIndex)` -> `(depends on rowIndex)`  

This works as expected with string arguments and is consistent with the specification for sha1 and md5.

Although the GREL documentation for md5 and sha1 describe both of these functions as having one argument of type string, these functions will in fact return results for one non-string argument.  
`md5(5)` -> `e4da3b7fbbce2345d7772b0674a318d5`  
`sha1(true)` -> `5ffe533b830f08a0326348a9160afafc8ada44db`  
`md5(PI)` -> `2836dd7cd23c875da1039ba2b4b4ccd0`  
`md5(toDate('04/26/2019 9:43:12pm'))` -> `797ccb12e2b76512cd8034eca2f9c6d2`  

The results for non-string arguments are obtained by hashing the java toString representation of the object argument:  
`sha1(5) == sha1('5')` -> `true`  
`sha1(true) == sha1('true')` -> `true`  
`sha1(PI) == sha1('3.141592653589793')` -> `true`  
`sha1('2019-04-26T21:43:12Z') == sha1(toDate('04/26/2019 9:43:12pm'))` -> `true`  

Using toString results in very undesirable behavior in some cases:  
`md5([1, 2, 3]) == md5([1, 2, 3])` -> `false`  

There are two similar issues here that I will address.  The first involves datatypes, the second involves Object uniqueness and representation.

First, I propose that datatype should be considered when hashing.  A date should not equate to a string.  Type is important in GREL and can be handled explicitly.  Proposing examples:  
`md5('2019-04-26T21:43:12Z') == md5(toDate('04/26/2019 9:43:12pm'))` -> `false`  
`md5(5) == md5('5')` -> `false`  
`md5(5) == md5(toNumber('5'))` -> `true`  
`md5(toString(5)) == md5('5')` -> `true`  

Second, I propose the hash of a value representing the same data should be consistent.  It should be based on the underling data, and not the object instantiation or memory address.  Proposing examples:  
`md5([1, 2, 3]) == md5([1, 2, 3])` -> `true`  
`md5([1, 2, 3]) == md5(sort([2, 3, 1]))` -> `true`  
`md5([1, 2, 3]) == md5(reverse([3, 2, 1]))` -> `true`  
`md5([[1], [2, 3]]) == md5([[1], [2, toNumber('3')]])` -> `true`  

**Proposed Solution**  
In order to address both of these issues I propose that instead of calling toString on non-string arguments to md5 and sha1, non-string arguments should be serialized (if they can be) and the byte array of that serialization should be hashed.  This will account for the datatype and it will also capture the uniqueness of the underlying data.  It requires minimal code changes and provides a fast means to hash non-string values consistently.

**Alternatives**  
Alternative methods to serialization that I considered did not meet the criteria stated above.  java.io.Serializable will work nicely with minimal code changes.  If anyone has concerns or suggestions, please advise.  Thanks.

**Additional context**  
I will update the wiki (https://github.com/OpenRefine/OpenRefine/wiki/GREL-String-Functions#md5string-s) to reflect the changes when this issue is resolved.

lemmingapex:
Closes #2037.  See the ticket for details.


wetneb:
Many thanks for looking into that! Sorry for the Travis issues, we need to fix our build.

I think hashing arrays is a perfectly valid motivation. Here are a few questions:
- would it not be preferable to make sure that `5` and `"5"` are hashed to the same thing as it was the case before? Some of our users might potentially rely on this behaviour, so I would consider this as a breaking change.
- concerning testing, would it not be preferable to assert equality (or difference) of hashes from two objects, rather than writing down the hash value in the tests? Your issue explains the motivation very well in terms of hashes of objects being equal or different, so I think the same sort of assertions should be used for the tests. Users probably do not care about specific hash values but rather about the collisions / non collisions.

lemmingapex:
Hey Antonin,

Thanks for your quick response.  Those are both great questions that I have spent some time thinking about.

First, concerning the tests, I think it is a great idea to add the assert equality (or difference) of hashes from two objects. I will add these tests once we resolve how types should be handled.  I think it is a good idea to keep the string equality tests in order to make sure sha1 and md5 are working as expected however.  For example, in a situation where md5 always returns `null`, the following test would pass although it does not really check that the md5 is created as expected:  
`Assert.assertEquals(new MD5().call(bindings, new Object[] { "5" }), new MD5().call(bindings, new Object[] { "5" }));`

Does this make sense?

Second, the question of whether or not hashing should consider type or not is a good questions.  There are pros and cons on both sides of the argument.  Here are some arguments:

We should try to equate different datatypes during hashing:  
* Some users might rely on the hash of a string and non-string equating.
* This change might be considered a breaking change.  `md5(5.0) == md5('5.0')` will now result in `false`.

We should consider different datatypes and different data during hashing:  
* The documentation never described sha1 or md5 as being able to take non-string arguments.  You could argue that because of this, the behavior was not well defined and should not be relied upon.
* A user savy enough to use md5 or sha1 almost certainly has some understanding of datatypes, and should be able to address the cases that hash non-string values using GREL functions like `toString` and `toNumber`.
* In GREL, if `md5(5) == md5('5')`, should that imply that `5 == '5'`?  Because `5 != '5'` in GREL, then shouldn't `md5(5) != md5('5')`?
* If we were to let `md5(5) == md5('5')` what should the following return:
`md5(5.0) == md5('5')`
`md5(5.) == md5('5.')`
`md5(5.) == md5('5')`
`md5(5.0) == md5('5.0')`
`md5(5.00) == md5('5.00')`
`md5(5.000000) == md5('5.000000')`
`md5(005.0000001000) == md5('005.0000001000')`
It should not be expected that users know that java `toString` of `5.000000` gives `"5.0"`.  I argue that `md5(5.00) == md5(5.000)` because `5.00 == 5.000`; and `md5("5.00") != md5("5.000")` because `"5.00" != "5.000"`.  So why should `md5(5.00) == md5("5.0")`, but not `md5(5.00) == md5("5.00")`?

* Type is important in GREL, and it should be considered.  toString can fundamentally change the data representation.  For example, what is the output of `toString(toNumber("4e3"))`?  `"4e3"`? `"4.0e3"`? `"4000"`? `"4000.0"`? `"4000.0000"`? `"0004000"`?  Hashing should not misrepresent your data.


What are your thoughts on all of this Antonin?

Thanks for your input!

wetneb:
Thanks for the reply! My personal preference would be to only change the serialization of arrays, so that the behaviour of hashing on other datatypes is preserved - that would avoid the breaking change. This is because OpenRefine already conflates objects with different datatypes but identical string representation in other places (such as the list facet), although this is something we are trying to fix too (but in a non-breaking way).

If you have a use case in mind where it is important that values with different datatypes are not hashed to the same thing, we could add a parameter to opt-in to this behaviour, or introduce new functions.

I know this might sound a bit over-cautious, but we have had issues in the past with seemingly innocuous breaking changes which ended up being severely damaging, so I prefer to err on the side of caution.

thadguidry:
Just catching up on this issue and read it through.
I agree with @wetneb where we want to be cautious on breaking changes. (changing behavior of previous OpenRefine functions and making them incompatible)  But certainly introducing NEW functions or hashing methods is safe and as @wetneb says, we just need the use case.

Historically, OpenRefine (as a powertool where we let the user decide often) we left it to the user to determine if they wanted to Type their data or not.  Some folks are fine with just Strings.

If this use case is to provide MORE functions that let a user decide how to hash, then +1.
If this use case is to provide LESS functions and begin to control how to hash or assume many things for the user, then -1.

lemmingapex:
Thanks for the feedback @wetneb and @thadguidry.

The use case is to provide MORE functionality, and let the user decide how to hash.  This change allows users to choose how they manipulate their data:  
`md5(value)`  
`md5(toNumber(value))`  
`md5(toString(value))`  

As the hash functions are currently, refine makes the assumption that your values should be converted to strings.  It controls how data is hashed in a non-intuitive manner.  There is less functionality.

If you consider these changes as 'breaking', is it possible to include them in a next major release?

lemmingapex:
If type is to not be considered in the hash of a value, but arrays are handled in a more practical way, then a user could effectively capture the type of a value in a hash using a transformation of the type:  
`md5([value, type(value)])`  

I don't like it, but it does the job.  😃

tfmorris:
What's the status of this? It's been inactive for a year. It looks like @wetneb gave a thumbs up to the last proposal, but that might not have triggered a notification.

Another approach to this would be to make sure that toString() returns reasonable values for arrays so that the existing practice of coercing to string would work more reasonably. That would be most in line with current OpenRefine practices (which aren't necessarily the best, but are established).

wetneb:
Let us close this since we have identified other ways to fix the issue. Thank you @lemmingapex for your contribution!

Issue #2041, opened 2019-05-15
Column Reorder leaves undeleted hidden cells.


stanakaj:
hi, I appriciates you for maintaining this nice tool.

**Describe the bug**
Column Reorder leaves undeleted hidden cells in case of removing column.
So you get unexpected cell data in a some condition.

**To Reproduce**
1. import a data set
![pic1](https://user-images.githubusercontent.com/44893868/57740654-de497f80-76f3-11e9-9faa-fec9e0de0c9a.png)

2. remove columns with Reorder
![pic2](https://user-images.githubusercontent.com/44893868/57740660-e73a5100-76f3-11e9-9853-c578de6a6072.png)
above instance : remove 'b', 'c' columns with Reoerder

3. split a left column
![pic3](https://user-images.githubusercontent.com/44893868/57740676-f7eac700-76f3-11e9-86a6-7f05b50299e0.png)
above instance : split 'a' column with separator '|'

4. you get some unexpected data
above instance : g,h,i is unexpected split data 

**Desktop :**
anywhere, I guess

**OpenRefine :**
 - Version 3.1

**Datasets**
project of above instance
[split-issue.openrefine.tar.gz](https://github.com/OpenRefine/OpenRefine/files/3180421/split-issue.openrefine.tar.gz)

I  made a patch for this issue.

thadguidry:
Hi @stanakaj Thanks for this bug report !  Are you going to be submitting the patch as a Pull Request in Github?

PR #2042, opened 2019-05-15
Fix column removal in reorder leaves undeleted hidden cells.

stanakaj:
fix for #2041


wetneb:
Thanks a lot for this @stanakaj. Sorry about Travis being broken, I am hoping to fix this soon. In the meantime do you think you could write a test to demonstrate the fix?

stanakaj:
I can make a test only for a bottom cause, not whole reported bug. I feel that there are too many things we need to test for whole.

wetneb:
@stanakaj sure, whatever you find easiest!

stanakaj:
@wetneb I appended a small test code.

wetneb:
Note to self:
- [ ] check is that old changes will still be parsed and applied/unapplied correctly after this.

PR #2048, opened 2019-05-24
Adding new Github Security Policy doc

thadguidry:
@wetneb Please review and provide feedback for accuracy and correctness.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,18 @@
+# OpenRefine Security Policy
+
+## Supported Versions
+
+OpenRefine versions being supported with security updates & vulnerability reports.
+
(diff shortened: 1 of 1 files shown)


wetneb:
Given our current workflows we are basically not able to release fixes for any previous version. We would need to use branches for the major versions we want to support on the long term. That's definitely something we can consider. Each security patch would be merged in all supported major release branches to create the corresponding versions. That would probably mean using more granular release numbers (such as 3.2.0).

thadguidry:
Closing since this need to go into an issue as @wetneb says so that we have a plan for supported versions of OpenRefine.
New issue for this: #2288

PR #2049, opened 2019-05-24
Creating new Github Sponsors FUNDING file

thadguidry:
This enables the new Github Sponsors button (we are beta testers of this feature now).  Will work with Open Collective on issues as they come up.  For user payouts, Github is starting with Stripe as the default payment processor.  But doesn't matter if some of us cannot use Stripe as we are beta testing this for contributors for can/cannot and will be providing feedback to Github.  Don't worry, I'll guide us through it and use core mailing list.

I'm super excited to see a Bountysource like system in Github finally becoming a reality!!!! (they have a lot planned for the future around Sponsorship and this is only the beginning)

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,2 @@
+open_collective: openrefine
+github: wetneb, ostephens, tcbuzor, thadguidry


sherria:
Don't know anything about what I'm doint

thadguidry:
> Don't know anything about what I'm doint

Hi @sherria are you a programmer or developer and learning Git?  Or do you have an interest in OpenRefine or use it?

wetneb:
So I guess merging this would amount to designating OpenCollective as our fiscal sponsor. Have we reached a consensus for that? I have recently come across Code for Science and Society (https://codeforscience.org/) which could be a good fit too (and they are a recognized 501(c)3).

thadguidry:
We have not come to consensus yet... and we need to.  We need SOME fiscal host that allows us to be recognized under 501(c)3.  And I would hate to switch or change fiscal hosts every year, so someone/somewhere that has longevity.  It seems that OpenCollective is a smart move for a number of reasons, one of which being that Github Sponsors is working with them even through the audit perspective.

For this merge its safe enough, but we still have to have the backend done with OpenCollective and that's just a bit of emailing with them if we all agree there ?

wetneb:
One of the issues with OpenCollective is that we would not get this 501(c)3 status AFAIK. I am getting in touch with people from Code for Science & Society to see how it would work with them.

thadguidry:
Correct, WE would not get that status without a lot of trouble as we already know, and that is perfectly fine.  You might be misunderstanding.  What we want IS a fiscal host that provides us 501(c)3 -like status.  For donors to us, this gives the advantage that they can still get a tax exemption for their donation to us because the fiscal host gives them a **charitable receipt**, OpenRefine Foundation.  But the 501(c)3 status and processing is done by a **fiscal host**.  It could be "Open Collective Foundation" or possibly "Open Collective Europe ASBL" as the real **fiscal host**, either way there's the 5% fee which is fine and we all agreed is fine.  OpenCollective as a **fiscal accounting platform** gives us that nice transparent ledger and auto reports and approval flows.

AND Code for Science & Society is already setup in OpenCollective :) as a fiscal host. https://opencollective.com/css

Like I said, is a no-brainer now to go with OpenCollective as our **fiscal accounting platform**...and then we just choose a **fiscal host**... which could be the two that I mentioned above...or CS&S as you prefer.  Doesn't matter on the **fiscal host** as long as they can provide US-backed businesses (like Google) with a **charitable receipt**...which OpenCollective gives automatically as a **fiscal accounting platform**.

Hopefully the above makes a bit more sense for you?

wetneb:
Yes I understand how OpenCollective works, the main issue for me was that I would not expect us to be fall into Open Collective Foundation's scope (and the european one is not technically a 501(c)3 since it is not based in the US - not that it would be a problem with me of course).

But I was not aware that CSS is registered there too - it could be a good choice for a fiscal host. Let's keep the discussion going and decide on something with the rest of the team soon.

thadguidry:
Just an update on this... we can't move on this until Joe from CS&S and others give us the OK (they have to ensure they can properly route payments for processing and lack ability like OpenCollective and so it would need OpenCollective as well - an additional 5% fee on top of CS&S 15% fee.)  So this might not be viable until next year 2021.

So that leaves us with small sponsors sending checks to CS&S directly....or clicking on a Donate Button on our Website (which is pending checkup on Joe's plate from last 2 meetings ago).

So until things get coordinated and clearly understood from CS&S, I'm closing this PR because it affords us no opportunity currently.  When things change, I'll open a new PR.

PR #2050, opened 2019-05-24
tweak Coveralls badge

thadguidry:
For some reason I think the whitespacing is causing a Badge detection issue on Coveralls site, but could be wrong and its something else?


wetneb:
@thadguidry badge detection on coveralls is broken, just ignore it.

Issue #2063, opened 2019-06-12
Add labels or descriptions without overriding existing ones


wetneb:
In Wikidata schema, it would be useful to have the possibility of adding a label or description to an item without overriding any existing one.

In batches where both new items and existing items are edited, users are encouraged to add descriptions and labels to the new items they create, but they often end up overriding those terms for the existing items they affect (example: https://tools.wmflabs.org/editgroups/b/OR/b0b128bd2/).

The current workaround is to create a new column which is only non-empty for new items (using facets). Or when adding terms to new items, the current terms need to be fetched first.
This is cumbersome and error-prone.

The downside of this change is that it would make QuickStatements exports less faithful (as QS does not have this feature).

PR #2065, opened 2019-06-13
Wikidata: Add labels or descriptions without overriding existing ones

wetneb:
This adds the possibility of specifying descriptions or labels for items in a Wikidata schema, such that they are only added to the items if no such term exist in the given language. See #2063 for the motivation of the feature.

The candidate UI for this feature looks like this:
image

Is this clear enough? I am not sure. Maybe it would be better to have a checkbox "override existing" on the right hand side of the label?


wetneb:
New version of the UI, much clearer I would say:
image

thadguidry:
Checkbox wins my vote.

PR #2066, opened 2019-06-13
Update to Wikidata-Toolkit 0.10.0.

### LINKED ISSUE #2064: Update Wikidata-Toolkit to 0.10.0
wdtk 0.10.0 was released with a useful bug fix to prevent from editing as an IP, we should include that before 3.2.

wetneb:
Closes #2064.


wetneb:
There is a serialization problem due to https://github.com/Wikidata/Wikidata-Toolkit/pull/394, do not merge yet.

wetneb:
This is blocked by issue https://github.com/Wikidata/Wikidata-Toolkit/issues/417

wetneb:
The bug is not actually critical, I have tried editing with WDTK 0.10.0 and it works fine. Even if I have fixed the bug in master in WDTK, I will stick to the released version of WDTK, to avoid relying on a custom release. I have temporarily disabled the failing test so this can be merged now.

PR #2072, opened 2019-06-19
Issue 608

impune-pl:
Hi @thadguidry, sorry I took so long - I'm in the middle of exams right now, and need to split my time between repeating some material and coding. I wrote function to parse unix time to OffsetDateTime, and some tests for ToDate class. I couldn't figure out test case that would test [lines 165 and 166](https://github.com/impune-pl/OpenRefine/blob/0ecf28174683bf9f7c443abaef72586f81be7789/main/src/com/google/refine/expr/functions/ToDate.java#L165), so I left them untested.

### DIFF
--- a/main/src/com/google/refine/expr/functions/ToDate.java
+++ b/main/src/com/google/refine/expr/functions/ToDate.java
@@ -38,6 +38,7 @@ import java.text.SimpleDateFormat;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.GregorianCalendar;
@@ -61,6 +62,7 @@ public class ToDate implements Function {
     public Object call(Properties bindings, Object[] args) {
         String o1;
         Boolean month_first = null;
+        Boolean unix = false;
         List<String> formats =  new ArrayList<String>();  
         OffsetDateTime date = null;
         
@@ -76,6 +78,8 @@ public class ToDate implements Function {
                 o1 =  ((Long) arg0).toString(); // treat integers as years
             } else if (arg0 instanceof String && arg0.toString().trim().length() > 0) {
                 o1 = (String) arg0;
+            } else if (arg0 instanceof Integer) {
+                o1 =  ((Long)((Integer) arg0).longValue()).toString();
             } else {
                 // ignore cell values that aren't Date, Calendar, Long or String 
                 return new EvalError("Unable to parse as date");
@@ -88,7 +92,12 @@ public class ToDate implements Function {
             if(args[1] instanceof Boolean) {
                 month_first = (Boolean) args[1];
             } else if (args[1] instanceof String) {
-                formats.add(StringUtils.trim((String) args[1]));
+                if(args[1].equals("unix")) {
+                    date = parse(o1);
+                    unix = true;
+                } else {
+                    formats.add(StringUtils.trim((String) args[1]));
+                }
             } else {
                 return new EvalError("Invalid argument");
             }
@@ -101,7 +110,7 @@ public class ToDate implements Function {
             }
             if(month_first != null) {
                 date = parse(o1,month_first,formats);
-            } else {
+            } else if(!unix) {
                 date = parse(o1,formats);
             }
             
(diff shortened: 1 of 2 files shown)


thadguidry:
Oh no problem @impune-pl !  I'd like to have @wetneb review this and @ostephens to test our assumptions here, since he worked on the original thought process with me.

wetneb:
@impune-pl just to make sure we are on the same page: at the moment the pull request fails the tests for recent versions of Java so we cannot merge this until the problem is solved.

main/src/com/google/refine/expr/functions/ToDate.java
```diff
@@ -76,6 +78,8 @@ public Object call(Properties bindings, Object[] args) {
                 o1 =  ((Long) arg0).toString(); // treat integers as years
```
ostephens: Probably should update or remove these comments

main/src/com/google/refine/expr/functions/ToDate.java
```diff
@@ -76,6 +78,8 @@ public Object call(Properties bindings, Object[] args) {
...
             } else if (arg0 instanceof String && arg0.toString().trim().length() > 0) {
                 o1 = (String) arg0;
+            } else if (arg0 instanceof Integer) {
+                o1 =  ((Long)((Integer) arg0).longValue()).toString();
             } else {
                 // ignore cell values that aren't Date, Calendar, Long or String
```
ostephens: This comment also outdated by these changes

main/src/com/google/refine/expr/functions/ToDate.java
```diff
@@ -187,15 +196,26 @@ private OffsetDateTime parse(String o1, DateFormat formatter) {
...
     @Override
     public String getDescription() {
-        return "Returns o converted to a date object, you can hint if the day or the month is listed first, or give an ordered list of possible formats using this syntax: http://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html";
+        return "Returns o converted to a date object, you can hint if the day or the month is listed first, use unix time (with \"unix\" flag) or give an ordered list of possible formats using this syntax: http://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html";
     }
     
     @Override
     public String getParams() {
-        return "o, boolean month_first / format1, format2, ... (all optional)";
+        return "o / integer unix_seconds (only with unix flag), boolean month_first / \"unix\" / format1, format2, ... (all optional) ";
```
ostephens: I don't think " / integer unix_seconds (only with unix flag)" is should be here - it is just another object. This is about the behaviour of the function and should be in the Description not the parameters

main/tests/server/src/com/google/refine/tests/expr/functions/ToDateTests.java
```diff
@@ -26,16 +26,99 @@
...
+    @Override
+    @BeforeTest
+    public void init() {
+        logger = LoggerFactory.getLogger(this.getClass());
+    }
+
+    @Test
+    public void testUnixTime()
+    {
+        Assert.assertEquals((OffsetDateTime)invoke("toDate", 123456l, "unix"), OffsetDateTime.parse("1970-01-02T10:17:36+00:00"));
```
ostephens: I don't think it's necessary, but for consistency would be nice to use "OffsetDateTime.parse("1970-01-02T10:17:36"+ZoneOffset.UTC))" as per other tests

main/src/com/google/refine/expr/functions/ToDate.java
ostephens: Also based on my reading of the code I think integers will be treated as Year if unix flag not set?

ostephens:
Apologies it's taken me so long to review this

impune-pl:
@wetneb, @ostephens thanks for reviews. I'll try to fix everything by the end of this week. No apologies necessary, this issue is around for 7 years, so it can wait a few weeks :-)

main/src/com/google/refine/expr/functions/ToDate.java
impune-pl: Yes. Based on comment in line 78 I think it's a desired behaviour.

main/src/com/google/refine/expr/functions/ToDate.java
thadguidry: Don't remove!  Update!  We actually need more comments in our code base to help explain more for new contributors coming in and make it more attractive to join in helping.

impune-pl:
Hi, @ostephens, I have found something that seems a bit weird to me and I'd like your opinion on it.
When this test
```java 
Assert.assertEquals((OffsetDateTime)invoke("toDate", "02001.July.04 AD 12:08 PM", "yyyyy.MMMMM.dd GGG hh:mm aaa", "1234", "unix"), OffsetDateTime.parse("2001-07-04T12:08:00.00"+ZoneOffset.UTC));
```
 is run, [these lines](https://github.com/OpenRefine/OpenRefine/blob/1064008cabb32cc78338b9947f3d3010604966a1/main/src/com/google/refine/expr/functions/ToDate.java#L145-L152) do something I do not understand.
First it loads two locales, default os locale (in my case pl-PL) and possibleLocale based on argument taken from list of formats (named f1), in this case it loads empty Locale object, with all fields being null, 0 or "", except for BaseLocale baseLocale (which is an object with all fields being "", null or 0).
Then it replaces the correct locale (pl-PL) with the empty one in the first iteration of foreach loop, and proceeds to put the f1 argument back into formats list around 200 times. 
The blank locale is later used to create SimpleDateFormat object.

The problem is that if the first argument after the date is "unix", like here
```java
Assert.assertEquals((OffsetDateTime)invoke("toDate", "02001.July.04 AD 12:08 PM","unix","yyyyy.MMMMM.dd GGG hh:mm aaa", "1234"), OffsetDateTime.parse("2001-07-04T12:08:00.00"+ZoneOffset.UTC));
```
then possibleLocale.baseLocale.language gets initialized with "unix", and possibleLocale.baseLocale.hash changes.
That seems to prevent locale from being replaced with possibleLocale, which leads to creating SimpleDateFormat object with pl-PL locale and EvalError being returned, because "02001.July.04 AD 12:08 PM" is not polish date format, and doesn't get parsed correctly.

Is it correct behaviour, or should I try to work around it?
I have attached current WIP version of ToDate class.
[ToDate.zip](https://github.com/OpenRefine/OpenRefine/files/3439450/ToDate.zip)

Issue #2018, opened 2019-04-12
Add new text facet type


ostephens:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
Add a new facet of type Text (as opposed to current "List" facet which is usually used for text values). This is in order to have a facet that facets on textual values while not attempting to automagically transform non-text value cells into text. The new Text facet should list all cell values that are text, and create buckets (with counts) for non-text values

**Describe the solution you'd like**
- [ ] Add support for a new facet type of "Text"
- [ ] Change option to "Create custom text facet" in menu to use this new facet (rather than the current use of List facet)
- [ ] Convert pre-defined facets to use new Text facet type where appropriate

**Describe alternatives you've considered**
See discussion in #1957 and #1662 and other linked PRs and issues

**Additional context**
See discussion in #1957 and #1662 and other linked PRs and issues

ostephens:
Just to note progress. I didn't manage to get as much time as I hoped in the last week to look at this,  but I did get some time to think through the problem. My plan currently is to:

* Create a new class for the current ListFacet behaviour which extends the ListFacet class 
* Create a new class for a new text fact which extends the ListFacet class
* Shared code between these will continue to live in the ListFacet class. Where there are differences the new classes will override methods to implement the appropriate behaviour

I may do some further refactoring of the ListFacet class once I've done this initial work, but it is this initial work that needs to be completed to get us to a point where we have the current List Facet behaviour and the new Text Facet behaviour supported with a reasonably minimum amount of code duplication.

ostephens:
In terms of actual progress I've written some extended tests for the current List Facet behaviour to make sure this continues to work in the current way

wetneb:
@ostephens what is your timeline for this? Is your current work pushed on some branch? If so maybe I can help?

ostephens:
@wetneb I've got time to look at this over this weekend - If I don't get the work completed, I'll share a branch with what I've done so we can share the work

wetneb:
Feel free to share your branch if you need a hand on this :)

ostephens:
Hi @wetneb - sorry I've been so slow on this.

You can see the progress (or lack of) I've made so far at https://github.com/ostephens/OpenRefine/tree/list-facet-refactor

I think the following commits are good & sensible steps:
https://github.com/ostephens/OpenRefine/commit/02501db87f21e2094cfa62f903b0bc02c0fe3e09
[...]

PR #2078, opened 2019-07-03
List facet refactor

wetneb:
Just creating this pull request to keep track of the work even if it is not ready yet.

### DIFF
--- a/main/src/com/google/refine/browsing/facets/TextListFacet.java
+++ b/main/src/com/google/refine/browsing/facets/TextListFacet.java
@@ -1,0 +1,172 @@
+/*
+
+Copyright 2010, Google Inc.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+    * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+    * Neither the name of Google Inc. nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,           
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY           
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+*/
+
+package com.google.refine.browsing.facets;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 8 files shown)


main/src/com/google/refine/browsing/facets/TextListFacet.java
```diff
@@ -0,0 +1,172 @@
...
+import com.google.refine.browsing.filters.AnyRowRecordFilter;
+import com.google.refine.browsing.filters.ExpressionEqualRowFilter;
+import com.google.refine.browsing.util.ExpressionNominalValueGrouper;
+import com.google.refine.expr.Evaluable;
+import com.google.refine.expr.MetaParser;
+import com.google.refine.expr.ParsingException;
+import com.google.refine.model.Column;
+import com.google.refine.model.Project;
+
+public class TextListFacet extends ListFacet {
```
wetneb: I am not sure it is a good idea to make TextListFacet extend ListFacet. In the current situation, ListFacet is not designed to be extended in this way, which causes the current NullPointerExceptions in the tests.

If the desired end state is to completely replace the original ListFacet with the new TextListFacet, then I think it would be cleaner to actually duplicate the existing code, with the understanding that on the long term the duplication will disappear when we drop ListFacet.

main/src/com/google/refine/browsing/facets/TextListFacet.java
```diff
@@ -0,0 +1,172 @@
...
+    
+    /*
+     * Computed results
+     */
+    protected int _stringCount;
+    protected int _numericCount;
+    protected int _datetimeCount;
+    protected int _booleanCount;
+    
+    TextListFacetConfig _config = new TextListFacetConfig();
```
wetneb: Here is a sign that inheritance is not the right tool for the job: re-declaring a `_config` attribute which was already defined in the base class is dangerous (and the cause of your NPEs). What happens in practice is that the class will have two differerent `_config` fields and methods will access one or another depending on which class they have been defined in.

main/src/com/google/refine/browsing/facets/TextListFacet.java
ostephens: > I am not sure it is a good idea to make TextListFacet extend ListFacet. In the current situation, ListFacet is not designed to be extended in this way, which causes the current NullPointerExceptions in the tests.

As I noted in #2018:
>My original idea was to have an abstract class for a 'list facet' which would support grouping by any of the supported data types then be implemented by different types of list facet including the existing 'list facet' and a new 'text list facet' which would exhibit the new desired behaviour. But each time I look at this I start to think the whole way facets are structured feels not quite right, and end up fluctuating between just doing something v simple but duplicating loads of code, and revisiting the whole structure of the facet code (at which point I have to lie down in a dark room for half an hour).

When I step away from the problem, it feels like having a general 'list facet' class which then could be specialised into types seems like a good idea. Especially when considering the point of this work is to end up with the old 'list facet' behaviour alongside the new 'text list facet' behaviour. However, when I try to implement I hit the problems you note.

I also had in mind this might make it easier to support different types of behaviour in different types of list facet going forward (since discussions in #1957 and #1662 show this is difficult to agree on)

I can't escape the feeling that the code around facets (of all types) could be more efficient and easier to extend, but I suspect I don't have the experience & knowledge to implement this, and I definitely haven't had the time

thadguidry: @ostephens I'd rather you not give up on this, given time this is a great opportunity to learn more!  For me being on the sidelines, and looking at the code and Antonin's comments I have learned more...and learned why I pulled away from Java in the first place long ago ;-)  

Perhaps a bit of guidance on a call with Antonin in a paired session for 30 mins - 1 hour could help OpenRefine and you?  I'll leave it to both of you to decide.  If you strongly feel like you want to pull away from this particular refactoring, then perhaps @wetneb can step in to finish this up?

wetneb:
@ostephens sorry I should have read your explanations more carefully - things make sense now.

I do not see a really clean and completely satisfactory solution at the moment, so the problem requires more thinking. It is indeed trickier than it looks!
My gut feeling is that we should not rush this before a stable release though. The chances that we get it wrong are too high. I will not have a lot of time to work on this in the coming weeks but I will keep thinking about it.

main/src/com/google/refine/browsing/facets/TextListFacet.java
antoine2711: Hi @ostephens, 
> My original idea was to have an abstract class for a 'list facet' which would support grouping by any of the supported data types then be implemented by different types of list facet including the existing 'list facet' and a new 'text list facet' which would exhibit the new desired behaviour.

I've just played a bit with the facets in order to implement the Minimize/Maximize functionality. And, I had to copy the EXACT same code in the 5 Facets JS object because they don't have a generic parent. I'd say I'm kind of compelled to create a common parent, let's say `GenericFacet` that would have all the common functionality of a facet.

So, in this situation, it would have a some benefits for this issue. 1) the ListFacet object would now be ridden of all the common functionality of a facet, so you would have less code to manage and refactorize for this TextListFacet.

Also, it would test all the logic of using a parent class in the context of facets. It would be kind of a practice.

And, last added benefit, in the future, such a common parent would surely help implementing things like saving facets or else… So, would you like to work on this with me?

Regards,
  Antoine

wetneb:
I would propose to close this PR, and instead change the text facet to convert all its non-blank non-error values to strings. This breaking change could be introduced in 4.0.

tfmorris:
Closing per @wetneb's suggestion and @ostephens' thumbs up.

Issue #2044, opened 2019-05-19
Labels for Text filter checkboxes are linked to first instance


yarl:
**Describe the bug**
Labels for checkboxes in text filters are resuing the same `ref`s and checkboxes are reusing the same `id`s

**To Reproduce**
1. Add multiple Text filters

**Current Results**
https://streamable.com/forh6

wetneb:
Thanks for the report! If you are interested in submiting a pull request for that I am available for any help you might want :) The relevant lines are here:
https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/scripts/facets/text-search-facet.js#L104-L105

mtricht:
@wetneb I believe this issue can be closed

PR #2084, opened 2019-07-11
Add a unique id to text filter facet checkboxes and labels.

mtricht:
My attempt to tackle https://github.com/OpenRefine/OpenRefine/issues/2044


wetneb:
Many thanks for this and sorry for the broken build, I have restarted travis. It looks great to me, it should make it into the coming release.

PR #2094, opened 2019-07-20
[DRAFT] Add recon and record to Help tab

thadguidry:
Began documenting missing Variables in Help tab within Expression Dialog.


thadguidry:
@wetneb you think the contained links I added are "good" or "bad" ?  I feel "good" since it's widely asked about on mailing list, and it doesn't crowd HELP too much and still view-able/scroll-able easily.

wetneb:
@thadguidry adding a documentation for `record` is good - it is a variable that is indeed available in the context. But I would be wary of adding `recon` as you did: it gives the impression that this is also a top-level variable while it is not: it is an attribute of the `cell` variable. So I would leave it out.

thadguidry:
yeah, I noticed that also, and both are mentioned on `cell` help.

PR #2095, opened 2019-07-21
Fix Find literal

### LINKED ISSUE #1833: In GREL `find` function treat quoted argument as literal string
**Describe the bug**
See discussion at https://groups.google.com/forum/#!topic/openrefine/bYOG-bn_2xw
@msaby reported:
> The GREL find() function can take a string as a parameter, but it seems the string is "translated" into regex.

**To Reproduce**
1. find ("ab", ".b")   --> ["ab"]  (because the . is not interpreted as a literal . but as a regex)

**Expected behavior**
The  dot '.' in the above expression should be treated as a literal character, not a regular expression dot. Hence expected behaviour is:

find("ab",".b") --> [ ]
find("ab",/.b/) --> ["ab"]

ostephens:
Fixes #1833


thadguidry:
Whoohooo !  Thanks @ostephens  !

msaby:
Thank you !

PR #2096, opened 2019-07-22
Use environment variables to detect Java location on Windows.

### LINKED ISSUE #2069: Windows version only works with Java 8 installed
It is reported by users that running OpenRefine 3.1 on Windows only works if Java 8 is installed.

This is probably related to the way Launch4J searches for the Java Runtime Environment: it relies on registry keys, which are no longer used by recent releases of Java:
https://sourceforge.net/p/launch4j/feature-requests/127/?page=0

The maintainer of Launch4J encourages to ship a bundled JRE in the releases. We already do that for Mac OS, we could consider doing that for Windows too.

wetneb:
This should hopefully fix #2069. By default launch4j only uses registry keys to locate Java installs on Windows and it appears that environment variables are now preferred.

This behaviour will be changed in the upcoming release but in the meantime it seems that this might do the trick. I do not have access to a Windows machine so I hope others will be able to try out this snapshot:
http://pintoch.ulminfo.fr/aaa8156eae/openrefine-win-3.2-SNAPSHOT.zip


thadguidry:
@wetneb double-clicking the openrefine.exe just starts the console window and then quickly closes for unknown reasons.  I suspect its not quite right yet.   My further suspicion is that launch4j is looking at PATH ? or not ?

```shell
C:\Users\THAD>java -version
java version "1.8.0_221"
Java(TM) SE Runtime Environment (build 1.8.0_221-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.221-b11, mixed mode)

C:\Users\THAD>echo %PATH%
E:\Python27\;E:\Python27\Scripts;E:\Python37\;E:\Python37\Scripts;C:\ProgramData\DockerDesktop\version-bin;C:\Program Files\Docker\Docker\Resources\bin;C:\Program Files (x86)\Common Files\Oracle\Java\javapath;C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v9.1\bin;C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v9.1\libnvvp;C:\Program Files\Java\jdk-11.0.1\bin;C:\Program Files (x86)\Intel\iCLS Client\;C:\Program Files\Intel\iCLS Client\;C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\Program Files (x86)\Intel\Intel(R) Management Engine Components\DAL;C:\Program Files\Intel\Intel(R) Management Engine Components\DAL;C:\Program Files (x86)\Intel\Intel(R) Management Engine Components\IPT;C:\Program Files\Intel\Intel(R) Management Engine Components\IPT;C:\Program Files (x86)\NVIDIA Corporation\PhysX\Common;C:\Users\THAD\.dnx\bin;C:\Program Files\Microsoft DNX\Dnvm\;C:\Program Files\Microsoft SQL Server\130\Tools\Binn\;C:\WINDOWS\System32\OpenSSH\;C:\Program Files\Microsoft VS Code\bin;C:\Program Files\NVIDIA Corporation\NVIDIA NvDLISR;E:\nodejs\;E:\Go\bin;C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\WINDOWS\System32\OpenSSH\;C:\Program Files\Git\cmd;E:\Downloads\lein;C:\Users\THAD\AppData\Local\Microsoft\WindowsApps;E:\Downloads\apache-maven-3.5.4-bin\apache-maven-3.5.4\bin;C:\Program Files\CMake\bin;C:\Program Files\Microsoft VS Code\bin;C:\Users\THAD\AppData\Roaming\npm;E:\Downloads\go-ipfs_v0.4.19_windows-amd64\go-ipfs;C:\Users\THAD\go\bin;E:\msys64\usr\bin;F:\flutter\bin;C:\Users\THAD\AppData\Local\GitHubDesktop\bin;

C:\Users\THAD>
```

wetneb:
@thadguidry it checks for `PATH` but in principle it should also check for `JAVA_HOME`. Can you check the value of that `JAVA_HOME` variable on your machine?

Anyway if it is not set on yours, it will not be set on others and that really needs to work out of the box…

thadguidry:
As you can see in the attached support.log that we now create ;-) , my JAVA_HOME is set correctly.
[support.log](https://github.com/OpenRefine/OpenRefine/files/3419166/support.log)

Let me give you the full log output from the console with `refine.bat` just as a reference:
[openrefine-3.2-SNAPSHOT-THAD.log](https://github.com/OpenRefine/OpenRefine/files/3419173/openrefine-3.2-SNAPSHOT-THAD.log)

wetneb:
@thadguidry can you try this one? It should work with your Java 8:
http://pintoch.ulminfo.fr/fc1011d173/openrefine-win-3.2-SNAPSHOT.zip
Ideally it would be great if you could try uninstalling Java 8 entirely and installing Java 10, 11 or 12, and see if it works too.

packaging/pom.xml
```diff
@@ -95,6 +194,25 @@
         <groupId>com.akathist.maven.plugins.launch4j</groupId>
         <artifactId>launch4j-maven-plugin</artifactId>
         <version>1.7.25</version>
+
+            <!-- This is a clean snapshot of launch4j's master branch,
```
thadguidry: Nice. OK, this makes sense now.  Was wondering the WHY, but this comment from you tells me that in the detail I was curious about. Thanks for this comment!  Let's do more of this!

thadguidry: Testing your provided binary now, just downloaded...

thadguidry: `openrefine.exe` started it just fine now and that was with my JAVA_HOME env variable pointing to Java 8.

One curious thing however that is different now... hitting CTRL-C in the console is exiting immediately and doesn't seem to be doing the Project Save prior to the shutdown?
Let me try other Java versions and also using the `refine.bat` and testing a few other things.

thadguidry: If I don't have Java installed...when starting `openrefine.exe` it opens my browser and takes me to https://www.java.com/en/download/
Perhaps this needs to be changed in the future after December 2020 when Oracle finally stops Java 8 patches?
or we can change it now to something better like https://jdk.java.net ?

It works fine with Java 11 and 12 where I downloaded and use OpenJDK and set my JAVA_HOME.
Annotation 2019-07-23 215914

packaging/pom.xml
```diff
@@ -114,7 +233,7 @@
               <jre>
                 <path>%JAVA_HOME%;%PATH%</path>
                 <minVersion>1.8.0</minVersion>
-                <jdkPreference>preferJre</jdkPreference>
+                <bundledJreAsFallback>true</bundledJreAsFallback>
```
thadguidry: Wait what? are we bundling the JRE now with Launch4j or no?

wetneb: No, it's a misleading option name. We do not ship the JRE, it's just that launch4j's way to inspect environment variables to look for Java relies on looking at a path containing environment variables, which is normally used to locate an embedded JRE. The launch4j maintainer is aware of this discrepancy.

packaging/pom.xml
wetneb: About changing to https://jdk.java.net/, the main downside I can see is that users will download a zip file and will have to set up their JAVA_HOME manually, whereas Oracle's Java comes with an installer.

I think it should not be a problem for us if Oracle no longer provides Java 8 support - people will just download newer versions of Java, which should be compatible with OpenRefine. Java 12 works fine for now except for Jython deprecation warnings, but that should be fixed by the upcoming release of Jython, which should hopefully happen before the cutoff date for Java 8.

wetneb:
Thanks a lot for the thorough testing Thad!
Yes I can confirm that launch4j closes the Java process gracefully (it waits for it to terminate and passes on its exit code). So it should be safe save-wise.

packaging/pom.xml
thadguidry: Thanks for clarifying.

packaging/pom.xml
thadguidry: Yes, I saw the ease of use with Oracle's installer but also opinionated for power users like me.  That's fine, the world needs easy installers as well as manual installation options.  Having both is a good thing in the Java ecosystem on Windows.

Hopefully the question of WHICH Java and WHERE to download gets a bit easier as time moves on next year.  God I hope so!

I am doubtful about Jython progress, its not moving and has not for quite some time because of lack of funding.  GraalVM might make it unnecessary by 2020...we shall see.

PR #2098, opened 2019-07-23
Wikidata: Use distinct tabs for QS and schema export.

### LINKED ISSUE #2097: Schema not exportable while running QuickStatements
**Describe the bug**
When you are running a QuickStatements job and want to back-up your schema for preparing  another one batch edit, you'll just get the QuickStatements windows focused instead of a local schema backup file.

**To Reproduce**
1. Open a QuickStatements window
2. Click on 'Export schema' in OpenRefine
3. The QuickStatements window gets focused instead of exporting the schema

**Current Results**
The QuickStatements window gets focused.

**Expected behavior**
The schema gets exported.

 - 3.2-beta [8d89a2a]

wetneb:
Closes #2097.

@sjoerddebruin could you try out this version to see if it works for you?
```
git clone https://github.com/OpenRefine/OpenRefine
cd OpenRefine
git checkout issue-2097-wikidata-export
./refine
```
(assuming you use Linux and have Maven installed)

### DIFF
--- a/extensions/wikidata/module/langs/translation-en.json
+++ b/extensions/wikidata/module/langs/translation-en.json
@@ -6,6 +6,7 @@
     "wikidata-extension/perform-edits-on-wikidata": "Upload edits to Wikidata",
     "wikidata-extension/export-to-qs": "Export to QuickStatements",
     "wikidata-extension/export-schema": "Export schema",
+    "wikidata-extension/export-wikidata-schema": "Export Wikidata schema",
     "wikidata-extension/quickstatements-export-name": "QuickStatements",
     "wikidata-schema/dialog-header": "Align to Wikidata",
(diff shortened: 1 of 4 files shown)


thadguidry:
@wetneb Personally, I like this better.  Less confusing :)
I'd also improve the label from "export schema" to "export Wikidata schema" ? To avoid confusion from other extensions that might customize menu also.

wetneb:
I think it is fairly clear that it applies to Wikidata given that it is in the corresponding menu… I would agree if the action was in the general "Export" menu but that is not the case at the moment.

thadguidry:
Umm, I see it in the general "Export" menu when pulling this change and seeing the menu... does it not look like this on yours ?
Annotation 2019-07-23 173733

I just don't see anything saying "Wikidata" when I clicked on our regular Export button, but I do see the subtle grey line.  To a user, they wouldn't know that the last section under that grey line is "Wikidata section"... make sense?

wetneb:
Oops, sorry about that! Totally makes sense, I will fix that.

sjoerddebruin:
I can't try it out locally, but I agree that it is expected behavior. Also thanks for the label change.

Issue #1724, opened 2018-09-06
Transform All Columns


allthebits:
When Transforming ALL rows of data:

![image](https://user-images.githubusercontent.com/2952327/45162956-2e636000-b1bd-11e8-9ed5-a0f4685fd872.png)

add a checkbox in the popup that allows you to "select all columns" to apply the transformation to all rows of data and all columns of data regardless of the column name, preventing the column names from being statically defined within the JSON representation of the transformation.

This request is based on the following discussion thread from the [OpenRefine google group](https://groups.google.com/forum/#!topic/openrefine/C3W-besDx4U)

thadguidry:
Done !  This was released as part of OpenRefine 2.7 !  Closing this issue now.

https://github.com/OpenRefine/OpenRefine/commit/20acc17c05c3d76a643908fe8918d788abc2cf6f#diff-350f28f0f68f7da753d1e8b086362e94

ostephens:
I think it would be an additional option on this screen:

<img width="822" alt="screenshot 2018-11-21 at 15 31 12" src="https://user-images.githubusercontent.com/576174/48851277-8401de80-eda2-11e8-8f63-11f618222267.png">

thadguidry:
I still have a need for this... a few datasets I imported recently, I needed to apply a Transform to all the columns except column 1 & 2.  Painful without the Column selector idea above from @ostephens to interactively ask which columns to apply the Transform expression on.

@wetneb I think this is probably a good candidate feature to prioritize, it comes up too often and would eliminate a lot of tedious user clicking currently that's really unnecessary if we added the column selector to All -> Transform.

PR #2110, opened 2019-08-04
Add menu for joining columns

msaby:
This patch adds a new "Join columns" entry in "Edit column" menu.

image

The dialog box looks like that:

image

To test : create a project with at least 10 columns and different type of values (strings, blank strings, numbers, dates, booleans, null) and play with the options.

Ex:

image

Under the hood it is a wrapper around GREL "join()" function.

For the UI, I did not keep the "traditional" layout of OR made with tables (gave me headache). Instead I used divs and custom css.

For me everything looks ok EXCEPT for the "delete joined columns" option. It seems that this code is wrong. It raises error message in alert box (I put them in comments), as if the successive Refine.postCoreProcess were "treading on there toes". Could someone help please?

```
columnsToJoin.forEach (function (columnName) {
        // FIXME ERREUR dans Refine.postProcess (project.js l. 359) : 
         // Index: 5, Size: 5
        // Index: 6, Size: 5
        // No column named e
        Refine.postCoreProcess(
          "remove-column", 
          {
            columnName: columnName
          },
          null,
          { modelsChanged: true }
        );
    });
    }
```


main/webapp/modules/core/langs/translation-en.json
```diff
@@ -680,6 +681,18 @@
     "core-views/use-values-as-identifiers2": "Mark cells as reconciled with their values as identifiers",
     "core-views/choose-reconciliation-service": "Choose a reconciliation service",
     "core-views/choose-reconciliation-service-alert": "Please choose a reconciliation service first.",
+    "core-views/column-join": "Join columns",
+    "core-views/column-join-before-column-picker": "Select and order columns to join",
+    "core-views/column-join-before-options": "Select options",
+    "core-views/column-join-field-separator": "Separator between the content of each column :",
+    "core-views/column-join-field-separator-advice": "Enter one or more character, or keep blank to juxtapose the columns.",
```
thadguidry: grammar improvement, replace "character" with "characters".
Also I think "juxtapose" is an often confusing term and translations will be more coherent if we use "append" or even just "connect" instead of "juxtapose".

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -680,6 +681,18 @@
...
+    "core-views/column-join-replace-nulls": "Replace nulls with...",
+    "core-views/column-join-skip-nulls": "Skip nulls.",
+    "core-views/column-join-replace-nulls-advice": "Enter one or more character, or keep blank to replace nulls with blank strings.",
```
thadguidry: grammar improvement, replace "character" with "characters"

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -680,6 +681,18 @@
...
+    "core-views/column-join-skip-nulls-advice": "Skipping nulls will make impossible the reverse the join.",
```
thadguidry: grammar improvement, "Skipping nulls will make it impossible to reverse the join."
But I am confused why we would mention this, since we have Undo/Redo for users to use?

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -680,6 +681,18 @@
...
+    "core-views/column-join-write-active-column": "Write result in active column.",
```
thadguidry: change word choice from "active" to "selected".  This will help with our future UI changes anticipated for multi-column support.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -680,6 +681,18 @@
     "core-views/use-values-as-identifiers2": "Mark cells as reconciled with their values as identifiers",
     "core-views/choose-reconciliation-service": "Choose a reconciliation service",
     "core-views/choose-reconciliation-service-alert": "Please choose a reconciliation service first.",
+    "core-views/column-join": "Join columns",
+    "core-views/column-join-before-column-picker": "Select and order columns to join",
+    "core-views/column-join-before-options": "Select options",
+    "core-views/column-join-field-separator": "Separator between the content of each column :",
```
thadguidry: Let's simplify the grammar here to "Separator to use between each column"

main/webapp/modules/core/langs/translation-en.json
msaby: ok for connect

main/webapp/modules/core/langs/translation-en.json
msaby: ok

main/webapp/modules/core/langs/translation-en.json
msaby: You are right! What I wanted was to describe the effects of join (["a",null,null,"b"],";") --> "a;b" in a few words, and I am not sure that "Skip nulls" is clear enough...
I believe that most use cases require the "coalesce" option, to keep trace of nulls, but in some cases it could be usefull to skip nulls.
Any idea of a better sentence? If not I will remove it, as it is purely informative.

main/webapp/modules/core/langs/translation-en.json
msaby: ok. Are you refering to an open issue?

main/webapp/modules/core/langs/translation-en.json
msaby: ok

msaby:
Thanks! I will submit a new version in a few days.
What annoys me the most is the bug with the successive calls to Refine.postCoreProcess(). When you reapply the history, you can smoothly remove 10 columns or more in less than 1 second. So why a succession of calls in a script is raising errors?

main/webapp/modules/core/langs/translation-en.json
thadguidry: I would remove it.  The effects of join will show quite clearly, and if they user didn't like what they saw, then they simply undo.

main/webapp/modules/core/langs/translation-en.json
thadguidry: Yes multi-column operations are captured as an enhancement in multiple issues as well as our project plans.

thadguidry:
@msaby I don't know why, sorry.  Perhaps @wetneb knows.

msaby:
I also tried to wrap the call with a setTimeout of 2 seconds, but has no effect...

msaby:
(By the way I also thought of a latter improvement : join ALL columns regardless to their name and number, with a forEach() loop!. This could improve reproducibilty. I think this is the same idea as https://github.com/OpenRefine/OpenRefine/issues/1724 )

thadguidry:
@msaby maybe we should have an "all" functor.  `all.cells.value`  But we handle this via fields currently and so it looks like `cells.all.value`.  I'm not sure what @wetneb has in mind for our support of this idea perhaps even using our existing [ColumnGroup](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/model/ColumnGroup.java) to represent that proposed "all" functor ?

msaby:
@ettorerizza gave this solution in a message to the support list : 
```
forEach(
forEach(row.columnNames, e, row.record.cells[e].value), c, c.join(" ")
).join("|")
```

main/webapp/modules/core/scripts/views/data-table/menu-edit-column.js
```diff
@@ -320,14 +334,204 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
+      }
+      // delete joined columns
+    if (self.deleteJoinedColumns) {
+      // do not delete the active cell if it contains the result
+      if (self.writeOrCopy !="copy-to-new-column") {
+        columnsToJoin = columnsToJoin.filter(function(item) {
+          return item !== column.name
+      });
+      }
+      columnsToJoin.forEach (function (columnName) {
```
wetneb: How about using just one operation to remove all the necessary columns at once? The `column-reorder` operation can do this.

About the nature of the error: due to the asynchronous nature of Javascript, making HTTP requests in a loop like here tends to be risky, as they will all be fired at the same time (so you do not control in which order they will reach the server). It is better to pass `postCoreProcess` a callback function which will continue making the next calls once the first one has completed. This also holds for the first call above (to `add-column`): you should wait for this HTTP request to complete before deleting the other columns. I suspect that is the reason why you are having issues here: some column removal operation might be executed before the join itself.

You can pass a callback to `Refine.postCoreProcess` as follows:

```javascript
var onJoinSuccess = function() {
    // Do other calls to Refine.postCoreProcess to remove columns here
}

var onJoinError = function(e) {
    // Report to the user that we could not join the columns for some reason (given in e)
}

Refine.postCoreProcess("add-column", parameters,
   { expression: self.expression },
   { modelsChanged: true },
   { onDone:onJoinSuccess, onError:onJoinError });
```

msaby: Thanks! I did not realized that the HTTP requests between frontend and backend were managed asynchronously like an ordinary web request

msaby:
An other point : Should I use simply **var variable** instead of **self.variable** ?

wetneb:
For local variables inside a function, I would use `var variable`.

msaby:
The removal seems to work with that recursive code. Is it ok for you ?
The array columnName is "shifted" at each step.


```
      var columnName = "";
      var onJoinError = function(e) {
        alert (e);
      }
      var recursiveRemoval = function() {
        if (columnsToJoin.length > 0) {
          columnName = columnsToJoin.shift();
          Refine.postCoreProcess(
              "remove-column", 
              {
                columnName: columnName
              },
              null,
              { modelsChanged: true },
              { onDone:recursiveRemoval, onError:onJoinError }
            );
          }
       }
      recursiveRemoval ();
```

wetneb:
Nice! I think it could potentially be cleaner to remove them all in one operation (to get a more compact history), but it's not a hill I will die on!

msaby:
How would you remove them all in one operation @wetneb ?
Here is a new commit with corrections.

msaby:
I'm trying to reuse code from column-reordering-dialog.js in order to remove all columns in one step.

In the process I found something disturbing : 

In the function used for removing columns, I begin to retrieve all the colums of the project : 

`var columnsToKeep = theProject.columnModel.columns.map (function (col) {return col.name;})
`

As this function is called in a callback in the  "Refine.postCoreProcess("add-column"...)" or the "doTextTransform(column.name,...", I thought theProject.columnModel.columns will contain the column created by "Refine.postCoreProcess("add-column"...)". It seems it is not the case...

I looked at project.js to understand. Refine.postProcess make this POST call : 

```
$.post(
    "command/" + moduleName + "/" + command + "?" + $.param(params),
    body,
    onDone,
    "json"
);
```

So if I understand well the callback (onDone) is called after the POST request is made, but there is no assurance that the Java backend process is completely executed when the callback is called.

Or maybe am I missing something?

wetneb:
@msaby I think for short-running operations (basically everything except reconciliation and fetching urls), the HTTP request completes after the operation is fully executed in the backend. But that does not necessarily mean that the browser has fetched the updates entirely! You might need to either trigger these updates manually, or use another function than `postCoreProcess` (or with another options) to ensure that your callback is triggered only once the UI has been updated.

msaby:
I have digged in project.js. 
In this block of Refine.postProcess we begin by calling custom callbacks, then we update the UI. The new value of theProject is gotten from the backend by Refine.update (which calls createUpdateFunction which calls reinitializeProjectData ) 


   ```
 } else {
      if ("onDone" in callbacks) {
        try {
          callbacks.onDone(o);
        } catch (e) {
          Refine.reportException(e);
        }
      }

      if (o.code == "ok") {
        Refine.update(updateOptions, callbacks.onFinallyDone);

        if ("historyEntry" in o) {
          ui.processPanel.showUndo(o.historyEntry);
        }
      } else if (o.code == "pending") {
        if ("onPending" in callbacks) {
          try {
            callbacks.onPending(o);
          } catch (e) {
            Refine.reportException(e);
          }
        }
        ui.processPanel.update(updateOptions, callbacks.onFinallyDone);
      }
    }
```

So even if I wait 1 day, theProject will not update, but it does not mean the backend has not finished to process the request ;-)
I've seen that there is a callback "onFinallyDone" that seems to behave differently from "onDone". I'm going to try to use it instead. It should be triggered after the UI is updated.

msaby:
Here is a 3rd commit

As suggested by  @wetneb I now remove all columns in one operation.

I also fixed an issue I found with backspace and quotes in column names and separator and replacement character, and added an option for interpreting \n and \t as newlines and tab (same logic as the replacement menu I wrote last year)

Result:

image

image

wetneb:
@msaby quick follow up on this: I have found myself using OR for a project and your Join operation was exactly what I needed, thanks a lot! It works great.

PR #2129, opened 2019-08-21
Fix typos in German translation

katrinleinweber:


main/webapp/modules/core/langs/translation-de.json
```diff
@@ -191,7 +191,7 @@
     "core-dialogs/cells-of-row": "Die Zellen der aktuellen Zeile. Dies ist eine Abkürzung für 'row.cells'. Eine bestimmte Zelle kann mit 'cells.<column name>' abgerufen werden, wenn der <column name> ein einzelnes Wort ist, oder andernfalls mit 'cells[\"<column name>\"].",
     "core-dialogs/for-include-drop-statement-checkbox": "Drop-Anweisung einbinden",
     "core-dialogs/ngram-radius": "Radius&nbsp;",
-    "core-dialogs/cell-fields": "Die aktuelle Zelle. Es hat einige Felder: 'Wert' und 'Aubstimmung'.",
+    "core-dialogs/cell-fields": "Die aktuelle Zelle. Es hat einige Felder: 'Wert' und 'Abstimmung'.",
```
katrinleinweber: I'm unsure about this one. The [original says "recon"](https://github.com/OpenRefine/OpenRefine/blob/3dc18eb89e28d6a96267f1c438ca8634f640b7a8/main/webapp/modules/core/langs/translation-en.json#L215) which seems very different from ["Abstimmung"](https://www.linguee.de/deutsch-englisch/search?source=deutsch&query=Abstimmung).

wetneb: @katrinleinweber thanks for this! Given that these field names are part of the GREL language, they are not translated at all, so we should keep "value" and "recon" there (they are basically language keywords, just like `if` or `while`).

katrinleinweber:
You're welcome, and thanks for merging :-)

Issue #2133, opened 2019-08-23
Reorganize test classes into correct packages


wetneb:
Currently, our tests are stored in their own package: for instance, `EvalErrorTests` lives in package `com.google.refine.tests.expr`, while the class it tests, `EvalError`, lives in `com.google.refine.expr`.

This does not follow established practice of using the same package name for the test class. Test classes are already isolated from main classes by using a different source directory for them. This also makes it harder to test `protected` methods or subclasses since the test class cannot access those. (See https://stackoverflow.com/questions/2388253/separation-of-junit-classes-into-special-test-package for more motivation).

So I think this should be reorganized to follow established practice. This change of namespace could potentially break extensions which rely on our test classes for their own tests, but that namespace change should be easy to fix for them (for instance the RDF extension would only have to adapt the namespace mentioned in https://github.com/stkenny/grefine-rdf-extension/blob/orefine/src/test/resources/testng.xml I think).

There are other issues with our tests, like having a mix of JUnit and TestNG tests, which should be cleaned up, and duplication of test classes for some classes.

PR #2134, opened 2019-08-23
Clean up test classes and packages

wetneb:
This does the cleanup described in #2133. No functional changes in this PR, just shuffling things around.


wetneb:
Merging relatively quickly since this is likely to conflict with many other pull requests impacting the backend.

Issue #2130, opened 2019-08-22
Add autofocus tag to separatorInput fields


katrinleinweber:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
I'm always frustrated when I need to move the mouse to the input field I'm using most often in a dialog. E.g. the `separatorInput` field in `Join...` or `Split multi-valued cells`. It would be easier if OpenRefine autofocussed the cursor onto that field, and others where similar user preference is expected.

**Describe the solution you'd like**

Apply [att_input_autofocus](https://www.w3schools.com/tags/att_input_autofocus.asp) to those fields. If you agree, I'll be happy to send a PR.

**Describe alternatives you've considered**
none, TBH

**Additional context**
Do you have numbers on which for example the popularity of the `by field lengths` split option?

wetneb:
Given that OpenRefine runs locally we don't have any way to collect usage statistics (we could add calls to some hosted service for that, but I think it would be contentious). But your proposal definitely makes sense, we can use our own experience and common sense to determine which fields should be focused by default.

PR #2139, opened 2019-08-26
Auto-focus input fields into which users likely type immediately

katrinleinweber:
As touched on in #2130, here are a few suggestions to start. Please feel free to review already, and change as you see fit.

~~I haven't tested nor built this, yet, and I can't promise to spend as much time setting up the prerequisites as might be necessary. Will try, though.~~

### DIFF
--- a/main/webapp/modules/core/scripts/project/history-panel.html
+++ b/main/webapp/modules/core/scripts/project/history-panel.html
@@ -7,7 +7,7 @@
   <a class="button button-pill-left" href="javascript:{}" bind="extractLink"></a><a class="button button-pill-right" href="javascript:{}" bind="applyLink"></a>
 </div>
 <div class="history-panel-filter" bind="bodyControlsDiv">
-<span bind="or_proj_filter"> </span><input type="text"  size="30" bind="filterInput" id="filterInput" />
+<span bind="or_proj_filter"> </span><input type="text"  size="30" bind="filterInput" />
 </div>
 <div class="history-panel-body" bind="bodyDiv">
   <div class="history-past" bind="pastDiv">
--- a/main/webapp/modules/core/scripts/project/history-panel.html
+++ b/main/webapp/modules/core/scripts/project/history-panel.html
@@ -7,7 +7,7 @@
   <a class="button button-pill-left" href="javascript:{}" bind="extractLink"></a><a class="button button-pill-right" href="javascript:{}" bind="applyLink"></a>
 </div>
 <div class="history-panel-filter" bind="bodyControlsDiv">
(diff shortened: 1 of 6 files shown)


wetneb:
Thanks for this! It seems that adding `autofocus` to the HTML elements in the templates does not do anything on my side. I suspect that is due to the way they are loaded and added to the DOM in JavaScript. I think you could try focusing these elements directly in JavaScript when the dialogs are shown. That would also let you control the focus in a finer way too (if two of these widgets happen to be shown on screen at the same time).

katrinleinweber:
OK, will try to implement https://stackoverflow.com/questions/17500704/

katrinleinweber:
Rather crude. Needs refactoring, maybe towards [attaching the `focus()`-`<script>` to a class](https://stackoverflow.com/questions/17500704/javascript-set-focus-to-html-form-element#comment55521723_17500718). Maybe a new one for all `<input>` elements that should be `autofocus`ed?

wetneb:
Nice! But I think it would be cleaner to reuse the existing JS code where these dialogs are shown. To find the appropriate location you can just search for the HTML filename in JS code. For instance, to figure out the location of the corresponding JS code for the Split column dialog, you can search for `split-column-dialog.html`.

katrinleinweber:
I'm not happy with the duplication, but I don't know enough JS/HTML/CSS details to find the correct placement. Somewhere in the definition of `DialogSystem`, maybe?

wetneb:
Thank you! Before worrying about duplication, have you tried the code? It does not seem to work for me (I am getting `TypeError: elmts.getElementById is not a function` in the console).

You can take inspiration from the code just above to retrieve a DOM element in this context:
https://github.com/OpenRefine/OpenRefine/blob/200d11cbef89904384f483598c6a11b5500b666b/main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js#L303

The way this works is by giving a `bind="filterInput"` attribute to the HTML element in the template, and then accessing it as `elmts.filterInput`. This avoids relying on global ids (if a dialog is opened twice, ids would not be unique).

So I would remove your `bind="filterInput"` and simply do: `elmts.filterInput.focus();`.

Concerning the duplication, I would say this is fine, I don't really see how to get rid of it in a generic way to be honest.

katrinleinweber:
> have you tried the code?

No. I worked on a Windows machine before and `refine.bat` ran into some errors during downloading dependencies. On a Mac now and it builds :-)

katrinleinweber:
OK, this works as I intended. Thanks for your patient advice!

I poked around some more dialogs, but didn't find any that would obviously benefit from the same change. Would you suggest more/others?

About merging, please feel free to squash ;-)

katrinleinweber:
Phew :-D

PR #2140, opened 2019-08-27
Fix scatterplot drawing command

### LINKED ISSUE #2117: Scatterplot
The plots are not displaying when I do a scatterplot facet. Using the current edition of OR. Tried it in both Firefox & Mozilla on a Windows 10 Machine (scatterplots failed to display in both browsers, though the failure to display looked different in the different browsers). I tried it with two different data sets, on three different machines.

wetneb:
closes #2117.

This is a pretty important bug so I think we should consider making a release fairly soon.


wetneb:
It looks like this does not fully solve the issue (plus Travis is failing due to these unmocked HTTP tests…)

PR #2151, opened 2019-09-09
Database Extension: input validation for database connection data

tcbuzor:
Validate Connection input data from frontend and backend. Sanitize database connection information before adding to the DOM.

### DIFF
--- a/extensions/database/module/scripts/index/database-source-ui.js
+++ b/extensions/database/module/scripts/index/database-source-ui.js
@@ -127,11 +127,11 @@ Refine.DatabaseSourceUI.prototype.attachUI = function(body) {
  this._elmts.saveConnectionButton.click(function(evt) {
       
        if(self._validateNewConnectionForm() == true){
-               var connectionNameInput = $.trim(self._elmts.connectionNameInput[0].value);
+            var connectionNameInput = $.trim(self._elmts.connectionNameInput[0].value);
             if (connectionNameInput.length === 0) {
                 window.alert($.i18n('database-source/alert-connection-name'));   
             } else{
-                    self._saveConnection(self._getConnectionInfo());
+                self._saveConnection(self._getConnectionInfo());
             }
              
        }
@@ -431,6 +431,9 @@ Refine.DatabaseSourceUI.prototype._getConnectionInfo = function() {
          var self = this;
         var jdbcConnectionInfo = {};
          jdbcConnectionInfo.connectionName = $.trim(self._elmts.connectionNameInput[0].value);
+         
+        // window.alert('input:' + jdbcConnectionInfo.connectionName + ' output:' + self._removeScriptTag(jdbcConnectionInfo.connectionName) );
+         
          jdbcConnectionInfo.databaseType = $.trim(self._elmts.databaseTypeSelect[0].value);
          jdbcConnectionInfo.databaseServer = $.trim(self._elmts.databaseHostInput[0].value);
          jdbcConnectionInfo.databasePort = $.trim(self._elmts.databasePortInput[0].value);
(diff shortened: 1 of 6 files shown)


extensions/database/module/scripts/index/database-source-ui.js
```diff
@@ -454,21 +469,45 @@ Refine.DatabaseSourceUI.prototype._validateNewConnectionForm = function() {
...
+        	return false;
+        }else if(connectionNameInput.contains("<script>") || connectionNameInput.contains("<img>")){
+        	window.alert($.i18n('database-source/alert-conn-name-invalid-character'));
+        	return false;
+        }else if (databaseHostInput.length === 0) {
             window.alert($.i18n('database-source/alert-server'));
             return false;
+        }else if(databaseHostInput.contains("<script>") || databaseHostInput.contains("<img>")){
+        	window.alert($.i18n('database-source/alert-db-host-invalid-character'));
+        	return false;
```
wetneb: Rather than doing this validation by checking for dangerous HTML tags, I think it would be simpler to just make sure the connection name (and other parameters) are HTML-escaped before being rendered to the DOM. This way, the user cannot insert any HTML tags.

extensions/database/src/com/google/refine/extension/database/cmd/SavedConnectionCommand.java
```diff
@@ -126,25 +132,36 @@ private void writeSavedConnectionResponse(HttpServletResponse response, Database
             writer.writeArrayFieldStart(DatabaseUtils.SAVED_CONNECTION_KEY);
             
             writer.writeStartObject();
-            writer.writeStringField("connectionName", savedConnection.getConnectionName());
+            
+            String sanitizedConnName = Encode.forHtml(savedConnection.getConnectionName());
```
wetneb: Intuitively I would rather do HTML escaping in the frontend. The backend should just expose the database credentials in their true form, it is the frontend's job to render them to the user.

extensions/database/tests/src/com/google/refine/extension/database/cmd/SavedConnectionCommandTest.java
```diff
@@ -154,8 +157,11 @@ public void testDoPost() throws IOException, ServletException {
         SUT.doPost(request, response);
         
         String result = sw.getBuffer().toString().trim();
- 
+        assertNotNull(result);
+        assertFalse(result.isEmpty(), "Valid response Message expected!");
+        
         ObjectNode json = ParsingUtilities.mapper.readValue(result, ObjectNode.class);
+        System.out.println("json:" + json);
```
wetneb: This looks like a debugging print that should be removed.

extensions/database/tests/src/com/google/refine/extension/database/cmd/SavedConnectionCommandTest.java
```diff
@@ -229,7 +235,7 @@ public void testDoPut() throws IOException, ServletException {
         Assert.assertEquals(savedConnections.size(), 1);
         
         ObjectNode sc = (ObjectNode)savedConnections.get(0);
-        System.out.println("sc" + sc);
+        //System.out.println("sc" + sc);
```
wetneb: I think we can remove this too.

extensions/database/module/scripts/index/database-source-ui.js
tcbuzor: This is just an extra check. The field will accept only Alphanumeric, period, underscore and hyphen only. I just added that line for extra check. As for html-escaping in the frontend, I'll have to check the javascript code again.

thadguidry:
@tcbuzor Thanks Tony for continuing to work on this!  Much appreciated!

extensions/database/module/scripts/index/database-source-ui.js
```diff
@@ -241,12 +241,15 @@ Refine.DatabaseSourceUI.prototype._editConnection = function(connectionInfo) {
...
               $.each(settings.savedConnections,function(index,savedConnection){
 
-                 items.push('<li class="pure-menu-item sc-list"><a href="#" class="pure-menu-link context-menu-one">'
-                    + '<span class="context-menu-text" >' + savedConnection.connectionName + '</span>'
-                    + '<span class="sc-context-more-vert pull-right"> </span></a></li>');
+                 var li = $('<li class="pure-menu-item sc-list"></li>').appendTo(menuList);
+                 var a = $('<a href="#" class="pure-menu-link context-menu-one"></a>').appendTo(li);
+                 $('<span class="context-menu-text"></span>').text(savedConnection.connectionName)
+                      .appendTo(a);
+                 $('<span class="sc-context-more-vert pull-right"> </span>').appendTo(a);
```
wetneb: This is a better way to construct DOM elements in jQuery: just use the `text()` method of a jQuery element, which already does the escaping for you.
In general, all user-supplied text should be inserted in this way in the UI. Manually concatenating HTML code in JavaScript is a bad sign.

Issue #2152, opened 2019-09-10
Cluster returning "groups" of 1 row/choice


emiraglia:
All of the cluster methods return clusters with one row/choice, which takes up processing time and makes using anything beyond ngram-fingerprint nearly impossible for larger sets.

![image](https://user-images.githubusercontent.com/20520821/64630920-f8962000-d3aa-11e9-8bec-0307449474a5.png)

**Desktop (please complete the following information):**
 - Windows 10
 - Browser Version: Chrome and Firefox

**OpenRefine (please complete the following information):**
 - Version 3.2

thadguidry:
First, any use of Clustering feature uses quite a bit of memory.
Try to increase the amount of memory that you allocate to OpenRefine.  Follow our guide here: 
https://github.com/OpenRefine/OpenRefine/wiki/FAQ:-Allocate-More-Memory

Have you tried different Clustering methods and functions such as kNN and Levenshtein?
Please review our wiki page that explains more in depth how to use our Clustering feature.
https://github.com/OpenRefine/OpenRefine/wiki/Clustering-In-Depth

My suggestion is perhaps to use Levenshtein and then start with an edit distance of 6 (the average length of most English words) and then try increasing or decreasing that edit distance value.

thadguidry:
You might also be expecting Clustering to find duplicates for you, which it sort of can do, but depending on your use case, you might find our Duplicates facet in the Column menu in what you are looking for.
If you want to treat your entire row (all columns) to look for Duplicates, then you can Concatenate all the columns into a new column and then run the Duplicates facet on that new concatenated column.

emiraglia:
Just for reference, when I use the exact same clustering method on the same dataset with version 3.1 I get this:
![image](https://user-images.githubusercontent.com/20520821/64637591-fdfa6700-d3b8-11e9-943f-5264eb90966d.png)

Which is much more in line with what I'd expect, and uses far less memory.

thadguidry:
I see no problem in 3.2 where you state that "All of the cluster methods return clusters with one row/choice"

There are clusters found with multiple rows using OpenRefine 3.2
3.1 version had several regressions and in hindsight we should not have released it.
So let's focus on 3.2 and expectations there.  Thanks.

So... here's what I get with fingerprint and Levenshtein with just 5 row test set that you can see on the grid additional to the Clustering dialog...

[...]

PR #2171, opened 2019-09-26
Skip only clusters with single rows

fsteeg:
See https://github.com/OpenRefine/OpenRefine/issues/2152


fsteeg:
Closing, see discussion starting at https://github.com/OpenRefine/OpenRefine/issues/2152#issuecomment-535406783

PR #2182, opened 2019-10-11
Add CSRF protection to POST endpoints

### LINKED ISSUE #2164: Add CSRF protection to all POST endpoints
We should add Cross Site Request Forgery protection to all our POST endpoints, to protect instances from being altered by other sites (which can be the first step in a XSS exploit).

- [x] add GET method to generate CSRF tokens
- [x] require and validate CSRF tokens in all POST endpoints. We probably want to do that in a generic way in the `Command` class.
- [x] update the frontend to request such tokens before POSTing

wetneb:
- [x] add GET method to generate CSRF tokens
- [x] change POST endpoints which do not modify any state into GET endpoints
- [x] require and validate CSRF tokens in all POST endpoints. We probably want to do that in a generic way in the `Command` class.
- [x] update the frontend to request such tokens before POSTing

This closes #2164.

### DIFF
--- a/extensions/database/module/scripts/database-extension.js
+++ b/extensions/database/module/scripts/database-extension.js
@@ -69,7 +69,7 @@ DatabaseExtension.handleConnectClicked = function(connectionName) {
                      databaseConfig.initialDatabase = savedConfig.databaseName;
                      databaseConfig.initialSchema = savedConfig.databaseSchema;
                     
-                        $.post(
+                        Refine.postCSRF(
                                 "command/database/connect",
                                  databaseConfig,
                                  
@@ -101,10 +101,10 @@ DatabaseExtension.handleConnectClicked = function(connectionName) {
                                     }
                                         
                                 },
-                                "json"
-                      ).fail(function( jqXhr, textStatus, errorThrown ){
-                       alert( textStatus + ':' + errorThrown );
-                   });
+                                "json",
+                                function( jqXhr, textStatus, errorThrown ){
+                                    alert( textStatus + ':' + errorThrown );
+                     });
                    
                  }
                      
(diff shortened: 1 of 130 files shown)


wetneb:
This will need a lot of testing and is likely to introduce bugs so we definitely want to call the next release a beta…

wetneb:
I have described these changes in the release notes:
https://github.com/OpenRefine/OpenRefine/wiki/Changes-for-3.3#csrf-protection-changes

This is going to break external clients such as https://github.com/opencultureconsulting/openrefine-client. This is unfortunately necessary to fix this important vulnerability. @felixlohmeier have you got any thoughts about these changes?

felixlohmeier:
Thanks for the heads up, @wetneb! Given that...
* the csrf token can be supplied either as GET or POST
* supplying one where it is not required does not fail

... I am going to take the easy way and obtain and supply a request token for each request. This can be implemented in the [method urlopen in refine-client-py](https://github.com/paulmakepeace/refine-client-py/blob/master/google/refine/refine.py#L57) and would be transparent to users.

I've [read that it's more secure](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#disclosure-of-token-in-url) to pass the CSRF token as POST parameter. That would require a change to the command create-project which only supports GET.

wetneb:
@felixlohmeier yes indeed, it could be preferable to only read the token from POST data. For commands like `create-project`, the POST body contains files and its reading is therefore handled differently. Interfering with that reading is a bit more complicated and likely to introduce bugs, so this is why I have gone down that route. Happy to change this if you can see a bullet-proof way to do this.

PR #2192, opened 2019-10-15
Remove useless condition

viniciusbds:
**It's known that m_current != 3 at this point**

This condition always produces the same result as the value of the involved variable that was narrowed before.


ostephens:
@viniciusbds thanks for this.

It looks to me like the `m_current == 3` condition is to ensure spellings like "MacGill" are correctly encoded. Can you point me at where this isn't possible due to previous conditions?

PR #2215, opened 2019-11-13
Update Wikidata-Toolkit to solve various issues

wetneb:
This adds support for client-side edit tagging, which will be enabled in the next release series.
It also improves the behaviour of OR when running on flaky network connections. For instance, when creating Wikidata items, if network goes down, the operation completes and the data grid is updated with the newly created items.


wetneb:
This was reviewed by a user and it seems to work for them: https://www.wikidata.org/wiki/User_talk:Bodhisattwa

Issue #533, opened 2012-10-15
Add support for Date only (ie w/o time) data type


tfmorris:
_Original author: danpaulsmith (February 14, 2012 15:03:47)_

When gathering user feedback on our extension for Refine the other day, a W3C member mentioned that when formatting a date (e.g. 07/05/2011) - information is created which isn't correct.

A formatted date gains the &quot;00:00:00Z&quot; (e.g. 2011-05-07T00:00:00Z) - stating that whatever this date is - happened at midnight - which is actually untrue.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=533_

thadguidry:
Java 8+ now has a class to store Dates without time or timezone (value-based, needs `equals` for comparisons).
https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/time/LocalDate.html

tfmorris:
The Java 8 date implementation is basically the same as the Joda package that I mentioned here: https://github.com/OpenRefine/OpenRefine/issues/533#issuecomment-9433294

thadguidry:
@tfmorris Yes, that I knew about.  (I still love Moment.js ways of working a bit more however, but our GREL is a bit better, but could be better still and closer to Moment's style of handling)

tfmorris:
I still believe that OpenRefine would benefit greatly from a Date datatype (and perhaps a Time datatype too).

Although what I said in 2012 about the internal format is true, what the user sees in Excel is the formatted version, so that's what's important to them. Even though they are all just floating point numbers internally, if the user sees it as a formatted date, it's a date to them.

Additionally, the original request didn't have anything to do with Excel at all.

ostephens:
>I still believe that OpenRefine would benefit greatly from a Date datatype (and perhaps a Time datatype too).

Completely agree. In library and archive data (which I work with, and train people to work with), there are a lot of dates without times and converting them to date times adds a spurious degree of specificity. The date information I encounter can be stated at year, year-month or year-month-day depending. To give a few examples from libraries:

* Books have a publication year. Representing these as a date in OR currently means choosing a specific date and time in that year
* Journal/serial issues can be dated by year-month (e.g. the December 2023 issue of a particular publication)
* Journal issues or specific articles might have a date of publication specified as year-month-day, but not a specific time

Currently the workflow needed tends to be like:`

emcaulay:
[...]

PR #2241, opened 2019-12-12
Fix ODS import for dates

### LINKED ISSUE #2224: Error importing date from ODS file
**Describe the bug**

A date in a cell in a ODS file is imported as a "java.util.GregorianCalendar" and displayed as 

 java.util.GregorianCalendar[time=1546297200000,areFieldsSet=true,areAllFieldsSet=true,lenient=true,zone=sun.util.calendar.ZoneInfo[id="Europe/Paris",offset=3600000,dstSavings=3600000,useDaylight=true,transitions=184,lastRule=java.util.SimpleTimeZone[id=Europe/Paris,offset=3600000,dstSavings=3600000,useDaylight=true,startYear=0,startMode=2,startMonth=2,startDay=-1,startDayOfWeek=1,startTime=3600000,startTimeMode=2,endMode=2,endMonth=9,endDay=-1,endDayOfWeek=1,endTime=3600000,endTimeMode=2]],firstDayOfWeek=2,minimalDaysInFirstWeek=4,ERA=1,YEAR=2019,MONTH=0,WEEK_OF_YEAR=1,WEEK_OF_MONTH=1,DAY_OF_MONTH=1,DAY_OF_YEAR=1,DAY_OF_WEEK=3,DAY_OF_WEEK_IN_MONTH=1,AM_PM=0,HOUR=0,HOUR_OF_DAY=0,MINUTE=0,SECOND=0,MILLISECOND=0,ZONE_OFFSET=3600000,DST_OFFSET=0] |  

**To Reproduce**

Create a ODS file with Libreoffice

Put a date in a cell (like 1/1/2000)

Import into openrefine

**Current Results**

**Expected behavior**

**Screenshots**

 - OS: Win10
 - Browser Version: Chrome
 - JRE or JDK Version: 1.8.0.231

 - Version :  OpenRefine 3.3 Beta

**Datasets**
If you are allowed and are OK with making your data public, it would be awesome if you can include or attach the data causing the issue or a URL pointing to where the data is.
If you are concerned about keeping your data private, ping us on our [mailing list](https://groups.google.com/forum/#!forum/openrefine)

**Additional context**

### LINKED ISSUE #2225: impossible to import a ODS file saved with Excel
**Describe the bug**

In Excel, save a XSLX file as ODS format, or open with Excel a ODS file created with Libreoffice and save it.

Try to import in Openrefine: the import is blocked in the 1st step

image

No Java errors in the console log

Maybe Excel is not fully compliant with the ODS format ?

**To Reproduce**
 

**Current Results**

**Expected behavior**

**Screenshots**

 - OS: Win10
 - Browser Version: Chrome
 - JRE or JDK Version: JRE 1.8.0_231

 - Version : OpenRefine 3.3 Beta

**Datasets**
If you are allowed and are OK with making your data public, it would be awesome if you can include or attach the data causing the issue or a URL pointing to where the data is.
If you are concerned about keeping your data private, ping us on our [mailing list](https://groups.google.com/forum/#!forum/openrefine)

**Additional context**

wetneb:
Closes #2224.

The importer now treats dates as strings - as a user I think I would prefer that rather than converting them to timestamps at midnight… I can still do that conversion later on with `toDate()` if necessary. @msaby, do you think this is the right approach?


msaby:
What is important for me is that ods and excel have the same behavior.
I also think that importing data from ods or Excel should use the more specific data type, so in that case the OR date fotmat, but maybe I am wrong...

thadguidry:
@msaby Your wrong.  :-) But let me explain why.  Dates have many patterns, sometimes the patterns are missing elements or it might be unclear if the elements are even in the right position.

Which is the day, month, and year in these date patterns?
10-01-10
02/02/01
01.01.02
We really have no idea until a fuller inspection of all rows and comparisons against other columns if they lend any hints.
This is especially true for datasets that were generated before Y2K.
The 1st date I gave is actually Oct. 10, 1910 but you might never know that just looking at the pattern.  It could have been Jan. 10 1910 or even Jan 10 2010 or 3010!!.
It is only from analysis of the dataset that a human might make sense of the data much more clearly than any algorithm could deduce.

Hence, we typically let humans use our power tool to coerce the String patterns into something meaningful and perhaps even more rigid by finally applying a DataType once they know for sure.

msaby:
Maybe I was not clear
Take this Excel file

image


of course "1/1/1888" is a string and should be imported as such
but 10/01/2020 is a date (= 43840 in number), and it has 1 meaning : january the 10th of 2020.

In an other flavor of Excel, maybe this day will be displayed as 01/10/2020, but the number behind will always be 43840.

So when we import a date encoded as  43840 in OR, I think it should be by defaut a date object meaning "january the 10th of 2020.", with an option for importing instead the string value. 

I took an Excel example, but I suppose it is the same for Libreoffice...

For hours, it's more complicated, as "10:02" is in fact the same value as "1/1/1900 at 10:02"

wetneb:
> I also think that importing data from ods or Excel should use the more specific data type, so in that case the OR date fotmat, but maybe I am wrong...

As far as I know we cannot store pure dates in a cell - we need to add hours, minutes and seconds to store them as OffsetDateTime. We can add these at midnight UTC, but I feel like that is distorting the data a bit… (but I really don't care much). @ostephens any opinion on that?

wetneb:
Related: #533 proposes to add support for storing dates (without time) in cells.

msaby:
What is most important for me : ODS import and Excel import should behave in the same way.

And for a "date" or "time" format value in the sheet (which are encoded in the same wayt, as a number),  I think the best would be to give the user the choice of 

* using textual representation used in the worksheet (ex: "2 october 2010" or "02/10/2010" or "10/02/2010" or "02/10/2010 00:00:00"), mapped as a string in OR

* or the numerical value (43840), mapped as date-time value in OR

thadguidry:
I don't think giving a choice of numerical value representing a date from an Epoch as shown in your bullet 2 would be useful.  Most folks are actually taking older data where dates are already stored using an Epoch (your numerical value) and want to convert them into a better date format like ISO 8601 or others.  Some good reading: https://stackoverflow.com/questions/31708622/how-to-represent-dates-before-epoch-as-a-unix-timestamp

In the Geologic and Chronostratigraphy domains, scientists do work with Epochs, but use simple decimal numbers or a String that represents "some date millions of years ago", such as 489.5 MYA  (489.5 million years ago).  The official system is called [GTS (Geologic Time Scale)](https://en.wikipedia.org/wiki/Geologic_time_scale).

msaby:
I am not sure to understand well.


Maybe some people have other uses cases, but what I mean is all the dates in the 1st line in this example are essentially the same date object in Excel (or Libreoffice), and I expect Openrefine to import them as the same OR date object  2000-01-03T00:00:00Z 

image

Currently it is **wrong with XLSX** file, because OR import those dates as a java.util.Date object (but denoting the same date)

image

And it is **wrong with ODS** because it produces a verbose error for each line

wetneb:
I am also in favour of ensuring the same behaviour for XLSX and ODS.
However I don't think importing to  2000-01-03T00:00:00Z is ideal, since it distorts the data by adding 00:00:00Z at the end, so I would either:
* import these as strings (so, as far as this PR is concerned, this means only the ODS importer remains to be fixed)
* or add proper support for date objects in OpenRefine (bigger task)

msaby:
ok but this need to be well documented, because people using MS Power Query or other tools won't expect that.

For the record in MS Power Query all my examples are analyzed as "01/03/2000 00:00:00", and it is up to the user to change the display : 

image

image

wetneb:
OK, if it is a normal behaviour in another tool we can definitely convert these dates to timestamps.

msaby:
@wetneb I think you are using tools like Pentaho or Talend. Could you tell us how they import excel or ods files ?

wetneb:
@msaby you probably did not mean to ping me? I don't use any of these tools :)

thadguidry:
@msaby Pentaho has all of its Trans Steps (transformation steps) [HERE](https://github.com/pentaho/pentaho-kettle/tree/master/engine/src/main/java/org/pentaho/di/trans/steps).  Where the specific transformation step for Excel Input is [HERE](https://github.com/pentaho/pentaho-kettle/tree/master/engine/src/main/java/org/pentaho/di/trans/steps/excelinput).

msaby:
Thanks 

@wetneb I thought you were using these tools but it is @thadguidry 

If I understand those lines, for Pentaho, if the Excel cell contains a date (with or without hours), it is converted to a date corresponding to the "numerical" underlying value, and the "display value" is not used in the process :

https://github.com/pentaho/pentaho-kettle/blob/cea3dcb52c1696fa66c90cde6d2721fc2d28ed38/engine/src/main/java/org/pentaho/di/trans/steps/excelinput/ExcelInput.java#L138

https://github.com/pentaho/pentaho-kettle/blob/cea3dcb52c1696fa66c90cde6d2721fc2d28ed38/engine/src/main/java/org/pentaho/di/trans/steps/excelinput/poi/PoiCell.java#L86

thadguidry:
@msaby Pentaho works this way:  If you don't specify a Format ([on the Fields Tab](https://help.pentaho.com/Documentation/8.2/Products/Data_Integration/Transformation_Step_Reference/Microsoft_Excel_Input)) then [the default handling kicks in](https://github.com/pentaho/pentaho-kettle/blob/cea3dcb52c1696fa66c90cde6d2721fc2d28ed38/engine/src/main/java/org/pentaho/di/trans/steps/excelinput/ExcelInput.java#L213) to convert the Date type that POI found and converts it to a String representation in the Pentaho row.

1. [Download PDI (Pentaho Data Integration)](https://sourceforge.net/projects/pentaho/files/Pentaho%208.3/client-tools/pdi-ce-8.3.0.0-371.zip/download)
2. Unzip and start data-integration/Spoon.bat  or Spoon.sh depending on your OS.
3. Click New Transformation
4. Expand Input 
5. Drag and Drop the "Microsoft Excel input" step into the empty Transformation window.
6. Double-click the step in the Transformation window to open its config.
7. Choose from dropdown the file type you are working with.  (xls, xlsx, ods, etc)
8. Click Browse and choose Filename to import
9. Then click the Add button (file is added into the grid to be loaded by the step. Pentaho supports multiple files being read)  (in my screenshot, the ODS file column 1 is a 
10. Go to Fields tab, and chose to get fields from header row (or just create the Schema there yourself if you know it, you can optionally change or specify the Format of the Field (column) - notice I did not specify a Format in the Fields tab)
11. Pentaho defaults to a String representation if you don't specify a Format (notice the Metadata structure of the r Row in the screenshot).

Annotation 2019-12-26 102834

My ODS file has the content within its XML structure as the following...
```xml
<table:table-row table:style-name="ro1">
	<table:table-cell office:value-type="date" office:date-value="2014-12-26" calcext:value-type="date">
		<text:p>12/26/14 12:00 AM</text:p>
	</table:table-cell>
	<table:table-cell office:value-type="string" calcext:value-type="string">
		<text:p>12-26-2019</text:p>
	</table:table-cell>
</table:table-row>
```

wetneb:
Closing this as there is no consensus around whether dates should be imported as strings or as a datetime object.

tfmorris:
I think the correct answer is to introduce pure dates to OpenRefine using Joda/Java 8 `LocalDate`s. Java's (previous) weak date handling has been improved in the past 10 years, so it makes sense to take advantage of the new support.

Issue #949, opened 2015-02-21
open refine can not export excel tables larger than 65536


daviddou82:
I am trying to export a 100k table to a excel file.  all the other export options seem to work except excel.

HTTP ERROR 500

Problem accessing /command/core/export-rows/concat-xlsx.xls. Reason:

```
Invalid row number (65536) outside allowable range (0..65535)
```

Caused by:

java.lang.IllegalArgumentException: Invalid row number (65536) outside allowable range (0..65535)
    at org.apache.poi.hssf.usermodel.HSSFRow.setRowNum(HSSFRow.java:232)
    at org.apache.poi.hssf.usermodel.HSSFRow.<init>(HSSFRow.java:86)
    at org.apache.poi.hssf.usermodel.HSSFRow.<init>(HSSFRow.java:70)
    at org.apache.poi.hssf.usermodel.HSSFSheet.createRow(HSSFSheet.java:205)
    at org.apache.poi.hssf.usermodel.HSSFSheet.createRow(HSSFSheet.java:71)
    at com.google.refine.exporters.XlsExporter$1.addRow(XlsExporter.java:96)
    at com.google.refine.exporters.CustomizableTabularExporterUtilities$1.visit(CustomizableTabularExporterUtilities.java:157)
    at com.google.refine.browsing.util.ConjunctiveFilteredRows.visitRow(ConjunctiveFilteredRows.java:76)
    at com.google.refine.browsing.util.ConjunctiveFilteredRows.accept(ConjunctiveFilteredRows.java:65)
    at com.google.refine.exporters.CustomizableTabularExporterUtilities.exportRows(CustomizableTabularExporterUtilities.java:171)
    at com.google.refine.exporters.XlsExporter.export(XlsExporter.java:138)
    at com.google.refine.commands.project.ExportRowsCommand.doPost(ExportRowsCommand.java:107)
    at com.google.refine.RefineServlet.service(RefineServlet.java:177)
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:820)
    at org.mortbay.jetty.servlet.ServletHolder.handle(ServletHolder.java:511)
    at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1166)
    at org.mortbay.servlet.UserAgentFilter.doFilter(UserAgentFilter.java:81)
    at org.mortbay.servlet.GzipFilter.doFilter(GzipFilter.java:132)
[...]

PR #2258, opened 2019-12-20
Update Apache POI to 4.1.1

wetneb:
This could potentially help with #949?


wetneb:
Confirmed by @thadguidry that it solves #949.

Issue #2194, opened 2019-10-15
Wikidata login: form not submitted when pressing Enter key


wetneb:
When pressing the Enter key in the Wikidata login form from the Wikidata extension, one would expect the form to be submitted, which currently does not happen.

Jataki:
Hi! Can I take this one?

wetneb:
@Jataki please!

Jataki:
Having an issue with my computer with no solution date. I'm sorry, but I'll have to leave this issue for others to have the opportunity to contribute. Thanks anyway and I'm so sorry!

Bhanu-mbvg:
Hey, can I take up this issue?

wetneb:
@Bhanu-mbvg sure, it would be great!

wetneb:
@Bhanu-mbvg let me know if you need any help :)

Bhanu-mbvg:
@wetneb How do I recreate Wikidata login form from the Wikidata extension

wetneb:
* Start OpenRefine
* Open any project (or create a new one, by opening any table from your computer for instance)
* Click on the Wikidata button in the top right corner
* Click "Manage Wikidata account"
This will create the login dialog where the form can be seen.

The HTML for this dialog is stored in extensions/wikidata/module/scripts/dialogs/manage-account-dialog.html.
The Javascript code for this dialog is extensions/wikidata/module/scripts/dialogs/manage-account-dialog.js.

Bhanu-mbvg:
#2259 Pull Request has been made. 
I resolved the issue using event listener for enter key inside the form
@wetneb can you see the PR

Bhanu-mbvg:
Hey @wetneb , I am still getting the error of form not getting submit on Enter keypress, can you check once?

wetneb:
When trying from the master branch, pressing Enter submits the form both with Firefox and Chromium, for me. Which browser are you using?

PR #2259, opened 2019-12-20
Wikidata login: form submittion by pressing Enter key

Bhanu-mbvg:
#2194 Issue Resolved


wetneb:
Thanks a lot for this! I have noticed a few issues, here they are.

You duplicated existing code: the code for the Enter key and the code for the click on the Login button are the same. It would be better to define a function and reuse it in both places.

Pressing Enter in the login form now triggers a login _twice_, because there are two `keydown` events for a given key press. You can check that with your developer tools: you will see that two requests to `/command/wikidata/login` are generated for each key press.

Also, your event listener stays registered even when the dialog has been closed, which leads to the following problem:
* Open the dialog and try to login with invalid credentials.
* Close the dialog.
* Open any other dialog, or add a text facet for instance, input a few things in a text input and press Enter;
* You will see that requests to `/command/wikidata/login` are still fired.

So this could lead to the user logging in when they do something unrelated!

I think this issue could be solved by adding a function on the `<form>` element with the `onsubmit` attribute.

Bhanu-mbvg:
Sure, thank you for mentioning the issues. Should I make a new pull request after making the required changes and close this pull request or make changes in this one only?

wetneb:
@Bhanu-mbvg as you prefer!

Bhanu-mbvg:
@wetneb I have worked and tresolved above 3 issues by : 

- Duplicated code converted into a `Login()` function
- Login form now triggers a login _once_ by having a variable `isOpen` variable
- keydown listener only active when this `isOpen` is set to _true_

wetneb:
Thank you!

I still think this would be cleaner via `onsubmit`, so with jQuery's `submit()` function on the form element:
https://api.jquery.com/submit/

Have you tried it?

Bhanu-mbvg:
Yes, @wetneb I tried that and on using the `onsubmit` the localhost page kept reloading and not show anything so chose to go with the pure Javascript approach.

wetneb:
Thanks! I have changed it slightly to use `jQuery.submit`: that handles form submissions by click and Enter key in the same go. I had to make sure the submit button was in the form for that. Do you see any problem with this solution?

Bhanu-mbvg:
The solution works fine just that we need to add a submit button inside form also one thing using jQuery increases the timespan for the login trigger.

wetneb:
Ok! I think this is a bit safer than registering an event listener that stays registered even when the form is not there anymore.

Issue #2260, opened 2019-12-23
Support CORS for reconciliation endpoints


wetneb:
We currently use JSONP to call reconciliation services from the browser. This is a problem since it is inherently insecure (the server can inject arbitrary javascript code in OpenRefine). The modern solution to call web services from a browser across domains is to use CORS.

The reconciliation CG is changing the standards to require CORS instead of JSONP (https://github.com/reconciliation-api/specs/issues/19) so we should support CORS too. Ideally we would support both during a transition period. When trying to add a service, check first if it can be reached via CORS, then with JSONP. We should probably remember in the service manifest whether the service could be accessed via CORS or not, so that we avoid trying CORS and falling back to JSONP for all subsequent queries.

PR #2261, opened 2019-12-24
Support CORS reconciliation services

wetneb:
This adds CORS support to the reconciliation UI as described in #2260. JSONP is still supported so users should not notice anything.

In a distant future we might consider removing JSONP support: we should encourage services to migrate to CORS.


thadguidry:
Thank you so much for adding this enhancement!
Can we self test this somehow, even for undefined?

wetneb:
For the "undefined" case, you can test this easily: any reconciliation service installed before this branch will fall into this case.
For the other cases, you can:
* install any service that supports CORS and check that suggest / flyout calls are made via CORS
* install any service that only supports JSONP and check that it still works
See https://reconciliation-api.github.io/testbench/ to get examples of services which support or do not support CORS.

PR #2275, opened 2020-01-07
Translate Sort link to the language of the UI

### LINKED ISSUE #2233: Translate the "Sort" link
**Describe the bug**

When a sort is made, the link "Sort" is displayed. It is not translated when the language of the UI is not english.

I suppose it could be done by editing this line 
https://github.com/OpenRefine/OpenRefine/blob/76162cfed150b32e1496bc07082302c35f56c6d7/main/webapp/modules/core/scripts/views/data-table/data-table-view.js#L152

**To Reproduce**

**Current Results**

**Expected behavior**

**Screenshots**

 - OS: Win10
 - Browser Version: Chrome
 - JRE or JDK Version:  JRE 1.8.0_231

 - Version :  OpenRefine 3.3 Beta

jamessspanggg:
Fixes #2233 

```
After a sort is made, the Sort link that is displayed does not
reflect to the corresponding language of the UI.

This PR adds the translation of the corresponding language
of the UI to the Sort link.
```

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
+++ b/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
@@ -149,7 +149,7 @@ DataTableView.prototype._renderSortingControls = function(sortingControls) {
 
   $('<a href="javascript:{}"></a>')
   .addClass("action")
-  .text($.i18n('core-views/first') + " ")
+  .text($.i18n('core-views/sort') + " ")
   .append($('<img>').attr("src", "../images/down-arrow.png"))
   .appendTo(sortingControls)
   .click(function() {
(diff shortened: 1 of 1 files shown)


main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -149,7 +149,7 @@ DataTableView.prototype._renderSortingControls = function(sortingControls) {
 
   $('<a href="javascript:{}"></a>')
   .addClass("action")
-  .text("Sort ")
+  .text($.i18n('core-views/first') + " ")
```
wetneb: This is a good start! However it's not enough for this string to be translatable - it also needs to say the right thing :)
At the moment this will show "First" instead of "Sort", which would be quite confusing. So the identifier for the translatable string needs to be different.

jamessspanggg: ahhh yes, that is really an obvious mistake 🤦‍♂ Have made the necessary change.

Issue #2277, opened 2020-01-07
Something is already running on http://127.0.0.1:3333/


Negens:
Hello ,

Distributor ID:	LinuxMint
Description:	Linux Mint 18.3 Sylvia
Release:	18.3
Codename:	sylvia

I've set no_proxy="localhost,127.0.0.1,127.0.0.1/8"
I also tried to set this no_proxy on refine.ini
I've tried to change the port
I've stopped the nginx server who was returning a 403 error page by default.
No services are running on the ports I've tried.

Whatever I do, I get this error.
I'm actually investigating the code to find the reason of this error.

thadguidry:
You might try asking in the [LinuxMint forums](https://forums.linuxmint.com/) or their [Chat room](https://linuxmint.com/contactus.php).

However to get OpenRefine to startup on another port, you can follow our [FAQ ](https://github.com/OpenRefine/OpenRefine/wiki/FAQ)which details how to change the Port that OpenRefine uses.
Once you change the port in the refine.ini then start OpenRefine with `./refine` or explicitly the ip and port you wish to use `./refine -i 127.0.0.1 -p 3334`

Negens:
Hello I also tried this option, it's the same result as modyfing the refine.ini file.

I launched some echo commands in the refine script
`$ ? = 22
URL = http://127.0.0.1:3333/
CURL -s URL | GREP =
Error: Something is already running on http://127.0.0.1:3333/ but doesn't seem to be OpenRefine. Maybe a proxy issue?
`

If the curl result is 22, the NOT_RUNNING viariable is supposed to be set, and it shouldn't go through the next steps of the check_running() function.

I checked the NOT_RUNNING "if" statement.
It doesn't enter the if statement and the NOT_RUNNING variable isn't populated.
`NOT RUNNING = `


To be sure, I tried this :
`        if [ "$?" = "7" ] || [ "$?" = "22" ] ; then
            echo "NOT RUNNING AT ALL"
            #NOT_RUNNING="1"
        fi
`
And I finally found something else that I can't explain.

if we do that :
`        curl -s -S -f $URL > /dev/null 2>&1

        echo "CURL RETURN value = "$?
        echo "CURL RETURN value = "$?
`
The result is 
`CURL RETURN value = 22
CURL RETURN value = 0
`

The $? value is changed by something or is reset to zero. I can't figure why for the moment.

Negens:
I found the reason of the bug !
[...]

PR #2278, opened 2020-01-08
Update the check_running() function in refine

Negens:
In strict shell environments, the "$?" is reset when evaluated.
Here, the "boolean if" statements compares it to "7", then reset it and compares it to "22".
If the "$?" is intially "22", it is not evaluated and the NOT_RUNNING variable is ignored while it shouldn't. this leads to the error
"Something is already running on http://127.0.0.1:3333/"

To avoid this, the solution is to first affect the value of "$?" to a non-system variable and then evaluate this new variable.

### DIFF
--- a/refine
+++ b/refine
@@ -157,7 +157,8 @@ check_running() {
 
     if [ "$CURL" ] ; then
         curl -s -S -f $URL > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


wetneb:
Closes #2277.

PR #2289, opened 2020-01-16
Inform user about download of Maven starting

thadguidry:


thadguidry:
@wetneb Thanks for fix... could we also tell the user WHERE it's going to be installed the PATH or MAVEN_HOME ?

wetneb:
I would not add the path in the message, since this local copy of Maven is not really meant to be used outside OpenRefine itself.

PR #2296, opened 2020-01-20
[#2280] Add fill down and blank down operations for all columns

### LINKED ISSUE #2280: Add "Fill down / Blank down" operation for all columns at once
There aren't many cell edits that make sense to apply to a whole dataset at once, but filling down is one of them (and I assume its sibling "Blank Down" too).

This would be a common operation needed after splitting multi-valued cells: depending on what the user wants to do with the exported data, they might need filling the resulting empty cells in _all_ other columns with the data from each record's first row.

The "All" dropdown could contain extra options "Edit rows > Fill down" and "Edit rows > Blank down", just like all other columns have.

I don't know of a different way to do that apart from applying it to each column one by one?

**Additional context**
Currently using OpenRefine 3.1 RC1.

jamessspanggg:
Resolves #2280 

```
There are use cases where users would want to apply
fill down/blank down operations for all columns. Users
would have to toggle the fill down/blank down operation
for each column in order to achieve that.

Let's add a fill down/blank down operation under the 
'All' > 'Edit Columns' drop down that applies the operation 
for all columns at once which will save the users' effort.
```

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
+++ b/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
@@ -875,11 +875,11 @@ DataTableView.prototype._createSortingMenu = function(elmt) {
 };
 
 var doAllFillDown = function() {
-  doFillDown(0);
+  doFillDown(theProject.columnModel.columns.length - 1);
 };
 
 var doFillDown = function(colIndex) {
-  if (colIndex < theProject.columnModel.columns.length) {
+  if (colIndex >= 0) {
     Refine.postCoreProcess(
         "fill-down",
         {
@@ -889,7 +889,7 @@ var doFillDown = function(colIndex) {
         {modelsChanged: true},
         {
           onDone: function() {
-            doFillDown(++colIndex);
+            doFillDown(--colIndex);
           }
         }
     );
--- a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
+++ b/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
@@ -765,34 +765,12 @@ DataTableView.prototype._createMenuForAllColumns = function(elmt) {
         {
           label: $.i18n('core-views/fill-down'),
           id: "core/fill-down",
-          click: function() {
-            for (var i = 0; i < theProject.columnModel.columns.length; i++) {
-              Refine.postCoreProcess(
-                  "fill-down",
-                  {
-                    columnName: theProject.columnModel.columns[i].name
-                  },
-                  null,
-                  { modelsChanged: true }
-              );
-            }
-          }
+          click: doAllFillDown
         },
         {
           label: $.i18n('core-views/blank-down'),
           id: "core/blank-down",
-          click: function() {
-            for (var i = 0; i < theProject.columnModel.columns.length; i++) {
-              Refine.postCoreProcess(
-                  "blank-down",
-                  {
-                    columnName: theProject.columnModel.columns[i].name
-                  },
-                  null,
-                  { modelsChanged: true }
-              );
-            }
-          }
+          click: doAllBlankDown
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


thadguidry:
@jamessspanggg Hi James, I'm getting some JS console errors when I test both functions on a simple datagrid:
```
a,b,c
1,2,3
1,,3
1,,3
```

```
XML Parsing Error: not well-formed
Location: http://127.0.0.1:3333/command/core/fill-down?columnName=b&project=2292583468875
Line Number 1, Column 1:
```

```
XML Parsing Error: not well-formed
Location: http://127.0.0.1:3333/command/core/blank-down?columnName=a&project=2292583468875
Line Number 1, Column 1:
```

jamessspanggg:
> @jamessspanggg Hi James, I'm getting some JS console errors when I test both functions on a simple datagrid:

Ok. Thanks for investigating. Will look up on it.

thadguidry:
@jamessspanggg Tested and works great on both functions! (The undo/redo history handles it as 3 separate operations instead of 1, but I think that's fine for now)

wetneb:
Yes the XML parsing error is just because the commands should add a `Content-Type: application/json` header but they do not always do it.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -863,6 +874,32 @@ DataTableView.prototype._createSortingMenu = function(elmt) {
...
+    Refine.postCoreProcess(
+        "fill-down",
+        {
+          columnName: theProject.columnModel.columns[i].name
+        },
+        null,
+        { modelsChanged: true }
+    );
+  }
+};
```
wetneb: I think it is fine to use the existing operations to implement this (so, resulting in multiple steps in the history).
However, doing a for loop like this will fire up HTTP queries to the backend in no particular order: the first column could well end up being filled down after the first one! That is a problem since the final state of the table can depend on the order of execution of these operations, if we are in records mode (simply because filling down the first column will change records, and filling down on another column in records mode respects the records boundaries).

So, the code should be adapted to use callbacks instead: the second column should only be filled down once the first one has. The `postCoreProcess` method accepts callbacks which can be used for that. Unfortunately that is going to lead to less readable code but I don't see how to avoid that to ensure correctness.

jamessspanggg: Thanks for the advise! I have tried to chain the operations with callbacks. 

However, I am not sure if my way is the simplest/cleanest way to do it. May I get your review on this?

wetneb: I would say this is a rather elegant way to do it!

wetneb:
Thanks! It looks good to me. I am still pondering whether to merge that before the final 3.3 release since this is a new feature and not a bug fix, but it looks like a really safe one, so I think it is doable.

wetneb:
After playing a bit with this, I wonder: when filling down all columns, do we really want to do this from left to right?

The problem with this approach is that we will first fill down the first column, which is used to define records.

Consider the following CSV dataset:

```
id,foo,bar
1,CN,45
,,56
,,87
2,DE,12
,,23
,,34
```

or in readable form:

```
id  foo   bar
1   CN    45
          56
          87
2   DE    12
          23
          34
```

If we use the new `All -> Edit cells -> Fill down` operation **in records mode**, we obtain:
```
id  foo   bar
1   CN    45
1         56
1         87
2   DE    12
2         23
2         34
```
whereas I would expect to obtain
```
id  foo   bar
1   CN    45
1   CN    56
1   CN    87
2   DE    12
2   DE    23
2   DE    34
```

So I think it would be worth doing the Fill down operations in reverse order, so that we obtain the expected result.
The Blank down operation should remain in the current order, such that it roughly corresponds to the inverse transform.

PR #2297, opened 2020-01-20
[#2220] Add facet for blank values/records per column

### LINKED ISSUE #2220: Add facet by blank per column in All menu

it can be useful to find all the columns with no value. We can use a facet with this formula filter(row.columnNames,cn,isBlank(cells[cn].value)), then sort by count to find the columns filled with blank values (for which the count will be the same as the total row of the dataset).

This facet could be prebuild on the same basis as https://github.com/OpenRefine/OpenRefine/issues/1977 

**Additional context**

jamessspanggg:
Resolves #2220 

```
It may be useful to have the break down of all columns 
with no values.

Let's add a facet for blank values/records per column.
```


thadguidry:
Does this look right for Blank records per column?

image

jamessspanggg:
> Does this look right for Blank records per column?

Hmm. The results for blank records per column should be `a: 0, b: 2: c: 3` instead right? 

In this case, the current implementation for non blank records per column is also incorrect. I've tried replicating your same test and it shows `a: 6, b:4, c:4`.

Will try to investigate on it.

wetneb:
It's not clear at all what "blank records per column" should even mean! Do you have an intuitive explanation of what this should be and in which scenarios it would be useful? The original issue makes a good case for the "blank values per column" facet but I can't work out the use case for records… And I suspect it might also be obscure for users as such :)

jamessspanggg:
> Do you have an intuitive explanation of what this should be and in which scenarios it would be useful? 

I think on what "blank records per column" should be is basically is the number of records where all of it's rows are null/empty. For the example above, it should be `a: 0, b: 2, c: 2`

Perhaps some users may find it useful to get the number of records where all of it's rows are empty? I'm unsure of the use case as well.

wetneb:
Great! That seems like a sensible thing to have. I am thinking of merging this before 3.3, since this is a well isolated feature which is very unlikely to break anything else.

thadguidry:
The reason I gave the example is that this is common to see in scientific data sets.  And often confusing to others on what is really being represented, and furthermore, what might the last 2 rows actually be considered as.

Context is important and without it, the values would be skewed with OpenRefine's single method of Record alignment.  In my example given, they are typically called "Record Rows"... sampling data on a DateTime, where that DateTime column sometimes initially is thrown out (Column A initially was a DateTime stamp) and you are left with just the Record Rows of Data.

In my example, OpenRefine thinks that I have 6 records, but it's actually 1 Record  (where column "a" is indicated by value 1".  The last 2 rows are outliers.
Folks have different expectations on Record alignment, and I have mentioned this many, many times before.
OpenRefine conforms to only 1 method of Record alignment on a single Column, but in reality, folks might have a few ways they align Records even within that single Column, as my explanation demonstrates.

In my context example, I have 0 "blank records per column".  But I do have 5 "incomplete" Record Rows, i.e. "some columns with no value".  Knowing that I have 5 "incomplete" Record Rows is what is important to me.  However, Ettore in #2220 wants to know "all columns with no value", where my case would be "some columns with no value".  Both provide usefulness.

PR #2308, opened 2020-02-10
Add Apache Maven download path to .gitignore

thadguidry:


wetneb:
Makes sense! But you could do that on the master branch, no? I will rebase the spark branch on master so that it starts off the 3.3 release, so that will be brought to the spark branch too.

PR #2324, opened 2020-02-22
enh: Text Filter waits until user finishes updating

### LINKED ISSUE #2112: Text filter should wait until user finishes before updating.
When entering a string or regex expression in the text filter, I'm often prevented from finishing the expression before OpenRefine attempts to update the results. This is frustrating, especially when trying to formulate complex regex.

My preferred solution would be a `Submit` button (ideally with a keyboard shortcut) so that OpenRefine doesn't keep trying to update while I'm typing.

The only alternative I've found is to write regex expressions in a text file and copy/paste into the text filter.

kushthedude:

Fixes #2112

### DIFF
--- a/main/webapp/modules/core/scripts/facets/text-search-facet.js
+++ b/main/webapp/modules/core/scripts/facets/text-search-facet.js
@@ -127,13 +127,13 @@ TextSearchFacet.prototype._initializeUI = function() {
   this._elmts.caseSensitiveCheckbox.bind("change", function() {
     self._config.caseSensitive = this.checked;
     if (self._query !== null && self._query.length > 0) {
-      self._scheduleUpdate();
+      setTimeout(function(){ self._scheduleUpdate(); }, 250);
     }
   });
   this._elmts.regexCheckbox.bind("change", function() {
     self._config.mode = this.checked ? "regex" : "text";
     if (self._query !== null && self._query.length > 0) {
-      self._scheduleUpdate();
+      setTimeout(function(){ self._scheduleUpdate(); }, 1500);
     }
   });
 
@@ -203,7 +203,7 @@ TextSearchFacet.prototype._scheduleUpdate = function() {
     this._timerID = window.setTimeout(function() {
       self._timerID = null;
       self._updateRest();
-    }, 500);
+    }, 2000);
   }
 };
 
--- a/main/webapp/modules/core/scripts/facets/text-search-facet.js
+++ b/main/webapp/modules/core/scripts/facets/text-search-facet.js
@@ -203,7 +203,7 @@ TextSearchFacet.prototype._scheduleUpdate = function() {
     this._timerID = window.setTimeout(function() {
       self._timerID = null;
       self._updateRest();
-    }, 2000);
+    }, 500);
   }
 };
 
--- a/main/webapp/modules/core/scripts/facets/text-search-facet.js
+++ b/main/webapp/modules/core/scripts/facets/text-search-facet.js
@@ -127,13 +127,13 @@ TextSearchFacet.prototype._initializeUI = function() {
   this._elmts.caseSensitiveCheckbox.bind("change", function() {
     self._config.caseSensitive = this.checked;
     if (self._query !== null && self._query.length > 0) {
-      setTimeout(function(){ self._scheduleUpdate(); }, 250);
+      self._scheduleUpdate();
     }
   });
   this._elmts.regexCheckbox.bind("change", function() {
     self._config.mode = this.checked ? "regex" : "text";
     if (self._query !== null && self._query.length > 0) {
-      setTimeout(function(){ self._scheduleUpdate(); }, 1500);
+      self._scheduleUpdate();
     }
   });
 
@@ -202,7 +202,12 @@ TextSearchFacet.prototype._scheduleUpdate = function() {
     var self = this;
     this._timerID = window.setTimeout(function() {
       self._timerID = null;
-      self._updateRest();
+      if(self._config.mode === 'regex') {
+        setTimeout(function(){ self._updateRest(); }, 1000);
+      }
+      else if (self._config.caseSensitive == true) {
+        setTimeout(function(){ self._updateRest(); }, 250);
+      }
     }, 500);
   }
 };
--- a/main/webapp/modules/core/scripts/facets/text-search-facet.js
+++ b/main/webapp/modules/core/scripts/facets/text-search-facet.js
@@ -202,13 +202,8 @@ TextSearchFacet.prototype._scheduleUpdate = function() {
     var self = this;
     this._timerID = window.setTimeout(function() {
       self._timerID = null;
-      if(self._config.mode === 'regex') {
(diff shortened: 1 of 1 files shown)


kushthedude:
@wetneb I was thinking of having PR Template for the Open Refine. Please tell me if it is good to go, I will go ahead and make PR

main/webapp/modules/core/scripts/facets/text-search-facet.js
```diff
@@ -147,7 +147,8 @@ TextSearchFacet.prototype._initializeUI = function() {
       return;
     }
     self._query = this.value;
-    self._scheduleUpdate();
+    setTimeout(function(){ self._scheduleUpdate(); }, 1500);
```
thadguidry: 1200 and 1300 feel a bit more natural to me, whereas 1500+ feels awkward. But then, the original issue was that of Regex entry into the Text Filter box, and when I test with a bit of Regex, indeed, 1500 is a good balance. Let's go with that!

kushthedude: So it's good to go then ?

thadguidry: We'll let a few other folks who use this often and see what they think.  It seems to work fine also and delay with the Regex error propagation also for instance if doing a `(a)(b)`

@ostephens @ettorerizza How does this feel for you when typing Regex?  too long? too short?

kushthedude:
Cool I will make the changes in the _scheduleUpdate function.

kushthedude:
@wetneb Done, as there was timeout of 500ms already and I added timeout of 1500ms additionally which was working quite fine. Hence I increased the original timeout to 2000. Please let me know if additional changes are required.

kushthedude:
@ostephens @thadguidry I have made separated timeout for both types of filters. I used the Reg-Ex and Case-Sensitive filters both, it works quite great.

wetneb:
Yes absolutely. At the moment the proposed change has two issues, in terms of actual behaviour:
* adding delays when boxes are checked;
* keeping the same 500ms delay on keystrokes, regardless of whether the regex checkbox is enabled.
The right place where to add this differentiated delay is indeed in the `_scheduleUpdate` function.

ostephens:
@kushthedude are you happy to make changes in line with the comment from @wetneb and me?

kushthedude:
I was working on recon issue meanwhile that's why is the delay. I will
complete it by tujhe.

kushthedude:
@wetneb @ostephens @thadguidry I tried to implement a timeout in the schedule_update function itself. Please have a look timeout is same for every keystroke and for case-sensitive/regex differential values have been set for an additional timeout.

kushthedude:
> At the moment there is a 500 ms timeout followed by either a 250 ms timeout or a 1000 ms timeout. Wouldn't it be cleaner to just have one timeout, whose length would vary depending on the checkbox's state? So, vary the existing 500 ms timeout instead of introducing another layer of timeouts, I would say.

@wetneb I could have used the ternary operator if we had two different types of time-outs but we have 3 timeouts right now:
- 500 ms for key-stroke.
- 750 ms for Case-Sensitive
- 1500 ms for RegEx.
Can you suggest a better way to utilise the same?

thadguidry:
@kushthedude Make a single timeout object which takes a delay of ms.   Based on if the user checks some boxes... set the appropriate delay in ms.  We don't want 3 timeout objects.  In other words, the input box has 1 single timeout... what the timeout delay is...will be based on criteria the user clicked on various boxes.  Does that make more sense?

kushthedude:
> @kushthedude Make a single timeout object which takes a delay of ms. Based on if the user checks some boxes... set the appropriate delay in ms. We don't want 3 timeout objects. In other words, the input box has 1 single timeout... what the timeout delay is...will be based on criteria the user clicked on various boxes. Does that make more sense?

Damn, I didn't get the idea to create an object. Thanks, @thadguidry for it!

thadguidry:
Cool.  And if you read above, it doesn't have to be an object.  As @ostephens states, it could just be you change the way the higher level function works.  `TextSearchFacet.prototype._scheduleUpdate` function.

And I agree with @ostephens the only criteria for now would be the Regex checkbox that increases or decreases the delay.  The case sensitive box shouldn't affect anything.

OK, now get to work and figure it all out! :-)

kushthedude:
> And I agree with @ostephens the only criteria for now would be the Regex checkbox that increases or decreases the delay. The case sensitive box shouldn't affect anything.

@thadguidry Please have a look now. Since the only Regex should vary the timeout, I didn't need to make a seperate variable for a timeout.

kushthedude:
> That's much better (and simpler!). If we agree on the specific values of the timeouts, then it works for me.

I think 1500ms for regex was right for everyone as can be seen from the above discussion. @ostephens @thadguidry What are your views on the same?

thadguidry:
1000 - 1500 is fine.  My preference seemed to be around 1200.  But up to you.

kushthedude:
> 1000 - 1500 is fine. My preference seemed to be around 1200. But up to you.

I think 1500 is fine as regex are quite complex than typing letter. But still, if 1200 is what all want I have no object for it.

thadguidry:
This will never end.
I think what the world wants here, especially with Regex, is that it should be a button to submit (or Enter key), and not auto-submitting in all honesty, since it could be 1 million rows to process or more eventually.  Some users will want 1.5 secs, some will want 5 or 10 secs.  A submit or Go button that is mapped to the ENTER key upon focus inside the input box would be the best of all, just like the original poster (OP) asked for in #2112.  The flip side is having a Cancel button for the processing (but this would be harder to implement and not as user friendly.)  A button keybinded will give the power back to the user, instead of us making a decision for him on how many seconds.  [Our core philosophy.](https://github.com/OpenRefine/OpenRefine/pull/2366/files)

kushthedude:
Got it @thadguidry . I will be making changes as per you asked !

thadguidry:
@kushthedude I'd show it as a the FILTER button (with ENTER keybinding) since it is a Text Filter operation ... something like this pretty picture :-)  I think this button should show always with the Text Filter box and would lend clarity to what the darn thing actually does!

image

wetneb:
@thadguidry hang on, that's a completely different change from what @kushthedude has been working on so far… It's not great to have him do many changes in the way timeouts are defined, just to change our mind at the last minute to switch to a totally different solution, no?

I would be inclined to merge this as such - replacing the timeout by a button is something we can discuss in another issue, as it brings other complications (how do you display a text filter that has been changed and not validated, for instance?)

thadguidry:
@wetneb absolutely.

kushthedude:
@thadguidry @wetneb What I suggest for the now is let's move ahead with this PR and see how it works for the users. But if the users still need a seperate text filter button after this, we can cover it in a cover-up PR.
Tell me how it sounds to you? :)

wetneb:
@thadguidry thanks, yes I am not opposed to a button at all.

thadguidry:
@kushthedude yes

ostephens:
Thanks @kushthedude - great!

PR #2325, opened 2020-02-23
fix: cannot export a file if blank tab opened is not closed

### LINKED ISSUE #1664: Cannot export a file if the "blank" tab opened after a first export is not closed

Version 3 beta

Windows 10

Chrome

When you export a file, a new blank tab ("about: blank") opens with the download window. As long as this tab remains open, you can't download anything again. This is particularly frustrating when using "Custom tabular export": not only it does not download anything, but you have to restart the settings configuration.

screencast

#### Expected Results:

Automatically close the new blank tab or allow multiple download tabs.

kushthedude:

Fixed #1664

### DIFF
--- a/main/webapp/modules/core/scripts/dialogs/custom-tabular-exporter-dialog.js
+++ b/main/webapp/modules/core/scripts/dialogs/custom-tabular-exporter-dialog.js
@@ -320,7 +320,7 @@ CustomTabularExporterDialog.prototype._postExport = function(preview) {
   
   document.body.appendChild(form);
 
-  window.open("about:blank", "refine-export");
+  window.open(" ", "refine-export");
   form.submit();
 
   document.body.removeChild(form);
--- a/main/webapp/modules/core/scripts/dialogs/sql-exporter-dialog.js
+++ b/main/webapp/modules/core/scripts/dialogs/sql-exporter-dialog.js
@@ -368,7 +368,7 @@ function SqlExporterDialog(options) {
    // alert("form::" + form);
     document.body.appendChild(form);
   
-    window.open("about:blank", "refine-export");
+    window.open(" ", "refine-export");
     form.submit();
   
     document.body.removeChild(form);
--- a/main/webapp/modules/core/scripts/dialogs/templating-exporter-dialog.js
+++ b/main/webapp/modules/core/scripts/dialogs/templating-exporter-dialog.js
@@ -169,7 +169,7 @@ TemplatingExporterDialog.prototype._export = function() {
 
     document.body.appendChild(form);
 
-    window.open("about:blank", "refine-export");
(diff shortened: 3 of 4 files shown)


kushthedude:
> Tested with Chromium and the behaviour seems identical to the original: exporting with the custom tabular exporter opens a tab, exporting a second time generates a download in the same tab. Ideally, no tab should be open at all, I would say.

 I didn't test with custom tabular export, my bad. I will fix it soon.

kushthedude:
@wetneb I looked deeper into it. There will definitely a need for new-tab since we generate a download URL. I have fixed the issue of tabular exporter tabs and other tabs, but if we will generate the download URL in the same tab then-current location will be lost once the download is successful. Also, most of the download files in other UI opens in a new tab for file-location and once the download is initiated or cancelled, the corresponding tab closes. I have implemented similar interface in here.
Please let me know if there are any more issues in the same.

thadguidry:
To me, it would be so much better if the Download button or Menu items themselves were anchor's that used HTML5 "download" attribute which has been supported in all browsers for the last 10 years [REF] .

But then we have Jquery here to deal with... or perhaps we don't need Jquery here and can have a more elegant solution, like those offered in a few answers and comments in here:
https://stackoverflow.com/questions/11620698/how-to-trigger-a-file-download-when-clicking-an-html-button-or-javascript

`window.location()` instead of `window.open()` perhaps?

REF:
https://developers.google.com/web/updates/2011/08/Downloading-resources-in-HTML5-a-download
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

kushthedude:
@thadguidry We can definitely use the onClick Event in Anchor Tag but this will make the UX poor. As once the user will click on the button, current window location will change and to get back to the previous tab he will have to navigate again or another option is we use `target=_blank` tag which will ultimately redirect to another window to choose the file location and same error will persist. 
I request you to please test the current changes once since the new tab closes once the user navigates away from pop.

thadguidry:
Still not working in Firefox.  The browser tab remains there after
1.  clicking Export -> Comma Separated Value 
2. clicking Save As on dialog that pops up.
3. about:blank in the URL for new tab that opened remains there.

maybe this is some issue I'm having on my dev env.  I've even reset hard on your branch and see the file changes, and yet ...

kushthedude:
I did check it on nightly🤔 I will check it once on firefox release version
and get back to you.
Thanks for the information.

kushthedude:
@wetneb Thanks for the review! Is it good to be merged then?

wetneb:
Yes, let's just wait a bit to make sure Thad agrees with my assessment above.

thadguidry:
I agreed...gave the 👍 already on your comment :)

PR #2327, opened 2020-02-23
Use release convention for Maven compiler config

thadguidry:
Can we move to using latest OpenJDK ?  (I'm using OpenJDK 13)
Regardless, we should use the newer "release" convention instead of source/target which wrecks havoc for me on Windows dev environment and switching Maven toolchains for other software.  Maybe Linux users are not prone to env issues as much (a reason Windows created WSL in the first place).  Anyways, some further context on using the "release" flag: https://stackoverflow.com/questions/43102787/what-is-the-release-flag-in-the-java-9-compiler

It's also possible to combine both properties and profile, and just use activation as shown: https://stackoverflow.com/questions/55404641/configure-maven-compiler-plugin-so-it-works-both-with-jdk-8-and-jdk-12-in-sprin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,8 +61,7 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.1</version>
         <configuration>
-          <source>1.8</source>
-          <target>1.8</target>
+          <release>13</release>
           <encoding>UTF-8</encoding>
           <showDeprecation>false</showDeprecation>
         </configuration>


mrcsparker:
I was thinking of submitting a pull with some Java 8 features that simplify the code. I wasn’t sure where to start - should I submit requests per extension?

wetneb:
@mrcsparker please do!

@thadguidry I am happy with using the `release` convention but would not upgrade Java to anything higher than 8 - this helps guarantee that we keep supporting Java 8 and newer.

wetneb:
Closing since we still want to support Java 8.

PR #2337, opened 2020-02-27
fix: Browser redirection for 0.0.0.0

### LINKED ISSUE #2336: browser should go to http://localhost not http://0.0.0.0
I like to open up my OpenRefine for colleagues (security? phew!),
- set this in `refine.ini`: `REFINE_HOST=0.0.0.0`
- or run `refine.bat -i 0.0.0.0`

This works fine, but after the server starts, it sends my browser to http://0.0.0.0:3333 which doesn't work. It should send me to http://localhost:3333/

kushthedude:
Fixes #2336 

cc: @thadguidry @wetneb

### DIFF
--- a/server/src/com/google/refine/Refine.java
+++ b/server/src/com/google/refine/Refine.java
@@ -463,8 +463,11 @@ class RefineClient extends JFrame implements ActionListener {
     
     public void init(String host, int port) throws Exception {
 
-        uri = new URI("http://" + host + ":" + port + "/");
-
+    	String cleanedHost = host;
+    	if("0.0.0.0".equals(host)) {
+            cleanedHost = "localhost";
+        }
+        uri = new URI("http://" + cleanedHost + ":" + port + "/");
         openBrowser();
     }
     


server/src/com/google/refine/Refine.java
```diff
@@ -464,7 +464,7 @@ static private String fixWindowsUnicodePath(String path) {
     public void init(String host, int port) throws Exception {
 
     	if(host.contains("0.0.0.0")) {
```
wetneb: Why `contains` and not string equality here? For instance `140.0.0.0` contains `0.0.0.0` but I don't think we want to rewrite that to `localhost`.

Also, it would perhaps be marginally cleaner to only set a variable storing the cleaned host in the `if` block, and then build the URI based on that cleaned host?

kushthedude: @wetneb Made the changes you requested!

kushthedude:
> Much simpler, thanks! Two small things about code style:
> 
> It is generally preferred to write `"0.0.0.0".equals(host)` rather than `host.equals("0.0.0.0")`. This is because it avoids null pointer exceptions (you know for sure that the string constant is never going to be null - but the variable `host` potentially could).
> 
> Also, ideally I would avoid reassigning method arguments (as a general good practice, not that it causes a bug in this specific situation).

Duly noted! Making the changes .

kushthedude:
> Also, it would perhaps be marginally cleaner to only set a variable storing the cleaned host in the if block, and then build the URI based on that cleaned host?

@wetneb Used a variable to set the cleaned host as you said and building the URI with the same.

kushthedude:
@wetneb Can this be merged ?

PR #2339, opened 2020-02-27
Closes dialogs when escape key is pressed

### LINKED ISSUE #1018: Close modal dialogs on ESC key
We need a possibility to close any modal window by pressing "Escape". Like we have this behaviour now on cells' "Edit", and "Rename this column" windows.

lisa761:
Closes #1018 
Closes all dialog boxes (created through `DialogSystem`) when escape key is pressed

### DIFF
--- a/main/webapp/modules/core/scripts/dialogs/clustering-dialog.js
+++ b/main/webapp/modules/core/scripts/dialogs/clustering-dialog.js
@@ -120,6 +120,22 @@ ClusteringDialog.prototype._createDialog = function() {
     this._elmts.applyCloseButton.click(function() { self._onApplyClose(); });
     this._elmts.closeButton.click(function() { self._dismiss(); });
 
+    var clickOutside = function(event) {
+        $target = $(event.target);
+        var dist = $target.closest(dialog).length;
+        if(dist == 0 && $(dialog).is(':visible')) {
+            self._dismiss();
+        }
+    }
+    $(document).click(clickOutside);
+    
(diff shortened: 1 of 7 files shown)


thadguidry:
@lisa761 Thanks!  Curious, was an outside click asked for in the issue?  Do you think this might cause a user some grief if they spent 10 mins configuring a dialog to their settings (looking up details on other screens, etc. SQL schema from Oracle, Templates from an outside spec) ... and then an errant click would lose the dialog info they patiently typed in?

lisa761:
@thadguidry You're right, I didn't think that far ahead (I just found it in the edit cell that was mentioned in the issue and thought it was a nice addition), it would definitely cause problems to those who did it by mistake. Should I remove it then?

thadguidry:
@lisa761 Try it out as a user on a few of the dialogs and then see how it feels for a user...then make your decision :-)

wetneb:
Congratulations for this PR!! \o/

One thing I am wondering is whether this could be implemented in a single place (the `DialogSystem`) which would apply to all the dialogs used in OpenRefine. That would avoid duplicating this logic in many places :) Have you considered moving some of this into the `DialogSystem.createDialog`, `DialogSystem.showDialog` and `DialogSystem.dismissUntil` methods?

lisa761:
@thadguidry @wetneb Thanks, for helping out :) I made the changes

lisa761:
@wetneb yes, I can do that by globally declaring `escapeKey`

PR #2340, opened 2020-02-28
Issue 2320 qs refs

### LINKED ISSUE #2320: Only first reference is exported for QuickStatements
**Describe the bug**
If your are adding multiple references to a statement, the QuickStatements export only contains the first reference. 

**To Reproduce**
1. Add two or more references to new statement in the Wikidata schema
2. Click on 'Export to QuickStatements'
3. View your QS-Export 
4. See that only the first Reference is exported.

**Expected behavior**
To handle multiple references in QuickStatements, the whole statement has to be duplicated just with the different reference. QuickStatements prevents to add a duplicate statement but is able to add a second or further reference to the pointing statement.

 - OS: Ubuntu 18.04
 - Browser Version: Firefox
 - JRE or JDK Version:. openjdk 11.0.6 

 - Version 3.3

**Datasets**
https://gitlab.com/LibrErli1/museen_burgenland/-/blob/master/museen_bgld.openrefine.tar.gz

afkbrb:
Fixes #2320 .

Before this PR, only the first reference is exported for QuickStatements.

For example, if we add a statement with two refs:

issue1-schema

The preview will work fine:

issue1-preview

However the QuickStatements output will only contain the first ref:

issue1-qs

To fix this, we can just duplicate the statement just with different references. QuickStatements prevents to add a duplicate statement but is able to add a second or further reference to the pointing statement.

After the fix, the QuickStatements export now works fine:

issue1-fixed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,9 @@ test-out/
 open-refine.log
 .vscode
 
+.idea
+*.iml
+
 main/target/
 main/webapp/WEB-INF/lib/
 server/target/
--- a/extensions/wikidata/src/org/openrefine/wikidata/exporters/QuickStatementsExporter.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/exporters/QuickStatementsExporter.java
@@ -151,21 +151,38 @@ public class QuickStatementsExporter implements WriterExporter {
         Value val = claim.getValue();
         ValueVisitor<String> vv = new QSValuePrinter();
         String targetValue = val.accept(vv);
-        if (targetValue != null) {
+        if (targetValue != null) { // issue #2320
             if (!add) {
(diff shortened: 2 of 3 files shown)


wetneb:
Fantastic, it looks great to me!
When making changes to the Java backend we generally try to add corresponding tests to demonstrate the change. Do you think you could try adding a test to demonstrate this? We already have a few tests about QuickStatements export, so you should be able to follow the existing pattern. Let me know if you need any help with that :)

afkbrb:
Thanks for the review, and sure that I should add corresponding tests. I can work on it now.

afkbrb:
I've added a test for the corresponding changes, new the coverage for QuickStatementsExporter is 100%!

PR #2347, opened 2020-02-29
remove auto launching Expression editor on 'change' click

### LINKED ISSUE #745: In the facet box display the facet expression used
When you combine multiple true/false facet on the same column it can be hard to remember what the action behind each one (isblank or a custom facet), see image attached

https mail-attachment googleusercontent com attachment u 0 ui 2 ik 39313f2a28 view att th 13f520c0a725495e attid 0 1 disp inl

Currently the work around is to click change, close the transform windows without doing anything, to display the GREL expression in the facet box.

This could be expended directly or display in a different manner

thadguidry:
Fixes #745

### DIFF
--- a/main/webapp/modules/core/scripts/facets/list-facet.js
+++ b/main/webapp/modules/core/scripts/facets/list-facet.js
@@ -174,11 +174,7 @@ ListFacet.prototype._initializeUI = function() {
 
   this._elmts.titleSpan.text(this._config.name);
   this._elmts.changeButton.attr("title",$.i18n('core-facets/current-exp')+": " + this._config.expression).click(function() {
-    self._elmts.expressionDiv.slideToggle(100, function() {
-      if (self._elmts.expressionDiv.css("display") != "none") {
-        self._editExpression();
-      }
-    });
+    self._elmts.expressionDiv.slideToggle(100);
   });
   this._elmts.expressionDiv.text(this._config.expression).hide().click(function() { self._editExpression(); });
   this._elmts.removeButton.click(function() { self._remove(); });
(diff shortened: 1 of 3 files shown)


kushthedude:
@thadguidry I had a doubt, I see there are branches which were updated more than a year ago. Do we really need them?

thadguidry:
@kushthedude The branch cleaning certainly needs to be done.  But I'd like to do that on a core team call here in the next few weeks with @wetneb @ostephens and myself who know the history of those branches, we can certainly add any other contributors.  We also certainly need to setup a brand new "OpenRefine Dev" call and begin hosting that 1 time a month (I used to do that on Sunday mornings and need to begin doing that again)

kushthedude:
> @kushthedude The branch cleaning certainly needs to be done. But I'd like to do that on a core team call here in the next few weeks with @wetneb @ostephens and myself who know the history of those branches, we can certainly add any other contributors. We also certainly need to setup a brand new "OpenRefine Dev" call and begin hosting that 1 time a month (I used to do that on Sunday mornings and need to begin doing that again)

Great, please let me know if I can be of any help !

@thadguidry One more thing, can we set-up netlify deploys for PRs?

thadguidry:
Let’s move all this discussion to dev mailing list please (unrelated to this PR)

ostephens:
It feels like this modifies the action of the "Change" button - making it a "View" button.
I don't have a problem with there being a two step process, but I think we need to make it more obvious that:

* first click displays the facet expression
* second click to change/edit the facet expression

At the moment, clicking "Change" does not give  the option for the user to Change anything, and there is no indication that a change can be made by clicking on the displayed expression

thadguidry:
@ostephens , when I clicked on "Change" on Firefox, it slides open the small box, and just moving my mouse over it highlights in Yellow and quickly displays the hover text quickly displays "Click to edit this expression".  I think the Yellow coloring and hover text seem to be enough indication?

Does it work like that for you or different?

thadguidry:
The more I look at a few design decisions we made in the early days of OpenRefine... I'm not happy with our current modal-based (popup and lock the main interface) Expression editor launching.  In fact, I don't like the modal-based dialogs that restrict the main interface.  I'd rather move to less modal-based interactions where the Expression editor (and other parts of OpenRefine) be redesigned later to make use of a right-hand sidebar (topbar, bottombar) where options and previewing can happen while directly seeing and interacting on the existing grid itself, as well as Reconciling previewing, etc, etc.

Closing this as "not needed".

PR #2349, opened 2020-03-01
Issue 2103 description validator

### LINKED ISSUE #2103: New wikidata validator: description lengths and endings
In the Wikidata extension, we could add validators to raise warnings when the following happens:
- adding a description which is too long (250 characters)
- adding a description which ends by punctuation signs (these should not be removed automatically but flagged, as they might indicate that full sentences were used as descriptions)
- any other sanity checks?

afkbrb:
fixes #2103 

I added a new item description validator.

It can do the following checks:
1. is a description too long
2. does a description end by punctuation signs
3. does a description begin with a uppercase letter
4. does a description begin with article ("a", "an" or "the")
5. is the description identical with corresponding label

If we create a schema with the following preview:

issue2-preview

The checks above will work, and the warning messages will look like this:

issue2-issue


thadguidry:
@afkbrb You are rocking!  Keep up the great work!  We absolutely love all this help you are providing to OpenRefine!

wetneb:
Actually I can even think of legitimate descriptions in English starting with capital letters: `London-based finance start-up` is fine. So perhaps it could be at a lower severity level (info, perhaps?).

The help page does mention that capitalized descriptions can happen in English:
https://www.wikidata.org/wiki/Help:Description

afkbrb:
> There might be languages where a legitimate description can start with "a", "an" or "the". Or languages where the first letter can be capitalized (for instance it is fine in German). 

Yes, I do need to think twice about the current implementation.

How about only doing the `is a description too long` check and `is the description identical with corresponding label` check in `DescriptionScrutinizer`. And create a new class `EnglishDescriptionScrutinizer` which extends `DescriptionScrutinizer` to do the checks for English only.

This way, whenever we need to implement a new language-specific scrutinizer, we just need to extend the `DescriptionScrutinizer` and model after `EnglishDescriptionScrutinizer`.

> Actually I can even think of legitimate descriptions in English starting with capital letters: London-based finance start-up is fine. So perhaps it could be at a lower severity level (info, perhaps?).

Agree again. info level should be more proper for the uppercase check.

> What do you think about adding a sample problematic description to the warnings raised by the scrutinizers?

Actually I didn't notice the format scrutinizer! I just copied some code from `NewItemScrutinizer` for the warning. I'll make the warning message more specific today.

afkbrb:
Done! Should be better now.

wetneb:
Brilliant, thank you so much!

Final thought (hopefully): when descriptions are too long, they will not be able to be uploaded to Wikidata at all. So for this specific warning we can probably raise the severity to the highest level (which will prevent from starting the upload in OpenRefine directly).

afkbrb:
> So for this specific warning we can probably raise the severity to the highest level (which will prevent from starting the upload in OpenRefine directly).

Yes, and that's easy to fix.

issue2-issue-final

extensions/wikidata/module/langs/translation-en.json
```diff
@@ -156,5 +156,15 @@
...
+    "warnings-messages/item-description-too-long/title": "Description is too long",
+    "warnings-messages/item-description-too-long/body": "Description ({lang}) such as <span class=\"wb-issue-preformat\">{description}</span> on {example_entity} is too long. Its length is {length}, which is more than {max_length}. Descriptions are not full sentences, but small bits of information. In most cases, the proper length is between two and twelve words. See the <a href=\"https://www.wikidata.org/wiki/Help:Description#Length\" target=\"_blank\">manual</a> for more information.",
+    "warnings-messages/item-description-identical-with-label/title": "Description is identical with label",
+    "warnings-messages/item-description-identical-with-label/body": "Both the description ({lang}) and the label ({label_lang}) on {example_entity} are <span class=\"wb-issue-preformat\">{description}</span>. Description are expected to be more specific than labels.",
```
thadguidry:
```suggestion
    "warnings-messages/item-description-identical-with-label/body": "Both the description ({lang}) and the label ({label_lang}) on {example_entity} are <span class=\"wb-issue-preformat\">{description}</span>. Description are expected to be more specific than labels. See the <a href=\"https://www.wikidata.org/wiki/Help:Description#Length\" target=\"_blank\">manual</a> for more information.",",
```

afkbrb:
@thadguidry 

There is no corresponding part for **Description is identical with label**  on the [manual](https://www.wikidata.org/wiki/Help:Description).  So linking to the manual is not necessary.

But I think it's reasonable to do this check, because it makes sense to require the description to be more specific than label.

thadguidry:
@afkbrb There is at top of the Wiki page, it's in the "**This page in a nutshell:**" section
>This page in a nutshell:

    A description describes and disambiguates a label using information that is free of controversy and bias
    Keep it short—descriptions need to only be long enough to allow people to easily grasp what an item's label refers to
    Follow style guidelines; punctuate and capitalize as if the description was appearing in the middle of a sentence
 as well as the first 2 sentences:
>The description on a Wikidata entry is a short phrase designed to disambiguate items with the same or similar labels. A description does not need to be unique; multiple items can have the same description, however no two items can have both the same label and the same description.

thadguidry:
@afkbrb What prevents an upload with 2 items having identical label and description? Something on Wikidata side, as well as our side? How does this quoted rule get enforced?  Curious, does the service do an indexing query and check on this rule for us?
>however no two items can have both the same label and the same description.

afkbrb:
> There is no corresponding part for Description is identical with label on the manual.

I mean there is no section (such as the "Capitalization" section and "Punctuation" section ) that explicitly express that description shouldn't be identical with label. But the references you mentioned above **do** indicate that a description should be more specific than the label. I'll add the manual for reference in the code then.

> however no two items can have both the same label and the same description.

We don't assure it on our side. It seems to be more difficult than other checks.

> Curious, does the service do an indexing query and check on this rule for us?

I am not sure about this. Probably it's just a convention.

wetneb:
Sorry @afkbrb I edited my earlier comment so you probably did not see this issue:

> This way, whenever we need to implement a new language-specific scrutinizer, we just need to extend the DescriptionScrutinizer and model after EnglishDescriptionScrutinizer.

I think it makes sense to use different classes, but the problem with using inheritance here is that checks done in the `DescriptionScrutinizer` will be made in all subclasses too. So if I have both an `EnglishDescriptionScrutinizer` and a `GermanDescriptionScrutinizer` then they will both run the maximum length check. So a single long description will generate two long description warnings.

afkbrb:
@wetneb You're right. That's a huge bug. I'll get it fixed.

PR #2350, opened 2020-03-01
Non printable characters

AghaSaad04:
I have added a feature of displaying non-printable characters. It is a fix for #1286.

## Steps:
- Import the messy data
- Create the project
- Toggling the non-printable character checkbox would show/hide the non-printable characters
#### Data grid without non-printable characters:
image

#### Data grid with non-printable characters:    
image


thadguidry:
@AghaSaad04 Can you change that first DEMO image to without using the Grammerly plugin? (Just disable it then screenshot again)
I'm pulling this in now and taking a look and testing.

AghaSaad04:
Done @thadguidry

thadguidry:
Ah, we also really needed this in the Project preview (prior to Create Project)
How difficult would that be to add the option there also?

thadguidry:
This is a good start, but it's not fulling mapped yet?  I typed in Control Chars using Windows Key + ALT + numpad 0001 - 0031 on various lines

SOURCE:
image

RESULT:
image

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -646,3 +657,56 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
...
+        var originalContent = $(columns[j]).text();
+        if (originalContent != "") {
+          var updatedContent = checkNonPrintable(originalContent);
+          $(columns[j]).html(updatedContent);
+        }
+      }
+    }
+  }
+  else {
+    $(".unprintableCharacaters").remove();
```
thadguidry: typo:  unprintableCharacters
```suggestion
    $(".unprintableCharacters").remove();
```

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -646,3 +657,56 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
...
+    else if ((content.charAt(character)).charCodeAt(0) == 32) {
+      unprintableChar = "<span class='unprintableCharacaters' style='background-color: orange'><b>NBSP</b></span>";
+    }
+    stringIncNonPrintable += unprintableChar + content.charAt(character);
+  }
+  return stringIncNonPrintable;
+}
+
+function parseToTextAreaFromDiv() {
+  $(".unprintableCharacaters").remove();
```
thadguidry: typo:
```suggestion
  $(".unprintableCharacters").remove();
```

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -646,3 +657,56 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
     MenuSystem.dismissAll();
   });
 };
+
+function checkNonPrintable(content){
+  var stringIncNonPrintable = "";
+  for (var character = 0; character < content.length; character++) {
+    var unprintableChar = "";
+    if ((content.charAt(character)).charCodeAt(0) == 13) {
+      unprintableChar = "<span class='unprintableCharacaters' style='background-color: orange'><b>CR</b></span>";
```
thadguidry:
```suggestion
      unprintableChar = "<span class='unprintableCharacters' style='background-color: orange'><b>CR</b></span>";
```

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -646,3 +657,56 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
...
+    }
+    else if ((content.charAt(character)).charCodeAt(0) == 10) {
+      unprintableChar = "<span class='unprintableCharacaters' style='background-color: orange'><b>LF</b></span>";
```
thadguidry:
```suggestion
      unprintableChar = "<span class='unprintableCharacters' style='background-color: orange'><b>LF</b></span>";
```

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -646,3 +657,56 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
...
+    } 
+    else if ((content.charAt(character)).charCodeAt(0) == 9) {
+      unprintableChar = "<span class='unprintableCharacaters' style='background-color: orange'><b>HT</b></span>";
```
thadguidry:
```suggestion
      unprintableChar = "<span class='unprintableCharacters' style='background-color: orange'><b>HT</b></span>";
```

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -646,3 +657,56 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
...
+    } 
+    else if ((content.charAt(character)).charCodeAt(0) == 16) {
+      unprintableChar = "<span class='unprintableCharacaters' style='background-color: orange'><b>LS</b></span>";
```
thadguidry:
```suggestion
      unprintableChar = "<span class='unprintableCharacters' style='background-color: orange'><b>LS</b></span>";
```

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -646,3 +657,56 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
...
+    } 
+    else if ((content.charAt(character)).charCodeAt(0) == 32) {
+      unprintableChar = "<span class='unprintableCharacaters' style='background-color: orange'><b>NBSP</b></span>";
```
thadguidry:
```suggestion
      unprintableChar = "<span class='unprintableCharacters' style='background-color: orange'><b>NBSP</b></span>";
```

main/webapp/modules/core/styles/views/data-table-view.less
```diff
@@ -356,3 +356,16 @@ ul.sorting-dialog-blank-error-positions > li {
   cursor: move;
   .rounded_corners(3px);
   }
+
+.interfaceForTextArea{
+  font: 400 13.3333px Arial;
```
thadguidry: Hmm, wondering if we should actually prefer first some Unicode fonts (Arial Unicode,etc.) in preference order, and then `Arial` ?  Do popular Linux distros have them?  What about Mac?

AghaSaad04:
@thadguidry Firstly thanks for reviewing, and yeah it is not fully mapped yet. That was a sort of prototype, I just wanted to make sure that the feature I was implementing was what you guys wanted. I will look into the Project preview part. Hopefully, that would not be tough. I will fix that typo. :)

thadguidry:
@AghaSaad04 Great!  Understood this is a prototype to see what was possible.  Now that we know a bit more, we can begin to make improvements.  To reduce repetition, I'd improve the code through the use of `.map`, arrow functions and perhaps `for..of`.  Experiment.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions

AghaSaad04:
Yeah, @thadguidry. Working on it to make it efficient and optimize.

thadguidry:
@AghaSaad04 I added my simple test file above the image in previous comments, if you want to use it or adapt it for test cases, up to you.

AghaSaad04:
Perfect Thanks @thadguidry. 
I was thinking of creating an array for control characters and then finding the character code in that array. 
A better-optimized version.
image

Imported the test file:
image

Result:
image

thadguidry:
@AghaSaad04 Would a dictionary make things easier?  `["SOH":"0001","STX":"0002"...]`

thadguidry:
Also, I'd replace char(9) `HT` with `TAB`, since this is a quite common convention worldwide, and char(11) is vertical tab and lesser known, so its fine to remain as `VT`.

thadguidry:
One last thing, that I think would be useful to users....
`SPACE`, where @ostephens and I discussed this on the issue and both thought it probably would be useful to show subtly.  Perhaps as Notepad++ and other text editing tools do by drawing white space with a single middle dot sometimes or other glyph.
`SPACE` is not in the Control Char range, but instead the start of the ASCII standard, it's printable, but hidden from our eyes often (2 different things, right :-) )

image

Sublime and other IDE's have the ability to toggle rendering white space in preferences.
I think we should have a 2nd toggle to do just that, but can be done in a separate issue.

thadguidry:
Lastly, the label "non-printable characters" is not accurate, since many Unicode chars also do not print with a glyph such as U+A0 and U+200B and U+FEFF , etc.
I think we need to be more direct and say "show control codes".

wetneb:
I would not add a "non printable characters" textbox at this place in the UI, I think it is way too visible. This is not something that people will need to toggle often - better keep that precious space for other things (such as extensions).

I think the option should be enabled by default and it should be possible to disable it in some menu, for instance in the preferences.

thadguidry:
@wetneb This was discussed before on the issue... Owen and Ettore also thought having a quick toggle was useful.  But I understand moving it outside of `div#tool-panel` probably would be best.  Perhaps here ?
image

wetneb:
> Owen and Ettore also thought having a quick toggle was useful.

Where? I can't find that in #1286.

thadguidry:
@wetneb maybe was another related issue or the mailing list.  We can ask them.

AghaSaad04:
> Ah, we also really needed this in the Project preview (prior to Create Project)
> How difficult would that be to add the option there also?
> The reason is because older data can use those Control Chars as separators (and often did!), and we'd like users to see them and could type in the Control Char(s) as a custom separator in the CSV/TSV/separator-based Importer.
> 
> image

@thadguidry I made some changes and got these results.  
image

But I am still trying to figure out when a user switches to another parsing format, it has to be unchecked and then checked again for the functionality to work. 

Once the data parsing format is changed, which function renders the data?

thadguidry:
@AghaSaad04 Don't know, but you should be able to watch the events, or network calls, to find that out?  No rush on this, you are doing great!

thadguidry:
I think prominent is important on this, it has affected many users.  But I guess what Antonin is saying is because of that, let's just make it default on and hide the toggle in our preferences.  I suspect we'll see later after release if users want to have that toggle exposed in the UI or not. (my hunch still says yes, but we'll see what the majority of users want and let them vote and speak for themselves, we trust them)

ostephens:
I'm in favour with this being the default view and controlled in preferences (as described by @wetneb)

If we want to have a place inside the project to control the view, the All->View menu would seem like the most appropriate place right now - but I prefer the idea of it being set in preferences as I think it's likely to be something people want to set across all projects

wetneb:
Yes, see for instance what @ostephens did to enable/disable displaying null values in cells in #1571. We could have a configuration menu for the grid's appearance where these options would be gathered, and that would be something that could be displayed relatively prominently, but not this checkbox on its own.

AghaSaad04:
Thanks, @wetneb @thadguidry  @ostephens for the suggestions. I will try to implement this.

AghaSaad04:
@thadguidry @wetneb, I have updated the PR. Kindly review :)

main/webapp/modules/core/scripts/index.js
```diff
@@ -244,3 +244,54 @@ $(function() {
 
   showVersion();
 });
+
+// Setting up control character as a default in preferences tab
+function set_preference_control_char() {
+  Refine.postCSRF(
+    "command/core/set-preference",
+    {
+      name: "Control characters",
```
thadguidry:
```suggestion
      name: "ui.show-control-characters",
```

main/webapp/modules/core/scripts/index.js
```diff
@@ -244,3 +244,54 @@ $(function() {
...
+
+// Getting value of control character
+function get_preference_control_char() {
+  var control_char = null;
+
+  $.ajax({
+    async: false,
+    type: "GET",
+    url: "command/core/get-preference?" + $.param({
+      name: "Control characters"
```
thadguidry:
```suggestion
      name: "ui.show-control-characters"
```

main/webapp/modules/core/scripts/preferences.js
```diff
@@ -107,12 +107,15 @@ function PreferenceUI(tr, key, value) {
 
   $('<button class="button">').text($.i18n('core-index/edit')).appendTo(td2).click(function() {
     var newValue = window.prompt($.i18n('core-index/change-value')+" " + key, value);
+    if (key == "Control characters"){
```
thadguidry:
```suggestion
    if (key == "ui.show-control-characters"){
```

main/webapp/modules/core/scripts/facets/list-facet.js
```diff
@@ -735,7 +735,40 @@ ListFacet.prototype._setChoiceCountLimit = function(choiceCount) {
...
+      stringIncNonPrintable += unprintableChar;
+    }
+    return stringIncNonPrintable;
+  }
+  return str;
+}
+
+// Converting tags into control characters
+function tags_to_control(str) {
+  if (localStorage.getItem('preference_control_char') == 'Enabled') {
```
thadguidry:
```suggestion
  if (localStorage.getItem('preference_control_char') == 'true') {
```

main/webapp/modules/core/scripts/index.js
```diff
@@ -244,3 +244,54 @@ $(function() {
...
+    "command/core/set-preference",
+    {
+      name: "Control characters",
+      value: JSON.stringify("Enabled")
+    },
+    function (o) {
+      if (o.code == "error") {
+        alert(o.message);
+      } else {
+        localStorage.setItem('preference_control_char', 'Enabled');
```
thadguidry:
```suggestion
        localStorage.setItem('preference_control_char', 'true');
```

main/webapp/modules/core/scripts/index/parser-interfaces/preview-table.js
```diff
@@ -104,7 +106,21 @@ Refine.PreviewTable.prototype._render = function() {
           .attr("target", "_blank")
           .appendTo(divContent);
         } else {
-          $('<span>').text(cell.v).appendTo(divContent);
+          if (localStorage.getItem('preference_control_char') == 'Enabled' ) {
```
thadguidry:
```suggestion
          if (localStorage.getItem('preference_control_char') == 'true' ) {
```

main/webapp/modules/core/scripts/facets/list-facet.js
```diff
@@ -735,7 +735,40 @@ ListFacet.prototype._setChoiceCountLimit = function(choiceCount) {
...
         "json"
-      );      
+      );
     }
   }
 };
+
+// Converting control characters into tags
+function control_to_tags(str) {
+  if (localStorage.getItem('preference_control_char') == 'Enabled') {
```
thadguidry:
```suggestion
  if (localStorage.getItem('preference_control_char') == 'true') {
```

main/webapp/modules/core/scripts/index.js
```diff
@@ -244,3 +244,54 @@ $(function() {
...
   showVersion();
 });
+
+// Setting up control character as a default in preferences tab
+function set_preference_control_char() {
+  Refine.postCSRF(
+    "command/core/set-preference",
+    {
+      name: "Control characters",
+      value: JSON.stringify("Enabled")
```
thadguidry:
```suggestion
      value: JSON.stringify("true")
```

main/webapp/modules/core/scripts/index.js
```diff
@@ -244,3 +244,54 @@ $(function() {
...
+  return control_char;
+}
+
+//  Checking whether preference of control character has been set or not
+if (get_preference_control_char() == null) {
+  if (localStorage.getItem('preference_control_char') == null) {
+    set_preference_control_char();
+  }
+  else {
+    localStorage.setItem('preference_control_char', 'Disabled');
```
thadguidry:
```suggestion
    localStorage.setItem('preference_control_char', 'false');
```

main/webapp/modules/core/scripts/index.js
```diff
@@ -244,3 +244,54 @@ $(function() {
 
   showVersion();
 });
+
+// Setting up control character as a default in preferences tab
```
thadguidry:
```suggestion
// Setting up control character as a default in global preferences tab
```

thadguidry:
Additionally, when testing the Edit cell.  I cannot backspace and remove (collapse) the control chars to clean them up manually.
This is something that many other editors do allow and would be awesome if we could implement it as well.

image

Additionally, we have broken functionality with Transforms in GREL Expression Editor.
Below, I am replacing the letter "b" with a regular space character.  Yet, the preview shows I am replacing with a NBSP (non-breaking space char), which is not correct.

image

AghaSaad04:
@thadguidry Thanks for reviewing, I will implement the suggested changes.

P.S for the edit option. It's weird as It's working on my side:

Before:
image

After pressing backspace:
image

And for the GREL bug, you have mentioned, I am sorry I made a mistake and assigned 'NBSP' to 32, will change that to 'space'

thadguidry:
> P.S for the edit option. It's weird as It's working on my side:
I'm testing on Firefox, and you?

On Chrome it looks like this and doesn't show much?  Are we dealing with a default font issue or something else?  Why the difference between Firefox and Chrome here?

image

AghaSaad04:
I am testing on Chrome and just checked on firefox. Yeah, that's weird backspace not working on firefox.

AghaSaad04:
So I just googled a bit about this problem, this is a bug in Firefox which has been reported in bug_zilla for more than 10 years. There are workarounds for this problem, will implement them.

AghaSaad04:
There are some libraries that are giving the same issues in firefox :(

thadguidry:
Would this be an easier way for now?  https://lonekorean.github.io/highlight-within-textarea/

AghaSaad04:
> Would this be an easier way for now? https://lonekorean.github.io/highlight-within-textarea/
The issue with this approach is that this plugin will not only highlight the control characters words but also those words having control characters in them:
Examples:
**TAB**LET
**ACK**NOWLEDGE

thadguidry:
yes, and couldn't you "adapt" that code to work with <tag> ? or ask Will how one would go about that?  perhaps opening an issue with his library to support that functionality?

main/webapp/modules/core/.fuse_hidden000008ab00000008
```diff
@@ -0,0 +1,87 @@
+<!doctype html>
```
wetneb: This looks like a file that was added by mistake.

AghaSaad04:
@wetneb, thanks for reviewing.
### For Firefox:
It's a bug in firefox you cannot add after the contenteditable="false" element if it is at the end of the text. Also, you cannot delete the contenteditable="false". 

### For chromium I will review again.
I am looking into the alternatives.

AghaSaad04:
So I just reviewed the Chromium issue, and got to know the localStorage function I am using is not supported by Chromium :/

wetneb:
I would not worry about caching the result of a command in local storage - you could run into issues of the cache being out sync with the preferences, and you can generally assume such a command is very quick, since the server is generally running locally.

If we want to minimize the number of queries (which could be very useful for hosted versions), we could refactor all calls fetching preferences such that all preferences are fetched just once and stored in a global scope (without caching them in local storage).

Anyway I think the problem you are tackling here is hard enough - don't make your life unnecessarily complicated with that sort of detail (that is a form of [premature optimization](https://wiki.c2.com/?PrematureOptimization)!)

main/webapp/modules/core/scripts/facets/list-facet.js
```diff
@@ -777,3 +785,162 @@ function tags_to_control(str) {
   }
   return str;
 }
+
```
thadguidry: Wow, that's a serious amount of custom code, just for FF support!  Maybe this is a code smell that we are doing something wrong in our approach? Don't know, not an expert.

main/webapp/modules/core/scripts/index/parser-interfaces/preview-table.js
```diff
@@ -37,6 +37,8 @@ Refine.PreviewTable = function(projectData, elmt) {
   this._render();
 };
 
+var controlCharacters = ["NUL", "SOH", "STX", "ETX", "EOT", "ENQ", "ACK", "BEL", "BS", "TAB", "LF", "VT", "FF", "CR", "SO", "SI", "DLE", "DC1", "DC2", "DC3", "DC4", "NAK", "SYN", "ETB", "CAN", "EM", "SUB", "ESC", "FS", "GS", "RS", "US", "SPACE"];
```
thadguidry: Let's replace `SPACE` with just `SP` to reduce verbosity

main/webapp/modules/core/scripts/index/parser-interfaces/preview-table.js
```diff
@@ -104,7 +106,22 @@ Refine.PreviewTable.prototype._render = function() {
...
         } else {
-          $('<span>').text(cell.v).appendTo(divContent);
+          if ($.cookie('preference_control_char') == 'true' ) {
+            var stringIncNonPrintable = "";
+            for (var character = 0; character < cell.v.length; character++) {
+              var unprintableChar = "";
+              var charCode = cell.v.charAt(character).charCodeAt(0);
+              if (charCode <= 32) {
+                var size = (controlCharacters[charCode].length + 2) * 2;
+                unprintableChar = "<input type='button' class='unprintableCharacters' disabled='disabled' width=" + size + "% value=" + controlCharacters[charCode] + ">";
```
thadguidry: To me it looks much better if change `unprintableCharacters` to have a `style="padding-left: 2px;padding-right: 2px;"` added.  Otherwise it seems there's a default padding of `8` on both sides and makes it too stretched?

main/webapp/modules/core/scripts/facets/list-facet.js
AghaSaad04: Not sure though. As there are some plugins I tried them as well those were not working on FF due to this backspace issue.

main/webapp/modules/core/styles/common.less
```diff
@@ -436,3 +436,10 @@ p.body-text {
   height: 100%;
   overflow: hidden;
   }
+
+.unprintableCharacters{
```
thadguidry: Also, for some reason, I think the margin-top is getting accumulated?  I preferred a setting of `margin-top: -3px;`

image

thadguidry: Double check all that in both FF and Chrome.

main/webapp/modules/core/scripts/index/parser-interfaces/preview-table.js
AghaSaad04: yeah, was thinking that as well. I will replace that now.

main/webapp/modules/core/scripts/index/parser-interfaces/preview-table.js
AghaSaad04: Sure. I will update the styling.

tfmorris:
This clearly represents a ton of work (and review), so I'm reluctant to close it, but there haven't been any updates in three months.

If @AghaSaad04 isn't available to push it over the finish line, is it close enough that someone else could pick it up? (I haven't looked at it myself, so asking @thadguidry and @wetneb who seem to have done most of the review).

thadguidry:
@tfmorris As a user, I've been pretty happy with it, and have a build with it merged into 3.3 myself that I have been using on occasion to have this long-standing needed feature in OpenRefine.  It's quite a useful thing actually and I have it defaulted on in preferences on that custom build.  I haven't had any problems other than a tiny small alignment issue between Chrome and Firefox, but its so minor and I bet @antoine2711 could find the issue on that sometime later after merge (or before if he can).

Anyways, that's my take on testing and using it.  Have no idea if it's performance bottle-necking for @lisa761 's ongoing work and didn't check all that, only against 3.3 tag and merge/build.  So I'd say it definitely needs more testing against master here.

wetneb:
Yes this is impressive work. On my side I am not sure we can pull this confidently, because of the complexity of the code to support editing.

Perhaps one way forward would be to highlight non-printable characters only in non-editable locations first, and leave the support in form inputs for a follow-up PR.

thadguidry:
@wetneb I agree with the way forward on this.  Now the question is who can make that happen?
@antoine2711 Do you think you might be able to volunteer an afternoon towards that, if @AghaSaad04 could not?

thadguidry:
I am now wondering...

For rendering in the Expression Editor:
I wonder if the best way forward would be to restart this and instead incorporate [ACE ](https://ace.c9.io) to handle the majority of the code editing experience within our Expression Editor (without having to maintain our own custom code for highlighting hidden characters, syntax highlighting, folding, indenting, etc.)

For rendering in the data grid:
I do not know if ACE (or a smaller part of it), could also be used to highlight hidden characters there?

(Syntax Highlighting in ACE is actually done with [a common standard](https://pcwalton.blogspot.com/2010/11/syntax-highlighting-specification.html) now in place that nearly all Javascript code editors use and developed by PCWalton who worked with Mozilla Labs and Rust)

wetneb:
I am going to close this as stale, given that there has not been any activity for many months.

I think what went wrong here is that @thadguidry did a lot of end-user reviewing upfront, but developers did not have time to review the PR code-wise. That pushed @AghaSaad04 to make a lot of changes, investing a lot of time in the PR to match the end user requirements, but building up a bit of a house of cards on the code side.

So:
* developers should aim to give feedback early on in a PR
* end-user reviewers (@thadguidry) should not push a contributor to do a lot of changes without the assurance of a developer that the PR is going in the right direction

If anyone wants to take that forward it is of course possible to open a new PR, possibly from the same branch.

thadguidry:
@wetneb ??? No pushing here.  It was a collaboration.  I do agree that within this @AghaSaad04 and I were both trying to see the limits of what could be done, that's all.  In reality it was an EPIC, as so many of our single issues are actually, but we don't spend the time to break them up that way.  That is something we should all try to improve on.

My opinion was that this was a terrific collaboration that met some technical hurdles for @AghaSaad04 with our current UI framework and the interactions of him coming to grips with some of it's pain points, along with learning new things along the way.  I definitely commend @AghaSaad04 on actually being successful in showing the non-printable chars in the Expression Editor...which was the primary pain point for users.  Since we give users other means to determine if there are non-printable (control chars) in the data grid like with the Unicode facet.

I think going forward for anyone on this PR and its issue, is that indeed they focus on 1 area at a time to display the non-printable chars, which @AghaSaad04 did get working quite well on that branch.

wetneb:
@thadguidry of course that was a collaboration: my point is that by bringing up a lot of small details in your reviews (which is very useful in general), you are encouraging them to invest more and more time into the PR. As a team, it is our responsibility to ensure that we only do that when we are confident that the PR can make it, otherwise it can be quite disappointing for contributors.

It's really the developers' fault here - we should have chimed in earlier on. But in these cases it would be good if you could slow down a tiny bit perhaps and remind us that we need to have a look at it?

Issue #898, opened 2014-09-04
Handle long column names better in Key/Value Columns dialog


baditaflorin:
![resize posibility](https://cloud.githubusercontent.com/assets/2330999/4147754/d5234208-341b-11e4-8c2c-2bdff2dab527.png)

obarabanov:
@thadguidry is this ticket going to be in review status for some more time, or it can be set as done ?

wetneb:
@obarabanov issues are closed when the corresponding PR is merged (automatically if they have been properly linked, by writing "closes #898" or "fixes #898" in the pull request or commit message).

obarabanov:
ah ok it's like in other collaboration platfrorms. But I haven't linked it properly, anyway.
My idea was just to give idea / to indicate this ticket is taken / being worked on.
Not sure who's going to merge the PR ?

wetneb:
Sure, we can assign you for that.

wetneb:
Closed by #2352.

PR #2352, opened 2020-03-02
Handle long column names in key/value columnize dialog

obarabanov:
changes for issue #898:
* made "Columnize by Key/Value columns" dialog resizable
* set overflow styling, which solves the issue in Chromium-based browsers (Chrome, Edge, Opera) but doesn't work in FireFox and WebKit-based browsers (like Epiphany @ Linux, not tested in Safari but most likely won't work either)

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js
+++ b/main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js
@@ -679,6 +679,16 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
 
       $('<option>').attr("value", column2.name).text(column2.name).appendTo(elmts.noteColumnSelect);
     }
+
+    var currentHeight = dialog.outerHeight();
+    var currentWidth = dialog.outerWidth();
+    dialog.resizable({
+      alsoResize: ".dialog-border .dialog-body",
+      handles: "e, w, se",
+      minHeight: currentHeight,
+      maxHeight: currentHeight,
+      minWidth: currentWidth
+    });
   };
 
   MenuSystem.appendTo(menu, [ "core/transpose" ], [
--- a/main/webapp/modules/core/scripts/views/data-table/key-value-columnize.html
+++ b/main/webapp/modules/core/scripts/views/data-table/key-value-columnize.html
@@ -9,9 +9,9 @@
           <td bind="or_views_noteCol"></td>
         </tr>
         <tr>
-          <td><select bind="keyColumnSelect" size="15" style="width: 100%;"></select></td>
-          <td><select bind="valueColumnSelect" size="15" style="width: 100%;"></select></td>
-          <td><select bind="noteColumnSelect" size="15" style="width: 100%;"></select></td>
+          <td><select bind="keyColumnSelect" size="15" style="width: 100%; overflow: auto;"></select></td>
+          <td><select bind="valueColumnSelect" size="15" style="width: 100%; overflow: auto;"></select></td>
+          <td><select bind="noteColumnSelect" size="15" style="width: 100%; overflow: auto;"></select></td>
         </tr>
       </table></div>
     </div>
--- a/main/webapp/modules/core/scripts/views/data-table/key-value-columnize.html
+++ b/main/webapp/modules/core/scripts/views/data-table/key-value-columnize.html
@@ -2,7 +2,7 @@
   <div class="dialog-border">
     <div class="dialog-header" bind="dialogHeader"></div>
     <div class="dialog-body" bind="dialogBody">
-      <div class="grid-layout layout-normal layout-full grid-layout-for-ui"><table>
+      <div class="grid-layout layout-normal layout-full grid-layout-for-ui"><table style="table-layout: fixed">
         <tr>
           <td bind="or_views_keyCol"></td>
           <td bind="or_views_valCol"></td>


main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js
```diff
@@ -679,6 +679,16 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
 
       $('<option>').attr("value", column2.name).text(column2.name).appendTo(elmts.noteColumnSelect);
     }
+
+    var currentHeight = dialog.outerHeight();
+    var currentWidth = dialog.outerWidth();
+    dialog.resizable({
```
thadguidry: Is this what causes the whole dialog to have a resize handle that seems on Firefox to allow expanding/shrinking left-to-right?
image

obarabanov: not really. To facts:
1. resizing is only secondary feature requested in issue #898 to support long column names. Primary one is horizontal **scrolling** which doesn't work in FireFox at all. Only resizing works in FF, but I wasn't able to see such 'righ-side-only' stretching as your screenshot demoes. In my FF versiion (70+) resizing works properly, proportionally resizing all 3 lists and making them wider / thinner. What FF version this screenshot has been taken in ?
2. Existent html markup doesn't support vertical resizing. But in fact, vertical stretching is not required because it's about too long column names, not about column numbers.
3. Again. primary change was to add horizontal list scrolling, which was added by introducing style.overflow=auto. In reality, the only way to show full column name to user is having horizontal scrolling withing every list (not dialog window). That was added and it works fine in Chromium-based browsers as mentioned in PR description.

In fact, this PR is an attempt to provide quick remedy to the issue upon existent dialog codebase.
In Chromium, it solves the issue completely. In FireFox - only partially. Firefox is only roughly 5% share of total browsers used.
Any ideas  what platforms most of OpenRefine users are upon ? The point is, Safari @ MacOs doesn't support that horizontal scrolling for <select> tag.
It's about browser quircks as it's always been. Chrome is recommended.
Hopefully this helps ;)

thadguidry: 1. Yes, I have same result for FF (73.0.1) and it does needed make wider or thinner, so that's OK.
2. Thanks fine.
3. Yeap, understood.  This is definitely an improvement!  Thanks!

obarabanov:
@wetneb yep @thadguidry has mentioned similar rendering issue. Unfortunately I cannot reproduce that as it renders properly for me in both Chrome and Firefox on Windows 10. See my detailed explanation at https://github.com/OpenRefine/OpenRefine/pull/2352/files#r386523554

As Thad proposed to describe, we now officially recommend Chrome - https://github.com/OpenRefine/OpenRefine/wiki/Installation-Instructions#recommended-browsers
So I'd say nothing changed in Firefox as it doesn't support this scrollbars improvement.

wetneb:
> As Thad proposed to describe, we now officially recommend Chrome

I think this move would deserve a more thorough discussion than a passing comment on a PR.

obarabanov:
> 
> 
> > As Thad proposed to describe, we now officially recommend Chrome
> 
> I think this move would deserve a more thorough discussion than a passing comment on a PR.

sure, that's why it was on dev discussion board: https://groups.google.com/d/msg/openrefine-dev/AgAB4eWXSds/DasZu07qBgAJ

wetneb:
Well, I don't see any consensus there to say that "we now officially recommend Chrome".

obarabanov:
neither objections are there.

wetneb:
The discussion was started 17 hours ago :) Your enthusiasm and boldness are very welcome, but for decisions like this one it is better to give a bit more time to others to contribute. We are a large community spread on many time zones.

obarabanov:
> The discussion was started 17 hours ago :) 

great attention to details, thanks!

> Your enthusiasm and boldness are very welcome, but for decisions like this one it is better to give a bit more time for others to contribute. We are a large community spread on many time zones.

well I didn't expect simple wiki update was supposed to be considered by entire community over many time zones. Looks like Thad didn't either. I'm just starting familirize myself with the project's ecosystem on my spare time and didn't want to postpone such a quick suggestion.
We can remove it quickly if there are objections.

wetneb:
Sure, there was no harm done! :)

Have you thought about my suggestion above?
> How about adding table-layout: fixed to the `<table>` element?

thadguidry:
> 
> 
> > As Thad proposed to describe, we now officially recommend Chrome
> 
> I think this move would deserve a more thorough discussion than a passing comment on a PR.

Tidbit:  I didn't say the words "officially" above.  And I forget to add the words "historically".  So sorry! Context is important here and thanks to Antonin for correcting me.

Historically, we have recommended Chrome.  That is a true fact.  When folks asked on the mailing list which browser to use, we responded, try Chrome first.  This was because of 2 things, the way David tested with Jquery custom functions and we knew they worked well on Chrome and because Google was paying the bill for development at that time.  I often did the testing on both FF and Chrome, but if things worked in Chrome mostly, we shipped a release.

Today, we have no reason to recommend 1 browser over another.
Today, if a user asked on the mailing list today, our response should be "use whatever browser you prefer!"  and if they came back and said XYZ doesn't work, we would respond "have you tried another browser?"

Thanks Antonin for expanding this discussion on our mailing list and completely agree on the changes in the Wiki about that.

obarabanov:
> 
> @ostephens approved this pull request.
> Fixes for chromium based browsers, and does not negatively impact on Ff or Safari so happy to see this merged

I couldn't express the essence of this PR more concisely.

@wetneb 
> Have you thought about my suggestion above?
> 
> > How about adding `table-layout: fixed` to the `<table>` element?

it's not what's requested in PR's target ticket #898 which is not about pixel-perfect design.
If you think this dialog can't stand next 10 years without this layout change, then you're welcome to make the change separately according to "contributing recommendations" which I'm sure you know very well - https://github.com/OpenRefine/OpenRefine/blob/master/CONTRIBUTING.md#how-to-submit-prs-pull-requests-patches-and-bug-fixes

wetneb:
I think it makes sense to do this change in the same PR since they are intrinsically related - both impact the size of these column selection widgets.

Issue #525, opened 2012-10-15
Add a property to the « cell object » to get access to error text


tfmorris:
_Original author: tfmorris (January 26, 2012 21:30:45)_

In cases where the EvalError.message contains useful information (e.g. the error returned from a URL fetch), there's currently no way to get access to it.

I propose adding a new pseudofield to Cell which, if the Cell contains an EvalError, will return EvalError.message instead of the EvalError itself (which can not be evaluated in the normal manner).

The proposed name is &quot;error&quot; which would make the GREL syntax:

  cell.error 

The value field (i.e. cell.value) would continue to return the error object.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=525_

thadguidry:
@wetneb I noticed this https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/model/Cell.java#L85 and is there anything more needed to suffice for Tom's need on this issue?  If so, can you add some implementation details of what is further needed here?

wetneb:
Sure, the method to change is https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/model/Cell.java#L71 (this is where GREL and Python fields are resolved). A new case for `"error"` must be added there, the value it returns will be available as `cell.error` for users.

The annotations you have noticed have nothing to do with GREL / Python, they are only there for JSON serialization.

antoine2711:
@thadguidry & @wetneb: here, what we are returning is more an errorMessage than an error (which would be the object that we already have, and that some day, we might also want).

I believe it's still time to correct that…

What do you think?

Regards,
   Antoine

thadguidry:
@antoine2711 Hi, sorry I've been away for about a week or so. 

Hmm, the main use case we have always heard was that of being able to easily surface HTTP errors, even to the point of actually parsing our the HTTP error codes themselves.  I've heard this numerous times as a wishlist from lots of data journalists in newsrooms and librarians that use OpenRefine against APIs in-house and through subscription services they use.  So that's the primary driving use case, really around Fetch URLs (HTTP) error reporting, so that they can re-run the Fetch for those that were missed etc. for only those with certain error codes, message text contains, etc.

It's not the only case for surfacing error messages however.  But I cannot off the top of my head think of other OpenRefine operations other than HTTP inclusive of Recon.

[...]

PR #2363, opened 2020-03-07
Add cell.error field for error messages

zengchu2:
Attempt to address issue #525 .

Enable returning error messages to users when typing `cell.error` and return objects when typing `cell.value`

Test As Follows:
value
error


main/src/com/google/refine/model/Cell.java
```diff
@@ -61,7 +61,7 @@
     final public Serializable   value;
     @JsonIgnore
     final public Recon          recon;
-    
+
```
kushthedude: unwanted change here!

zengchu2:
Sure, no problem!

wetneb:
Nice, thank you! Currently, unit tests for the `Cell` class are gathered in `CellTests`, do you think you could add your test there instead of creating a dedicated class for it?
Also, you should not need to use `createCSVProject` for this: just creating a `Cell` with an `EvalError` as value, and checking that getting the `error` field returns the corresponding error message should do.

zengchu2:
Thanks for the quick suggestion, I will address this issue immediately!

PR #2366, opened 2020-03-07
Let's just put our philosophy on our README.md

thadguidry:


wetneb:
Good idea to use the README, but I would prefer to discuss this on user and dev mailing lists first.

thadguidry:
Sure, it's a PR, it can sit for ages (we had old PR's sit for over 5 years, remember? ;-)  until we decide how it should be worded, etc.

PR #2367, opened 2020-03-07
Migrate local dependencies to Maven Central.

### LINKED ISSUE #2365: Publish dependencies on Maven Central
Some of our dependencies are not available on Maven Central - they are old unmaintained libraries or our own forks of them. We currently ship their jars in the repository itself, but this is a problem since it requires installing these in the user's own local Maven repository first. This impedes workflows such as importing OpenRefine in Eclipse with m2e.

We should publish all our dependencies under our own `org.openrefine` groupId. This will make #2254 easier.

wetneb:
Instead of shipping jars in the local repository, we upload them to Maven Central under `org.openrefine.dependencies`.
The goal is to get rid of these dependencies eventually, or turn them into properly maintained projects.

I have left the launch4j jars as it should be easy to remove them and point to the next release when it is out. (The jars we have are only snapshots of their master branch - no added patches there).

Closes #2365.


thadguidry:
@wetneb Hey Antonin, so where's the publish going to happen exactly?  I don't see a [GitHub Action workflow](https://github.com/OpenRefine/OpenRefine/actions) for the publish event either.  Am I missing something somewhere?

wetneb:
The publishing is done, it's just a one off thing as we are not maintaining these libraries. The code for these libraries is not hosted in our repository so there would not be any use for GitHub actions here anyway.
We should migrate to properly maintained alternatives.

One exception is Butterfly, which I have already updated in the past, so I have set it up for deployment to Maven Central in our own fork:
https://github.com/OpenRefine/simile-butterfly

wetneb:
Merging this now to make the onboarding of interns easier.

PR #2375, opened 2020-03-08
fix: List facets dialogue should be non editable

### LINKED ISSUE #2374: List Facets dialogue box is editable
**Describe the bug**

List Facets dialogue box is editable, which should not be.

**To Reproduce**
1. Go to `Customised Facets`
2. Click on `Word Facet`
3. List all choices 

**Current Results**

You can edit the dialogue and remove or add anything which should not be applicable as it is listing the facets.

**Screenshots**

Screenshot 2020-03-09 at 4 04 34 AM
Screenshot 2020-03-09 at 4 04 39 AM

kushthedude:
Fixes #2374


thadguidry:
I agree with @wetneb I was not one of those users (who probably don't even use OpenRefine anymore honestly) that wanted to edit in-situ, instead I prefer a full featured text editor to handle the job after copying.

kushthedude:
@wetneb Is it good to be merged ?

Issue #1975, opened 2019-03-05
Extract Wikidata Q-numbers from reconciled items


ookgezellig:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
I often need to extract the Wikidata Q-values from reconciled items. I now do this via "Edit column --> Add column based on this column" and then using the GREL expression "cell.recon.match.id" to retrieve the Q-numbers

![afbeelding](https://user-images.githubusercontent.com/6546244/53798106-dcb34c80-3f37-11e9-9bd8-a076c824dcb8.png)

**Describe the solution you'd like**
It would be handy to have a function ("Extract Q-value", or some other description) under "Edit column" to automatically extract the Q-value and put these in a new adjecent column.

wetneb:
Related: https://github.com/wetneb/openrefine-wikibase/issues/57.

wetneb:
To do this, one would need to add an action in the menu, probably in `main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js`. This action should prompt the user for a column name and use the existing operation to create the column (no backend change should be needed).

madham32:
Hi! I'd like to be assigned this.

madham32:
![Screenshot from 2020-03-09 00-36-29](https://user-images.githubusercontent.com/47317143/76169839-1b5c9200-619e-11ea-9a4a-34aee1c81cb1.png)
Just got to add a function with the query expression now and we're good to go!

madham32:
Here's how the fix works:

![Screenshot from 2020-03-09 11-03-28](https://user-images.githubusercontent.com/47317143/76187016-027fcb00-61f6-11ea-91a9-3d22a230da9f.png)
![Screenshot from 2020-03-09 11-04-15](https://user-images.githubusercontent.com/47317143/76187019-07447f00-61f6-11ea-8f72-778449d9a185.png)
![Screenshot from 2020-03-09 11-09-42](https://user-images.githubusercontent.com/47317143/76187194-876ae480-61f6-11ea-8401-9f75e64fc938.png)
![Screenshot from 2020-03-09 11-04-59](https://user-images.githubusercontent.com/47317143/76187029-0ca1c980-61f6-11ea-837e-cde8fbc53be1.png)

CennoxX:
I think the menu item / column /... should be renamed to "Add ID values column" / "ID values" /... or something like that. Because this extracts all ids of the cells, only in the case of Wikidata this is a Q...-value.

[...]

PR #2378, opened 2020-03-09
Added feature to create column containing Q values

madham32:
Fix for issue https://github.com/OpenRefine/OpenRefine/issues/1975


madham32:
Understood! That should be easy

madham32:
Screenshot from 2020-03-10 17-30-18

madham32:
@ostephens Hope this is better!

ostephens:
Thanks @madham32. Almost there. Sorry to be really picky but could you:

* Lower case the first letter of "Entity" - we usually only use uppercase for the first word in the phrase (except for proper names)
* Use "identifiers" instead of "ids" (looking at what we've done elsewhere then this is more aligned with our current practice)

Then it's good to go for me

madham32:
Screenshot from 2020-03-10 18-12-58
@ostephens  Thats's alright. Here you go!

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -188,6 +189,49 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
     var level = DialogSystem.showDialog(frame);
   };
 
+  var doAddIdcolumn = function() {
+    var frame = $(
+      DOM.loadHTML("core", "scripts/views/data-table/add-q-column-dialog.html"));
+
+    var elmts = DOM.bind(frame);
+    elmts.dialogHeader.text($.i18n('core-views/add-id-col')+" " + column.name);
```
wetneb: Localizing text in this way makes it impossible for translators to reorder the column name to a different position in the message. I am really not blaming you since there are loads of places where we do that in the UI at the moment - I'll open an issue to clean these up.

Because we are introducing a new message here, let's try to get this right. We use the Wikimedia jQuery i18n plugin, which lets you define templates like this:
https://github.com/wikimedia/jquery.i18n#jqueryi18n-plugin

So you should be able to pass `column.name` as an extra argument to the function and insert it with `$1` in the message text.

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -188,6 +189,49 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
+        return;
+      }
+
+      Refine.postCoreProcess(
+        "add-column", 
+        {
+          baseColumnName: column.name,  
+          newColumnName: columnName, 
+          columnInsertIndex: columnIndex + 1,
+          onError: "cell.recon.match.id"
```
wetneb: I think `onError` is meant to determine what to store when evaluating the expression gives an error, so I don't think you can have an expression here? According to the "Add column" dialog it can be one of `set-to-blank`, `store-error` or `keep-original` (which probably only makes sense when editing an existing column). I would intuitively go for `set-to-blank` here.

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
madham32: Got it!

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
madham32: Oh sorry my bad. Fixed!

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
madham32: Not sure if I was supposed to click Resolve on my own xD Sent a final pull request!

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -518,7 +518,7 @@
     "core-views/word-facet": "Word facet",
     "core-views/add-id-column": "Add entity identifiers column",
     "core-views/add-id-column2": "Add a column containing entity identifiers from items",
-    "core-views/add-id-col": "Add column containing entity identifiers on column",
+    "core-views/add-id-col": "Add column containing entity identifiers on $1 {{plural:$1|column}}",
```
wetneb: Why use `{{plural}}` here? Isn't plural intended to be used for numbers? `$1` represents a string, not a number here.

madham32: I see. I thought that was a i18n parameter, my bad. Should I remove it all and keep just $1?

wetneb: `Add column containing entity identifiers on column $1` should do: `$1` will just be replaced by the column name.

madham32: Yep that's what I meant :+1:

madham32: Fixed :)

PR #2382, opened 2020-03-10
fix Google Sheets importer when URL is missing a trailing slash (#2380)

### LINKED ISSUE #2380: inifite preparing if missing slash after Google Spreadsheet URL
**Describe the bug**

When importing a Google spreadsheet, if we add a URL like `https://docs.google.com/spreadsheets/d/id_of_spreadsheet` (without "/" after it), then we will see an infinite "preparing" (you must login first).

issue3-preparing

We can see the following error in the console.

issue3-console

If we add "/" or "/any-path" after the URL above, it will works fine.

**To Reproduce**
1. Go to 'Create Project'
2. Click on 'Google Data'
3. Add spreadsheet URL like `https://docs.google.com/spreadsheets/d/id_of_spreadsheet`
4. Click 'Next'
5. Authorize OpenRefine with Google
6. See the error

**Expected behavior**

Even if we don''t have a "/" after URL like `https://docs.google.com/spreadsheets/d/id_of_spreadsheet`, the gdata extension should works fine. 

 - OS: Windows 10
 - Browser Version: Chrome
 - JRE or JDK Version: openjdk version "1.8.0_232"

I am using the dev version, because I have to manually add client_id / client_secret to `GoogleAPIExtension` java file and recompile it.

BTW, are there other options to add the id/secret pair?

**Datasets**

https://docs.google.com/spreadsheets/d/16L0JfpBWPfBJTqKtm-YU5-UBWLpkwXII-IRLMLnoKw0

afkbrb:
Close #2380 .

I changed the regex to extract the spreadSheetId from spreadsheet url and add corresponding tests for the new regex.

### DIFF
--- a/extensions/gdata/tests/src/com/google/refine/extension/gdata/GoogleAPIExtensionTest.java
+++ b/extensions/gdata/tests/src/com/google/refine/extension/gdata/GoogleAPIExtensionTest.java
@@ -1,48 +1,38 @@
 package com.google.refine.extension.gdata;
 
+import com.google.refine.ProjectManager;
 import org.testng.annotations.Test;
 
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static org.mockito.Mockito.mock;
 import static org.testng.Assert.*;
 
-public class SpreadsheetRETest {
+public class GoogleAPIExtensionTest {
 
-	/**
-	 * We cannot call {@link GoogleAPIExtension#extractSpreadSheetId(String url) } here,
-	 * otherwise, we'll get NullPointerException during its class initialization.
-	 *
-	 * So we just copy the regex and test it here.
-	 */
 	@Test
 	public void reTest() {
+		ProjectManager.singleton = mock(ProjectManager.class);
+
 		String spreadSheetId = "16L0JfpBWPfBJTqKtm-YU5-UBWLpkwXII-IRLMLnoKw0";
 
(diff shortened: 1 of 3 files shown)


extensions/gdata/tests/src/com/google/refine/extension/gdata/GoogleAPIExtensionTest.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.regex.Pattern;
+
+import static org.mockito.Mockito.mock;
+import static org.testng.Assert.*;
+
+public class GoogleAPIExtensionTest {
+
+	@Test
+	public void reTest() {
+		ProjectManager.singleton = mock(ProjectManager.class);
```
wetneb: This assignment of `ProjectManager.singleton` does not seem to be needed. Perhaps the test could also have a more explanatory name (making it clearer that it relates to `extractSpreadSheetId`).

afkbrb: We do need `ProjectManager.singleton = mock(ProjectManager.class)`, because GoogleAPIExtension.&lt;clinit&gt; will call `ProjectManager.singleton.getPreferenceStore()`, which will result in NullPointerException if `singleton` is null.

I'll write a comment for this in the test.

PR #2385, opened 2020-03-10
Corrects wrong trim label for json importer config

lisa761:
On clicking the `Trim leading & trailing whitespace from strings` checkbox in json importer configuration, the next checkbox i.e. `Parse cell text into numbers, dates, …` gets clicked. This is due to the wrong id in label for trim. This PR fixes that.


wetneb:
Closes #2388

Issue #1778, opened 2018-10-27
Edit summaries for Wikidata edit batches should have auto-complete


wetneb:
Previous edit summaries should be suggested when adding a summary to an edit group in the Wikidata extension, just like the Wikipedia edit interface suggests previous edit summaries.

wetneb:
The problem comes from the fact that the form isn't submitted:
https://stackoverflow.com/questions/15462991/trigger-autocomplete-without-submitting-a-form

wetneb:
This does not seem to be working with Chrome (only Firefox). https://github.com/wetneb/openrefine-wikibase/issues/51

PR #2387, opened 2020-03-11
Fix: Enable auto complete in chrome

zengchu2:
Addressed issue #1778. Now auto-fill works in both Chrome and Firefox.

**Test With Chrome:**
Before submitting:
chromB
After submitting with text **t1**:
chromA

**Test With Firefox**
Before submitting:
fb
After submitting with text **t 3**:
fa


wetneb:
This is great! By the way - make sure the email address you are using in your commits is associated to your GitHub account, so that GitHub can link your commits to your account. You might want to fix this commit before it gets merged. (either by changing your email address in the commit, by configuring git and  `git commit --amend` and `git push -f`, or by adding the email address used in the commit to your GitHub profile)

PR #2389, opened 2020-03-11
Fixed wrong and undefined stylings

AghaSaad04:

I have fixed LESS styling and removed the undefined ones.


AghaSaad04:
Yeah @wetneb though I have checked, but I think it would be better somebody else review the changes as well.

Issue #2307, opened 2020-02-10
Can't select a different sheet than the default in a gSheet importation


antoine2711:
**Describe the bug**
You can't select a different sheet in a gSheet (tabs at the bottom) in a gSheet importation with OR. All the names of the sheets of spreadsheets are there and good, but choosing one of them just make the default one reload.

**To Reproduce**
1. Go to 'Create project' -> 'Google Data'
2. Sign in a Google Account
3. Choose a gSheet' spreadsheet.
4. Try to change the sheet of the chosen spreadsheet. Does not work.

**Current Results**
Nothing. Just a feeling of reload.

**Expected behavior**
See the data related to the chosen sheet.

**Screenshots**
Here's a short video:
![OpenRefine-Import-gSheet-sheets](https://user-images.githubusercontent.com/12217525/74133520-bcaff280-4bb6-11ea-8532-6929bdf95659.gif)

**Desktop (please complete the following information):**
 - OS: Mac OS X.14.6
 - Browser Version: Chrome latest
 - JRE or JDK Version: 13.0.1

**OpenRefine (please complete the following information):**
 - Version 3.3

**Additional context**
Relates to this Google Groups' discussion : [Problem changing the sheet to import with Google Sheets (thru import creating a project)](https://groups.google.com/forum/#!topic/openrefine/otgLKSkpQgM)

buntamas:
Hi everyone. I would like to solve this bug. Can I work on this issue?

wetneb:
Yes you can!

antoine2711:
> Hi everyone. I would like to solve this bug. Can I work on this issue?

I can't wait to try this out! Great. Thank you.

Regards, Antoine

PR #2390, opened 2020-03-11
Fixed that can't select a different sheet #2307

buntamas:
**Refers**: _Can't select a different sheet than the default in a gSheet importation_ #2307


antoine2711:
Great job. I've tried it, and so far, everything works well.

Thank you.

Regards,
   Antoine

PR #2401, opened 2020-03-12
Use parentheses for JAVA_OPTIONS on Unix/Linux

### LINKED ISSUE #2396: Misleading JAVA_OPTIONS configuration samples in refine.ini
In [refine.ini](https://github.com/OpenRefine/OpenRefine/blob/master/refine.ini), we have two samples for JAVA_OPTIONS configuration:

```ini
#JAVA_OPTIONS=-XX:+UseParallelGC -verbose:gc -Drefine.headless=true
#JAVA_OPTIONS=-Drefine.data_dir=C:\Users\user\AppData\Roaming\OpenRefine
```

These configurations works fine on Windows. But on linux, if we uncomment the first sample, we'll see the following error:

issue4-error

That's because `JAVA_OPTIONS=-XX:+UseParallelGC -verbose:gc -Drefine.headless=true` is processed as `export JAVA_OPTIONS=-XX:+UseParallelGC -verbose:gc -Drefine.headless=true` in the [refine](https://github.com/OpenRefine/OpenRefine/blob/master/refine) shell script. But the export result is `JAVA_OPTIONS=-XX:+UseParallelGC`, `-verbose:gc` and `-Drefine.headless=true` are processed as identifiers.

Seems we can fix this by rewriting the configuration:

```ini
JAVA_OPTIONS="-XX:+UseParallelGC -verbose:gc -Drefine.headless=true"
```
However, we'll get another error:

issue4-error1

That's because the new configuration is interpreted as `"-XX:+UseParallelGC -verbose:gc -Drefine.headless"="true"`.

Actually, we should use `JAVA_OPTIONS=(-XX:+UseParallelGC -verbose:gc -Drefine.headless=true)`. Because `JAVA_OPTIONS` is processed as an array of JVM options in `refine` script.

So, my opinion is that we should tell Unix/Linux users to use parentheses for this configuration.

afkbrb:
Close #2396 .

Just added some comments to clarify the configuration of JAVA_OPTIONS on Unix/Linux.

### DIFF
--- a/refine.ini
+++ b/refine.ini
@@ -15,11 +15,8 @@ REFINE_MIN_MEMORY=1400M
 
 # Some sample configurations. These have no defaults.
 #JAVA_HOME=C:\Program Files\Java\jdk1.8.0_151
-# For Windows users, configure JAVA_OPTIONS like below
 #JAVA_OPTIONS=-XX:+UseParallelGC -verbose:gc -Drefine.headless=true
 #JAVA_OPTIONS=-Drefine.data_dir=C:\Users\user\AppData\Roaming\OpenRefine
-# For Unix/Linux users, remember to add parentheses
-#JAVA_OPTIONS=(-XX:+UseParallelGC -verbose:gc -Drefine.headless=true)
 
 # Uncomment to increase autosave period to 60 mins (default: 5 minutes) for better performance of long-lasting transformations
 #REFINE_AUTOSAVE_PERIOD=60
--- a/refine.ini
+++ b/refine.ini
@@ -15,8 +15,11 @@ REFINE_MIN_MEMORY=1400M
 
 # Some sample configurations. These have no defaults.
 #JAVA_HOME=C:\Program Files\Java\jdk1.8.0_151
+# For Windows users, configure JAVA_OPTIONS like below
 #JAVA_OPTIONS=-XX:+UseParallelGC -verbose:gc -Drefine.headless=true
 #JAVA_OPTIONS=-Drefine.data_dir=C:\Users\user\AppData\Roaming\OpenRefine
+# For Unix/Linux users, remember to add parentheses
+#JAVA_OPTIONS=(-XX:+UseParallelGC -verbose:gc -Drefine.headless=true)
 
 # Uncomment to increase autosave period to 60 mins (default: 5 minutes) for better performance of long-lasting transformations
 #REFINE_AUTOSAVE_PERIOD=60
--- a/refine.ini
+++ b/refine.ini
@@ -15,6 +15,7 @@ REFINE_MIN_MEMORY=1400M
 
 # Some sample configurations. These have no defaults.
 #JAVA_HOME=C:\Program Files\Java\jdk1.8.0_151
+# Notice, only the last uncommented JAVA_OPTIONS will work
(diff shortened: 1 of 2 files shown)


fpompermaier:
I personally don't like to have different JAVA_OPTIONS in the .ini file. Woulnd't be better to change only the .bat or .sh (actually .refine) files in order to handle them appropriately?

wetneb:
So you would edit the scripts themselves, rather than the `.ini` file? Intuitively it is better to separate configuration from code, no?

afkbrb:
Yeah, I'll try to edit `refine`.

thadguidry:
The syntax works on Windows, however, what Stefano did was to just show some examples in the .ini file.  They were to be used as a guide.  Also, as a user, you cannot uncomment 2 `JAVA_OPTIONS` lines, the refine.bat only processes 1 single `JAVA_OPTIONS` line, if both are uncommented, the last one specified overrides the others.

I think it would be wise to remove one of them, such as the `JAVA_OPTIONS` line with `-XX:+UseParallelGC`

**Please don't change the `refine.bat`, it's fine actually.**
Instead, just put more comments into the `refine.ini` itself to explain the above and anything else.

fpompermaier:
> So you would edit the scripts themselves, rather than the `.ini` file? Intuitively it is better to separate configuration from code, no?

I'm not saying that the .bat or .sh are currently bugged, I'm just saying that there's should be only one syntax for the java options. If they are handled differently in Windows vs Linux it's simpler to fix the .bat or .sh file in order to handle them correctly.

The PR just add brackets around the JAVA_OPTS in the Linux case..In that case it's MUCH safer to do that in the code wrt expect the user to configure that correctly IMHO.

afkbrb:
@fpompermaier Agree. Now I've made it possible.

wetneb:
Makes sense, thanks both!

refine.ini
```diff
@@ -15,6 +15,7 @@ REFINE_MIN_MEMORY=1400M
 
 # Some sample configurations. These have no defaults.
 #JAVA_HOME=C:\Program Files\Java\jdk1.8.0_151
+# Notice, only the last uncommented JAVA_OPTIONS will work
```
thadguidry: Improve the comment with the following:
```suggestion
# Use a single JAVA_OPTIONS that includes any JVM options you need upon OpenRefine startup
```

refine.ini
```diff
@@ -15,6 +15,7 @@ REFINE_MIN_MEMORY=1400M
...
 #JAVA_OPTIONS=-XX:+UseParallelGC -verbose:gc -Drefine.headless=true
```
thadguidry:
```suggestion
#JAVA_OPTIONS=-XX:+UseParallelGC -verbose:gc -Drefine.headless=true -Drefine.data_dir=C:\Users\user\AppData\Roaming\OpenRefine
```

thadguidry:
```suggestion

```

refine.ini
```diff
@@ -15,6 +15,7 @@ REFINE_MIN_MEMORY=1400M
...
 #JAVA_OPTIONS=-Drefine.data_dir=C:\Users\user\AppData\Roaming\OpenRefine
```
thadguidry:
```suggestion

```

afkbrb:
Wow, three approvals now! Thanks for your suggestions and approvals:)

Issue #1792, opened 2018-10-31
Data loss/ column mismatch when importing multiple sheets from xlsx


burgerga:
**Describe the bug**
I want to import data from multiple sheets in one excel file, but somehow data from column 2 ends up in column 17 and is partly lost, because there is already data in column 17

**To Reproduce**
1. Open attached xlsx file
2. Uncheck "Parse next 1 line(s) as column headers"
3. Uncheck "Store blank rows"
4. (Optional) Check "Store file source (file names, URLs) in each row
5. Preview shows unexpected results
6. Uncheck "test.xlsx#Sheet1" from "Worksheets to Import"
7. Preview shows expected Sheet2 results

(If you click "Create Project" you'll end up with the situation as in the preview, so it's not a preview problem)

**Current Results**
- Data from Column 2 in Sheet2 ends up in Column 17, however first part of Column 2 is lost because there is already data in Column 17 in Sheet2.
- Also, (not sure if related), Sheet2 shows a ridiculous amount of rows to import

**Expected behavior**
Data from Column 2 in Sheet2 should stay in Column 2

**Screenshots**
What happens:
![screenshot from 2018-10-31 10-06-02](https://user-images.githubusercontent.com/907362/47777248-9d0ff600-dcf4-11e8-8c78-850f6861dbd5.png)

What Sheet2 should look like: 
![screenshot from 2018-10-31 10-09-28](https://user-images.githubusercontent.com/907362/47777469-19a2d480-dcf5-11e8-9c49-429f509f8bd5.png)

**Desktop:**
 - OS: Manjaro Linux
 - Browser Version: Chrome 70.0.3538.77 (Official Build) (64-bit)
 - JRE or JDK Version: openjdk version "1.8.0_192"

**OpenRefine:**
 - Version 3.0 [TRUNK]

**Datasets**
Sort of minimal working example: 
[test.xlsx](https://github.com/OpenRefine/OpenRefine/files/2533471/test.xlsx)
Derived from a larger, but private, data set.

wetneb:
I can reproduce this bug on master.

zengchu2:
Hi, I am now currently working on this issue. It seems that the problem occurs due to the false update for request `update-format-and-options` is sent and so the incorret row model is returned. 
Are there any hints or docs related to the server side job updating, that would speed up my process quickly ! 

Thanks in advanced!

tfmorris:
I think this was intended to be closed when #2404 was merged.

PR #2404, opened 2020-03-13
Fix: Data losses when importing multiple sheets from same Excell file

zengchu2:
Issue as noted as issue #1792 . 
Generally speaking, this error can occur when importing multiple sheets from one source file with some later sheets have more columns than the a previous one and the columnns share the same name.

**Cause of Issue**
In the test file, _sheet2_ has more columns than _sheet1_ and each column has the same name in both sheets.

When loading the second sheet, in lin 113 of `TabularImportingParserBase`, the columnNames is initialized to `[]` while the names of _column1_ to _column15_ are stored already in the _project_ variable.  

Without reusing the previous column names, _column 16_ is then treated as the 1st column when reading data from _sheet 2_ (since columnNames is initialized to `[]` instead of being `[column1 ... column15]`)

**Testing**
- Pass `./refine test`

- Belows are the same test case reproduced from as described in issue #1792 . I changed some values from the test.xls to make it more readable.
Preview after setting up the read options:
after
Showing sheet two only:
single

attached test file: 
[test.xlsx](https://github.com/OpenRefine/OpenRefine/files/4327594/test.xlsx)

### DIFF
--- a/main/src/com/google/refine/importers/ImporterUtilities.java
+++ b/main/src/com/google/refine/importers/ImporterUtilities.java
@@ -139,6 +139,10 @@ public class ImporterUtilities {
                         // Already taken name
                         i++;
                     } else {
+                        // Want to update currentFileColumnNames
+                        if(! currentFileColumnNames.contains(columnName)){
+                            currentFileColumnNames.add(columnName);
+                        }
                         return column;
                     }
                 } else {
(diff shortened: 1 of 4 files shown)


zengchu2:
Absoultely, my bad for not giving some extra test on this!
I would suggest adding test in `ExcellImporterTest` in this pr. 
If there are further test needed in other importers, I might need another PR and that will be done in short time!

zengchu2:
Hi, I have managed to fix this issue by modifying `main/src/com/google/refine/importers/ImporterUtilities.java` instead of `TabularImportingParserBase`.

The issue introduced in issue #1792 is due to the fact that `columnNames` is not updated properly under specific condition and hence the cells are put in the wrong column.  I tried  to test with importing multile csv/xls/xlsx files and it seems good now. 

Any suggestions will be appreciated !

Issue #626, opened 2012-10-15
Butterfly classloader deadlock


tfmorris:
_Original author: tfmorris (October 03, 2012 15:03:13)_

This is the second time I've seen this deadlock.  The first time I encountered it, pausing and resuming the server was enough to clear it, so I'm going to record all the info this time before debugging more!

The lines that I've marked with asterisks represent the synchronized methods/blocks according to Eclipse's display.  There are many threads blocked with the stack in the state of the last two traces below, but I've only include two.  The third trace is the odd man out and is the other half of the deadlock.  The &quot;classes&quot; attempting to be loaded vary, but always seem to be package names, including &quot;com.google.refine&quot; &quot;com&quot; &quot;org&quot; &quot;edu&quot; 

Thread [pool-1-thread-13](Suspended)  
    URLClassLoader$1.run() line: not available  
    URLClassLoader$1.run() line: not available  
    AccessController.doPrivileged(PrivilegedExceptionAction&lt;T&gt;, AccessControlContext) line: not available [native method] 
    Launcher$AppClassLoader(URLClassLoader).findClass(String) line: not available  
    Launcher$AppClassLoader(ClassLoader).loadClass(String, boolean) line: not available 
    Launcher$AppClassLoader.loadClass(String, boolean) line: not available  
    Launcher$AppClassLoader(ClassLoader).loadClass(String) line: not available  
-   WebAppClassLoader.loadClass(String, boolean) line: 401  
  WebAppClassLoader.loadClass(String) line: 363  
  ButterflyClassLoader.loadClass(String, boolean) line: 61  
  ButterflyClassLoader(ClassLoader).loadClass(String) line: not available 
  Kit.classOrNull(ClassLoader, String) line: 92  
-   NativeJavaTopPackage(NativeJavaPackage).getPkgProperty(String, Scriptable, boolean) line: 159  
  NativeJavaTopPackage(NativeJavaPackage).get(String, Scriptable) line: 114  
  NativeJavaTopPackage.init(Context, Scriptable, boolean) line: 128  
  GeneratedMethodAccessor32.invoke(Object, Object[]) line: not available  
  DelegatingMethodAccessorImpl.invoke(Object, Object[]) line: not available  
  Method.invoke(Object, Object...) line: not available  
  ScriptableObject.buildClassCtor(Scriptable, Class&lt;T&gt;, boolean, boolean) line: 1030  
  LazilyLoadedCtor.buildValue() line: 110 
[...]

Issue #416, opened 2012-10-15
IllegalStateException: Committed when fetching favicon


tfmorris:
_Original author: tfmorris (June 23, 2011 20:40:35)_

Got this weird error message on the console when starting up Refine today.  Don't have time to investigate right now...

16:00:16.846 [                butterfly] GET /images/logo-googlerefine40.png [ain](0ms)
16:00:16.877 [          org.mortbay.log] /favicon.ico (31ms)
java.lang.IllegalStateException: Committed
        at org.mortbay.jetty.Response.resetBuffer(Response.java:1024)
        at javax.servlet.ServletResponseWrapper.resetBuffer(ServletResponseWraper.java:202)
        at org.mortbay.servlet.GzipFilter$GZIPResponseWrapper.resetBuffer(GzipFilter.java:310)
        at org.mortbay.servlet.GzipFilter$GZIPResponseWrapper.sendError(GzipFilter.java:319)
        at edu.mit.simile.butterfly.Butterfly.error(Butterfly.java:1020)
        at edu.mit.simile.butterfly.Butterfly.service(Butterfly.java:528)
        at com.google.refine.RefineServlet.service(RefineServlet.java:180)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:820)
        at org.mortbay.jetty.servlet.ServletHolder.handle(ServletHolder.java:51x)
        at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1166)
        at org.mortbay.servlet.UserAgentFilter.doFilter(UserAgentFilter.java:81
```
    at org.mortbay.servlet.GzipFilter.doFilter(GzipFilter.java:132)
    at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(Servle
```

Handler.java:1157)
        at org.mortbay.jetty.servlet.ServletHandler.handle(ServletHandler.java: 88)
        at org.mortbay.jetty.security.SecurityHandler.handle(SecurityHandler.ja a:216)
        at org.mortbay.jetty.servlet.SessionHandler.handle(SessionHandler.java: 82)
        at org.mortbay.jetty.handler.ContextHandler.handle(ContextHandler.java: 65)
        at org.mortbay.jetty.webapp.WebAppContext.handle(WebAppContext.java:418
```
    at org.mortbay.jetty.handler.HandlerWrapper.handle(HandlerWrapper.java:
```

52)
[...]

PR #2405, opened 2020-03-13
Remove duplicate metadata command registration

thadguidry:
This removes a duplicate `set-project-metadata` command during Butterfly initialization, which was already done on L72.
Hmm, wondering if this might also fix any issues that we already have regarding startup?


thadguidry:
Ah found the one I was thinking of #416 but its unrelated.
And #626 is probably fixed now since we use Java 8 and ThreadPool issues were fixed in Java 7/8? (dunno, it could probably be closed, haven't seen Tom's error ever since 2013)

PR #2407, opened 2020-03-13
Fixes no selected record element error

### LINKED ISSUE #2402: 'Updating Preview' gets stuck in an infinite loop on clicking any config option in XML
**Describe the bug**
The 'Update Preview' gets stuck in an infinite loop if the user does not select a record element first before clicking on any configuration option when working with XML datasets.

**To Reproduce**
1. Go to Create Project
2. Open any XML dataset (e.g. https://www.govinfo.gov/bulkdata/FR/2020/03/FR-2020-03-09.xml)
3. Click any configuration option (e.g. Load at Most checkbox)

**Current Results**
Update Preview gets stuck and there is an error in console `Index 0 out of bounds for length 0`

**Expected behavior**
There should be a message that tells the user to first select an element

**Screenshots**

lisa761:
Closes #2402

### DIFF
--- a/main/webapp/modules/core/scripts/index/parser-interfaces/xml-parser-ui.js
+++ b/main/webapp/modules/core/scripts/index/parser-interfaces/xml-parser-ui.js
@@ -43,7 +43,6 @@ Refine.XmlParserUI = function(controller, jobID, job, format, config,
   this._dataContainer = dataContainerElmt;
   this._progressContainer = progressContainerElmt;
   this._optionContainer = optionContainerElmt;
-  this._recordElementSelected = false;
 
   this._timerID = null;
   this._initialize();
@@ -132,7 +131,7 @@ Refine.XmlParserUI.prototype._initialize = function() {
     this._optionContainerElmts.includeFileSourcesCheckbox.prop("checked", true);
   }
   this._optionContainerElmts.pickRecordElementsButton.click(function() {
-    self._recordElementSelected = false;
+    self._config.recordPath = undefined;
     self._showPickRecordElementsUI();
   });
 
@@ -227,7 +226,6 @@ Refine.XmlParserUI.prototype._showPickRecordElementsUI = function() {
         div.removeClass('highlight');
       })
       .click(function(evt) {
-        self._recordElementSelected = true;
         if (hittest(evt)) {
           self._setRecordPath(path);
         }
@@ -261,7 +259,7 @@ Refine.XmlParserUI.prototype._setRecordPath = function(path) {
 Refine.XmlParserUI.prototype._updatePreview = function() {
   var self = this;
   
-  if(!self._recordElementSelected) {
+  if(!this._config.recordPath){
     window.alert($.i18n('core-index-import/warning-record-path'));
     return;
   }
--- a/main/webapp/modules/core/scripts/index/parser-interfaces/xml-parser-ui.js
+++ b/main/webapp/modules/core/scripts/index/parser-interfaces/xml-parser-ui.js
@@ -31,6 +31,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
  */
 
+var recordElementSelected = false;
+
 Refine.XmlParserUI = function(controller, jobID, job, format, config,
     dataContainerElmt, progressContainerElmt, optionContainerElmt) {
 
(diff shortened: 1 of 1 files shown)


main/webapp/modules/core/scripts/index/parser-interfaces/xml-parser-ui.js
```diff
@@ -31,6 +31,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
  */
 
+var recordElementSelected = false;
```
wetneb: I think it'd be cleaner if this was not a global variable, I would store it as an attribute of the importer.

main/webapp/modules/core/scripts/index/parser-interfaces/xml-parser-ui.js
```diff
@@ -257,6 +260,11 @@ Refine.XmlParserUI.prototype._setRecordPath = function(path) {
 
 Refine.XmlParserUI.prototype._updatePreview = function() {
   var self = this;
+  
+  if(!self._recordElementSelected) {
+    alert("No record element selected! Please select one first.");
```
wetneb: Oh, and that's clearly something we want to localize using i18n. And perhaps avoid the exclamation mark in the message.

wetneb: Hang on, we already have this message: `core-index-import/warning-record-path`.

wetneb: This seems to duplicate the error message here: https://github.com/OpenRefine/OpenRefine/pull/2407/files#diff-92cdef36b4cf54a220fa4a4a7acc5107R65

main/webapp/modules/core/scripts/index/parser-interfaces/xml-parser-ui.js
```diff
@@ -262,7 +262,7 @@ Refine.XmlParserUI.prototype._updatePreview = function() {
   var self = this;
   
   if(!self._recordElementSelected) {
-    alert("No record element selected! Please select one first.");
+    window.alert($.i18n('core-index-import/warning-record-path'));
```
wetneb: It's still not clear to me why we have to duplicate the existing check, where the same message is used?

lisa761: The existing check works only when the user clicks the `Create Project` button. The `Update Preview` gets stuck if the user does not select one and clicks on any other configuration option. Once that happens, the user cannot do anything (and this happens before Create Project), there isn't any check for this right now. Maybe we can remove the one in `Create Project` because the check in `Update Preview` will make sure that a Record Element is always selected?

lisa761: I think this happens because once `_updatePreview` calls `._getOptions` where the recordPath is set to a defualt value when there isn't already one, which in this case is an empty array. Once the flow of control reaches `XMLImportUtilities`, an error ArrayIndexOutOfBoundsException is generated in `findRecord` because `recordPath` is 0 and we cannot access its 0th index.

wetneb: It looks like we could potentially remove the boolean you introduce and do the same check as for the other `window.alert`, no?

lisa761: Yes okay, I wll do that then. Should I also remove the check from confirmCreateProject then?

wetneb: If that is redundant, sure.

main/webapp/modules/core/scripts/index/parser-interfaces/xml-parser-ui.js
```diff
@@ -131,6 +131,7 @@ Refine.XmlParserUI.prototype._initialize = function() {
     this._optionContainerElmts.includeFileSourcesCheckbox.prop("checked", true);
   }
   this._optionContainerElmts.pickRecordElementsButton.click(function() {
+    self._config.recordPath = undefined;
```
lisa761: I did this because if I did `setRecordPath`, the flow reaches `updatePreview` first and then the same thing happened. Also I let the check in `confirmCreateProject` be beacuse that would be required when someone directly clicks that button.

PR #2408, opened 2020-03-13
Adds trim whitespace option to separator based files

### LINKED ISSUE #791: Option to strip leading/trailing whitespace on import of csv/tsv/seperator format files
When creating a project from a csv/tsv/seperator format file, there is no checkbox to strip leading/trailing whitespace. This is already an option on importing JSON and XML - would be good if it could be an option for csv/tsv as well

lisa761:
Closes #791

### DIFF
--- a/main/src/com/google/refine/importers/SeparatorBasedImporter.java
+++ b/main/src/com/google/refine/importers/SeparatorBasedImporter.java
@@ -76,6 +76,7 @@ public class SeparatorBasedImporter extends TabularImportingParserBase {
         JSONUtilities.safePut(options, "guessCellValueTypes", false);
         JSONUtilities.safePut(options, "processQuotes", true);
         JSONUtilities.safePut(options, "quoteCharacter", String.valueOf(CSVParser.DEFAULT_QUOTE_CHARACTER));
+        JSONUtilities.safePut(options, "trimStrings", false);
 
         return options;
     }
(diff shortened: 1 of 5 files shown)


main/src/com/google/refine/importers/TabularImportingParserBase.java
```diff
@@ -161,6 +163,9 @@ static public void readTable(
                             
                             Object value = cells.get(c);
                             if (value instanceof Cell) {
+                                if(trimStrings) {
+                                    value = ((Cell) value).value.toString().trim();
```
wetneb: If `value` is an instance of `Cell`, then you should not replace it by a `String`. This will cause runtime type errors.

lisa761:
@wetneb Sure, I will get to it

lisa761:
@wetneb Tell me if there needs to be more changes. I understand the problem now, so `trimString` now works only for string values, as it should have before (according to the checkbox). Thanks for the review :)

main/src/com/google/refine/importers/SeparatorBasedImporter.java
```diff
@@ -76,6 +76,7 @@ public ObjectNode createParserUIInitializationData(ImportingJob job,
         JSONUtilities.safePut(options, "guessCellValueTypes", false);
         JSONUtilities.safePut(options, "processQuotes", true);
         JSONUtilities.safePut(options, "quoteCharacter", String.valueOf(CSVParser.DEFAULT_QUOTE_CHARACTER));
+        JSONUtilities.safePut(options, "trimStrings", false);
```
wetneb: We could potentially set this to `true` instead - this would enable the option by default in the UI. The importer would still behave in the same way if using an importer configuration generated by a previous version, since you use `false` as default value in `getBoolean` in `readTable`.

I don't have strong feelings about whether this should be enabled by default in the UI. It would perhaps make it more consistent with the other importers which already support trimming strings. And I would expect that workflows where preserving whitespace around strings are quite rare for TSV/CSV. So leaning towards enabling… @ostephens what do you think?

lisa761: Yes, that would be a better option, I think.

thadguidry: 👍 "enabled", it's a smart default.

PR #2410, opened 2020-03-14
Mock reconciliation service

### LINKED ISSUE #2398: Mock reconciliation calls in Wikitext importer tests
The Wikitext importer makes calls to the Wikidata reconciliation service at import time.
We should make sure these calls are mocked in its tests.

afkbrb:
Close #2398 

Now the reconciliation service used by WikitextImporter is successfully mocked.


afkbrb:
BTW, `PreviewWikibaseSchemaCommandTest.testValidSchema` makes call to Wikidata during inspection, and it's the last test that depends on external services. Should I open a new issue for that?

PR #2412, opened 2020-03-14
Mock Wikidata service

### LINKED ISSUE #2411: Mock Wikidata service for PreviewWikibaseSchemaCommandTest
PreviewWikibaseSchemaCommandTest calls Wikidata service during test. We should mock the service, similar to #2398 .

afkbrb:
Close #2411 

Glad to announce that OpenRefine doesn't depend on any external web service during test now!


wetneb:
Nice! But it looks like PowerMock does not quite work as it should on Windows, according to Appveyor.

In general I think we should avoid PowerMock as much as we can - when possible, refactoring the code to make it more testable might be a better option (but I know it can be quite tricky in many cases).

afkbrb:
Weird. I'm using Windows to test and build it, works fine. I'll try to figure out what happened.

afkbrb:
@wetneb I couldn't find what happened from the log AppVeyor provided. Is it possible to run AppVeyor build again without a new commit?

wetneb:
Looks like it was a passing failure, the new build passes. Perhaps there is some sort of race condition there. Here is [the Appveyor log](https://github.com/OpenRefine/OpenRefine/files/4332932/log.1.txt) for the passing build.

wetneb:
Ok, I re-run the build a couple of times and got a green light each time, so I guess we can merge this. If we experience unstable Appveyor builds in the future we might want to revert though (I have observed issues with PowerMock on Appveyor before).

afkbrb:
Thanks, seems we do need to avoid using PowerMock if possible.

thadguidry:
@afkbrb You'll love JMockit, we did at Ericsson for enforcing consistent structure in blocks after switching from Mockito.

Issue #2230, opened 2019-11-22
Reconciliation UI: checkboxes not taken into account


wetneb:
**Describe the bug**

Open the reconciliation dialog on some column (Reconcile -> Start reconciling).
In the panel on the right-hand side, tick a column to include it in the information used for the reconciliation. Enter a property and validate it. Untick the checkbox.
If you run the reconciliation, you can observe that the property is still included in the reconciliation settings! So the status of the checkbox is ignored.

Instead, we should:
* clear the property field
* not include it in the reconciliation settings

Or, we could also decide that these checkboxes are redundant: a column should be included if and only if the user has input a corresponding property name in the field next to it. In this case, we should make sure that if the service has a property suggest service, the UI makes it clear whether a given input has been validated by the suggest service (i.e., whether we have obtained a property id for the user input).

PR #2413, opened 2020-03-14
Reconciliation UI: property not included if checkbox unchecked

darecoder1999:
solves #2230


main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -287,9 +288,9 @@ ReconStandardServicePanel.prototype.start = function() {
   var columnDetails = [];
   $.each(
     this._panel.find('input[name="property"]'),
-    function() {
+    function(index) {
       var property = $(this).data("data.suggest");
-      if (property && property.id) {
+      if (property && property.id && include[index].checked) {
```
darecoder1999: Condition is added to include only those fields that are checked.

Issue #2403, opened 2020-03-12
Silent errors when using JSON importer


thadguidry:
**Describe the bug**
Errors in parsing JSON files with Json Importer are not shown to user, but instead are silent and shown only in console. 

**To Reproduce**
1. Use Clipboard with the following:
```
{
"stats":[
"goal":"1",
"time":".4"
],
"stats":[
"goal":"1",
"time":".3"
],
"stats":[
"goal":"1",
"time":".2"
]
}
```
2. Click Next
3. Choose JSON importer
4. The preview of some elements are shown (although not completely because of a JSON structure problem in the provided test Clipboard, but the user doesn't know this quite yet)
5. Click checkbox "Load at most"
6. See error in console, and JSON elements disappear from preview (Silent Error)

**Current Results**
```
09:10:45.530 [       XmlImportUtilities] Exception from XML parse (124ms)
com.google.refine.importers.tree.TreeReaderException: com.fasterxml.jackson.core.JsonParseException: Unexpected character (':' (code 58)): was expecting comma to separate Array entries
 at [Source: (com.google.refine.importers.ImporterUtilities$2); line: 3, column: 8]
        at com.google.refine.importers.JsonImporter$JSONTreeReader.next(JsonImporter.java:361)
        at com.google.refine.importers.tree.XmlImportUtilities.processSubRecord(XmlImportUtilities.java:493)
        at com.google.refine.importers.tree.XmlImportUtilities.processSubRecord(XmlImportUtilities.java:495)
        at com.google.refine.importers.tree.XmlImportUtilities.processRecord(XmlImportUtilities.java:361)
        at com.google.refine.importers.tree.XmlImportUtilities.findRecord(XmlImportUtilities.java:323)
        at com.google.refine.importers.tree.XmlImportUtilities.importTreeData(XmlImportUtilities.java:263)
        at com.google.refine.importers.tree.TreeImportingParserBase.parseOneFile(TreeImportingParserBase.java:223)
        at com.google.refine.importers.JsonImporter.parseOneFile(JsonImporter.java:206)
        at com.google.refine.importers.tree.TreeImportingParserBase.parseOneFile(TreeImportingParserBase.java:120)
        at com.google.refine.importers.tree.TreeImportingParserBase.parse(TreeImportingParserBase.java:86)
        at com.google.refine.importing.ImportingUtilities.previewParse(ImportingUtilities.java:961)
[...]

PR #2414, opened 2020-03-14
Fix silent error in JSON/XML importers

zengchu2:
Address issue #2403 .
**Cause of issue**
Exception is caught but not thrown at `XmlImportUtilities.java`.

**Implementation**
1. Add exception handler at  `XmlImportUtilities.java`, `TreeImportingParserBase.java `
2. Pass more user friendly message in `DefaultImportingController.java ` when writing exceptions into response.

**Test**
- Try with error mentioned in issue #2403, and the expected result is aquired.
getError

- Implement test method at `JsonImporterTests.java `


main/src/com/google/refine/importing/DefaultImportingController.java
```diff
@@ -290,7 +292,12 @@ static public void writeErrors(JsonGenerator writer, List<Exception> exceptions)
             e.printStackTrace(new PrintWriter(sw));
             
             writer.writeStartObject();
-            writer.writeStringField("message", e.getLocalizedMessage());
+            String errorMessage = e.getLocalizedMessage();
+            if(e instanceof TreeReaderException) {
+                errorMessage += "\nPlease correct the format of your input file, or choose another" +
```
thadguidry: We should probably add 1 more new line to separate from the error itself more in the dialog and make it stand out.
Also, will this get localized properly for our translations support?

zengchu2: Sadly the whole error message is shown only in English under other languages. 
Should we only localized the latter part ("Please correct your format ... ") and keep the error message unchanged ?

thadguidry: Yes, please, if it's not too much trouble.

zengchu2: Hi, I manged to fix this problem in front end .
Here is my test result with incorrect json format input string:

image
微信图片_20200315020831

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
```diff
@@ -375,7 +375,8 @@ Refine.TagsManager._getAllProjectTags = function() {
 
 // Helper function to provide more localized message from plain Java Exception message.
 function getLocalizedErrorMessage(message) {
-    if(message.contains("JsonParseException")){
+    console.log(message);
```
wetneb: `console.log` should be removed.

zengchu2:
Hi, my code has passed the test locally but not the test provided on github. 
Are there any hints or guidence on this ?

thadguidry:
At the bottom of this PR you will see the merge checks passing or failing.  We have builds on Travis CI and Appveyor.  Next to each one you can click on the blue colored "Details" link for the failing checks of the build.  This will open a new browser page that shows the build job details.  Then click on any of the build job links within it, and scroll down the log to see where and why it failed.

wetneb:
@zengchu2 could it be that you still have some modified files that have not been committed? Travis fails because of a build error, so it should be relatively easy to reproduce the same error on your side…

main/src/com/google/refine/importing/DefaultImportingController.java
```diff
@@ -290,7 +293,15 @@ static public void writeErrors(JsonGenerator writer, List<Exception> exceptions)
             e.printStackTrace(new PrintWriter(sw));
             
             writer.writeStartObject();
-            writer.writeStringField("message", e.getLocalizedMessage());
+            String message = e.getLocalizedMessage();
+            
+            if(e.getCause() instanceof JsonParseException) {
```
wetneb: This code is specific to the JSON importer, but it was added to a general purpose importing class. This is a problem: writing importers should not require making changes to generic classes like this.

Instead, we should have a generic way to report parsing exceptions to the user. So I propose we do this:
* Catch the `JsonParseException` in the JSON importer;
* Extract the message as you are doing here;
* Store this message in a new exception, that we could introduce in this PR but could then be reused by other importers (we could call it `ImportingParserException`);
* Then we can still have your `isImportingParserException` logic with `"core-views/check-format"` as message that we concatenate.

main/src/com/google/refine/importers/tree/XmlImportUtilities.java
```diff
@@ -265,13 +268,13 @@ static public void importTreeData(
             }
         } catch (TreeReaderException e) {
             logger.error("Exception from XML parse",e);
-            throw e;
+            throw new ImportingParserException(((JsonParseException)e.getCause()).getOriginalMessage());
```
wetneb: I don't think we can assume that all `TreeReaderException` instances are `JsonParseException`… Especially since `JsonParseException` is not a subclass of `TreeReaderException`! It would be good to have a unit test to show how this works. (This sort of mistake should be caught by any unit test covering these lines)

main/src/com/google/refine/importers/tree/XmlImportUtilities.java
```diff
@@ -265,13 +268,13 @@ static public void importTreeData(
...
         }
     }
 
     /**
      *
-     * @param project
+     * @param pro
```
wetneb: Why?

main/src/com/google/refine/importers/ImportingParserException.java
```diff
@@ -0,0 +1,13 @@
+package com.google.refine.importers;
+
+/**
+ * Exception thrown by importers.  
+ * This exception should contain
```
wetneb: "This exception should contain"… what?

zengchu2:
**My Fix** 
Hi, I managed to reuse the  `TreeReaderException` instead of wrapping the original error message using a new type of exception since the source of error `next()` from `JsonImporter.java` is a Override function and I will try to reuse the original exception here. 
**Souce Of Error** 
The source of casuing the exception is at line `JsonImporter.java` line 349. When calling `parser.getText()` with some invalid text, there will be a `JsonParseException`. 
**Test Case:**
My implemented test `canThrowError` is more of a integrated test, it checks wheter the message of thrown exception contains only the original message (which is being displayed to users in front end)

Issue #2338, opened 2020-02-27
Code style improvements


wetneb:
We have not been enforcing a consistent code style lately. This causes some issues in inconsistent indentation, diverging name conventions, etc.

- [x] decide which style code should follow (probably broken down by languages)
- [x] convert the existing code to match these guidelines
- [x] set up continuous integration to enforce these styles

This would need to be done both for Java and Javascript.

thadguidry:
"I think", we have some prior discussion on mailing lists...or even old issues, that had some of this discussion (Tom and Stefano had preferences)...but it's 2020 now...so best to start fresh and make decisions on current practices. (just thinking about ECMA and all those debates they themselves had back in the day, hahaha)

thadguidry:
Actually, yeah, we had some decisions and some of those landed in the /IDE path

kushthedude:
@wetneb @thadguidry What about implementing ESLint for JS and Checkstyle for Java >

thadguidry:
@kushthedude yeah probably Checkstyle for Java...doesn't SonarQube and IntelliJ both use that as a default now?  Can you check?

kushthedude:
Can't checkstyle also enforce the coding styles such as Linting, White-Spaces and other Google Java Style Standard Rules

kushthedude:
Also, ESLint can do the same for the js code with an option to add tests into the CI.

kushthedude:
@thadguidry Shall I go ahead implementing the same ?

thadguidry:
There's a style guide and then there's enforcing compliance of the style.
It sounds like you want to make the enforcement and IDE setup for developers easier?
Optionally enforcing checks on our CI/CD for pull requests?

You can look to a few Apache projects and GitHub projects to see how they do this (only requires us to supply a few extra files historically, but today, I think Eclipse and IntelliJ ship with a few default styles like Google Java Style ...check into this yourself to see. )  http://help.eclipse.org/2019-12/index.jsp
Here's the official location of Google Style Guides for reference (but tools already grab from here): https://github.com/google/styleguide
Eclipse Help setup: http://help.eclipse.org/2019-12/index.jsp

And here's a snippet of instructions that we could also borrow and put somewhere in our docs, perhaps in the CONTRIBUTOR.md or README.md  (I don't know...check on how Apache projects set this up and give developers the info, I would say, even though we are not an Apache project, historically we have pulled in their best practices)
[...]

PR #2418, opened 2020-03-16
Addition of ESLint

kushthedude:
Refers #2338
- Added npm to the project.
- Initialised ESLint.
- Enforced the Standard JS Style.
- Manually edited the files to enforce the style guide which can be followed by vanilla js and jquery add-on as being used by Open Refine.


fpompermaier:
I didn't test the PR but I can't find where the npm is integrated..have the user to manually run npm install or something? I don't see any modification to any pom.xml..it seems a bit strange to me..but maybe I'm wrong (I'm not an UI expert at all..)

thadguidry:
> I didn't test the PR but I can't find where the npm is integrated..have the user to manually run npm install or something? I don't see any modification to any pom.xml..it seems a bit strange to me..but maybe I'm wrong (I'm not an UI expert at all..)

@fpompermaier If using IntelliJ, you would setup ESLint and configure something like this: https://www.jetbrains.com/help/idea/eslint.html#  Each IDE has a similar kind of configuration, and most perform ESLinting with codestyle verificaton out-of-the-box because they look at the `.eslintrc` and `package.json` files in the parent folder and self setup.

thadguidry:
@kushthedude Can you also do an additional PR for updating our CONTRIBUTING.md file with information about ESLint setup.  Look at how other popular projects mention about IDE setups and code style enforcement settings, and let's add some similar paragraph language to our own CONTRIBUTING.md file to make it easier for developer onboarding.

kushthedude:
> I didn't test the PR but I can't find where the npm is integrated..have the user to manually run npm install or something? I don't see any modification to any pom.xml..it seems a bit strange to me..but maybe I'm wrong (I'm not an UI expert at all..)

Once you have npm already on your computer, running `npm install` will do the work for you.
I will update the contributing.md and readME concerning our code style for the same @thadguidry .

kushthedude:
> * removing included libraries and adding configuration to fetch them via npm. That's a great idea, however it does add a build dependency: npm. At the moment Maven is all that is needed to build the project, so such a change should not be taken lightly. We would need to add functionality to the `refine` and `refine.bat` scripts to download npm on the fly if it is not present, probably. This impacts the integration in IDEs, the release process… Big enough to be a solid PR on its own.

@wetneb That seems nice to me, How about we integrate the refine scripts into `npm start` or we can make use of `makefile`, simply doing `make build` will do everything for end-user from fetching the dependencies to running the project.


> * linting the .js files to a common standard: that's a good thing, however I don't think this should be done for external libraries that are still included in the repository, as it makes it harder to diff with the original or any new versions.

Yeah, I just noticed that while applying the changes I didn't look upon the external-scripts we are using. It is really inappropriate to edit those external-sources as standardisation shouldn't be broken. I will fix it in upcoming commit.

wetneb:
> How about we integrate the refine scripts into `npm start` or we can make use of `makefile`, simply doing `make build` will do everything for end-user from fetching the dependencies to running the project.

I would not rely on a Makefile: that would add yet another dependency to `make`.
Integrating the refine scripts in `npm start` would make it impossible to use the refine scripts to fetch npm in the first place…

thadguidry:
I think just providing instructions to developers (CONTRIBUTING.md) and the 2 ESLint files are enough.  The rest our smart IDE's can handle as long as developers know how to apply the code style checks in their dev environment.

Then for code style enforcement (if we really want to go there), we can talk about that separately on our dev mailing list (that would likely use GitHub Actions with its builtin NPM container and ESLint.  Read more here: https://github.com/marketplace/actions/eslint-action)

kushthedude:
> I think just providing instructions to developers (CONTRIBUTING.md) and the 2 ESLint files are enough. The rest our smart IDE's can handle as long as developers know how to apply the code style checks in their dev environment.
> 
> Then for code style enforcement (if we really want to go there), we can talk about that separately on our dev mailing list (that would likely use GitHub Actions with its builtin NPM container and ESLint. Read more here: https://github.com/marketplace/actions/eslint-action)

we can integrate the lint tests with travis too that won't be a big issue

thadguidry:
Sure, Travis, whatever, anywhere you want to do the automation.

fpompermaier:
> > How about we integrate the refine scripts into `npm start` or we can make use of `makefile`, simply doing `make build` will do everything for end-user from fetching the dependencies to running the project.
> 
> I would not rely on a Makefile: that would add yet another dependency to `make`.
> Integrating the refine scripts in `npm start` would make it impossible to use the refine scripts to fetch npm in the first place…

Why not just using maven webjars if you want abetter dependancy management of js jars? https://www.baeldung.com/maven-webjars

wetneb:
Sounds like a good fit given OpenRefine's existing architecture.

thadguidry:
I always forget about those damn things.

kushthedude:
> Why not just using maven webjars if you want abetter dependancy management of js jars? https://www.baeldung.com/maven-webjars

No web jars won't be useful in installing the npm. Webjars specifically installs the dependencies into the web jars folder when we need to use them we can call them using the external src and specifying the path. But that's not how npm and the node_modules works going into any of the basic web projects needs you to have npm installed on your pc or rather you can fetch with any external sh script or makefile rule. Once you got npm, the node_modules will be fetched into the node_modules folder which is already specified the path by the node engine to use from. Using the web jars would ease the dependencies but the dependencies management will get more worse.

wetneb:
> No web jars won't be useful in installing the npm.

If we use web jars, we would not use npm at all: we would use Maven to fetch frontend dependencies directly.

kushthedude:
> > No web jars won't be useful in installing the npm.
> 
> If we use web jars, we would not use npm at all: we would use Maven to fetch frontend dependencies directly.

Considering this webjar

<dependencies>
    <dependency>
        <groupId>org.webjars</groupId>
        <artifactId>bootstrap</artifactId>
        <version>3.1.0</version>
    </dependency>
</dependencies>
Now whenever we need to use the bootstrap in any of our file we will have to use this line above

<link rel='stylesheet' href='webjars/bootstrap/3.1.0/css/bootstrap.min.css'>
but rather if we use npm engine, Once the bootstrap is specified in our package.lock, it is available to our project globally without the need of specifying it in every file.

Also again, not all peer dependencies for ES-Lint are available in web jars like we need eslint-plugin-node, there is not web jar for the following. Now considering the eslint-plugin-import, the web jar for the following has not been updated since 2017.

fpompermaier:
I think that the problem here is that with this pull request you're assuming that refine is using nmp that is not at all at the moment. Of couse I'm absolutely in favour of moving to npm and to a better UI framework like Vue/React/Angular but this is something that should be carefully designed and accepted by the community. As you said, webjars has the inconvinient that you have to import the js in every page, but this is how Refine works right now so it's not a problem for current committers. Actually, wrt the current version of Refine is still a small improvement!

kushthedude:
> I think that the problem here is that with this pull request you're assuming that refine is using nmp that is not at all at the moment.

I am well aware that we are not using npm, hence I initialised the package.json where there are some details still needed to be incomplete. Webjars are mainly packaged for java application assuming our complete server runs on Java and the Client-Side is based on vanilla js with jquery. With moving to ESLint we are making our first step to move towards Ecmascript 2018. Also not mention that ES-Lint modules are very old and incomplete in web jars. Currently, there is only a jar available for ESLint, ESLint-Plugin-import. To enforce the rules and styling in our CI as well as on the editor config we will need some extra peer dependencies like Plugin-Node to successfully enforce the code-style in javascript.
Introducing the npm and node engine in our web UI will be the first step of Open Refine towards moving to a Progressive Application.

Also earlier as @thadguidry mentioned we are actually looking towards moving to Vue.JS, now to use vue.js we can use the CDN or we can manage and install the vue-cli and vue using the node engine.

wetneb:
We are conflating two discussions here, so I propose we close this PR and move this to two separate issues:
* #2338 for code style
* #2423 for dependency management.
Any PR should address only one of these as both are far-reaching topics on their own.

kushthedude:
@thadguidry @wetneb I think it would be better if I start a new topic in
dev-mailing list regarding the npm vs webjar.

wetneb:
Sure, that's also an appropriate place to discuss things.

Issue #1869, opened 2018-11-28
CSV Export, always quote text (always include text qualifier, quoteAll option)


lucacanella:
There should be a way to force text quoting in the export for every field.
Example given:
Normal CSV export behaviour
```
a,b,cc
1,2,"1,2,3"
```
With quote all option:
```
"a","b","cc"
"1","2","1,2,3"
```

I looked into the source code and found that there already is a "quoteAll" parameter in the CSV Exporter: https://github.com/OpenRefine/OpenRefine/blob/9f7d5b878628c40d68a445b751fd50140d922e7d/main/src/com/google/refine/exporters/CsvExporter.java#L84 
A flag in the frontend could activate this parameter. This flag should be included at least in the Custom Tabular Exporter interface.

Just as demostration of how it could work, I've done a little hacking into the Chrome Console (this assumes that you can inject code into the console and can debug the code by stopping the execution at a give breakpoint):
1. Launch OpenRefine, open a project and open the browser console.
2. From the browser console look for the "bundle.js" file and put a breakpoint just before the line with `form.submit();`, in the `ExporterManager.handlers.exportRows` function.
3. From the OpenRefine user interface, select "Export > Comma-separated value"; the browser should stop execution just at `form.submit()`.
4. Inject the following code, so that the "quoteAll" parameter is set in the request, then press "continue" to start execution.
5. The CSV fields in the result should all be quoted.

```javascript
$("<input />")
  .attr("name", "options")
  .attr("value", "{\"quoteAll\": true}")
  .appendTo(form);
```

bizKOr:
Perhaps a good idea to add such option when exporting a CSV.

bizKOr:
@lucacanella is possible to show where you put the breakpoint and the code. And I only find "project-bundle.js".

wetneb:
`project-bundle.js` is what @lucacanella meant, this is where all the JS code for the project UI is distributed.

lucacanella:
Sorry, I've been missing the notifications.
Back to the issue: **Yes**, I put the breakpoint on `http://127.0.0.1:3333/project-bundle.js`, row 70710. 
Here is where i put the breakpoint:

```js
ExporterManager.handlers.exportRows = function(format, ext) {
  /* BREAKPOINT */ 
  var form = ExporterManager.prepareExportRowsForm(format, true, ext);
  $('<input />')
  .attr("name", "contentType")
  .attr("value", "application/x-unknown") // force download
[...]

PR #2427, opened 2020-03-17
Custom Tabular Exporter: add option to quote all strings #1869

Abbe98:


Abbe98:
@wetneb sure, will however do that on a separate branch as this as you mentions does not modify the backend.

Picking issues that affect me as an user so tomorrow I will aim to do a WD import with OpenRefine to then being able to take on additional issues.

Issue #1164, opened 2016-12-09
Tab character in a value in a JSON file results in a truncated data import


ostephens:
When importing JSON to OpenRefine through 'Create Project', if there is a tab inside a value in the JSON file, the import succeeds but only imports data up to the tab character. e.g. JSON:

`{
  "rows": [
    {
      "Content": "A"
    },
    {
      "Content": "	B"
    },
    {
      "Content": "C"
    }
  ]
}`
(tab character is before the 'B')

Since tab is an illegal character in this position (should be escaped as \t) it makes sense for the import to fail - but failing silently is a problem. 

Not sure what the best solution is - three possibilities:
1. Import should fail completely with an error
2. Import should succeed with the data truncated as currently but with feedback to the user that the truncation has occurred and why
3. Import should succeed with full data, converting illegal character to escaped version

thadguidry:
I think 3 would be most appropriate given OpenRefine is a data cleansing tool.  And we should try to cleanse a bit of that.  Folks expect it not to fail for a String value and we can just escape for those cases.

thadguidry:
Looks like we might be able to just expand on the CASE logic for VALUE_EMBEDDED_OBJECT
https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/importers/JsonImporter.java#L125
using JsonToken
http://fasterxml.github.io/jackson-core/javadoc/2.2.0/index.html?com/fasterxml/jackson/core/JsonParseException.html
or perhaps using JsonParser.Feature methods might do it
http://fasterxml.github.io/jackson-core/javadoc/2.2.0/index.html?com/fasterxml/jackson/core/JsonParseException.html

zengchu2:
Hi, I am now working with this issue and had make some progress. 

Just double check if I understand the expected effect correct:  file with `{ "rows": [ { "Content": "A" }, { "Content": " B" }, { " Content": "C" } ] }`  should be able to get loaded and there are two types of `Content`, one with a tab and one do not. And there is a tab before value `B` .

wetneb:
I am not sure what you mean by "two types of `Content`": I would expect `A`, ` B` and `C` to end up in the same column.

zengchu2:
I see, so unquoted control chars (like \t, \n) should be ignored during the parsing right?

wetneb:
I would not ignore these unescaped control characters - see the 3 options mentioned by Owen above.

tfmorris:
[...]

PR #2431, opened 2020-03-17
Convert illegal characters into legal ones.

zengchu2:
Attempt to fix issue #1164 .
**Source of error:**
The JacksonParser does not accept unquoted control chars (ASCII characters with value less than 32) unless udpate it's features. 
Parsing unquoted chontrol chars will cause an  `Illegal Unquoted character` exception. 

**Tests**:
1.  Implement new test cases to ensure parser works as expected.
2.  Using the attached test json file, illegal character (the tab) is converted into legal ones as shown in the preview.

testJson file: 
[testJson.txt](https://github.com/OpenRefine/OpenRefine/files/4345718/testJson.txt)

### DIFF
--- a/main/tests/server/src/com/google/refine/importers/JsonImporterTests.java
+++ b/main/tests/server/src/com/google/refine/importers/JsonImporterTests.java
@@ -380,6 +380,52 @@ public class JsonImporterTests extends ImporterTest {
     }
 
     @Test
+    public void testCanParseTab() throws Exception {
+        // Use un-escaped tab here. 
+        String sampleJson = "{\"field\":\"\tvalue\"}";
+        String sampleJson2 = "{\"\tfield\":{}}";
+        
+        
+        JSONTreeReader parser = new JSONTreeReader(new ByteArrayInputStream(sampleJson.getBytes("UTF-8")));
+        Token token = Token.Ignorable;
+        int i = 0;
+        try {
+            while (token != null) {
+                token = parser.next();
+                if (token == null) {
+                    break;
+                }
+                i++;
+                if (i == 3) {
+                    Assert.assertEquals(Token.Value, token);
+                    Assert.assertEquals("field", parser.getFieldName());
+                }
+            }
+        } catch (Exception e) {
+            Assert.fail();
+        }
+
+        parser = new JSONTreeReader(new ByteArrayInputStream(sampleJson2.getBytes("UTF-8")));
+        token = Token.Ignorable;
+        i = 0;
+        try{
+            while(token != null){
+                token = parser.next();
+                if(token == null) {
+                    break;
+                }
+                i++;
+                if(i == 3){
+                    Assert.assertEquals(Token.StartEntity, token);
+                    Assert.assertEquals(parser.getFieldName(), "\tfield");
+                }
+            }
+        }catch(Exception e){
+            Assert.fail();
+        }
+    }
+    
+    @Test
     public void testJsonDatatypes(){
         RunTest(getSampleWithDataTypes());
 
--- a/main/tests/server/src/com/google/refine/importers/JsonImporterTests.java
+++ b/main/tests/server/src/com/google/refine/importers/JsonImporterTests.java
@@ -38,11 +38,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
+import java.lang.reflect.Method;
 import java.util.LinkedList;
 import java.util.List;
 
 import org.slf4j.LoggerFactory;
 import org.testng.Assert;
+import org.testng.ITestResult;
 import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.BeforeTest;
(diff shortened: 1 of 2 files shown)


thadguidry:
@zengchu2 This is great, however 2 things:
1. We didn't inform the user of the problem and instead silently converted some data? OpenRefine, as a powertool, we try to avoid being silent.
2. We didn't give the power to the user to control how they want to handle this.  (Like exposing an option in the UI).

Example message dialog popup:
"Could not Preview the file.  Detected unquoted control chars while attempting to parse.  You can try checking the option `While parsing, ignore unquoted control chars` to have OpenRefine allow parsing with unquoted control chars."

By the way, does this `JsonParser.Feature.ALLOW_UNQUOTED_CONTROL_CHARS` actually do a conversion of the data somehow (like changing Control Chars to something displayable like a whitespace or something else)?  In OpenRefine, whenever we do ANY data conversion of original files, we will always inform the user of that and never are silent.

main/tests/server/src/com/google/refine/importers/JsonImporterTests.java
```diff
@@ -296,6 +296,52 @@ public void EnsureJSONParserHandlesgetLocalNameCorrectly() throws Exception{
...
+        try {
+            while (token != null) {
+                token = parser.next();
+                if (token == null) {
+                    break;
+                }
+                i++;
+                if (i == 3) {
+                    Assert.assertEquals(Token.Value, token);
+                    Assert.assertEquals("field", parser.getFieldName());
```
tfmorris, line 318: I don't see anything that tests for the value `"\tvalue"` which what the original bug was about.

tfmorris:
@thadguidry I'm not sure an importer option to control this is necessary. The file is illegal JSON, so isn't parseable at all unless we use a permissive parser. The control characters will get escape correctly on output, so it's totally reversible, but that seems fine to me.

thadguidry:
@tfmorris I was just mentioning the importer option...not standing upon its heels.  If you re-read my point 1... we should not be silent to the user when things FAIL behind the scenes. I.E Illegal JSON should be thrown up to the user and stated as such.  And that was my point, that OpenRefine should be less silent about that.  Agree?

wetneb:
As a user I would find it frustrating to see an error by default and have to discover that there is a checkbox somewhere to disable this error.

The only reason to make this user-configurable would be that the option could corrupt the parsing of files which would be valid for another importer, but I cannot think of how this could happen with this particular change.

thadguidry:
I need to be more explicit in my writing on these comments.  I guess I expect everyone to understand me, but talking and reading are 2 modes of input that don't always agree, and I'm better at talking as everyone knows, often too much. :-)

So... I don't think we need this change.  I mean an option that converts invalid JSON formats.  I prefer to warn about the invalid formats, so that the user can then clean it up...perhaps even using OpenRefine itself in another mode like Line-based to perform the cleanup.  I have done that often enough.

We clean messy data.  We DO NOT clean messy formats?  I guess that's what we should adhere to since we are dependent on a large ecosystem of helper libraries for parsing and validating before importing the data.

So 👎 for introducing this change on this PR.  And instead we improve the error messaging to users on the JSON importer (and maybe others) trying to use a particular importer that cannot parse the data provided by the user and generously warn the user of that (and not silently fail or get stuck).

In general, I take back [my comment](https://github.com/OpenRefine/OpenRefine/issues/1164#issuecomment-266079114) that we should offer an option to cleanse unescaped values to clean messy JSON on the issue #1164 , since I was not aware that this was in fact an invalid or unknown format parsing error for JSON (Thanks to @tfmorris for helping me understand now on this) 

All in favor?

wetneb:
I don't understand your comment above. Could you try to describe a concrete situation, from a user point of view, where it is undesirable that OpenRefine silently accepts to parse unescaped tabs?

thadguidry:
@wetneb Sure.  @ostephens already did.  It truncated the import, silently, without warning him.  Just re-read his exact original issue comment:  https://github.com/OpenRefine/OpenRefine/issues/1164#issue-194632653

thadguidry:
Oh if it accepts parsing the unescaped tabs.... hmm... if it parses, GREAT, and we can then fully import the data and not truncate, GREAT.  Then that would be fine.  The parsing, then silent failure of that parsing, is the issue we are all familiar with and want to fix.

wetneb:
Finally we got there, yipee!

thadguidry:
HOWEVER:  I hope everyone understands that I always worry about this:

When OpenRefine finds something about the data during import.... converts it WITHOUT telling the user...and then 1 week later the user finds out about the conversion where there was data truncation or important data removal, that he/she didn't actually know about... we then look like the crappiest program out there.

tfmorris:
@zengchu2 I took the liberty of updating the test for you and pushing the update to your branch. LGTM. Thank you again!

@wetneb Can you please do a quick review? I made the tests less chatty and want to make sure that the logging that I removed isn't needed for anything. My first attempt at a fix a few weeks ago passed tests locally, but not on Travis, apparently because of an order dependency in the tests. That's fixed now and it passes on Travis as well as locally.

PR #2440, opened 2020-03-19
Show correct data type in data table cell editor

### LINKED ISSUE #2424: Data Type setted not remembered by application
**Describe the bug**

Change the data type of any cell, and see again the type of data is not remembered by the application.

**To Reproduce**

1. Go to 'Edit Cell'
2. Click on 'Boolean Datatype'
3. Set the data to `True/False`.
4. Again click on edit cell and see data type as text rather than boolean.

**Expected behavior**

Data-Type preference should be remembered

**Screenshots**

Screenshot 2020-03-17 at 4 03 45 PM

joanneong:
Fixes #2424 .

The application does not remember the data type of a cell as a cell only stores information related to the cell value. Hence, to show the correct data type on the front end, we can just use `typeof` to get the data type of the cell before rendering the HTML. This way of using `typeof` is also consistent with the rest of the application (e.g. the rest of the code in `cell-ui.js` etc).

This works with different languages as well since the `select` values are in English regardless how OpenRefine is localised.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
@@ -543,7 +543,10 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
   elmts.or_views_ctrlEnter.html($.i18n('core-views/ctrl-enter'));
   elmts.cancelButton.html($.i18n('core-buttons/cancel'));
   elmts.or_views_esc.html($.i18n('core-buttons/esc'));
-  
+
+  var cellDataType = typeof originalContent === "string" ? "text" : typeof originalContent;
+  elmts.typeSelect.val(cellDataType);
+
   MenuSystem.showMenu(menu, function(){});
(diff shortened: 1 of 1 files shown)


joanneong:
Sorry for the delay (been rushing my thesis for the past month)! My bad for missing out the date - I've fixed it and checked all the cell types this time round. 

As for extracting it into a separate type checking function, I did consider it as well. However, I am not sure where the best place to do that would be. Looking at the code base, it seems like serializing the cell data type in `Cell.java` when a cell is retrieved as a JSON object is the 'cleanest' way of doing things. But I am not sure if doing this would affect other parts of the application, and whether this is indeed the best way to do things. In the meantime, I have added an additional line to check for the date type using existing cell properties instead.

thadguidry:
@joanneong Thanks! And best of luck on your thesis!

wetneb:
Hmm, actually this breaks the editing of null cells, I should have checked it more thoroughly. The fix is on me.

PR #2446, opened 2020-03-20
Recently selected choice will appear as default choice

### LINKED ISSUE #2316: "Match other cells with same content" keeps appearing as the default choice
**Describe the bug**

**Expected behavior**

**Screenshots**
Bez názvu

**Desktop**
 - OS: [Windows 10]
 - OpenRefine 3.3 Beta

khyati-agarwalss:
Fixes #2316

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
@@ -360,11 +360,13 @@ DataTableCellUI.prototype._searchForMatch = function(suggestOptions) {
       if (elmts.radioSimilar[0].checked) {
         params.similarValue = self._cell.v;
         params.columnName = Refine.cellIndexToColumn(self._cellIndex).name;
+        elmts.radioSimilar[0].setAttribute(checked, "true");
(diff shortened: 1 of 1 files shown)


khyati-agarwalss:
> Thanks! However I don't understand how this is supposed to work - where is the recently selected choice stored? Also, I tried running this and now the "Match" button does not do anything anymore.

@wetneb I thought it worked, but you are right, the Match button isn't working. I tried other methods of doing this, but nothing is leading to what is expected.

wetneb:
The 2 options are radio buttons, so one of them must be selected by default, I would expect. You could replace both radio buttons by a single checkbox, but it would probably not be as clear as the current UI.

khyati-agarwalss:
> The 2 options are radio buttons, so one of them must be selected by default, I would expect. You could replace both radio buttons by a single checkbox, but it would probably not be as clear as the current UI.

True. Also, changing that might break other functions.

wetneb:
Closing per https://github.com/OpenRefine/OpenRefine/issues/2316#issuecomment-606469655

PR #2450, opened 2020-03-21
Create manual_devfile.yaml

thadguidry:
Still testing this on Eclipse SaaS workspaces to find out issues and further customization's.

### DIFF
--- a/IDEs/eclipse_che/devfile.yaml
+++ b/IDEs/eclipse_che/devfile.yaml
@@ -1,27 +1,25 @@
 apiVersion: 1.0.0
-
 metadata:
-    name: OpenRefineChe-
-
+  name: OpenRefine_Che
 projects:
-  - name: OpenRefine_Che
+  - name: OpenRefine
     source:
-      location: 'https://github.com/OpenRefine/OpenRefine'
+      location: 'https://github.com/OpenRefine/OpenRefine.git'
       type: github
       branch: master
-
-# A single development container having a small OS, Maven, and JDK
+attributes:
+  persistVolumes: 'false'
 components:
-  - alias: maven
-    id: redhat/java11/latest
+  - id: redhat/java11/latest
     type: chePlugin
-    image: 'quay.io/eclipse/che-java11-maven:nightly'
-    mountSources: true
+  - mountSources: true
     memoryLimit: 1536M
     type: dockerimage
     volumes:
       - name: m2
         containerPath: /home/user/.m2
+    alias: maven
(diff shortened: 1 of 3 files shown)


IDEs/eclipse_che/devfile.yaml
```diff
@@ -0,0 +1,60 @@
...
+        name: JAVA_TOOL_OPTIONS
+  - id: ms-vscode/vscode-github-pullrequest/latest
+    type: chePlugin
+
+commands:
+  - name: maven build
+    actions:
+      - workdir: '${CHE_PROJECTS_ROOT}/OpenRefine'
+        type: exec
+        command: mvn clean install
```
thadguidry: @wetneb what would be the real commands we want to expose for `mvn` here in the `devfile.yaml`?  First step is just to ensure that the dev environment can build/test inside Che, then I can work on exposing the UI endpoint as well, which I kinda started in the `manual_devfile.yaml`.  I have 3 files here so that I don't lose track of testing (I hit my quota today on Eclipse SaaS with all the testing! lolol)

wetneb: I don't know, I haven't looked into Eclipse Che configuration - I am only aware of the existence of Eclipse Che, but have never tried to use it myself.

thadguidry: Its OK, I found my answer... at the bottom of the `refine.bat` more or less.  I'll see if I can make progress on this today or tomorrow.
```
:gotMvnHome
set MVN_ACTION=""%ACTION%""
if ""%ACTION%"" == ""build"" set MVN_ACTION=compile test-compile dependency:build-classpath
if ""%ACTION%"" == ""test"" set MVN_ACTION=test dependency:build-classpath
if ""%ACTION%"" == ""server_test"" set MVN_ACTION=test -f main
if ""%ACTION%"" == ""extensions_test"" set MVN_ACTION=test -f extensions
call "%MAVEN_HOME%\bin\mvn.cmd" process-resources
call "%MAVEN_HOME%\bin\mvn.cmd" %MVN_ACTION%
goto end
```

thadguidry:
So, the more I thought about it.  And what @tfmorris said on the dev mailing list about stretching our support on various things.  We probably **do not** want this config in our repo directly.  It will add overhead to our limited bandwidth for supporting users.
I think what is better is that this config can live in someone else's repo or a blog post if needed.

So closing this as "not needed".
And since [GitHub Codespaces](https://help.github.com/en/github/developing-online-with-codespaces/about-codespaces) is coming along quite nicely actually in Beta.

Issue #2234, opened 2019-11-26
Extending the cross() function - please give advice


felixlohmeier:
Two years ago a colleague and I created a feature request (#1289) and a pull request (#1294) to extend the cross function to support multi-valued cells which was followed by a long discussion. The original cross function expects normalized data (one foreign key per cell in base column). If you have multiple key values in one cell you need to split them first in multiple rows before you apply cross (and join results afterwards). This can be quite "expensive" if you work with bigger datasets.

Finally, we all agreed upon keeping the cross function unchanged and aim for general performance improvements (https://github.com/OpenRefine/OpenRefine/projects/1#card-5380183). In the meanwhile I created a fork including the pull request to solve my use case. It is based on the OpenRefine development version of October 10, 2017 (something between OpenRefine 2.7 and 2.8):
* Code: https://github.com/opencultureconsulting/openrefine/
* Documentation: https://github.com/opencultureconsulting/openrefine/wiki

We use this fork within a data pipeline for the German Literature Archive to extract, transform, enrich and load data from an integrated library system into a search index for a new catalog, which will be published in a few months. Slides from a library conference are available here: http://doi.org/10.5281/zenodo.2678113 (cross on p. 16). The amount of data is big enough that it just don't work (within a time frame of 24 hours / daily updates of the catalog) without the extended cross (e.g. split, cross and join separately takes too much time, see some stats here https://github.com/OpenRefine/OpenRefine/issues/1289#issuecomment-340018966).

The project will end in June 2020 and I would like to find a sustainable solution by then. Currently the staff at the German Literature Archive is stuck at OpenRefine ~2.7 and merging the pull request #1294 with the current master got a bit more difficult because of other changes to Cross.java and the InterProjectModel.java.

Regarding the recent OpenRefine funding and plans to work on the core data model in 2020 what approach would you recommend?
- merge fork with current master (not very easy to maintain)
- just wait until the end of 2020 (because general performance improvements with Apache Beam will solve the issue)
- create an extension with an additional GREL function, e.g. crossSplit (not sure if this is possible because of the changes to InterProjectModel.java; will be broken in 2020 due to planned changes to the core?)

OpenRefine team, could you please give an advice here?

wetneb:
Thanks for opening this! I would like to understand better why @ostephens' GREL-based solution is so slow - I haven't looked in detail but I think there should be ways to make `cross` more efficient in that respect.

ostephens:
[...]

Issue #1204, opened 2017-06-23
Extend cross() function to take either a cell or a value


felixlohmeier:
On Saturday, 13 October 2012 01:12:23 UTC+2, Tom Morris wrote:
> Currently the cross() function is one of the few (only) that takes a cell as its first argument instead of value.
>
> Can anyone think of a good reason that we shouldn't extend it to take either a cell or a value?
> 
> Tom

On Sunday, 14 October 2012 23:13:57 UTC+2, David Huynh wrote:
> If I remember correctly--and I should have documented this, the reasons for taking a cell rather than an arbitrary expression were to make the caching logic easy, and also to avoid any expression that yields something other than a single literal value (e.g., an array of value, a "cell" or "row" or "record" object).
> 
> We should definitely make it consistent with the other functions if we can work out the caching logic (which has been buggy).
> 
> David

Current state:
* Code: [OpenRefine/main/src/com/google/refine/expr/functions/Cross.java](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/expr/functions/Cross.java)
* User documentation: https://github.com/OpenRefine/OpenRefine/wiki/GREL-Other-Functions#crosscell-c-string-projectname-string-columnname
* User documentation "recipes": https://github.com/OpenRefine/OpenRefine/wiki/Recipes#merging-all-or-more-than-two-columns-in-a-project

jackyq2015:
Pushed the change f03be76, it covers:

- Take the cell value instead of the cell as the first argument
- Added the necessary action to flush the stale cache. They were missed before.

tfmorris:
@felixlohmeier Thanks for the summary. Do you have links the the original sources of the quotes? Was it one of the mailing lists or another issue?

@jackyq2015 The referenced commit appears to be to master, not a branch. Does it resolve this issue? Who reviewed it? (Apparently Thad has reviewed it post-facto based on his typo comment)

tfmorris:
"Thad" being @thadguidry, of course (so he gets notified)

felixlohmeier:
@tfmorris quotes from https://groups.google.com/forum/#!topic/openrefine/7ESn_Xh0tRU

wetneb:
Nice, that was quick! If you have time to write one or two unit tests that would be fantastic too!

felixlohmeier:
[...]

PR #2456, opened 2020-03-22
Make cross() function work for all columns

### LINKED ISSUE #1950: Cross() function not working if applied to a column different from the column used for matching
**Describe the bug**
Cross() function not working if applied to a column different from the column used for matching.
Not sure it is fully reproducible, but happended to me twice, with big OR projects (a lot of transformations, including moving columns, before applying the cross function)

I cannot join the files because they weigh 200 Mo...

**To Reproduce**

**Current Results**

The normal behavior (matching 2 project, based on "IDAloes", the 1st column of each project, and applying the formula to the 1st column)
image

The error (matching the same projects on the same column, but applying the formula to the 2d column)
image

No warning in console

**Expected behavior**

**Screenshots**

 - OS: Mac and PC, same behavior
 - Browser Version: FF
 - JRE or JDK Version:[output of "java -version" e.g. JRE 1.8.0_181]

 - Version : OpenRefine 3.1

**Datasets**
If you are allowed and are OK with making your data public, it would be awesome if you can include or attach the data causing the issue or a URL pointing to where the data is.
If you are concerned about keeping your data private, ping us on our [mailing list](https://groups.google.com/forum/#!forum/openrefine)

**Additional context**

afkbrb:
Fix #1950 

Now the cross() function works for all columns.


afkbrb:
Yeah, you're right. 

I just studied the join mechanism used currently. Another problem is that whether we update the source project or the target project, the join cache will be removed.  If we use the new mechanism you mentioned, we'll only deal with the target project, so the join cache is likely to be removed, and accordingly it may be more efficient actually. 

I'd be glad to make the change if you agree.

wetneb:
That would be fantastic!

It would be worth making sure `cross` can still take Cells as first argument, to maintain backwards compatibility.

thadguidry:
> That being said, I still don't see why the cross function should necessarily take a Cell as first argument. Why can't we just have a function taking:
> 
>     * a value
> 
>     * a project name
> 
>     * a column where to look for the value in the project
> 

Questions:
1. "a project name" is ANY project (current or another)?
2. a column where to look for the value in ANY project (current or another)?

wetneb:
I guess it's a violation of the row-wise computation of expression values if you use `cross` on the same project, so we could potentially want to forbid that (I haven't checked if it is currently allowed).
But there are other GREL functions which introduce dependencies outside the row/record context (such as `facetCount`) so I am not sure it's worth forbidding.

thadguidry:
I don't think we want to forbid it allowing where to lookup.  I have always wanted it to work more a bit more similarly to VLOOKUP() in Excel and LibreOffice.  David said at the time there was some modeling limit or GREL snafu that limited our ability when he first hacked on it. Can't recall but I remember him mentioning it to me.  Dunno, would have to ask him.

I think its also important that we document the differences in handling just for posterity:

VLOOKUP() | cross()
------------|-----------
value to lookup | value to lookup
lookup range is same workbook or another, containing full table or partial table range Ex. `VLOOKUP(D8,'<filepath>'#$<sheetname>.$B$1:$I$65255,8,0` | lookup range is full table in another project
column index | column name
exact or similar match | exact match (hard coded)

Cosmetically, do we agree the above is our current functioning of our `cross()` ?

antoine2711:
> * "a project name" is ANY project (current or another)?
> * a column where to look for the value in ANY project (current or another)?

Yes, I think it important this it keeps this functionality, it now works that way (you can self-join your project).
And of course, the column parameter can be any column, as long as it's in the project referred by the « project's name » ;-)

Regards, Antoine

antoine2711:
> Another problem is that whether we update the source project or the target project, the join cache will be removed. If we use the new mechanism you mentioned, we'll only deal with the target project, so the join cache is less likely to be removed, and accordingly it may be more efficient actually.

If we use the cross() function on the same project, then, from your explanation, I understand that you will have to build the local « join cache » on the spot? But you could do that if it's a self-join, and not if it's an external project, so the time price is not always paid?

Regards, A.

antoine2711:
> if you use `cross` on the same project, so we could potentially want to forbid that (I haven't checked if it is currently allowed).

Yes, @wetneb, you can do that right now. See this self-join:
image

afkbrb:
> If we use the cross() function on the same project, then, from your explanation, I understand that you will have to build the local « join cache » on the spot? But you could do that if it's a self-join, and not if it's an external project, so the time price is not always paid?

@antoine2711 , sorry I didn't make it clear. I am talking about the current join mechanism, not the one I am about to implement (which is not join-based). The `join cache` can be used between any two projects, including the same project. 

If it's a self-join, the efficiency of the two mechanism won't differ too much, because only one project is involved in this situation.  So, yes, the time price is not always paid.

But I assume that the cross() function is usually used between two different projects (the source project and the target project). In that case, if we change the source project, we'll need to remove the cache in the current implementation and **won't** need to do it in the new implementation. So in the new implementation, the cache will be more likely to stay valid and less likely to be computed again. That's why I think the new implementation will be more efficient.

afkbrb:
Done!

afkbrb:
@wetneb Thanks for the approval! 

I just reformated  the code, since most files in the project are using 4 spaces instead of a tab.

antoine2711:
Nice work! Well done @afkbrb.

Little question for you: is your new cross() function will work on other types of data than strings? If my key column is composed of numbers, will your function work? Because right now, if it's not text, it won't work…

Regards, Antoine

afkbrb:
@antoine2711 The behavior is almost the same as before, it's an implementation change, not an API change. It's possible to support number, but the previous implementation explicitly requires the first argument to be String or Cell, not sure why we don't support number in that implementation. Maybe we should ask @wetneb .

wetneb:
I think it'd be nice to support numbers too (especially since the underlying index uses `Object` to store values already), but I agree it's a different change, so we can do it in another PR.

thadguidry:
Please support numbers also.  Many scientific datasets have a KEY field that is composed of Long numbers.  Great job thus far @afkbrb !

afkbrb:
Great! Now this PR is merged and I can work on supporting number then.

antoine2711:
@afkbrb: good job. I'll go test that right now.

Regards, Antoine

antoine2711:
@afkbrb: do you thing you could improve the cross function by making the second parameter optional (i.e. if empty…)? 

That way, it would represent the current project, and it would make the code more readable. One other added benefit would be that the same transformation could would on similarly based project without change, and would still work in a same project that would have changed name.

Regards, Antoine

Regards, Antoine

afkbrb:
@antoine2711 Good idea! It's not difficult to do it and it's helpful for reusing. I think you could open a new issue for it since this PR has alreay been closed.

antoine2711:
> @antoine2711 Good idea! It's not difficult to do it and it's helpful for reusing. I think you could open a new issue for it since this PR has already been closed.

I'll get to work for the PR…

felixlohmeier:
@afkbrb and all, I just want to say a big thank you for this new feature! cross supporting values (instead of only cells) is what I dreamed of since 2017: #1204, #1289, #2234 . 

I have only recently found a [solution for my use case with Jython](https://gist.github.com/felixlohmeier/a5a893190e4aa8e26091664908d04e20). But it was quite a hack. It's great that with OpenRefine 3.4 it now works in GREL natively.

felixlohmeier:
@wetneb you wrote on https://github.com/OpenRefine/OpenRefine/wiki/Changes-for-3.4:
> The cross function is no longer restricted to source values from the column where it is invoked

I didn't realize it at first. Should we highlight the new options, e.g. "The cross function now supports any value for input (instead of just cells) and is no longer restricted to the column where it is invoked"?

wetneb:
@felixlohmeier Yes feel free to tweak the changelog!

antoine2711:
> @wetneb you wrote on https://github.com/OpenRefine/OpenRefine/wiki/Changes-for-3.4:
> 
> > The cross function is no longer restricted to source values from the column where it is invoked
> 
> I didn't realize it at first. Should we highlight the new options.

Yes, the overhaul of the cross function is major, in v3.4.

But in v3.5 (Issue « Enhance cross() with empty "project name" calling parameter to represent the current project and empty column name to represent the index #2504 »), you can even do:
`cross(index+1)[0].cells.ColName.value` to get the value of the next row of the cell « ColName ».

It's because, this call is like `cross(index+1, "ProjectName", "")[0].cells.ColName.value`where an empty column name actually means « using the internal index ».

If you think about it, it's pretty cool. The index can now be a column for the `cross()` function.

How do you like that? ;-)

Regards,
   Antoine

PR #2458, opened 2020-03-22
Added google sheet feature and moved features to the export drop-down

### LINKED ISSUE #2453: Have separate menu items for « Export to file… » and « Export to Google Drive… »
Now, if you want to export your project (to file), you go in the menu, and you have to answer a dialog where you choose either « to file » or « to Google Drive ».

Also, using the term « Export to local » is vague, « Export to file » (or even « Export to OpenRefine ZIP » would be better tho the latter is a bit too verbose… ;-)

It would be much better if there where 2 items in the menu, named « Export to file… » and « Export to Google Drive… ». It would save clicks for exporting to file, which is the operation most of our user do, and it would be clearer in the interface that OpenRefine can export the whole project to Google Drive. 

Doing the same as now. But if we want to add interface elements for Export to gDrive, it will clutter the interface.

**Additional context**
image

AghaSaad04:

### FIX #2453  

I have added the feature of exporting the OpenRefine project to google sheets, and have moved the export features out of dialogs and placed into the export menu.

image


main/webapp/modules/core/langs/translation-en.json
```diff
@@ -343,12 +344,12 @@
...
-    "core-project/html-table": "HTML table",
-    "core-project/excel": "Excel (.xls)",
-    "core-project/excel-xml": "Excel 2007+ (.xlsx)",
-    "core-project/odf": "ODF spreadsheet",
+    "core-project/tab-value": "Tab-separated value file",
+    "core-project/comma-sep": "Comma-separated value file",
+    "core-project/html-table": "HTML table file",
+    "core-project/excel": "Excel (.xls) file",
+    "core-project/excel-xml": "Excel 2007+ (.xlsx) file",
+    "core-project/odf": "ODF spreadsheet file",
```
wetneb: Is it really worth adding "file" at the end of all these labels? For instance, "ODF spreadsheet file" sounds quite redundant for me. I would keep these labels as they were.

main/webapp/modules/core/scripts/project/exporters.js
```diff
@@ -150,90 +160,95 @@ ExporterManager.prepareExportRowsForm = function(format, includeEngine, ext) {
...
+
+  document.body.appendChild(form);
+
+  window.open(" ", "refine-export");
+  form.submit();
+
+  document.body.removeChild(form);
+};
+
+ExporterManager.handlers.exportProjectToGoogleDrive = function () {
```
wetneb: Menu items related to Google services should be added in the gdata extension. It is possible to add menu items to common menus in extensions (see for instance how it is done in the Wikidata extension).

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -284,8 +284,9 @@
     "core-dialogs/for-include-if-exist-drop-stmt-checkbox": "Include 'IF EXISTS' in DROP statement",
     "core-dialogs/for-null-cell-value-to-empty-str-label": "Convert null value to NULL in INSERT",
     "core-dialogs/choose-export-destination": "Please choose the destination for project export",
-    "core-dialogs/export-to-local": "Export to local",
-    "core-dialogs/export-to-google-drive": "Export to Google Drive",
+    "core-dialogs/export-to-local": "OpenRefine project archive to file",
+    "core-dialogs/export-to-google-drive": "OpenRefine project archive to Google Drive...",
+    "core-dialogs/export-to-google-sheets": "OpenRefine project archive to Google Sheets...",
```
wetneb: The Google Sheets option should be in the second section and should not be named "OpenRefine project archive to Google Sheets", since it does not export the project archive, it just exports the current state of the table like the other exporters. I think we could just use "Google Sheets" as label for that item.

Also these translations should be in the `gdata` extension (since the menu items should be added in the `gdata` extension, see below).

extensions/wikidata/module/langs/translation-en.json
```diff
@@ -3,10 +3,10 @@
     "wikidata-extension/edit-wikidata-schema": "Edit Wikidata schema",
     "wikidata-extension/import-wikidata-schema": "Import schema",
     "wikidata-extension/manage-wikidata-account": "Manage Wikidata account",
-    "wikidata-extension/perform-edits-on-wikidata": "Upload edits to Wikidata",
-    "wikidata-extension/export-to-qs": "Export to QuickStatements",
+    "wikidata-extension/perform-edits-on-wikidata": "Wikidata edits...",
+    "wikidata-extension/export-to-qs": "QuickStatements file...",
     "wikidata-extension/export-schema": "Export schema",
-    "wikidata-extension/export-wikidata-schema": "Export Wikidata schema",
+    "wikidata-extension/export-wikidata-schema": "Wikidata schema to file",
```
wetneb: This label is also used in the Wikidata extension menu, and in that context we do need to keep "Export" in the label because the menu contains all sorts of actions, not just export ones. So I would keep this existing string and add a new one without "export" (and without adding "to file" either, see below).

main/webapp/modules/core/langs/translation-en.json
antoine2711: > Is it really worth adding "file" at the end of all these labels? For instance, "ODF spreadsheet file" sounds quite redundant for me. I would keep these labels as they were.

   Also, spreadsheet file is not redundant: the spreadsheet can be saved to the cloud or elsewhere. But in my proposition, the menu item is precise, it tells you when you will generate files.

extensions/wikidata/module/langs/translation-en.json
antoine2711: > This label is also used in the Wikidata extension menu, and in that context we do need to keep "Export" in the label because the menu contains all sorts of actions, not just export ones. So I would keep this existing string and add a new one without "export" (and without adding "to file" either, see below).

The menu where « Wikidata schema to file » is located, is called « Export ». So, it's not necessary to repeat it, as you do not do with all the export above that also export (though not the same thing, that why I did put in my proposition this item in a section by itself.

AghaSaad04:
@wetneb I have made the suggested changes. Let me know if there is something needs to be done.

PR #2466, opened 2020-03-23
Thad no longer supports Netbeans config

thadguidry:


wetneb:
These config files are probably massively outdated, so it probably does not help to keep them. If anyone uses Netbeans and would be keen to contribute their configuration files, it would be great to have them.

Issue #2304, opened 2020-02-03
Make maxlag configurable in Wikidata extension


wetneb:
Maxlag is a parameter that controls how aggressive a mass-editing tool should be when uploading edits to Wikidata.

With the current issues around Wikidata's stability it is often impossible to edit with maxlag set to 5 seconds, the default value recommended by the guidelines (and the one set in OpenRefine by default).

As part of  #1640 we want to make the configuration of the Wikibase instance user-configurable (instead of being hard-coded to Wikidata) and part of that includes letting users set the maxlag value as they wish.

james-cui:
Hey @wetneb, I want to take a look at this as the potential GSoC applicant. May I know where the variable was set to 5 sec by default?


Thanks!

wetneb:
Sure!
Maxlag is set by default to 5 in Wikidata-Toolkit itself and can be changed here:
http://wikidata.github.io/Wikidata-Toolkit/org/wikidata/wdtk/wikibaseapi/WikibaseDataEditor.html#setMaxLag(int)

User preferences can be stored in the `PreferenceStore` (which can be accessed from the start page, in the "Preferences" section).

james-cui:
@wetneb Here is the request to merge 
https://github.com/OpenRefine/OpenRefine/pull/2469/commits/af09b05d502157d9b4314ec71adec79e1d7e550a

tfmorris:
> With the current issues around Wikidata's stability it is often impossible to edit with maxlag set to 5 seconds

That's the whole point, as I understand it. Allowing users to defeat this opens OpenRefine to being used as a tool for abuse.

Should this be added to the set of Wikibase configuration parameters so that a sysadmin can advertise the maximum value allowed and then users can set it to a lower, but not higher, value? That way if Wikidata advertises 5, users won't be able to change it to 10.

wetneb:
In my opinion there is no reason for OpenRefine to forbid its users to do something that the API allows. Other bot editing frameworks (pywikibot) also let the user configure this parameter.

If Wikibase admins want to enforce a specific throttling, they can do that on their side. Editing requests made from OpenRefine are all marked with `bot=true` and tagged appropriately.

WMDE is indeed looking into changing this throttling mechanism: https://phabricator.wikimedia.org/T240442

tfmorris:
That graph is frightening. 
[...]

PR #2469, opened 2020-03-24
make maxlag configurable

james-cui:
./refine build

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
@@ -43,6 +43,9 @@ import org.wikidata.wdtk.datamodel.interfaces.MonolingualTextValue;
 import org.wikidata.wdtk.wikibaseapi.WikibaseDataEditor;
 import org.wikidata.wdtk.wikibaseapi.WikibaseDataFetcher;
 import org.wikidata.wdtk.wikibaseapi.apierrors.MediaWikiApiErrorException;
+import com.google.refine.ProjectManager;
+import com.google.refine.preference.PreferenceStore;
+
 
 /**
  * Schedules and performs a list of updates to items via the API.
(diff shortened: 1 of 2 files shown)


extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
```diff
@@ -67,6 +70,14 @@
     private int globalCursor;
     private Map<String, EntityDocument> currentDocs;
     private int batchSize;
+    private static final String MAX_LAG_KEY = "maxLag";
```
wetneb: I would replace `"maxLag"` by something a bit more specific, to make sure there is no collision with anything else: for instance `"wikibase:upload:maxLag"`.

extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
```diff
@@ -67,6 +70,14 @@
...
+    private static final int MAX_LAG_DEFAULT = 5000;      // 5 second default maxLag
+    private PreferenceStore prefStore = ProjectManager.singleton.getPreferenceStore();
+
+    private int getMaxLag() {
+        return prefStore.get(MAX_LAG_KEY) == null ? MAX_LAG_DEFAULT :
+                Integer.parseInt((String) prefStore.get(MAX_LAG_KEY));
```
wetneb: What happens if `parseInt` fails?  Should it fall back to the default?

extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
james-cui: will improve the naming.

extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
james-cui: good catch, I should wrap the code in a try catch and use default if user put something invalid.

james-cui:
@wetneb, resolved the review comments, do you want to take a look?

wetneb:
Great! Can we have an accompanying test perhaps? Either of your `getMaxLag` function (which could be `protected` instead of `private`), or just by checking that `setMaxLag` is called on the mocked `WikibaseDataEditor`. Both with preferences set and without.

james-cui:
@wetneb  resolved your comments
./refine extensions_test 
passed all tests

wetneb:
Closes #2304.

antoine2711:
@wetneb: here, we use semi columns in the key, but this is not a pattern used elsewhere. Shouldn't dots be used instead?

i.e. `wikibase.upload.maxLag` instead of `wikibase:upload:maxLag`?

Regards, A.

tfmorris:
Yes, we should make this consistent before the release of v3.4.

PR #2470, opened 2020-03-24
remove Freebase reconciliation from Excel Importer

Abbe98:


Abbe98:
@wetneb my thoughts exactly.

PR #2471, opened 2020-03-24
[FEAT]Adds new options for split

### LINKED ISSUE #2238: add new option for spliting multi-valued cells: by transition between lowercase and uppercase, and number and text

I discoverd interesting options in MS PowerQuery. It could be nice to have the same in OR : split by transition between lowercase and uppercase, split by transition between number and text

Here is the menu (in french) in PowerQuery : 

image

**Additional context**

lisa761:
Closes #2238 

Now splitting multi-cell values can have two more options:
* On transition from lowercase to uppercase
* Split between numbers and letters


main/src/com/google/refine/commands/cell/SplitMultiValueCellsCommand.java
```diff
@@ -84,6 +84,20 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
                 Process process = op.createProcess(project, new Properties());
                 
+                performProcessAndRespond(request, response, project, process);
+            } else if ("cases".equals(mode)) {
+                AbstractOperation op = new MultiValuedCellSplitOperation(columnName, 
+                                                                         keyColumnName,
+                                                                         true);
+                Process process = op.createProcess(project, new Properties());
+
+                performProcessAndRespond(request, response, project, process);
```
wetneb: Instead of duplicating the two lines `Process process = op.createProcess(project, new Properties());` and `performProcessAndRespond(request, response, project, process);` in each case of the `if`/`else` block, can we move them outside so that they are shared for all blocks? (The original code already had that duplication, it just feels more painful now that we have more cases)

lisa761: @wetneb Yes, I should do that.

lisa761:
@wetneb You're right, that would have been easier and much cleaner than this, I can make the changes. It would make sense to keep the values of `regex` in the front end itself.

main/tests/server/src/com/google/refine/operations/cell/SplitMultiValuedCellsTests.java
```diff
@@ -163,5 +163,54 @@ public void testSplitMultiValuedCellsLengths() throws Exception {
...
+        Assert.assertEquals(project.rows.get(3).getCellValue(keyCol), null);
+        Assert.assertEquals(project.rows.get(3).getCellValue(valueCol), "Eight;nine91011twelve thirteen 14Fifteen");
+    }
+
+    @Test
+    public void testSplitMultiValuedCellsTextNumber() throws Exception {
+        AbstractOperation op = new MultiValuedCellSplitOperation(
+            "Value",
+            "Key",
+            "(?<=[0-9]|[0-9][\\s])(?=[A-Z]|[a-z])|(?<=[a-z]|\\s|[A-Z])(?=[0-9])",
```
lisa761: I let the tests remain just for the sake of different mode names w.r.t UI, otherwise there already exists a test for separator with regex.

msaby:
@wetneb Here is the MS Powerquery behavior : 

split by transition from lowercase to uppercase  : 

original string : fooBarABABFooBar

image

the function in the GUI generates a line of code  :  = Table.SplitColumn(#"Type modifié", "test", Splitter.SplitTextByCharacterTransition({"a".."z"}, {"A".."Z"}), {"test.1", "test.2", "test.3", "test.4"})

Now split by transition from uppercase  to lowercase :

image

the line of code generated : = Table.SplitColumn(Source, "test", Splitter.SplitTextByCharacterTransition({"A".."Z"}, {"a".."z"}), {"test.1", "test.2", "test.3", "test.4"})

wetneb:
Ok great! And what about the transition between letters and numbers, is that also unidirectional?

msaby:
image


From number to non-number

image

= Table.SplitColumn(#"Type modifié", "test", Splitter.SplitTextByCharacterTransition({"0".."9"}, (c) => not List.Contains({"0".."9"}, c)), {"test.1", "test.2", "test.3", "test.4", "test.5"})

from non-number to number

image

= Table.SplitColumn(Source, "test", Splitter.SplitTextByCharacterTransition((c) => not List.Contains({"0".."9"}, c), {"0".."9"}), {"test.1", "test.2", "test.3", "test.4", "test.5"})

lisa761:
So, should I add all 4 options?

thadguidry:
Maybe radio buttons for each primary option, with reverse toggle checkboxes?

image

lisa761:
@thadguidry I will go ahead with that then

wetneb:
Hmm, it's not clear to me that the UI above is really clear? Giving examples like "Ab" is a good idea but perhaps it might be confusing too, since by splitting between upper then lower you get values which look like "aB"… >_<

Perhaps it would help to have an idea of the sort of data this function is generally used on?

thadguidry:
@msaby needs to speak up about his examples.  I was only trying to help flesh out a possible UI method (which can be completely blown away and done differently based on what he is really looking for)

thadguidry:
Or perhaps we should show users a fuller string example in the dialog to avoid confusion?
image

antoine2711:
> Or perhaps we should show users a fuller string example in the dialog to avoid confusion?

Yes, longer string help to better see the pattern.

I think we could also have just one checkbox, below the 2 radios buttons. And the text of the checkbox would change to reflect the choice of the radio buttons.

Regards, A.

thadguidry:
@antoine2711 now your getting all fancy and turning into a UI expert.  I like it, harder to implement, but I like it!

antoine2711:
> @antoine2711 now your getting all fancy and turning into a UI expert. I like it, harder to implement, but I like it!

I actually am a Mac user since 1985… Human interface has always been an area of concern and interest…

Let's just say that OpenRefine has a lot of potentiel on the interface side! ;-)

Regards, Antoine

lisa761:
> Also, could we take the text of the current selected cell to show the result?  

There are no selected cells per se. When we chose the split multi-valued cells, the option is available through that cell's corresponding column, and it splits all the cells that can be split in a similar manner (choosing from the available options) in that column.

antoine2711:
But you can still take the value of that column for row index 0.

Regards, Antoine

Le ven. 27 mars 2020 à 07:47, Lisa Chandra <notifications@github.com> a
écrit :

lisa761:
I think I will wait for any other suggestions, just in case. I'd be happy to go ahead with what @antoine2711 suggested if everybody agrees.

Personally though, instead of using row 0 and the corresponding column value for an example, we would have to traverse through the row and find the first cell that can be split in the manner that the user wants it to, which I think might get a little confusing for the users, in case there is no cell present on the current page which can be split that way. If that happens, then the split example, first of all, would need to render the value of a cell that has not yet been loaded into the DOM, and I am afraid it would be equivalent to just showing a default example to the user. 
Another thing is that I think it might be confusing if there's a single checkbox for both radio buttons, it might be easy to overlook for first time users (especially if it will be below both of them). We might need to indicate that the checkbox will work for both options. Although, the users should be able to figure it out if they play with the options a little bit. But in case a user first checks the reverse checkbox, we will also have to tell the users that they first have to select one of the two options above.

main/tests/server/src/com/google/refine/operations/cell/SplitMultiValuedCellsTests.java
```diff
@@ -163,5 +163,54 @@ public void testSplitMultiValuedCellsLengths() throws Exception {
...
+        int valueCol = project.columnModel.getColumnByName("Value").getCellIndex();
+
+        Assert.assertEquals(project.rows.get(0).getCellValue(keyCol), "Record_1");
+        Assert.assertEquals(project.rows.get(0).getCellValue(valueCol), "one:two;three four;five");
+        Assert.assertEquals(project.rows.get(1).getCellValue(keyCol), null);
+        Assert.assertEquals(project.rows.get(1).getCellValue(valueCol), "Six ");
+        Assert.assertEquals(project.rows.get(2).getCellValue(keyCol), null);
+        Assert.assertEquals(project.rows.get(2).getCellValue(valueCol), "Seven");
+        Assert.assertEquals(project.rows.get(3).getCellValue(keyCol), null);
+        Assert.assertEquals(project.rows.get(3).getCellValue(valueCol), "Eight;nine91011twelve thirteen 14Fifteen");
```
tfmorris: Test cases including accented upper and lower case characters are needed here. Probably also Unicode digits outside the 0-9 range.

main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js
```diff
@@ -347,6 +350,12 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
           alert($.i18n('core-views/warning-format'));
           return;
         }
+      } else if (mode === "cases") {
+        config.separator = "(?<=[a-z]|[a-z][\\s])(?=[A-Z])";
```
tfmorris: This won't work for characters with diacritics e.g. Éè

antoine2711:
> I think I will wait for any other suggestions, just in case. I'd be happy to go ahead with what @antoine2711 suggested, otherwise.
> 
> Personally though, instead of using row 0 and the corresponding column value for an example, we would have to traverse through the row and find the first cell that can be split in the manner that the user wants it to (…)

Yes, I would also do that. And, I would put that value in a textbox that the user can edit.
Also, I would have a result textbox (not editable) that would preview the result for this single value.

And, I think that @thadguidry first proposition for the checkboxes might be in line with this, where I did also add a checkbox to reverse the column creation order, which is not the same thing as reversing the splitting pattern.

Here a picture of a possible interface (of course the spacing is way not enough):
image

Regards, Antoine

lisa761:
@antoine2711 Thanks, it makes the UI much clearer! I think some checkboxes here are different from the issue this PR solves, so maybe we could create another issue for it first? I think the input and result textfield is a good idea, but it would make more sense to do it for all the available options (which I guess is what you mean anyway) and not just the last two .

antoine2711:
> I think some checkboxes here are different from the issue this PR solves, so maybe we could create another issue for it first?

The two reverse the two « between » radio buttons are in the logic of this PR/Issue. But the last one, that applies to ALL options, this is something new I just got out of my hat. I really can't fool you, it seams… ;-) 

> I think the input and result textfield is a good idea, but it would make more sense to do it for all the available options (which I guess is what you mean anyway) and not just the last two.

Yes, of course, it's for all. And note that the input field is editable, so the user can even put his own input to see the results with his choices.

Regards, Antoine

P.S. Here the PR: [« Split into several columns » should have the option to fill the columns created in a reverse order #2505](https://github.com/OpenRefine/OpenRefine/issues/2505)

msaby:
Sorry I don't have time for taking a closer look at that. For the record, here was the original use case : a colleague wanted to separate
"Eglise catholique  Aspect social" in 2 columns "Eglise catholique" + "Aspect social"
these words are subject headings used in a library catalog. The capitalization in the original data was used to separate the 2 headings

antoine2711:
> Sorry I don't have time for taking a closer look at that. For the record, here was the original use case : a colleague wanted to separate
> "Eglise catholique Aspect social" in 2 columns "Eglise catholique" + "Aspect social"
> these words are subject headings used in a library catalog. The capitalization in the original data was used to separate the 2 headings

@msaby: but in your example, there was never spaces… So, we should ignore spaces with this operation?

Regards, Antoine

msaby:
After the request of this colleague (which can be answered with GREL), I discovered the options for spliting string by transition beetween lowercase and uppercase in MS Powerquery, and I thought it could have been useful for this use case. But you are right, now I realize this use case was a bit more complex, and should not guide the developpement of this new feature.

So you would rather develop something generic that can be useful in several occasions.

By the way, I am not fan of the "Reverse aB" label. As a user I will maybe understand it as "the feature will reverse the order of the letters", and that is not the idea... Is it possible to find a more explicit label ?


> 
> @msaby: but in your example, there was never spaces… So, we should ignore spaces with this operation?
> 
> Regards, Antoine

antoine2711:
> After the request of this colleague (which can be answered with GREL), I discovered the options for splitting string by transition between lowercase and uppercase in MS PowerQuery, and I thought it could have been useful for this use case. But you are right, now I realize this use case was a bit more complex, and should not guide the development of this new feature.

Well, I think there is a clear case that this can be a useful tool. But the spaces are usually dealt with sooner.

As for the « Reverse "aB" », do this labeling « Reverse splitting order "aB" » would be clearer?

Regards, Antoine

msaby:
Yes I think « Reverse splitting order "aB" » would be clearer, and easier to translate

antoine2711:
@lisa761: I guess that with the presence of spaces, the processus should ignore them and split like if spaces didn't exists, while retaining them. But, if the were a split where _some_ spaces are, (i.e. with "Ab cd    Ef gh"), then, the process would trim them (in my last example: ["Ab cd", "Ef gh"]). I also think that covers will @msaby example of names: `"Église catholique Aspect social" in 2 columns "Église catholique" + "Aspect social"`.

I means, we could go on and on add many options, but if we want to convert the most ground with reasonable effort, I think we are at a good spot. Because there is also testing (and maybe other bug fixes ?!) and documentation that will come with all of this.

I'm curious to hear a bit from the others about the question of spaces…

One last question: does your code handle accentuated letters like «É» and «ç» and treat them as upper case and lower case characters?

Regards, Antoine

lisa761:
Hi @antoine2711 currently, the code does ignore spaces, that is it would split ['Ab cd Ef gh'] to '[Ab cd ', 'Ef gh']. But it does not trim the last space as can be seen in the 'Ab cd ' split which still has the remaining space (I didn't trim space because the other options available currently also let trailing spaces remain, plus there is an option to trim all trailing/leading whitespace if the user wishes to do so). The code can also handle  transition from lower case accented letters to normal uppercase but I have to modify it to handle all transitions including accented letters.

antoine2711:
Hi @lisa761, 

> (I didn't trim space because the other options available currently also let trailing spaces remain, plus there is an option to trim all trailing/leading whitespace if the user wishes to do so).

That makes total sense. I agree with you. And good luck with the remaining accentuated letters! ;-) Being a french speaking guy, I use them much.

I see good work with this feature.

Regards, Antoine

main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js
```diff
@@ -351,10 +353,18 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
           return;
         }
       } else if (mode === "cases") {
-        config.separator = "(?<=[a-z]|[a-z][\\s])(?=[A-Z])";
+        if(elmts.reversTranistionCases[0].checked) {
+          config.separator = "(?<=[A-Z]|[À-Ý]|[A-Z\\s][À-Ý\\s])(?=[a-z]|[à-ÿ])";
+        } else {
+          config.separator = "(?<=[a-z]|[a-z\\s]|[à-ÿ]|[à-ÿ\\s])(?=[A-Z]|[À-Ý])";
```
wetneb, line 359: I would avoid hard-coding the character ranges. We should be able to use unicode character ranges as specified here: https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html. It should just be a matter of enabling unicode in the Pattern declaration and then using the corresponding escape sequences.

lisa761:
@wetneb Hi, I was wondering if I should do the input and result thing in a different PR as it is kinda a new feature and related to all split modes?

lisa761:
Is there anything more required in this PR?

PR #2472, opened 2020-03-24
Fix org.eclipse.m2e lifecycle-mapping not found

thadguidry:


wetneb:
Do we really need any of that? I don't think we use the install plugin since I removed all local jars anyway (in #2367). So you can probably remove the lifecycle mappings altogether.

thadguidry:
Dunno.  But this fixes the vastly annoying messages in IntelliJ and VSCode for developers now!  :)  As a bonus, this does show how we can use `profiles` quite well for future needs.  Up to you if you want to gut it all.

wetneb:
Yes let's remove it all and keep profiles in mind - we should definitely use them if we ever need to add IDE-specific config.

Issue #2031, opened 2019-04-26
Add column by fetching URLs fails on a HTTP 200 response


ostephens:
**Describe the bug**
In some cases "Add column by fetching URLs" fails to populate content into the column, even though the URL request gets a 200 OK response and accessing the URL via curl or a browser returns content

**To Reproduce**
1. Create a cell populated with the URL "https://www.trismegistos.org/place/18340/json"
2. Use "Add column by fetching URLs", check option to store error
3. Click OK
4. See new column added with cell displaying error `HTTP error 200 : OK | `

**Expected behavior**
Expected cell to be populated with JSON returned from https://www.trismegistos.org/place/18340/json

**OpenRefine (please complete the following information):**
 - tested with OpenRefine 3.1 and OpenRefine 3.2 beta

**Additional context**
Reported on Google Group https://groups.google.com/d/msg/openrefine/YAXrXTIXk-k/MQjxxuI_DAAJ
@thadguidry did a CURL dump for the URL
```
THAD's CURL DUMP on Windows 10...

THAD@DESKTOP-TJUMQT0 MINGW64 ~
$ curl -v https://www.trismegistos.org/place/18340/json
* Expire in 0 ms for 6 (transfer 0x66ac80)
* Expire in 1 ms for 1 (transfer 0x66ac80)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Expire in 0 ms for 1 (transfer 0x66ac80)

*   Trying 134.58.134.15...
* TCP_NODELAY set
* Expire in 200 ms for 4 (transfer 0x66ac80)
* Connected to www.trismegistos.org (134.58.134.15) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: C:/Program Files/Git/mingw64/ssl/certs/ca-bundle.crt
  CApath: none
} [5 bytes data]
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
} [512 bytes data]
* TLSv1.3 (IN), TLS handshake, Server hello (2):
{ [106 bytes data]
* TLSv1.2 (IN), TLS handshake, Certificate (11):
{ [3160 bytes data]
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
{ [333 bytes data]
[...]

PR #2475, opened 2020-03-25
Enable gzip compression

zengchu2:
**Realted Issue**
As mentioned in issue #2031 , the InputStreamReader does not support GZIP compressed contents. 

Tried to use a GZIPInputStream to address response with `content-encoding:gzip`.


zengchu2:
No problem, I would need some time to come up with a better solution.

zengchu2:
Hi, I added a web independent test case for testing `ParsingUtilitiesTests.inputStreamToString`.

Another change I made for `setRequestProperty("Accept-Encoding", "gzip")` seems quite difficult to test for the following reasons.
- The purpose of this line is to enable response returned with a gzip compressed body if available.
-  The response is dependent on behaviors of remote server (like https://www.google.com) 
-  We can only access the response header  within the `fetch()` function from `ColumnAdditionByFetchingURLsOperation`, storing only the response as a public field only for testing does not seem good to me .

Any suggestions are appreciated towards this fixes, thanks !

Issue #2306, opened 2020-02-09
gSheet export thru « Custom tabular exporter… » doesn't work with empty values


antoine2711:
**Describe the bug**
I'm trying to export with « Custom tabular exporter… » and choosing « Upload to -> A new Google spreadsheet », and in some rows, if a value is missing, then the cell value disapear and the cell on the right shift left.

**To Reproduce**
1. « Custom tabular exporter… » 
2. choose « Upload to -> A new Google spreadsheet »
3. See the resulting document with mess up data.

**Current Results**
Some cell are shifted left when there is no data in the cell on the left.

**Expected behavior**
A clear and concise description of what you expected to happen or to show.

**Screenshots**
![Export-gSheet_decal_col_VueGA](https://user-images.githubusercontent.com/12217525/74112812-974ac680-4b6d-11ea-93b1-0b26463fc695.png)
![Export-gSheet_decal_col_VueOR](https://user-images.githubusercontent.com/12217525/74112813-987bf380-4b6d-11ea-86f2-f5669f7874d5.png)

**Desktop (please complete the following information):**
 - OS: MacOS 10.14.6
 - Browser Version: Chrome latest
 - JRE 13.0.1

**OpenRefine (please complete the following information):**
 - Version 3.3

Relates to Google Groups « OpenRefine » discussion : [This seams to be a bug in exporting to gSheet](https://groups.google.com/forum/#!topic/openrefine/8BjfWdLwUhs)

linderbarbara:
Hi everyone. I would like to solve this bug. Can I work on this issue?

wetneb:
Sure! Let us know if you get stuck.

Nishtha3512:
Hey,where i get the code related to this issue? please share link with me if any
Thanks :)

wetneb:
Are you still working on this @linderbarbara?

The relevant code is in `extensions/gdata/src/org/openrefine/extension/gdata/SpreadsheetSerializer.java`.

SubhamPramanik:
Hi! I would like to try this issue.
I'm getting this error during the OAuth call to Google while trying to recreate the issue.

Google
400. That’s an error.
Error: invalid_request
Missing required parameter: client_id

I've built the project locally on Eclipse. I use a Windows system.

wetneb:
You will need to set up Google OAuth credentials as described here:
https://github.com/OpenRefine/OpenRefine/wiki/Google-Extension

SubhamPramanik:
[...]

PR #2476, opened 2020-03-25
Issue 2306: Fixes

SubhamPramanik:
The for loop had a null checker, if the cell is empty it wouldn't be added to the Sheets Cell ArrayList, hence the empty cells weren't recognized and the remaining cells were shifted to the left. 

Removing the null checker allows for all the cells to be added to the Sheet, including the ones which are empty.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ target/
 # Ignore Apache Maven default download path
 apache-maven-*/
 apache-maven-*-bin.tar.gz
+/.metadata/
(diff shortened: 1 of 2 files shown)


SubhamPramanik:
> Very nice, thank you!
> 
> Do you think you could write a test for this? You should be able to test the behaviour of the `SpreadsheetSerializer` in isolation, without actually sending the generated spreadsheet to Google.
> There is currently no test class for this class, so we would need to introduce one.

Sure. I'm looking into it.

tfmorris:
@SubhamPramanik  Thank you for working on this and isolating the cause of the problem. A fix for #2306 was included in #2782, so I'm closing this, but feel free to choose a different issue to tackle.

Issue #875, opened 2014-05-13
Network error kills status updates


tfmorris:
While running a long "Add column by fetching URLs" my laptop went to sleep and I got:

Failed to load resource: net::ERR_NETWORK_IO_SUSPENDED http://127.0.0.1:3333/command/core/get-processes?project=1445522723935

when the laptop resumed, the server job continued, but the front-end was stuck, no longer getting status updates.  It should retry on a network error (and/or inform the user).

Rookev:
Same error here. Sometimes, after a bit of inactivity, the front end is stuck. I can see graphical changes like hover effects for buttons but the connection seems to be lost and won't come back. Error console shows:

Failed to load resource: net::ERR_NETWORK_IO_SUSPENDED

jackyq2015:
@Rookev, According to this:
http://stackoverflow.com/questions/23725085/failed-to-load-resource-neterr-network-io-suspended
It might caused by the ajax call. 

If you can exactly put the steps to reproduce this error, I am happy to pin point it and root cause and fix it.

tfmorris:
It looks like adding error/retry code here: https://github.com/OpenRefine/OpenRefine/blob/1638d68e352a89c3481d661b0cc950e667447da2/main/webapp/modules/core/scripts/util/ajax.js#L45-L53
would be a likely fix. The timeout added in #2482 will keep the client from getting hung forever, but probably won't be a complete solution.

wetneb:
This is fixed in 4.0 by adding a global jQuery error handler which notifies the user if any query to the backend fails due to a connection error:
https://github.com/OpenRefine/OpenRefine/blob/c87d082fe27b1bcdbd2227c107d7e12b77d49cdb/main/webapp/modules/core/scripts/util/ajax.js#L37-L60

PR #2482, opened 2020-03-26
Set client timeout limit in chainGetJSON()

Abbe98:
By setting a client side timeout one can avoid net::ERR_NETWORK_IO_SUSPENDED errors caused by the OS network layer being suspended [1]. This allows the client to handle higher level errors.

The jQuery.getJSON() function does not have a parameter for timeout management.

[1]: https://cs.chromium.org/chromium/src/net/http/http_network_layer.cc?q=ERR_NETWORK_IO_SUSPENDED&sq=package:chromium&dr=C&l=42

### DIFF
--- a/main/webapp/modules/core/scripts/util/ajax.js
+++ b/main/webapp/modules/core/scripts/util/ajax.js
@@ -42,10 +42,15 @@ Ajax.chainGetJSON = function() {
       var data = a[i++];
       var callback = a[i++];
 
-      $.getJSON(url, data, function(o) {
+      $.ajax({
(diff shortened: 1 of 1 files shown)


Abbe98:
While there are other client requests that could benefit from this the functionality using `chainGetJSON()` seems most likely to run during actual suspension of the OS network layer.

Related: #875

Abbe98:
@wetneb I'm afraid this change wont solve anything, but future issues will likely throw other high level errors that one could possibly handle in the future.

PR #2486, opened 2020-03-27
Issue 2272 portable package windows

Shubhangi2901:
2 files were updated for generating a portable windows package.

In packaging/windows.xml, I added configuration for copying JRE from the build system inside the portable package for windows when openRefine is built.

In packaging/pom.xml, I configured launch4j for using the JRE copied over from the previous step, instead of the version installed on the system. The previous version of packaging/pom.xml was incorrectly referring to %JAVA_HOME% for finding the installed JRE, whereas it should use ${java.home}, as I have added in the windows.xml file above.

I have tested the package on a machine with no JRE installed, and was able to get openRefine running on the system.


wetneb:
Also, importantly: our Maven configuration should generate *two* Windows packages: one with the JRE, and one without it, to cater for all needs.

thadguidry:
@Shubhangi2901 what was your test machine (OS,etc.) ?

Shubhangi2901:
> 
> 
> @Shubhangi2901 what was your test machine (OS,etc.) ?

I tested it on windows platform

Shubhangi2901:
@wetneb maven configures the JRE resource in server/target folder inside the **openrefine-win-3.4-SNAPSHOT** created at the time of **Maven build**. This JRE resource is then copied to the user system while installation. So this way the application always uses the JRE bundled in the application.

I have not incorporated this feature for MAC or Linux platform.

wetneb:
Sorry, my bad! Then this `server/target/jre` path is fine.

However, you have used `${java.home}` as source path for the JRE, so that will only work if you are doing the packaging on Windows. It should be possible to do this packaging step on any OS, provided we download a Windows JRE in a specific location.

Then, my comment above applies to that path: this should be configurable to point to a local copy of a Windows JRE, just like we do for Mac.

Shubhangi2901:
> 
> 
> Sorry, my bad! Then this `server/target/jre` path is fine.
> 
> However, you have used `${java.home}` as source path for the JRE, so that will only work if you are doing the packaging on Windows. It should be possible to do this packaging step on any OS, provided we download a Windows JRE in a specific location.
> 
> Then, my comment above applies to that path: this should be configurable to point to a local copy of a Windows JRE, just like we do for Mac.

Thanks for your suggestion @wetneb. I initially was trying to use %JAVA_HOME%;%PATH% in that place (as it exists that way already in packaging/pom.xml), however, that does not work as intended, and while building Maven gives an error about JRE incorrectly specified (since %JAVA_HOME% actually points to the JDK, not JRE), and hence I had used ${java.home}.

I will now try and use a settings.xml as you pointed out for the packaging process, and update my PR.

For the Mac hardcoded path "/opt/jre1.8.0_181.jre", I would prefer if we create a separate issue for that, which I can fix in a separate PR, since this issue was for creating a portable windows package. Does that work for you?

thadguidry:
@Shubhangi2901 Yes please, separate PR,  Thanks!  I'll test on my Windows 10 once it's all complete.

Shubhangi2901:
Hi @wetneb and @thadguidry I have added a new settings file to the project which includes path to my windows/jre, this path is then used by maven to bundle JRE with **openrefine-win-3.4-SNAPSHOT**.

However for mvn to use this settings file you will need to run **mvn install -s settings.xml**. I have tested installation on a windows machine with no JRE installed and was able to get it running.

packaging/pom.xml
```diff
@@ -231,7 +231,7 @@
                 <postCp>server/target/lib/*.jar</postCp>
               </classPath>
               <jre>
-                <path>%JAVA_HOME%;%PATH%</path>
+                <path>server/target/jre</path>
```
wetneb, line 234: We should be able to produce two different Windows packages: one with embedded JRE, one without embedded JRE (for people who already have Java on their machine). So instead of modifying the existing Launch4j configuration, we would need to duplicate it, to be able to run with both `<path>%JAVA_HOME%;%PATH%</path>` and with `<path>server/target/jre</path>`. Perhaps you could take inspiration from the profiles introduced by @afkbrb in #2498?

Shubhangi2901: Hi @wetneb thanks for the suggestion. I looked the up the issue #2498 , i understand one has to use profiles to generate two different types of exec depending on the option given during compilation. I will update my commit with the suggested inclusion.

Shubhangi2901:
@wetneb @thadguidry did you guys get a chance to review my PR? Thanks!

For the other issue referenced above about the hardcoded path ["/opt/jre1.8.0_181.jre"] for a mac build, should I create an issue on OpenRefine? Any guidelines on creating the issue?

wetneb:
I have not forgotten this, I will try to get this merged soon. Thanks for your patience!

wetneb:
I have restored the `./refine dist` functionality: by default, without passing any profile, the packaging script needs to work. By default it builds a Windows package without an embedded JRE. To build a windows package with embedded JRE, the syntax is still `mvn package -s settings.xml -P embedded-jre`. I have also changed the filename of the resulting package to `openrefine-win-with-java-3.4-SNAPSHOT.zip` to distinguish it from `openrefine-win-3.4-SNAPSHOT.zip`.

PR #2488, opened 2020-03-28
Provides more intuitive representation for arrays in GREL

### LINKED ISSUE #2040: toString() for array arguments in GREL returns a memory address.  Consider giving a more intuitive representation.
In GREL, `toString( [ 3 ,4])`, will return something like: `[Ljava.lang.Object;@79738ce6`.  It would be helpful to provide a more intuitive and natural representation, such as: `[ 3, 4 ]`.
Something like `java.util.Arrays.deepToString` might give a better meaning when running toString on an array in GREL.

Thoughts?

**Some Additional Information:**
You will also notice that the expression preview of the array ` [ 3 ,4]` already shows a more natural representation: `[ 3, 4 ]`.  Because arrays are not storable however, attempting to run this transformation leads to the error: `Object[] value not storable`.  This is conflicting behavior.  The preview shows a valid transformation, yet the execution of the transformation results in an error.  See issue #1088 for more information on storing and displaying arrays in GREL.

darecoder1999:
Closes #2040

sample

### DIFF
--- a/main/tests/server/src/com/google/refine/util/StringUtilsTests.java
+++ b/main/tests/server/src/com/google/refine/util/StringUtilsTests.java
@@ -1,0 +1,25 @@
+package com.google.refine.util;
+
+import com.google.refine.RefineTest;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+import java.time.OffsetDateTime;
+import com.google.refine.util.StringUtils;
+
+public class StringUtilsTests extends RefineTest {
+
+    @Test
+    public void objectToString() {
+        Object nullObject = null;
+        Object[] emptyArray = {};
+        Object[] objArray = {4, "hello", true, 0.01};
+        Object[][] multiArray = {{"OpenRefine", 12}, {13, 4.6}, {"data", "mining"}};
+        OffsetDateTime time = OffsetDateTime.parse("2017-01-02T01:02:03Z");
+
+        Assert.assertEquals("", StringUtils.toString(nullObject));
+        Assert.assertEquals("[]", StringUtils.toString(emptyArray));
+        Assert.assertEquals("[4, hello, true, 0.01]", StringUtils.toString(objArray));
+        Assert.assertEquals("[[OpenRefine, 12], [13, 4.6], [data, mining]]", StringUtils.toString(multiArray));
+        Assert.assertEquals("2017-01-02T01:02:03Z", StringUtils.toString(time));
+    }
+}
--- a/main/src/com/google/refine/util/StringUtils.java
+++ b/main/src/com/google/refine/util/StringUtils.java
@@ -27,6 +27,7 @@
 package com.google.refine.util;
 
 import java.time.OffsetDateTime;
+import java.util.Arrays;
 
 public class StringUtils {
     /**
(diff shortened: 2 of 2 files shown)


darecoder1999:
@ostephens @wetneb kindly please review this and suggest if any changes are needed to be done. :smile:

main/tests/server/src/com/google/refine/util/StringUtilsTests.java
```diff
@@ -0,0 +1,20 @@
...
+
+public class StringUtilsTests extends RefineTest {
+
+    @Test
+    public void objectToString() {
+        Object[] nums = {4, "hello", true, 0.01};
+        OffsetDateTime time = OffsetDateTime.parse("2017-01-02T01:02:03Z");
+
+        Assert.assertEquals("[4, hello, true, 0.01]", StringUtils.toString(nums));
+        Assert.assertEquals("", StringUtils.toString(null));
```
thadguidry: Shouldn't this also test an Array that is empty, i.e `[]`?  Instead of a String that is empty, `""`?

Also, I am wondering about multi-dimensional arrays which is kinda the point of deepToString... If the array contains other arrays as elements, the string representation contains their contents and so on.  https://www.geeksforgeeks.org/arrays-deeptostring-in-java-with-example/

thadguidry:
OK good...
`toString([1,[2,[1,2,3]3]])`  = `Parsing error at offset 22: Missing ]`

But this...interesting.  @darecoder1999 Do you feel these should be a Parsing error also?
`toString([1[2,[1,2,3]]])`  outputs `[null]`
`toString([1,2[2,[1,2,3]]])`  outputs `[1, null]`

wetneb:
Those are very weird behaviours indeed, but they are beyond the scope of this PR I would say. Let's open a separate issue for them.

thadguidry:
Opened issue #2507 for the weirdness with GREL parsing.

antoine2711:
> OK good...
> `toString([1,[2,[1,2,3]3]])` = `Parsing error at offset 22: Missing ]`
> 
> But this...interesting. @darecoder1999 Do you feel these should be a Parsing error also?
> `toString([1[2,[1,2,3]]])` outputs `[null]`
> `toString([1,2[2,[1,2,3]]])` outputs `[1, null]`
> 
> More interesting.
> `[1,2[2,[1,2,3]]].type()` outputs `array` <-- other programs would say this is an incomplete array or throw a syntax error, anyone know what's going on with GREL here?
> 
> `[1,2,[2,[1,2,3]]].type()` outputs `array`

@thadguidry: I'm not sure I get this… 

How on earth can `[1[1]]` be an array? This should NOT, IMOO, be converted to an array.

When I put this in Cell->Transform, I get:
Input | Result
------------ | -------------
[1[1]]|Error: java.lang.StringIndexOutOfBoundsException: begin 1, end 2, length 1
[1,2[2,1]]|[ 1, "" ]
[1,2[2]]|Error: java.lang.StringIndexOutOfBoundsException: begin 1, end 2, length 1
[1[1,2]]|[ "" ]
[1,[1,2]]|[ 1, [ 1, 2 ] ]

I think you found a very nice bug! This is a +2 bug in my book. Bravo!

Regards,
   Antoine

thadguidry:
@antoine2711 let's talk in #2507

darecoder1999:
@thadguidry great observation, I will surely try to resolve #2507.

@wetneb: Are there any other changes, I'm supposed to do? Please do let me know, I'll do the needful.

Thanks, everyone for such a quick review.

Issue #1062, opened 2015-09-04
Edits made via "Edit this cell" occasionally fail to apply


joewiz:
In some circumstances that I'm not able to consistently reproduce, changes made a cell via "Edit this cell" and "Apply" are not applied. 

I will update this issue if I am able to supply reproducible test, but I'm submitting this in the hope that others can contribute tests/observations if they've seen this too.

(I'm using 2.6RC1 with current versions of Mac OS X and Google Chrome.)

Blakko:
It happened to me a couple of times, but it was just related to the interface not being updated. Next time try to refresh the page to see if it's the case :)

Opennat:
In addition to what have been said above, sometimes your edit is really applied to the cells, and sometimes not (you could check this in another browser tab with current project). In both cases you will see yellow "popup": "Edit ..... on row X, column Y Undo".

This bug occurs (I believe) in 100% cases, when you have 2 or more filters/facets active.
I'll check this later and make screencast.

joewiz:
@Opennat Thanks for what you suggested. I agree that this often happened when I had 2 or more filters/facets active. (And sorry I didn't spot your reply sooner.) Were you able to find a reproducible case? It's been a while since I was working intensively with OR, but if you had anything more to add, esp. with 2.6RC2, that would be worth keeping the issue open for.

tfmorris:
We've fixed the case where this was being triggered by text filter facets in #2846 (as demonstrated by @zengchu2's excellent analysis in #2492). Unfortunately, although I wrote the patch back in May 2016, it got lost in the shuffle and just got committed recently.

I'm going to close this, but if anyone finds any other reproducible cases where this happens, please open separate bug reports for them.

PR #2492, opened 2020-03-28
Fix cannot edit cell

zengchu2:
Address issue #1062.
**Bug Description**
When there is two or more filters and we clicked edit button as soon as the page is loaded, the cell is not updated as expected.
If we refresh the page, we can see the cell correctly updated.

**Cause of Error** 
When we click the edit button as soon as the filtered table is loaded,  a new `DataTableCellUI` is being created and rendered in the front-end. The edit is processed as expected but the front-end is actually displaying another object.

Once we refresh, a new CellUI object is created using `dataTableView` (with expected updated values) so user can see the expected cell.

**Fix** 
Once **dataTableView** has update user's expected value, we use **dataTableView** to re-rendered the cells in the page.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
@@ -614,8 +614,8 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
 
             self._cell = o.cell;
             self._dataTableView._updateCell(self._rowIndex, self._cellIndex, self._cell);
-            self._render();
             self._dataTableView._adjustDataTables();
+            self._dataTableView.render();
           }
         }
       );


zengchu2:
No problem, here are the steps in screen shots:
1.  Add `  console.log("Line34 : DataTableCellUI gets called", cell);
`,  `  console.log("Line 528: Start Edit Cell");
` to  line 34 and line 528 of `cell-ui.js`.
2.  Build and run the app,  use Clipborad to import the following text:
T1
T2
UNUSED

3. Create the two project and add 1 filter, typing `T`.
4. Create another filter, type `2`, and click on the cell and edit it. In console, we are expcting to see these outputs:
output
Note, after we click on the edit, code in line 34 gets called again, a new object is being displayed in the front-end. 
5. Edit the cell, in my case I set to `T3` and click _Apply_, we can still see T2 but not T3 and a yellow popup box at the top.
after
 6. If we refresh, we can now see the update. (Having T3 and no T2)
refresh

wetneb:
It's funny, in my case the cell value gets updated immediately, no need to refresh. Tested on master with both Firefox and Chromium. So I still cannot reproduce the issue. But your fix works too, and I trust your screenshots, so let's merge this.

lisa761:
I have tried to reproduce this too, but I don't need to refresh it to see the applied changes.

tfmorris:
That's a great analysis!

> Note, after we click on the edit, code in line 34 gets called again, a new object is being displayed in the front-end.

I think that's actually the root cause of the problem and should be the focus of the fix. It appears to be an interaction with the text facets and I think that the text facets are (were) generating too many updates.

I was able to reproduce the problem using the commit immediately before this patch, but am not able to reproduce it on the current master. I suspect that the change that I made to the text filter facet in #2846 fixed the underlying issue that was triggering this behavior.

My theory is that clicking the Edit button caused an event in the text filter's input field which should have been ignored, but was causing the facets to trigger a refresh of the data grid. Now that we ignore events that don't change the value, this no longer happens.

I'm going to close this as unnecessary now, but I want to thank you for all the work you put into the analysis!

PR #2494, opened 2020-03-30
load correct format

james-cui:
Tested in the following steps
1 ./refine build
2 ./refine
3. Creat project by using http://api.worldbank.org/countries/all/indicators/SP.POP.TOTL?date=2000:2001
4. verify XML is loaded instead of line-based text

### DIFF
--- a/main/src/com/google/refine/importing/ImportingUtilities.java
+++ b/main/src/com/google/refine/importing/ImportingUtilities.java
@@ -770,6 +770,10 @@ public class ImportingUtilities {
         for (int i = 0; i < count; i++) {
             ObjectNode fileRecord = JSONUtilities.getObjectElement(fileRecords, i);
             String format = JSONUtilities.getString(fileRecord, "format", null);
+            if (format == null) {
+                format = guessFormatFromMimeType(JSONUtilities.getString(fileRecord, "declaredMimeType", null));
+            }
+
             if (format != null) {
                 if (formatToCount.containsKey(format)) {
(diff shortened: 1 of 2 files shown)


main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -846,6 +850,22 @@ public int compare(String o1, String o2) {
...
+        }
+
+        if (mimeType.contains("/xml")) {
+            return "text/xml";
+        }
+
+        if (mimeType.contains("/json")) {
+            return "text/json";
+        }
+
```
wetneb: It seems to me that this function is duplicating functionality that should already be implemented elsewhere, since we are already storing a mapping from MIME types to formats in `main/webapp/modules/core/MOD-INF/controller.js`.

Could the issue be solved simply by adding the relevant MIME types in `controller.js`? If patterns are necessary, such as `.*/xml.*`, we could consider adding support for patterns in this MIME type registry.

james-cui: @wetneb this is really a good idea. update it now

james-cui:
@wetneb updated the diff based on your comments, can you take a look once you have time?

james-cui:
tested using

Tested in the following steps
1 ./refine build
2 ./refine
3. Creat project by using http://api.worldbank.org/countries/all/indicators/SP.POP.TOTL?date=2000:2001
4. verify XML is loaded instead of line-based text

wetneb:
Closing as a duplicate of #2496.

Issue #2420, opened 2020-03-16
Importing an XML file from a URL does not auto detect the correct importer


thadguidry:
**Describe the bug**
XML importer is not selected by default

**To Reproduce**
1. **Create Project** and get data from Web Addresses (URLs)
2. Paste the following URL: 
 `http://api.worldbank.org/countries/all/indicators/SP.POP.TOTL?date=2000:2001`
3. Click **NEXT**
4. Notice that **Line-based text files** is the importer detected instead of **XML files**

**Current Results**
The Line-based importer is selected

**Expected behavior**
XML importer should be automatically selected for preview.  XML files are typically detected based on file extension such as `.xml`, but retrieving a file from a URL will sometimes need to resort to additional methods such as the following:

1. the response header `Content-Type` could be `application/xml` or `text/xml`
```http
HTTP/1.1 200 OK
Date: Mon, 16 Mar 2020 16:53:17 GMT
Content-Type: text/xml; charset=UTF-8
Content-Length: 13184
Connection: keep-alive
Set-Cookie: TS019266c8=017189f9470f813515282caf2315619d07a80cfb8e6a6cf9f09b164f0c7b7c86cca54015e18e7d6df654b2231545f75eee48d6eb6d; Path=/
Strict-Transport-Security: max-age=31536000; includeSubDomains
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
X-Frame-Options: deny
``` 

2. the first line of a file contains:
 ```xml
<?xml version=
```

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Version: Firefox
 - JRE or JDK Version: JDK8

**OpenRefine (please complete the following information):**
 - Version: OpenRefine (master) trunk

buntamas:
Hi ! I would like to solve this. Can I work on this issue?

thadguidry:
Sure @buntamas , let us know if you need assistance.  We'd also expect to see some tests added as well.

james-cui:
@thadguidry here is my pull request on the issue https://github.com/OpenRefine/OpenRefine/pull/2494
Can you review it?

PR #2496, opened 2020-03-30
Fixed that does not auto detect XML files from URL #2420

buntamas:
**Refers**: _Importing an XML file from a URL does not auto detect the correct importer_ #2420

I tested with a number of URLs, all of which worked.
The **content type** response header was: **application / xml** or **text / xml**.

### DIFF
--- a/main/webapp/modules/core/MOD-INF/controller.js
+++ b/main/webapp/modules/core/MOD-INF/controller.js
@@ -275,6 +275,8 @@ function registerImporting() {
   IM.registerMimeType("text/n3", "text/rdf/n3");
   IM.registerMimeType("text/rdf+n3", "text/rdf/n3");
   IM.registerMimeType("text/turtle", "text/rdf/ttl");
+  IM.registerMimeType("application/xml", "text/xml");
+  IM.registerMimeType("text/xml", "text/xml");
   IM.registerMimeType("application/rdf+xml", "text/rdf/xml");
   IM.registerMimeType("application/ld+json", "text/rdf/ld+json");
 


wetneb:
Nice! I am not sure which PR to merge though, since @james-cui opened another PR #2494 in the mean time (you were assigned to the issue, but it looks like this PR was influenced my review on  #2494). I do not want to be the arbiter of whose PR gets merged… It would be good to have a bit more communication to make sure efforts are not duplicated.

Can you both discuss and let me know if either @james-cui updates his PR, or we merge this?

thadguidry:
@wetneb While I appreciate @james-cui 's effort now and in the future, the issue was properly assigned and implemented by @buntamas.  Next time @james-cui needs to ensure that he checks issue assignment.

We don't want to get into an situation where we allow issue hijacking; that will not benefit our project, nor will would it be fair to all parties.  I would also say that if someone assigned doesn't respond after 1 week, then that might be a case for reassignment but never before.  We should probably mention something like that in our CONTRIBUTING.md guide; that we expect some loose responsiveness from contributors, otherwise they risk their issues being reassigned.

buntamas:
I was planning to finish the issue in this week, because last week I debug the problem, only I did not make commit. I saw interested in this issue and I wanted me to solve the problem because I am a university student and my subject matter will be judged on the basis of the solved issues.

james-cui:
@wetneb  @thadguidry @buntamas 
just saw this thread, for sure, I come after @buntamas, I was thinking no one was touching this for sometime and propose a fix. But since @buntamas has a diff, I am OK if his is get prioritized. 
Next time, should I ask permission to work for an issue before I go ahead? :)

main/webapp/modules/core/MOD-INF/controller.js
```diff
@@ -275,6 +275,8 @@ function registerImporting() {
   IM.registerMimeType("text/n3", "text/rdf/n3");
   IM.registerMimeType("text/rdf+n3", "text/rdf/n3");
   IM.registerMimeType("text/turtle", "text/rdf/ttl");
+  IM.registerMimeType("application/xml", "text/xml");
+  IM.registerMimeType("text/xml", "text/xml");
```
tfmorris: It would be good to also recognize the `+xml` suffix as a fallback when we don't have a more specific type, e.g. `application/atom+xml` but that would require introducing additional generality into the underlying code, so perhaps we can spin that off as a separate issue. We already support a few of the more specific ones such as `application/rdf+xml`.

antoine2711:
> Next time, should I ask permission to work for an issue before I go ahead? :)

I think, @james-cui, the best way to go is to first check if there is an « assignee » to the original request ([https://github.com/OpenRefine/OpenRefine/issues/2420](#2420)), then if someone is there, ask them if they are still working on the issue. If after a week, you get no answer, then you can go ahead. Otherwise, you could work in team with the original assignee?

That's the process I read in other threads about issue assignment. Of course, if no one is assigned, then you can put your name, and can even validate it with a senior developer here (Antonin, Thad, Owen or Tom).

Regards, Antoine

buntamas:
Should I do something more about this issue?

PR #2498, opened 2020-03-30
Support windows_dist & linux_dist for packaging

### LINKED ISSUE #2477: The packaging of just the Windows or Linux version doesn't work (./refine windows_dist | linux_dist)
**Describe the bug**
If you try to build the executable for Mac, you can use `./refine mac_dist "3.4-SNPASHOT"` or for all the platform, you can use `./refine dist "3.4-SNPASHOT"`.

The doc also says you can do the same for Windows and Linux:
```
   mac_dist <version> .................. Make MacOSX binary distribution
   windows_dist <version> .............. Make Windows binary distribution
   linux_dist <version> ................ Make Linux binary distribution
   dist <version> ...................... Make all distributions
```

But, these:
`./refine windows_dist "3.4-SNPASHOT"`
`./refine linux_dist "3.4-SNPASHOT"`
Do not work…

```
antoine$ ./refine windows_dist "3.4-SNPASHOT"
./refine: line 920: windows_dist: command not found

antoine$ ./refine linux_dist "3.4-SNPASHOT"
./refine: line 921: linux_dist: command not found
```
**To Reproduce**
1. Go to your code folder
2. Type: `./refine linux_dist "3.4-SNPASHOT"`
4. See error: « **./refine: line 921: linux_dist: command not found** »

**Current Results**
« **./refine: line 921: linux_dist: command not found** »

**Expected behavior**
Compile for that platform. Use profiles to invoke Maven with specific executions steps. (see below)

 - Version 3.3

**Additional context**
@wetneb wrote: « At the moment, it is not figured out how to invoke Maven for specific executions in the package stage »

@afkbrb wrote back: « We can use profiles: https://stackoverflow.com/questions/1791305/maven-assembly-plugin-run-only-one-descriptor. And I've made `windows_dist` and `linux_dist` work locally.

afkbrb:
Close #2477 

`./refine windows_dist xxx` and `./refine linux_dist xxx` work now.


wetneb:
There is *some* duplication of configuration here, although it's lightweight since linux and windows packages are completely described by external files. What if we wanted to do the same for `mac_dist`, would that imply duplicating the `appbundle` configuration both in the `default` and `mac` profiles? Is there no way around this?

afkbrb:
I guess we need something like "function" in pom.xml, so we can  "call the functions" for different profiles. I'll try to get it fixed.

afkbrb:
@wetneb I can't find a way to include other profiles in the default profile :(

We can discard the default profile and have three profiles for mac, linux, windows respectively. This way, we need to change `mvn package` in dist() function of `./refine` to `mvn package -Pmac,linux,windows`. But it seems better if `mvn package` can activate these three profiles together.

In the new commit I use pluginManagement to solve the problem. But this solution seems not good enough. Do you have any ideas?

thadguidry:
@afkbrb There are different types of Profiles.  You are dealing with just 1 at the moment, a Project. You are also missing the most important parts of Profile activation.  Specified by `-P`.  Within Maven settings you can list the `<activeProfiles>`.  The idea is that within any script or CI/CD, you can specify all the profiles you need for a particular build environment.   Please read through the first 10-20 paragraphs (ideally the whole page) of the [Maven - Intro to Profiles](https://maven.apache.org/guides/introduction/introduction-to-profiles.html)

afkbrb:
Thanks @thadguidry , I'll read it.

wetneb:
Using `<pluginManagement>` is a good idea, that works for me.

PR #2500, opened 2020-03-30
Remove unused imports & optimize them

thadguidry:
...because I'm sick of 100+ problems always being reported... in 2 different IDEs!


fpompermaier:
Good reason (y)

wetneb:
Fine in principle, but I would make sure it still compiles :)

thadguidry:
Oops!  Would help if I verified my classpath before I did this. :-)  Let me close this and clean that up.

PR #2501, opened 2020-03-30
Organize and remove unused imports

thadguidry:
...because I'm sick of 100+ problems in 2 different IDE's!!!

### DIFF
--- a/extensions/gdata/src/com/google/refine/extension/gdata/SpreadsheetSerializer.java
+++ b/extensions/gdata/src/com/google/refine/extension/gdata/SpreadsheetSerializer.java
@@ -2,7 +2,6 @@ package com.google.refine.extension.gdata;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
-import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.List;
 
--- a/extensions/gdata/tests/src/com/google/refine/extension/gdata/GoogleAPIExtensionTest.java
+++ b/extensions/gdata/tests/src/com/google/refine/extension/gdata/GoogleAPIExtensionTest.java
@@ -3,9 +3,6 @@ package com.google.refine.extension.gdata;
 import com.google.refine.ProjectManager;
 import org.testng.annotations.Test;
 
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 import static org.mockito.Mockito.mock;
 import static org.testng.Assert.*;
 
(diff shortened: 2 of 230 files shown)


tfmorris:
Do we have a coding standard which describes the correct order (and, preferably, tooling support for the 2 IDEs)?

It looks like the Eclipse settings for import order were removed in c812b6118c62d868517e36baf171d1fc14a62dd6. Is there a reason for that?

thadguidry:
These are useful references that I myself glanced over today (I myself use IntelliJ and VSC) :
https://google.github.io/styleguide/javaguide.html#s3.3-import-statements
https://stackoverflow.com/questions/54676137/how-to-make-vs-code-organize-java-imports-like-intellij

Hi @tfmorris During the Maven migration @wetneb had a problem with settings in Eclipse, if I recalled and found it easier to cleanup the Eclipse configs.  I was using VSC at the time, and didn't have similar issues.  I agree that a code style standard would help in the future, as we're discussing in that other issue, once we agree on one, like Google Java Style, or whatever.  This reorganization took effect under VSC while verifying also in IntelliJ.  The main use case for this PR was to reduce some of the noise in the IDE's which had grown much noisier over the last 3 years from my perspective.

wetneb:
> It looks like the Eclipse settings for import order were removed in c812b61. Is there a reason for that?

The intention was that people would just import the Maven configuration to configure their IDE. I should have been more careful not to get rid of style guidelines. 

I have no problem with restoring that but I think these things should not be specified in some IDE configuration file, since this should apply to all IDEs (so it should be enforced in the CI, for instance). I haven't checked if it is possible to specify that in the Maven configuration - if that is recognized IDEs, it would be great.

tfmorris:
I think the steps are:
1. Write down what we expect for a standard
1. Do a 1-time cleanup to match the standard
1. Enforce the standards using a CI static checker
1. Provide conforming IDE settings for those IDEs which support them and are popular with developers. There's no reason developers should need to think about such trivialities.

The original import order for the project was:

5=com.google.refine
4=com.google
3=com
2=org
1=javax
0=java

I'll add this to a code style wiki page as a starting point for discussion.

wetneb:
Thanks very much Tom for the initiative!

wetneb:
Since there are merge conflicts with this already, let's close this and redo it the way @tfmorris proposed: only do the one-time cleanup once when the relevant guidelines are in place, and set up CI to enforce it immediately after (ideally, in the same PR).

PR #2503, opened 2020-03-30
Replace deprecated method for LoadingCache

thadguidry:
@wetneb Do you think this is wise? To discard the key from the cache when not found? or am I assuming too much here? Need your help to clean this since I am finding it extremely hard to test the Wikidata stuff. It probably would be useful to have a doc that helps understand how to test the Wikidata extension properly, as you do.

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/utils/EntityCache.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/utils/EntityCache.java
@@ -23,7 +23,12 @@
  ******************************************************************************/
 package org.openrefine.wikidata.utils;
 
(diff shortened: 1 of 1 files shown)


wetneb:
It's kind of you to try to clean things up but as you noticed, one generally needs to understand the code to decide what to do about warnings raised by IDEs. Our Outreachy and GSoC students who will work on the Wikidata extension this summer will have a chance to work on the code and understand it in depth, so why don't you leave this to them?

The Wikidata extension is covered by much more tests than the rest of the code, so I would recommend you just run these tests.

wetneb:
I propose to close this soon to clear up the backlog

PR #2506, opened 2020-03-31
Created preferences to fix #2316

### LINKED ISSUE #2316: "Match other cells with same content" keeps appearing as the default choice

leoshrey:
1. Additionally, we can take a choice from users, whether to save their preference. Not implementing it based on https://github.com/OpenRefine/OpenRefine/issues/2316#issuecomment-600245240
2. Using cookies to store preference as of now. If needed, can move it to cache too.

### DIFF
--- a/main/webapp/modules/core/MOD-INF/controller.js
+++ b/main/webapp/modules/core/MOD-INF/controller.js
@@ -358,6 +358,7 @@ function init() {
       "scripts/util/encoding.js",
       "scripts/util/sign.js",
       "scripts/util/filter-lists.js",
+      "scripts/util/pref.js",
 
       "scripts/index.js",
       "scripts/index/create-project-ui.js",
@@ -447,6 +448,7 @@ function init() {
       "scripts/util/custom-suggest.js",
       "scripts/util/encoding.js",
       "scripts/util/sign.js",
+      "scripts/util/pref.js",
 
       "scripts/widgets/histogram-widget.js",
       "scripts/widgets/slider-widget.js",
--- a/main/webapp/modules/core/scripts/util/pref.js
+++ b/main/webapp/modules/core/scripts/util/pref.js
@@ -1,0 +1,43 @@
+/*
+This utility is used for storing user preferences to enhance user experience.
+Inspiration: https://github.com/OpenRefine/OpenRefine/issues/2316
+
+Further to fulfilling the above requirement, this utility can be extended to:
+1. Storing any user preference which saves rework from user's side,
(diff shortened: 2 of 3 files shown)


leoshrey:
Hi @wetneb ,
 My thought was more on the lines that the user can change it one time when he/she starts working on a project and it will be modified anyway.
While for the users who intend to have the same preference in their working style, it can be kept on the client side irrespective of sessions.

But I agree that having this specific flag on session level makes more sense to curb the issues coming with change of code of consumption.

My general intent was to work towards creating a preference manager sort of a thing, and for each preference, we could set the level of persistence (or no persistence), i.e. session, instance, user levels.
Also, storing the preference in the backend is more robust and would help extend the horizon, but it could be an overkill for a trivial flag like above.

I had commented the intentions in _pref.js_, however, I will make necessary changes and update the same. I'll also make the nomenclature changes to be more descriptive.

wetneb:
> Also, storing the preference in the backend is more robust and would help extend the horizon, but it could be an overkill for a trivial flag like above.

Are you aware that we already have backend-side storage preference? We already use it for things like this (such as the maximum number of values to store in a text facet).

leoshrey:
Yes I went through the same. I wanted to persist this flag on the browser earlier so I went ahead with this, do you think it should be moved to backend?

1) After your comment https://github.com/OpenRefine/OpenRefine/pull/2506#pullrequestreview-384681756, I was thinking of moving it to a browser session only.

2) In general, I'm planning to categorise preferences into Session Level, User Account Level, Universal. For the latter two, I will try and integrate with the existing store.

I will not include the 2) in this PR and take it separately if you think 1) is good enough.

wetneb:
1. I have no strong feelings about whether it should persist across sessions or not. If it should persist, I would use the existing preference store.

2. Improvements on the preferences system itself would be welcome - note that we don't even have a notion of user so far, so "User Account Level" preferences will be hard to introduce. Perhaps it is worth discussing these sort of changes on an issue or on the openrefine-dev mailing list first.

tfmorris:
I don't think this merits a preference setting. It seemed like the consensus from #2316 was to making the dialog's selection "sticky" across multiple invocations, which seems like a simple and straightforward solution to me.

tfmorris:
I'm going to close this, but @leoshrey please open another PR if you'd like to implement the suggested solution. Thank you for your efforts on this!

Issue #2511, opened 2020-04-02
Setup repository for product documentation


ostephens:
To better support OpenRefine we are undertaking a project to improve documentation. Documentation source files will live in a separate Github repo in the OpenRefine org, and from there be published to a documentation platform.

This issue is to setup the new repository

See https://docs.google.com/document/d/1VtzXjSqrliAp1R8Xy3yARXIiN5NuV0dNf4apvJz4rl8/edit# for more detail

kushthedude:
@wetneb @ostephens Didn't we agree to put the documentation in the OpenRefine Repository itself? And to publish the website we can use an outsourcing script or script to generate docs dynamically from the given URL. 

> To better support OpenRefine we are undertaking a project to improve documentation. Documentation source files will live in a separate Github repo in the OpenRefine org, and from there be published to a documentation platform.
>

ostephens:
@kushthedude for the technical reference documentation we'll definitely be doing this - via javadocs. I think for the product reference documentation we need a separate repo. I think we can still encourage contributors to update the docs as they change behaviour of the tool.

(definitions of technical reference vs product reference given at https://docs.google.com/document/d/1VtzXjSqrliAp1R8Xy3yARXIiN5NuV0dNf4apvJz4rl8/edit#)

wetneb:
Did we not also agree to store the source for the user manual in the main repository?

My understanding from the meeting we just had was that I would set up a repository to host GitHub Pages deployment for the docs, but the source of the docs would stay in the main repository.

tfmorris:
One advantage to having a separate repo is that permissions can be managed separately from the main source code repo.

Also, re:

> we can encourage contributors to update the docs as they change the behaviour of the tool

writing good quality code and writing good quality user documentation are two very different skill sets. It's not uncommon for people to have one or the other, but not both.

wetneb:
About permissions, I am not really sure why we would want to grant a user access to the docs and not the code or conversely. Changes to both are done through pull requests anyway, the master branch is protected. If someone becomes disruptive in any area we can easily remove them. We are currently trying to invite contributors to the organization as soon as they start contributing, which lets them contribute to issue triage and become part of the team. I think this works pretty well so far. In my experience there is little risk that people start abusing permissions and start messing up in an area they don't know about.

> writing good quality code and writing good quality user documentation are two very different skill sets. It's not uncommon for people to have one or the other, but not both.

That is true. I would actually expect two sorts of pull requests:
[...]

Issue #2273, opened 2020-01-06
Migrate to a new documentation platform


wetneb:
We want to get ourselves better docs than those we have on the GitHub wiki at the moment.

We need to decide which documentation system to use, where to host it, and so on. Let's use this issue to map the possibilities we are aware of. I will start:
* Sphinx: documentation system based on reStructuredText or Markdown, which can be hosted for free on ReadTheDocs. Example docs: https://editgroups.readthedocs.io/en/latest/
* Gitbook: based on Markdown, it looks like it can be hosted on gitbook.com. Example docs: https://devdocs.foodsharing.network/
* I really like Django's docs - not sure if the framework is reusable though: https://docs.djangoproject.com/en/3.0/topics/

I think our docs should be:
- [ ] versioned: one sub-site per version, ideally without duplicating content too much to keep things maintainable?
- [ ] translated: it should be easy for people who currently contribute on Weblate to also translate the docs.

Any other requirements we should have? Which existing docs should we take inspiration from?

thadguidry:
Here is my copy/paste from what I added into the EOSS 2020 1st Quarter Roadmap doc:

Doc Platform Evaluation criteria (a start) -
- Dynamic loader based, or Static generator ?  Examples of each would be:
  - https://docsify.now.sh/
  - https://v2.docusaurus.io/
-  Allows Semantic Tag versioning for documents, as well as tagging content sections by appropriate OpenRefine versions (3.1, 3.2, 3.3…) , this would allow filtering all new content tagged 3.3
- Full-text search (plugin or built-in) but often needs hosted to work ?
- Hosted or self/non-hosted support ? ([Gitbook no longer has CLI self-hosted](https://github.com/mozilla-l10n/localizer-documentation/issues/145))
- Multilingual & easy translation support ?
  - https://docs.readthedocs.io/en/stable/gsoc.html#improve-translation-workflow
- Math or [Mathjax](https://www.mathjax.org) support ?
- [Mermaid](https://mermaid-js.github.io/mermaid/#/) support for easy graphs ?
  - https://sphinxcontrib-mermaid-demo.readthedocs.io/en/latest/

thadguidry:
Hmm, there might be something else to say here about Sphinx:

> We could embed comment blocks (i.e., docstrings) inside the model code and use Sphinx to dynamically generate the model documentation by parsing these comments. This helps to ensure that the model documentation remains up-to-date.

fpompermaier:
Hi all,
[...]

PR #2513, opened 2020-04-03
Initial structure for Docusaurus docs

wetneb:
This is a prototype of reference manual using Docusaurus, for #2273. The generated docs can be [previewed here](http://pintoch.ulminfo.fr/e7c523cfa5a/).

The intention is to add a GitHub action to automatically build these docs and deploy them to a static host. I have created [OpenRefine/docs-prototype](https://github.com/OpenRefine/docs-prototype) (for #2511) to host docs at https://openrefine.github.io/docs-prototype/ (which is not set up yet), and later on with a cleaner domain such as `docs.openrefine.org`. As @kushthedude mentioned, we can consider other static hosts such as Netlify.

If we go for this set up, this is what the process should look like for contributors:
* the documentation can be edited directly through the Markdown source. GitHub embeds a markdown editor so they can preview their changes directly there.
* documentation changes are submitted as pull requests to this repository. We could set up a GitHub action to automatically tag PRs which only edit files within the `docs/` folder with the "documentation" tag.
* changes can be reviewed in GitHub using the markdown diffs. We could also set up CI to create previews of the full docs for each PR.

Contributors can also render the documentation locally (but should not have to):
```
cd docs
npm install
npm start
```
which serves the documentation at `http://localhost:3000/` and to build the static html, `npm run build`.

I have spent quite some time investigating how translation can be set up. My goal would be to add another Weblate component where documentation could be translated, using the same workflow as for UI text. I am confident we can set up something like this, but it is a significant piece of work and there is no point doing that until we actually have some stable content in the system.

Setting up versioning should be doable, using [the workflows recommended by Docusaurus](https://docusaurus.io/docs/en/versioning).


kushthedude:
Ok, so as I can understand from above. We are going to keep docs at OpenRefine Repository and the docusaurus, static site generator and the website-related code at https://github.com/OpenRefine/docs-prototype . Am I getting it right? @wetneb

wetneb:
In this set-up, the `docs-prototype` repository would only be used as a GitHub Pages host. If we use netlify, we would not need this repository at all. As you can see in this PR, all docusaurus-related files would still be in this repository.

docs/siteConfig.js
```diff
@@ -0,0 +1,81 @@
...
+  docsUrl: '',
+
+  /* path to images for header/footer */
+  headerIcon: 'img/openrefine_logo.png',
+  footerIcon: 'img/openrefine_logo.png',
+  favicon: 'img/openrefine_logo.png',
+
+  /* Colors for website */
+  colors: {
+    primaryColor: '#196581',
```
thadguidry, line 29: I'd would change the primary color to be one of the 2 within the inner facets of our new logo.
Annotation 2020-04-03 103147

#00a0ff  <-- a bit darker
#00c8ff  <-- a bit lighter

wetneb:
Yes we can change `docs/source` to `docs/src` or anything else.

We are not following the classic structure as it is designed to render both a website and the docs. Since we are only using Docusaurus for the docs, I tweaked the default structure to remove the unnecessary pages (see https://github.com/facebook/docusaurus/issues/459).

docs/siteConfig.js
wetneb: Hmmm, actually, even your dark one seems too agressive to me:
image

kushthedude:
> In this set-up, the `docs-prototype` repository would only be used as a GitHub Pages host. If we use netlify, we would not need this repository at all. As you can see in this PR, all docusaurus-related files would still be in this repository.

Oh got it. Are we trying netlify or we need more opinions?

docs/siteConfig.js
thadguidry: yeah I saw that also...and why I updated the comment to say "or even a darker shade...".  just turn it down.  Just increase the K value (black).  Perhaps `#0080cc` or even darker...whatever begins to make it look nicely contrasted.

wetneb:
> Oh got it. Are we trying netlify or we need more opinions?

Let's just wait to hear what others think, that should not have a huge impact on other decisions I think.

thadguidry:
Netlify simplifies deployments considerably for larger projects. (Vuejs.org uses them now also, I know)  But ours is not that large (yet) or complex.  I feel once we need some of those benefits of Netlify, than we can begin to take advantage of them.  In other words, I need to see the problems that we have with deployment with Docusaurus and GitHub before I'd venture into a conversation about Netlify.  We have enough things to worry about and learn for now.

kushthedude:
> I don't see any search capability. Would we need to use something like Google Custom Search?

We can use Lunrjs for this, it won't be a big deal for now.

wetneb:
> Have we decided on Docusaurus? As of a few days ago, new entries were still being added to the evaluation matrix and I haven't looked closely at all the candidates.

From the comments of @ostephens, @thadguidry and @joanneong on the issue, it seems to me that Docusaurus is the most popular choice. I would also be happy with it: my main reservation was the native use of Crowdin for translation, but manually integrating it with Weblate seems doable. But this is not set in stone - the purpose of this PR is to see if we like it. So do have a look at the candidates and let us know what you would go for.

> I don't see any search capability. Would we need to use something like Google Custom Search?

They integrate with [Algolia](https://docsearch.algolia.com/). I haven't set it up because the docs need to be hosted before they can be indexed. But I assume it should be easy to replace this by a Google Custom Search too.

> I think using a separate doc repo would minimize the cross talk noise from coders & writers and allow us to be more permissive with granting permissions to writers.

I think automatic tagging of PRs can address the cross talk noise, and I would prefer not to have to categorize contributors into rigid groups of documentation/code contributors (as I wrote in https://github.com/OpenRefine/OpenRefine/issues/2511#issuecomment-608087111).

Many large projects use a single repository approach and this seems to work quite well: [Spark](https://github.com/apache/spark), [Django](https://github.com/django/django/), [Weblate](https://github.com/WeblateOrg/weblate), [Matplotlib](https://github.com/matplotlib/matplotlib), [Flink](https://github.com/apache/flink), [PyTorch](https://github.com/pytorch/pytorch)…

docs/src/operations/key_value_columnize.md
```diff
@@ -0,0 +1,152 @@
+---
+id: key_value_columnize
+title: Columnize by key/value columns
```
thadguidry, line 3: technically, its "Columinze by key/value values"

docs/src/operations/key_value_columnize.md
```diff
@@ -0,0 +1,152 @@
...
+sidebar_label: Columnize by key/value
+---
+
+This operation can be used to reshape a table which contains *key* and *value* columns, such that the contents of the key column become new column names, and the contents of the value column are spread in the new columns. This operation can be invoked from
```
thadguidry, line 7: Nicely done explanation!

wetneb:
I have added an example of the sort of documentation I am thinking about for operations:
 * [Rendered version](http://pintoch.ulminfo.fr/e7c523cfa5a/operations/key_value_columnize.html)
* [Source](https://github.com/OpenRefine/OpenRefine/blob/3d6960200374783fc841cbfc1f8fe1bb72f1fedf/docs/src/operations/key_value_columnize.md)

I have of course no intention to start writing the documentation in this PR, this is just an opportunity to test the system on something a bit more realistic than a lorem ipsum, and see if it works as expected.

When writing this, I thought about the following guidelines:
* We are writing a reference manual, not a tutorial, but that does not mean we should not use examples. Unlike tutorials where trainers often demonstrate end-to-end workflows on real-world data, these examples should be carefully crafted to demonstrate only the operation at hand, in isolation, so they can be artificial.
* We should avoid including screenshots as much as possible. Screenshots become quickly outdated, they are heavy and hard to translate. Instead of including screenshots of tables in OpenRefine, include the table in Markdown itself. (Hence #2532 could be useful.) Of course, there are parts of the UI that are hard to describe without screenshots, so it should also be possible to use some sparingly.
* Any other guiding principles?

docs/src/operations/key_value_columnize.md
```diff
@@ -0,0 +1,152 @@
...
+| Q1727024    | Narcissus cyclamineus | Yellow   | 161899  |
+
+If extra columns do not contain identical values for all old rows spanning an entry, this can
+be fixed beforehand by using the [fill down operation](fill_down).
+
+Row order
+---------
+
+In the absence of extra columns, it is important to note that the order in which 
+the key/value pairs appear matters. Specifically, the operation will use the first key it encounters as delimiter for entries:
```
thadguidry, line 121:
```suggestion
the key/value pairs appear matters. Specifically, the operation will use the first key it encounters as the delimiter for entries:
```

thadguidry:
>Any other guiding principles?

You say the term "reference manual", is that the same as a "user guide" that other programs often have?
I think we need to lock down the terms of kinds of documentation that we are going to produce (or hire someone to produce).  Once we lock down those kinds, then I'll have a better understanding.

For instance, here is [Apache NiFi's User Guide](https://nifi.apache.org/docs.html), they have some screenshots explaining the feature areas.  They then have another Guide for "Expression Language Guide".  They also have individual Processor reference pages. 

For LibreOffice, the book guides are more tutorial written.  Their "reference manual" is actually their [Help Pages](https://help.libreoffice.org/latest/en-US/text/shared/05/new_help.html?&DbPAR=WRITER&System=WIN) (which is linked to from within the program itself)

I think what you have done so far, could serve as a Help Page for the "columnize as key/value" ? or would we need different documentation for that need?

In general, the term "reference" is really throwing me off, because I equate that to something like Apache NiFi's individual Processors docs.

Once we solidify the naming of the documentation kinds, then I can give feedback on "guiding principles" for each kind of documentation kind.

wetneb:
This would be a "product reference" according to Owen's classification in the planning document. Or "reference manual", that's a synonym for me. "User guide", "help page" also work. (I am personally not a huge fan of calling OpenRefine a "product", it sounds too commercial to me, that's why I use "reference manual", but that is clearly not a hill I will die on)

The details of what it should actually be are here:
https://docs.google.com/document/d/1C9Vm-oQcCYvyBZ9PtBIhybRfZc2pterf8H71L9P1M7s/edit

thadguidry:
Thanks!  Agree, so "reference manual" or colloquially, "user guide" can be what we use going forward!

tfmorris:
>> I think using a separate doc repo would minimize the cross talk noise from coders & writers and allow us to be more permissive with granting permissions to writers.

> I think automatic tagging of PRs can address the cross talk noise, and I would prefer not to have to categorize contributors into rigid groups of documentation/code contributors (as I wrote in https://github.com/OpenRefine/OpenRefine/issues/2511#issuecomment-608087111).

Sorry, I wasn't intentionally ignoring your other comment. I hadn't seen it yet when I wrote my note. While I'd prefer a separate repo, I'm OK with a single mongo repo if that's what everyone else wants.

> This would be a "product reference" according to Owen's classification in the planning document. Or "reference manual", that's a synonym for me.  "User guide", "help page" 

A "User Guide" is something very different from a "Reference Manual" in my mind. The "Getting Started" section of this prototype is the kind of thing I'd expect to see in a User Manual. A list of Operator descriptions is reference material.

> I am personally not a huge fan of calling OpenRefine a "product", it sounds too commercial to me, that's why I use "reference manual", but that is clearly not a hill I will die on.

The fact that OpenRefine is free doesn't make it any less of a product. Lots of folks depend on it being a high quality product with high quality documentation. Personally, I don't think the OpenRefine project should take an "anti-commerce" stance. I volunteer my time, but have, in the past, offered OpenRefine training, done custom OpenRefine development (open & closed source), done bespoke data wrangling, etc. I think there's room for both open source and commercial development as long as contributors are transparent and above board.

wetneb:
Yes, a "user guide" sounds a bit more like a tutorial to me. Happy to call this a "product reference" then!

(And yes, there is totally room for consulting activities and commercial providers around the tool - I think it's a great thing.)

ostephens:
So my view is:
* As far as we have a decision making process, Docusaurus was the preferred product based on #2273 - so we should go ahead with this. I specifically feel versioning is important here
* In terms of whether we have a separate repo for the documentation source - originally this is what I was thinking with #2511 but looking at this PR my feeling is we should move ahead with the documentation in the main repo and we can review once we have the basic documentation established - if we want to go with a separate repo at some point, then we can migrate easily I think

tfmorris:
That sounds good to me. If @ostephens is happy, I'm happy! :) Incremental progress is good.

tfmorris:
@thadguidry From a process point of view, if you re-review and approve the PR, it'll make Github happy that you're happy with @wetneb's changes to address your comments. Since he appeared to have addressed them, I just merged the PR using admin privs.

Issue #200, opened 2012-10-15
Accessing other rows than the current row


tfmorris:
_Original author: sanderse...@gmail.com (November 13, 2010 22:18:28)_

I've searched through the issues and all the documentation I could find and come up blank, sorry if this has been answered elsewhere.

Is there any way of accessing rows other than the current row? Something like rows[rowIndex-1].cells[&quot;Column1&quot;].value would be priceless. Unfortunately it doesn't work.

I'm looking for ways of computing moving averages etc and I need this feature. Thanks, and keep up the good work, this software looks great!

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=200_

wetneb:
For use cases like this, where it is not clear how to achieve the desired result via the records mode, I would be in favour of letting users run code (Python, R…) on the entire table, modifying it in arbitrary ways. It would avoid the need to export the project, run it through a script or some other tool, and import it back again.

We would need to expose the entire table in a suitable object in the guest language: pandas dataframe in Python, table in R…

thadguidry:
@wetneb or Apache Drill ;-)

wetneb:
Drill is a query engine so it's less clear to me what its integration as an operation would look like - would you use it to replace the table by the query results? Or would you try to add support for `INSERT` statements which modify the table? It'd be good to have concrete workflow examples to make sure we are on the same page.

(But I agree their integration of structured data in a columnar environment is clearly related)

thadguidry:
Supporting Selections (similar to how we allow users to select subdata various ways with GREL functions now, via parseJson() and parseHtml(), etc.)

I initially thought of "wouldn't it be cool if" somehow exposing a new `select()` and letting Apache Drill handling the heavylifting?  It would be useful to know how others view this issue however like @ettorerizza

`select()` could be exposed as a Data Model transform function via an inverse `insert()`.
And it doesn't have to be just at a GREL level...Our Expression editor UI could even be enhanced for support of this in a better fashion. (given time/money)

antoine2711:
> Is there any way of accessing rows other than the current row? Something like rows[rowIndex-1].cells["Column1"].value would be priceless. Unfortunately it doesn't work.

Hi @tfmorris, while reading this, it came to me that now that the cross() function has been enhanced, I think you could get `rows[rowIndex-1].cells["Column1"].value` this way:

1. Have a numerical key column where `value = row.index + 1`;
[...]

Issue #2461, opened 2020-03-23
GREL cross function should support lookup by numbers and other datatypes (date, boolean, etc.)


wetneb:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
The GREL `cross` function only works for string values at the moment. It could also support numbers (`int`, `long`). Looking up by floating point number is probably not a great idea because of rounding errors.

**Describe alternatives you've considered**
The `cross` function could directly convert values to string (in its index, and in its first argument).

**Additional context**
Suggested in #2456.

antoine2711:
I think that cross() should support any type (possible, easy) of value. Boolean would be pretty important, and date, time, datetime could also be useful.

Is there a big cost to doing them? I think that should guide our choice.

Regards, Antoine

wetneb:
The PR #2468 already allows for any cell value to be used in `cross`.

antoine2711:
@afkbrb: we have a bug in our code.

If I do this: 
`cross(cells.fakeIndex.value + 1, "TEST-CrossFunc_v3-4", "Col1_NumberInt")[0].cells.rowName.value`
![image](https://user-images.githubusercontent.com/12217525/79672865-90f74d00-81a3-11ea-8bb4-1f606568d0c5.png)
Everything is good. 

But if I do this:
`cross(cells.fakeIndex.value, "TEST-CrossFunc_v3-4", "Col1_NumberInt")[0].cells.rowName.value`
![image](https://user-images.githubusercontent.com/12217525/79672866-98b6f180-81a3-11ea-8178-de544c9bfd40.png)

In the last example, ONLY the first row should return `null`. All the other rows should return the same thing as the example 1 (it should only be offset by 1…). Can you please check this weird thing?

@thadguidry: since this is a bug in code that is in the master branch, but not in v3.3, should I reopen this Issue with this new problem, or should I create a new issue? (I'd rather not, it really concerns this Issue).

Regards, 
   Antoine

thadguidry:
I cannot comment on that bug.  `cross()` has now been changed from how it worked for the past 10 years.  Its overloaded and in my opinion is breaking how philosophy.  Let me know when you have something documented on how its supposed to work now, then I and others in the community will be happy to test...until then we cannot test something that isn't properly designed and documented yet.

thadguidry:
[...]

PR #2518, opened 2020-04-06
Support default project name and column name for GREL cross() function

### LINKED ISSUE #2504: Enhance cross() with empty "project name" calling parameter to represent the current project and empty column name to represent the index
The `cross()` function can get data from an external project, but also from the current project (kind of a self-join). In order to minimize the text to write (and read), if the project name parameter could be set to an empty string, it would help or visibility, and for reusability, has the SAME transformation formula can be used in 2 different projects that share some data structure (or on the same project that would have changed name since).

Also, in order to be able to do a cross with the index of the row, I would let empty value for the column name. In that case, the value (first parameter of `cross()`) should be of type number.

if `cross()` is called with an empty project name, then it should mean that the current project is the one to use. Ex.: `cross(1, "", "ID")[0]`.
if `cross()` is called with an empty column name, then the cross should be done not with a column, but with the internal index of the row, in a numerical match.

Writing the full column name, like now. Creating a column in the related project with the index as the value of each row.

afkbrb:
(Closes #2504) 

Can #200 be closed by this PR?

Currently, cross() can take 3 arguments: 
```
cross(cell or value, string projectName, string columnName)
```

In this PR, I added the following enhancements as mentioned in #2504 :

1. projectName will be optional and will be interpreted as the current project name if omitted or set to "".
2. columnName will also be optional and will be interpreted as the index column if omitted or set to "". Note that I don't mean a column with the name "index" by saying "the index column", it's actually a column with name "_OpenRefine_Index_Column_Name_", which is the inner name of the index column. See [Cross.java](https://github.com/OpenRefine/OpenRefine/pull/2518/commits/5d063f7ca965d1a8b8346fa528ed5a4bff0e687a#diff-d4f3fbf156f6686e6d9b4358766804edR50) for more details.

With these two enhancements, the cross() function will be more flexible.

We can use `cross(targetValue, "", targetColumn)` to lookup the given value in the given column of the current project.

We can use `cross(targetRowIndex)` to access another row in the current project. Similarly, we can use `cross(targetRowIndex, targetProjectName)` or `cross(targetRowIndex, targetProjectName, "")` to access a specified row in the target project. So this PR can be taken as a solution for #200 .

One interesting thing is that we already have an embedded "rowIndex" variable for GREL, so we can use `cross(rowIndex + offset)` to access a row relative to the current row. 

new

### DIFF
--- a/main/src/com/google/refine/expr/functions/Cross.java
+++ b/main/src/com/google/refine/expr/functions/Cross.java
@@ -41,24 +41,37 @@ import com.google.refine.expr.EvalError;
 import com.google.refine.expr.WrappedCell;
 import com.google.refine.grel.ControlFunctionRegistry;
 import com.google.refine.grel.Function;
+import com.google.refine.model.Project;
 import com.google.refine.util.GetProjectIDException;
 import com.google.refine.util.LookupException;
 
 public class Cross implements Function {
 
+    public static final String INDEX_COLUMN_NAME = "_OpenRefine_Index_Column_Name_";
+
     @Override
     public Object call(Properties bindings, Object[] args) {
-        if (args.length == 3) {
+        if (1 <= args.length && args.length <= 3) {
             // 1st argument can take either value or cell(for backward compatibility)
             Object v = args[0];
-            Object targetProjectName = args[1];
-            Object targetColumnName = args[2];
+            // if 2nd argument is omitted or set to "", use the current project name
+            Object targetProjectName = "";
+            boolean isCurrentProject = false;
+            if (args.length < 2 || args[1].equals("")) {
+                isCurrentProject = true;
+            } else {
+                targetProjectName = args[1];
+            }
+            // if 3rd argument is omitted or set to "", use the index column
+            Object targetColumnName = args.length < 3 || args[2].equals("") ? INDEX_COLUMN_NAME: args[2];
+
             long targetProjectID;
             ProjectLookup lookup;
 
             if (v != null && targetProjectName instanceof String && targetColumnName instanceof String) {
                 try {
-                    targetProjectID = ProjectManager.singleton.getProjectID((String) targetProjectName);
+                    targetProjectID = isCurrentProject ? ((Project) bindings.get("project")).id :
+                            ProjectManager.singleton.getProjectID((String) targetProjectName);
                 } catch (GetProjectIDException e) {
                     return new EvalError(e.getMessage());
                 }
@@ -76,17 +89,21 @@ public class Cross implements Function {
                 }
             }
         }
-        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects a cell or value, a project name to look up, and a column name in that project");
+
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects a cell or value, a project name to look up (optional), and a column name in that project (optional)");
     }
 
     @Override
     public String getDescription() {
-        return "Looks up the given value in the target column of the target project, returns an array of matched rows, cell will be interpreted as cell.value. Two values match if and only if they have the same string representation";
+        return "Looks up the given value in the target column of the target project, returns an array of matched rows. Two values match if and only if they have the same string representation. " +
+                "The first argument will be interpreted as cell.value if set to cell. " +
+                "The second argument will be interpreted as the current project name if omitted or set to \"\". " +
+                "The third argument will be interpreted as the index (starts from 0) column if omitted or set to \"\"";
     }
 
     @Override
     public String getParams() {
-        return "cell or value, string projectName, string columnName";
+        return "cell or value, string projectName (optional), string columnName (optional)";
     }
 
     @Override
--- a/main/tests/server/src/com/google/refine/expr/functions/CrossTests.java
+++ b/main/tests/server/src/com/google/refine/expr/functions/CrossTests.java
@@ -164,6 +164,49 @@ public class CrossTests extends RefineTest {
     }
 
     @Test
+    // lookup the row with index 0 in the current project
+    public void crossFunctionOneArgumentTest() throws Exception {
+        Row row = (((WrappedRow) ((HasFieldsListImpl) invoke("cross", 0)).get(0)).row);
+        String address = row.getCell(1).value.toString();
+        Assert.assertEquals(address, "mary");
+    }
+
+    @Test
+    public void crossFunctionOneArgumentTest1() throws Exception {
+        Row row = (((WrappedRow) ((HasFieldsListImpl) invoke("cross", 0, "")).get(0)).row);
+        String address = row.getCell(1).value.toString();
+        Assert.assertEquals(address, "mary");
+    }
+
+    @Test
+    public void crossFunctionOneArgumentTest2() throws Exception {
+        Row row = (((WrappedRow) ((HasFieldsListImpl) invoke("cross", 1, "", "")).get(0)).row);
+        String address = row.getCell(1).value.toString();
+        Assert.assertEquals(address, "john");
+    }
+
+    @Test
+    public void crossFunctionTwoArgumentTest() throws Exception {
+        Row row = (((WrappedRow) ((HasFieldsListImpl) invoke("cross", "lamp", "", "gift")).get(0)).row);
+        String address = row.getCell(1).value.toString();
+        Assert.assertEquals(address, "mary");
+    }
+
+    @Test
+    public void crossFunctionTwoArgumentTest1() throws Exception {
+        Row row = (((WrappedRow) ((HasFieldsListImpl) invoke("cross", 0, "My Address Book")).get(0)).row);
+        String address = row.getCell(1).value.toString();
+        Assert.assertEquals(address, "120 Main St.");
+    }
+
+    @Test
+    public void crossFunctionTwoArgumentTest2() throws Exception {
+        Row row = (((WrappedRow) ((HasFieldsListImpl) invoke("cross", 0, "My Address Book", "")).get(0)).row);
+        String address = row.getCell(1).value.toString();
+        Assert.assertEquals(address, "120 Main St.");
+    }
+
+    @Test
     public void crossFunctionOneToOneTest() throws Exception {
         Row row = (((WrappedRow) ((HasFieldsListImpl) invoke("cross", "mary", "My Address Book", "friend")).get(0)).row);
         String address = row.getCell(1).value.toString();
@@ -328,7 +371,7 @@ public class CrossTests extends RefineTest {
     @Test
     public void crossFunctionNonLiteralValue() throws Exception {
         Assert.assertEquals(((EvalError) invoke("cross", null, "My Address Book", "friend")).message,
-                "cross expects a cell or value, a project name to look up, and a column name in that project");
+                "cross expects a cell or value, a project name to look up (optional), and a column name in that project (optional)");
     }
     
     /**
@@ -346,10 +389,14 @@ public class CrossTests extends RefineTest {
             return function.call(bindings,args);
         }
     }
-    
+
     @Test
     public void serializeCross() {
-        String json = "{\"description\":\"Looks up the given value in the target column of the target project, returns an array of matched rows, cell will be interpreted as cell.value. Two values match if and only if they have the same string representation\",\"params\":\"cell or value, string projectName, string columnName\",\"returns\":\"array\"}";
+        String json = "{\"description\":\"Looks up the given value in the target column of the target project, returns an array of matched rows. Two values match if and only if they have the same string representation. " +
+                "The first argument will be interpreted as cell.value if set to cell. " +
+                "The second argument will be interpreted as the current project name if omitted or set to \\\"\\\". " +
+                "The third argument will be interpreted as the index (starts from 0) column if omitted or set to \\\"\\\"\"," +
+                "\"params\":\"cell or value, string projectName (optional), string columnName (optional)\",\"returns\":\"array\"}";
         TestUtils.isSerializedTo(new Cross(), json);
     }
 }
--- a/main/src/com/google/refine/LookupCacheManager.java
+++ b/main/src/com/google/refine/LookupCacheManager.java
@@ -3,6 +3,7 @@ package com.google.refine;
 import com.google.refine.expr.ExpressionUtils;
 import com.google.refine.expr.HasFieldsListImpl;
 import com.google.refine.expr.WrappedRow;
+import com.google.refine.expr.functions.Cross;
 import com.google.refine.model.Column;
 import com.google.refine.model.Project;
 import com.google.refine.model.Row;
@@ -74,6 +75,14 @@ public class LookupCacheManager {
         ProjectMetadata targetProjectMetadata = ProjectManager.singleton.getProjectMetadata(lookup.targetProjectID);
         if (targetProject == null) {
             return;
+        }
+
+        // if this is a lookup on the index column
+        if (lookup.targetColumnName.equals(Cross.INDEX_COLUMN_NAME)) {
+            for (int r = 0; r < targetProject.rows.size(); r++) {
+                lookup.valueToRowIndices.put(String.valueOf(r) , Collections.singletonList(r));
+            }
+            return; // return directly
         }
 
         Column targetColumn = targetProject.columnModel.getColumnByName(lookup.targetColumnName);


antoine2711:
@thadguidry: when you have the opportunity, give a look at this. We await your input.

Regards, A.

thadguidry:
@antoine2711 checking it out now....stay tuned...

main/src/com/google/refine/expr/functions/Cross.java
```diff
@@ -41,18 +41,25 @@
...
     public Object call(Properties bindings, Object[] args) {
-        if (args.length == 3) {
+        if (args.length == 1 || args.length == 2 || args.length == 3) {
             // 1st argument can take either value or cell(for backward compatibility)
             Object v = args[0];
-            Object targetProjectName = args[1];
-            Object targetColumnName = args[2];
+            // if 2nd argument is omitted or set to "", use the current project name
+            Object targetProjectName = args.length < 2 || args[1].equals("") ?
+                    ((Project) bindings.get("project")).getMetadata().getName() : args[1];
```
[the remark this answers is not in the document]
afkbrb: Got it!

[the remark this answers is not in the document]
antoine2711: > Project's can actually have the same name. I have 25 projects with unique Project IDs but they all have a common name "clipboard". And that's fine, we allow that. `cross()` however, without a Project name now supplied here will get confused as shown, so we need to improve this handling (in other words, it should know that I am about to perform a `cross()` on the current `binding` to a Project id.

@thadguidry: I know we are not keen on overloading function parameters, but, in this case, I really think we should accept number to identify project. Because, it's just 2 ways of identifying the exact same thing.

What do you think about that?

Regards, A.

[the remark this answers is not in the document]
thadguidry: @antoine2711 Users can be explicit about the project they want to cross with by specifying name.  If the project names are the same, it is trivial for them to rename a project and then be explicit with the names.  So I'd say let's not introduce extra worry about project ids in `cross()` for now.

thadguidry:
@afkbrb Thanks...checked out...reviewing now.

main/src/com/google/refine/expr/functions/Cross.java
```diff
@@ -71,22 +84,27 @@ public Object call(Properties bindings, Object[] args) {
...
-        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects a cell or cell value, a project name to look up, and a column name in that project");
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects a cell or cell value, a project name to look up (optional), and a column name in that project (optional)");
     }
 
     @Override
     public String getDescription() {
-        return "Looks up the given value in the target column of the target project, returns an array of matched rows, cell will be interpreted as cell.value";
+        return "Looks up the given value in the target column of the target project, returns an array of matched rows. " +
+                "The first argument will be interpreted as cell.value if set to cell. " +
+                "The second argument will be interpreted as the current project name if omitted or set to ''. " +
```
thadguidry, line 101: @afkbrb Looks like we need to clean up the help text just a bit?  missing the single quote chars `''` on the end of those 2 sentences?  Looks like some kind of rendering issue?

image

thadguidry: What if we flipped this around and instead just got smarter with inspecting the argument values themselves?

GIVEN Only 1 argument
WHEN argument represents a valid column name in current project
THEN  perform the cross with against valid column name in current project

Instead of this unneeded verbosity...

image

We make it easier with simpler syntax because we know a few things already from argument inspection (where `cell` and `project` are not specified)

image

thadguidry:
What if we flipped this around and instead just got smarter with inspecting the argument values themselves?

GIVEN Only 1 argument
WHEN argument represents a valid column name in current project
THEN  perform the cross against valid column name in current project

Instead of this unneeded verbosity...

image

We make it easier with simpler syntax because we know a few things already from argument inspection (where `cell` and `project` are not specified)

image

main/src/com/google/refine/expr/functions/Cross.java
```diff
@@ -41,24 +41,37 @@
...
+            // if 2nd argument is omitted or set to "", use the current project name
+            Object targetProjectName = "";
+            boolean isCurrentProject = false;
+            if (args.length < 2 || args[1].equals("")) {
+                isCurrentProject = true;
+            } else {
+                targetProjectName = args[1];
+            }
+            // if 3rd argument is omitted or set to "", use the index column
+            Object targetColumnName = args.length < 3 || args[2].equals("") ? INDEX_COLUMN_NAME: args[2];
```
thadguidry, line 66: Hmm...  what are we missing here?

I perform a `cross()` that has valid syntax and 3 args.
But there's no try wrapping here...so... silent error?
My expectation was that of some kind of error GIVEN the below scenario:

image

antoine2711: I don’t think it’s an error.

Indexes being numerical, you got to match them with numbers.

Regards, A.

thadguidry: Sorry @antoine2711 I don't quite understand what you mean.  Please explain what I am asking GREL for with that syntax in my image and given the underlying data grid.

antoine2711: > Sorry @antoine2711 I don't quite understand what you mean. Please explain what I am asking GREL for with that syntax in my image and given the underlying data grid.

You are doing `cross(cell.value, "CurrentProject", Use the internal index)`.

Since internal index is ALWAYS a number, cell.value MUST be a number to match & cross. In your example, 111 or 222 « could be » numbers, but since the preview doesn't use the green coloring to confirm it, can't know for sure, but at the look of the grey table in the background, neither 111 or 222 or 333 seams to be cast as number, but rather like strings, which will never match an index.

But, even if 111 or 222 where numbers, you would need 112 or 223 rows in your project to have a match.

Is it clearer? Regards, A.

main/src/com/google/refine/expr/functions/Cross.java
```diff
@@ -71,22 +84,27 @@ public Object call(Properties bindings, Object[] args) {
...
+                } else if (v instanceof Number && targetColumnName.equals(INDEX_COLUMN_NAME)){
+                    return lookup.getRows(((Number) v).intValue());
                 } else {
                     return lookup.getRows(v);
                 }
             }
         }
-        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects a cell or cell value, a project name to look up, and a column name in that project");
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects a cell or cell value, a project name to look up (optional), and a column name in that project (optional)");
     }
```
antoine2711, line 95: I would change `cell value` for just value everywhere in the doc. It's just a value, it can come from anywhere (even literals values are accepted!).

main/src/com/google/refine/expr/functions/Cross.java
```diff
@@ -71,22 +84,27 @@ public Object call(Properties bindings, Object[] args) {
...
+                "The first argument will be interpreted as cell.value if set to cell. " +
+                "The second argument will be interpreted as the current project name if omitted or set to \"\". " +
+                "The third argument will be interpreted as the index (starts from 0) column if omitted or set to \"\"";
     }
 
     @Override
     public String getParams() {
-        return "cell c or object value, string projectName, string columnName";
+        return "cell c or object value, string projectName (optional), string columnName (optional)";
     }
```
antoine2711, line 108: I would change `object value` for just value everywhere in the doc. It's just a value, it can come from anywhere (even literals values are accepted!). Also, the use of object is misleading, not pertinent and confusing.

main/tests/server/src/com/google/refine/expr/functions/CrossTests.java
```diff
@@ -263,12 +306,12 @@ public void crossFunctionMatchNotFoundTest() throws Exception {
      
     /**
      *  
-     *  rest of cells shows "Error: cross expects a cell or cell value, a project name to look up, and a column name in that project"
+     *  rest of cells shows "Error: cross expects a cell or cell value, a project name to look up (optional), and a column name in that project (optional)"
      */
```
antoine2711, line 310: Same.

main/tests/server/src/com/google/refine/expr/functions/CrossTests.java
```diff
@@ -263,12 +306,12 @@ public void crossFunctionMatchNotFoundTest() throws Exception {
...
     @Test
     public void crossFunctionNonLiteralValue() throws Exception {
         Assert.assertEquals(((EvalError) invoke("cross", null, "My Address Book", "friend")).message,
-                "cross expects a cell or cell value, a project name to look up, and a column name in that project");
+                "cross expects a cell or cell value, a project name to look up (optional), and a column name in that project (optional)");
     }
```
antoine2711, line 315: Same

main/tests/server/src/com/google/refine/expr/functions/CrossTests.java
```diff
@@ -289,7 +332,11 @@ private static Object invoke(String name,Object... args) {
     
     @Test
     public void serializeCross() {
-        String json = "{\"description\":\"Looks up the given value in the target column of the target project, returns an array of matched rows, cell will be interpreted as cell.value\",\"params\":\"cell c or object value, string projectName, string columnName\",\"returns\":\"array\"}";
+        String json = "{\"description\":\"Looks up the given value in the target column of the target project, returns an array of matched rows. " +
+                "The first argument will be interpreted as cell.value if set to cell. " +
+                "The second argument will be interpreted as the current project name if omitted or set to \\\"\\\". " +
+                "The third argument will be interpreted as the index (starts from 0) column if omitted or set to \\\"\\\"\"," +
+                "\"params\":\"cell c or object value, string projectName (optional), string columnName (optional)\",\"returns\":\"array\"}";
         TestUtils.isSerializedTo(new Cross(), json);
```
antoine2711, line 340: Same.

main/src/com/google/refine/expr/functions/Cross.java
thadguidry: they are Strings (not numbers)

antoine2711: > they are Strings (not numbers)

K. So, unless they are numerical and between 0 and « Max(row.index) », you can always expect `null` result value, and not an error.

Regards, A.

thadguidry: No, it's much more confusing now.  I need a test case (a fuller test case) that shows how the 3rd argument works or doesn't.  I'd expect that single test case to show a single cross operation that is using the 3rd argument working against a test data grid of 3 rows x 2 columns.

I was also confused by the issue #2504 and still am, so that's part of the problem....me actually seeing some usefulness...which I have not yet, so I need everyone to help me here.

thadguidry: @antoine2711 Would you mind helping everyone understand how the new `cross()` is supposed to work on a new wiki page or some Google document or Hackmd.io so that the community can leave comments?  Call it "Proposed Cross() in Issue #2504"  The issue #2504 is not clear to me, sorry.  Thanks.

antoine2711: > No, it's much more confusing now. 
Your confusion seams to come from GREL cross function should support lookup by numbers #2461, which is committed into the master branch since almost a month.

Now, when we use the `cross()` function, matches can be done even if the datatype is not a string. 

> I need a test case (a fuller test case) that shows how the 3rd argument works or doesn't. I'd expect that single test case to show a single cross operation that is using the 3rd argument working against a test data grid of 3 rows x 2 columns.

I think you are raising a good point, that #2461 could have more tests.
What should we add to this file? `main/tests/server/src/com/google/refine/expr/functions/CrossTests.java`

> I was also confused by the issue #2504 and still am, so that's part of the problem....me actually seeing some usefulness...which I have not yet, so I need everyone to help me here.
The real added value to Issue #2504 is to be able to do cross() with the internal index of rows, that is not a column, but can easily behave like one. It is very powerful as it actually enables every project to be used like an index array. I basically makes it possible to access every cells of every projects. It's very flexible.

Should we discuss of datatypes with the cross() function in Issue #2461 or here? It much more relevent there.

Regards, Antoine

antoine2711: > @antoine2711 Would you mind helping everyone understand how the new `cross()` is supposed to work on some Google document or Hackmd.io so that the community can leave comments? Call it "Proposed Cross() in Issue #2504" The issue #2504 is not clear to me, sorry. Thanks.

Yes, but like I wrote, it would be about Issue #2461. The situation you have here is the same you would have with an added column called ID that would contain numeric values of index (`row.index`).

This is the first thing I was planning to do on the new documentation system. And it's important to have that ready BEFORE v3.4 goes out, since it already committed in the master branch. I could go in the Wiki, and start a new page, that could be exposed once v3.4 is out.

Regards, 
   Antoine

thadguidry: The community can't leave comments on a GitHub Wiki page, unfortunately.  Please use a collaborative environment.  My suggestion is https://hackmd.io

main/src/com/google/refine/expr/functions/Cross.java
afkbrb: @antoine2711 Good point! I'll change it. The "jsonize()" function also uses "value" , so I think it's ok to do the same here.

jsonize

thadguidry: `value` is loosely defined here: https://github.com/OpenRefine/OpenRefine/wiki/Variables#variables

We probably should give a small mention of `value` somewhere here:
https://github.com/OpenRefine/OpenRefine/wiki/Server-Side-Architecture#data-model
>raw data: a list of rows, each row consisting of a list of cells

antoine2711: In the previous version of `cross()` even a value was somehow invisibly « bounded » to the column it came from. So, to name it « object value », or « cell value », had a certain logic in it. But now, it's really a pure value, that can be anything. So, might as well just call it that, a value.

@thadguidry: yah, we have a few definition here and there, so a nice clean up in our new documentation will surely come handy. my favorite is datatype definitions in the Cell page… But hey! At one point, it made a lot of sens, I guess. ;-)

I'm eager to see what's coming in that way.

Regards, Antoine.

wetneb:
Clearing this milestone as we only add issues, not PRs, to milestones.

afkbrb:
Can I close this PR? It has conflicts with #2571 . I'd like to make the required changes in a new branch based on #2571 .

wetneb:
Sure, thanks for taking care of this!

tfmorris:
It's possible to rebase a branch and force push it, rather than starting over. That makes it easier to maintain continuity of the discussion.

What is the new PR/branch that includes this functionality (if there is one)?

antoine2711:
> What is the new PR/branch that includes this functionality (if there is one)?

I think that this PR is close, and that the code has been moved to #2571 and merge there to master.

Regards, Antoine

wetneb:
No #2571 was different (and merged earlier)

afkbrb:
I haven't used git rebase before. I'll learn about it and continue on this branch based on the new changes in #2571 .

afkbrb:
Seems the rebase worked. Now this PR is based on the latest master branch.

tfmorris:
Great! Thank you.

main/src/com/google/refine/expr/functions/Cross.java
afkbrb: Hi, @thadguidry , with #2571, we are now using string match for the cross() function now. So the result of your test above will be:

test1

Here the index range is [0, 9], so in column "#@!c", only the value "3" got a match.

But there is another problem here: 

Both the embedded GREL variable "rowIndex" and the index used by the cross() function start from 0, so "cross(rowIndex)" will match the same column, which is what we expected. 

test3


However the row index in the UI starts from 1. This is confusing, because "cross(3)" actually match the 4th row.

test2

My suggestion is that we change the "rowIndex" and the index used by the cross() function both to start from 1, so they will be consistent with the UI index. But I am not sure if changing the "rowIndex" variable will result in some unexpected behaviour of other GREL functions.

antoine2711:
@afkbrb: K, I'm sorry. I did something (c4b088b), but I don't know what exactly. If you want to delete that, or tell me how to do it. Very sorry.

For the rest, I did a relatively complete test of the code, and it's working good. Great job. I would just put back the `cell or value` in the docs, and then I guess it would be finished.

Regards,
   Antoine

tfmorris:
I got rid of the stray merge commit, dropped the commit that removed the mention of "cell or value" so that it's back to the original and rebased on the current master.

Once the CI tests run, it should be good to merge now that we've cut the 3.4 branch.

main/src/com/google/refine/expr/functions/Cross.java
antoine2711: > My suggestion is that we change the "rowIndex" and the index used by the cross() function both to start from 1, so they will be consistent with the UI index. But I am not sure if changing the "rowIndex" variable will result in some unexpected behaviour of other GREL functions. What's your opinion on this?

@afkbrb: rowIndex is 0-bound everywhere, except in the UI in the preview and the data-grid. I think it would be less breaking and complicated to just change the UI display, and make it 0-bound! ;-)

I'm joking, but I'm also serious. It would be more coherent.

Regards,
   Antoine

tfmorris: This has been reviewed and approved. We're not making breaking changes to things. Programmers and computers expect 0-based. Humans expect 1-based. Thus we have two different domains.

Issue #2197, opened 2019-10-15
Derive default enter-separator from previous split-cells 


katrinleinweber:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
I'm always frustrated when the ["Join multi-valued cells" prompt suggests `,`](https://github.com/OpenRefine/OpenRefine/blob/9d72de1874726be82e458335deef7cd4aee0302f/main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js#L119-L120) to me, even though:

1. I used a different symbol in the previous ["Split multi-valued cells" operation](https://github.com/OpenRefine/OpenRefine/blob/9d72de1874726be82e458335deef7cd4aee0302f/main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js#L288-L323), and
2. the `,` is present at least in some of the still split cells.

**Describe the solution you'd like**
Would both operations be more convenient & less error-prone if OpenRefine:

a) derived the default character from 1. or at least,
b) not suggested a symbol for which 2. was true?

**Describe alternatives you've considered**
I'm not sure.

**Additional context**
Related to #1113 & #2139.

wetneb:
For 1., we could consider storing the last value used for the Join/Split operations in the preferences, and propose that as default value in both dialogs.

ostephens:
Either using the last value as the default or being able to set a default value in preferences would be approaches that worked for me (I always have to change this, and it's always a pain!)

Resolving issue (2) is more problematic - I don't think we can predict what the user wants to do here (maybe they specifically do want to use that separator even though it is in the cells already) and checking all values in the column for a separator character is likely to lead to performance problems for large projects.

So my vote would be to put (2) aside for the moment and focus on solving (1)

wetneb:
Given @ostephens' comment above let's say that point 2. is out of scope because heuristics are likely to be flaky. So this is closed by #2520.

PR #2520, opened 2020-04-06
Adds a default separator preference for split/join multi valued cells

lisa761:
Closes part 1 of #2197

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js
+++ b/main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js
@@ -117,6 +117,8 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
   };
 
   var doJoinMultiValueCells = function(separator) {
+    var defaultValue = Refine.getPreference("defaultSeparatorValue", ",");
+    var separator = window.prompt($.i18n('core-views/enter-separator'), defaultValue);
     if (separator !== null) {
       Refine.postCoreProcess(
         "join-multi-value-cells",
@@ -128,7 +130,7 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
         null,
         { rowsChanged: true }
       );
-      setDefaultSeparatorValuePreference(separator);
+      Refine.setPreference("defaultSeparatorValue", separator);
     }
   };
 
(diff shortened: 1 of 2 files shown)


thadguidry:
@wetneb Hmm, I don't know.  This might actually end up being a bit useful as a permanent preference depending on your daily life in your domain.  I'm thinking of subfields in archival formats in GLAM.  Examples of separator pattern they might want to store: `##$a` or `##$z`

lisa761:
Can we ask for some more opinions then?

wetneb:
To merge this, I think it would be worth using the new JS interface to access the preferences introduced by @antoine2711. I am not opposed to storing it in the preferences (for the sake of getting out of this stalemate).

antoine2711:
> To merge this, I think it would be worth using the new JS interface to access the preferences introduced by @antoine2711. I am not opposed to storing it in the preferences (for the sake of getting out of this stalemate).

@lisa761 & @wetneb: I can confirm that #2624 implements these two JS functions and is now merged in the master branch (v3.5):
```
Refine.getPreference(key, defaultValue) => return a string
Refine.setPreference(key, newValue)
```

Regards,
   Antoine

wetneb:
@tfmorris any thoughts about the preference key to use for this? (should we prefix it like the others?)

tfmorris:
Yes, a prefix would be good. It provides a simple namespacing mechanism. Our naming scheme is kind of a mess, but let's try to not make it worse.

How about something along the lines of `ui.edit.defaultSeparator`? Do we need to qualify separator any more in this context (ie will we have multiple types of editing default separators)?

Here's a sampling of what we have now (along with suggested cleaner names, but it might be too late to rename things):

```
cell-ui.previewMatchedCells => ui.cell.previewMatchedCells
exporters.templating.template
reconciliation.standardServices
ui.gridPaginationSize => ui.browsing.pageSize
ui.browsing.listFacet.limit
userMetadata

```

lisa761:
@tfmorris There's also a split/join operation for columns. But we don't have a preference setting for it's default value yet.

ostephens:
I think `ui.cell` better than `ui.edit.cell`

Anticipating that we might have a preference for split/join columns at some point I'd suggest we are specific with this naming, so suggest something like:

`ui.cell.rowSplitDefaultSeparator`

This allows us to have `ui.cell.columnSplitDefaultSeparator` in future if we want. I think having these separate is a good idea because when you are doing data manipulations that uses both operations together you need different separators for each way you are splitting

Issue #80, opened 2012-10-15
Is it possible to to add a row of data ?


tfmorris:
_Original author: premato...@gmail.com (June 21, 2010 00:50:20)_

<b>What steps will reproduce the problem?</b>
<b>1.</b>
<b>2.</b>
<b>3.</b>

<b>What is the expected output? What do you see instead?</b>

&gt;&gt;&gt;&gt;&gt;&gt; No where to add a row of data to a project once in Gridworks

<b>What version of the product are you using? On what operating system?</b>

&gt;&gt;&gt;&gt;&gt;&gt; Version 1.1-r878

<b>Please provide any additional information below.</b>

We are in the process of cleaning up some data base data. The data base in question is connected to a Live UI and is updated manually, so needs to be kept up to data with new data that comes in while we clean up the original data. We also need to add that new data to the corresponding Grid works project so that when we load the cleaned up data back into the data base, it's up to date. 

Our question is : Is it possible to add data to a Gridworks project or do we need to export the project to csv, add the new records, then import it back into the data base? 

Thank you

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=80_

PR #2522, opened 2020-04-06
FIX: #2521 Replaced Project Delete to use red color text instead of blue 'X'

### LINKED ISSUE #2521: Change Project Delete to use red color text instead of blue 'X' png
Users are not seeing how to delete a Project in OpenRefine.
User tweeting about this:
https://twitter.com/ruthbrarian/status/1243243600223391745
>Took me forever to see the little Xs to delete them too

We also use currently use differing styles for Actions to perform on a Project:
* Img
* Text

Make the style consistent.  Change the style to use red text instead:
Before change:
```
<td><a class="delete-project" title="Delete this project" href=""><img src="images/close.png"></a></td>
```
After change:
```
<td><a class="delete-project" title="Delete this project" href="" style="color: red;">Delete</a></td>
```

Do nothing and confuse users more.

**Additional context**
Before change:
image
After change:
image

AghaSaad04:

Fix #2521 
A minor change of replacing the Blue 'X' to Red-colored Delete text as it was confusing users

### DIFF
--- a/main/webapp/modules/core/scripts/index/open-project-ui.js
+++ b/main/webapp/modules/core/scripts/index/open-project-ui.js
@@ -218,7 +218,8 @@ Refine.OpenProjectUI.prototype._renderProjects = function(data) {
       .addClass("delete-project")
       .attr("title",$.i18n('core-index-open/del-title'))
       .attr("href","")
-      .html("<img src='images/close.png' />")
+      .css("color","red")
+      .html("Delete")
       .click(function() {
         if (window.confirm($.i18n('core-index-open/del-body') + project.name + "\"?")) {
           Refine.postCSRF(


AghaSaad04:
Should I change .html("Delete") to .text("Delete") ?

thadguidry:
@AghaSaad04 No idea.  You need to see what the implications are for translation or styling this element...then make the best decision.

AghaSaad04:
**.html() is faster!**

The .html() function is only a wrapper to .innerHTML, but in the .text() function jQuery adds an "entity filter", and this filter naturally consumes time.
.html() treats the string as HTML while .text() treats the content as text

Reference:
https://stackoverflow.com/questions/1910794/what-is-the-difference-between-jquery-text-and-html

thadguidry:
It's not always about speed.  But sometimes about accessibility and usability!  Can a red trashcan image allow longer text to be displayed along Project rows, and still make it easy to spot for users?  Think about the original problem... a user couldn't see how to delete a project because the blue colored X was stuck at the beginning of the row and likely just looked like some bulleted element along with all the other blue | grey lines, etc.  Step back from the UI, zone out on it...what pops out quickly to your eyes?  what feels less contrasting?  UI/UX is more than just color, it's sometimes about shape, and sometimes providing lots of white spacing or pushing a bunch of crowed interaction elements into ... a hamburger menu.

How would Delete and About work if moved into a hamburger menu?  Would would be gained? What would be lost?

antoine2711:
Hi @thadguidry & @wetneb, 

  I read this issue, and I was also baffled at first with this « X », and I didn't use it the first times. Now, I know it there, so I used it.

   I also read the comment on the red color being aggressive, and I also share that opinion. 

   I'm wondering: could a balanced solution be to have « delete », written in black, with the trash can on the left? That way, I think all angles would be covered.

Regards, Antoine

AghaSaad04:
Oh sorry @thadguidry I miss interpreted that. For better usability, I was thinking of red trash can icon?

It would look something like this:
image

Suggestions are welcome

wetneb:
I think we should pick a color that is consistent with the existing palette used in the UI. If the trash can icon is clear enough, this does not need to be in a flashy color.

thadguidry:
The original problem for users was that they could not see the blue X because it was "TOO CONSISTENT" as well as using symbology that is not as universal , so we need to be careful there.

I love colors.  We should begin introducing a bit more color to OpenRefine where it makes sense for our users.  We all need to get used to these changes.

For now, I'd suggesting using a vibrant blue shade, so that it is highly contrasting with our subtle blue color harmonies we have now.
`#1134A6` - Egyptian
`#0080FE` - Azure
`#0F52BA` - Sapphire
`#000080` - Navy

or combining blue and red:
`#6F2DA8` - Grape
`#B43757` - Hibiscus
`#B5338A` - Fandango
`#710193` - Violet
`#67032F` - Jam

When James Home (and Dae Park) from Google was originally helping us with our UI/UX design he would provide color harmonies, he would give us some previews..., taking screenshots, throwing the images into a single long document attached so we can review and then we vote on the best looks.

wetneb:
I propose to close this PR due to the lack of consensus around the icon's appearance.

tfmorris:
I agree. Apologies to @AghaSaad04 for the confusion, but thank you for pitching in to help.

Let's move the discussion to the issue #2521 and get to consensus there.

Issue #2451, opened 2020-03-21
Columns named « toString » are displayed as collapsed


antoine2711:
**Describe the bug**
Using the name « toString » for a column creates a column with no name and not popup arrow. You must use « Re-order / remove columns... » to clean that up…

**To Reproduce**
1. Go to 'Add column based on this column...'
2. Enter the text for the column name to « toString »
3. See the error in the table view.

**Current Results**
Unusable column.

**Expected behavior**
Seeing a normal column named « toString »

**Screenshots**
Command:
![image](https://user-images.githubusercontent.com/12217525/77234992-b3884d00-6b88-11ea-97ad-c79a225aa5e3.png)

Result (see the column after column « ValDiv »:
![image](https://user-images.githubusercontent.com/12217525/77235005-ca2ea400-6b88-11ea-8acc-fa564eee263f.png)

**OpenRefine (please complete the following information):**
 - Version 3.3

antoine2711:
@wetneb Hi Antonin, I created this issue, but I still can't apply labels, unlike you said. 

Regards, Antoine

wetneb:
That's my fault, I did not add you to the relevant team. It should be fixed now.

james-cui:
@wetneb and @antoine2711 , can I work on this issue?

antoine2711:
> @wetneb and @antoine2711 , can I work on this issue?

Sure James, be my guest. Don't hesitate if you have questions.

Regards, Antoine

james-cui:
@antoine2711  here is the pull request https://github.com/OpenRefine/OpenRefine/pull/2531

james-cui:
@antoine2711  @wetneb   just updated the pull request

wetneb:
Thanks @antoine2711 for discovering and reporting this!

PR #2531, opened 2020-04-09
Fix that columns named « toString » are displayed as collapsed (#2451)

james-cui:
1) ./refine build to build

2) ./refine to run an instance

3) add column by using existing column

4) verify column header and content are added correctly

Root cause is there are default properties in a empty dict of javascript
we have to enhance the element existence by checking the value type

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
+++ b/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
@@ -229,7 +229,7 @@ DataTableView.prototype._renderDataTables = function(table, headerTable) {
       for (var c = 0; c < columns.length; c++) {
         var column = columns[c];
         var td = tr.insertCell(tr.cells.length);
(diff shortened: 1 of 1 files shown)


antoine2711:
Hi @james-cui: I see your modification.

I wonder, do you think it's possible to fix it before?
We have this: `typeof(self._collapsedColumnNames[column.name]) !== 'function'`

But, instead of checking AFTER that it's a function, could the title of the column be stored in `self._collapsedColumnNames[column.name]` as a **string**, rather that a **function**?

One of the reason I'm asking is because we are moving towards Vue.js for our main grid, so this patch could have to be done again in the new code. But if we clean this before so it a normal string, we might not have this problem…

Regards, Antoine

james-cui:
@wetneb improved based on code review comments, can you take a look?

antoine2711:
@james-cui: good job and thank you. Nice fix.

Regards, Antoine.

Issue #2516, opened 2020-04-04
Creating a project from a URL: URLs with leading whitespace are rejected


wetneb:
**Describe the bug**

The input field to create a project from a URL does not trim its input when validating it.

**To Reproduce**
1. Go to 'Create Project' -> 'Web Addresses (URL)'
2. Paste "  http://api.worldbank.org/countries/all/indicators/SP.POP.TOTL?date=2000:2001" with the leading whitespace
3. Validate
4. The URL is rejected

**Expected behavior**

The value of this input should be trimmed before validation.

**OpenRefine (please complete the following information):**
 - master branch (3.4 snapshot)

Nishtha3512:
![Screenshot from 2020-04-04 14-36-17](https://user-images.githubusercontent.com/51858166/78423107-c8cda300-7681-11ea-9517-37e59309b3ac.png)
Hey,Where is the option to validate?
As you can see in above screenshot when i click on create project then project is simply created without rejecting any URL.

wetneb:
Are you sure you included whitespace before the URL?
![image](https://user-images.githubusercontent.com/309908/78423384-408dd480-7666-11ea-8481-4e99a9f57284.png)

Nishtha3512:
Ohh now i understand your issue.Can you please tell me where  is code related to issue?I will try to fix it. 
Is it in OpenRefine/main/src/com/google/refine/importers/XmlImporter.java file?

wetneb:
No, it is a UI issue, so it is in a Javascript file.

Here is how you can find it:
 * Search for the error message "You must specify a valid web address" in the source tree: `grep -R "You must specify a valid web address" .`
 * You will find this message in the transations: `"core-index-import/warning-web-address": "You must specify a valid web address (URL) to import."`
* Now search for the identifier of the message: `grep -R "core-index-import/warning-web-address" .`.
* The only Javascript file that contains this is `./main/webapp/modules/core/scripts/index/default-importing-sources/sources.js`, so the error is coming from there.
* Open this file and look for `warning-web-address`.

With this method you do not need to understand OpenRefine's structure much: you can locate the code you are interested in just by following references, starting from UI text.

kushthedude:
I think we can just add a function in UI to trim the whitespaces in the input URL before it is sent to the Regex for validation, or can we just add a trim function in the regex function?

Nishtha3512:
[...]

PR #2534, opened 2020-04-11
Fixes issue #2516 trim-input-url-whitespace

Nishtha3512:


main/webapp/modules/core/scripts/index/default-importing-sources/sources.js
```diff
@@ -91,7 +91,17 @@ UrlImportingSourceUI.prototype.attachUI = function(bodyDiv) {
...
+      validUrl = true;
+    } else {
+      importUrl = self._elmts.urlInput[0].value.trim();
+      self._elmts.urlInput[0].value = importUrl;
+    }  
+    
+    if(!isUrlValid(importUrl)) {
       window.alert($.i18n('core-index-import/warning-web-address'));
     } else {
       self._controller.startImportJob(self._elmts.form, $.i18n('core-index-import/downloading-data'));
```
wetneb, line 107: Make sure you also trim the URL in `startImportJob`.

main/webapp/modules/core/scripts/index/default-importing-sources/sources.js
```diff
@@ -91,7 +91,17 @@ UrlImportingSourceUI.prototype.attachUI = function(bodyDiv) {
...
-    if(!isUrlValid(self._elmts.urlInput[0].value)) {
+    var importUrl = self._elmts.urlInput[0].value;
+    var validUrl =  False;
+    
+    if(isUrlValid(importUrl)) {
+      validUrl = true;
+    } else {
+      importUrl = self._elmts.urlInput[0].value.trim();
+      self._elmts.urlInput[0].value = importUrl;
+    }
```
wetneb, line 102: I think this code could be a lot simpler. You can trim the value no matter if it is valid initially or not - trimming whitespace around a URL will never hurt.

wetneb:
@antoine2711 it's a good point that we should check for a valid URL not just in the first input, but all of them. It could be worth opening an issue for that.

antoine2711:
> @antoine2711 it's a good point that we should check for a valid URL not just in the first input, but all of them. It could be worth opening an issue for that.

checking the other input fields is a point I was raising, but I also have that question: where can I (or what the good way to) retrieve the value of the other fields?

My first intuition is using our binded variable, `self._elmts.urlInput[1].value` (like what's used else where), but with an incremented index (1 instead of 0). But it raise an error, as the other fields don't seam to be binded with our variable. We can clearly see that with `self._elmts.form.elements`, we can get to those input text field, but is it a « clean » way? Shouldn't they also be binded in the addButton.click() event handler?

 ```
 this._elmts.addButton.click(function(evt) {
    self._elmts.buttons.before(self._elmts.urlRow.clone());
  });
```

Here my variables just after I click the « Next » button.
image

I know my question might be totally beside the reality, I'm not familiar with the binding mechanism we use in OR. Is it from jQuery?

Regards, Antoine

**Comment by wetneb** [comment]
Yes the mechanism to obtain elements using their `bind` id does not follow updates, so if a form input is added later on by the user we are not able to access it with this mechanism. I think we have to resort to standard jQuery here.

**Comment by antoine2711** [comment]
> Yes the mechanism to obtain elements using their `bind` id does not follow updates, so if a form input is added later on by the user we are not able to access it with this mechanism. I think we have to resort to standard jQuery here.

```
this._elmts.addButton.click(function(evt) {
   self._elmts.buttons.before(self._elmts.urlRow.clone());

    inputElements = [];
    elementCount = self._elmts.form.context.elements.length
    
    for (var elementIndex = 0; elementIndex < elementCount; elementIndex++) {
      var element = self._elmts.form.context.elements[elementIndex];
      
      if(element.localName.toLowerCase() == "input") {
        inputElements.push(element);
      }
    }
 });
```

Could something like this work for rebinding?

Regards, A.

**Comment by wetneb** [comment]
I would find it cleaner to use an appropriate jQuery selector, rather than things like `element.localName.toLowerCase() == "input"` which are quite low level.

**Comment by tfmorris** [comment]
@antoine2711 please don't hijack PRs for random questions. Let's keep the conversation focused on the code review.

@wetneb were all your comments resolved? It looks to me like this (important) line got dropped:

       self._elmts.urlInput[0].value = importUrl;

so that the trimmed value isn't actually going to be used by the importer.

**Comment by antoine2711** [comment]
> @antoine2711 please don't hijack PRs for random questions. Let's keep the conversation focused on the code review.

I don't see highjacking here. All the URL will have to be trimmed, not just `self._elmts.urlInput[0].value`. Despite all his good will, @Nishtha3512 was only trimming the first one. (For the record, I was also having email with him to help him with this issue.)

It is logical to want to manipulate this from the code, not the DOM. `urlInput` is bound to `self._elmts.urlInput[0]`, but [self._elmts.urlInput[1]](url) is bound to nothing.

But, maybe there is something I need to learn or don't understand, and if it's the case, I'm listening.

Antoine

**Comment by wetneb** [comment]
@tfmorris I think you might have seen just one commit out of the two. The complete change looks good to me although it does not address the point that only the first URL is checked for errors, but that's arguably a different issue.

File: main/webapp/modules/core/scripts/index/default-importing-sources/sources.js [hunk 3]
```diff
@@ -91,7 +91,9 @@ UrlImportingSourceUI.prototype.attachUI = function(bodyDiv) {
 
   this._elmts.form.submit(function(evt){
     evt.preventDefault();
-    if(!isUrlValid(self._elmts.urlInput[0].value)) {
+    var importUrl = self._elmts.urlInput[0].value.trim(); 
+
```
tfmorris, line 95:
```suggestion
    self._elmts.urlInput[0].value = importUrl;
```

PR #2535, opened 2020-04-11
Fix TSV mediatype & add support for atom+xml and .ntriples

thadguidry:


main/webapp/modules/core/MOD-INF/controller.js
```diff
@@ -250,6 +251,8 @@ function registerImporting() {
   IM.registerExtension(".ods", "text/xml/ods");
   
   IM.registerExtension(".nt", "text/rdf/nt");
+  IM.registerExtension(".ntriples", "text/rdf/nt");
```
thadguidry, line 254: @wetneb this is because of stupid DBpedia.  
image

main/webapp/modules/core/MOD-INF/controller.js
```diff
@@ -266,7 +269,7 @@ function registerImporting() {
   IM.registerMimeType("text/plain", "text/line-based");
   IM.registerMimeType("text/csv", "text/line-based/*sv");
   IM.registerMimeType("text/x-csv", "text/line-based/*sv");
-  IM.registerMimeType("text/tab-separated-value", "text/line-based/*sv");
+  IM.registerMimeType("text/tab-separated-values", "text/line-based/*sv");
```
thadguidry, line 272: IANA confirmed at https://www.iana.org/assignments/media-types/text/tab-separated-values

main/webapp/modules/core/MOD-INF/controller.js
```diff
@@ -278,6 +281,7 @@ function registerImporting() {
   IM.registerMimeType("text/xml", "text/xml");
   IM.registerMimeType("application/rdf+xml", "text/rdf/xml");
   IM.registerMimeType("application/ld+json", "text/rdf/ld+json");
+  IM.registerMimeType("application/atom+xml", "text/xml");
```
thadguidry, line 284: I actually have some feeds aggregated using Mozilla FF addon's and would like this to be easier to just feed OpenRefine the URL's and then I can just quickly XML box select the Entry Elements in OpenRefine for sanitizing and reconciling within some domains.

main/webapp/modules/core/MOD-INF/controller.js
wetneb: It's probably useful to keep the singular version around too: if it was added, it might be in use somewhere. There is no harm in having both.

PR #2539, opened 2020-04-12
chore: Cleanup sorting controllers

kushthedude:
- We are not using sortingControllers in View-Panel Header.


kushthedude:
I double checked it, there is no use for this neither the UI is rendered
for this.

wetneb:
Click on any column, and sort it by any criterion. This works on the master branch and breaks with your change.

kushthedude:
I didnt know we had sort on click too.

wetneb:
I am not sure how you double checked that this was dead code but please use a different method in the future.

kushthedude:
> I am not sure how you double checked that this was dead code but please use a different method in the future.

@wetneb I overlooked the following conditional 
```js 
if (this._sorting.criteria.length > 0) {		
     this._renderSortingControls(elmts.sortingControls);		
   }
``` 
& presumed there was no call to renderSortingControls. Hence made the PR.

PR #2540, opened 2020-04-12
fix: looksLikeURL Function catches wrong URLs

### LINKED ISSUE #2519: Hyperlink is created for cell value even when invalid and text datatype is choosen
**Describe the bug**
Hyperlink is created for cell value, even when text is specified as the datatype.

**To Reproduce**
1. Edit a cell's value
2. Choose datatype: Text
3. Paste the following text exactly (includes whitespace):
```
```

4. Click Apply

**Current Results**
Notice that cell value is shown completely as a clickable hyperlink rather than just text.

**Expected behavior**
The cell's value should be rendered with only those pieces which represent valid URLs hyperlinked. For the example above, there should be four hyperlinks with unlinked text " are now mapped to " and " via " between them.

Browsers now have native support for URL detection and we should take advantage of that.  See comment https://github.com/OpenRefine/OpenRefine/issues/2519#issuecomment-612645821

**Screenshots**
image

 - OS: Windows 10
 - Browser Version: Firefox
 - JRE or JDK Version: JDK8

 - Version: Trunk (master)

**Additional Context**
The use case that I actually had here in #2519 was that of a cell of text that might contain multiple URLs would be convenient to be able to click on the URLs and it open a browser for quick verification during reconciling processes.

As long as scrolling the grid still works, editing the cells still works, running GREL operations on the cell text, and exporting the cells still works, then I think it satisfies most of the needs that will come up within GLAM reconciling processes with multiple URLs in a cell text for clicking and verifying.

kushthedude:
Fixes #2519

### DIFF
--- a/main/webapp/modules/core/scripts/util/url.js
+++ b/main/webapp/modules/core/scripts/util/url.js
@@ -90,15 +90,7 @@ URL.getParameters = function() {
 };
 
 URL.looksLikeUrl = function(s) {
-  if (s.length > URL.minSchemeLength + 1) {
-    var sep = s.substring(0, URL.maxSchemeLength + 3).indexOf(":");
-    if (sep >= URL.minSchemeLength) {
-      var scheme = s.substring(0, sep).toLowerCase();
-      if (scheme in URL.schemes) {
-        return (URL.schemes[scheme] === 0) || (s.substring(sep + 1, sep + 3) == "//");
-      }
-    }
-  }
-  return false;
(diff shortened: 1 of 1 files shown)


kushthedude:
> If browser support is not a problem, using `new URL(s)` as suggested by Thad sounds like a cleaner approach than relying on a complicated regular expression.

Noted, I will look into it !

tfmorris:
We also already have this function: https://github.com/OpenRefine/OpenRefine/blob/5ddf4d6278ed5386e786dfc8538b0df0af467463/main/webapp/modules/core/scripts/index/default-importing-sources/sources.js#L39-L43

The URL constructor browser support matrix is here: https://developer.mozilla.org/en-US/docs/Web/API/URL
Internet Explorer isn't included, but I suspect we don't care. Do we have a supported browser list?

wetneb:
> Do we have a supported browser list?

As far as I know, all we have is https://github.com/OpenRefine/OpenRefine/wiki/Installation-Instructions#recommended-browsers for now.

tfmorris:
I've added a note that Internet Explorer is unsupported to the wiki, to make it official, so we can go ahead and use the URL() constructor.

tfmorris:
@kushthedude Are you still planning to update this PR?

wetneb:
Probably not, since the fork was deleted, therefore closing this.

PR #2541, opened 2020-04-12
Docs: Add GitHub Actions CI Badge in ReadME

kushthedude:
#### Screenshot :
Screenshot 2020-04-12 at 10 30 54 AM


wetneb:
We are not really using GitHub as CI, we just use it to publish the snapshots, so for me it's redundant with Travis.

kushthedude:
> We are not really using GitHub as CI, we just use it to publish the snapshots, so for me it's redundant with Travis.

Cool!

Issue #2514, opened 2020-04-03
set focus on find-field for replace command


LibrErli:
**Enhancement Suggestion**
If you click on `Edit Cells > Replace` it would be very great if the cursor focus will be set automatically to the input field for the `find`-string (`id#text-to-find-..`.)

**Environment**
* OpenRefine 3.3
* Firefox 74
* Ubuntu 18.04

kovacskarolina:
Hello, I think I've solved this issue. Can you assign it to me?

HeroicHitesh:
@LibrErli Does this issue need further fixes or is it resolved? I see there's a merged PR, so if the issue is resolved by @kovacskarolina PR kindly close this issue to avoid confusion.

wetneb:
Thanks for catching this, this one should have been closed indeed.

PR #2545, opened 2020-04-13
Fixed to focus on find-field

kovacskarolina:
Refers: set focus on find-field for replace command #2514

Added a function to focus on input field


kovacskarolina:
I've corrected it according to your suggestions, @wetneb!

wetneb:
I think we can even avoid the `frame.ready` by calling the focus method just after the dialog is shown:
https://github.com/OpenRefine/OpenRefine/blob/8e68c39bba80485a1b7a45b61cce383766872cc7/main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js#L247

kovacskarolina:
I removed that part. Sorry, I think I've overcomplicated this issue.

PR #2546, opened 2020-04-13
Add our Org attributes to parent POM

thadguidry:
and full details on scm

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
   <description>OpenRefine is a free, open source power tool for working with messy data and improving it</description>
   <url>http://openrefine.org/</url>
   <organization>
-    <name>Code for Science & Society Inc.</name>
+    <name>Code for Science &amp; Society Inc.</name>
     <url>https://codeforscience.org/</url>
   </organization>
   <licenses>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -10,6 +10,38 @@
   <name>OpenRefine</name>
   <description>OpenRefine is a free, open source power tool for working with messy data and improving it</description>
   <url>http://openrefine.org/</url>
+  <organization>
+    <name>OpenRefine</name>
+    <url>http://openrefine.org/</url>
```
wetneb, line 15: Perhaps the organization should be CS&S instead?

thadguidry: Dunno.  This statement could be anything:
>Most projects are run by some sort of organization (business, private group, etc.). Here is where the most basic information is set.

I think the key word is "run".  CS&S is a fiscal sponsor, they don't "run" OpenRefine project.

examples: https://github.com/search?l=&p=2&q=%3Corganization%3E+extension%3Apom&ref=advsearch&type=Code

Some have foundations:
```xml
  <organization>
    <name>Apache Software Foundation</name>
    <url>http://groovy-lang.org</url>
  </organization>
```
Others have the project:
```xml
      <organization>
        <name>XmlBeans</name>
        <url>http://xmlbeans.apache.org/</url>
    </organization>
```

wetneb: Since we already have the project name and URL in the main fields for that, I think it'd be nice to put CS&S there - it would make the metadata a bit richer

thadguidry: agreed

PR #2550, opened 2020-04-15
Update perform-edits-dialog.js to trap [RETURN] in editSummary and proceed to action

antoine2711:
Adding PerformEditsDialog.doFormSubmit() to share with keypress() [RETURN] from editField and « Edit Uploads ».

Questions:
* Should I move PerformEditsDialog.doFormSubmit() inside PerformEditsDialog.launch()?
* Should I also trap [RETURN] when the focus is not in `this._elmts.editSummary`?

Regards, Antoine

### DIFF
--- a/extensions/wikidata/module/scripts/dialogs/perform-edits-dialog.js
+++ b/extensions/wikidata/module/scripts/dialogs/perform-edits-dialog.js
@@ -31,21 +31,16 @@ PerformEditsDialog.launch = function(logged_in_username, max_severity) {
     formCopy.submit();
 
     if(elmts.editSummary.val().length == 0) {
-        elmts.editSummary.focus();
+      elmts.editSummary.focus();
     } else {
-        Refine.postProcess(
+      Refine.postProcess(
         "wikidata",
         "perform-wikibase-edits",
         {},
-        {
-            summary: elmts.editSummary.val(),
-        },
+        { summary: elmts.editSummary.val(), },
         { includeEngine: true, cellsChanged: true, columnStatsChanged: true },
-        { onDone:
-            function() {
-            dismiss();
-            }
-        });
+        { onDone: function() { dismiss(); }
+      });
     }
   }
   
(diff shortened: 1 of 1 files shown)


antoine2711:
> Why do we need to catch the enter key manually here? Is it not enough to configure the form to be submitted with `doFormSubmit`?

Before I added this trapping of the return, a simple return just closes the windows without doing any action.

I though the best course of action was the trap this event, and do the same as the submit.

This is clean, reuse the code, and is logical. It might not be the best fix, but I bet it will hold fine.

That being said, if you want me to code it another way, tell me what to do.

Regards,
   Antoine

antoine2711:
> Why do we need to catch the enter key manually here? Is it not enough to configure the form to be submitted with `doFormSubmit`?

Before I added this trapping of the return, a simple return just closes the windows without doing any action.

I though the best course of action was the trap this event, and do the same as the submit.

This is clean, reuse the code, and is logical. It might not be the best fix, but I bet it will hold fine.

That being said, if you want me to code it another way, tell me what to do.

Regards,
   Antoine

antoine2711:
> Why do we need to catch the enter key manually here? Is it not enough to configure the form to be submitted with `doFormSubmit`?

I relooked into it, and the way I see it, the submit mechanism WAS already active, but, for a reason or another, I didn't processed. I remember 20 year ago, I would check whether a submit was sent thru [RETURN] or the submit button by the fact the [RETURN] didn't put in the param list the pair submit=submit, there would just be no submit… It it the case now? I don't know. But this way of coding it just makes sure we follow the same code when doing [RETURN] or pressing the « Edit Wikidata » button.

Regards, Antoine

extensions/wikidata/module/scripts/dialogs/perform-edits-dialog.js
```diff
@@ -30,34 +57,18 @@ PerformEditsDialog.launch = function(logged_in_username, max_severity) {
      dismiss();
   });
 
-  var hiddenIframe = $('#hiddenIframe').contents();
+  this._elmts.editSummary.keypress(function (evt) {
+    evt.preventDefault();
```
wetneb, line 61: you should not call `evt.preventDefault()` here since it blocks all key presses in the form. Test your PR: try adding a summary in the field, your key presses will not be recorded.

PR #2555, opened 2020-04-16
Issue #2306 - null value fix in tables

linderbarbara:
URL of the original issue:
https://github.com/OpenRefine/OpenRefine/issues/2306
Description of the original issue:
"I'm trying to export with « Custom tabular exporter… » and choosing « Upload to -> A new Google spreadsheet », and in some rows, if a value is missing, then the cell value disapear and the cell on the right shift left."


thadguidry:
@linderbarbara Hi and thanks for your first PR !   It looks like you accidentally deleted our `.gitignore` file.  Please add that back in.  Also I noticed that our `.gitignore` is missing the exclusion for Eclipse `.metadata` folder so I just issued a PR #2556 to fix that.  If you could please cleanup this PR (add our `.gitignore` back in place, and ensure this PR does not include pushing the `.metadata` Eclipse folder) then we should be ready for further review!

wetneb:
New version of this PR is in #2560.

PR #2556, opened 2020-04-16
Add .metadata to ignore Eclipse plugins meta

thadguidry:
I noticed on PR #2555 that we should have ignored the Eclipse plugins ,metadata folder


thadguidry:
Merging this in quick so we don't foul up more PR's from corporate Eclipse users.

PR #2561, opened 2020-04-17
2306 null value fix in tables

linderbarbara:
Bugfix and related changes


wetneb:
Hi @linderbarbara, there is still the same issue. Please seek git help for this :)

PR #2564, opened 2020-04-18
feat: Move to Yarn Package Manager

kushthedude:
- Most of the packages were fixed to a specific version rather they should not be fixed at a particular and should be specified (>= version number which is done by `^` sign) hence most of this discrepancies occurred in the first place.

Refers #2273


thadguidry:
@kushthedude Disagree.  Modification of dependency versions can introduce undesirable changes and offers NO GUARANTEE.  That's why they are fixed to a version.  That is the actual intent of locked packages.  The file `package-lock.json` is generated.  If you feel that there is some regression or bug introduced by some specific versions of these packages, then it should be resolved with the author of that particular package, who is then responsible for his packages' dependency tree.

Please read thoroughly why NPM package lock files exist, what they are used for and why  https://docs.npmjs.com/configuring-npm/package-locks.html
https://docs.npmjs.com/configuring-npm/package-lock-json

kushthedude:
@thadguidry As soon as you will set up the docs using `npm i` same changes will get staged in your local. Also in` lock.json` peer dependencies specify a minimum compatible version, they don't lock the packages to a fixed version as because dependency upgrading tools like Dependabot, Renovate-bot may upgrade the dependencies. If the lock.json dependencies are fixed to a particular version, not only there will be conflicts on local when using `npm install` but we would also need to upgrade the new packages manually.


> Actually I am not sure package-lock.js should be included in git - looking at other projects which use Docusaurus, they don't seem to include it. So I would just delete this: the Docusaurus version is already pinned in `package.json` anyway.

@wetneb Package-lock.json should be included in git because it keeps the track of peer-dependencies of the devDependencies specified in `package.json`. Usually, we ignore `node_modules` in our node-based project so that no packages get staged into our repository.

More info here : 
https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md
https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

wetneb:
Yes I understand why it generally makes sense to include `package-lock.json` for npm projects, but in our case this is a documentation project: we are not developing our own code on top of existing libraries. We are just using npm to call the Docusaurus build process: Docusaurus is the only dependency. In this context I am not sure we care about peer dependencies: we only care about Docusaurus' version, and that is pinned in `package.json`.

Otherwise, how do you explain that the following repositories do not include `package-lock.json` in their repositories:
* https://github.com/facebook/docusaurus
* https://github.com/facebook/create-react-app/tree/master/docusaurus/website
* https://github.com/facebookincubator/profilo/tree/master/website
* https://github.com/facebook/jest/tree/master/website

kushthedude:
@wetned
That is because all of those repository are of facebook and facebook is
using its own node engine called yarn. Therefore you would see yarn.lock in
those Repositories at place of package-lock.json .
Yarn is quite faster than npm and is created by facebook.
Nonetheless there is no difference in yarn.lock and package-lock.json
except that both are created by difference node engine yarn and npm
respectively.

kushthedude:
Even we can use yarn engine. Only benefit yarn has over npm is yarn caches the packages so that the installation time is significantally lower than npm engine.

wetneb:
> Therefore you would see yarn.lock in those Repositories at place of package-lock.json

Well, I don't see any `yarn.lock` either! So I don't think this is due to them using Yarn rather than NPM.

kushthedude:
> > Therefore you would see yarn.lock in those Repositories at place of package-lock.json
> 
> Well, I don't see any `yarn.lock` either! So I don't think this is due to them using Yarn rather than NPM.

Here are all -:
https://github.com/facebook/docusaurus/blob/master/yarn.lock
https://github.com/facebook/create-react-app/blob/master/docusaurus/website/yarn.lock
https://github.com/facebookincubator/profilo/blob/master/website/yarn.lock
https://github.com/facebook/jest/blob/master/yarn.lock
All of the four repositories you mentioned have yarn.lock respectively.

kushthedude:
Also one more thing, Can we change the netlify domain? I think openrefine-docs is right now available, the current one is quite inappropriate (musing-fermat-1fee39) 😅 .

wetneb:
Ok thanks for the yarn.lock examples, not sure why I did not see them.

Yes, I have changed the netlify domain earlier today but this will probably only apply once you push other changes to this PR.

I am not sure if it is worth using yarn rather than npm, I guess looking at the examples above it is vaguely more useful.

kushthedude:
It is our choice to use yarn/npm. Yarn is relatively new and has more features than npm, including package fetching, package caching and is more popular than npm, has almost twice the number of stars😆. 
I suggest using yarn as both docusaurus and yarn is made by facebook.

wetneb:
Fine with me - can you update the `docs/README.md` in the `docs-netlify` branch to use yarn rather than npm, then?

kushthedude:
> Fine with me - can you update the `docs/README.md` in the `docs-netlify` branch to use yarn rather than npm, then? It will be merged in master by #2567.

Sure

kushthedude:
@wetneb is it good to be merged now?

PR #2571, opened 2020-04-19
Use string representation for matching for GREL cross() function

### LINKED ISSUE #2461: GREL cross function should support lookup by numbers and other datatypes (date, boolean, etc.)

afkbrb:
Close #2461 

As a conclusion to the discussion under #2461, when doing a match with the GREL cross() function, two values will be taken as identical if and only if they have the same string representation.

That's to say, the Integer 1 and Long 1 will match, while 1 and 1.0 won't, because they don't have the same representation.

Also changed the help text for the cross() function to make it more clear.

### DIFF
--- a/main/src/com/google/refine/LookupCacheManager.java
+++ b/main/src/com/google/refine/LookupCacheManager.java
@@ -87,8 +87,9 @@ public class LookupCacheManager {
             Row targetRow = targetProject.rows.get(r);
             Object value = targetRow.getCellValue(targetColumn.getCellIndex());
             if (ExpressionUtils.isNonBlankData(value)) {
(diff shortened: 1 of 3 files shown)


thadguidry:
Nicely done.  Help text is perfect for explaining current handling.

matching column `!@#b` with `!@#c`
image

matching column `!@#b` with itself
image

thadguidry:
However, curious where it gets confused with the record index here?

Seems OK...

image

and the record handling is unchanged I think with this PR (a good thing)

image

PR #2573, opened 2020-04-20
add archive column

james-cui:
S1 ./refine build
S2 ./refine test
S3 ./refine
S4. upload a compressed file and verify the archive file name is added as column


james-cui:
@wetneb  just updated the pull request added several tests.
Following is the screen when running with the pull request
Screen Shot 2020-04-24 at 8 42 46 PM
Screen Shot 2020-04-24 at 8 43 18 PM
Screen Shot 2020-04-24 at 8 43 32 PM

Can you take a look when you have time?

tfmorris:
Does this address an open issue? Can you link it please?

tfmorris:
This is a breaking change to a public extension API. We don't make many promises to extension writers, but the Importer API is one of the few.

The fact that this requires changes to every single one of our importers is a clear sign that something is afoot. All those importers that we don't control will break and be incompatible with the next release.

We need to figure out a different way to do this which preserves the compatibility of the public API.

PR #2575, opened 2020-04-21
remove Freebase AGENT_ID

thadguidry:
Removes AGENT_ID from Servlet.
In the past, Freebase editors used Google Refine for making edits to its database and the internal identifier was "/en/google_refine" which equated to a Software Application type with attached metadata and also had ownership privileges for certain Freebase Apps.  Since Freebase is no longer around, this identifier, only used by Freebase, can now be removed.  (This is not a User-Agent header string but was an internal identifier for the Freebase database which no longer exists)

### DIFF
--- a/main/src/com/google/refine/RefineServlet.java
+++ b/main/src/com/google/refine/RefineServlet.java
@@ -71,9 +71,6 @@ public class RefineServlet extends Butterfly {
     static public String FULL_VERSION = "";
     static public String FULLNAME = "OpenRefine ";
 
-
-    static public final String AGENT_ID = "/en/google_refine"; // TODO: Unused?  Freebase ID
-    
     static final long serialVersionUID = 2386057901503517403L;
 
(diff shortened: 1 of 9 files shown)


thadguidry:
@wetneb OK, ready for review now. :-)

main/src/com/google/refine/RefineServlet.java
```diff
@@ -71,9 +71,6 @@
     static public String FULL_VERSION = "";
     static public String FULLNAME = "OpenRefine ";
 
-
-    static public final String AGENT_ID = "/en/google_refine"; // TODO: Unused?  Freebase ID
-    
     static final long serialVersionUID = 2386057901503517403L;
```
tfmorris, line 74: I doubt this class ever gets serialized, but, technically, the `AGENT_ID` constant is part of the public API and removing it should probably change the `serialVersionUID`.

thadguidry: @tfmorris Can you generate the new UID and submit a PR?

PR #2581, opened 2020-04-22
Change "reference manual" to "product reference"

wetneb:
This is mostly an attempt to check if this pull request gets automatically labelled as "documentation".


wetneb:
Let's try again with a fixed configuration file.

Issue #2562, opened 2020-04-17
type(an error) should return the string "error"


antoine2711:
**Describe the bug**
`type(value)` on a column with an error stored as a value raise the error instead of returning the string "error".

`type()` should never result in an error. At the worst, it should return an empty string or `undefined` (as it does now, most of the time). This is a utility function, and as such, it should be robust.

Current code (probably since a lots of years…):
https://github.com/OpenRefine/OpenRefine/blob/d7d567439eb39957e149e1b67576fe67c7836a10/main/src/com/google/refine/expr/functions/Type.java#L51-L66

**To Reproduce**
1. Go to 'Edit Cells -> Transform'
2. Type '`type(parseJson("a"))`'
4. See error and not "error"

**Current Results**
An error is raised, interrupting the process.

**Expected behavior**
I think `type()` should also intercept errors like this (where value is a previously stored error):
Expression|Result|Expected result
-- | -- | --
value.type()|raise the error IF value is a stored error|"error"
type(parseJson("a"))|raise the error: `parseJson failed: Unrecognized token 'a': was expecting ('true', 'false' or 'null') at [Source: (String)"a"; line: 1, column: 3]`|"error"
type(vfdsf)|"undefined"|"undefined"

**Screenshots**
![image](https://user-images.githubusercontent.com/12217525/78468983-ac198580-76ea-11ea-854f-cf43ef45c353.png)

**OpenRefine (please complete the following information):**
 - Version 3.3

antoine2711:
In the PR « Fix the true.type() == "boolean" #2586 », @wetneb noted that `type()` being a function, it will never have the opportunity to inspect its arguments if they contain errors. He suggested it « might » be converted to a control. 

But, @tfmorris said that this choice would be breaking, as the Dot Syntax functionality would no longer work.

So, the idea of extending the Dot Syntax for control is now discussed (Extending the dot syntax for controls #2618). This would open a non-breaking path to this issue, by transforming it into a control, while still retaining its Dot Syntax functionality.

Regards, Antoine

tfmorris:
Evaluation of parameters happens before the `type()` function gets control, so it's never going to see the error. Also, an error isn't a data type. Data types, which is what `type()` returns are things like String, Date, etc.

If you want to know whether an expression will result in an error (or have a cell which contains an error, rather than a value), use the `isError()` control. Because it's a control rather than a function, it doesn't have its operands evaluated ahead of time.

PR #2586, opened 2020-04-24
(Closes #2510) Fix the true.type() == "boolean"

### LINKED ISSUE #2510: true.type() return « java.lang.Boolean » and not « boolean »
**Describe the bug**

Expression|Result|Expected result
-- | -- | --
true.type()|**"java.lang.Boolean"**|_"boolean"_
(1).type()|"number"|"number"
"A".type()|"string"|"string"
now().type()|"date"|"date"
[].type()|"array"|"array"
null.type()|"undefined"|~~"null"~~ "undefined"

**To Reproduce**
1. Go to 'Edit cells -> Transforms…'
2. Enter 'true.type()' in the Expression field
3. See « java.lang.Boolean » in the preview

**Current Results**
true.type() -> "java.lang.Boolean"
null.type() -> "undefined"

**Expected behavior**
true.type() -> "boolean"
~~null.type() -> "null"~~

 - Version v3.3

antoine2711:
(Closes #2510)
Fix the true.type() == "boolean" instead of "java.lang.Boolean".

Regards,
   Antoine

### DIFF
--- a/main/src/com/google/refine/expr/functions/Type.java
+++ b/main/src/com/google/refine/expr/functions/Type.java
@@ -57,8 +57,10 @@ public class Type implements Function {
                     return "number";
                 } else if (v.getClass().isArray() || v instanceof List<?>) {
                     return "array";
-                } else if (v instanceof EvalError) {
-                    return "error";
+                } else if (v instanceof Boolean) {
+                    return "boolean";
+//                } else if (v instanceof EvalError) { // @ToDo fix this with issue #2562
+//                    return "error";
                 } else {
                     return v.getClass().getName();
                 }
@@ -70,7 +72,7 @@ public class Type implements Function {
 
     @Override
     public String getDescription() {
-        return "Returns the type of o";
+        return "Returns the type of o as a string ('string', 'date', 'number', 'array', 'boolean' or a class name)";
     }
     
     @Override
--- a/main/tests/server/src/com/google/refine/expr/functions/TypeTests.java
+++ b/main/tests/server/src/com/google/refine/expr/functions/TypeTests.java
@@ -26,15 +26,83 @@
  ******************************************************************************/
 package com.google.refine.expr.functions;
 
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
+
+import org.slf4j.LoggerFactory;
+import org.testng.Assert;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.BeforeTest;
 import org.testng.annotations.Test;
 
+import java.util.Arrays;
+import java.util.List;
+import java.util.Properties;
+
+import com.google.refine.RefineTest;
+import com.google.refine.expr.EvalError;
 import com.google.refine.expr.functions.Type;
+import com.google.refine.grel.ControlFunctionRegistry;
+import com.google.refine.grel.Function;
 import com.google.refine.util.TestUtils;
 
-public class TypeTests {
+
+public class TypeTests extends RefineTest {
+    private static Properties bindings;
+    static final List<String> listArray = Arrays.asList("v1", "v2", "v3");
+    private static OffsetDateTime dateTimeValue = OffsetDateTime.parse("2017-05-12T05:45:00+00:00", DateTimeFormatter.ISO_OFFSET_DATE_TIME);
+
+    @Override
+    @BeforeTest
+    public void init() {
+        logger = LoggerFactory.getLogger(this.getClass());
+    }
+
+    @BeforeMethod
+    public void setUp() {
+        bindings = new Properties();
+    }
+
+    @AfterMethod
+    public void tearDown() {
+        bindings = null;
+    }
+
+    /**
+     * Lookup a control function by name and invoke it with a variable number of args
+     */
+    private static Object invoke(String name,Object... args) {
+        // registry uses static initializer, so no need to set it up
+        Function function = ControlFunctionRegistry.getFunction(name);
+        if (function == null) {
+            throw new IllegalArgumentException("Unknown function "+name);
+        }
+        if (args == null) {
+            return function.call(bindings,new Object[0]);
+        } else {
+            return function.call(bindings,args);
+        }
+    }
+
+    @Test
+    public void testTypeInvalidParams() {
+        Assert.assertTrue(invoke("type") instanceof EvalError);
+    }
+
+    @Test
+    public void testType() {
+        Assert.assertEquals(invoke("type", (Object) null),"undefined");
+        Assert.assertEquals(invoke("type", 1),"number");
+        Assert.assertEquals(invoke("type", true),"boolean");
+        Assert.assertEquals(invoke("type", "a string"),"string");
+        Assert.assertEquals(invoke("type", dateTimeValue), "date");
+        Assert.assertEquals(invoke("type", listArray), "array");
+    }
+
     @Test
     public void serializeType() {
-        String json = "{\"description\":\"Returns the type of o\",\"params\":\"object o\",\"returns\":\"string\"}";
+        String json = "{\"description\":\"Returns the type of o as a string ('string', 'date', 'number', 'array', 'boolean' or a class name)\",\"params\":\"object o\",\"returns\":\"string\"}";
         TestUtils.isSerializedTo(new Type(), json);
     }
 }


antoine2711:
@weblate: I don't understand why it doesn't build. On my machine, it build fine.

Regards, Antoine

wetneb:
@tfmorris We can squash commits when merging via GitHub's UI, I routinely do it (and I would propose to make it a default - I think we can configure this in the repository somewhere).

tfmorris:
@wetneb I understand that, but it's an all or nothing process and sometimes there are significant commits which should be kept separate. I've got a strong preference for developers creating clean PRs which are as easy as possible for the reviewers to review without distraction.

antoine2711:
> If the only test change was to account for the changed description, that would imply that we don't have any unit test coverage for the type functionality itself. Can you add one?

Hi @tfmorris: yes, I had the same thought. Can you point me to a file that does the same for another function, and I will base myself from that.
 
> Also, can you squash all the commits which consist of nothing but white space changes? This is small enough that a single commit makes sense.

Yes, I want to merge them. Can you point me to a documentation page on how to do that? I have no clue, I don't know Git.

Thank you for you feedback. Regards, Antoine

wetneb:
@antoine2711 you can imitate this test class, which tests another GREL function:

`main/tests/server/src/com/google/refine/expr/functions/CoalesceTests.java`

antoine2711:
> @antoine2711 you can imitate this test class, which tests another GREL function:
> `main/tests/server/src/com/google/refine/expr/functions/CoalesceTests.java`

Thank you for the information. So I did phase one of test units for this type function. Tell me what to add.

Regards,
   Antoine

main/tests/server/src/com/google/refine/expr/functions/TypeTests.java
```diff
@@ -26,15 +26,84 @@
...
+    
+    @Test
+    public void testType() {
+        Assert.assertEquals(invoke("type", (Object) null),"undefined");
+        Assert.assertEquals(invoke("type", 1),"number");
+        Assert.assertEquals(invoke("type", true),"boolean");
+        Assert.assertEquals(invoke("type", "a string"),"string");
+        Assert.assertEquals(invoke("type", dateTimeValue), "date");
+        Assert.assertEquals(invoke("type", listArray), "array");
+//        Assert.assertEquals(invoke("type", error), "error");
```
wetneb, line 101: I would remove this commented out line.

main/src/com/google/refine/expr/functions/Type.java
```diff
@@ -70,7 +72,7 @@ public Object call(Properties bindings, Object[] args) {
 
     @Override
     public String getDescription() {
-        return "Returns the type of o";
+        return "Returns the type of o as a string ('string', 'date', 'number', 'array', 'boolean', 'error' or a class name)";
```
tfmorris, line 75:
```suggestion
        return "Returns the type of o as a string ('string', 'date', 'number', 'array', or 'boolean'). The string 'error' is returned if the value can't be evaluated";
```
Now that the bug is fixed, can it ever return a Java class name? I'd be tempted to just pare this down to simply "Returns the type of o as a string."

main/tests/server/src/com/google/refine/expr/functions/TypeTests.java
tfmorris, line 101:
```suggestion
```

main/tests/server/src/com/google/refine/expr/functions/TypeTests.java
```diff
@@ -26,15 +26,84 @@
...
+    }
+    
     @Test
     public void serializeType() {
-        String json = "{\"description\":\"Returns the type of o\",\"params\":\"object o\",\"returns\":\"string\"}";
+        String json = "{\"description\":\"Returns the type of o as a string ('string', 'date', 'number', 'array', 'boolean', 'error' or a class name)\",\"params\":\"object o\",\"returns\":\"string\"}";
```
tfmorris, line 106: Fix test to match above.

antoine2711: Is what to be understood here is that type(an error) will only return "error" for certains types of errors? And if so, shouldn't it be documented which error are trap and which one are « reraised »?

Regards, Antoine

antoine2711: I also think this should be taken in consideration with PR #2562.

In particular, if `isError(aParticularValue) == true`, then `type(aParticularValue) == "error"` should be true.

Regards, A.

wetneb:
Ok, the case where we obtain the type of an `EvalError` does not work because `type` is currently a function, not a control, so it cannot catch evaluation errors.
If the argument of a function evaluates to an error, the function call evaluates to that error again.

So we should document this properly: in the code, remove the `EvalError` case, and in the tests, add a comment saying that for now it does not work because `type` is a function and not a control.

Or you can also decide to change `type` from a function to a control in the same PR if you feel like it.

antoine2711:
> Or you can also decide to change `type` from a function to a control in the same PR if you feel like it.

Hi @wetneb. Thank you very much for following up on this. Yes, I think that change `type()` to a control would be the best solution. 

1. It doesn't break anything;
2. It respects the original intent of the devs;
3. It make `type()` more useful as a control.

Regards, Antoine

tfmorris:
> > Or you can also decide to change `type` from a function to a control in the same PR if you feel like it.
> Yes, I think that changing `type()` to a control would be the best solution.
> 
> 1. It doesn't break anything;

Are you sure? Doesn't that mean that `true.type()` will no longer work? Users will have to switch to `type(true)` if it's a control.

antoine2711:
> Are you sure? Doesn't that mean that `true.type()` will no longer work? Users will have to switch to `type(true)` if it's a control.

Thanks @tfmorris for this accurate and important correction.

So, taking in account that we can't change the function's behavior with error arguments for many reasons, would it be possible/easy to code in order to:
- change type() to a control, but also implement the .dot.function() feature;
- or change the behavior of every control by implementing the .dot.function() feature for every control. It would be nice to be able to be able to write: `true.if(eTrueExpr, eFalseExpr)`… It could help making the reading of GREL's code easier.

What would be the other pros and cons of these two options?

Regards,
   Antoine

ostephens:
I'm in favour of looking at allowing the dot syntax for controls, but it probably needs working through carefully to check it doesn't break anything. I'd suggest we keep this PR to its original scope (and improvements to .type() are welcome after all even if there are some cases where it doesn't work), and tackle changes to controls separately

antoine2711:
> I'm in favour of looking at allowing the dot syntax for controls, but it probably needs working through carefully to check it doesn't break anything. I'd suggest we keep this PR to its original scope (and improvements to .type() are welcome after all even if there are some cases where it doesn't work), and tackle changes to controls separately

Hi @ostephens, thank you for your feedback.

So, I created an issue: « Extending the dot syntax for controls #2618 ». Also, for the `type()` function not returning "error", I created last month an issue just on that subject « type(an error) should return the string "error" like the code was designed to do but is buggy #2562 », but no one responded/contributed. I will do a summary there of what we've discussed here.

Regards,
   Antoine

main/src/com/google/refine/expr/functions/Type.java
antoine2711: > Now that the bug is fixed, can it ever return a Java class name? I'd be tempted to just pare this down to simply "Returns the type of o as a string."

@tfmorris: the question you raise is far more complex, from what I see. In Issue « 
true.type() return « java.lang.Boolean » and not « boolean »
#2510 », you will find some discussion on that, but here's a table I did:
Expression|Result|Expected result
--- | -- | :------:
row.type()|com.google.refine.expr.WrappedRow|"object: row" 
row.record.type()| com.google.refine.expr.WrappedRow$WrappedRecord |"object: record"
cell.type()| com.google.refine.expr.WrappedCell |"object: cell"
cells.type()| com.google.refine.expr.CellTuple |"object: cells"
cell.recon.features.type()| com.google.refine.model.Recon$Features |"object: features"
row.record.cells.type()| com.google.refine.expr.WrappedRow$RecordCells |"object: cells"

Because in the code, we have this: `return v.getClass().getName()`, we can never be sure of all the value that can be returned. I guess the only thing we can be sure it's that it's going to be a string…
```
                } else if (v instanceof EvalError) {
                    return "error";
                } else {
                    return v.getClass().getName();
                }
```
I think we should close this issue, and maybe open another one on how to return a type from an object. But closing this issue, we finalize the type() function with basic datatypes.

For this particular change you suggest: « The string 'error' is returned if the value can't be evaluated"; », I wouldn't write that because I don't believe it's accurate. Imagine the case of use `type(value)`, where value contains a previous stored error. But this discussion should be in issue « type(an error) should return the string "error" like the code was designed to do but is buggy #2562 ». Do you share this opinion?

Regards, Antoine

main/tests/server/src/com/google/refine/expr/functions/TypeTests.java
antoine2711: Done.

antoine2711: Done.

antoine2711:
@tfmorris: 
So, the last question standing in this issue is:
« String json = "{\"description\":\"Returns the type of o as a string ('string', 'date', 'number', 'array', 'boolean', 'error' or a class name)\",\"params\":\"object o\",\"returns\":\"string\"}"; »

I've remove the "error" test, like you asked.

I think we should keep it like this, deal with 'error' in Issue #2562, deal with object class name in a new issue, and close this issue. But if you insist on changing the description, I will go in your path.

Regards,
   Antoine

wetneb:
I think it would be cleaner to:
- remove the `EvalError` branch in the function - at the moment this is dead code. We can replace it by a comment explaining that we would like to add support for this, but it requires changing `type` into a control (with a link to the issue)
- update the description, to remove the mention of the `error` value, since it cannot be returned

These things will be easy to add back once #2562 is tackled.

antoine2711:
> * remove the `EvalError` branch in the function - at the moment this is dead code. We can replace it by a comment explaining that we would like to add support for this, but it requires changing `type` into a control (with a link to the issue)
> * update the description, to remove the mention of the `error` value, since it cannot be returned

Done. Added this text: `// @ToDo fix this with issue #2562` and the two lines where commented.

Anything else before merging this PR?

Regards,
   Antoine

main/src/com/google/refine/expr/functions/Type.java
tfmorris: Marking this as resolved. As discussed elsewhere, evaluation errors happen before ever reaching this function.

PR #2587, opened 2020-04-24
Fix null cell editing, followup for #2440

wetneb:
Fixes the editing of null cells, followup for #2440.


wetneb:
No problem at all!

Issue #2593, opened 2020-04-25
ExcelImporter uses too much memory to import


thadguidry:
I often have to work OUTSIDE of OpenRefine to down convert large Excel spreadsheets into CSV files that OpenRefine can import because our current ExcelImporter doesn't use memory efficient ways to handle this.

Example Large XLSX file in the public domain (46 MB , 917000 rows x 15 cols):
https://usace.contentdm.oclc.org/digital/api/collection/p16021coll2/id/4706/download

Many folks have this problem with Apache POI itself such as mentioned here: https://stackoverflow.com/questions/11891851/how-to-load-a-large-xlsx-file-with-apache-poi

There's also a guy who forked the original https://github.com/monitorjbl/excel-streaming-reader and maintains that now here: https://github.com/pjfanning/excel-streaming-reader

Apache POI recognizes that using an InputStream does use more memory than using File.
https://poi.apache.org/apidocs/dev/org/apache/poi/xssf/usermodel/XSSFWorkbook.html#XSSFWorkbook-java.io.InputStream-
but it's not the complete reason for large memory usage;  that stems from the overhead of retaining all the large structures and the content of cells, instead of minimal structure and just the content.

In researching further, I came to know that there are examples in Apache POI package that Nick Burch wrote that uses the EventUserModel code to efficiently (although with fewer options) parse and convert XLS files to CSV (of course while throwing away structures and minimizing Java Objects creation in the process)

POI EXAMPLES:
XSSF (.xlsx and ooxml):
https://svn.apache.org/repos/asf/poi/trunk/src/examples/src/org/apache/poi/xssf/eventusermodel/
HSSF (.xls):
https://svn.apache.org/repos/asf/poi/trunk/src/examples/src/org/apache/poi/hssf/eventusermodel/examples/
SS (either):
https://svn.apache.org/repos/asf/poi/trunk/src/examples/src/org/apache/poi/ss/examples/

Further research lead me to look at the EventUserModel a bit more and I noticed that POI has support for Text Extraction with Excel files in constrained environments (at a cost of losing Java Objects that retain the data structure, but the benefit being less memory is consumed, but there's little to no structure, so its mostly useful for dropping into text mining or Lucene indexes)
https://poi.apache.org/text-extraction.html

I'm wondering 2 things now:

1. Should we refactor ExcelImporter to be more memory efficient? How, with EventUserModel as shown in some of those examples?
[...]

PR #2594, opened 2020-04-26
Use SXSSFWorkbook in XlsExporter

pjfanning:
* to improve memory usage when exporting xlsx files
* indirectly related to https://github.com/OpenRefine/OpenRefine/issues/2593

### DIFF
--- a/main/pom.xml
+++ b/main/pom.xml
@@ -253,12 +253,12 @@
     <dependency>
       <groupId>org.apache.poi</groupId>
       <artifactId>poi</artifactId>
-      <version>4.1.1</version>
+      <version>4.1.2</version>
     </dependency>
     <dependency>
       <groupId>org.apache.poi</groupId>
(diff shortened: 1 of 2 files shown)


wetneb:
Nice! It's convenient that the API remains the same even in streaming mode.
Looking at `main/tests/server/src/com/google/refine/exporters/XlsExporterTests.java` we do not seem to have tests for XLSX export so I am a bit worried about test coverage here (although it's not your fault of course). So we need to test this manually and add a unit test which covers the change.

pjfanning:
There is XlsxExporterTests and this tests xlsx cases.

The XlsExporterTests (for xls files) seems not to have much coverage.

wetneb:
Ah thanks, funny that they were split into two test classes, that's something I should clean up. Yes overall we have a pretty pathetic test coverage in this project. Ok, I'll test this manually and merge.

Issue #2595, opened 2020-04-26
XlsExporter mime types for excel are not standard


pjfanning:
**Describe the bug**

XlsExporter.java has https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/exporters/XlsExporter.java#L68

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

* .xls application/vnd.ms-excel
* .xlsx application/vnd.openxmlformats-officedocument.spreadsheetml.sheet

I can provide a PR if this issue is deemed worth fixing.

**OpenRefine (please complete the following information):**
 - Version - latest master branch

wetneb:
Thanks for reporting this, it would indeed be worth fixing this.

It would also be worth adding these MIME types to `main/webapp/modules/core/MOD-INF/controller.js`, in the map from MIME types to importers (without replacing the existing ones - we should just add new entries in the map).

wetneb:
Closed by #2596

PR #2596, opened 2020-04-26
correct excel mime types

pjfanning:
https://github.com/OpenRefine/OpenRefine/issues/2595

### DIFF
--- a/main/webapp/modules/core/MOD-INF/controller.js
+++ b/main/webapp/modules/core/MOD-INF/controller.js
@@ -223,6 +223,8 @@ function registerImporting() {
 
   IM.registerFormat("text/xml", "XML files", "XmlParserUI", new Packages.com.google.refine.importers.XmlImporter());
   IM.registerFormat("binary/text/xml/xls/xlsx", "Excel files", "ExcelParserUI", new Packages.com.google.refine.importers.ExcelImporter());
+  IM.registerFormat("application/vnd.ms-excel", "Excel files", "ExcelParserUI", new Packages.com.google.refine.importers.ExcelImporter());
+  IM.registerFormat("application/vnd.openxmlformats-officedocument.spreadsheetml.*", "Excel files", "ExcelParserUI", new Packages.com.google.refine.importers.ExcelImporter());
   IM.registerFormat("text/xml/ods", "Open Document Format spreadsheets (.ods)", "ExcelParserUI", new Packages.com.google.refine.importers.OdsImporter());
   IM.registerFormat("text/json", "JSON files", "JsonParserUI", new Packages.com.google.refine.importers.JsonImporter());
   IM.registerFormat("text/marc", "MARC files", "XmlParserUI", new Packages.com.google.refine.importers.MarcImporter());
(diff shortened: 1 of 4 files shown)


main/webapp/modules/core/MOD-INF/controller.js
```diff
@@ -223,6 +223,8 @@ function registerImporting() {
 
   IM.registerFormat("text/xml", "XML files", "XmlParserUI", new Packages.com.google.refine.importers.XmlImporter());
   IM.registerFormat("binary/text/xml/xls/xlsx", "Excel files", "ExcelParserUI", new Packages.com.google.refine.importers.ExcelImporter());
+  IM.registerFormat("application/vnd.ms-excel", "Excel files", "ExcelParserUI", new Packages.com.google.refine.importers.ExcelImporter());
+  IM.registerFormat("application/vnd.openxmlformats-officedocument.spreadsheetml.*", "Excel files", "ExcelParserUI", new Packages.com.google.refine.importers.ExcelImporter());
```
wetneb, line 227: This should be removed - we should only need to add a mapping from the MIME type to the internal format identifier, with `IM.registerMimeType` (see below).

main/webapp/modules/core/MOD-INF/controller.js
```diff
@@ -245,8 +247,8 @@ function registerImporting() {
   IM.registerExtension(".json", "text/json");
   IM.registerExtension(".js", "text/json");
 
-  IM.registerExtension(".xls", "binary/text/xml/xls/xlsx");
-  IM.registerExtension(".xlsx", "binary/text/xml/xls/xlsx");
+  IM.registerExtension(".xls", "application/vnd.ms-excel");
+  IM.registerExtension(".xlsx", "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet");
```
wetneb, line 251: This should be kept: here, "binary/text/xml/xls/xlsx" serves as an internal identifier for the format, and does not need to be an official MIME type. The correct place to add the mapping is further down, with `IM.registerMimeType`.

main/webapp/modules/core/MOD-INF/controller.js
```diff
@@ -291,7 +291,8 @@ function registerImporting() {
   IM.registerMimeType("application/x-excel", "binary/text/xml/xls/xlsx");
   IM.registerMimeType("application/xls", "binary/text/xml/xls/xlsx");
   IM.registerMimeType("application/x-xls", "binary/text/xml/xls/xlsx");
-  
+  IM.registerMimeType("application/vnd.openxmlformats-officedocument.spreadsheetml.*", "binary/text/xml/xls/xlsx");
```
pjfanning, line 294: * I'm not sure if wildcards work here
* application/vnd.ms-excel already has an entry

wetneb: good catch, wildcards are indeed not supported yet, sadly. Perhaps it is still worth keeping here for reference, it does not hurt: I'll add an issue to add support for wildcards which would enable it. Unless there is a small number of possible values for this wildcard, in which case we could add all the variants directly?

wetneb: If we keep this line for now, we should add a comment about the fact that the wildcard is not supported yet: https://github.com/OpenRefine/OpenRefine/issues/2598

pjfanning:
I removed the wildcard

thadguidry:
@pjfanning Thank you so much for putting a few hours into helping us out with all of this! I was blown away waking up this morning and seeing all your activity!

PR #2607, opened 2020-05-02
Fix imprecise facet statistics in records mode

### LINKED ISSUE #2588: Imprecise facet statistics in records mode
**Describe the bug**

The blank values in a column are counted incorrectly in records mode.

**To Reproduce**
Import the following CSV dataset:
```
a,b,c
1,2,3
,,4
,,5
,6,7
3,4,8
,2,9
```
Create a text facet on the second column, **in records mode**.
Observe that it counts a single blank value, whereas there are two of them.

**Current Results**
The count for `(blank)` is 1.

**Expected behavior**
The count for `(blank)` should be 2.

 - master branch (but applicable to all versions since at least 2.6 probably)

joanneong:
Fixes #2588 

The bug results because the `ExpressionNominalValueGrouper` used in `ListFacet` treats row data in the same way in both row and record modes.

To fix this, we have to make sure we do not double count instances of unique non blank data in record mode. I added some additional scratch pad variables instead of creating additional methods (e.g. `visitRecordRow`) because it feels cleaner this way.

I also added a test for the `ExpressionNominalValueGrouper`. Although I tried to reuse the project set up in `@BeforeMethod`, I was facing a lot of trouble changing the project to run in record mode. Hence, I am currently using `parseCSVProject` for a small CSV example that I came up with for this issue.

### DIFF
--- a/main/src/com/google/refine/browsing/util/ExpressionNominalValueGrouper.java
+++ b/main/src/com/google/refine/browsing/util/ExpressionNominalValueGrouper.java
@@ -36,7 +36,6 @@ package com.google.refine.browsing.util;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
--- a/main/src/com/google/refine/browsing/util/ExpressionNominalValueGrouper.java
+++ b/main/src/com/google/refine/browsing/util/ExpressionNominalValueGrouper.java
@@ -86,7 +86,7 @@ public class ExpressionNominalValueGrouper implements RowVisitor, RecordVisitor 
      */
     protected boolean hasBlank;
     protected boolean hasError;
-    protected boolean isVisitingRecord;
+    protected boolean isRowPartOfRecord;
     protected HashSet<String> uniqueNonBlankData;
 
     public ExpressionNominalValueGrouper(Evaluable evaluable, String columnName, int cellIndex) {
@@ -112,8 +112,7 @@ public class ExpressionNominalValueGrouper implements RowVisitor, RecordVisitor 
 
         Properties bindings = ExpressionUtils.createBindings(project);
 
-        uniqueNonBlankData = new HashSet<>();
-        isVisitingRecord = false;
+        isRowPartOfRecord = false;
 
         visitRow(project, rowIndex, row, bindings, rowIndex);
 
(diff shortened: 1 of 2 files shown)


thadguidry:
Some of your comments above... should be useful as code comments additionally, so we don't lose some of that context.

@joanneong 
1. could you add a few code comments to clarify ?  perhaps above L115, L137, and L190 to explain why.  Example:  "the following generates a unique hashset for Rows, to ensure we don't double count instances of non-blank values" ...etc.  "For Records, we also generate a unique hashset of non-blank values in Rows within Records" ...etc.  Or however you want to phrase it to make it explicit what and why we are checking.
2. `isVisitingRecord` and it's usage, perhaps `isVisitingRecord` could be more simply named as `isRecord` ? My thoughts are that we know we `visitRow` and we are then inspecting a Row it see if it is "part of a Record" , YES ?  `rowIsPartOfRecord` ?

joanneong:
@thadguidry I fully agree with what you have suggested, and I have updated my code to reflect that! My only concern now is whether the comments are clear enough/too repetitive.

wetneb:
Intuitively I would have done it the other way round: counting blanks and errors multiple times rather than deduplicating non-blank values (which is how I fixed it in my spark-based datamodel). But happy with this approach too. It's all undocumented anyway, so we can do pretty much what we want.

thadguidry:
@joanneong Yes, @wetneb has a good point, and building the hashset does take up some extra memory with larger datasets, whereas counting reduces the memory overhead and it will scale and align with Spark (which de-duplicates already for us) and its handling later.

wetneb:
I have thought a bit more about this, and I am a bit conflicted:

I agree with Joanne that as intuitively you want to count only distinct values per record, since it matches the spirit of the records mode a bit better. The numbers appearing in the facets tell you how many records are going to be selected when you select only that option. 

But then I also see two reasons not to do this:
* Counting distinct values per record for non-blank non-error values is more of a breaking change than the other way around. Users could potentially rely on these values in their scripts, with the `facetCount` function, and I would expect that if they do, they are looking for counts of non-blank non-error values generally. So it is more dangerous to change the tool's behaviour for this class of values than to tweak the handling of blanks and errors.
* If we count distinct values in this list facet, then we should have a consistent behaviour in other facets. But what does it mean for a numeric facet to count distinct values? We are going to run into issues with equality of floating-point numbers here.

And perhaps a third reason which is efficiency (as Thad pointed out): having to introduce a HashSet for this will definitely not speed things up, and facet computation is already a performance bottleneck with big datasets.

So, my vote goes to counting each value as often as it appears (so, for a single facet in records mode we should have the same facet statistics as in the rows mode).

joanneong:
@thadguidry @wetneb Sorry for the delayed response - I just cleared my last exam earlier today. Thanks for articulating the advantages and disadvantages of different approaches so clearly. Although I do find counting distinct values per record to be more intuitive, it does seem to be the case that the disadvantages outweigh the advantages of this approach.

As such, I am also all for switching to counting each value as often as it appears. I will update this PR in a while to reflect this!

PR #2608, opened 2020-05-03
bump Clojure version to support its new features

thadguidry:
In particular for my use case with `datafy` and Clojure's other new features with better error reporting, etc.

image


thadguidry:
@wetneb freaking hell, again!   Sorry, please squash this when merged!

PR #2611, opened 2020-05-07
Update CHANGES.txt

romy421kumari:


romy421kumari:
update

wetneb:
Thanks for offering this, but I think the original wording is more faithful to the features of the tool. I am not sure what is the point in fixing an old changelog anyway.

Issue #2598, opened 2020-04-26
Support wildcards in importer MIME type registry


wetneb:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
We hold a mapping from MIME types to our supported import formats in `main/webapp/modules/core/MOD-INF/controller.js`.
MIME types can vary and so it would be useful to support wildcards in this mapping: https://github.com/OpenRefine/OpenRefine/pull/2596#discussion_r415296082

**Describe the solution you'd like**

Improve `ImportingManager` to support wildcards in the keys of `mimeTypeToFormat`. The lookup should still be efficient.

**Additional context**

If I remember correctly this was also brought up by @tfmorris recently in another context but I cannot remember where.

tfmorris:
The specific case that I was talking about (and I don't remember the context off-hand) was supporting the use of [Structured Syntax Name Suffixes](https://tools.ietf.org/html/rfc6838#section-4.2.8) as a fallback for importers.

Currently we support `rdf+xml` and know that it's more specific than the generic `text/xml` so we should try the RDF importer before the XML importer, but if we encounter an unknown type with a syntax suffix, we don't prioritize the appropriate importer.

The idea would be that if we were to encounter `application/3gpdash-qoe-report+xml` or `application/activity+json` (to pick the first ones on the [list](https://www.iana.org/assignments/media-types/media-types.xhtml#application)), we would suggest the XML and JSON importers, respectively, as the first fallback options.

This is something we could implement directly in the importer framework, so perhaps it's orthogonal to the idea that @wetneb had for this issue.

tfmorris:
In looking back at the original usage https://github.com/OpenRefine/OpenRefine/pull/2596#discussion_r415296082, the wildcard was used in place of two different types, and was a different use case than the one that I was discussing. Since MIME types are registered individually and are bounded in number, I think wild card support is pretty low priority.

tfmorris:
I'd like to see us drop this (and the associated PR) now that we have #2805. I can't think of a use case that justifies it and any implementation is going to add complexity to the simple table lookup that we have now.

PR #2612, opened 2020-05-08
Fix issue #2598 to support wildcards

CJrZhang:
I wonder if this is the kind of feature that you required in issue #2598. I just implement it for the case that the key is ended with ".*" symbol.


wetneb:
It's not as generic as what I was thinking about but perhaps it can suit most use cases! And it's simple and efficient, so I like it. I am not versed enough in MIME types to understand if the sort of variations we would want to capture with a wildcard always occur at this location, after the last dot.

wetneb:
I would be happy to merge this if we had an accompanying test to demonstrate the feature.

wetneb:
We want to use a different strategy for this: see #2805

PR #2615, opened 2020-05-11
Only show the database login fields when necessary

antoine2711:
The database login field’s visibility are now controlled by CSS styling.

Choosing sqlite will only show the database field.

Regards,
   Antoine


extensions/database/module/scripts/index/database-source-ui.js
```diff
@@ -91,24 +91,44 @@ Refine.DatabaseSourceUI.prototype.attachUI = function(body) {
...
+    } else if(type === "mariadb") {
+        $( "#databaseUser" ).val("root");
+        $( "#databasePort" ).val("3306");
+        databaseClass = "mardb"
+      
+    } else if(type === "sqlite") {
+        $( "#databaseUser" ).val("na");
+        $( "#databasePort" ).val("0");
+        $( "#databaseHost" ).val("na");
+        databaseClass = "sqlit"
```
wetneb, line 116: Is there a particular reason not to reuse the `type` directly as class (perhaps with a prefix, to be on the safe side)? That would simplify the code a bit. You would just have one line `databaseClass = "dbtype" + type`.

antoine2711: > Is there a particular reason not to reuse the `type` directly as class (perhaps with a prefix, to be on the safe side)? That would simplify the code a bit. You would just have one line `databaseClass = "dbtype" + type`.

Yes. The full name would be fine, just 2x longer. I also though of just have TWO classes network_db and file_db. It would even more simply the code, but wouldn't permit to have specific fields per database, just between database types.

So, I go for the full name?

Regards,
   Antoine

antoine2711: > Is there a particular reason not to reuse the `type` directly as class (perhaps with a prefix, to be on the safe side)? That would simplify the code a bit. You would just have one line `databaseClass = "dbtype" + type`.

@wetneb: It's done. You can review it again.

Regards, A.

antoine2711: > Is there a particular reason not to reuse the `type` directly as class (perhaps with a prefix, to be on the safe side)? That would simplify the code a bit. You would just have one line `databaseClass = "dbtype" + type`.

Hi @weblate; I wrote too fast when I said I did what you asked. I forgot half of what you asked! ;-) So here's the other half. For the class name `options`, I changed it to: `dbtype-options`. For the name of the db, I took the full name like you asked, but only prefixed it with `dbt-`. Some names are long, and imagine a user with 10 databases type (core-OR + extensions), it would make a long class attribute value… Anyway, it drastically reduce the risk of collisions. 

Regards, A.

antoine2711:
@wetneb & @thadguidry: so, I guess it's pretty done.
- Each field can be show/hidden with 1 class name (dbtype value prefixed with "dbt-") for every dbtype
- Placeholders of input fields are now all getting they values from i18n.
- Edit Connections now works with show/hide fields
- Default value for dbtype is now queried from the HTML, and used for the first connections or « Reset Connections ».
- Default value for dbHost: it is now localhost everywhere, stored in a property of the object `DatabaseSourceUI`.
If issue _« Access to a JS object on the client side with all the preferences, that would not be updated (frequently) #2619 »_ is implemented, it will be easy for a user to create two preferences, defaultDatabaseType and defaultDatabaseHost, and they would be used by the interface.
- Default connection name is now given by i18n, so it changes in the user language. (`"database-source/connectionNameDefaultValue": "New Connection Name"`).

If you fell there is something else to do, let me know.

Regards,
   Antoine

antoine2711:
@thadguidry: when you have the time, I really think you should check this enhancement. I actually think it's more a bug fix for interface. Anyway, I think we should put it in v3.4, since it's to make SQLite very simple to use.

And now that we have « Access to a JS object on the client side with all the preferences, that would not be updated (frequently) #2619 », we could even store a default value for database server, and database type.

Regards,
   Antoine

extensions/database/module/langs/translation-en.json
```diff
@@ -21,7 +21,17 @@
...
+    "database-source/databaseFileNameLabel": "Database file:",
     "database-source/databaseSchemaLabel": "Schema:",
+    "database-source/connectionNamePlaceholder": "Enter a new Connection Name",
+    "database-source/databaseHostPlaceholder": "Enter Database Host Address",
+    "database-source/databasePortPlaceholder": "Enter Database Port e.g. 3306",
+    "database-source/databaseUserPlaceholder": "Enter Database User",
+    "database-source/databasePasswordPlaceholder": "Enter Database Password",
+    "database-source/databaseNamePlaceholder": "Enter Database",
+    "database-source/databaseFileNamePlaceholder": "Enter the full path to the Database file",
+    "database-source/databaseSchemaPlaceholder": "Enter Initial Schema",
```
thadguidry, line 33:
```suggestion
    "database-source/databaseSchemaPlaceholder": "Enter initial schema",
```

extensions/database/module/scripts/index/database-import-form.html
```diff
@@ -29,44 +29,43 @@
...
-                                           <input bind="connectionNameInput" id="connectionName" type="text" placeholder="Enter Connection Name" value="127.0.0.1" class="pure-input-1-3" required>
+                                           <input bind="connectionNameInput" id="connectionName" type="text" placeholder="" value="" class="pure-input-1-3" required>
                                        </div>
                                
                                        <div class="pure-control-group">
                                            <label id="databaseTypeLabel" for="databaseType">Database Type:</label>
                                             <select name="selection" id="databaseTypeSelect" bind="databaseTypeSelect" class="pure-input-1-3" required>
                                             	  <option value="postgresql">PostgreSQL</option>  
-                                               <option value="mysql">MySQL</option>
+                                               <option value="mysql" selected="selected">MySQL</option>
```
thadguidry, line 39: I guess I'm OK with a default selection here.

extensions/database/module/scripts/index/database-source-ui.js
```diff
@@ -64,6 +64,10 @@ Refine.DatabaseSourceUI.prototype.attachUI = function(body) {
   this._elmts = DOM.bind(this._body);
   var self = this;
   
+  self._defaultDatabaseHost = "localhost";
+  self._defaultDatabaseType = $( "select#databaseTypeSelect" ).val();
+  if (self._defaultDatabaseType == "") { self._defaultDatabaseType = "mariadb"; }
```
thadguidry, line 69: In which scenario would the defaultDatabaseType be empty?  On a brand new installation of OpenRefine?

antoine2711: The only scenario where it would be empty is if a coder changes the HTML code.

But, since it can happen, I check it. What I would really want to do here, is add a preference, for both the IP (host) and DB-type. But that feature is not yet merged in master.

Regards,
   Antoine

thadguidry: OK, understood.

thadguidry:
@antoine2711 It would be really nice to have both Travis or Appveyor, actually test create the sqlite and test... as well as fix the `travis-pgsql.sql` failure on Appveyor.  Whenever you have the time. :-)

antoine2711:
> @antoine2711 It would be really nice to have both Travis and Appveyor, actually test create the sqlite and test... as well as fix the `travis-pgsql.sql` failure on Appveyor. Whenever you have the time. :-)

Sure, I could learn how to do that and implement it. I wouldn't be fast, though. I will follow the discussion here. If you have links to docs on that matter, please share.

Would you like me to create a new issue: « Fix Travis and Appveyor tests for databases ». Would that be good? I think it could deserve a dedicated attention.

Regards,
   Antoine

thadguidry:
@antoine2711 yes please, new issue with my facts above.

antoine2711:
From another thread, « might » be fixed here:

> snussik: By the way, don’t know if it is a bug or feature but if I choose “SQLite” engine and enter any string to database textfield, “test” button shows “Ok”.

antoine2711:
> @antoine2711 yes (…) use a new branch, and iterate on it until db tests are working well on Appveyor and Travis. Looks like mostly just some .yml file updates and .sql file updates in various places are all that's needed. You can see the previous pqsql and mysql as examples… oh wait, we have sqlite already by that fellow a few months ago… so just ensure Appveyor and Travis are indeed testing the DB extension against those 3, `postgres`, `mysql`, `sqlite` properly.

@thadguidry: I don't understand where it fails. If I do: `./refine test`, I get this at the end:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for OpenRefine 3.4-SNAPSHOT:
[INFO] 
[INFO] OpenRefine ......................................... SUCCESS [  0.696 s]
[INFO] OpenRefine - main .................................. SUCCESS [ 37.771 s]
[INFO] OpenRefine - server ................................ SUCCESS [  0.391 s]
[INFO] OpenRefine - extensions ............................ SUCCESS [  0.006 s]
[INFO] OpenRefine - Jython extension ...................... SUCCESS [  3.744 s]
[INFO] OpenRefine - Wikidata extension .................... SUCCESS [  5.949 s]
[INFO] OpenRefine - Database extension .................... SUCCESS [  1.107 s]
[INFO] OpenRefine - Gdata extension ....................... SUCCESS [  1.875 s]
[INFO] OpenRefine - PC-axis extension ..................... SUCCESS [  0.062 s]
[INFO] OpenRefine - Phonetic clustering extension ......... SUCCESS [  1.152 s]
[INFO] OpenRefine - packaging ............................. SUCCESS [  0.846 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  53.817 s
[INFO] Finished at: 2020-05-21T03:07:38-04:00
[INFO] ------------------------------------------------------------------------
Mac-de-Antoine:OpenRefine antoine$ 
```

Where can I see the failing you are talking about?

Regards,
   Antoine

thadguidry:
On GitHub.  The PR when viewed will show near the bottom the checks passing/failing.
image


When you click on Details blue link it will take you to Appveyor or Travis and there you can see the logs from each respectively:
image

tfmorris:
The first thing to fix is to make sure that the run gets marked as failed when it hits an error like this.

wetneb:
@thadguidry the errors you mention are unrelated to this PR. Yes it would be worth cleaning up this SQL script but this is completely independent from these changes.

antoine2711:
> There seems to be an issue with database path validation for the SQLite set-up. If I try to connect to a SQLite database via its full path, I get the error `Connection Name Input Error: Illegal Character in Input. Only [a-zA-Z0-9._-] Allowed`, which does not happen on master branch.

@wetneb: oh! I never saw that. My first impression, is that some data in other fields (hidden for sqlite) are containing bad data. I'll check it.

> Some of the placeholders added in the form inputs seem redundant with the field labels. It does not feel natural to me to have all these "Enter …" prefixes there (these are text fields - the user already knows that they need to enter something there). Also I would not capitalize every word there - these are not titles, so let's not use title case for them.
> 
> For instance having "Enter Database" as placeholder for a "Database" field is not helpful. I would prefer to have something like "database name" instead, since that is more precise - but to be honest this could just be the label of the field itself.

I hear you, and I share your opinion, but, before going further here, I must say that I didn't write this stuff, all of that was previously there in the HTML (the english version). I only factorized it with i18t code.

The french version, on the other hand, is all of my writing. So, in english, I can remove all capitals, except the first, and change « database name ». For the rest, I wouldn't put now too much time, as the i18t data is now exposed, and easier to follow/clean.

Regards,
   Antoine

extensions/database/module/scripts/index/database-import-form.html
```diff
@@ -29,44 +29,43 @@
...
+                                        <div class="pure-control-group dbtype-options dbt-postgresql dbt-mysql dbt-mariadb">
+                                          <label id="databasePortLabel" for="databasePort">Database Port:</label> 
+                                          <input bind="databasePortInput" id="databasePort" type="text" placeholder="" value="5432" class="pure-input-1-3" required>
+                                        </div>
+                                        <div class="pure-control-group dbtype-options dbt-postgresql dbt-mysql dbt-mariadb">
+                                          <label id="databaseUserLabel" for="databaseUser">Database User:</label> 
+                                          <input bind="databaseUserInput" id="databaseUser" type="text" placeholder="" value="postgres" class="pure-input-1-3" required>
+                                        </div>
+                                        <div class="pure-control-group dbtype-options dbt-postgresql dbt-mysql dbt-mariadb">
+                                          <label id="databasePasswordLabel" for="databasePassword">Database Password:</label>
```
wetneb, line 58: How about we remove this text from the HTML file, since it is overwritten by i18n in Javascript anyway?

antoine2711: Done.

antoine2711:
@wetneb: It's fixed. The problem was the name of the connection, it can't have space. So I changed the default to `New_Connection_Name`. Much better than `127.0.0.1` that was there before this PR.

What do we do about the test button which always succeed, even when the path is bad? I think I should open another PR for that.

Regards,
   Antoine

antoine2711:
@thadguidry: per your request, here's the new issue to bring back the Travis and Appveyor tests.
« Bring back Travis & Appveyor tests for Database Import #2637 »

Don't hesitate to modify it, of course. I tried to edit it a bit, but still not perfect.

Regards,
   Antoine

antoine2711:
> Thank you very much for that!

You're welcome. It was a pleasure working with you on that. Thanks for your support and guidance.

Best Regards,
   Antoine

PR #2616, opened 2020-05-11
(I #1038) Add a Show/Hide left panel button

### LINKED ISSUE #1038: Allow facet-undo/redo panel to be shown/hidden
It would be nice to be able to minimize the left panels(that show the facets). This give more space to maneuver without having to use the scroll up and down. I know we can resize the panels but doubleclicking the title area and having it toggle to regular size or minimize would be nice.

antoine2711:
(Closes #1038)
Add a Show/Hide left panel (Facets and Undo/Redo).


antoine2711:
@thadguidry: here, I think a better icon for the button would be a wise choice. What do you think?

Regards,
   Antoine

antoine2711:
> Anyways... Here, we are talking about the entire Modal drawer being collapsed or not, in order to show more grid area. That should be a standard collapse arrow, preferably slightly larger than you have, and not boxed in, but instead open (something like [THIS](https://images.app.goo.gl/KRYSmALmpdFZrHtf7)).

Do you think you could modify the map-picture file yourself? You are much better than me at this graphical level stuff, and I don't really feel I can provide you with better input than what you see/do. Picture file: 
https://github.com/OpenRefine/OpenRefine/blob/8da9eb5095034e40de2545f75e56629e39cd8d53/main/webapp/modules/core/images/leftpanel-showhide.png

If this size changes, come CSS must be updated, and I could do that for you.

Regards, 
   Antoine

thadguidry:
This is how we have OpenRefine designed.  It is a 3 div app with
1. a header
2. a left panel
3. a right panel

and then all other elements fall into those (or should).

The position of the collapse arrow and handling needs to move from the `summary-bar` to the `left-panel`.

image

let me push the updated graphic arrow...

antoine2711:
> The position of the collapse arrow and handling needs to move from the `summary-bar` to the `left-panel`.

Yes, @thadguidry, I though of putting the button in the `left-panel`, but if I do that, it will be hidden when I hide the whole `left-panel`…

> let me push the updated graphic arrow...

I don't really understand most features of Git/GitHub, but I think this is because this branch is on my GitHub, so I've invited you. 

Regards,
   Antoine

thadguidry:
@antoine2711 just merge in my PR. it should auto update this branch PR request against upstream OpenRefine master.  here: https://github.com/antoine2711/OpenRefine/pull/3

thadguidry:
Keep left-panel as a modal...in other words it still shows a sliver. I.E. replace the div with a collapsed div element with minimal width a bit wider than the arrow image.

antoine2711:
> @antoine2711 just merge in my PR. it should auto update this branch PR request against upstream OpenRefine master. here: [antoine2711#3](https://github.com/antoine2711/OpenRefine/pull/3)

Done.

antoine2711:
> Keep left-panel as a modal...in other words it still shows a sliver of space, not completely hidden, but replaced. I.E. with the event, replace the div with a collapsed div element with minimal width a bit wider than the arrow image. Or maybe just resize it very minimally instead of replacing to fake the collapse effect ? ... https://www.npmjs.com/package/@kanety/jquery-simple-resize
> 
> If you want to make this pretty and fancy... Jquery and Jquery UI have a few plugins/extensions that can give you a collapse-able div
> https://api.jquery.com/category/selectors/jquery-selector-extensions/
> https://jqueryui.com/
> https://www.npmjs.com/search?q=keywords:jquery-plugin

Yes, @thadguidry, I see how being resizable would be better, but it would be more complicated. Because of the way it's designed, the width is actually locked in code. I want to change the less possible the code, so this was a « simple » feature. I understand much better solution can be found. If I go more complicated, then it will be more prone to bugging/crashing.

Are you pushing for a custom resize feature here?

Regards,
   Antoine

thadguidry:
I'm pushing for a lot more... but Jquery is not a space that we should push into much...I'd prefer Vue.js and ideally floating panels are what has really been needed, as other SPA grid applications do.  We lack a highly skilled Frontend dev to make all that magic happen.

For now, since you are limited in skillset and time I imagine... we can continue with your current design.  It doesn't break anything that I see, and was a simple enough change for you.

Finally, do you think the background color of the arrows is fine for the current location? (I had changed to match the left-panel div color `#e3e9ff` )

antoine2711:
> I'm pushing for a lot more... 

If we could have all that we want… ;-)

> but Jquery is not a space that we should push into much...I'd prefer Vue.js and ideally floating panels are what has really been needed, as other SPA grid applications do. We lack a highly skilled Frontend dev to make all that magic happen.

Yes, that's also my take. Let's wait for Vue.js for the real coding.
 
> For now, since you are limited in skillset and time I imagine... 

More than I care to admit… ;-)

> we can continue with your current design. It doesn't break anything that I see, and was a simple enough change for you.
> Finally, do you think the background color of the arrows is fine for the current location? (I had changed to match the left-panel background color `#e3e9ff` )

Yes, I think this is a much better improvement. I think that matching the left-panel background is more consistent with our color scheme, so it's a +.

I think it could be improved by shrinking the arrow of 2px (or 3px) in height (not thickness), and putting the border at 2px instead of 1px. But it's merely a suggestion.

Regards,
   Antoine

thadguidry:
I actually increased the height to make it more visible as a collapse arrow.  **Let's leave it as is now**, I really prefer it over the smaller arrow which seemed a bit unnoticeable.  We have a few icons and features that are not as noticeable for some folks, like the small blue X delete on the Projects screen (I still think RED color is the best there, but Antonin disagreed).  We'll get there.

antoine2711:
> I actually increased the height to make it more visible as a collapse arrow. **Let's leave it as is now**, I really prefer it over the smaller arrow which seemed a bit unnoticeable. 

Don't worry, I'm leaving this into your hands. What you did is already a much better version.

> We have a few icons and features that are not as noticeable for some folks, like the small blue X delete on the Projects screen (I still think RED color is the best there, but Antonin disagreed). We'll get there.

Yah, I remember this discussion. I think your red is too agressive, but putting the word « Delete » and a trash can would make the interface clear, which we lack now. Anyway, I'll leaving you two with the pleasure of that discussion! ;-)

Good team work on this. Thanks @thadguidry.

Regards,
   Antoine

antoine2711:
@thadguidry: I surprised myself wanting to experiment other variations. I think you infected me… ;-) This is very close to your last one:

leftpanel-showhide_ab3

If you like it, tell me, I push-it to the codebase.

Regards,
   Antoine

thadguidry:
Oh ohhh... we're creating a Frontend dev!  :-)  Push it if you like it, I'm fine with it.

antoine2711:
> Oh ohhh... we're creating a Frontend dev! :-) Push it if you like it, I'm fine with it.

;-)

Done. It's pushed.

I actually think that your use of the light-blue for hover background color would make the _Column Menu_ arrow much better if applied… We don't really feel the _mouseover_… I'm just saying… ;-)

Regards,
   Antoine

antoine2711:
> First off: thank you very much for tackling this. This is obviously extremely welcome and can make a huge difference to users with small screens. We should have done this a long time ago!

It's a pleasure working with you, @wetneb, to make OR better for our users. You're welcome.
 
> I would prefer to avoid using a clickable `<span>`: I think it is better for accessibility to use an HTML element which really indicates a clickable element (either `<a>` or `<button>`). Similarly, the element should have an alternate text that screen readers can use to make sense of this button (see for instance [this SO question](https://stackoverflow.com/questions/21625672/what-is-an-accessible-way-to-mark-up-buttons-whose-only-content-is-an-icon)).

K. Changed for `<a>`. But, for the `alt=""` attribute, I would need a `` or `<input type="image"/>`. But my picture comes from a background CSS property…

This also rise a few questions:
- where is a model that I can copy/imitate?
- should the `alt=""` be i18n?
- if so, how do I call the key? `core-views/leftPanel-showHide`?

> I am really not saying that the rest of the UI is accessible, but it does not cost much to try to follow accessibility guidelines in the changes we introduce.

I fear that on that regard, we are very much behind…

> About the design, I definitely agree with @thadguidry and you that it would feel much more natural to have the button in the left panel, as shown by [Thad's screenshot](https://user-images.githubusercontent.com/986438/82458729-aef2ee80-9a7c-11ea-9351-f74c25a357cf.png)). How about putting the button in the left panel when it is visible, and moving it to the very beginning of the grid header when the panel is collapsed? 

I see 2 main reasons to keep it as it is: 1) for the user, changing places is distracting and 2) make a moving part increase the possibility and chances of it breaking. he code that is here is as such that I'm confident it won't break.

>I think the "standard" way to implement this in a UI is to have a "View" menu where you can restore any panel that you closed, but we don't have anything like that at the moment, and it would be a bit weird to introduce such a menu in the current layout, especially if it only has a single action in it.

Yes, most probably, a call from a menu would be the standard way, but we don't have that.
 
> Another problem with the current behaviour is that when you open the Wikidata extension, the button gets inserted inside the first tab, whereas it should be outside of that tab in my opinion. This is mostly due to the ugly way I set up the tabs in the Wikidata extension, but we can fix this easily (separate the `summaryText` from the button with some containers and update the Wikidata extension to only wrap the summaryText in the tab).

Well, this one really slipped by me. I've should've pay much more attention to Thad's screenshot, I would've seen it. K, I fix it by moving it BEFORE the `div#summary`. I think it's clean. I also check for it existence as to not create it twice.

I think we now have with this a good balance effort/result, with little breaking possibility.

Regards,
   Antoine

wetneb:
> I see 2 main reasons to keep it as it is: 1) for the user, changing places is distracting and 2) make a moving part increase the possibility and chances of it breaking. he code that is here is as such that I'm confident it won't break.

I don't think it is distracting for the user to have it in different places. The user discovers this button first when the left panel is visible: in that case, it is much easier to understand that the button collapses this panel if it is in the top right corner of that panel. At the moment, because it is next to the row/record count, one can think that it has something to do with that, which is wrong.

When you click on the button, you then see that the arrow changed sides and moved to the new top left corner of the view - I think it is then pretty easy to understand that this button has the opposite effect.

All three of us had this intuitive expectation about the initial location of this button. That is a strong sign: just follow that intuition.

thadguidry:
When I said minimally shown collapsed, I meant something like this, where a very small sliver of the Facet frame (faked or not) is shown always to users.  That way there is a dedicated place for the Arrow and inline with other common applications and user expectations.  My example here shows it enlarged a bit, but you get the idea... a small rectangle just slightly bigger than the Arrow image itself.

Collapsed

wetneb:
That could also work! Obviously we are losing a bit of space there, but it would be consistent and easy to understand. It's probably worth reducing the size of the collapse button a bit, so that when collapsed the bar isn't too wide.

thadguidry:
@wetneb Yes, my example was fat just to get the idea across, but ideally the collapsed rectangle to be skinny, like ultra skinny, like some other apps do.  Folks will easily recognize it.  Bonus points if we can have a bit of CSS animation upon shrinking (if not a slide effect, then just a quick glow/highlight effect around the skinny rectangle would work as well, just so users know what happened, a bit.

antoine2711:
> My example was fat just to get the idea across, but ideally the collapsed rectangle to be skinny, like ultra skinny, like some other apps do. Folks will easily recognize it.

@thadguidry: Leaving a collapse rectangle would mean to change some CSS logic I'm using. To make it simple, I just « hide » the left panel. Keeping it would mean making sure that nothing shows.

Since @wetneb insisted, I coded it to « change » side. And of course, I managed the interaction with the wikidata extension. This is what it looks like:
image

image

image

The Rows count can even be pushed a bit…
image

image

> Bonus points if we can have a bit of CSS animation upon shrinking (if not a slide effect, then just a quick glow/highlight effect around the skinny rectangle would work as well, just so users know what happened, a bit.

I would like the animation, but it would take me time, and would make it more possible to break.

Maybe we can ship this simple and working version, and hear user feedbacks, in order to improve in the next version itineration.

Regards,
   Antoine

thadguidry:
@antoine2711 What you have is great for now, indeed.  We'll get there.

wetneb:
Yes, animation is likely to be extremely hard with the current system that computes element sizes in JS directly. It would not be efficient at all.

wetneb:
Thank you very much for moving the button. It feels extremely natural to me now.
Testing this out, I still have one issue with column headers getting out of sync with the table body.

This happens in quite a few other situations (independent from this PR) and should be fixed by @kushthedude 's fantastic PR that @lisa761 is going to take over, but in the meantime it would be good to add the required JS call to put these two back in sync.

Before collapsing the panel, columns headers and table body are aligned:
image
After collapsing, they get out of sync:
image

The project where this occurs is attached.
[tournagesdefilmsparis2011-csv.openrefine.tar.gz](https://github.com/OpenRefine/OpenRefine/files/4705361/tournagesdefilmsparis2011-csv.openrefine.tar.gz)

antoine2711:
> Thank you very much for moving the button. It feels extremely natural to me now.

Excellent.

> Testing this out, I still have one issue with column headers getting out of sync with the table body.
> 
> This happens in quite a few other situations (independent from this PR) and should be fixed by @kushthedude 's fantastic PR that @lisa761 is going to take over, but in the meantime it would be good to add the required JS call to put these two back in sync.

Now, I did also experience that, in many way, mostly just a little. I already have the code to synch it. I just fails sometimes. In the browser, I can play, and call `document.resize()` multiple times, and each time it gets better a bit or all. And, overall, I find it to happen 10 % of the time on my machine. At the same rate as with other operations.

I could do a « for loop » that calls `document.resize()` as many times as there is columns, but that would definitely be ugly coding. A little 3px widening or shrinking of the window fixes this.

I really on't want to fix `document.resize()`… would be such a waste of time. And to prevent this feature from going out because of that… hum… It's really not a simple situation, in my mind.

Regards,
   Antoine

wetneb:
I have changed your call to `resize()` to a `resizeAll()` which covers resizing the grid UI as well. That does not completely solve the issue, but after clicking the collapse button multiple times the column widths start to converge to the right values! (It's pretty extraordinary that this behaviour is happening).

I agree fixing the resizing logic is beyond the scope of this PR, and would create unnecessary conflicts with Lisa's work.

PR #2622, opened 2020-05-18
Create an access in all the JS code for preferences

### LINKED ISSUE #2619: Access to a JS object on the client side with all the preferences, that would not be updated (frequently)
A direct access to the value of the preferences for reading in the interface at the JS level would be most useful. Right now, you have to do an async call to retrieve them.

Those preferences would be, in the context of this issue, only updated when the user changes of page. Other processes could want to modify/update these preferences, but it's not the scope of this issue.

Two functions, accessible from anywhere in the UI JS code, that would get and set the user preferences. It could be:

`Refine.getPreference(key, defaultValue)` => return a string
`Refine.setPreference(key, newValue)`

antoine2711:
(Closes #2619)
issue-2619-preferences-everywhere will let you access a « relatively » fresh copy of the preferences and their values.

This is not meant to be accurate, just fast and easy to access.

Regards,
   Antoine

### DIFF
--- a/main/webapp/modules/core/scripts/project.js
+++ b/main/webapp/modules/core/scripts/project.js
@@ -230,11 +230,11 @@ Refine.reinitializeProjectData = function(f, fError) {
 Refine.getPreference = function(key, defaultValue) { 
   if(!thePreferences.hasOwnProperty(key)) { return defaultValue; }
   
-  return thePreferences.key;
+  return thePreferences["key"];
 }
 
 Refine.setPreference = function(key, newValue) { 
-  thePreferences.key = newValue;
+  thePreferences["key"] = newValue;
   
   Refine.wrapCSRF(function(token) {
     $.ajax({
--- a/main/webapp/modules/core/scripts/project.js
+++ b/main/webapp/modules/core/scripts/project.js
@@ -32,6 +32,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
 var theProject;
+var thePreferences;
 var ui = {};
 
 var lang = (navigator.language|| navigator.userLanguage).split("-")[0];
@@ -209,9 +210,9 @@ Refine.reinitializeProjectData = function(f, fError) {
                     }
                   } else {
                     if (preferences != null) {
-                      theProject.preferences = preferences;
+                      thePreferences = preferences;
                     }
-                    f();  // f(null);
+                    f();
                   }
                 },
                 'json'
--- a/main/webapp/modules/core/scripts/project.js
+++ b/main/webapp/modules/core/scripts/project.js
@@ -230,11 +230,11 @@ Refine.reinitializeProjectData = function(f, fError) {
 Refine.getPreference = function(key, defaultValue) { 
   if(!thePreferences.hasOwnProperty(key)) { return defaultValue; }
   
-  return thePreferences["key"];
+  return thePreferences[key];
 }
 
 Refine.setPreference = function(key, newValue) { 
-  thePreferences["key"] = newValue;
+  thePreferences[key] = newValue;
   
   Refine.wrapCSRF(function(token) {
     $.ajax({
--- a/main/webapp/modules/core/scripts/project.js
+++ b/main/webapp/modules/core/scripts/project.js
@@ -188,12 +188,35 @@ Refine.reinitializeProjectData = function(f, fError) {
         $.getJSON(
           "command/core/get-models?" + $.param({ project: theProject.id }), null,
           function(data) {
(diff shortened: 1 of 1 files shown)


wetneb:
Thanks! Just a side remark, can you try to link your pull requests to the issues, by mentioning "Closes #2619" in the description of the PR or the commit message?

I am not sure the preferences should be stored in a subfield of `theProject`: these preferences are independent of this specific project, right?

antoine2711:
> I am not sure the preferences should be stored in a subfield of `theProject`: these preferences are independent of this specific project, right?

Yes @wetneb, it's only global preferences. So, should I create `thePreferences`? Or the `thePrefs`?

Regards,
   Antoine

wetneb:
Linking pull requests to issues only works if you do it in the *description* (body, not title) of your pull request or in your commit messages.

antoine2711:
@wetneb: done.

Regards, A.

wetneb:
Thanks a lot! Should we migrate some existing calls to the preference API to use `thePreferences` instead, in the same go?
For instance, this would be a good candidate:
https://github.com/OpenRefine/OpenRefine/blob/8414a82df8dac9941ff918f217649020e99487f5/main/webapp/modules/core/scripts/views/data-table/cell-ui.js#L44-L59

antoine2711:
> Thanks a lot! Should we migrate some existing calls to the preference API to use `thePreferences` instead, in the same go?

@wetneb: interesting proposition, as it will speed up this code. 

I wonder, is there a list of all the preferences that can be set?

But, to answer your question, first, I would rather do this: « Add a preference to control the choice of row quantity displayed (i.e. other than [5, 10, 25, 50]) #2624 ». I really think you should try this.

 Anyway, I created a new issue, « Use the new JS variable « thePreferences » in the interface code #2625 » which is about your request, so it gets put in the system!

Regards,
   Antoine

tfmorris:
> interesting proposition, as it will speed up this code.

I'm having a hard time seeing this as a performance optimization. The AJAX fetch takes 8 msec on my system.

Rather than living with "relatively fresh," which is another way of saying "stale," why don't we just encapsulate the AJAX call in a convenience function like getPreference(key,default) (and refactor the existing code to use it as @wetneb suggests).

wetneb:
> I'm having a hard time seeing this as a performance optimization. The AJAX fetch takes 8 msec on my system.

Although OpenRefine isn't currently designed for that, people do run instances online, so reducing the number of API calls makes sense in general.

I do like the idea of exposing a function `getPreference(key,default)` rather than directly a JS object, since it gives us more control over the underlying implementation. We could do the same for setting preferences too.

About updating the existing preference-fetching parts of the code to use the new API, I think it would be good to do a bit of that in this PR, as it would demonstrate and test the feature better.

antoine2711:
> I do like the idea of exposing a function `getPreference(key,default)` rather than directly a JS object, since it gives us more control over the underlying implementation. We could do the same for setting preferences too.
> 
> About updating the existing preference-fetching parts of the code to use the new API, I think it would be good to do a bit of that in this PR, as it would demonstrate and test the feature better.

@wetneb: I would like to keep this Issue simple and wrap it, because I need it for other features/issues.

That said, I really see two uses for this: one is a rapid querying of preferences, and the other is a need for the current value of a preferences. For the latter, re-querying the pref is important, but for the former, speed is more important, as there could be A LOT of querying (I think of things in the UI related to rows…).

Also, one thing possible would be to update the prefs every 5 min. It would not be complicated, and it would ensure a « fresher » copy of the preferences.

Regards,
   Antoine

tfmorris:
> I would like to keep this Issue simple and wrap it, because I need it for other features/issues.

What depends on this? It seems like a convenience, not a hard dependency.

I'm strongly opposed to a half-baked solution which is what I see a kinda, sorta consistent cache as being and I can't image preference fetching as being on the critical path for performance of any reasonable code.

antoine2711:
> I do like the idea of exposing a function `getPreference(key,default)` rather than directly a JS object, since it gives us more control over the underlying implementation. We could do the same for setting preferences too.
> 
> About updating the existing preference-fetching parts of the code to use the new API, I think it would be good to do a bit of that in this PR, as it would demonstrate and test the feature better.

@wetneb: do you have something like this in mind?

https://github.com/OpenRefine/OpenRefine/blob/6e40b92c16fd3fa3b910be67f116c8520bbeabf9/main/webapp/modules/core/scripts/project.js#L230-L252

Regards,
   Antoine

antoine2711:
I'm closing this issue, since #2624 implements the code described in this issue.

```
Refine.getPreference(key, defaultValue) => return a string
Refine.setPreference(key, newValue)
```

Regards,
Antoine

Issue #1677, opened 2018-07-13
Need a list of all preferences that can be set via UI or openrefine.l4j.ini


quirkiest:
I work with very large datasets (~20MM rows x 12 cols) and set cache & memory to take advantage of the big EC2 server I am using.  Glad to see that in V3 this seems to be very easy.

I also set preferences:
-Drefine.data_dir=D:\OpenRefine\working
-Dui.browsing.listFacet.limit=5000

I'd love to know what the range of prefs actually are - what is available?  For example I set display rows default via data-table-view.js but if it were a preference that'd make life much easier.

thadguidry:
Unfortunately, we were not consistent in how we wired up the PreferenceStore.  But if you look in each of these files from this search, you will see many of them, like the listFacet.limit you found:

https://github.com/OpenRefine/OpenRefine/search?p=2&q=getPreferenceStore%28%29&unscoped_q=getPreferenceStore%28%29

TODO:

- [ ] Update our code so that its easy for us to know what so we can pre-populate the PreferencesStore and UI grid and always show the user the preferences available that can be set in the UI. Probably extend this along with updates where PreferenceStore is used. https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/commands/GetAllPreferencesCommand.java

Expressions (Json Array)
- scripting.starred-expressions (https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/commands/expr/ToggleStarredExpressionCommand.java#L23)
Example: `"expressions":["code":"value.toString()"]`
- scripting.expressions (https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/commands/expr/LogExpressionCommand.java#L57)
- globalExpressions (https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/commands/expr/GetExpressionHistoryCommand.java#L81)

Language
- userLang (Default is `UTF-8`)

Facets
- ui.browsing.listFacet.limit

Google Drive Timeouts
- googleConnectTimeOut
- googleReadTimeOut

Export Template
- exporters.templating.template

Reconciliation (Json Array)
- reconciliation.standardServices

Metadata (Json Array)
- USER_NAME
- projectName
- projectTags
- title
- homepage
- image
- license
- encoding
- userMetadata
- customMetadataColumns
- ??? @jackyq2015 will have to document these.

Wikidata (Json Array)
[...]

PR #2626, opened 2020-05-19
(I #2624) Add preferences for the the grid's page size

### LINKED ISSUE #2624: Add a preference to control the choice of row quantity displayed (i.e. other than [5, 10, 25, 50])
I would like to be able to decide what's would be my choices. A simple preference would do the trick, where I could enter: `[100, 500, 1000, 5000]`.

Set `ui.gridPaginationSize = [100, 500, 1000, 5000]` in the preferences.

Just more paging.

antoine2711:
(Closes #2624) 

Be able to control the choices for the quantity of rows displayed with a preference.

@thadguidry: what do you think of that? I think a lot of users could love this feature.

Regards,
   Antoine


antoine2711:
@wetneb: this code is working, but I still have to write code to check and validate user input:
- is an array
- all elements are positives, uniques numbers (remove what's not)
- the max doesn't exceed 10000 (it works on my machine, but very slow)

Regards, A.

antoine2711:
> Is this a bit of pre-setup work that's needed for the GSoC proposal of Infinite Rows and Pagination enhancement?

This feature is a « goodie » : it’s an easy feature to add, and it gives more power to the user at a minimal cost.

It is not related to GSoS in anyway, and should the interface change, we may not even have to modify the code.

It’s really a small thing. I would put it in v3.4, personally.

Regards, 
   Antoine

wetneb:
I totally understand that you are keen to get features released as soon as possible but at some point we need to draw a line and decide that we only add bug fixes to a release branch. Otherwise we run the risk of introducing bugs in a release.

I am not sure about `displayRowQuantity` as a name: `quantity`. This applies to both rows and records, and isn't really a quantity (it is more of a "count" or "size", which makes it clearer that it is an integer). How about `gridPaginationSize`?

Also, do we have a list of project-level preferences somewhere? If not we should consider creating it and add this new preference there.

antoine2711:
> I totally understand that you are keen to get features released as soon as possible but at some point we need to draw a line and decide that we only add bug fixes to a release branch. Otherwise we run the risk of introducing bugs in a release.

@weblate: I totally agree that the main objective now for v3.4 is to be sure there is no bug. I mean, I won't overpush anything, but still, if a feature is ready and « safe »… it's tempting. Especially if it can get more people to install the beta version and try the new feature. Nevertheless, I see your point and the gamble involved.

> I am not sure about `displayRowQuantity` as a name: `quantity`. This applies to both rows and records, and isn't really a quantity (it is more of a "count" or "size", which makes it clearer that it is an integer). How about `gridPaginationSize`?

Yes, `gridPaginationSize` is much better.

> Also, do we have a list of project-level preferences somewhere? If not we should consider creating it and add this new preference there.

I have made this list informally, for what it's worth. I would put it in the new doc system, but my issues there are stalled.

Key | Since | Default Value
-- | -- | --
reconciliation.standardServices | ? v3.2 |  
cell-ui.previewMatchedCells | ? v3.2 |  
exporters.templating.template | ? v3.2 |  
ui.browsing.listFacet.limit | ? v3.2 |  
userLang | ? v3.2 | en

Regards,
   Antoine

wetneb:
Thank you so much for looking into the supported project-level preferences. It would be worth following the established convention of prefixing the preference key by its domain, so perhaps `ui.gridPaginationSize`?

thadguidry:
For finding the preferences, I just do a GitHub Search, both against the Wiki and Code and then click each result:
https://github.com/OpenRefine/OpenRefine/search?p=1&q=preferences&type=Wikis

antoine2711:
> Thank you so much for looking into the supported project-level preferences. It would be worth following the established convention of prefixing the preference key by its domain, so perhaps `ui.gridPaginationSize`?

Hi @wetneb: I would be cautious to go too fast for this. I think that prefs with a  Dot «.» are actually preferences HIDDEN to the user (i.e. not shown/editable in our preferences panel).

This is a nice conversation going into the work tfmorris started here: https://github.com/OpenRefine/OpenRefine/pull/2630

Regards,
   Antoine

tfmorris:
Based on @antoine2711's comment that this is incomplete, I've converted it to a draft. Please use draft PRs for work in progress.

We definitely need to focus on bugs over features for wrapping up v3.4. I'm guilty of getting distracted by the preferences stuff instead of fixing the expression history.

Our preferences UI sucks, quite frankly, and this not only continues the tradition of requiring the user to magically know the name of the preference, but also to write their desires in the form of a valid JSON array. We'd be better off fixing the UI before going wild adding new preferences. 

If we did do something like this, the default needs to change from 10 to something more appropriate like the first entry on the list.

10000 rows/records is an extremely large upper limit. It is only remotely feasible for the narrowest of tables. From a performance point of view, table width has a significant impact and may influence how we want to handle this.

We're already changing this piece of the UI with the infinite scrolling work, which represents a completely different approach.

All of that is a long-winded way of saying that I think this needs more thought.

p.s. There's a much more comprehensive list of preferences in #1677. Things like the facet limit have been around since the beginning of time (and that one actually has a reasonable UI to set it).

antoine2711:
> Based on @antoine2711's comment that this is incomplete, I've converted it to a draft. Please use draft PRs for work in progress.

I wouldn't say it complete or incomplete, I just asked questions.
 
> We definitely need to focus on bugs over features for wrapping up v3.4. I'm guilty of getting distracted by the preferences stuff instead of fixing the expression history.

Agreeded.

> Our preferences UI sucks, quite frankly, and this not only continues the tradition of requiring the user to magically know the name of the preference, but also to write their desires in the form of a valid JSON array. We'd be better off fixing the UI before going wild adding new preferences.
> 
> If we did do something like this, the default needs to change from 10 to something more appropriate like the first entry on the list.

I also think loading the smallest entry would be the best choice.

> 10000 rows/records is an extremely large upper limit. It is only remotely feasible for the narrowest of tables. From a performance point of view, table width has a significant impact and may influence how we want to handle this.

10,000 would not be a recommended value, mearly a possibility the user could choose, and a choice HE would have to live with. It is also easy to undo.
 
> We're already changing this piece of the UI with the infinite scrolling work, which represents a completely different approach.
> 
> All of that is a long-winded way of saying that I think this needs more thought.

It depends on what we want to give to the user, and the time/resources we have to implement them, also…

> p.s. There's a much more comprehensive list of preferences in #1677. Things like the facet limit have been around since the beginning of time (and that one actually has a reasonable UI to set it).

Thanks. Good info.

Regards,
   Antoine

wetneb:
I definitely agree it feels important to improve our preferences system. At the moment preferences settings are really hard to discover for users, and preferences which expect a JSON value are brittle - we should not expect users to edit that JSON manually.

I personally would not invest much in tweaking this since @lisa761 is going to revamp the grid UI in a much more dramatic way. If we end up only having infinite scrolling around, this preference setting will disappear anyway.

The main question for me is: what is the risk with introducing this preference setting? In what situation is it going to be problematic for us to have it around? If that can help some power users display more rows as a temporary workaround for the lack of infinite scrolling, I am not opposed to merging this. It's not introducing any new interface that we are going to struggle maintaining down the line, even if it is probably going to be a short-lived preference.

thadguidry:
>  If we end up only having infinite scrolling around, this preference setting will disappear anyway.

I will never vote for that.  We need to always have pagination available **for at least** records mode, and in fact offer more options in that mode.  This is incredibly important in the Science and GLAM domains.

wetneb:
> I will never vote for that. We need to always have pagination available for at least records mode, and in fact offer more options in that mode. This is incredibly important in the Science and GLAM domains.

It would be great if you could explain why this is needed, but not here: on the openrefine-dev mailing list, please :)

antoine2711:
@thadguidry & @wetneb: I've implemented all that was talked about, validation of user input included.

Regards,
   Antoine

antoine2711:
@thadguidry: How would you feel about such an option:
image

It's just FOUR lines of code, and a new preference (`ui.toolPanelWidth=1000`)!

I'm half kidding here… ;-)

Regards,
   Antoine

thadguidry:
@antoine2711 Does it set the Grid size?  or the Facet panel size?

antoine2711:
> @antoine2711 Does it set the Grid size? or the Facet panel size?

ui.toolPanelWidth=1000 would be a new issue, with another PR related to it. This issue is about the choice for the number of rows that are displayed in the grid.

Regards,
   A.

thadguidry:
I'll let @wetneb or @tfmorris review this PR now, whenever they have time.

antoine2711:
> The code changes look good to me, but when testing this on a workspace where the preference setting has not been set, I get an error in `_checkPaginationSize`: `gridPageSize is null`.
> 
> Sadly `typeof(null)` evaluates to `"object"` in javascript… so you need to adjust your check in that function.

Yah, sorry for having let this one slip. I swear I do my testing, but not well enough it seams…

K, @wetneb, it is fixed.

Regards,
   Antoine

wetneb:
@antoine2711 happy to merge this when the merge conflicts are solved.

antoine2711:
> @antoine2711 happy to merge this when the merge conflicts are solved.

@wetneb: how can I solved merge's conflicts?

Regards,
   Antoine

antoine2711:
> Testing this, I get a JS error: `tableHeader is not defined` (still in the case where preferences have not been set).

@wetneb: I will look at it now. I did the Merge Resolve here on GH, I should've tried it on my computer. I will do it now and next time.

A.

antoine2711:
> Testing this, I get a JS error: `tableHeader is not defined` (still in the case where preferences have not been set).

@wetneb: done. It was an error in the conflict resolve.

wetneb:
@lisa761 I'll leave this to you, so you can merge/request tweaks as it suits you, since this can potentially interfere with your work.

antoine2711:
> @lisa761 I'll leave this to you, so you can merge/request tweaks as it suits you, since this can potentially interfere with your work.

@lisa761: I don't think there is overlap here between your code (or anybody else’s code) and the changes in this PR. Also, I did one of the 3 three changes you suggested me in the other issue.

Tell me if you see anything else!

Regards,
   Antoine

PR #2628, opened 2020-05-19
Changed cell.error to cell.errorMessage & added help data.

### LINKED ISSUE #525: Add a property to the « cell object » to get access to error text

antoine2711:
(Closes #525)

Changed cell.error to cell.errorMessage and added the informations into the internal help system.

@wetneb: I did the correction, and took the initiative to update the internal help system.

Regards, A.


antoine2711:
@wetneb: can we merge it or should another person review it? If so, who?

Regards, A.

wetneb:
Given the discussion on the issue we are good to go.

wetneb:
Added to 3.4

Issue #2354, opened 2020-03-03
Implement more constraint checks in the Wikidata extension


wetneb:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**

When uploading data to Wikidata, OpenRefine checks for common issues in the uploaded data, and reports these to the user before the upload. Many of these checks rely on Wikidata's own constraint system, which lets Wikidata contributors specify how each Wikidata property should be used (for instance by providing a regular expression for its format).

![Screenshot of the Wikidata issues UI](https://upload.wikimedia.org/wikipedia/commons/5/50/OR-WD-editing-tutorial-issues.png)

The Wikidata extension in OpenRefine only supports [some of the constraints that Wikidata uses](https://www.wikidata.org/wiki/Wikidata:WikiProject_property_constraints/implementations). This means that some problems in data imports can go undetected and get flagged up as constraint violations later on in Wikidata itself.

**Describe the solution you'd like**

We could implement more constraint checks. This could include constraints defined in Wikidata but also other generic checks such as those implemented in #2103.

**Additional context**

Some constraints are expensive to check as they require communicating with Wikidata itself. Since constraint checks are run in real time (to provide quick feedback to the user), we should be careful not to add any expensive operations in new constraint checks.

The architecture of constraint checks in OpenRefine can evolve - for instance to accommodate for more expensive checks transparently, better warnings reported to the user, better handling of multiple constraint declarations of the same type on the same property… The current design is not set in stone.

There is also an interest in developing a generic data validation system, not specific to Wikidata, where all sorts of issues could be reported (think validation against any tabular schema, for instance as defined by the [Data Package](https://frictionlessdata.io/specs/) or [CSVW](https://www.w3.org/2013/csvw/wiki/Main_Page) specs).

--------------------------
_This is a [proposed Outreachy project in 2020](https://www.outreachy.org/apply/project-selection/#openrefine). If you are **not** planning to apply for an internship via Outreachy, we kindly ask that you do not work on this task yet, in order to leave the floor to potential interns._

thadguidry:
Also, regarding a data validation system or "checks", using Schemas and Shape Expressions (ShEx) for validation/rules is something that many folks inside and outside the Wikidata ecosystem have been talking about for a long time and been experimenting with, such as @VladimirAlexiev 
Lots of discussion on approaches here:
https://www.wikidata.org/wiki/Wikidata_talk:WikiProject_Schemas

ostephens:
[...]

PR #2629, opened 2020-05-20
Add Multi-value Scrutinizer

darecoder1999:
Implemented Multi-value constraints as part of #2354


extensions/wikidata/module/langs/translation-en.json
```diff
@@ -126,6 +126,10 @@
     "warnings-messages/no-edit-generated/body": "There might be something wrong with your schema.",
     "warnings-messages/no-issue-detected/title": "No issue was detected in your edits.",
     "warnings-messages/no-issue-detected/body": "Note that OpenRefine cannot detect all the types of problems Wikidata edits can have.",
+    "warnings-messages/multi-valued-property-is-required-for-new-item/title": "{property_entity} should have more than one statements.",
```
wetneb, line 129: "one statements" -> "one statement". Also, it could be good to change the title to something like `{property_entity} should have more than one statement on new items` to make it clear what the difference is with the second issue. (This is mostly useful if both issues get reported at the same time).

The same applies to the other issue title.

extensions/wikidata/module/langs/translation-en.json
```diff
@@ -126,6 +126,10 @@
...
+    "warnings-messages/multi-valued-property-is-required-for-new-item/body": "This property is expected to have more than one statement on each item but it has single statement, for instance on {example_entity}.",
+    "warnings-messages/multi-valued-property-is-required-for-existing-item/title": "{property_entity} should have more than one statements.",
+    "warnings-messages/multi-valued-property-is-required-for-existing-item/body": "This property is expected to have more than one statement on each item but it has single statement, for instance on {example_entity}. It could be a false warning that would be removed once the item is uploaded to wikidata as it might have additional values over there already.",
```
wetneb, line 132: In the second sentence, capital letter for "Wikidata". I would also rephrase the sentence to something like `If the item already has statements with this property in Wikidata, then this warning can be ignored.`

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/MultiValueScrutinizer.java
```diff
@@ -0,0 +1,50 @@
...
+                    QAWarning issue = new QAWarning(new_type, pid.getId(), QAWarning.Severity.WARNING, 1);
+                    issue.setProperty("property_entity", pid);
+                    issue.setProperty("example_entity", update.getItemId());
+                    addIssue(issue);
+                }
+            }
+        } else {
+            for (PropertyIdValue pid : propertyCount.keySet()) {
+                if (propertyCount.get(pid) == 1) {
+                    QAWarning issue = new QAWarning(existing_type, pid.getId(), QAWarning.Severity.WARNING, 1);
```
wetneb, line 41: Should we use a lower severity level for warnings on existing items?

darecoder1999: Very sorry, I didn't get your point. Can you please explain a bit more?

wetneb: I just thought that because the warnings we raise on existing items can potentially be false positives, we could set them to a lower severity level than `QAWarning.Severity.WARNING`, so that they do not look too scary to users. (The severity level determines the icon used in the issues tab). But that is only a suggestion, it is not a big deal if we keep the same level.

darecoder1999:
@wetneb: I have made all the changes as directed. Please let me know if any other requirements are there. :)

PR #2630, opened 2020-05-21
Issue 2619 preferences everywhere

tfmorris:
@antoine2711 @wetneb 

I took a crack extending #2622 to refactor the existing uses of get-preference / set-preference (which was a little more fiddly than I anticipated) and also refactored the cache loading to use Promises so that it's a) not so deeply nested and b) parallel instead of serial.

I haven't reviewed it thoroughly, so there is probably some additional cleanup to be done.

One thing I'd list to do is build a "thenable" PostCSRF(), but I don't know how to do it off the top of my head without doing a little reading first.

### DIFF
--- a/main/src/com/google/refine/commands/GetAllPreferencesCommand.java
+++ b/main/src/com/google/refine/commands/GetAllPreferencesCommand.java
@@ -46,12 +46,17 @@ import com.google.refine.model.Project;
 import com.google.refine.preference.PreferenceStore;
 
 public class GetAllPreferencesCommand extends Command {
-	/**
-	 * The command uses POST (not sure why?) but does not actually modify any state
-	 * so it does not require CSRF.
-	 */
+
     @Override
+    @Deprecated
     public void doPost(HttpServletRequest request, HttpServletResponse response)
+            throws ServletException, IOException {
+        doGet(request, response);
+    }
+
+
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response)
             throws ServletException, IOException {
         
         Project project = request.getParameter("project") != null ? getProject(request) : null;
@@ -63,6 +68,7 @@ public class GetAllPreferencesCommand extends Command {
         
         for (String key : ps.getKeys()) {
             Object pref = ps.get(key);
+            // This skips prefs of type TopList. Not sure if that's intentional or an oversight
             if (pref == null || pref instanceof String || pref instanceof Number || pref instanceof Boolean) {
                 map.put(key, pref);
             }
--- a/main/webapp/modules/core/scripts/index.js
+++ b/main/webapp/modules/core/scripts/index.js
@@ -73,7 +73,7 @@ var dictionary = "";
 $.ajax({
 	url : "command/core/load-language?",
 	type : "POST",
-	async : false,
+	async : false, // FIXME: blocking
 	data : {
 	  module : "core",
 //		lang : lang
--- a/main/webapp/modules/core/scripts/project.js
+++ b/main/webapp/modules/core/scripts/project.js
@@ -40,7 +40,7 @@ var dictionary = "";
 $.ajax({
 	url : "command/core/load-language?",
 	type : "POST",
-	async : false,
+	async : false, // FIXME: blocking
 	data : {
 	  module : "core",
 //		lang : lang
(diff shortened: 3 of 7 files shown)


main/webapp/modules/core/scripts/index.js
```diff
@@ -73,7 +73,7 @@ var dictionary = "";
 $.ajax({
 	url : "command/core/load-language?",
 	type : "POST",
-	async : false,
+	async : false, // FIXME: blocking
```
antoine2711, line 76: Do you mean here that we shouldn't use a synch call?

main/webapp/modules/core/scripts/project.js
```diff
@@ -175,35 +176,58 @@ Refine.setTitle = function(status) {
...
-  $.getJSON(
-    "command/core/get-project-metadata?" + $.param({ project: theProject.id }), null,
-    function(data) {
+  Promise.all([
+    Promise.resolve($.getJSON("command/core/get-project-metadata", { project: theProject.id }))
+    .then(data => {
       if (data.status == "error") {
-        alert(data.message);
-        if (fError) {
-          fError();
```
antoine2711, line 184:
```
       if (fError) {
          fError();
```

I'm curious, why stop using this mechanism?

main/webapp/modules/core/scripts/project.js
```diff
@@ -175,35 +176,58 @@ Refine.setTitle = function(status) {
 };
 
 Refine.reinitializeProjectData = function(f, fError) {
-  $.getJSON(
-    "command/core/get-project-metadata?" + $.param({ project: theProject.id }), null,
-    function(data) {
+  Promise.all([
+    Promise.resolve($.getJSON("command/core/get-project-metadata", { project: theProject.id }))
```
antoine2711, line 180: Do we use Promise somewhere else in the OR's code?

main/src/com/google/refine/commands/GetAllPreferencesCommand.java
```diff
@@ -46,13 +46,18 @@
 import com.google.refine.preference.PreferenceStore;
 
 public class GetAllPreferencesCommand extends Command {
-	/**
-	 * The command uses POST (not sure why?) but does not actually modify any state
-	 * so it does not require CSRF.
-	 */
```
antoine2711, line 52: @tfmorris: I'm not sure I would change that. I bet that the programer's concern at the time with a GET request is that a non-authorized user could have access to private preference's data.

That would be the purpose of the CSRF. Prevent intruders access.

Regards, A.

main/src/com/google/refine/commands/GetAllPreferencesCommand.java
```diff
@@ -63,6 +68,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
         
         for (String key : ps.getKeys()) {
             Object pref = ps.get(key);
+            // This skips prefs of type TopList. Not sure if that's intentional or an oversight
```
antoine2711, line 71: Maybe it's normal, so that the user only sees the first level, that are returned by this call. Internal function that uses the preference mechanism must always be hidden by being TopList…

Maybe I'm wrong.

Regards, A.

antoine2711:
@tfmorris: So, I checked all of your code, and I think this TopList thing is not a trivial matter, and that we cannot with just a read from Get-All-Preferences assume all the data got in…

Then, this solution also does not take into account the multi-tabs situation, whereas the old way did (no caching). Not sure of the implications of what I just wrote, but it seams worthy to mention it.

Now, let's say we go further.
- We could always query `get-preference` when the key contains a dot (.), if the cache is empty.
- For the multi-tabs situation, we could use the `window.sessionStorage`: that's easy and effective.

But, it might be better to tackle all of that in another issue…

Regards,
   Antoine

antoine2711:
I saw this about the `get-all-preferences`:
```
The following commands can still be used with POST but do not require CSRF tokens (although supplying one will not fail), as they do not incur a state change in the backend:

compute-clusters;
compute-facets;
export-project;
export-rows;
get-all-preferences;
get-importing-configuration;
get-importing-job-status;
get-rows;
load-language.
```

https://github.com/OpenRefine/OpenRefine/wiki/Changes-for-3.3

Regards, A.

main/webapp/modules/core/scripts/index.js
tfmorris: Correct. These all block user interactivity of the browser (and you'll see warnings on your Javascript console due to them).

main/webapp/modules/core/scripts/project.js
tfmorris: I didn't. I just moved it to the `catch()`

main/webapp/modules/core/scripts/project.js
tfmorris: No idea. Search for `Promise` or `then` in your IDE to find out.

main/src/com/google/refine/commands/GetAllPreferencesCommand.java
tfmorris: That comment is, I believe, from @wetneb saying that he's not going to add CSRF protection to that post, so it's not currently protected. Also `get-preference` is already available via GET. This makes the API more regular, in my opinion, but I'm open to other opinions. @wetneb ?

tfmorris:
> Then, this solution also does not take into account the multi-tabs situation, whereas the old way did (no caching). Not sure of the implications of what I just wrote, but it seams worthy to mention it.

That was inherent in your original approach and was my main reservation with it, but I decided it's not a big deal since it's possible to have tabs out of sync regardless.

> Now, let's say we go further.

Absolutely not. This is a trivial corner of the product which doesn't deserve this level of attention.

antoine2711:
> > Then, this solution also does not take into account the multi-tabs situation, whereas the old way did (no caching). Not sure of the implications of what I just wrote, but it seams worthy to mention it.
>
> That was inherent in your original approach and was my main reservation with it, but I decided it's not a big deal since it's possible to have tabs out of sync regardless.

My original approche never intended to manage multi-tabs or having a current value. I'm just mentioning this is a broader scope, where we would manage correctly and accurately prefs.
 
> > Now, let's say we go further.
> 
> Absolutely not. This is a trivial corner of the product which doesn't deserve this level of attention.

It's your idea to have a big preference implementation. The avenue I was exploring was much narrower. That being said, the concept of preference will come back, for sure, in the future, even if there is a lot more important features.

Regards, 
   Antoine

main/src/com/google/refine/commands/GetAllPreferencesCommand.java
antoine2711: @tfmorris: this is documented here. I was wrong, CSRF are not needed for this POST, so it's not a security thing.

https://github.com/OpenRefine/OpenRefine/wiki/Changes-for-3.3

```
The following commands can still be used with POST but do not require CSRF tokens (although supplying one will not fail), as they do not incur a state change in the backend:
- compute-clusters;
- compute-facets;
- export-project;
- export-rows;
- get-all-preferences;
- get-importing-configuration;
- get-importing-job-status;
- get-rows;
- load-language.
```

Regards, Antoine

wetneb: Yes it's a good idea to deprecate this - however I am worried that just adding a Java annotation will not be very visible to any client that calls this method using POST. We should make the deprecation more visible, for instance by a warning-level log line, or an additional message added to the JSON response. Perhaps it would be worth keeping the comment (or adapt it) to explain why we are deprecating this.

tfmorris:
Closing this since the suggestion for a cleaner implementation was rejected.

wetneb:
Did we? I did not register that. I think I have been confused by many PRs implementing this new JS preference API in different ways, I was not aware that they were all competing to be merged.

antoine2711:
> Closing this since the suggestion for a cleaner implementation was rejected.

@tfmorris: I think your implementation is superior, and if you are OK with it, I would now take this issue, reopen it, and merge your code to replace the less well written (though stable) code I have done. Do this seams a good path to you?

Regards,
   Antoine

PR #2631, opened 2020-05-21
Make parameter naming consistent - wikibase.upload.maxLag

### LINKED ISSUE #2304: Make maxlag configurable in Wikidata extension

tfmorris:
Fixes #2304 & #2469


wetneb:
Added to the `3.4` branch.

PR #2633, opened 2020-05-22
Restore copyrights

tfmorris:


thadguidry:
Did this get removed only by @jackyq2015 if you know Tom as part of the Gsheet work? or from other people/places also?  Just so I know.

PR #2634, opened 2020-05-22
Add difference-within-range Scrutinizer

darecoder1999:
Implemented difference-within-range constraint as part of #2354

### DIFF
--- a/extensions/wikidata/module/langs/translation-en.json
+++ b/extensions/wikidata/module/langs/translation-en.json
@@ -120,6 +120,8 @@
     "warnings-messages/disallowed-qualifiers/body": "Statements using {statement_property_entity} such as the one on {example_item_entity} should not have a {disallowed_property_entity} qualifier as they are incompatible.",
     "warnings-messages/single-valued-property-added-more-than-once/title": "{property_entity} added more than once on the same item.",
     "warnings-messages/single-valued-property-added-more-than-once/body": "This property is expected to be used at most once on each item but has been added multiple times on the same item, for instance on {example_entity}.",
+    "warnings-messages/difference-of-the-properties-is-not-within-the-specified-range/title": "Inconsistent {source_entity} and {target_entity}",
+    "warnings-messages/difference-of-the-properties-is-not-within-the-specified-range/body": "The difference between {source_entity} and {target_entity} is expected to be within the range [{min_value}, {max_value}], but this fails on items such as {example_entity}.",
     "warnings-messages/identical-values-for-distinct-valued-property/title": "Identical values for {property_entity}",
     "warnings-messages/identical-values-for-distinct-valued-property/body": "This property should have distinct values, but the same value was found on {item1_entity} and {item2_entity} for instance.",
     "warnings-messages/no-edit-generated/title": "No edit was generated.",
--- a/extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
@@ -25,6 +25,7 @@ package org.openrefine.wikidata.qa;
 
 import org.wikidata.wdtk.datamodel.interfaces.ItemIdValue;
 import org.wikidata.wdtk.datamodel.interfaces.PropertyIdValue;
+import org.wikidata.wdtk.datamodel.interfaces.QuantityValue;
 import org.wikidata.wdtk.datamodel.interfaces.Value;
 
 import java.util.Set;
@@ -139,4 +140,38 @@ public interface ConstraintFetcher {
      * Can this property be used on items?
      */
     boolean usableOnItems(PropertyIdValue pid);
+
+    /**
+     * Retrieves the lower bound of the range
+     * required in difference-within-range constraint
+     *
+     * @param pid
+     * @return minimum value
+     */
+    QuantityValue getMinimumValue(PropertyIdValue pid);
+
+    /**
+     * Retrieves the upper bound of the range
+     * required in difference-within-range constraint
+     *
+     * @param pid
+     * @return maximum value
+     */
+    QuantityValue getMaximumValue(PropertyIdValue pid);
+
+    /**
+     * Retrieves the lower value property for calculating the difference
+     * required in difference-within-range constraint
+     *
+     * @param pid:
+     *            the property to calculate difference with
+     * @return the pid of the lower bound property
+     */
+    PropertyIdValue getLowerPropertyId(PropertyIdValue pid);
+
+    /**
+     * Is this property expected to have a value whose difference
+     * with its lower bound property should be in a range?
+     */
+    boolean hasDiffWithinRange(PropertyIdValue pid);
 }
--- a/extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
@@ -25,6 +25,7 @@ package org.openrefine.wikidata.qa;
 
 import org.openrefine.wikidata.utils.EntityCache;
 import org.wikidata.wdtk.datamodel.helpers.Datamodel;
+import org.wikidata.wdtk.datamodel.implementation.QuantityValueImpl;
 import org.wikidata.wdtk.datamodel.interfaces.*;
 
 import java.util.ArrayList;
@@ -78,6 +79,11 @@ public class WikidataConstraintFetcher implements ConstraintFetcher {
     public static String DISTINCT_VALUES_CONSTRAINT_QID = "Q21502410";
 
     public static String MULTI_VALUE_CONSTRAINT_QID = "Q21510857";
+
+    public static String DIFFERENCE_WITHIN_RANGE_CONSTRAINT_QID = "Q21510854";
+    public static String DIFFERENCE_WITHIN_RANGE_CONSTRAINT_PID = "P2306";
+    public static String MINIMUM_VALUE_PID = "P2313";
+    public static String MAXIMUM_VALUE_PID = "P2312";
 
     public static String NO_BOUNDS_CONSTRAINT_QID = "Q51723761";
     public static String INTEGER_VALUED_CONSTRAINT_QID = "Q52848401";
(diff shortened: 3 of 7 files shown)


extensions/wikidata/module/langs/translation-en.json
```diff
@@ -120,6 +120,8 @@
     "warnings-messages/disallowed-qualifiers/body": "Statements using {statement_property_entity} such as the one on {example_item_entity} should not have a {disallowed_property_entity} qualifier as they are incompatible.",
     "warnings-messages/single-valued-property-added-more-than-once/title": "{property_entity} added more than once on the same item.",
     "warnings-messages/single-valued-property-added-more-than-once/body": "This property is expected to be used at most once on each item but has been added multiple times on the same item, for instance on {example_entity}.",
+    "warnings-messages/difference-of-the-properties-is-not-within-the-specified-range/title": "{property_entity} is not having a difference within the range.",
+    "warnings-messages/difference-of-the-properties-is-not-within-the-specified-range/body": "{property_entity} is not having the specified difference w.r.t the lower value property within the range.",
```
wetneb, line 124: These messages are not very natural. As a user I would expect something like this:

"Inconsistent date of birth (P1234) and date of death (P3456)"
"The difference between date of birth (P1234) and date of death (P2345) is expected to be within the range [0 year, 150 year], but this fails on items such as Jane Doe (Q18934) where the difference is -18 year.

extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
```diff
@@ -134,4 +135,35 @@
      * Can this property be used on items?
      */
     boolean usableOnItems(PropertyIdValue pid);
+
+    /**
+     * Retrieves the minimum difference value of the range
```
wetneb, line 140: It would be good to mention that this is used by the difference between range constraint (otherwise one might think this is for another constraint).

extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
```diff
@@ -134,4 +135,35 @@
...
+     *
+     * @param pid
+     * @return minimum value
+     */
+    QuantityValue getMinimumValue(PropertyIdValue pid);
+
+    /**
+     * Retrieves the maximum difference value of the range
```
wetneb, line 148: Same here.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/DifferenceWithinRangeScrutinizer.java
```diff
@@ -0,0 +1,53 @@
...
+            PropertyIdValue pid = statement.getClaim().getMainSnak().getPropertyId();
+            Value value = statement.getClaim().getMainSnak().getValue();
+            propertyIdValueValueMap.put(pid, value);
+        }
+
+        for(PropertyIdValue propertyId : propertyIdValueValueMap.keySet()){
+            if (_fetcher.hasDiffWithinRange(propertyId)){
+                PropertyIdValue lowerPropertyId = _fetcher.getLowerPropertyId(propertyId);
+                QuantityValue minRangeValue = _fetcher.getMinimumValue(propertyId);
+                QuantityValue maxRangeValue = _fetcher.getMaximumValue(propertyId);
```
wetneb, line 27: We should handle the case where only some of these parameters have been defined: we should make sure we do not fail with a null pointer exception, for instance…

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/DifferenceWithinRangeScrutinizer.java
```diff
@@ -0,0 +1,53 @@
...
+
+                if (propertyIdValueValueMap.containsKey(lowerPropertyId)){
+                    Value startingValue = propertyIdValueValueMap.get(lowerPropertyId);
+                    Value endingValue = propertyIdValueValueMap.get(propertyId);
+
+                    if (startingValue instanceof TimeValue && endingValue instanceof TimeValue){
```
wetneb, line 33: This constraint might also be used on quantities: it would be good to support that case too.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/DifferenceWithinRangeScrutinizer.java
```diff
@@ -0,0 +1,53 @@
...
+                        TimeValue lowerDate = (TimeValue)startingValue;
+                        TimeValue upperDate = (TimeValue)endingValue;
+
+                        if (upperDate.getYear() - lowerDate.getYear() < minRangeValue.getNumericValue().longValue()
+                            || upperDate.getYear() == lowerDate.getYear() && upperDate.getMonth() < lowerDate.getMonth()
+                            || upperDate.getYear() == lowerDate.getYear() && upperDate.getMonth() == lowerDate.getMonth() && upperDate.getDay() < lowerDate.getDay()
+                            || upperDate.getYear() - lowerDate.getYear() > maxRangeValue.getNumericValue().longValue()){
```
wetneb, line 40: I think this would be cleaner if we first compute the difference between the two values, in some unit, and then compare it to the bounds supplied in the constraint.

Also, it is not clear to me how this works in the Wikibase implementation, but I think the constraint system automatically converts units: differences can be specified in any unit, as far as I can tell.

See the official docs:

> WikibaseQualityConstraints normalizes the range endpoints before checking this constraint; for example, the intervals [ 0 days , 14 days] and [ 0 weeks , 2 weeks] are equivalent.

https://www.wikidata.org/wiki/Help:Property_constraints_portal/Diff_within_range

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/DifferenceWithinRangeScrutinizer.java
darecoder1999: When I went through the docs It was mentioned that these are [mandatory parameters](https://www.wikidata.org/wiki/Help:Property_constraints_portal/Diff_within_range) for the constraint. Therefore, I didn't add the conditions. If you say so, I'll add those checks.

extensions/wikidata/module/langs/translation-en.json
darecoder1999: Yes, I also felt the same, that I should change the messages. For the time being, I was more concerned about the implementation. I'll surely update these messages, it's must.

extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
darecoder1999: Sure.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/DifferenceWithinRangeScrutinizer.java
darecoder1999: I took a look at the [uses](https://www.wikidata.org/w/index.php?title=Special:WhatLinksHere&target=Q21510854&namespace=120) of this constraint, and all of them are for date & year. Therefore, I made this specific condition for TimeValue only. 

If there are chances to use this constraint on other quantities then should I make a different class extending this one for TimeValue and so on for other quantities?

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/DifferenceWithinRangeScrutinizer.java
wetneb: They are mandatory indeed, but there is nothing preventing users from not adding them to the constraint definition.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/DifferenceWithinRangeScrutinizer.java
wetneb: Good question. Perhaps it is worth checking the code of WikibaseQualityConstraints to see how it was implemented there (perhaps quantities are not supported indeed!)
https://github.com/wikimedia/mediawiki-extensions-WikibaseQualityConstraints

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/DifferenceWithinRangeScrutinizer.java
darecoder1999: If I'll even calculate the difference that would be in terms of year-month-day and the range is provided in years. So, 1st thing that comes in my mind is it might increase the complexity of the code to understand the working and 2nd thing is the comparison with range. 

for instance: here it's quite simple to even check for same death and birth of year but an invalid month and day.  In case of difference, it will be like 0 years -2 month something which doesn't sound good to me. I also checked wikibase implementation before implementing this. They only checked for the year difference which I guess not so appropriate in cases like:
D.O.B : 2020-10-09
D.O.D: 2020-02-02
(ps: I'm not too good in PHP maybe I have overlooked any condition)
Do you want to have that implementation, I can proceed with that. No issues :)

wetneb: Oh indeed, they only support years on their end too. That's not really what the docs say!

Still, I would prefer to compute the difference in years first, and then check how it compares to the bounds, which is what they do:
https://github.com/wikimedia/mediawiki-extensions-WikibaseQualityConstraints/blob/master/src/ConstraintCheck/Checker/DiffWithinRangeChecker.php#L166

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/DifferenceWithinRangeScrutinizer.java
darecoder1999: I checked their implementation, quantities are not really supported over there too

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/DifferenceWithinRangeScrutinizer.java
darecoder1999: Done!

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
```diff
@@ -340,4 +336,84 @@ public boolean usableOnItems(PropertyIdValue pid) {
...
+    /**
+     * Retrieves the lower value property for calculating the difference
+     * required in difference-within-range constraint
+     *
+     * @param pid:
+     *            the property to calculate difference with
+     * @return the pid of the lower bound property
+     */
+    @Override
+    public PropertyIdValue getLowerPropertyId(PropertyIdValue pid) {
```
wetneb, line 369: We have the same problem as the Conflicts With constraint here: as a parameter, we should pass the target property id (the one against which the difference is checked), and then find the corresponding constraint definition instead of only reading the first one.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/DifferenceWithinRangeScrutinizer.java
wetneb: Apparently they are actually supported! :)

darecoder1999: Yes, they do.  I'm trying to implement something similar to rangeCheckerHelper and UnitsConverter. Once it gets implemented very well and validated I'll update the PR. implementing these kinds of classes will provide a good help in range constraints too. :)

Issue #1858, opened 2018-11-22
Improve use of Wikimedia i18n library to provide translated UI


ostephens:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
Following #1846 we need to make further modifications to make use of the more powerful translation options offered by this library

A minimum start would be to amend the dictionary JSON returned by http://127.0.0.1:3333/command/core/load-language? (i.e. LoadLanguageCommand.java) to include multiple languages so the UI can fall back to an appropriate language if a term is not present in their preferred language.

More significant changes would include using smarter translations and pluralisation etc. to make translating the interface more effective and efficient, rather than translating whole phrases (this isn't an area I know much about, but the library definitely opens up new ways of approaching this challenge)

wetneb:
We already have a language fallback (#2209) so the first point has been addressed.

We still need to improve our use of the localization library ([Wikimedia jQuery i18n plugin](https://github.com/wikimedia/jquery.i18n)), mostly:
* to use their support for templates rather than concatenating translated strings to other strings;
* to use support for plurals where appropriate.

For instance, we have things like `$.i18n('core-views/custom-numeric')+'...'`, which is not ideal: there might be languages where adding `...` at the end of such a message does not make sense!

Similarly, we have things like `$.i18n('core-views/split-col')+" " + column.name + " "+$.i18n('core-views/several-col')` which should rather use the template mechanism.

wetneb:
The syntax for templates is described in [jquery.i18n's documentation](https://github.com/wikimedia/jquery.i18n#placeholders).

For instance: we currently have: `$.i18n('core-views/split-col')+" " + column.name + " "+$.i18n('core-views/several-col')` in the code and
```json
   "core-views/split-col": "Split column",
   "core-views/several-col": "into several columns",
```
in the translations.

We should  rather have `$.i18n('core-views/split-column-into-several-columns', column.name)` in the code and
```json
   "core-views/split-column-into-several-columns": "Split column $1 into several columns"
```

tfmorris:
Most packages support named parameters/placeholders, but it looks like the Wikimedia system only allows $1, $2, etc, which is unfortunate because the names are helpful to provide context for the translators in non-obvious cases.

antoine2711:
I have a question:
If I have this in the json-lang files for FR:
`"core-views/goto-page": "$1 de $2 {{plural:$2|page|pages}}"`
I get an error. 

[...]

Issue #1312, opened 2017-11-08
Allow column resize and auto-resize on browser resize


thadguidry:
A community request was mentioned on our mailing list https://groups.google.com/d/topic/openrefine/H3FAf5DSHPk/discussion

This example demonstrates the basic need of having resize-able columns and also auto-resizing the columns if the user changes the size of their browser window.
http://6pac.github.io/SlickGrid/examples/example15-auto-resize.html

jackyq2015:
there is one for AugularJS. just for the records:
https://tympanix.github.io/angular-table-resize/

kushthedude:
@thadguidry I would like to work on the following

thadguidry:
@kushthedude This is a baby step issue, to the larger goal of having a better grid technology that supports many features we would like such as draggable re-orderable columns, resizable columns, etc.  But sure, if you want to isolate on only the "resize columns" feature that's fine.  But how you do it and which technology choosen will impact the larger goal later (or you'll have to refactor/redo).

kushthedude:
> @kushthedude This is a baby step issue, to the larger goal of having a better grid technology that supports many features we would like such as draggable re-orderable columns, resizable columns, etc. But sure, if you want to isolate on only the "resize columns" feature that's fine. But how you do it and which technology choosen will impact the larger goal later (or you'll have to refactor/redo).

@thadguidry Hmm, I was first thinking to introduce `table-layout` property to our container and then go for column resize property for each column. But I just looked into code and saw the table is bound differently, All the column headers are coming from the `DataTableHeader Container` and the table data is coming from `TableDataContainer` which renders the data on row basis by labelling them with `odd/even` id.

wetneb:
Yes, it's a big mess! Resizing is done in JavaScript explicitly, which is massively inefficient since it triggers many DOM reflows. Also the size of the headers often get out of sync with the size of the grid.

kushthedude:
I was thinking to reimplement the tables and bind the content to a single table container if this can be done efficiently. Then the goal of having a better grid technology will be much more easier as we can directly apply our styling to the container rather worrying about how it would affect other columns and their headers.

thadguidry:
[...]

PR #2639, opened 2020-05-24
(I #2638) Feature to Goto a page directly

### LINKED ISSUE #2638: Feature to Goto a page directly
We can go to the first, the last, a next or previous page, but we can't jump to any page.

A simple link, with a prompt, telling you your current page, the max # of pages, and asking to what page you want to go would be a nice feature.
 
image

Clicking many times the next button. I works.

antoine2711:
(Closes #2638)

First implementation. Is functional.


wetneb:
Thanks a lot! I am thinking that as a user it might be more natural to go to a particular row or record number, since the pagination changes depending on the filters applied and the page size.

antoine2711:
> Thanks a lot! I am thinking that as a user it might be more natural to go to a particular row or record number, since the pagination changes depending on the filters applied and the page size.

Good question @wetneb. As a programer, I try to stay in the same logic that was there before. With the « page » concept, if you choose 2, and you go back one page, you end up strait with the first record at the top. I only mimicked the way the « last page » works, which takes into account the hidden rows.

https://github.com/OpenRefine/OpenRefine/blob/1cdf81b4f0fa7b1034fe0762e60989dde376d676/main/webapp/modules/core/scripts/views/data-table/data-table-view.js#L501

It really follow thru the current's OR logic, so I fell this way is less disruptive to the user. But I might be wrong.

As with the other features I coded, my aim is to be as sure as possible it won't break. I know I'm not setting new marks in term of UI, but it's decent, simple, and… working. Also, no dirty tricks in the code, it really a lot on existing code/way OR works.

Regards,
   Antoine

wetneb:
Yes, the API exposed by the backend is not really suited to go to a specific row at the moment: this would require backend changes.

In terms of UI, if we just do it based on page numbers as proposed in this PR, I think we could potentially have a number input directly in the table header, which would show the current page number: the user could edit it directly and confirm with the Enter key or a small button next to it.

Let's not rush this and ask @ostephens and @lisa761 what they think about it.

antoine2711:
> Yes, the API exposed by the backend is not really suited to go to a specific row at the moment: this would require backend changes.

We could circumvent having to change the API by bringing the user to the page that contains the row or the first row of the record. It wouldn't be the first row of the page, but the user would have what he wants.
 
> In terms of UI, if we just do it based on page numbers as proposed in this PR, I think we could potentially have a number input directly in the table header, which would show the current page number: the user could edit it directly and confirm with the Enter key or a small button next to it.

Yes, in terms of UI, that would be better. I didn't choose that path because it's a little bit more complex to code, and much less robust, even if it's not important or with effects if it breaks.

I also think we should consider this: by giving a small feature asked from a long time, effective even if in a basic form, we give value to the user at a minimal cost. Then, user usually feel entitled to come and comment about that (usually in the form: « thanks for this feature (…) but (…) ;-) ). Then, it's easy to get a feel of our user base, and act accordingly, with a possible incremental change in the next version. It gives a feeling to the user that they are listened to. And, in fact, it's not just a feeling, it's a processus and it's real.

It might not seem like that to a seasoned developer on GitHub, but there is a step learning curve for a user not too tecky to come here and start to interact. It's a bit intimidating also.

> Let's not rush this and ask @ostephens and @lisa761 what they think about it.

Of course. Anyway, it's out now, and people can make their own mind and see if they like it! ;-)

Best Regards, @wetneb, 
   Antoine

P.S. I feel I've now pushed enough features in the last month, and it's now time to pause. But… if I was to continue, the next feature on the path would probably be to let a user change the name of a facet. It's also something that could be done simply, and without too much risks of breaking, and that would help a good bunch of users, I feel. (That is also another feature that would be MUCH easier to code with a GenericFacet JS parent class…)

But, instead, let's go in debug/testing mode for v3.4! ;-)

wetneb:
Yes you have been incredibly productive, thank you very much! And I have the impression that you are tackling issues which are directly motivated by your own experience with the tool - those contributions are extremely valuable since they make a big difference to users. So thank you again!

lisa761:
This would be nice to keep if we keep a toggle between infinite scrolling and pagination. It's a useful feature for the users and the code doesn't seem to introduce anything particularly hard to work with for scrolling.

thadguidry:
@antoine2711 Any way for the down arrow in the input box to not rollover into negative integers, but instead 1 would being the lowest?  I like the fluidity of being able to hold down the down arrow... letting it rollover back down to 1 at the lowest, releasing my mouse button, and page 1 would show.  Right now, it doesn't work that way, if I hold down the mouse on down arrow, it goes past 1 into negative numbers and if I release my mouse button, the current page is selected.

thadguidry:
@antoine2711 Any way for the down arrow in the input box to not rollover into negative integers, but instead 1 would being the lowest? I like the fluidity of being able to hold down the down arrow... letting it rollover back down to 1 at the lowest, releasing my mouse button, and page 1 would show. Right now, it doesn't work that way, if I hold down the mouse on down arrow, it goes past 1 into negative numbers and if I release my mouse button in the negative number range, the current page is selected.

antoine2711:
Hi Thad,

> 1. Any way for the down arrow in the input box to not rollover into negative integers, but instead 1 would be the lowest? I like the fluidity of being able to hold down the down arrow... letting it rollover back down to 1 at the lowest, releasing my mouse button, and page 1 would show. Right now, it doesn't work that way, if I hold down the mouse on down arrow, it goes past 1 into negative numbers and if I release my mouse button in the negative number range, the current page is selected.

Very interesting question. Yesterday, I coded so that if the user go over the max, that it sticks to the max. I coded it yesterday, but I didn't pushed it yet to GitHub. I believe, I could do the same with the lower limit, that is, if you go below 1, it will stick at 1. I think this will solve the issue you are raising here.

That code is now pushed.

> 2. I'd also try increasing the height of the input box to 16 pixels, as long as it doesn't adversely affect display of rows (this allows the up/down arrows to slightly increase for easier clicking)

Done.

> 3. The input box width of 50px cuts off display of 1 million + rows, so we should probably bump that to 70px width?  Or can/should we auto-size that width?

Auto-size is difficult. It could be calculated… There. Did it. 20px + 8px per number.

> 4. The wording of `[#####] of #####` leaves me wondering what the context is of those numbers... we should probably add the term `pages` after it?  `[#####] of ##### pages`

Yes indeed. What would still be missing, is the i18n + TEMPLATING ! @wetneb, could you point me to an example of templating? Or a documentation link to implement it? I think I remember you talked about that in another issue.

> 5. Allow the focus to remain in the input box (until users click outside of it), so that users can easily use their Arrow keys on keyboard to go forward or back on page numbers.  Currently, the input box loses focus the moment that the up or down arrow on my keyboard is pressed and it goes to the next page.  More useful to stay focused in input box, so that I can tap my up arrow on keyboard to continually go forward until I click outside of the input box and lose the focus?

Yes, I also think this would be great. I tried to implement it, but failed :-( Any suggestion as to what I should do? Trapping the key down, setting an internal variable to « refocusAfterChange == true » and then, after change, put back the focus?

I'm dealing with internal browser behavior that I would have to modify… That maybe easier to ask than to code…

> Everything else works fantastic and this is a major improvement!

Thank you! I'm happy you like it. It's actually not much code for a lot of user benefit!

Regards,
   Antoine

thadguidry:
Hmm, negative numbers continue to show instead of stopping on 1.
Tried on both Edge (Chromium) and Firefox, depending on if you increase, release mouse button, and then decrease as shown:

PageInput

antoine2711:
> This seems like it would be easier to just use the control attributes themselves?
> `<input type="number" id="something" name="something" min="1">`
> REF: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number

I think I did a mistake… I've put `min` and `max` as CSS styles… ;-) Silly me… Anyway, it's now fixed.

Regards, A.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -173,7 +177,23 @@ DataTableView.prototype._renderPagingControls = function(pageSizeControls, pagin
...
-  $('<span>').addClass("viewpanel-pagingcount").html(" " + from + " - " + to + " ").appendTo(pagingControls);
+  lastPageSpan = $('<span>').attr("id", "viewpanel-paging-last").text(self._lastPageNumber);
+  pageControlSpan = $('<span>').attr("id", "viewpanel-paging-current");
+  
+  currentPageInput = $('<input type="number">')
+    .change(function(evt) { self._onChangeGotoPage(this, evt); })
+    .attr("id", "viewpanel-paging-current-input")
+    .attr("min", 1)
+    .attr("max", self._lastPageNumber)
+    .attr("id", "viewpanel-paging-current-input")
```
thadguidry, line 188: I'd also add `required` so we don't accept an empty value
```suggestion
    .attr("required") 
```
REF: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Validation

thadguidry, line 188: I'd also add `required` so we don't accept an empty value
REF: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Validation
```suggestion
    .attr("id", "viewpanel-paging-current-input")
    .attr("required")
```

thadguidry: Why is "id" attr repeated?

antoine2711: Dumb mistake. I do them sometimes… ;-)

antoine2711: > I'd also add `required` so we don't accept an empty value
> REF: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Validation

Not sure if the required will change something, because I was managing it, but I guess it can't hurt. Added.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -524,6 +526,17 @@ DataTableView.prototype._onChangeGotoPage = function(elmt, evt) {
   this._showRows((gotoPageNumber - 1) * this._pageSize);
 };
 
+DataTableView.prototype._onKeyDownGotoPage = function(elmt, evt) {
+  var keyDownCode = event.which;
+  
+  if([38, 40].indexOf(keyDownCode) == -1) return;
+
+  if(keyDownCode == 38) this._onClickPreviousPage(elmt, evt);
```
thadguidry, line 534: Nice.  It's a bit sluggish, and not quick firing like on the Mozilla Dev example URL I gave earlier, perhaps nothing you can do about that since we're evaluating a bunch on the key press now it seems?

Also, we need to check if currentPageNumber is already 1 ... and we press Down Arrow key...then DO NOTHING (I.E. do not onClickNextPage.  

Because now, if it's on 1 and I press Down Arrow key on my keyboard...it increments to 2, instead of staying on 1.
Same goes for if i press Up Arrow key and the currentPageNumber = Max page number, it should stay at Max page number and not increase or decrease.

thadguidry: And...you need to flip the Up/Down arrow keypress to do the opposite... DOWN is go down the sheet.  Users will expect it to increment, and when you press UP arrow, users would expect it to scroll up on row/record index, or in other words, decrement the page number.

thadguidry: Going to bed... will check on updates in morning.

antoine2711: > Nice. It's a bit sluggish, and not quick firing like on the Mozilla Dev example URL I gave earlier, perhaps nothing you can do about that since we're evaluating a bunch on the key press now it seems?

It's because since we delete/rebuilt control with each change of page, I must resort to an unorthodox method. I wait 1,5 s. for the data-panel to be rebuilt. This is ok, but not more. I'm not proud of that, but I believe you bear the responsability to send me to those murky waters… I'm kidding, of course, but still, this arrow thing is a bit stretching our current table management.

> Also, we need to check if currentPageNumber is already 1 ... and we press Down Arrow key...then DO NOTHING (I.E. do not onClickNextPage).
> 
> Because now, if it's on 1 and I press Down Arrow key on my keyboard...it increments to 2, instead of staying on 1.
> Same goes for if I press Up Arrow key and the currentPageNumber == Max page number, it should stay at Max page number and not increase.

It is now fixed. It stops at bounds.

> And...you need to flip the Up/Down arrow keypress to do the opposite... DOWN is go down the sheet. Users will expect it to increment, and when you press UP arrow, users would expect it to scroll up on row/record index, or in other words, decrement the page number.

I understand that, BUT… just beside are the `<input type="number">` arrows. And they follow the human number convention, where higher # has a meaning. So, we are doing the contrary. This is weird. You think about it.

> Going to bed... will check on updates in morning.
> 
> We're so close to perfection now!!!

Well, I guess you will have fun testing that tomorrow!

Regards,
   Antoine

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -178,10 +177,10 @@ DataTableView.prototype._renderPagingControls = function(pageSizeControls, pagin
...
   }
 
-  lastPageSpan = $('<span>').attr("id", "viewpanel-paging-last").text(self._lastPageNumber);
-  pageControlSpan = $('<span>').attr("id", "viewpanel-paging-current");
+  var pageControlsSpan = $('<span>').attr("id", "viewpanel-paging-current");
   
-  currentPageInput = $('<input type="number">')
+  var pageInputSize = 20 + (8 * ui.dataTableView._lastPageNumber.toString().length);
+  var currentPageInput = $('<input type="number">')
     .change(function(evt) { self._onChangeGotoPage(this, evt); })
```
thadguidry, line 184: I think **something** like this for example would be more efficient... 
```suggestion
    /* Don't fire the GotoPage until the user has stopped typing or pressing keys...
    
    // .change(function(evt) { self._onChangeGotoPage(this, evt); })
    // .keydown(function(evt) { self._onKeyDownGotoPage(this, evt); })
    
   // So, we check the keyup, and delay for some configurable Timeout, default .7 secs, and then attempt to GotoPage...
    .bind('keyup', function(evt) {self._onChangeGotoPage(this, evt); })
```

REFS: 
1. https://stackoverflow.com/questions/1909441/how-to-delay-the-keyup-handler-until-the-user-stops-typing
2. https://github.com/OpenRefine/OpenRefine/blob/3ca08f6ff189cc1bd806a1d49d2ead538882c0e2/main/webapp/modules/core/scripts/facets/text-search-facet.js#L221

antoine2711: @thadguidry: Are you sure this is needed/useful?

I'm not sure I understand you. You want to get the change of page to auto-execute after a certain period of time? What is the added benefit? Not pressing [RETURN]?

Regards,
   Antoine

thadguidry: Lets drop that need however.  Automatically hitting [RETURN] for folks.  They get that with a mouse click on the existing Prev / Next buttons.  I was thinking of accessibility as other apps work this way by holding down up or down arrow to go up or down to pages.  Bit it currently doesn't feel fluid to me with this current code change for some reason.  So let's park that need for now.

The input box should require the user to press [RETURN] key to go to the page they have typed in.  DONE.
If the users want to cycle forward or backward, they will have to use mouse clicks on our Prev / Next buttons.  DONE.

If you want to revert those last changes for now, that's fine.  And then we can merge.

antoine2711: > Lets drop that need however. Automatically hitting [RETURN] for folks. 

If the user presses [RETURN] after changing the number, the browser does the out-of-focus (blur) automatically. I didn't code anything for that.

>They get that with a mouse click on the existing Prev / Next buttons. I was thinking of accessibility as other apps work this way by holding down up or down arrow to go up or down to pages. But it currently doesn't feel fluid to me with this current code change for some reason. So let's park that need for now.

Are you talking of the Arrow Keys, or the Arrow button? image)

Maybe if you give me a precise example of an application and the place in the UI that handles that, I could understand better.

> The input box should require the user to press [RETURN] key to go to the page they have typed in. DONE.
> If the users want to cycle forward or backward, they will have to use mouse clicks on our Prev / Next buttons. DONE.

And don't forget the Keyboard Arrows! That's a nice add-on you proposed.

> If you want to revert those last changes for now, that's fine. And then we can merge.

K. I'll wrap up and finish this issue.

Regards,
   Antoine

antoine2711:
So, I guess it's review time.

@thadguidry, thanks for you guidance, support and contribution. I really believe that the management of the arrow keys is outstanding, and is an advanced and polished feature as we brought it up in this issue. By myself, I wouldn't have got to that, I would have feared its coding complexity/breaking vulnerability. Not sure if it's rock-solid, but I don't see flaws or too much code dirtyness now that it's coded. There's one weakness, and that is the fluidity of the arrow scrolling, but that could be fixed after the major restructuration of the table display code. I'll talk to @lisa761 about that. We know Infinite Scrolling is the ultimate solution for that, and it's on the way.

Anyway, I coined the arrow feature of this issue: « Infinite Paging ». I think it deserves it. ;-)

Best Regards @thadguidry,
   Antoine

thadguidry:
In simple terms.  If you can go back to making it work as that example https://xaksis.github.io/vue-good-table/guide/configuration/pagination-options.html#mode

Then the user just needs to type in the integer and hit [RETURN].
Then we have the Prev / Next Page buttons (and First/Last) additionally.
(no need for that keyboard arrow key up/down I was asking for ... I would remove that functionality and let's keep it simple and straightforward for now as vue-good-table example does)

Thanks for trying to experiment with me and see how far we could make the arrow key up/down work.  Let's remove that functionality and go back to onlythe basic input box as integer and [RETURN] key as the vue-good-table example does.

antoine2711:
> I was talking of the sluggishness of the Arrow Keys using my keyboard to go up/down by a page while focused inside the input box. The alternative is using the Prev / Next - First/Last buttons. Which also let you advance or retreat page by page.
> 
> Sure, If you want to go back to making it work as that example https://xaksis.github.io/vue-good-table/guide/configuration/pagination-options.html#mode without the Keyboard Arrow "infinite paging" feature that's fine.
> 
> Then the user just needs to type in the integer and hit [RETURN].
> Then we have the Prev / Next Page - First/Last buttons additionally.
> 
> Thanks for trying to experiment with me and see how far we could make the "infinite paging" work. If you think we should remove that functionality and go back to only the basic input box as integer and [RETURN] key as the vue-good-table example does...then that's fine. If you think the arrow key up/down works well enough...then that's fine also.
> 
> Your decision.

Well @thadguidry, like I wrote, I think our « Infinite Paging » with the arrow keys is cool, even if it's « sluggish ». I also think that yours propositions and feedbacks helped me do a better job, and « Infinite Paging » is a good example.

> If you think "infinite paging" might get even better later on with @lisa761 's work later on, that's fine and we can keep it around, it works, just a bit sluggish, but works.

Yes, I think that once Vue.js is implemented, with just one table, we could then upgrade more easily this feature. Now, it would be a mess, hard, and would require a LOT of after-testing. I don't think we can afford it.

> I have seen on other apps that whenever the mouse is in the grid window...the scroll wheel can be used up/down to go page by page. So that might be an additional way, where the control is binded to both up/down arrow keys, and mouse scroll wheel, and it pages.

Oh! Well, we have vertical scrolling integrated over the table (up/down, thru the standard browser behavior), but it would be nice to implement horizontal scrolling that would go page to page… And then, left is the page before, and right is the page after! That could be nice. Now, I don't have a mouse with scrolling, but once I get one, I check that. Now, we would still have the sluggishness anyway, just like with arrows keys.

Regards,
   Antoine

antoine2711:
@thadguidry: ok, I wondered a bit about our « Infinite Paging » and I went on Firefox to test it a bit more. It seams Firefox handled with less accuracy the `.focus()` call. So I wrapped it around a try/catch, and it fixed a problem where FF would stop the scrolling for no reason. This part is what I call the weakness in the code, so it's now reinforced. I think it's another little gain toward code solidity. 

Are you having a better experience on FF?

Regards,
   Antoine

thadguidry:
Much better now on Firefox!  Kudos!

I did notice a few times that it still goes out of focus, but that only seems to occur when I hold down the up or down arrow keys on my keyboard and then stopped for a split-second or less and hit the opposite arrow key.  Hmm...Yeah, confirmed, it only seems to exhibit the out of focus when I sometimes hold down the up or down arrow key on my keyboard for say 5 or 6 seconds and then quickly keyup that arrow key and then keydown with the opposite up or down arrow key.

Perhaps there is still something you can do about that?

**Context**: I have a mechanical keyboard ([Corsair K95 RGB Platinum](https://www.corsair.com/uk/en/Categories/Products/Gaming-Keyboards/RGB-Mechanical-Gaming-Keyboards/K95-RGB-PLATINUM-XT-Mechanical-Gaming-Keyboard/p/CH-9127411-UK) that is super fast with German-made CHERRY MX RGB Blue switches), and this might be the real issue here where it fires keyup's and keydown's much faster than most other keyboards, so it exposes these corner case issues.

thadguidry:
UPDATE: The out of focus happens even while holding down an up or down key on my keyboard sometimes.

thadguidry:
If it helps, here is my capture of the performance in Firefox (which you can import into it's dev tools as shown by red arrow)

[profile.zip](https://github.com/OpenRefine/OpenRefine/files/4688871/profile.zip)

image

antoine2711:
> Much better now on Firefox! Kudos!

Thanks. It really a lot of fine tuning now.
 
> I did notice a few times that it still goes out of focus, but that only seems to occur when I hold down the up or down arrow keys on my keyboard and then stopped for a split-second or less and hit the opposite arrow key. Hmm...Yeah, confirmed, it only seems to exhibit the out of focus when I sometimes hold down the up or down arrow key on my keyboard for say 5 or 6 seconds and then quickly keyup that arrow key and then keydown with the opposite up or down arrow key.

Yes, it also happened to me. Just much less. The thing is, I'm now doing now all the error checking I can think about. So it's an unreliability of the browser, but what we are doing, actually the way we are doing it, is unorthodox at best, so we can't really blame the browser.

> Perhaps there is still something you can do about that?
> This **only happens** on Firefox.
> On Chromium browsers, the value change in the input box while I hold down the up/down keys on my keyboard is much slower.
> Firefox, in contrast, is rapid fast with the value change in the input box. Like 3 times faster.

Well, rarely, but still, it happens with Chrome. And you can try with 50 rows. I would love to see this with 500 rows (my other feature #2626)…

> **Context**: I have a mechanical keyboard ([Corsair K95 RGB Platinum](https://www.corsair.com/uk/en/Categories/Products/Gaming-Keyboards/RGB-Mechanical-Gaming-Keyboards/K95-RGB-PLATINUM-XT-Mechanical-Gaming-Keyboard/p/CH-9127411-UK) that is super fast with German-made CHERRY MX RGB Blue switches), and this might be the real issue here where it fires keyup's and keydown's much faster than most other keyboards, so it exposes these corner case issues.

I have a Bluetooth keyboard. Logitech.

@thadguidry, I'm happy with what we have now.

Regards,
   Antoine

thadguidry:
Agree, I think the PR can get merged now actually if someone else tests and gives a 👍 as well.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -173,7 +177,27 @@ DataTableView.prototype._renderPagingControls = function(pageSizeControls, pagin
...
+    .keydown(function(evt) { self._onKeyDownGotoPage(this, evt); })
+    .attr("id", "viewpanel-paging-current-input")
+    .attr("min", 1)
+    .attr("max", self._lastPageNumber)
+    .attr("required", "required")
+    .val(self._currentPageNumber)
+    .css("width", pageInputSize +"px")
+    .appendTo(pageControlsSpan);
+
+  pageControlsSpan.append(' of ');
```
wetneb, line 194: This `' of '` should be localized, and we need to rework the code a bit so that the page numbers that are prepended / appended to the message are inserted in a template instead (so that translators can choose to reorder them if necessary). If this is unclear let me know, I can give examples.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -493,20 +517,80 @@ DataTableView.prototype._showRows = function(start, onDone) {
...
+    this._onClickPreviousPage(elmt, evt);
+  }
+    
+  if(keyDownCode == 40) {  // Down arrow
+    if(newPageValue >= this._lastPageNumber) return;
+    this._onClickNextPage(elmt, evt);
+  }
+  
+  var rehookDelay = 50;
+  var rehookCurrentPageInput = function() {
```
wetneb, line 554: Why do we need to rehook anything? Are you working around a specific browser bug? It would be good to add some comments to explain what is going on here.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
antoine2711: Yes. Somewhere, I asked you an example of code using templates. There is the concept of « placeholder » -> « [FIELD] of [SPAN] pages », and the concept of singular/plural: « [FIELD] of [1] page ».

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
antoine2711: > Why do we need to rehook anything? Are you working around a specific browser bug? It would be good to add some comments to explain what is going on here.

Here, the code logic is simple. The <input> field, from what I understand, gets deleted with a page change. The data gets requeried for each page change, so a new field is created « sometimes » after. The joy with client/server roundtrip is that is take a very different time for the exact same task… network contribution and it's overhead I guess…

So, I don't know « when » the new field will exists. So, I come back quickly, but if the new field doesn't exist, then, I'll just comeback latter, as long as it takes…

It is not the best coding choice, it's the less disruptive code change choice. Less change, potentially less breaking, but maybe not. It's not a hack, but it's not pure. A refactoring of the UI would be better, but development path wise, this now, and the Vue.js change after that, then fixing this would be very easy, I think. More than now anyway.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
wetneb: See https://github.com/OpenRefine/OpenRefine/issues/1858#issuecomment-636313972

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
wetneb: I have heard you talking about "the Vue.js change" before but I am not aware of anyone planning to do that soon, unless @lisa761 has expressed the wish to do this during the GSoC project?

I understand you do not want to introduce code changes that are too involved but I am not comfortable with merging a hack like that, it could easily break if the update takes longer then 50ms.

thadguidry: @wetneb But is it an actual "break" that's not quickly recoverable by a user? Where the UX is completely locked up?  I don't think so, right?  This would just appear to the user that things don't update...they click in another area of the UI to refocus in some other area, then the UX is recovered.  No?  Or am I missing something?

antoine2711: > I have heard you talking about "the Vue.js change" before but I am not aware of anyone planning to do that soon, unless @lisa761 has expressed the wish to do this during the GSoC project?

I was referring to #1312. But it seams it's dropped since a few days…

> I understand you do not want to introduce code changes that are too involved but I am not comfortable with merging a hack like that, it could easily break if the update takes longer then 50ms.

This is not what it does: if after 50ms, it doesn't see the <input> it reschedule a check in the next 50 ms. And it will do that as long as it takes.

What I don't like about this, is that if there is a recurring error, then the mechanism will always be retriggered. That it's limitation. Now, that « shouldn't » happen, but it could in particular circonstances.

Like, if this feature was to be pushed in v3.4, I wouldn't not put « Infinite Paging » in it. But, in v3.5, I would put it.

Regards,
   Antoine

antoine2711:
@thadguidry & @wetneb: I realized that I had some minor changes on my machine but not here. Also, I added a mechanism so that if after 1s. if it can't find the <input>, it forgets about try to keep the focus.

Now, this is flawless on Chrome, but for some reason stops on Firefox after 100 or 200 in average. I think this is clean, and not prone to bug.

Regards,
   Antoine

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
antoine2711: Oups. I'm sorry. My comment is about code not pushed on server. 

Now, what I have is clean. It's recursive, and has a fail safe to stop after _x_ times.

Regards,
   Antoine

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
antoine2711: > See [#1858 (comment)](https://github.com/OpenRefine/OpenRefine/issues/1858#issuecomment-636313972)

Thanks @wetneb for the pointer. It is now coded and working, FR, EN & ES.

I think I have answered all questions and fulfilled all requests for the PR. Is there anything next?

Regards,
   Antoine

antoine2711:
@wetneb: after coding i18n with PLURALS, I tested FR, and it was crashing. So I put something neutral. The thing is, PLURAL work OK in EN and ES…

What do you think? Bug in the library? Bad calling of mine?!

Regards,
   Antoine

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
antoine2711: @wetneb: K, I fixed the problem by using « page(s) » in french. The problem is only in French (AFAIK) and I described the problem in #1858.

For this issue, everything should be K.

Regards,
   Antoine

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -493,20 +513,85 @@ DataTableView.prototype._showRows = function(start, onDone) {
...
+  }
+    
+  if(keyDownCode == 40) {  // Down arrow
+    if(newPageValue >= this._lastPageNumber) return;
+    this._onClickNextPage(elmt, evt);
+  }
+  
+  // For the arrow keys, in order to have a keyDownRepeat, we have to put the focus back on the « currentPageInput »
+  // The <input> seams to be deleted and recreated, so we wait with a few setTimeout() until it is there, unless 1 s. elapse, then we stop
+  const refocusDelay = 50;
```
wetneb, line 551: By principle I do not think we should start relying on timeouts like this to build the UI. If the DOM element is deleted and re-added somewhere, then the code to bind events on it should be placed there, in the code that sets it up.

antoine2711: > By principle I do not think we should start relying on timeouts like this to build the UI.
Yes, I agree with you that.

> If the DOM element is deleted and re-added somewhere, then the code to bind events on it should be placed there, in the code that sets it up.

@wetneb: rebinding is not the problem, it's probably already done. What we have here is the need to « refocus » inside the next input field.

Me, I think the best way would just be to not delete the input, then, you have nothing to do more. The setTimeout is just compensation required by that fact that the paging control are deleted/recreated each time a page changes. It could be much more gentle in the code.

So, I guess we should just drop this « Infinite Paging » with the arrow keys. It's not important, and I coded it because @thadguidry wanted to try that, but it is beyond the scope of this issue. When @lisa761 will be done with the table code overhaul, I can revisit this. But now, either we live with it as is, or we remove it. I doesn't break at all on Chrome, so it's not a bad coding.

Maybe I don't understand what you mean by « then the code to bind events on it should be placed there, in the code that sets it up. » because from what I see, it gets rebinded after being rebuilt.

Regards,
   Antoine

wetneb: What I mean is that we should not rely on any timeouts to refocus / rebind / do any other type of set-up on the UI. Just use callbacks that are executed when some elements are created / deleted / other lifecycle events. This is just to avoid building up technical dept, running into situations were we have to deal with various race conditions between the timeouts, and so on.

antoine2711: > What I mean is that we should not rely on any timeouts to refocus / rebind / do any other type of set-up on the UI. Just use callbacks that are executed when some elements are created / deleted / other lifecycle events. This is just to avoid building up technical dept, running into situations were we have to deal with various race conditions between the timeouts, and so on.

@wetneb: ok, I get something in my mind now with what you write. K. A very simple fix would be: everytime the page is loaded, that focus goes to the field. That way, I could just remove my coding, and all will work.

What do you think of that?

Regards,
   Antoine

wetneb: It does sound better to me!

antoine2711: > It does sound better to me!

Et voilà !

@wetneb: It's fixed. And now, it even works perfectly in FF, not just Chrome.

Now, I guess it's done! ;-)

Regards,
   Antoine

thadguidry: Yeap, can confirm this works well in FF now for me.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -173,8 +179,32 @@ DataTableView.prototype._renderPagingControls = function(pageSizeControls, pagin
...
+  pageControlsSpan.append($.i18n('core-views/goto-page', '<span id="currentPageInput" />', self._lastPageNumber));
+  pageControlsSpan.appendTo(pagingControls);
 
+  $('span#currentPageInput').replaceWith($(currentPageInput));
+  
+  if(self._refocusPageInput == true) { 
+    self._refocusPageInput = false;
+    var currentPageInput = $('input#viewpanel-paging-current-input');
+    currentPageInput.ready(function(evt) { 
+      (new Promise(resolve => setTimeout(resolve, 1000))).then(currentPageInput.focus());
```
wetneb, line 204: There is still a `setTimeout` here :)

antoine2711: > There is still a `setTimeout` here :)

@wetneb: yes, but this one is very clean. It's to make the « scrolling » thru pages not too fast. If you have a better delay strategy, please advise me.

Regards,
   Antoine

antoine2711:
@tfmorris: I've pull the master branch with PR #2717, I've put back the french plural, and I still the same « Uncaught ReferenceError: pluralRuleParser is not defined » for the french.

Can you advise me on how to get that?

Regards,
   Antoine

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
wetneb: ok fair enough, I will test this as a user now, and might go back to this later on when I understand its role better.

wetneb:
Hi @antoine2711,

@lisa761 is interested in getting this PR merged because it could be useful for her to build on it in the infinite scrolling project (specifically for "speed scrolling", so dragging the scrollbar to a particular position quickly, without seeing the intermediate rows).

Of course, now that #2719 is merged, we have merge conflicts here… Could one of you take care of the conflict resolution?

thadguidry:
@wetneb that "2" in your screenshot looks like a completely different font which might be affecting the padding/margins.  Hmm, why is it different on your system? Which font does it show using for that element?

wetneb:
These are the computed attributes of this element in Firefox, for me:

```
font-family: Sans;
font-size: 13.3333px;
height: 34px;
margin-bottom: 0px;
margin-left: 0px;
margin-right: 0px;
margin-top: 0px;
padding-bottom: 0px;
padding-left: 0px;
padding-right: 0px;
padding-top: 0px;
position: relative;
top: -2px;
vertical-align: middle;
width: 28px;
```

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
antoine2711: > ok fair enough, I will test this as a user now, and might go back to this later on when I understand its role better.

I'm all available to explain everything to speed up this. I'm very confident that all the requirements are there. We can also go thru the other features that are still waiting for approval.

Regards,
   Antoine

thadguidry:
It looks and works great on mine with Firefox (defaults) on Windows.
image


```
font-family: MS Shell Dlg 2;
font-size: 13.3333px;
height: 16px;
margin-bottom: 0px;
margin-left: 0px;
margin-right: 0px;
margin-top: 0px;
padding-bottom: 0px;
padding-left: 0px;
padding-right: 0px;
padding-top: 0px;
position: relative;
top: -2px;
vertical-align: middle;
width: 28px;
```

Firefox settings:
image

tfmorris:
> I've pull the master branch with PR #2717, I've put back the french plural, and I still the same « Uncaught ReferenceError: pluralRuleParser is not defined » for the french.

Please put steps to reproduce, including the branch/commit that you're using on #2700 so that it doesn't get lost here and I'll have a look. You could also look at the French plurals on that branch to see how they differ, if at all, from what you are doing.

antoine2711:
> Hi @antoine2711,
> 
> @lisa761 is interested in getting this PR merged because it could be useful for her to build on it in the infinite scrolling project (specifically for "speed scrolling", so dragging the scrollbar to a particular position quickly, without seeing the intermediate rows).
> 
> Of course, now that #2719 is merged, we have merge conflicts here… Could one of you take care of the conflict resolution?

Hi @wetneb, I don't understand. For my, this is working, both on GC and FF on Mac. I haven't tested on Windows.

FireFox
image

Chrome
image

I've put back « page(s) » in french, let's deal with that some other time.

@wetneb: is it also expected from a contributor like me to test multi-platform?

Regards,
   Antoine

wetneb:
Ok then this must be something specific to my own setup, it is curious. No, there is no expectation that you are testing things on multiple platforms unless you are specifically working on things that are platform-dependent in nature (such as packaging).

antoine2711:
> Ok then this must be something specific to my own setup, it is curious.

I will test it on Windows 10. Is that close to your setup, @wetneb? Also, I changed the way the delay was working, now, it really delay.

Regards,
   Antoine

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -133,8 +138,7 @@ DataTableView.prototype.render = function() {
     this._renderSortingControls(elmts.sortingControls);
   }
 
-  this._renderDataTables(elmts.table[0], elmts.tableHeader[0]);
-  this._div.empty().append(html);
+  this._renderDataTables(elmts.table[0], elmts.headerTable[0]);
```
lisa761, line 141: This line throws an error as there is no `headerTable` anymore, it was changed to `tableHeader` in #2719. 

So if I pull the PR now, the projects don't load:

main/webapp/modules/core/styles/views/data-table-view.less
```diff
@@ -55,7 +55,53 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+  }
+
+.data-header-table-container {
+  border-top: 1px solid @chrome_primary;
+  border-bottom: 2px solid @chrome_primary;
+  background: @fill_primary;
+  overflow: visible;
+  }
+
+.data-header-table {
```
lisa761, line 104: `.data-header-table` was also changed to `data-table-header` in #2719.

main/webapp/modules/core/styles/views/data-table-view.less
```diff
@@ -55,7 +55,53 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+.viewpanel-paging input#viewpanel-paging-current-input {
+  width: 60px;
+  margin: 0px;
+  padding: 0px;
+  height: 16px;
+  position: relative;
+  top: -2px;
+  }
+
+.data-header-table-container {
```
lisa761, line 97: We don't need this style for ` .data-header-table-container` either, as that container was also removed (unless you're using it somewhere?). I missed it earlier.

antoine2711: @lisa761: thank you so much for pointing me your changes in my PR. I would have had a very hard time to find them by myself.

I'm now pretty good, and you, can you run it smooth? Have you tried the Up/Down Arrow Keys?

Regards,
   Antoine

lisa761: Yes, it's pretty great actually!

antoine2711:
@wetneb: Window 10

**GC**:
image

**FF**:
image

This all seams good.

Regards,
   Antoine

wetneb:
Yeah, not on my Debian, but that's a cosmetic issue that should not be a blocker for @lisa761's work.

tfmorris:
@antoine2711 If this:

> I've put back « page(s) » in french, let's deal with that some other time.

implies that you don't want to help test French plurals support, I really wish you'd reconsider, since I spent the time fixing it specifically for you.

main/webapp/modules/core/langs/translation-fr.json
```diff
@@ -537,6 +537,7 @@
     "core-views/view": "Aperçu",
     "core-views/extend-not-supported": "Ce service de réconciliation de supporte pas l'extension de données. Essayez de supprimer le service et de l'ajouter à nouveau. Si le problème persiste, contactez le fournisseur de service.",
     "core-views/to-text": "En texte",
+    "core-views/goto-page": "$1 de $2 page(s)",
```
tfmorris, line 540: This should be pluralized the same way Spanish and English is.

antoine2711:
> @antoine2711 If this:
> 
> > I've put back « page(s) » in french, let's deal with that some other time.
> 
> implies that you don't want to help test French plurals support, I really wish you'd reconsider, since I spent the time fixing it specifically for you.

Hi @tfmorris, please, when you see one of my action and you don't think it's inline with « OR standard » or your way of seeing things, do NOT assume it's true or it's my intention.

In this particular case, the reason I removed this (after putting IT BACK a few days ago), it's because @wetneb wanted this issue resolve for his GSoS.

Also, I didn't want to make « noise » here about this, since it's not directly related. My intentions are to tackle this with PR #2717, where you will soon see me requesting your attention in order to fixed. From there, I will then fix this.

Regards,
   Antoine

P.S. 2 things, OT, but still: 
1. I have much respect for all the work I see you doing in the last 2 months. I have the feeling you have done in that period of time 6 months worth of work.
2. You live in Boston, right? You might have heard of french canadians nationalists (just North) who love their language and culture, and who work hard to promote it, or just have on par with other languages/cultures. I will probably never miss the opportunity of making french a first class citizen in the languages world. ;-)

PR #2641, opened 2020-05-25
Add conflicts-with constraint

darecoder1999:
Implemented conflicts-with scrutinizer as part of #2354

### DIFF
--- a/extensions/wikidata/module/langs/translation-en.json
+++ b/extensions/wikidata/module/langs/translation-en.json
@@ -104,6 +104,8 @@
     "warnings-messages/remove-statements-with-invalid-format/body": "If these statements currently exist on Wikidata, this will solve constraint violations.",
     "warnings-messages/missing-inverse-statements/title": "Inverse statements missing for {added_property_entity}.",
     "warnings-messages/missing-inverse-statements/body": "Any {added_property_entity} statement such as the one from {source_entity} to {target_entity} should also be added in reverse with {inverse_property_entity}: in this case, {target_entity} {inverse_property_entity} {source_entity}.",
+    "warnings-messages/having-conflicts-with-statements/title": "{property_entity} conflicts with statement {added_property_entity}.",
+    "warnings-messages/having-conflicts-with-statements/body": "Remove the conflicting statements with {property_entity} such as item with property {example_entity}.",
     "warnings-messages/self-referential-statements/title": "Self-referential statements.",
     "warnings-messages/self-referential-statements/body": "While not forbidden, self-referential statements are generally suspicious. You have some on {example_entity}.",
     "warnings-messages/unsourced-statements/title": "Statements without references.",
--- a/extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
@@ -28,6 +28,8 @@ import org.wikidata.wdtk.datamodel.interfaces.PropertyIdValue;
 import org.wikidata.wdtk.datamodel.interfaces.QuantityValue;
 import org.wikidata.wdtk.datamodel.interfaces.Value;
 
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 /**
@@ -142,6 +144,7 @@ public interface ConstraintFetcher {
     boolean usableOnItems(PropertyIdValue pid);
 
     /**
+<<<<<<< HEAD
      * Retrieves the lower bound of the range
      * required in difference-within-range constraint
      *
@@ -174,4 +177,14 @@ public interface ConstraintFetcher {
      * with its lower bound property should be in a range?
      */
     boolean hasDiffWithinRange(PropertyIdValue pid);
+
+     /*
+     * Returns the Map of all the conflicting pid and their item values
+     *
+     * @param pid:
+     *            the property having conflicts-with constraint
+     * @return
+     */
+    Map<PropertyIdValue, List<Value>> getParamConflictsWith(PropertyIdValue pid);
+
 }
--- a/extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
@@ -28,9 +28,7 @@ import org.wikidata.wdtk.datamodel.helpers.Datamodel;
 import org.wikidata.wdtk.datamodel.implementation.QuantityValueImpl;
 import org.wikidata.wdtk.datamodel.interfaces.*;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -94,7 +92,10 @@ public class WikidataConstraintFetcher implements ConstraintFetcher {
     public static String ALLOWED_ENTITY_TYPES_QID = "Q52004125";
     public static String ALLOWED_ITEM_TYPE_QID = "Q29934200";
     public static String ALLOWED_ENTITY_TYPES_PID = "P2305";
-    
+
+    public static String CONFLICTS_WITH_CONSTRAINT_QID = "Q21502838";
+    public static String CONFLICTS_WITH_PROPERTY_PID = "P2306";
+    public static String ITEM_OF_PROPERTY_CONSTRAINT_PID = "P2305";
 
     // The following constraints still need to be implemented:
 
@@ -423,4 +424,36 @@ public class WikidataConstraintFetcher implements ConstraintFetcher {
         return null;
     }
 
+    /**
+     * Returns the Map of all the conflicting pid and their item values
+     *
+     * @param pid:
+     *            the property having conflicts-with constraint
+     * @return
+     */
+    @Override
+    public Map<PropertyIdValue, List<Value>> getParamConflictsWith(PropertyIdValue pid) {
+        List<Statement> statementList = getConstraintsByType(pid, CONFLICTS_WITH_CONSTRAINT_QID).collect(Collectors.toList());
+        Map<PropertyIdValue, List<Value>> propertyIdValueListMap = new HashMap<>();
+        for (Statement statement : statementList) {
+            List<SnakGroup> specs = statement.getClaim().getQualifiers();
+            PropertyIdValue conflictingPid = null;
+            List<Value> items = new ArrayList<>();
+            for(SnakGroup group : specs) {
+                for (Snak snak : group.getSnaks()) {
+                    if (group.getProperty().getId().equals(CONFLICTS_WITH_PROPERTY_PID)){
+                        conflictingPid = (PropertyIdValue) snak.getValue();
+                    }
+                    if (group.getProperty().getId().equals(ITEM_OF_PROPERTY_CONSTRAINT_PID)){
+                        items.add(snak.getValue());
+                    }
+                }
+            }
+            if (conflictingPid != null) {
+                propertyIdValueListMap.put(conflictingPid, items);
+            }
+        }
+
+        return propertyIdValueListMap;
+    }
 }
--- a/extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizer.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizer.java
@@ -1,0 +1,61 @@
+package org.openrefine.wikidata.qa.scrutinizers;
+
+import org.openrefine.wikidata.qa.QAWarning;
+import org.openrefine.wikidata.updates.ItemUpdate;
+import org.wikidata.wdtk.datamodel.interfaces.PropertyIdValue;
+import org.wikidata.wdtk.datamodel.interfaces.Statement;
+import org.wikidata.wdtk.datamodel.interfaces.Value;
+
+import java.util.*;
+
+public class ConflictsWithScrutinizer extends EditScrutinizer {
+
+    public static final String type = "having-conflicts-with-statements";
+
+    @Override
+    public void scrutinize(ItemUpdate update) {
+        Map<PropertyIdValue, Set<Value>> propertyIdValueValueMap = new HashMap<>();
+        for (Statement statement : update.getAddedStatements()){
+            PropertyIdValue pid = statement.getClaim().getMainSnak().getPropertyId();
+            Value value = statement.getClaim().getMainSnak().getValue();
+            Set<Value> values;
+            if (value != null) {
+                if (propertyIdValueValueMap.containsKey(pid)) {
+                    values = propertyIdValueValueMap.get(pid);
+                } else {
+                    values = new HashSet<>();
+                }
+
+                values.add(value);
+                propertyIdValueValueMap.put(pid, values);
+            }
+        }
+
+        for(PropertyIdValue propertyId : propertyIdValueValueMap.keySet()){
+            Map<PropertyIdValue, List<Value>> conflictingPropertyMap = _fetcher.getParamConflictsWith(propertyId);
+            for (PropertyIdValue conflictingPid : conflictingPropertyMap.keySet()) {
+                if (propertyIdValueValueMap.containsKey(conflictingPid) && raiseWarning(propertyIdValueValueMap, conflictingPid, conflictingPropertyMap)) {
+                    QAWarning issue = new QAWarning(type, propertyId.getId()+conflictingPid.getId(), QAWarning.Severity.WARNING, 1);
+                    issue.setProperty("property_entity", propertyId);
+                    issue.setProperty("added_property_entity", conflictingPid);
+                    issue.setProperty("example_entity", update.getItemId());
+                    addIssue(issue);
+                }
+            }
+        }
+    }
+
+    private boolean raiseWarning(Map<PropertyIdValue, Set<Value>> propertyIdValueValueMap, PropertyIdValue conflictingPid, Map<PropertyIdValue, List<Value>> conflictingPropertyMap) {
+        if (conflictingPropertyMap.get(conflictingPid).isEmpty()){
+            return true;
+        }
+
+        for (Value value : conflictingPropertyMap.get(conflictingPid)) {
+            if (propertyIdValueValueMap.get(conflictingPid).contains(value)){
+                return true;
+            }
+        }
+
+        return false;
+    }
+}
--- a/extensions/wikidata/tests/src/org/openrefine/wikidata/qa/MockConstraintFetcher.java
+++ b/extensions/wikidata/tests/src/org/openrefine/wikidata/qa/MockConstraintFetcher.java
@@ -33,6 +33,9 @@ import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Set;
+import org.wikidata.wdtk.datamodel.interfaces.*;
+
+import java.util.*;
 import java.util.stream.Collectors;
 
 public class MockConstraintFetcher implements ConstraintFetcher {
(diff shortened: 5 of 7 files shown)


extensions/wikidata/tests/src/org/openrefine/wikidata/qa/MockConstraintFetcher.java
```diff
@@ -59,6 +60,11 @@
     
     public static PropertyIdValue propertyOnlyPid = Datamodel.makeWikidataPropertyIdValue("P372");
 
+    public static PropertyIdValue conflictsWithPid = Datamodel.makeWikidataPropertyIdValue("P50");
+    public static PropertyIdValue pidConflictingStatement = Datamodel.makeWikidataPropertyIdValue("P31");
+    public static Value conflictingStatementValue = Datamodel.makeWikidataItemIdValue("P31");
```
wetneb, line 65: `Datamodel.makeWikidataItemIdValue("P31")` is incorrect, you are trying to create an `ItemIdValue` with a property identifier.

extensions/wikidata/module/langs/translation-en.json
```diff
@@ -104,6 +104,8 @@
     "warnings-messages/remove-statements-with-invalid-format/body": "If these statements currently exist on Wikidata, this will solve constraint violations.",
     "warnings-messages/missing-inverse-statements/title": "Inverse statements missing for {added_property_entity}.",
     "warnings-messages/missing-inverse-statements/body": "Any {added_property_entity} statement such as the one from {source_entity} to {target_entity} should also be added in reverse with {inverse_property_entity}: in this case, {target_entity} {inverse_property_entity} {source_entity}.",
+    "warnings-messages/having-conflicts-with-statements/title": "{property_entity} is having conflicts with statement {added_property_entity}.",
```
wetneb, line 107: I do not think the "is having" is necessary here: `"{property_entity} conflicts with {added_property_entity}"` is simpler.

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
```diff
@@ -340,4 +333,47 @@ public boolean usableOnItems(PropertyIdValue pid) {
...
     }
+
+    /**
+     * Returns the list of PropertyIdValues of conflicting statements
+     *
+     * @param pid:
+     *            the property having conflicts-with constraint
+     * @return
+     */
+    public List<Value> getConflictsWithProperties (PropertyIdValue pid){
```
wetneb, line 344: Style: no space between the method name and the opening parenthesis, one space between the closing parenthesis and the opening curly brace.

extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
```diff
@@ -134,4 +135,27 @@
...
+    List<Value> getConflictsWithProperties (PropertyIdValue pid);
+
+    /**
+     * Returns the list of all the conflicting values
+     *
+     * @param pid:
+     *            the property having conflicts-with constraint
+     * @return
+     */
+    List<Value> getItemwithConflicts (PropertyIdValue pid);
```
wetneb, line 155: Didn't we agree that we needed two parameters here: the property on which the constraint is defined, and the property it conflicts with?

darecoder1999: That's what I want to discuss with you, that If we are following wikibase quality constraints implementation so I guess they have used the same kind of implementation, they are fetching the item list at once and then checking that if the conflicting property is having a value from that list. https://github.com/wikimedia/mediawiki-extensions-WikibaseQualityConstraints/blob/master/src/ConstraintCheck/Checker/ConflictsWithChecker.php#L119-L139

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
darecoder1999: Okay, I'll do the changes.

extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
wetneb: Their architecture is quite different from what we have here: yes they fetch items in a single go, but in this context they are already looking at a single constraint definition anyway. We should make sure examples like https://www.wikidata.org/wiki/Property:P4013 are handled correctly.

wetneb: Happy to discuss that at our next meeting on tuesday if that helps :)

extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
```diff
@@ -134,4 +135,27 @@
...
     boolean usableOnItems(PropertyIdValue pid);
+
+    /**
+     * Returns the list of PropertyIdValues of conflicting statements
+     *
+     * @param pid:
+     *            the property having conflicts-with constraint
+     * @return
+     */
+    List<Value> getConflictsWithProperties (PropertyIdValue pid);
```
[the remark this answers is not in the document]
darecoder1999: I did that so I can make use of findValues method else we have to create a new method doing nearly same thing.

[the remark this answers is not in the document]
wetneb: I think it is better to have a clean signature for an interface: you can convert the values with a map (from the streaming API).

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
```diff
@@ -340,4 +333,40 @@ public boolean usableOnItems(PropertyIdValue pid) {
...
+
+    /**
+     * Returns the list of PropertyIdValues of conflicting statements
+     *
+     * @param pid:
+     *            the property having conflicts-with constraint
+     * @return
+     */
+    public List<PropertyIdValue> getConflictsWithProperties(PropertyIdValue pid) {
+        List<SnakGroup> specs = getSingleConstraint(pid, CONFLICTS_WITH_CONSTRAINT_QID);
```
wetneb, line 345: Because you are using `getSingleConstraint` here, this means that for properties with multiple conflicts-with constraints, you will only check the first one.

darecoder1999: Okay, let me try the alternatives. Thanks for the review. :)

extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
darecoder1999: @wetneb: I have made the changes proposed, still finding an issue. When there are multiple issues to be reported for conflicts with constraint only the one is reported at once. I have checked that the raiseWarning method is returning true for all the possible issues but still the issues are not flagged in the issues tab. Please help. Thank you.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizer.java
```diff
@@ -0,0 +1,58 @@
...
+public class ConflictsWithScrutinizer extends EditScrutinizer {
+
+    public static final String type = "having-conflicts-with-statements";
+
+    @Override
+    public void scrutinize(ItemUpdate update) {
+        Map<PropertyIdValue, Value> propertyIdValueValueMap = new HashMap<>();
+        for (Statement statement : update.getAddedStatements()){
+            PropertyIdValue pid = statement.getClaim().getMainSnak().getPropertyId();
+            Value value = statement.getClaim().getMainSnak().getValue();
```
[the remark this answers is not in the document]
darecoder1999: In the case of NoValueSnak & SomeValueSnak, it will make a call to the method implemented in SnakImpl which will return null. So, I guess it will not create any issues. If it is an issue, please let me know, I'll use some other way to implement the functionality.

darecoder1999:
@wetneb: Please review the changes made. I have removed all the bugs I have encountered until now. This one was great fun :)

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
```diff
@@ -337,4 +342,57 @@ public boolean usableOnItems(PropertyIdValue pid) {
...
+     */
+    public List<Value> getItemWithConflicts(PropertyIdValue pid, PropertyIdValue conflictingPid) {
+        List<Statement> statementList = getConstraintsByType(pid, CONFLICTS_WITH_CONSTRAINT_QID).collect(Collectors.toList());
+        List<Value> itemList = new ArrayList<>();
+        for (Statement statement : statementList) {
+            List<SnakGroup> specs = statement.getClaim().getQualifiers();
+            boolean flag = false;
+            for(SnakGroup group : specs) {
+                for (Snak snak : group.getSnaks()) {
+                    if (snak.getValue() instanceof PropertyIdValueImpl && snak.getValue().equals(conflictingPid)){
```
wetneb, line 384: Two things: it is better to avoid referring to `PropertyIdValueImpl`, you can just use `PropertyIdValue` which is the interface. Also, this type check is actually redundant with the equality check afterwards. You should just be able to do `conflictingPid.equals(snak.getValue())` (since you know `conflictingPid` is not null and its equals method will only return true for instances of `PropertyIdValue`).

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
```diff
@@ -337,4 +342,57 @@ public boolean usableOnItems(PropertyIdValue pid) {
...
+                        flag = true;
+                    } else if (snak.getValue() instanceof PropertyIdValueImpl && !snak.getValue().equals(conflictingPid)) {
```
wetneb, line 386: Same here.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizer.java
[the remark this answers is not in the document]
wetneb: That is fine then! It would be worth adding a test case with a `NoValueSnak` to demonstrate that, perhaps?

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
darecoder1999: Changing PropertyIdValueImpl to PropertyIdValue sounds good to me but removing that check will not be good I guess. Because I have used that as in Snaks we get both Property and Item so we have to update flag value only in case when we get Property as Snak. So, in the 1st if condition it will work but in 2nd else if part it will not work as expected.

wetneb: Can you give me an example of a case where `snak.getValue().equals(conflictingPid)` evaluates to `true` but `snak.getValue() instanceof PropertyIdValue` does not?

darecoder1999: As I mentioned, it will work for sure but in the 2nd else if condition `snak.getValue() instanceof PropertyIdValue && !snak.getValue().equals(conflictingPid)`  if we remove the check to check the instance class then there might be the case that it belongs to ItemIdValue and it will make get executed as this will be true in that case `!snak.getValue().equals(conflictingPid)`. 

As an alternative, if we really want to remove the instanceof check then one thing that comes to mind is we can place the last if condition at 2nd position. It will be better I guess?

wetneb: If `snak.getValue()` is an `ItemIdValue` then `snak.getValue().equals(conflictingPid)` will evaluate to `false` - check the `equals` implementation of `ItemIdValue` :)

darecoder1999: I feel we both are looking at the issue differently.

> If `snak.getValue()` is an `ItemIdValue` then `snak.getValue().equals(conflictingPid)` will evaluate to `false` - check the `equals` implementation of `ItemIdValue` :)

I do agree with your explanation, that's why I'm saying it. If it evaluates to false for other ItemIdValue the `!` at the starting `!snak.getValue().equals(conflictingPid)` will make it true and it will get executed which I don't want for now. I have tried to remove the instanceof comparison but it didn't worked as expected. If you are insisting I'll try it once again for sure.

wetneb: Could you explain what you mean by "it didn't worked as expected"? Perhaps the best thing to do in this case is to put a breakpoint on this `if` and use the debugger to find out what is happening here exactly :)

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
wetneb: Here, instead of relying on the `PropertyIdValue` type of the snak value, you should rather rely on the `snak.getProperty()`'s value.

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
darecoder1999: Because I want the 2nd else if condition to execute only when the case is it is not and ItemIdValue, it is a PropertyIdValue but its value is not equal to the conflictingPid. But when I remove the instanceof check from it, the condition executes even for ItemIdValue as expected according to your explanation. 

> If `snak.getValue()` is an `ItemIdValue` then `snak.getValue().equals(conflictingPid)` will evaluate to `false` - check the `equals` implementation of `ItemIdValue` :)

But I want to it to execute for PropertyIdValue only when its value isn't equal to conflictingPid.

wetneb: > But I want to it to execute for PropertyIdValue only when its value isn't equal to conflictingPid.

Yes but that is not how you should do it: you should look at the property of the snak, not the type of its values (because there could potentially be some other parameter which would also have the same type of values). So: remove all `instanceof` checks and use `snak.getProperty()` to know what sort of parameter you are handling.

darecoder1999: I have removed the instanceof checks. Is it okay now?

wetneb: It is much better! Now I guess we still have the issue that you are relying on the order of the qualifiers (you expect the property first, and then the items, but there is no reason why they should necessarily appear in this order).

darecoder1999: I tried a lot of properties having conflicts-with constraint and observed this pattern, hence proceeded with this implementation. I won't say testes all of them but some of them. The item values don't seem to have anything that relates it to the conflicting Pid I guess. What else I should do? Any suggestions?

wetneb: I would do as follows:
- create a variable to store the conflicting pid and a list of conflicting values
- loop over all qualifier snaks (as you are doing)
  + if the property of the snak is the one that specifies the conflicting pid, then set the conflicting pid to the qualifier value
  + if the property of the snak is the one that specifies the conflicting values, then add the value to the list of conflicting values

darecoder1999: Hope, I correctly understand what you tried to explain. :)

extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
```diff
@@ -142,6 +144,7 @@
     boolean usableOnItems(PropertyIdValue pid);
 
     /**
+<<<<<<< HEAD
```
tfmorris, line 147: This looks like a missed merge conflict

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
```diff
@@ -28,9 +28,7 @@
 import org.wikidata.wdtk.datamodel.implementation.QuantityValueImpl;
 import org.wikidata.wdtk.datamodel.interfaces.*;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
tfmorris, line 31: Did your IDE do this without you asking? Can you configure a higher limit for wild card imports and/or review the diff to make sure unnecessary changes aren't included?

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
```diff
@@ -94,7 +92,10 @@
     public static String ALLOWED_ENTITY_TYPES_QID = "Q52004125";
     public static String ALLOWED_ITEM_TYPE_QID = "Q29934200";
     public static String ALLOWED_ENTITY_TYPES_PID = "P2305";
-    
+
+    public static String CONFLICTS_WITH_CONSTRAINT_QID = "Q21502838";
+    public static String CONFLICTS_WITH_PROPERTY_PID = "P2306";
+    public static String ITEM_OF_PROPERTY_CONSTRAINT_PID = "P2305";
```
tfmorris, line 98: This appears to duplicate `ALLOWED_ENTITY_TYPES_PID` just above.

extensions/wikidata/tests/src/org/openrefine/wikidata/qa/MockConstraintFetcher.java
```diff
@@ -66,6 +69,11 @@
     public static QuantityValue minValuePid = Datamodel.makeQuantityValue(new BigDecimal(0));
     public static QuantityValue maxValuePid = Datamodel.makeQuantityValue(new BigDecimal(150));
 
+    public static PropertyIdValue conflictsWithPid = Datamodel.makeWikidataPropertyIdValue("P50");
+    public static PropertyIdValue pidConflictingStatement = Datamodel.makeWikidataPropertyIdValue("P31");
+    public static Value conflictingStatementValue = Datamodel.makeWikidataItemIdValue("Q5");
+    public static Value conflictsWithValue = Datamodel.makeWikidataItemIdValue("Q36322");
```
tfmorris, line 75: I recognize Q5, but the rest would have to look up. For the benefit of future readers, it might be nice to give them mnemonic variable names or to put the string values in constants.

extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java
wetneb: whoops, thanks for catching that!

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
wetneb: While these happen to have the same value for Wikidata, it might not be true of all Wikibase instances (these constants should become user-configurable once we add proper Wikibase support). That being said the name of the constant could probably be adapted to reflect that it is specifically for that constraint.

extensions/wikidata/src/org/openrefine/wikidata/qa/WikidataConstraintFetcher.java
darecoder1999: Yes, it got updated on its own by the IDE. Sorry, I'll take care of it next time onwards for sure. For now, I'll create a PR resolving all the flaws. Thanks for the review. :)

PR #2657, opened 2020-05-30
Save preferences JSON using UTF-8 encoding. Bulletproof prefs load.

### LINKED ISSUE #2543: Loss of expression history and language at restart
**Describe the bug**

There are reports that OpenRefine 3.3 forgets language settings and expression history when restarted:
https://twitter.com/daieuxdailleurs/status/1248911911036628992

I am unable to reproduce this on 3.3. If anyone could get in touch and ask for more details (platform info, perhaps), it would potentially help to debug this.

 - OS: Windows 64bits [version not specified]
 - Browser Version: FF ou Chrome [no versions]
 - JRE or JDK Version: ?

**OpenRefine:**
 - Version 3.3

### LINKED ISSUE #2544: Loss of Starring an Expression in History
What breaks:
* Star an expression in history

**Details:**
* OpenRefine win 3.3 release
* Windows 10
* JDK 8
* Firefox, Chrome, Edge
* The last part of `workspace.json` is formed as such:
```json
            ],
            "ui.browsing.listFacet.limit": 22000,
            "scripting.starred-expressions": null,
            "userLang": "en",
            "scripting.expressions": {
                "top": 100,
                "list": ["grel:value.startsWith(\"http\")", "grel:value.replace(\"/wiki/Property:\",\"/entity/\")", "grel:value.contains(\"Property\")"],
                "class": "com.google.refine.preference.TopList"
            },
            "ui.show-control-characters": "true"
        }
    }
```

**Error:** 
* Clicking on a Star next to a GREL in the Expression History **does not work** and throws an NPE here:
```
10:06:28.438 [          org.mortbay.log] /command/core/toggle-starred-expression (0ms)
java.lang.NullPointerException
        at com.google.refine.commands.expr.ToggleStarredExpressionCommand.doPost(ToggleStarredExpressionCommand.java:53)
        at com.google.refine.RefineServlet.service(RefineServlet.java:189)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:820)
        at org.mortbay.jetty.servlet.ServletHolder.handle(ServletHolder.java:511)
        at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1166)
        at org.mortbay.servlet.UserAgentFilter.doFilter(UserAgentFilter.java:81)
        at org.mortbay.servlet.GzipFilter.doFilter(GzipFilter.java:132)
<snip>
```

**Additional context:**

Wondering if any leftover TODO's from this issue #2086 ?

### LINKED ISSUE #2627: added Reconciliation Services disappear on restart
**Describe the bug**
All added Reconciliation Services other than the default "Wikidata (en)" disappear on restart.
The only hint I get in the console could be this line:  [FileProjectManager] com.fasterxml.jackson.databind.JsonMappingException: Invalid UTF-8 middle byte 0x72
 at [Source: (File); line: 1, column: 2280] (through reference chain: com.google.refine.io.FileProjectManager["preferences"]->com.google.refine.preference.PreferenceStore["entries"]) (368ms)

This only happens with OpenRefine 3.3 and 3.4 (Beta). I also tested 3.1 and 3.2 and encountered no problems there!

I tested it on two different Windows machines with the same bug persisting on both.

**To Reproduce**
Just add any reconciliation service, e.g. https://lobid.org/gnd/reconcile (GND service), shut down OpenRefine and restart

**Expected behavior**
all added Reconciliation Services should be preserved on restart

 - OS: Windows 10
 - Browser Version: Chrome Version 81.0.4044.138 (64-Bit)
 - JRE or JDK Version: java version "1.8.0_251"

 - Version 3.3 

**Datasets**
all of my datasets are affected

tfmorris:
Fixes #2543 Fixes #2544 Fixes #2627.

Always use UTF-8 to write JSON because platform default encoding
might not be legal JSON (e.g. ISO 8859-1).

Also be more conservative about keeping backups if we fail to write.

Two open questions:
- Are we using the correct filter on write? Looks like it could be wrong.
- Do we want to make sure that we can deserialize the written file before declaring success? That would have saved us from the user data loss due to the write/read mismatch here.

### DIFF
--- a/main/src/com/google/refine/io/ProjectMetadataUtilities.java
+++ b/main/src/com/google/refine/io/ProjectMetadataUtilities.java
@@ -70,11 +70,15 @@ public class ProjectMetadataUtilities {
         File file = new File(projectDir, "metadata.json");
         File oldFile = new File(projectDir, "metadata.old.json");
 
-        if (file.exists() && file.length() > 0) {
-            if (oldFile.exists()) {
-                oldFile.delete();
-            }
+        if (file.exists()) {
+            if(file.length() > 0) {
+                if (oldFile.exists()) {
+                    oldFile.delete();
+                }
             file.renameTo(oldFile);
+            } else {
+                file.delete();
+            }
         }
 
         tempFile.renameTo(file);
--- a/main/src/com/google/refine/io/ProjectMetadataUtilities.java
+++ b/main/src/com/google/refine/io/ProjectMetadataUtilities.java
@@ -39,6 +39,7 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
@@ -60,15 +61,19 @@ public class ProjectMetadataUtilities {
     public static void save(ProjectMetadata projectMeta, File projectDir) throws IOException  {
         File tempFile = new File(projectDir, "metadata.temp.json");
         saveToFile(projectMeta, tempFile);
+        if (tempFile.length() == 0) {
+            throw new IOException("Failed to save project metadata - keeping backups");
+        }
+
+        // TODO Do we want to make sure we can successfully deserialize the file too?
 
         File file = new File(projectDir, "metadata.json");
         File oldFile = new File(projectDir, "metadata.old.json");
 
-        if (oldFile.exists()) {
-            oldFile.delete();
-        }
-        
-        if (file.exists()) {
+        if (file.exists() && file.length() > 0) {
+            if (oldFile.exists()) {
+                oldFile.delete();
+            }
             file.renameTo(oldFile);
         }
 
@@ -76,9 +81,9 @@ public class ProjectMetadataUtilities {
     }
     
     protected static void saveToFile(ProjectMetadata projectMeta, File metadataFile) throws IOException   {
-        Writer writer = new OutputStreamWriter(new FileOutputStream(metadataFile));
+        Writer writer = new OutputStreamWriter(new FileOutputStream(metadataFile), StandardCharsets.UTF_8);
         try {
-            ParsingUtilities.defaultWriter.writeValue(writer, projectMeta);
+            ParsingUtilities.saveWriter.writeValue(writer, projectMeta);
         } finally {
             writer.close();
         }
@@ -121,7 +126,7 @@ public class ProjectMetadataUtilities {
      * creation and modification times based on whatever files are available.
      * 
      * @param projectDir the project directory
-     * @param id the proejct id
+     * @param id the project id
      * @return
      */
     static public ProjectMetadata recover(File projectDir, long id) {
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,7 @@
     <jee.port>3333</jee.port>
     <refine.data>/tmp/refine</refine.data>
     <surefire.version>2.22.2</surefire.version>
-    <surefireArgs></surefireArgs>
+    <surefireArgs>-Dfile.encoding=cp1252</surefireArgs>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <jackson.version>2.11.0</jackson.version>
   </properties>
--- a/main/src/com/google/refine/io/FileProjectManager.java
+++ b/main/src/com/google/refine/io/FileProjectManager.java
@@ -40,6 +40,8 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map.Entry;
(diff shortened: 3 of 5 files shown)


main/src/com/google/refine/io/ProjectMetadataUtilities.java
```diff
@@ -60,25 +61,35 @@
...
         tempFile.renameTo(file);
     }
     
     protected static void saveToFile(ProjectMetadata projectMeta, File metadataFile) throws IOException   {
-        Writer writer = new OutputStreamWriter(new FileOutputStream(metadataFile));
+        Writer writer = new OutputStreamWriter(new FileOutputStream(metadataFile), StandardCharsets.UTF_8);
         try {
             ParsingUtilities.defaultWriter.writeValue(writer, projectMeta);
+            // TODO: Do we want the below instead? Doesn't filter recon candidates
+//            ParsingUtilities.saveWriter.writeValue(writer, projectMeta);
```
wetneb, line 92: There is no reason for recon candidates to end up in the project metadata, so both writers are equivalent. Perhaps we could use `saveWriter` because we are *saving* things to disk, but it is a style change.

main/src/com/google/refine/io/ProjectMetadataUtilities.java
```diff
@@ -60,25 +61,35 @@
...
+                file.delete();
+            }
         }
 
         tempFile.renameTo(file);
     }
     
     protected static void saveToFile(ProjectMetadata projectMeta, File metadataFile) throws IOException   {
-        Writer writer = new OutputStreamWriter(new FileOutputStream(metadataFile));
+        Writer writer = new OutputStreamWriter(new FileOutputStream(metadataFile), StandardCharsets.UTF_8);
```
wetneb, line 88: Perhaps we could look into making the deserializer more flexible with respect to encoding issues (just like it can be configured to accept unescaped tabs)? I have not checked if it is possible though.

wetneb:
Thanks a lot for looking into this. I am a bit puzzled by this - on top of my mind I don't think we enforce UTF-8 anywhere else, so I would have expected this bug to show up earlier (on project data, changes, any other JSON data that we serialize to disk). Perhaps we use other abstractions which enforce UTF-8 themselves? It would be worth reviewing that.

> Do we want to make sure that we can deserialize the written file before declaring success? That would have saved us from the user data loss due to the write/read mismatch here.

If we do it here, I guess we should do it everywhere where data loss could be an issue… It sounds pretty uninviting!

main/src/com/google/refine/io/ProjectMetadataUtilities.java
tfmorris: Jackson does it's own encoding guessing, but it's got to be one of the JSON legal encodings (UTF-8, 16, 32). The ASCII is legal UTF-8, but the upper half of ISO 8859-1 isn't, which is where, I suspect, the problem comes in.

tfmorris:
>  I am a bit puzzled by this - on top of my mind I don't think we enforce UTF-8 anywhere else, so I would have expected this bug to show up earlier (on project data, changes, any other JSON data that we serialize to disk). Perhaps we use other abstractions which enforce UTF-8 themselves? It would be worth reviewing that.

It's really never safe to depend on the default platform encoding, so we should definitely review to make sure there aren't other places where we do that, but since we haven't seen other issues, I suspect we're OK.

The reason that the problem likely wasn't more prevalent is that it's a) Windows only and b) requires non-ASCII data in the preferences to trigger it.  Hmmm, speaking of which, I guess I should write a test case for it too.

wetneb:
Yes but I would expect we serialize non-ASCII data in many places in the data grid, for instance, and I never heard of project data being corrupted on Windows…

main/src/com/google/refine/io/ProjectMetadataUtilities.java
tfmorris: I switched to `saveWriter`

tfmorris:
OK, I think this is ready for review @wetneb.  I switched all the tests to run with cp1252 encoding (the old Windows default) to catch problems like this in the future and wrote a test that I confirmed fails with this setup against `master`.

Good thing I added the test because I found that the problem I fixed wasn't actually the one causing the issue, so I fixed the second bit of code and did a non-exhaustive review of other places where JSON serializations are saved to file and all the rest appear to have the correct UTF-8 encoding set.

I did **not** go the extra mile of making sure that we can re-read the `workspace.json` after we write it out.

In terms of severity, I think this will affect any Windows user who has non-ASCII data in any of:
- expression history (column names, replacement strings, etc)
- preferences
- reconciliation service metadata (e.g. GND reconciliation service)

pom.xml
```diff
@@ -57,7 +57,7 @@
     <jee.port>3333</jee.port>
     <refine.data>/tmp/refine</refine.data>
     <surefire.version>2.22.2</surefire.version>
-    <surefireArgs></surefireArgs>
+    <surefireArgs>-Dfile.encoding=cp1252</surefireArgs>
```
wetneb, line 60: I guess we should ideally be able to apply this only for specific tests, no? I am not sure how that could be done, though. I am trying to think about the risks we are taking by switching to this encoding everywhere. Intuitively it should help us surface more bugs rather than hide ones which would only happen if the system encoding is UTF-8, so it is probably fine?

tfmorris: That was my thinking. We assume/need UTF-8 generally, but our most popular platform doesn't use that as its default encoding, so neither should our tests. I think all tests should be run with this setting.

PR #2658, opened 2020-05-30
Update issue templates

tfmorris:
Make boilerplate/suggestions into comments so they don't show up in final issue if people forget to edit them out.

Add default labels to identify issues needing triage.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,7 @@
 name: Bug report
 about: Create a report to help us improve OpenRefine
 title: ''
-labels: bug, triage needed
+labels: bug, to be reviewed
 assignees: ''
 
 ---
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,6 +1,9 @@
 ---
 name: Bug report
 about: Create a report to help us improve OpenRefine
+title: ''
+labels: bug, triage needed
```
wetneb, line 5: Is there any difference between `triage needed` and `to be reviewed` or should we merge these two tags?

tfmorris: Happy to use the historical `to be reviewed`. I just wasn't sure of the current semantics.

wetneb: Ok, that's pretty much how we have been using it so far, so let's do that.

Issue #1612, opened 2018-05-18
Add OAuth support to Wikidata extension


wetneb:
With @lucaswerkmeister we are exploring hosting OpenRefine instances for the Wikidata community. As part of that, we would need to migrate Wikidata authentication to OAuth.

https://github.com/scribejava/scribejava/pull/852 adds support for MediaWiki OAuth to the scribejava OAuth library. We also need support in Wikidata-Toolkit (see https://github.com/Wikidata/Wikidata-Toolkit/issues/268).

--------------------------
_This is a [proposed Google Summer of Code project in 2020](https://github.com/OpenRefine/OpenRefine/wiki/GSoC-2020%C2%A0Ideas). If you are **not** planning to apply for an internship via GSoC, we kindly ask that you do not work on this task yet, in order to leave the floor to potential interns._

lucaswerkmeister:
Note that, as far as I’m aware, it’s not yet clear what this will mean for the instances we’re *not* hosting :) I currently see two possibilities:

1. Keep username/password authentication and use that if no OAuth consumer has been configured.
2. Automatically register an owner-only consumer (which does not require review by Wikidata administrators) and use that if no other OAuth consumer has been configured.

lucaswerkmeister:
(The advantage of option 2 would be that we don’t need to store the user’s password indefinitely, we only need it once to register the owner-only consumer and then store the consumer/access key/secret instead.)

lucaswerkmeister:
BTW the Phabricator task for a hosted OpenRefine instance is [T194767](https://phabricator.wikimedia.org/T194767).

wetneb:
I thought about the workflow for the consumer-only tokens and it still looks a bit blurry to me. To request an owner-only token, I assume you must be logged in as the owner. So, the first time a user uses OpenRefine, they would first need to login with login / password via OpenRefine so that we can request n owner-only token, and then have them go through the OAuth process with these credentials… so they would log in twice.

lucaswerkmeister:
Yeah, and apparently owner-only tokens don’t support the usual “three-legged OAuth flow”, you *have* to use the “one-legged OAuth flow” instead, so we won’t even be able to reuse all the code from full OAuth… it’s probably much simpler to just stick to normal username/password authentication for local instances.

wetneb:
[...]

Issue #2015, opened 2019-04-10
Support Authentication in Reconciliation API


broom9:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
It would be helpful to support the Authorization HTTP header when adding a reconciliation service. In this way, a private reconciliation service can be supported.

Currently, Authorizaiton and other headers are supported when adding a column by fetching URLs, but not supported when adding a reconciliation service.

**Describe the solution you'd like**
Support HTTP header customization when adding reconciliation services.

**Describe alternatives you've considered**
Workarounds I considered include to deploy the private reconciliation service in a private network or to use URL parameters for authentication.

**Additional context**
Adding a screenshot of the current HTTP header customization function in the "Add Column by Fetching URL" function.
<img width="457" alt="screenshot-hucdk" src="https://user-images.githubusercontent.com/28372/55908068-70ff7780-5ba6-11e9-8917-e62464660db1.png">

wetneb:
See related PR #758.

broom9:
@wetneb thanks. It's indeed a very similar issue. But looks like that pull request didn't make to the master branch?

wetneb:
Indeed not! If you want to give it a go, it could be a starting point, though.

wetneb:
The specific mechanism to use (headers, parameters…) is being discussed in the W3C group:
https://lists.w3.org/Archives/Public/public-reconciliation/2019Jul/0010.html

PR #2661, opened 2020-06-01
OAuth support for the Wikidata extension

afkbrb:
See #1612


afkbrb:
Hi @wetneb @tfmorris, the GSoC coding period has started, I'll work on this PR to add OAuth support to the Wikidata extension, so you can see my progress here.

afkbrb:
Hi @wetneb @tfmorris , I've finished the OAuth prototype with [commit#de080e6](https://github.com/OpenRefine/OpenRefine/pull/2661/commits/de080e60e804dd98189551e31c973b5a84bf1ec7). I've tested it locally and written some unit tests for it. You can see that I've uploaded some edits using the Wikidata extension with OAuth:

openrefine-wikidata-oauth

You can test it with the following steps:

1. install the latest [Wikidata-Toolkit](https://github.com/Wikidata/Wikidata-Toolkit) locally, since the version with OAuth functionality hasn't been released it. You need to set the version to "0.12.0".

2. propose an OAuth consumer at https://meta.wikimedia.org/wiki/Special:OAuthConsumerRegistration/propose, the callback should be set to "http://localhost:3333/command/wikidata/authorized". For "Applicable grants", check "High-volume editing", "Edit existing pages" and "Create, edit, and move pages" (I still need to figure out the rights needed).

3. configure WIKIDATA_CLIENT_ID and WIKIDATA_CLIENT_SECRET in `refine.ini`. Or you can manually set the environment variables "ext.wikidata.clientid" and "ext.wikidata.clientsecret".

4. use the Wikidata extension to login and upload some edits to Wikidata. You should see this dialog when you login to Wikidata:

oauth-login

extensions/wikidata/module/scripts/dialogs/manage-account-dialog.js
```diff
@@ -33,89 +45,135 @@ ManageAccountDialog.display = function(logged_in_username, saved_credentials, ca
...
+              function () {
+                $.get(
+                    "command/wikidata/login",
+                    function (data) {
+                      if (data.logged_in) {
+                        dismiss();
+                        callback(data.username);
+                      } else {
+                        frame.show();
+                        elmts.invalidCredentials.text("Invalid credentials.");
```
tfmorris, line 89: User visible strings need to be internationalized using `$.i18n()`

extensions/wikidata/src/org/openrefine/wikidata/commands/AuthorizeCommand.java
```diff
@@ -0,0 +1,39 @@
...
+                String remember = request.getParameter("remember-credentials");
+                manager.setRememberCredentials("on".equals(remember));
+                String authorizationUrl = manager.getAuthorizationUrl();
+                response.sendRedirect(authorizationUrl);
+            } else {
+                throw new IllegalStateException("You must configure Wikidata OAuth client id/secret in refine.ini " +
+                        "in order to use OAuth to login");
+            }
+        } catch (Exception e) {
+            // TODO: use respondWithErrorPage after fixing its relative path bug
```
tfmorris, line 31: Is this an OpenRefine bug or Wikidata Toolkit bug?

extensions/wikidata/src/org/openrefine/wikidata/commands/AuthorizeCommand.java
```diff
@@ -0,0 +1,39 @@
...
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException {
+        ConnectionManager manager = ConnectionManager.getInstance();
+        try {
+            if (manager.supportOAuth()) {
+                String remember = request.getParameter("remember-credentials");
+                manager.setRememberCredentials("on".equals(remember));
+                String authorizationUrl = manager.getAuthorizationUrl();
+                response.sendRedirect(authorizationUrl);
+            } else {
+                throw new IllegalStateException("You must configure Wikidata OAuth client id/secret in refine.ini " +
```
tfmorris, line 27: This probably should be an error code or something which gets interpretted on the front end so that we can take advantage of the i18n support there.

extensions/wikidata/src/org/openrefine/wikidata/commands/AuthorizeCommand.java
afkbrb: OpenRefine bug. Reported at #2678

extensions/wikidata/src/org/openrefine/wikidata/editing/ConnectionManager.java
```diff
@@ -168,4 +203,12 @@ protected BasicApiConnection createNewConnection() {
         conn.setReadTimeout(READ_TIMEOUT);
         return conn;
     }
+
+    public String getAuthorizationUrl() throws InterruptedException, ExecutionException, IOException {
+        System.out.println("Fetching the Request Token...");
```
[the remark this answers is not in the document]
afkbrb: Thanks for the review. I'll try using logger.debug() in later developing.

afkbrb:
@wetneb @tfmorris In the implementation of [commit#de080e6](https://github.com/OpenRefine/OpenRefine/pull/2661/commits/de080e60e804dd98189551e31c973b5a84bf1ec7), I use `ConnectionManager` to hold an ` ApiConnnection` instance (can be `BasicApiConnection` or `OAuthApiConnection`). This way, only one connection is supported at the same time. This makes sense since OpenRefine currently is not for multiple users to share, and it also makes it possible to save/restore the login credentials (username/password or OAuth access token/secret) to/from the preferences. 

 But this is a weird design. To enable the OAuth functionality, the user has to register an OAuth consumer (the consumer cannot be [owner-only consumer](https://www.mediawiki.org/wiki/OAuth/Owner-only_consumers), since for owner-only consumers, the usual authorization flow will not work). Typically, OpenRefine is used locally, so the user must set the callback URL to, for example, "http://localhost:3333/command/wikidata/authorized". But as mentioned at [the OAuth consumer regsitration of wikimedia](https://meta.wikimedia.org/wiki/Special:OAuthConsumerRegistration/propose):
> Proposals which look like test apps (have "test" in the name or use localhost as the callback URL) will be ignored; you can use them with your own account, for 30 days.

So the consumer will be invalid after a month, unless OpenRefine is hosted in public and the registration is approved. For local use, it's better to use the owner-only consumer. In that case, to login, the user can just give OpenRefine the consumer id/secret and the access token/secret, just like the username/password.

So I'm going to change the design (which is similar to the design in first version of my GSoC proposal):

1. If consumer id and consumer secret are not configured in `refine.ini`, keep the original login dialog, and add a new form to support login with owner-only consumer:
proposal-2
In this case, two-legged OAuth is supported. We still use `ConnectionManager` to hold one connection and can save/restore the credentials.

2. Otherwise, three-legged OAuth is supported. This is about enabling multiple users to uploading edits to Wikidata at the same time, it's still far from full multi-user support (for example, access control and collaborative editing are not supported). In this case, we only provide the following login form:
oauth
If the user clicks on the "Log in" button, the three-legged OAuth flow then begins. The access token/secret will be saved in cookie, and the connection is kept in session. OpenRefine will not serialize the credentials (since the "Remember credentials" option is gone).

wetneb:
@afkbrb that sounds like a very sensible plan! Yes the current design is far from ideal for a multi-user setting, I am glad you are pushing the architecture in the right direction.

refine.bat
```diff
@@ -188,6 +188,11 @@ if "%GDATA_CLIENT_SECRET%" == "" goto skipGDataCredentials
 set OPTS=%OPTS% -Dext.gdata.clientid=%GDATA_CLIENT_ID% -Dext.gdata.clientsecret=%GDATA_CLIENT_SECRET%
 :skipGDataCredentials
 
+if "%WIKIDATA_CLIENT_ID%" == "" goto skipWikidataCredentials
+if "%WIKIDATA_CLIENT_SECRET%" == "" goto skipWikidataCredentials
```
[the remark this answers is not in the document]
afkbrb: Emm, the if statement doesn't support logical operators (https://stackoverflow.com/a/2143203/10464202), but we can rewrite it like 
```
if not "%WIKIDATA_CLIENT_ID%" == "" if not "%WIKIDATA_CLIENT_SECRET%" == "" set OPTS=%OPTS% -Dext.wikidata.clientid=%WIKIDATA_CLIENT_ID% -Dext.wikidata.clientsecret=%WIKIDATA_CLIENT_SECRET%
```
But that line will be too long.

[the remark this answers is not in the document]
tfmorris: Oops, sorry! I was thinking of shell scripts and totally missed the fact that it's the Windows .bat file.

It's been so long since I've written Windows batch files that I'm not an expert at all, but if something like this would do the trick:

```
if not "%WIKIDATA_CLIENT_ID%" == "" (
  if not "%WIKIDATA_CLIENT_SECRET%" == "" (
    set OPTS=%OPTS% -Dext.wikidata.clientid=%WIKIDATA_CLIENT_ID% -Dext.wikidata.clientsecret=%WIKIDATA_CLIENT_SECRET%
  )
)
```

I think it would be slightly cleaner control flow, but don't spend more than 5 minutes on it.

tfmorris:
I like the new proposal better. It's much cleaner to store the secrets client side than server side. I'd even go as far as suggesting that perhaps we don't want to allow the saving of passwords in `refine.ini` since we have no way to encrypt them reliably.

refine.bat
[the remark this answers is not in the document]
afkbrb: Terrific! I didn't notice that we can use parentheses here...

afkbrb:
To host OpenRefine in public, we'll only support using OAuth to login, the host must configure the consumer id/secret somewhere that OpenRefine can reach. refine.ini should be visible only to the host, so I think it's safe to keep the credentials in refine.ini, though encrypting them would be better.

afkbrb:
Hi @wetneb @tfmorris , I've almost finished the new design. But I don't know how to deal with the serialization/deserialization of operations properly.

To be more specific, we are changing the wikidata extension to support multiple users to upload edits to Wikidata at the same time, so each [PerformWikibaseEditsOperation](https://github.com/OpenRefine/OpenRefine/blob/master/extensions/wikidata/src/org/openrefine/wikidata/operations/PerformWikibaseEditsOperation.java) must be bind to a user/connection. But how to serialize/deserialize the operation with the connection properly? I think it's not secure to include the connection in the serializing result. So when deserializing, we need to provide the operation with a connection somehow. The deserializing logic is in [ApplyOperationsCommand](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/commands/history/ApplyOperationsCommand.java), which is in the main module. We could check if the deserializing result is an instance of  PerformWikibaseEditsOperation, if it is, we could get the connection from session and set it to the result. But in this way, we are mixing the main module and the wikidata extension together, so it's not a good solution. 

Maybe we could add an " initWithRequest(HttpServletRequest request)" method to the [AbstractOperation](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/model/AbstractOperation.java) class, the method is used to init an operation with the data from the request (such as the connection we need). This method should be implemented in AbstractOperation as an empty method, since most operations now don't need to be inited with the request. But an operation can choose to override it if needed. For example, for the PerformWikibaseEditsOperation, it can override the method like this:

```java
    @Override
    public void  initWithRequest(HttpServletRequest request) {
        connection = (ApiConnection) request.getSession().getAttribute("wikidata-connection");
    }
```
So after it is deserialized, it can get the connection needed from the request.

Do you think this is a proper solution? Or can we solve this in a better way? Looking forward to your suggestions.

wetneb:
This is an interesting problem! I would say that full multi-user support is beyond the scope of this PR. It is of course useful to think about it as we are doing things, but it is good to keep the scope of the PR as small as it can be.

To address your point specifically: the JSON serialization of operations must be changed with a lot of care, because this JSON format is exposed to the user (when they export their workflow in the Undo/Redo tab). So the crucial question here is: when exporting a workflow to a JSON blob, do we want to also serialize the connection information? I would say no, because this information is generally private, and users expect to be able to share JSON workflows, or re-execute them when logged in via other methods.

At the moment, when a user applies a JSON workflow, we just expect them to be logged in already and we fetch the connection via a static instance. If they are not logged in, the operation will fail. This is of course not ideal, but is mostly down to the fact that operations do not have a good API to expose their requirements and dependencies before being applied (so many other operations will fail similarly, for instance if the column they are supposed to be applied on does not exist).

For now I propose that we keep this architecture: the connection is fetched from the `ConnectionManager`, which can return either a password-based connection instance or an OAuth one, depending on how the user has logged in. We can (and should) change this later on, but I would say that this problem is distinct from adding OAuth support.

tfmorris:
If we go that route, I think we might want to create a separate subclass of `AbstractOperation` with a different constructor, similar to the way that `EngineDependentOperation`, but it seems unusual that this is an `Operation` at all, rather than a `Command`, since it doesn't modify the data. My apologies for not noticing this sooner. 

Both the Freebase [UploadDataCommand](https://github.com/OpenRefine/OpenRefine/blob/ee38f9edde3041a0bb1287a0faa7e1c8510f6641/extensions/freebase/src/com/google/refine/freebase/commands/UploadDataCommand.java), which this is probably most analogous to, and the various cloud exports in the [GData module](https://github.com/OpenRefine/OpenRefine/blob/515d7019731ba721f96e4d086ab2441d6509f517/extensions/gdata/src/com/google/refine/extension/gdata/UploadCommand.java#L135) are implemented as `Command`s. 

@wetneb is there a specific reason that this is structured as an `Operation` instead of a `Command`?

On the multi-user front, OpenRefine is decidedly not multi-user, so I don't think we should attempt a half solution here. We should minimize the exposure of credentials, but other than that follow the existing pattern of it being single user.

wetneb:
> @wetneb is there a specific reason that this is structured as an Operation instead of a Command?

Yes, the idea is that uploading the edits is an operation as part of a larger workflow. You could argue that it should be an exporter instead, but we do not have long-running exporters and there are data upload workflows where it makes sense to do multiple uploads (for instance, creation of new items and updates of existing items separately), so capturing this as an exporter would not let JSON workflows capture them.

afkbrb:
> If we go that route, I think we might want to create a separate subclass of AbstractOperation with a different constructor, similar to the way that EngineDependentOperation, but it seems unusual that this is an Operation at all, rather than a Command, since it doesn't modify the data. My apologies for not noticing this sooner.

@tfmorris  Thanks for your suggestion, that could be a good solution for the serialization/deserialization problem. That problem occurs only if we want to support multiple connections at the same time, but now I am think that maybe we should still support only one connection/user after reading both your comments.


> For now I propose that we keep this architecture: the connection is fetched from the ConnectionManager, which can return either a password-based connection instance or an OAuth one, depending on how the user has logged in. We can (and should) change this later on, but I would say that this problem is distinct from adding OAuth support.

This way, we don't need to take care of the `PerformWikibaseEditsOperation`. But only one connection is supported at the same time. This is not a big problem because OpenRefine is usually used by a single user. Here's one question: **when will we introduce a full multi-user support to OpenRefine?** I guess that won't happen in the near future that should appear in a 4.x version.

> OAuth support is arguably not very useful in a setting when the rest of the software is decidedly single-user, but there is still a case for it (I use OpenRefine but I don't trust it to handle my password securely, or I want to control the action OpenRefine can do by limiting the scopes of my OAuth application).

 So I begin to think about the final goal of this PR, **do we really need the three-legged OAuth support?**

If we only introduce two-legged OAuth support here (i.e., add support for using owner-only consumer to login), it can already benefit a lot to users who want to limit the rights they give to OpenRefine and don't want to give OpenRefine their passwords. And we only need to deal with the single connection held by ConnectionManager.

Three-legged OAuth makes sense only when multi-user support is available. That's why we have to add a very limited multi-user support here. Ideally, OpenRefine should have a user system first, then we bind a user's Wikidata account to their OpenRefine user account via OAuth. Also, in my new design above, for the three-legged OAuth support, I divide OpenRefine to local mode (single user) and public mode (multiple users), different modes result in different behaviors both in the backend and the frontend. Since the public mode is almost useless for now or in the near future, the design is adding unnecessary complexity to OpenRefine.

I guess we should focus on single-user support on 3.x for now, and only add support for using owner-only consumer to login in this PR. The three-legged OAuth support should be introduced in a 4.x version when multi-user support is avaiable. But this doesn't seem like the goal of #1612. By changing the goal, I'm a little worried about my first evaluation of GSoC :)

We can still use three-legged OAuth support to authorize a single user, as I've already finished with [commit#de080e6](https://github.com/OpenRefine/OpenRefine/pull/2661/commits/de080e60e804dd98189551e31c973b5a84bf1ec7). But if the simpler two-legged OAuth can do the same thing, why don't we choose it?

I really need your suggestions on whether we should change the goal or not.

tfmorris:
@afkbrb It's a little difficult to have this discussion because we are in 3 different geos widely separated in space and time, but please let me reassure you that changing the goal in a mutually agreed way will have absolutely no negative effect on your evaluation. Adapting to evolving requirements is a key part of software engineering.

From my point of view, I think the two-legged OAuth support would be sufficient, but let's hear what @wetneb has to say. Either way, I think we want to preserve the work that you've done on the three-legged support regardless.

As for multi-user support, I think it's orthogonal to the scalability work being targeted for what's being called "4.0," but definitely something bigger than we'd be comfortable fitting into 3.x, so definitely not in the immediate future.

wetneb:
> when will we introduce a full multi-user support to OpenRefine?

That definitely requires a lot of thought as it brings up many important design questions! I agree with @tfmorris that we are highly unlikely to ship a full multi-user support soon, but I would not be opposed that you start thinking about some small steps towards that. But I would really like that we do this in PRs that are as granular as possible.

I think there is a good case for letting operations rely on authentication information obtained from user cookies. The current mechanism to store user credentials is really not satisfactory (storing credentials in plain text in the preferences is bad), so it is worth working on this even if it does not actually enable multi-user support: this would just improve the security of the tool a lot. As we are also thinking about adding authentication support for other operations (such as reconciliation, with endpoints requiring some sort of authentication, #2015), now feels like a good time to start thinking about that more seriously.

> do we really need the three-legged OAuth support?

How about this:
* in this PR we do not add three-legged OAuth support, just the two-legged version. The PR does not need to change the architecture of `ConnectionManager`
* in a separate issue (to be opened) we discuss architectural changes to allow operations to rely on user cookies;
* once we have agreement about this issue we implement it in another PR;
* finally we add three-legged OAuth support building up on the new architecture.

Does this sound like a sensible plan? Of course if we do not reach a consensus around the new architecture we can also skip the last three bullet points and let @afkbrb move on to other aspects of the GSoC project (wikibase support), but in any case we will have the consensual parts of OAuth (two-legged workflows) implemented.

I think this is a great opportunity to get @afkbrb involved in interesting architectural discussions, so I would be keen to take advantage of this opportunity :)

afkbrb:
> * in this PR we do not add three-legged OAuth support, just the two-legged version. The PR does not need to change the architecture of `ConnectionManager`.

> ...but in any case we will have the consensual parts of OAuth (two-legged workflows) implemented.

Yes, that's what I want to do in this PR in the next. 

> The current mechanism to store user credentials is really not satisfactory (storing credentials in plain text in the preferences is bad)

So in the following commits, when we add the two-legged OAuth support, do we still store the credentials in preferences? In the future, when we add multi-user support, user's own credentials should definitely be stored in the client side. But now, if we store the credentials in the server side, `ConnectionManager` is able to read them and try restoring a connection with them automatically. In this case, the user doesn't need to login before replaying a serialized `PerformWikibaseEditsOperation`. If the credentials are store in the client side, the user does need to login before replaying it.

> * in a separate issue (to be opened) we discuss architectural changes to allow operations to rely on user cookies;
> * once we have agreement about this issue we implement it in another PR;

Agree, and I think that could be easy (the proposed `initWithRequest(reqeust)` method above should actually be able to do it).

> * finally we add three-legged OAuth support building up on the new architecture.

Even if we agree on some changes to allow operations to rely on user cookies, it's just solving the serialization/deserialization problem. The three-legged OAuth is still not very useful in the absence of full multi-user support, and I don't expect it to be merged into the master branch. But if you agree, I can create a new branch and push my work on three-legged OAuth support to it for later reference.

tfmorris:
I agree with @wetneb that storing credentials in plain text is undesirable and I don't think it's too much of a burden to ask users to login. I'd rather prioritize security over convenience.

> But if you agree, I can create a new branch and push my work on three-legged OAuth support to it for later reference.

Yes, I think that would be good.

afkbrb:
Hi, I finished the owner-only consumer support just now. 

To test it, we still need to clone https://github.com/Wikidata/Wikidata-Toolkit.git and install it locally. The version should be set to "0.12.0".

I changed the login dialog in the commit. Hope you like it.

login1

login2

afkbrb:
@tfmorris @wetneb , waiting for your review :)

tfmorris:
@afkbrb Sorry for the delay! I've been trying to finish up some critical bug fixes which are blocking the 3.4 release. I will get to it before the end of the weekend.

afkbrb:
No hurry, I just want to remind you that the OAuth support is almost done, I'll begin to prepare for the wikibase support in the following days. The preparation won't take me too much time, so we can still improve the OAuth support during that period.

tfmorris:
When I attempt to build wdtk with the `pom.xml` edited to version 0.12.0, I get the following errors when I try to build with Maven

```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for com.squareup.okhttp3:okhttp-urlconnection:jar must be a valid version but is '${okhttpVersion}'. @ line 38, column 13
[ERROR] 'dependencies.dependency.version' for com.squareup.okhttp3:mockwebserver:jar must be a valid version but is '${okhttpVersion}'. @ line 43, column 13
[ERROR] 'dependencies.dependency.version' for se.akerfeldt:okhttp-signpost:jar must be a valid version but is '${okhttpSignpostVersion}'. @ line 49, column 13
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project org.wikidata.wdtk:wdtk-wikibaseapi:0.11.1 (/Users/tfmorris/git/Wikidata-Toolkit/wdtk-wikibaseapi/pom.xml) has 3 errors
[ERROR]     'dependencies.dependency.version' for com.squareup.okhttp3:okhttp-urlconnection:jar must be a valid version but is '${okhttpVersion}'. @ line 38, column 13
[ERROR]     'dependencies.dependency.version' for com.squareup.okhttp3:mockwebserver:jar must be a valid version but is '${okhttpVersion}'. @ line 43, column 13
[ERROR]     'dependencies.dependency.version' for se.akerfeldt:okhttp-signpost:jar must be a valid version but is '${okhttpSignpostVersion}'. @ line 49, column 13
```

I'm sure I could debug it, but perhaps you could fork that repo, make the version edit and whatever other edits are necessary and send us a link to it so it works out of the box? I'm not going to work on it any more tonight, so no rush.

afkbrb:
Sorry I didn't make it clear. I just tried again, simply replace all "<version>0.11.1</version>" with "<version>0.12.0</version>" should work.

extensions/wikidata/src/org/openrefine/wikidata/editing/ConnectionManager.java
```diff
@@ -23,114 +23,100 @@
...
+ *
  * @author Antonin Delpeuch
+ * @author Lu Liu
  */
 
 public class ConnectionManager {
-    
-    final static Logger logger = LoggerFactory.getLogger("connection_mananger");
 
-    public static final String PREFERENCE_STORE_KEY = "wikidata_credentials";
```
wetneb, line 58: For people upgrading from previous versions, should we try to delete any existing preference setting stored at this key? That would ensure we clean up credentials stored in plain text there.

extensions/wikidata/src/org/openrefine/wikidata/commands/LoginCommand.java
```diff
@@ -23,61 +23,177 @@
...
+            }
+        }
+
+        if (isNotBlank(username) && isNotBlank(password)) {
+            manager.login(username, password);
+            // Once logged in with new credentials,
+            // the old credentials in cookies should be cleared.
+            if (manager.getConnection() != null && remember) {
+                setCookie(response, USERNAME, username);
+                setCookie(response, PASSWORD, password);
```
wetneb, line 139: Instead of storing the password as cookie, could we store Wikidata's cookie as a cookie ourselves? That should let us restore the connection later on, without storing the user's password at all.

extensions/wikidata/module/langs/translation-en.json
```diff
@@ -65,11 +65,24 @@
     "wikidata-preview/new-id": "new item",
     "wikidata-account/dialog-header": "Wikidata account",
     "wikidata-account/explain-log-in": "Logging in to <a href=\"https://www.wikidata.org/\" target=\"_blank\">Wikidata</a> lets you to upload edits directly from OpenRefine.",
+    "wikidata-account/explain-owner-only-consumer-wiki": "See this <a href=\"https://github.com/OpenRefine/OpenRefine/wiki/Wikidata-owner-only-consumer\" target=\"_blank\">wiki</a> to get your owner-only consumer if you don't have one.",
+    "wikidata-account/explain-password-login": "You can also <a href=\"javascript:void(0)\">login with your username/password.</a>",
+    "wikidata-account/explain-owner-only-consumer-login": "You can also <a href=\"javascript:void(0)\">login with your owner-only consumer.</a>",
```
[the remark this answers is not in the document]
afkbrb: I'll update the code tomorrow (my local time).

wetneb:
@tfmorris one easy way to build and install WDTK for this PR is:
```
mvn versions:set -DnewVersion=0.12.0
mvn install
```
After that you should be able to build this PR.

To get this merged, we could:
* ask to get a new version of WDTK released (in which case it would be good to make sure we do not foresee any other change on that side at the moment)
* install a local jar in the repo
* publish a snapshot on Maven with our `org.openrefine` domain.
Have you got preferences?

afkbrb:
@wetneb 

> The discrepancy between the appearance of the two login fields is something that was there before, just throwing that in if you happen to know how to solve this cleanly (but not a requirement to merge the PR obviously).

Interesting, I can't reproduce it on the latest Firfox (I mean the different size of the two inputs). Part of the discrepancy is due to the fact that we did define the style of "input[type=text]" in [common.less](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/styles/common.less#L62) but the definition of "input[type=password]" is missing there. So probably setting the styles of "input[type=text]" and "input[type=password]" to the same could solve the problem.

The following screenshots are also made on Firfox (I've swapped the two explaining messages):

Before defining the style of "input[type=password]":

old


After:

new


Besides, I don't know why the image right to "Remember me" was missing on your Firefox. I failded to reproduce it on my Chrome and Firefox. Can you see that image on another browser?

extensions/wikidata/src/org/openrefine/wikidata/commands/LoginCommand.java
afkbrb: Good idea! This is helpful for logging in with password, but for the owner-only consumer, we still need to store the credentials in cookies.

tfmorris:
@afkbrb @wetneb I was finally able to build by editing `wdtk-wikibaseapi/pom.xml` to hardcode the values for '${okhttpSignpostVersion}' and '${okVersion}'. I'm not sure if it was a platform dependent difference in evaluation order or something else, but I'm not going to attempt to debug it. I'm running the latest mvn 3.6.3 on Mac OS X.

@wetneb Either a locally hosted jar or self-published maven artifact works for me. I guess I have a slight preference for the latter because it's closest to our target, but it's not a strong preference.

wetneb:
@tfmorris now the PR should compile out of the box for you, hopefully (I am using the Maven Central snapshot repository for the first time so I hope I did not get things wrong).

afkbrb:
"information.png" should definitely be "Information.png", but it's weird that the image was loaded normally when I was testing. I'll try figuring out what happened.

afkbrb:
Thanks for your approvals 🎉

tfmorris:
@wetneb Thanks for publishing the artefact! That'll definitely make things easier, although I did get this built yesterday after I figured out that I needed to edit *all* the POM versions.

@afkbrb Looks great! I made a few clarifying edits to the wiki page with the consumer registration instructions. Can we restrict the target to just `www.wikidata.org` instead of `*`? It seems like that's all we should need. In general, the more restrictions the users can put in place, the more comfortable they will be.

The icon looks good now. The only nit that I have about it is that I expected it to be clickable. I clicked it several times before giving up, only to realize later that I was resetting the hover timeout every time I clicked. Perhaps we should think about the best way to inform the user. We should also make an explicit statement that they shouldn't use "Remember" on a shared computer they don't control.

extensions/wikidata/module/langs/translation-en_GB.json
```diff
@@ -95,17 +95,31 @@
...
+    "wikidata-account/access-token-label": "Access token:",
+    "wikidata-account/access-token-placeholder": "access token",
+    "wikidata-account/access-secret-label": "Access secret:",
+    "wikidata-account/access-secret-placeholder": "access secret",
+    "wikidata-account/remember-me": "Remember me",
+    "wikidata-account/password-remember-me-title": "Your password won't be stored.",
+    "wikidata-account/owner-only-consumer-remember-me-title": "Consumer credentials are stored encrypted in cookies.",
+    "wikidata-account/close": "Close",
+    "wikidata-account/log-in": "Log in",
+    "wikidata-account/logged-in-as": "You are logged in as:",
```
tfmorris, line 121: Include the username parameter here so that the translators can use language specific ordering for the phrase:
```suggestion
    "wikidata-account/logged-in-as": "You are logged in as: $1",
```
You'll need to change the Javascript code with the $.i18n call too.

extensions/wikidata/module/scripts/dialogs/manage-account-dialog.js
```diff
@@ -2,120 +2,210 @@ var ManageAccountDialog = {};
...
+  elmts.cancelButton.text($.i18n('wikidata-account/close'));
+};
 
-  var dismiss = function() {
-    DialogSystem.dismissUntil(self._level - 1);
+ManageAccountDialog.displayLoggedIn = function (logged_in_username, callback) {
+  var frame = $(DOM.loadHTML("wikidata", "scripts/dialogs/logged-in-dialog.html"));
+  var elmts = DOM.bind(frame);
+  ManageAccountDialog.initCommon(elmts);
+  elmts.loggedInAs.text($.i18n('wikidata-account/logged-in-as'));
```
tfmorris, line 46: This should include the username as a parameter if it's not too complex to manage with the rest of the HTML.

extensions/wikidata/module/styles/dialogs/manage-account-dialog.less
```diff
@@ -55,10 +56,39 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 }
+
+
+.wikibase-login-dialog-footer {
+  text-align: center;
+}
+
+.wikibase-login-dialog-footer span {
+  cursor: pointer;
+}
```
tfmorris, line 94: Missing newline at EOF

extensions/wikidata/src/org/openrefine/wikidata/commands/ConnectionManager.java
```diff
@@ -0,0 +1,219 @@
...
+        connection = new OAuthApiConnection(WIKIBASE_API_ENDPOINT,
+                consumerToken, consumerSecret,
+                accessToken, accessSecret);
+        setupConnection(connection);
+        try {
+            // check if the credentials are valid
+            connection.checkCredentials();
+            return true;
+        } catch (IOException | MediaWikiApiErrorException e) {
+            logger.error(e.getMessage());
```
tfmorris, line 140: Is this message displayed to the user? It looks like it just gets put into the server log.

extensions/wikidata/module/scripts/dialogs/manage-account-dialog.js
```diff
@@ -2,120 +2,210 @@ var ManageAccountDialog = {};
...
+    ManageAccountDialog.displayLoggedIn(logged_in_username, callback);
   } else {
-      elmts.logoutArea.remove();
+    ManageAccountDialog.displayPasswordLogin(callback);
   }
+};
+
+ManageAccountDialog.tryLoginWithCookies = function (callback) {
+  var logged_user_name = null;
+  $.ajaxSetup({async: false});
```
tfmorris, line 20: This is going to block all interactivity. Normally it should return quickly, but it would be better to do it asynchronously and allow the user to cancel the operation if it's taking too long.

I know there are a bunch of other places that this is done in OpenRefine, so it's natural that you'd follow the pattern. 

Since this is new code, perhaps we should consider using Promises. They provide a much cleaner pattern.

extensions/wikidata/src/org/openrefine/wikidata/commands/ConnectionManager.java
```diff
@@ -0,0 +1,219 @@
...
+
+    final static Logger logger = LoggerFactory.getLogger("connection_manager");
+
+    /**
+     * We used this key to read/write credentials from/to preferences in the past, which is insecure.
+     * Now this key is kept only to delete those credentials in the preferences.
+     */
+    public static final String PREFERENCE_STORE_KEY = "wikidata_credentials";
+
+    public static final int CONNECT_TIMEOUT = 5000;
```
tfmorris, line 67: This seems like it might be too short, but I guess we can find out when users test.

afkbrb:
> Can we restrict the target to just www.wikidata.org instead of *? It seems like that's all we should need.

Limiting to www.wikidata.org  should work, since we only deal with it. I'll test it.

> The only nit that I have about it is that I expected it to be clickable. I clicked it several times before giving up, only to realize later that I was resetting the hover timeout every time I clicked. Perhaps we should think about the best way to inform the user. 

The cursor doesn't become a pointer when it's moved to the information icon, I think this might be enough to inform the user that it's not a clickable element.

tfmorris:
> Limiting to www.wikidata.org should work, since we only deal with it. I'll test it.

Actually, I tested it with my registration and it seemed to work fine (at least as far as authentication goes). I'll update the wiki page.

Thanks for all your work! It looks really good.

afkbrb:
Uploading edits also works.

extensions/wikidata/module/scripts/dialogs/manage-account-dialog.js
afkbrb: I fixed this directly in https://github.com/OpenRefine/OpenRefine/tree/wikidata-extension-oauth.

tfmorris: Thank you!

extensions/wikidata/src/org/openrefine/wikidata/commands/ConnectionManager.java
afkbrb: Yes, only in server log. If the login fails, the frontend can know by checking the response and then asks the user to try again.

afkbrb:
Hi @tfmorris , can you merge https://github.com/OpenRefine/OpenRefine/tree/wikidata-extension-oauth to master so we can have more people test it with our nightly releases. Or do I need to create a PR for that?

tfmorris:
@afkbrb Sorry, when I did the merge I didn't notice that the original PR wasn't against `master`, but @wetneb's development branch. Can you please create a PR from your branch to OpenRefine/master so that we have a record of what was merged and I'll merge it for you? Thanks

afkbrb:
I created that branch on OpenRefine actually...

I'll create a PR right now.

Maybe next time when I work on the Wikibase support, I should directly create the PR against master.

tfmorris:
> Maybe next time when I work on the Wikibase support, I should directly create the PR against master.

Yes, please.

PR #2668, opened 2020-06-01
Fix duplicate i18n keys for FR & ES

tfmorris:
remove-facet was mentioned twice with 2 different messages


antoine2711:
@tfmorris: thanks for fixing this. 

@wetneb: this should go also in v3.4 as it touches code there.

Regards,
   Antoine

PR #2691, opened 2020-06-11
Replace Apache Ant with Commons Compress

### LINKED ISSUE #2690: Refactor - replace Apache Ant tools for tar & bzip2 with CommonsCompress
As discussed in https://github.com/OpenRefine/OpenRefine/pull/2645#issuecomment-635622749, we currently bundle the Apache Ant tools because that was, bizarrely, the historical location that the tar archive utilities and bzip2 compressor were bundled. Now that they have a proper home in Apache Commons Compress, using the same implementations, we should migrate to those.

This should save us some size in our release distributions as well as being less confusing. It is also where all new updates will occur and easily gives us access to other compressors like 7zip.

tfmorris:
NOTE: Changes this changes the Project Manager public API where some of the old types were embedded which means that any extensions that extend these interfaces will have to be updated.

Fixes #2690. Replaces #2645

### DIFF
--- a/main/pom.xml
+++ b/main/pom.xml
@@ -236,9 +236,9 @@
       <version>2.4</version>
     </dependency>
     <dependency>
-      <groupId>org.apache.ant</groupId>
-      <artifactId>ant</artifactId> <!-- for org.apache.tools.bzip2 -->
-      <version>1.8.0</version>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-compress</artifactId>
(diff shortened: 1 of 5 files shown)


tfmorris:
It turns out that Apache Commons Compress was **not** a drop in replacement for Ant's implementation of bzip2 (see [COMPRESS-13](https://issues.apache.org/jira/browse/COMPRESS-13)) and we had no test coverage, so this PR broke our bzip2 support on import.

PR #2692, opened 2020-06-11
Use Apache HTTP Commons for Fetch URL

### LINKED ISSUE #1410: Fetch URLs should follow redirections

### LINKED ISSUE #1904: Mock the Url Fetching Tests (because they fail too often as not reachable)
**Describe the bug**
When running `refine test` the UrlFetchingTests often fail.

**To Reproduce**
1. refine test

**Current Results**
[ERROR] Tests run: 558, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 60.912 s <<< FAILURE! - in TestSuite
[ERROR] testInvalidUrl(com.google.refine.tests.operations.column.ColumnAdditionByFetchingURLsOperationTests)  Time elapsed: 5.006 s  <<< FAILURE!
java.lang.AssertionError: expected [false] but found [true]
        at com.google.refine.tests.operations.column.ColumnAdditionByFetchingURLsOperationTests.testInvalidUrl(ColumnAdditionByFetchingURLsOperationTests.java:226)

**Expected behavior**
URL tests should be mocked against an http server and pass, and not depend on real sites availability.

 - OS: Windows 10 64 bit
 - JRE or JDK Version: JDK 1.8.0_181 or JDK 11.0.1

 - Version: Trunk

### LINKED ISSUE #2680: Network tests being silently skipped
One of the HTTP tests relies on the external endpoint `http://headers.jsontest.com` but if this service is overquota and returning 503s (like it is today), the related tests are silently skipped.

https://github.com/OpenRefine/OpenRefine/blob/5351e9f41c9188006f156c27eadbd6a838cb152d/main/tests/server/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperationTests.java#L285-L298

We should figure out a better way to do this, which doesn't rely on an unreliable external service. One possibility would be to self-host the servlet https://github.com/jsontest/jsontest

tfmorris:
* Fixes #1410 by virtue of HTTP Commons having built-in support for redirects. It also supports retries, gzip decompression, and a bunch of other stuff out of the box.

* Replaced external network tests with mockwebserver based tests which should be faster and 100% reliable. Refactored to remove redundancy and make quicker. Fixes #2680. Fixes #1904.

* Replaces use of deprecated methods in ImportingUtilities which had already been converted to use Apache HTTP Commons for project creation.

### DIFF
--- a/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
+++ b/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
@@ -33,6 +33,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package com.google.refine.operations.column;
 
+import static com.google.common.base.Strings.isNullOrEmpty;
+
 import java.io.IOException;
 import java.io.Serializable;
 import java.net.MalformedURLException;
@@ -155,7 +157,7 @@ public class ColumnAdditionByFetchingURLsOperation extends EngineDependentOperat
         List<Header> headers = new ArrayList<Header>();
         if (_httpHeadersJson != null) {
             for (HttpHeader header : _httpHeadersJson) {
-                if (header.name != null && header.name != "" && header.value != null && header.value != "") {
+                if (!isNullOrEmpty(header.name) && !isNullOrEmpty(header.value)) {
                     headers.add(new BasicHeader(header.name, header.value));
                 }
             }
@@ -167,6 +169,7 @@ public class ColumnAdditionByFetchingURLsOperation extends EngineDependentOperat
                 .setConnectionRequestTimeout(30 * 1000)
                 .setSocketTimeout(10 * 1000).build();
 
+        // TODO: Placeholder for future Basic Auth implementation
 //        CredentialsProvider credsProvider = new BasicCredentialsProvider();
 //        credsProvider.setCredentials(new AuthScope(host, 443),
 //                new UsernamePasswordCredentials(user, password));
@@ -175,15 +178,14 @@ public class ColumnAdditionByFetchingURLsOperation extends EngineDependentOperat
                 .setUserAgent(RefineServlet.getUserAgent())
                 .setDefaultRequestConfig(defaultRequestConfig);
 //               .setConnectionBackoffStrategy(ConnectionBackoffStrategy)
-//                .setDefaultCredentialsProvider(credsProvider);
-
+//               .setDefaultCredentialsProvider(credsProvider);
     }
 
     @JsonProperty("newColumnName")
     public String getNewColumnName() {
         return _newColumnName;
     }
-    
+
     @JsonProperty("columnInsertIndex")
     public int getColumnInsertIndex() {
         return _columnInsertIndex;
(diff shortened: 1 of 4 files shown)


main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
```diff
@@ -134,8 +151,34 @@ public ColumnAdditionByFetchingURLsOperation(
         _delay = delay;
         _cacheResponses = cacheResponses;
         _httpHeadersJson = httpHeadersJson;
+
+        List<Header> headers = new ArrayList<Header>();
+        if (_httpHeadersJson != null) {
+            for (HttpHeader header : _httpHeadersJson) {
+                if (header.name != null && header.name != "" && header.value != null && header.value != "") {
```
wetneb, line 158: Shouldn't checks like `header.name != ""` use `equals` instead?

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
```diff
@@ -134,8 +151,34 @@ public ColumnAdditionByFetchingURLsOperation(
...
+        httpHeaders = headers.toArray(httpHeaders);
+
+        defaultRequestConfig = RequestConfig.custom()
+                .setConnectTimeout(30 * 1000)
+                .setConnectionRequestTimeout(30 * 1000)
+                .setSocketTimeout(10 * 1000).build();
+
+//        CredentialsProvider credsProvider = new BasicCredentialsProvider();
+//        credsProvider.setCredentials(new AuthScope(host, 443),
+//                new UsernamePasswordCredentials(user, password));
```
wetneb, line 172: Why is this code commented out?

wetneb:
Thanks a lot for tackling this, of course!

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
tfmorris: Yes, of course! I'll fix it.

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
tfmorris: It's not part of the current functionality, so not included in the refactoring. I left it as a placeholder for future feature enhancement, but I can remove it if you think that would be cleaner.

wetneb: Ah yes, I thought we supported this already, but we only expose an `Authorization` header in the UI. I generally try to avoid committing commented out code, but if you want to leave it as a demo for future improvements, it could be worth explaining that intention with a comment, perhaps?

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
tfmorris: Done.

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
tfmorris: Comment added

tfmorris:
Just in time because our Fetch URL tests are failing due to random.org taking more than 5 seconds to return a result.

PR #2693, opened 2020-06-11
New implementation for Conflicts-With Scrutinizer

darecoder1999:
Part of #2354

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizer.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizer.java
@@ -3,18 +3,47 @@ package org.openrefine.wikidata.qa.scrutinizers;
 import org.openrefine.wikidata.qa.QAWarning;
 import org.openrefine.wikidata.updates.ItemUpdate;
 import org.wikidata.wdtk.datamodel.interfaces.PropertyIdValue;
+import org.wikidata.wdtk.datamodel.interfaces.Snak;
+import org.wikidata.wdtk.datamodel.interfaces.SnakGroup;
 import org.wikidata.wdtk.datamodel.interfaces.Statement;
 import org.wikidata.wdtk.datamodel.interfaces.Value;
 
-import java.util.Set;
-import java.util.Map;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 public class ConflictsWithScrutinizer extends EditScrutinizer {
 
     public static final String type = "having-conflicts-with-statements";
+    public static String CONFLICTS_WITH_CONSTRAINT_QID = "Q21502838";
+    public static String CONFLICTS_WITH_PROPERTY_PID = "P2306";
+    public static String ITEM_OF_PROPERTY_CONSTRAINT_PID = "P2305";
+
+    class ConflictsWithConstraint {
+        final PropertyIdValue conflictingPid;
+        final List<Value> itemList;
+
+        ConflictsWithConstraint(Statement statement) {
+            List<SnakGroup> specs = statement.getClaim().getQualifiers();
+            PropertyIdValue pid = null;
+            this.itemList = new ArrayList<>();
+            for(SnakGroup group : specs) {
+                for (Snak snak : group.getSnaks()) {
+                    if (group.getProperty().getId().equals(CONFLICTS_WITH_PROPERTY_PID)){
+                        pid = (PropertyIdValue) snak.getValue();
+                    }
+                    if (group.getProperty().getId().equals(ITEM_OF_PROPERTY_CONSTRAINT_PID)){
+                        this.itemList.add(snak.getValue());
+                    }
+                }
+            }
+            this.conflictingPid = pid;
+        }
+    }
 
     @Override
     public void scrutinize(ItemUpdate update) {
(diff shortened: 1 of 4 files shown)


darecoder1999:
@wetneb: I hope this is what you were trying to explain about the new constructor implementation?

Stuck with getStatementByType implementation in MockContraintFetcher but trying to make it work ASAP.

wetneb:
That's exactly it, you nailed it!
For MockConstraintFetcher, one way to do it would be to just return one big list containing all sorts of constraint declarations - that would have the same sort of effect as all the other methods. But as we discussed in the call, I think the proper solution for this is to remove this class and switch to proper mocking with Mockito in each test class.

extensions/wikidata/tests/src/org/openrefine/wikidata/qa/MockConstraintFetcher.java
```diff
@@ -211,10 +218,42 @@ public boolean hasDiffWithinRange(PropertyIdValue pid) {
...
+    public Stream<Statement> getConstraintsByType(PropertyIdValue pid, String qid) {
+        EntityIdValue entityIdValue = Datamodel.makeWikidataItemIdValue("Q21502838");
+        PropertyIdValue propertyIdValue = Datamodel.makeWikidataPropertyIdValue("P2302");
+        Snak snak = Datamodel.makeValueSnak(propertyIdValue,entityIdValue);
+
+        PropertyIdValue property = Datamodel.makeWikidataPropertyIdValue("P2306");
+        Value propertyValue = Datamodel.makeWikidataPropertyIdValue("P31");
+        Snak snak1 = Datamodel.makeValueSnak(property,propertyValue);
+        List<Snak> group1 = new ArrayList<>();
+        group1.add(snak1);
```
wetneb, line 231: One way to make this slightly more concise is to use `Collections.singletonList` which creates a list with a single element.

extensions/wikidata/tests/src/org/openrefine/wikidata/qa/MockConstraintFetcher.java
```diff
@@ -211,10 +218,42 @@ public boolean hasDiffWithinRange(PropertyIdValue pid) {
...
+        Snak snak2 = Datamodel.makeValueSnak(item,itemValue);
+        List<Snak> group2 = new ArrayList<>();
+        group2.add(snak2);
+
+        SnakGroup snakGroup1 = Datamodel.makeSnakGroup(group1);
+        SnakGroup snakGroup2 = Datamodel.makeSnakGroup(group2);
+
+        List<SnakGroup> list = new ArrayList<>();
+        list.add(snakGroup1);
+        list.add(snakGroup2);
```
wetneb, line 244: When you have more than one element, you can use `Arrays.asList(first, second)` to construct the list in a more concise way.

extensions/wikidata/tests/src/org/openrefine/wikidata/qa/MockConstraintFetcher.java
```diff
@@ -211,10 +218,42 @@ public boolean hasDiffWithinRange(PropertyIdValue pid) {
...
+        Claim claim = Datamodel.makeClaim(entityIdValue,snak, list);
+
+        Reference reference = Datamodel.makeReference(list);
+        List<Reference> referenceList = new ArrayList<>();
+        referenceList.add(reference);
+
+        Statement statement = Datamodel.makeStatement(claim,referenceList, StatementRank.NORMAL, "P50$77BD7FE4-C051-4776-855C-543F0CE697D0");
```
wetneb, line 251: It might be worth ensuring that the statement id uses the same property id as the claim ("P2302" instead of "P50"), because there are plans to validate statement ids in forthcoming versions of WDTK.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizer.java
```diff
@@ -3,18 +3,50 @@
...
+    public static String CONFLICTS_WITH_CONSTRAINT_QID = "Q21502838";
+    public static String CONFLICTS_WITH_PROPERTY_PID = "P2306";
+    public static String ITEM_OF_PROPERTY_CONSTRAINT_PID = "P2305";
+
+    class ConflictsWithConstraint {
+        PropertyIdValue conflictingPid;
+        List<Value> itemList;
+
+        ConflictsWithConstraint(Statement statement) {
+            this.setParams(statement);
```
wetneb, line 31: I would inline the `setParams` method (pull its contents into the constructor and delete the method), because we do not expect to have to change the constraint definition after constructing it. In general, when we can make immutable classes, it is a good idea to build them as such (so you could also mark the two fields above as final).

wetneb:
For the LGTM issues: you should change things like `@param pid:` to `@param pid` (no colon at the end).

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizer.java
```diff
@@ -3,18 +3,48 @@
...
+    public static String CONFLICTS_WITH_CONSTRAINT_QID = "Q21502838";
+    public static String CONFLICTS_WITH_PROPERTY_PID = "P2306";
+    public static String ITEM_OF_PROPERTY_CONSTRAINT_PID = "P2305";
+
+    class ConflictsWithConstraint {
+        final PropertyIdValue conflictingPid;
+        final List<Value> itemList;
+
+        ConflictsWithConstraint(Statement statement) {
+            System.out.println(statement.getStatementId());
```
wetneb, line 31: We should remove this print.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizer.java
```diff
@@ -36,25 +66,29 @@ public void scrutinize(ItemUpdate update) {
         }
 
         for(PropertyIdValue propertyId : propertyIdValueValueMap.keySet()){
-            Map<PropertyIdValue, List<Value>> conflictingPropertyMap = _fetcher.getParamConflictsWith(propertyId);
-            for (PropertyIdValue conflictingPid : conflictingPropertyMap.keySet()) {
-                if (propertyIdValueValueMap.containsKey(conflictingPid) && raiseWarning(propertyIdValueValueMap, conflictingPid, conflictingPropertyMap)) {
+            List<Statement> statementList = _fetcher.getConstraintsByType(propertyId, CONFLICTS_WITH_CONSTRAINT_QID).collect(Collectors.toList());
+            for (Statement statement : statementList) {
+                ConflictsWithConstraint constraint = new ConflictsWithConstraint(statement);
```
wetneb, line 71: What happens when the statement does not contain the necessary qualifiers to extract constraint parameters? For instance, a conflicts with constraint without any parameter? We should make sure such declarations are ignored. This will be easier to test with once we migrate to proper mocking, because you will be able to control more easily the return value of `getConstraintsByType` in each test.

darecoder1999: For now, In that case conflictingPid will be null and `propertyIdValueValueMap.containsKey(conflictingPid)` this will return false and the if condition will not execute in that case. Is it fine for now or it needs some changes?

wetneb: Let's just try to remember to add a test for it once we have proper mocking in place :)

darecoder1999: Sure!

PR #2711, opened 2020-06-12
Preserve international characters on import/export

### LINKED ISSUE #1352: Project name with national characters
Hi all,

image

tfmorris:
Fixes #1352. Preserve non-ASCII characters in project names on
project creation and filenames on export.


tfmorris:
Note that there are a couple of additional places where the \W character class (equivalent to [^A-Za-z0-9]) is used, which will drop all non-ASCII characters. They are in the schema alignment and the SQL exporter, but I don't know if they're actually needed or not.

tfmorris:
Closing in favor of #2713 which combines this and #2712, as well as dealing with one piece of the SQL exporter issues mentioned above. I've created #2716 to track the rest of the SQL exporter cleanup.

PR #2712, opened 2020-06-12
Fixed HTML & CSV export of non-BMP UTF-8

tfmorris:
Fixes #1197

PR #2713, opened 2020-06-12
Clean up i18n support for import / export

### LINKED ISSUE #1197: HTML/CSV export corrupts UTF-8 characters outside of Basic Multilingual Pane (BMP) ie code point >10000
OpenRefine 2.7 rc2

After reading UTF 8 file and executing export as UTF 8 file, garbled characters occurred.

displayed characters
image
有限会社なべ茶屋あさ𡌛

Exported garbled characters
image
有限会社なべ茶屋あさ����

other garbled export charactor sample
𣘺𣳾

### LINKED ISSUE #1352: Project name with national characters

### LINKED ISSUE #2710: Non-ascii characters in project name are lost
**Describe the bug**

Non-ascii characters in project name are lost

**To Reproduce**

I tried with both OR3.4beta on current Mac OS 10.15 AND java-embedded OR3.4beta on Windows10.

1. When starting project with file name in Japanese, all Japanese characters are lost in suggested (initial) project name, usually leaving only extention such as csv. (In the test file case, "2020 xls" is left )

2. So, I input manually a new project name in Japanese, like "選挙2020" (= election2020). The project is safely saved. The project list shows the Japanese project name perfectly. Reloading the project works fine.

3. When exporting csv/xls/html/odf, Japanse characters in exported file name are lost. BUT exporting project ( 選挙2020.openrefine.tar ) works fine, with Japanese characters intact.

**Desktop:**

 - OS:  Mac10.15/Windows10
 - Browser Version: 
 - JRE or JDK Version: 
java 11.0.1 2018-10-16 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.1+13-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.1+13-LTS, mixed mode)

**OpenRefine :**
 - Version  3.4beta

**Datasets**

P.S. I added a test file with Japanese file name/ Japanese cell value for dev team.
[選挙2020.xls.zip](https://github.com/OpenRefine/OpenRefine/files/4772265/2020.xls.zip)

Presumably this has been happening for a long time (<- sorry for my poor language)
I usually use ascii project name (is it bad habit?). But when I was teaching OR, they wanted to use original names.

tfmorris:
This pull request combines the two previous pull requests (#2711 & #2712) since they ended up being intertwined.

Fixes #1197. Fixes #1352. Those issues contain test data files. The basic pieces are:
- #1197 - use `ContentDisposition: attachment` to force download instead of `ContentType: application/x-unknown`. The funky content type is what was interfering with the non-BMP UTF-8 characters. Took a while to figure that out!
- As part of the above, switch to consistently using option `preview: true` to control preview vs download. It was used in some place, but others used the content type.
- While investigating 1197 I tried upgrading our HTML output to XHTML 1.1, which didn't help, but seemed useful, so I left it in.
- #1352 - Multinational characters are now supported on both import and export. On import just the characters "\._-" are converted to spaces and everything else, including, notably, international characters, are left untouched. On output, a multi-OS filename character blacklist is used, including "\*/:;,?"<>|#". Some are not technically illegal, but just a pain to deal with. The cleaning is centralized in `ExporterManager.stripNonFileChars()` which we now use wherever this is needed.
- The SQL exporter still needs more love. It will output multinational characters in the SQL filename, but still aggressively cleans the table, but doesn't appear to clean the column names. One improvement that I made is that the user can now see the cleaned tablename during their preview.

### DIFF
--- a/extensions/wikidata/module/scripts/dialogs/schema-alignment-dialog.js
+++ b/extensions/wikidata/module/scripts/dialogs/schema-alignment-dialog.js
@@ -32,11 +32,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
 var SchemaAlignment = {};
-
-SchemaAlignment._cleanName = function(s) {
-  return s.replace(/\W/g, " ").replace(/\s+/g, " ").toLowerCase();
-};
-
 var SchemaAlignmentDialog = {};
 
 /**
--- a/main/src/com/google/refine/commands/project/ExportRowsCommand.java
+++ b/main/src/com/google/refine/commands/project/ExportRowsCommand.java
@@ -100,6 +100,9 @@ public class ExportRowsCommand extends Command {
                 contentType = exporter.getContentType();
             }
             response.setHeader("Content-Type", contentType);
+            String path = request.getPathInfo();
+            String filename = path.substring(path.lastIndexOf('/') + 1);
+            response.setHeader("Content-Disposition", "attachment; filename=" + filename);
             
             if (exporter instanceof WriterExporter) {
                 String encoding = params.getProperty("encoding");
--- a/main/src/com/google/refine/commands/project/ExportRowsCommand.java
+++ b/main/src/com/google/refine/commands/project/ExportRowsCommand.java
@@ -100,9 +100,13 @@ public class ExportRowsCommand extends Command {
                 contentType = exporter.getContentType();
             }
             response.setHeader("Content-Type", contentType);
-            String path = request.getPathInfo();
-            String filename = path.substring(path.lastIndexOf('/') + 1);
-            response.setHeader("Content-Disposition", "attachment; filename=" + filename);
+
+            String preview = params.getProperty("preview");
+            if (!"true".equals(preview)) {
+                String path = request.getPathInfo();
+                String filename = path.substring(path.lastIndexOf('/') + 1);
+                response.setHeader("Content-Disposition", "attachment; filename=" + filename);
+            }
             
             if (exporter instanceof WriterExporter) {
                 String encoding = params.getProperty("encoding");
--- a/main/src/com/google/refine/exporters/HtmlTableExporter.java
+++ b/main/src/com/google/refine/exporters/HtmlTableExporter.java
@@ -60,14 +60,16 @@ public class HtmlTableExporter implements WriterExporter {
             @Override
             public void startFile(JsonNode options) {
                 try {
-                    writer.write("<html>\n");
+                    writer.write("<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.1//EN\" " +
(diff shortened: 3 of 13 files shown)


thadguidry:
@tfmorris OK, waiting on this to finalize then...before testing again.

tfmorris:
@thadguidry the non-BMP UTF-8 export wasn't really touched with the additional work & merge, so shouldn't need to be retested, but I'll leave that up to you.

Loading up a project name with not only multinational characters, but also as many potentially troublesome punctuation characters would be good for testing the various exporters. The one I used for my testing was **選挙2020/-)(*:;!@#$^&*():"<>;',.[]{}**

thadguidry:
@wetneb I would be more worried about not delivering these important (longstanding) bug fixes into 3.4 for our users.  It gives the wrong impression to the community, that we care to deliver good quality, not less.  From my limited testing, it helps regardless of a an additional very small risk of some corner-case bug arising.

tfmorris:
@wetneb Thanks for the review. Sorry it's so big. If you reviewed all the changes together, reviewing by commit might provide more insight into what's what.

Having said that, since I went back and reordered & merged commits before finalizing this joint PR, it might not be that difficult now to split it back apart and get at least the fixes for #1352 into 3.4. Also the XHTML 1.1 support could clearly be pulled out. It's all in a single commit.

As far as 

> we will need end user testing to avoid any regression.

goes, how will we do that if people don't test the betas?

tfmorris:
OK, after a quick review, it looks like there's only one commit, for the SQL exporter, that straddles the two issues and needs to be split. I think I can relatively easily split this into four pieces: one for each of the bug fixes, XHTML 1.1, and then all the miscellaneous cleanups. They're all in separate commits (except for SQL), so should be straightforward.

I probably won't get to it for 6-8 hours though.

tfmorris:
I've put up 4 new branches/PRs for review.

thadguidry:
@tfmorris slow down.  We might end up owing you a million dollars at this rate! :-)  Thanks so much!

wetneb:
Thanks a lot for splitting up the PR! It really helps.

> how will we do that if people don't test the betas?

My aim is just to make sure that any patch that could potentially have unintended non-trivial side effects stays in master for long enough before being released. Obviously safe fixes can make it into the release branch directly, but it is of course a difficult call to make for each PR. I think only merging patches which fix regressions is common practice to ensure the stability of releases.

If you have suggestions about how we could better encourage people to use the beta more, I am keen to hear them :) (beyond announcing the releases properly on the mailing list…)

We already have quite a few Wikidata edits made with 3.4-beta (166,099 as of now):
https://www.wikidata.org/w/index.php?hidebots=1&hidecategorization=1&tagfilter=openrefine-3.4&limit=50&days=7&enhanced=1&title=Special:RecentChanges&urlversion=2
But it is true that only 1 746 downloads is not that much.

PR #2422, opened 2020-03-17
Pave way for table revamping, binding the table to single container

kushthedude:
Refers: #1312

PR #2714, opened 2020-06-13
Refactoring the data table into a single container

lisa761:
Continuation for #2422 

- Fixed the orange bars
- Added a sticky header
- Fixed border of header elements in Firefox


lisa761:
Switched to #2719 (rebase instead of merge for the conflict).

PR #2719, opened 2020-06-14
Refactoring the data table into a single container

lisa761:
Continuation for #2422 

- Fixed the orange bars
- Added a sticky header
> Only for `column-header` row (one with column names). The entire header group for XML/JSON (with the node structure) is not sticky yet.
- Fixed border of header elements in Firefox

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
@@ -619,7 +619,6 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
             self._cell = o.cell;
             self._dataTableView._updateCell(self._rowIndex, self._cellIndex, self._cell);
             self._render();
-            self._dataTableView._adjustDataTables();
           }
         }
       );
--- a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
+++ b/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
@@ -91,6 +91,7 @@ DataTableView.prototype.render = function() {
   var scrollLeft = (oldTableDiv.length > 0) ? oldTableDiv[0].scrollLeft : 0;
 
   var html = $(
+  	
     '<div class="viewpanel-header">' +
       '<div class="viewpanel-rowrecord" bind="rowRecordControls">'+$.i18n('core-views/show-as')+': ' +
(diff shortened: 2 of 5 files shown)


wetneb:
Could you give a few screenshots to explain the issue @thadguidry? I do not seem to have the same problem (it works great with both Firefox and Chrome for me).

thadguidry:
Not sticking header. 
Probably because I still see `position: relative` for .data-table-header on L60 in the `data-table-view.less` file

FF_No_Sticky

thadguidry:
@wetneb  Or instead of a bottom border, remove it ... and perhaps give it a very slight gradient shadow along the bottom edge of the header, to make it even more clear on the delineation between the two. 
`box-shadow: 0px 4px 6px 0px #b2b9cc;`

image

wetneb:
@thadguidry I prefer the existing solid border, it matches the rest of the design better I think.

wetneb:
Thanks, this is much better!

I have just noticed another difference, in the importer preview:
Before:
image
After:
image

thadguidry:
@wetneb 
It's a limited consensus from a non-user group.  I personally will enforce the stickyness with some small JS changes on my end whenever I have the need.

I'm not happy with the Column Groups not sticking.  It is a feature that users use for detecting messy data within a **node** (keys and values - sometimes inside branches).  For those that work with Tree-based data its very useful to see, especially when scrolling down where I usually see the messy-ness.

But all of the Column Group visualization (and Tree-based data viz) needs an overhaul anyways as you and I discussed before, and David never did finish making the **node key** field (dropdown arrow) have functionality to help support transforms and pivoting and most importantly, flattening.  We can do all that later or next year.

wetneb:
I personally do not care much about which option we choose, but notice in the Gitter discussion how you talked about very different things in that thread: @lisa761 initially asked for a choice between two options, and you answered with 7 consecutive unrelated messages.

If you have a strong opinion about which option we should go for, then it should really help if you voice that directly instead (just like @darecoder1999 did in the first reply to the thread). By keeping your messages on topic, you would ensure that your opinion is taken into account right from the start :)

wetneb:
Because this is a prerequisite for @lisa761's work on infinite scrolling, let us merge this now - we can tweak things like the stickiness of the column groups later on if need be.

PR #2720, opened 2020-06-14
Preserve international characters in project/file names on import/export

### LINKED ISSUE #1352: Project name with national characters

tfmorris:
Fixes #1352. First piece of replacement for #2713.


tfmorris:
@wetneb Does this look OK? I think it's low enough risk to include in 3.4.

wetneb:
Added to 3.4.

tfmorris:
I've updated the description with the relevant text from #2713.

PR #2721, opened 2020-06-14
Update HTML exporter to XHTML 1.1

tfmorris:
This turned out not to have anything to do with the bug I was trying to fix, but it's a useful enhancement, so might as well keep it.


tfmorris:
Actually, what we have is pretty much HTML 5 since they've dropped a lot of the DTD and namespace declaration non-sense, so why don't we just skip this.

Issue #2462, opened 2020-03-23
Improvements to our nightly releases


wetneb:
I have fixed the GitHub Actions workflow to publish snapshots of our master branch to [this repo](https://github.com/OpenRefine/OpenRefine-nightly-releases/releases) (the releases are currently empty but the next commit on master will trigger the upload of a release there).

Things to improve:

- [x] Add JRE in the Mac build, it is currently missing
- [x] Delete old releases after some time? Otherwise this is likely to become quite crowded.
- [x] We could potentially try to switch the build environment to Mac so that we can compress the Mac `.dmg` with `hdiutil`. Some commands (such as `apt-get install`) would need to be adapted.
- [ ] Use authentication header rather than get parameter for the API key to create releases

The release process is documented here:
https://github.com/OpenRefine/OpenRefine/wiki/Releasing-Version

thadguidry:
Instead of time based, I would use the Maven convention of keeping X number of SNAPSHOT's and perhaps only the last 3.

thadguidry:
Hmm, seems it did run and work...sorta kinda...but not fully.

When I expand the hierarchy for this here:
https://github.com/OpenRefine/OpenRefine/runs/781393861?check_suite_focus=true
it shows this:
```
Delete older releases
9s
    GITHUB_TOKEN: ***
Run dev-drprasad/delete-older-releases@v0.1.0
  with:
    repo: OpenRefine/OpenRefine-nightly-releases
    keep_latest: 10
  env:
    JAVA_HOME: /opt/hostedtoolcache/jdk/8.0.252/x64
    JAVA_HOME_8.0.252_x64: /opt/hostedtoolcache/jdk/8.0.252/x64
    OR_VERSION: 3.4-beta-139-gae58bc0
    API_RELEASE: https://uploads.github.com/repos/OpenRefine/OpenRefine-nightly-releases/releases/27651323/assets{?name,label}
    GITHUB_TOKEN: ***
💬  found total of 30 active releases
🍻  found 20 older release(s)
👍🏼  20 older release(s) deleted successfully!
```

Ah, I bet he needs to change his code!  Digging in...
I read here:  https://help.github.com/en/enterprise/2.16/user/github/administering-a-repository/editing-and-deleting-releases#deleting-a-release
that 
> Tip: You must delete all binary files attached to a release before you can delete the release.


Ah, files are considered "assets"?  hmm.
https://developer.github.com/v3/repos/releases/#list-release-assets

[...]

PR #2742, opened 2020-06-16
Delete older snapshot releases

wetneb:
For #2462.
Since the workflow is not active on pull requests, CI will not actually test the PR before it is merged.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -82,3 +82,12 @@ jobs:
         asset_path: ./packaging/target/openrefine-mac-${{ env.OR_VERSION }}.dmg
         asset_name: openrefine-mac-${{ env.OR_VERSION }}.dmg
         asset_content_type: application/x-apple-diskimage
+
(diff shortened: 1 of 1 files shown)


wetneb:
It looks like the GitHub action will only delete 20 releases at a time, so it will take a few commits before we only have 10 snapshot releases left.

tfmorris:
This needs to be restricted to only run on the OpenRefine organization, not forks. It's currently failing on my fork. It should be conditionalized the same way the various "Upload release" steps are.

>  Create Release0s
>  Upload Release Asset Windows0s
>  Upload Release Asset Linux0s
>  Upload Release Asset Mac0s
>   Delete older releases1s
> 🔴 no GITHUB_TOKEN found. pass `GITHUB_TOKEN` as env
> Run dev-drprasad/delete-older-releases@v0.1.0
> 🔴 no GITHUB_TOKEN found. pass `GITHUB_TOKEN` as env
>

Issue #1134, opened 2016-05-17
First pagination action after applying text filter fails


ostephens:
After applying a text filter to a project, if you then try to go to the next/last page, the project goes to the next/last page then refreshes back to the first page of the project.

If you then try the same action it is successful.
If you update the text filter the same thing happens again.

Tested on Mac using Linux build of 2.6 RC2 on Safari 9.1.1 and Chrome 50.0.2661.102

Issue #1027, opened 2015-07-10
Implement Clusterize.js or similar library


Opennat:
With (or instead of) current pagination

http://nexts.github.io/Clusterize.js/

jackyq2015:
From which aspects do you think the current implementation of  Refine need
to be replaced? Or what's the benefit we can gain from the Clusterize.js
<http://nexts.github.io/Clusterize.js/> you mentioned?

Opennat:
We work, I think as many of OpenRefine users, at large datasets, say, current project is 18.000.000+ rows with 17 columns. Even when you limit data with some filters/facets, there still many rows in result. Before applying any transforms over the results' cells, there is a good practice to look at these results: to add/delete/move/replace some data in some columns or some data depends on data in another column(s) and you need to apply another filter, and so on.

Now, we have ability to jump over pages in strictly predefined way: "X rows(records) before/after".

Yes, in our Refine we have already "tuned" data-table-view.js for
> this._pageSize = 200;

and

> var sizes = [ 100, 200, 500, 1000, 5000 ];

to get 200 rows in default result and to have ability to get more via filter menu.

But many rows in result table overloading and overheating browsers' DOM (browser lags getting worse as you increase count of results). So I think, that Clusterize or similar library will be very useful to avoid hard impact at DOM.

Plus, in current implementation, every time you click "previous/next" links (you must use mouse/touchpad for this), focus is getting lost from results table, and you need to use click to the table to be able of using PgUp/PgDown again or should use mouse to scrollbar (or use touchpad for this). Each of these movements is loss for faster and effective work.

And I did not said, that current pagination should be replaced, mind it as UI enhancement. Pressing "X" times on PgDown without clicking over pages is far faster and simple, that the way we have now.

So I think, with library, that I mentioned above, we could get almost free implementation of "infinitive scrolling page" of results (in limits of browser, sure), and all of the results will be sit in browser window always visible (in limits of Refine UI), so you dont need to scroll it in way, as it happens now.

P.S. Maybe it's time to change default limit of showable rows to 30-50, as minimum, instead of current 10? I dont think, that users who use Refine, using such small limit at all, may be in testing purposes only. Anyway, default limit of 50 rows would not affect anyone on 10 years or more modern computer.
P.P.S. Pardon about language, I hope you would get the idea.

magdmartin:
Couple of comments:
[...]

Issue #572, opened 2012-10-15
Undo/Redo page reset problem


tfmorris:
_Original author: ericjarv...@gmail.com (May 05, 2012 19:02:28)_

This is related to some of the other recent posts I've made.  Again, this is related to how the page reloads when edits have been committed.  In this particular case, when using the 'Undo' tool to undo some recent changes, upon selecting the respective undo, the page is refreshed back to #  1.  This would not be a big deal if the user had only a few pages to deal with, but when the user is editing page 14,389 of 25,000... its a bummer, because the 'only' way to get back to page 14,389 is to manually click 'next... next... next..', which of course is counter productive.

So, I would suggest that Refine returns the user to the last view, or the exact view of that given undo, after performing any 'undo', this way the user is returned ot the exact spot s/he was at when s/he made that particular edit in the first place, but now is there because s/he undid it, or redid it if the case may be.  Always returning the user to page 1/record 1 isa real pain.

Thanks,

Eric Jarvies

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=572_

wetneb:
Similar to #33 but for undo/redo

wetneb:
My change in #5411 has not fixed that yet, because an undo/redo action can not always preserve pagination: it depends on the nature of the operations involved. If all of the operations between the previous point and the new one preserved the row/record structure, then we should indeed preserve the pagination, otherwise we need to go back to the first page.

To solve this we need more operation metadata, which is something I will be introducing in the project to improve OpenRefine's reproducibility.

Issue #570, opened 2012-10-15
Editing records and advancing forward page by page- how the list refreshes and displays the records


tfmorris:
_Original author: ericjarv...@gmail.com (May 05, 2012 17:33:44)_

An annoying behavior when editing lots of records, page by ascending page;
1. Load the attached .tsv file into refine.
2. For this demonstration, make sure your page view is 10 records per page(default view).
3. facet &gt; Facet &gt; Text facet.
4. On the Facet, click 'include' for the 'facetCriteria(30)' item.
5. Click 'next' to go to second page.
6. At this point, record #  11 should show 'start of new page' record, and record #  20 should show 'end of page' record, as is seen in the 'notes' column.
7. in the column names 'facet', remove the 'facetCriteria' text from records 12, 14, 16 and 18.
8. Click the 'next' link to go to page 3 which should display records 21 to 30.
9. You will now notice on this page, there is no record that says 'start of new page', and the reason for this is because it is now on the previous page, a result of those 4 records(12, 14, 16, and 18) being edited, and thus no longer being bound to the respective Facet.  And herein is the problem.

When editing hundreds of pages of data, this is a real pain, because instead of the user being able to simply click 'next' and have all un-revised/un-edited records be displayed, the user instead must first click the 'back' link to commit the changes, and then click the 'next' link so that those 4 records that took the place of the above modified 4 records, can be reviewed and edited if need-be prior to moving on to the new page.  If any of those 4 records are edited, then the problem occurs again, and is further compounded, which is to say, the user has to click the 'back' link again for a measly &lt;4 records.

The correct behavior should allow the user to click the 'next' link without having to worry about un-revised records ending up on the previous page.  This should be accomplished by taking the last edited record on the page(e.g.- #  18), and using the next record up(e.g.- #  19) as the record that will display FIRST on the refreshed 'next' page, accounting for any/all records that dropped out of the Facet as a result of the edits.

[...]

PR #2746, opened 2020-06-17
Infinite scrolling

lisa761:
This is for the [Infinite Scrolling Project](https://github.com/OpenRefine/OpenRefine/wiki/GSoC-2020%C2%A0Ideas#replace-row-pagination-by-infinite-scrolling) as described by #1027.
_____________________
A rough roadmap for the project:
- [x] Downward scrolling that loads next set of rows
- [x] Scrollbar length to reflect the size of the data-set
- [x] Same for `Record` mode
- [x] Speed scrolling, i.e. dragging the scrollbar to a particular position quickly without loading the intermediate rows
- [x] Virtual scrolling i.e. elements (rows) that are not visible, will not be present in the DOM
- [ ] Smooth scrolling
- [x]  Performance
- [ ] #33, #570, #572, #1134 
> Currently, the scrollbar does not reflect changes made through `previous`, `next` and `goto page`.
_____________________
Datasets:
- https://query.data.world/s/7ccggdq62yhyndv2cgkws5m4kzs7ja
- https://correlatesofwar.org/data-sets/state-system-membership/system2016/
- https://correlatesofwar.org/data-sets/world-religion-data/wrp-national-data-1/@@download/file/WRP_national.csv
_____________________
[Here](https://github.com/OpenRefine/OpenRefine/compare/master...lisa761:scroll) is a rougher version of the changes with more detailed commit history.
> It includes a lot of commented out code to keep track of some alternative methods.

Link to the weekly blogs: https://lisa761.github.io/categories/gsoc/
______________________
P.S. I have let some console outputs (browser) remain for better understanding of flow of control.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
+++ b/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
@@ -108,9 +108,9 @@ DataTableView.prototype.render = function() {
       '<div class="viewpanel-rowrecord" bind="rowRecordControls">'+$.i18n('core-views/show-as')+': ' +
         '<span bind="modeSelectors"></span>' + 
       '</div>' +
-      '<div class="viewpanel-pagesize" bind="pageSizeControls"></div>' +
+      // '<div class="viewpanel-pagesize" bind="pageSizeControls"></div>' +
       '<div class="viewpanel-sorting" bind="sortingControls"></div>' +
-      '<div class="viewpanel-paging" bind="pagingControls"></div>' +
+      // '<div class="viewpanel-paging" bind="pagingControls"></div>' +
     '</div>' +
     '<div bind="dataTableContainer" class="data-table-container">' +
       '<table class="data-table">'+
--- a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
+++ b/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
@@ -460,14 +460,8 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
 
     clearTimeout($.data(this, 'scrollTimer'));
     $.data(this, 'scrollTimer', setTimeout(function() {
-      if(positionLastElement.top <= 0 && positionLastElement.bottom >= 0) {
-        var goto = self.getPageNumberSrcolling(self._scrollTop);
-        self._onChangeGotoScrolling(self._scrollTop, goto, table);
-      }
-
-      if(positionFirstElement.top < self._headerTop + 1 && positionFirstElement.bottom >= window.innerHeight) {
-        var goto = self.getPageNumberSrcolling(self._scrollTop);
-        self._onChangeGotoScrolling(self._scrollTop, goto, table);
+      if((positionLastElement.top <= 0 && positionLastElement.bottom >= 0) || (positionFirstElement.top < self._headerTop + 1 && positionFirstElement.bottom >= window.innerHeight)) {
+        self.getPageNumberSrcolling(self._scrollTop, table);
       }
       prevOperationSet = false;
     }, 250));
@@ -475,9 +469,19 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
   });
 };
 
-DataTableView.prototype.getPageNumberSrcolling = function(scrollPosition) {
-  var num = Math.floor(scrollPosition / this._sizeSinglePage);
-  return num;
+DataTableView.prototype.getPageNumberSrcolling = function(scrollPosition, table) {
+  // Loading sign
+  var width = 150 + 0.5 * window.innerWidth;
+  window.img = document.createElement('img');
+  img.setAttribute('src', 'images/large-spinner.gif');
+  img.style.zIndex = '10';
+  img.style.top = '55%';
+  img.style.left = width + 'px';
+  img.style.position = 'fixed';
+  table.appendChild(img);
+
+  var goto = Math.floor(scrollPosition / this._sizeSinglePage);
+  this._onChangeGotoScrolling(goto, table);
 };
 
 DataTableView.prototype._adjustNextSetClasses = function(start, top) {
@@ -507,7 +511,7 @@ DataTableView.prototype._adjustNextSetClasses = function(start, top) {
   this._addHeights(heightToAddTop, heightToAddBottom);
 };
 
-DataTableView.prototype._addHeights = function(heightToAddTop, heightToAddBottom) {
+DataTableView.prototype._addHeights = function(heightToAddTop, heightToAddBottom, table) {
   $('.data-table tbody tr:first').css('height', heightToAddTop);
 
   document.querySelector('.data-table').insertRow();
@@ -522,16 +526,17 @@ DataTableView.prototype._addHeights = function(heightToAddTop, heightToAddBottom
       $('.data-table tbody tr').eq(this._pageSize / 2).addClass('load-next-set');
     }
   }
+   if(table !== undefined) table.removeChild(img);
 };
 
-DataTableView.prototype._adjustNextSetClassesSpeed = function(modifiedScrollPosition, start) {
+DataTableView.prototype._adjustNextSetClassesSpeed = function(modifiedScrollPosition, start, table) {
   var heightToAddTop = modifiedScrollPosition;
   var heightToAddBottom = Math.max(0, this._sizeRowsTotal - (modifiedScrollPosition + this._sizeSinglePage));
 
   $('.data-table tbody tr').slice(1, $('.data-table tbody tr').length - theProject.rowModel.rows.length).remove();
   this._pageStart = this._totalSize - this._pageSize;
 
-  this._addHeights(heightToAddTop, heightToAddBottom);
+  this._addHeights(heightToAddTop, heightToAddBottom, table);
 };
 
 DataTableView.prototype._showRows = function(start, onDone) {
@@ -545,7 +550,7 @@ DataTableView.prototype._showRows = function(start, onDone) {
   }, this._sorting);
 };
 
-DataTableView.prototype._showRowsBottom = function(scrollPosition, table, start, onDone) {
+DataTableView.prototype._showRowsBottom = function(table, start, onDone) {
   var self = this;
 
   this._totalSize = start + this._pageSize;
@@ -562,7 +567,7 @@ DataTableView.prototype._showRowsBottom = function(scrollPosition, table, start,
   }, this._sorting);
 };
 
-DataTableView.prototype._showRowsTop = function(scrollPosition, table, start, onDone) {
+DataTableView.prototype._showRowsTop = function(table, start, onDone) {
   var self = this;
 
   this._totalSize = start + 2 * this._pageSize;
@@ -579,7 +584,7 @@ DataTableView.prototype._showRowsTop = function(scrollPosition, table, start, on
   }, this._sorting);
 };
 
-DataTableView.prototype._showRowsSpeed = function(modifiedScrollPosition, scrollPosition, table, start, onDone) {
+DataTableView.prototype._showRowsSpeed = function(modifiedScrollPosition, table, start, onDone) {
   var self = this;
 
   this._totalSize = start +  this._pageSize;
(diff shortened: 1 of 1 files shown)


wetneb:
When you feel like you are ready for some early feedback about the user experience, we can make snapshot packages and advertise them on the user mailing list.

lisa761:
@wetneb Great, thanks!

tfmorris:
@lisa761 I discovered a 4 year old abandoned commit which I thought at the time fixed #1134. I haven't looked at it again or tested it with the current code, but it's in #2846 if you want to have a look.

lisa761:
Updates on performance of rendering: https://github.com/lisa761/OpenRefine/pull/11

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -464,15 +468,147 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
...
+  this._sizeRowFirst = total / this._pageSize;
+  this._sizeRowsTotal = this._sizeRowFirst * theProject.rowModel.filtered;
+  this._sizeSinglePage = this._sizeRowFirst * this._pageSize;
+  document.querySelector('.data-table tbody').insertRow(0).setAttribute('class', 'first-row');
+  this._headerTop = $('thead').offset().top + $('thead').height();
+  this._pageStart = 0;
+  this._totalSize = this._pageSize;
+  this._adjustNextSetClasses();
+
+  var flag = 0;
```
tfmorris, line 503: More descriptive name please

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -36,13 +36,21 @@ function DataTableView(div) {
 
   this._gridPagesSizes = JSON.parse(Refine.getPreference("ui.gridPaginationSize", null));
   this._gridPagesSizes = this._checkPaginationSize(this._gridPagesSizes, [ 5, 10, 25, 50 ]);
```
tfmorris, line 38: These two lines are obsolete as well, correct?

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -464,15 +468,147 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
...
+        var goto = self.getPageNumberSrcolling(self._scrollTop);
+        self._onChangeGotoScrolling(self._scrollTop, goto, table);
+      }
+
+      if(positionFirstElement.top < self._headerTop + 1 && positionFirstElement.bottom >= window.innerHeight) {
+        var goto = self.getPageNumberSrcolling(self._scrollTop);
+        self._onChangeGotoScrolling(self._scrollTop, goto, table);
+      }
+      flag = 0;
+    }, 250));
```
tfmorris, line 550: This feels too long to me. If I'm scrolling with the scroll wheel, the screen basically goes blank until I stop scrolling. If I'm using the scroll wheel, I want to be seeing at least a few updates a second. I'm not sure if reducing this to 100 msec would help or if a different heuristic is needed (e.g. throttling the update rate rather than shutting off updates entirely if the scroll rate is too high).

tfmorris:
Changing the timeout from 250 ms to 100ms doesn't improve things. Reducing the page size from 100 to 40 didn't provide a significant improvement either. Even at that size, on a pretty fast laptop, it takes over 500 msec to populate the table with the 40 rows for a page, during which time the UI is blocked. Of course this time is longer using the default 100 row page size.

Looking at the profile when scrolling quickly using the Page Down key, over the course of 8 seconds there's 4 seconds of idle time, but each of the page loads blocks the UI for over half a second for a total of 2.5 seconds. A third of the time is spent in `DataTableCellUI.prototype._render()` when it's calling (eventually) `buildFragment`, so that could probably use some optimization, but I think the main problem is that the work is too "chunky" and we should try and split it into smaller tasks that can be overlapped more to smooth things out. This probably means decoupling the row loading increment from the page size and loading/pruning smaller chunks of rows based on the current scroll direction and how close the viewport is to the end of the buffer.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
tfmorris: Actually, on further review, there's a whole bunch of pagination related code that appears to be obsolete: lines 180-244, `onClick*Page` functions, etc.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -464,15 +468,147 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
...
+  var flag = 0;
+  $(table.parentNode.parentNode).bind('scroll', function(evt) {
+    self._downwardDirection = self._scrollTop < $(this).scrollTop();
+
+    try {
+      var element = document.querySelectorAll('.load-next-set');
+      element = element[element.length - 1];
+      var position = element.getBoundingClientRect();
+      var element2 = document.querySelector('.load-next-set');
+      var position2 = element2.getBoundingClientRect();
```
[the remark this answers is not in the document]
lisa761: Yeah, sure! There are two `load-next-set`s that we are working with, one for the previous set and one for the next state. So I guess I could go for `previousSet` and `nextSet`?

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -464,15 +468,147 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
...
+      }
+    }
+
+    if(!self._downwardDirection && !flag) {
+      if(position2.top >= 0 && position2.bottom <= window.innerHeight || 
+        (positionFirstElement.bottom > 0 && positionFirstElement.bottom < window.innerHeight)) {
+        flag = true;
+        self._onTopTable(self._scrollTop, table, this, evt);
+      }
+    }
```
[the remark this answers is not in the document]
lisa761: Yes, I'll add that.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -464,15 +468,147 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
...
+
+  var flag = 0;
+  $(table.parentNode.parentNode).bind('scroll', function(evt) {
+    self._downwardDirection = self._scrollTop < $(this).scrollTop();
+
+    try {
+      var element = document.querySelectorAll('.load-next-set');
+      element = element[element.length - 1];
+      var position = element.getBoundingClientRect();
+      var element2 = document.querySelector('.load-next-set');
```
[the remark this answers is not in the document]
lisa761: Right, I will change that, thanks.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
lisa761: Yup, I will get to removing them.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
lisa761: The reason I didn't go for throttling is because the amount of time required to render rows limits the fps to a great extent, even if we allow something like 60 fps (the most common rate followed).

lisa761:
> * There's a boundary condition bug where dragging the scrollbar thumb will only render a partial page if a row divisible by 100 falls in the middle of the display. This happens both dragging up and dragging down and the partial display can be as small as a single row.

Yes this is something I have noticed as well, I have recorded it here https://github.com/lisa761/OpenRefine/issues/1 with an idea of how to solve it. In such a case the rendering function starts executing only after you move a little bit as the condition for `last-row` or `first-row` being visible becomes true. 

> * Having the screen go blank for long periods of time is disconcerting. I think perhaps adding a spinner or "Loading..." message would be good, even if it's only displayed for a second.

Yes, we can do this.

> Looking at the profile when scrolling quickly using the Page Down key, over the course of 8 seconds there's 4 seconds of idle time, but each of the page loads blocks the UI for over half a second for a total of 2.5 seconds. A third of the time is spent in DataTableCellUI.prototype._render() when it's calling (eventually) buildFragment, so that could probably use some optimization, but I think the main problem is that the work is too "chunky" and we should try and split it into smaller tasks that can be overlapped more to smooth things out. This probably means decoupling the row loading increment from the page size and loading/pruning smaller chunks of rows based on the current scroll direction and how close the viewport is to the end of the buffer.

I will look into it, thanks!

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -464,15 +468,147 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
...
+      var position2 = element2.getBoundingClientRect();
+    } catch (err) {
+      var position = {top: undefined, bottom: undefined};
```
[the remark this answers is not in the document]
lisa761: Not really, this happens sometimes when the scroll function starts looking for these elements before they were added to the UI.

wetneb:
Thanks a million Tom for the in depth review!

tfmorris:
@lisa761 Do you want bug reports here or on your personal repo or ...?

I like the new spinner, but I've seen a few of these:

```
project-bundle.js:37334 Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at DataTableView._addHeights (http://127.0.0.1:3333/project-bundle.js:37334:33)
    at DataTableView._adjustNextSetClassesSpeed (http://127.0.0.1:3333/project-bundle.js:37344:8)
    at http://127.0.0.1:3333/project-bundle.js:37405:10
    at Object.success (http://127.0.0.1:3333/project-bundle.js:31316:9)
    at fire (http://127.0.0.1:3333/project-bundle.js:3121:30)
    at Object.fireWith [as resolveWith] (http://127.0.0.1:3333/project-bundle.js:3233:7)
    at done (http://127.0.0.1:3333/project-bundle.js:9277:14)
    at XMLHttpRequest.callback (http://127.0.0.1:3333/project-bundle.js:9687:8)
```

which I suspect are due to a race with multiple img elements getting created, but only one global variable to hold them/it.

lisa761:
Hey @tfmorris, I haven't been able to reproduce the issue but it makes sense. I have added a commit to change the scope of the image so can you check if that's still the case?

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -464,15 +383,159 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
...
+        self.getPageNumberSrcolling(self._scrollTop, table);
+        prevOperationSet = false;
+      }, 250));
+    }
+
+    self._scrollTop = $(this).scrollTop();
+  });
+};
+
+DataTableView.prototype.getPageNumberSrcolling = function(scrollPosition, table) {
```
wetneb, line 469: Typo: "Srcolling"

lisa761:
I have added changes for large screen sizes, feel free to check it and let me know of any bugs. Thanks!

wetneb:
I tried setting the page size to 20 to artificially simulate a taller screen, but I still get the previous behaviour (rows being deleted even if they are still visible, leaving white space behind):
image

Also, when I try to switch from records to rows mode, the UI starts generating tons and tons of requests to the backend (to the `get-rows` command), which eats up a lot of memory and eventually crashes my browser.

lisa761:
The records mode as such hasn't been changed yet. Because the `pageSize` is small, the rendering is more frequent and there seems to be white space until the rows load. I will check for why deletion is happening for rows that are still visible.

wetneb:
Ok! At the moment I cannot even test the rows mode on this project because of the issue I mention at the end of my last message.

wetneb:
I think we should acknowledge that this is not getting worked on further and there are little chances we can merge this as things stand, with those outstanding problems. I still think the GSoC internship was successful: we learned that it is not an easy thing to get right, and that none of the existing libraries feel suitable for it as things stand. Happy to reopen if people want to work on this further.

wetneb:
@lisa761 I still think this was really great work. I think it would be worth revisiting it and I have made a proposal about this in the forum:
https://forum.openrefine.org/t/improving-the-project-grids-layout/730

In short, I think the crucial issue to fix before this is the sizing of columns (#4806) and of rows (#1440), which would then make cell dimensions much more predictable. Do you think this makes sense?

PR #2747, opened 2020-06-17
Add LGTM badges to README

tfmorris:


wetneb:
I think the LGTM integration with GitHub would also deserve a discussion (as an issue or in the dev ML), but let's try it and see if we like it.

PR #2748, opened 2020-06-17
Fix race in Process Manager

tfmorris:
No issue. One of the Process Manager tests can fail intermittently (It failed on an edit of the README)  due to a race where the process under test finishes between the two test points so that the status changes from `running` to `done` causing the serialized JSON comparison to fail.

This PR fixes the race and also cleans up the test utility methods which were printing two copies of the diff.


wetneb:
It looks like LGTM is running the full Maven build including packaging, which causes the build failure in the PR checks - it would be worth configuring LGTM more like Travis, or perhaps tweaking our POMs to avoid that?

PR #2751, opened 2020-06-18
Fix LGTM build

tfmorris:
Attempt to fix LGTM build so that it doesn't attempt to do packaging.

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -3,4 +3,4 @@ extraction:
   java:
     index:
       build_command:
-      - mvn clean compile
+      - mvn clean test
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -1,0 +1,6 @@
+
+extraction:
+  java:
+    index:
+      build_command:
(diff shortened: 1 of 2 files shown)


tfmorris:
Merged without review because the LGTM build failure is on master, not the PR branch, so I need the config merged to be able to test it. Only affects LGTM build.

wetneb:
@tfmorris I think this should be `lgtm.yml` and not `.lgtm.yml`:
https://help.semmle.com/lgtm-enterprise/user/help/lgtm.yml-configuration-file.html

tfmorris:
> @tfmorris I think this should be `lgtm.yml` and not `.lgtm.yml`:
> https://help.semmle.com/lgtm-enterprise/user/help/lgtm.yml-configuration-file.html

No, actually both are supported, as noted on that page, but `lgtm.yml` takes precedence over `.lgtm.yml`, which I chose to keep the directory cleaner. So now that you've introduced `lgtm.yml`, the PR builds are broken again. :(

PR #2755, opened 2020-06-18
Restore character encoding guesser

### LINKED ISSUE #486: Character encoding not being set in new projects

tfmorris:
Fixes #486


wetneb:
Thanks for looking into this, it is indeed weird to have an architecture for encoding guessers without using it.

tfmorris:
Ready for review. Added some tests for at least a few relatively popular (~0.5-1% prevalence) non-UTF-8 multi-byte character sets.

Bumped the confidence threshold from 10% to 50% to weed out low confidence matches and turned on the HTML tag filter.

tfmorris:
> I suspect the reason for removing this was the size of the dependency: icu4j weigh 13MB on its own, that is pretty big!

No, the history is on the original ticket, but that was a reason I was reluctant to merge the original PR https://github.com/OpenRefine/OpenRefine/pull/1094#issuecomment-162073770 for fear that it would put us over the (at the time) 100MB limit for Github releases.

13MB is a lot though and we're only using a tiny fraction of the functionality, so perhaps investigating [juniversalchardet]( https://github.com/albfernandez/juniversalchardet) or look into Java bindings for [compact_enc_det](https://github.com/google/compact_enc_det) would be worthwhile.

tfmorris:
I've switched to [juniversalchardet](https://mvnrepository.com/artifact/com.googlecode.juniversalchardet/juniversalchardet/1.0.3) which is 0.2 MB instead of 13MB.

PR #2756, opened 2020-06-19
Wikidata extension two-legged OAuth support

afkbrb:
We should merge the "wikidata-extension-oauth" branch to master so we can test it in our nightly releases.

See #2661


tfmorris:
It needed to be rebased against the current master due to merge conflicts, so I'm going to wait until the CI tests complete after the rebase, but it should be merged shortly.

afkbrb:
Of course.

afkbrb:
I was about to complain about the secure cookie check, but I see #2757, now we don't need to deal with it :)

PR #2757, opened 2020-06-19
Ignore insecure cookie alert in LGTM

wetneb:
In #2756 I think we have a spurious alert, because OpenRefine can be used over HTTP locally and therefore cookies should not need to be sent over HTTPS by default. This attempts to disable this check in LGTM.

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,2 @@
+queries:
+  - exclude: java/insecure-cookie


wetneb:
HTTPS always for an app run on localhost?

tfmorris:
> HTTPS always for an app run on localhost?

Yes. The browsers are going to get more and more aggressive about encouraging users to always use HTTPS. Already Chrome says "Your connection is not secure" in red letters, followed by "don't enter any sensitive info" when you click on the info icon. For unsophisticated users, it's easier for them to just always look for the padlock icon than explain when no padlock is OK and when it's not. Yes, it's extra overhead, but I predict we'll get to the point where we'll just accept it.

wetneb:
But I do not even know how we would do it! Is it possible to obtain HTTPS certificates for localhost? I would be quite surprised!

wetneb:
I think you can only do self-signed certificates for localhost, and that should generate more warnings from browsers than a bare HTTP connection:
https://letsencrypt.org/docs/certificates-for-localhost/

thadguidry:
Agree.  It already causes some Library IT staff to make big frowns on their faces. :-(

The way I've handled this in the past (at Ericsson) is provide scripts that can be run to guide the user through creating the self-signed cert and then also auto-installing it for them into the OS trust store.
So we could make a generator/installer script for each OS we support, Linux, Windows, Mac.

Yeah, @wetneb kinda like that.

tfmorris:
I didn't realize that the loopback address had a special exemption, so perhaps it'll continue to be OK.

tfmorris:
I should have looked more carefully at the filename. I didn't notice that you'd created `lgtm.yml` instead of adding this to `.lgtm.yml'.

Since it seems likely that someone else will make the same mistake in the future, I guess the safest thing is to standardize on the dot-less version even though it leaves the directory a little messier.

PR #2761, opened 2020-06-19
Include coverage statistics for the main module too

wetneb:
Coverage statistics are now produced again but somehow the main module is not taken into account. And our badge points to coverage for the `coveralls` branch, which does not make sense.


wetneb:
This time master is included, yay! And coverage is back to 46%… I will not celebrate that.

Issue #2700, opened 2020-06-11
Plurals don't work for i18n package


tfmorris:
**Describe the bug**

The i18n package is missing the CLD rule parser, so plurals don't work for many languages.

**To Reproduce**
Per https://github.com/OpenRefine/OpenRefine/issues/1858#issuecomment-637274292
```
$.i18n().locale = 'en';
$.i18n("$1 de $2 {{plural:$2|page|pages}}", 5, 10);
$.i18n().locale = 'fr';
$.i18n("$1 de $2 {{plural:$2|page|pages}}", 5, 10);
```

The last line gives this: 
```
project-bundle.js:28751 Uncaught ReferenceError: pluralRuleParser is not defined
    at Object.getPluralForm (project-bundle.js:28751)
    at Object.convertPlural (project-bundle.js:28751)
    at Object.plural (project-bundle.js:28744)
    at Object.emit (project-bundle.js:28744)
    at project-bundle.js:28744
    at Function.map (project-bundle.js:504)
    at Object.emit (project-bundle.js:28744)
    at Object.parse (project-bundle.js:28739)
    at a.parse (project-bundle.js:28723)
    at Function.e.i18n (project-bundle.js:28723)
```

**Proposed solution**
Include https://github.com/santhoshtr/CLDRPluralRuleParser/blob/8baf9aedc428924fe6ee508b3d952cb5564efb3a/src/CLDRPluralRuleParser.js
(and perhaps https://github.com/santhoshtr/CLDRPluralRuleParser/blob/8baf9aedc428924fe6ee508b3d952cb5564efb3a/data/plurals.json)

(check for the appropriate version to match the version of Wikimedia i18n that we're using)

**OpenRefine:**
 - Version 3.4 beta - current master

antoine2711:
@tfmorris: it tested the french with this example, and it still doesn't work.

so I created a simple web page with jQuery 
https://raw.githubusercontent.com/antoine2711/jQuery_i18n/master/index.html, with jQuery 3.2.1, and jQuery.i18n v1.0.7. And this works fine. But, on our codebase, jQuery v1.13.1 `jQuery.fn.jquery "1.11.1"`, French is still not working.

But, if in file `jquery.i18n.language.js`, If remove the "fr" definition, then, the english fallback works fine.

So, is it possible to ask dependabot to update jQuery?

Regards,
   Antoine

thadguidry:
[...]

PR #2762, opened 2020-06-19
(I #2700) Code to fix french plurals

antoine2711:
Hi @tfmorris: like I said I would do, this is the code that will fix french plurals.

Unfortunatly, it doesn't fix any other languages, and it's more a trick than nice coding/understanding the real problem. But it fixes french plurals and is very simple.

What do you think of that?

Regards,
   Antoine

### DIFF
--- a/main/webapp/modules/core/externals/jquery.i18n.language.js
+++ b/main/webapp/modules/core/externals/jquery.i18n.language.js
@@ -80,9 +80,6 @@
 			fil: {
 				one: 'v = 0 and i = 1,2,3 or v = 0 and i % 10 != 4,6,9 or v != 0 and f % 10 != 4,6,9'
 			},
-			fr: {
-				one: 'i = 0,1'
-			},
 			ga: {
 				one: 'n = 1',
 				two: 'n = 2',
--- a/main/webapp/modules/core/langs/translation-fr.json
+++ b/main/webapp/modules/core/langs/translation-fr.json
@@ -537,7 +537,7 @@
     "core-views/view": "Aperçu",
     "core-views/extend-not-supported": "Ce service de réconciliation de supporte pas l'extension de données. Essayez de supprimer le service et de l'ajouter à nouveau. Si le problème persiste, contactez le fournisseur de service.",
     "core-views/to-text": "En texte",
-    "core-views/goto-page": "$1 de $2 page(s)",
+    "core-views/goto-page": "$1 de $2 {{plural:$2|page|pages}}",
     "core-views/first": "première",
     "core-views/word-facet": "Facette par mot",
     "core-views/check-format": "Merci de vérifier le format du fichier.",


antoine2711:
> @antoine2711 Thank you for looking into this, but my only request before was the you provide a clear reproducer (branch and/or commit) that I could use to test. I wasn't asking you for a fix.

Finding a fix is fun enough for me, so I don't mind. I don't really understand what you expect of me, and, also, I lack the time to explain thoroughly.

For me, 
```
$.i18n().locale = 'fr';
$.i18n("$1 de $2 {{plural:$2|page|pages}}", 5, 10);
```
is self-explanatory. Someone can see that it doesn't work. So, I don't really understand what you mean by: « a clear reproducer (branch and/or commit) that I could use to test ». This issue is very front-end, so test/branch code that demonstrate the problem, are maybe not the most efficient way to go, if I understand what you mean. I've read that you used to have tests case for front end, but it stoped working, and it's just another issue to put time.

> As Antonin has mentioned before, please use "Fixes #xxx" in the PR if you think it fixes an outstanding bug, so that Github will a) cross reference the two and b) will close the bug when the PR gets merged.

I use `(Closes #xxxx)`, it also works, but the related issue was close, and I don't reopen issues closed by people. You suggested that I do that in the future, so next time, I will do, and I will also map the PR that will close it again.

> Having said all that, #2774 contains the fix that's needed for #2700, so closing this.

As a user that tries to help here, I'm just glad we got this covered, and we also have improvement issues that we identified for the future, so I see positive points here.

Thanks and glad to have done this with you.

Best regards,
   Antoine

tfmorris:
> I lack the time to explain thoroughly.

That means that I have to invest extra time to try and figure out what you are talking about. That is time that I would **much** prefer to not have to spend, because it means less time for fixing bugs.

> For me,
> ```
>     $.i18n().locale = 'fr';
>     $.i18n("$1 de $2 {{plural:$2|page|pages}}", 5, 10);
> ```
> is self-explanatory. Someone can see that it doesn't work. 

But it doesn't need to be self-explanatory for you. It needs to be clear to the person who's going to do the work. That code doesn't exist anywhere in OpenRefine, so we know it's not the actual problem.

In this case the bug was location dependent. It would work in some parts of the UI (including those that I tested) and not in others (including your new code). That's why precision is important.

The type of instructions that are useful to someone trying to reproduce a problem are:

1. Check out branch `antoine2711/french-plural-bug` (with the code in the *exact* state needed to reproduce the bug)
1. Start Refine and set your language to Français
1. Go to page XYZ and click the button labelled "Truc"
1. Note that the page fails to render and on the Javascript console is the following stack trace:

If you Google "How to file a good bug report", you'll find any number of variations that include basically the same guidance.

If you ever don't understand one of my requests, please ask, instead of doing something completely different.

antoine2711:
> > I lack the time to explain thoroughly.
> 
> That means that I have to invest extra time to try and figure out what you are talking about. That is time that I would **much** prefer to not have to spend, because it means less time for fixing bugs.

What ever I do here, I always try to make it so it generates the less actions that will take time to you, Antonin, or Thad.

Since I don't code in Java, I can't be useful and I would ask more question (meaning you having to take time to answer them or just read them to ignore them) than deliver functionality. So, I stay in the JS user code, where I have knowledge that enables me to be autonomous.

> > For me,
> > ```
> >     $.i18n().locale = 'fr';
> >     $.i18n("$1 de $2 {{plural:$2|page|pages}}", 5, 10);
> > ```
> > 
> > 
> > is self-explanatory. Someone can see that it doesn't work.
> 
> But it doesn't need to be self-explanatory for you. It needs to be clear to the person who's going to do the work. That code doesn't exist anywhere in OpenRefine, so we know it's not the actual problem.

The 2 lines provided are actually exactly code used in OR, but yes, it lacks the code context. But this is what's broken. So, in 10 s. you can open the JS console, type them, and see the broken part. If some would get that to work, then the problem is most probably fixed, and I would make sure to take the ball from there and complete what was left standing by because of the problem.

> In this case the bug was location dependent. It would work in some parts of the UI (including those that I tested) and not in others (including your new code). That's why precision is important.

The feeling I got, was that it didn't work ANYWHERE in FR. But, have I known the FR plurals where working at some places, I would have directed my investigation towards that. I would have like to know that precision.

> The type of instructions that are useful to someone trying to reproduce a problem are:
> 
> 1. Check out branch `antoine2711/french-plural-bug` (with the code in the _exact_ state needed to reproduce the bug)
> 2. Start Refine and set your language to Français
> 3. Go to page XYZ and click the button labelled "Truc"
> 4. Note that the page fails to render and on the Javascript console is the following stack trace:
> 
> If you Google "How to file a good bug report", you'll find any number of variations that include basically the same guidance.

Yes, I understand that it's important to have clear and simple instruction to reproduce a problem, in order to fix it or test it. You can count on my collaboration in terms of improving that quality of Bug Report I do.

> If you ever don't understand one of my requests, please ask, instead of doing something completely different.

Yes, I try to balance my interaction here between providing added value, and being a contributor that have questions and needs counseling in order to improve it's skills.

I'm looking forward to work in an optimal manner with you, @tfmorris, to squash more bugs, add features and improve my Git/GitHub skills which are currently minimals.

Regards,
   Antoine

PR #2764, opened 2020-06-20
Add new option for deleting tags associated to releases

thadguidry:
Graciously, Reddy has added a new option for us to `delete_tags`
https://github.com/dev-drprasad/delete-older-releases#delete_tags

Also, Reddy mentioned in his PR we can just do manual cleanup (via Git cli) for the old tags that are not associated to a release.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -89,7 +89,11 @@ jobs:
       uses: dev-drprasad/delete-older-releases@v0.1.0
       with:
         repo: OpenRefine/OpenRefine-nightly-releases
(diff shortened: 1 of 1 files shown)


thadguidry:
@wetneb GitHub Actions work like this...with simple Git tags.  His `v0.1.0` tag already has all the commits pushed to his `master` (he **did not** make a new tag, which is bad practice on him actually).  https://github.com/dev-drprasad/delete-older-releases/commits/master
And you can further see that the tag `v0.1.0` has that latest commit and that it contains that new functionality if you click on the little commit hash. https://github.com/dev-drprasad/delete-older-releases/tags

In summary:  That new `delete_tags` parameter will work with the existing `v0.1.0` now.

Issue #1011, opened 2015-06-26
Custom facet names


esiegerman:
It would be great if one could give a facet a custom name.  This would help immensely when there are multiple facets against the same column, for example.

(This is related to #575 and #745, but not identical to either of them.)

magdmartin:
I keep supporting this suggestion.

thadguidry:
I think something like this functionality would work well:
1. allow a user to click on the titleSpan.text on a Facet and update its config.name.
2. allow a user to hover on the Facet titleSpan.text and see the original column name on the hover event that would show `"Column: " + config.columnName`

antoine2711:
Here's a VERY simple implementation of renaming a facet. Could be added in v3.4 as it's so simple.
PR #2767 

Regards,
   Antoine

thadguidry:
Hmm, I think I view this differently now.  After seeing lots of overlapping issues regarding "remembering things" for Facets and Columns.  For users, this would be exposed as **metadata** in various places of OpenRefine.

@antoine2711  I think what the issue was asking was pertaining to temporary **metadata**.  Primarily, tons of folks want to see not only "name", but also "title" and "descriptor" at a minimum for Facets and Columns.

I would **not** put a lot of emphasis on what I suggested earlier above in #1011 because I've learned that the real use cases are more about **metadata** in a lot of places of OpenRefine that users want to edit and store to avoid ambiguity and **remember things**, especially when sharing but also while exploring their data.

And so we should probably re-design the approach in that issue #1011 

Imagine:

1. A user can give alternative titles to Facets.  (and be able to save that info as part of exporting and sharing a project)
2. A user can give alternative titles to Column names. (the **metadata** and saving and exporting)

We can't do that currently.
From the Frictionless Data Specs - Table Schema
```
"fields": [
    // a field-descriptor
    {
      "name": "name of field (e.g. column name)",
      "title": "A nicer human readable label or title for the field",
      "type": "A string specifying the type",
      "format": "A string specifying a format",
      "description": "A description for the field"
      ...
    },
```
IF using only the frontend, you can expose the temporary solution that a "title" on a facet can be set, while holding on to the original "name" of the column, that that's what I see #1011 as asking for.

"name",  <-- we have this now
"title" or "label"  <-- currently, we cannot set this on Facets and give it visibility in place of the "name", but also being able to see the original "name" if they hover over it.

Later I would like to be able to save that info permanently and have it persist with the project.
[...]

PR #2767, opened 2020-06-20
(I #1011) Rename facets

antoine2711:
Here's a VERY simple implementation of renaming a facet.

image

image

I uses document.prompt() JS function to minimize the risk and difficulty. This count be easily updated to an `input type="text"` in a next revision, incremental upgrade.

Regards,
   Antoine


antoine2711:
This is a partial implementation of I #1011.

main/webapp/modules/core/scripts/facets/list-facet.js
```diff
@@ -161,7 +161,7 @@ ListFacet.prototype._initializeUI = function() {
           '<a href="javascript:{}" class="facet-choice-link" bind="resetButton">'+$.i18n('core-facets/reset')+'</a>' +
           '<a href="javascript:{}" class="facet-choice-link" bind="invertButton">'+$.i18n('core-facets/invert')+'</a>' +
           '<a href="javascript:{}" class="facet-choice-link" bind="changeButton">'+$.i18n('core-facets/change')+'</a>' +
-          '<span bind="titleSpan"></span>' +
+          '<span class="facet-title-span" bind="titleSpan" title="Click here to edit the name of the facet"></span>' +
```
thadguidry, line 164:
```suggestion
          '<span class="facet-title-span" bind="titleSpan" title="Click to rename facet"></span>' +
```

thadguidry: Simple English for easier translations.  Can we also allow hover text to be translatable?

main/webapp/modules/core/scripts/facets/list-facet.js
```diff
@@ -525,6 +526,18 @@ ListFacet.prototype._doEdit = function() {
   new ClusteringDialog(this._config.columnName, this._config.expression);
 };
 
+ListFacet.prototype._editTitle = function() {
+  var currentFacetTitle = this._config.name;
+  
+  var promptText = "The current name of the facet is: "+ currentFacetTitle +".\n What name do you want now?";
```
thadguidry, line 532:
```suggestion
  var promptText = "Current name of facet is: "+ currentFacetTitle +".\n Rename to: ";
```
Let's use Simple English as much as possible for easier translations.  Can we translate this also?

antoine2711:
@thadguidry: both of your request have been forfilled.

The last thing is that now, this ONLY applies to ListFacets. Now, the proper way to add it to all would be to subclass the Facets object in JS.

I have already opened an issue about that, but it is stalled. I think we could just go with this, and later extended it to other facets with the subclassing power.

Regards,
   Antoine

main/webapp/modules/core/styles/project/facets.less
```diff
@@ -185,15 +185,21 @@ li.facet-state-minimize a.facet-title-minimize:hover {
...
   color: black;
   }
+  
 .facet-choice-link.facet-mode-inverted {
   text-decoration: line-through;
   color: black;
   }
 
+div.facet-title span.facet-title-span {
+  cursor: pointer;
```
wetneb, line 200: I would rather use `cursor: text` to better suggest the editing.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -285,6 +285,8 @@
     "core-dialogs/for-null-cell-value-to-empty-str-label": "Convert null value to NULL in INSERT",
     "core-dialogs/choose-export-destination": "Please choose the destination for project export",
     "core-dialogs/export-to-local": "OpenRefine project archive to file",
+    "core-facets/facet-current-title": "The current name of the facet is: $1.\n To what name do you want to change it?",
```
wetneb, line 288: I would make the question even simpler: instead of "To what name do you want to change it?", I would just use "Enter the new name" or something like that.

main/webapp/modules/core/langs/translation-fr.json
```diff
@@ -280,6 +280,8 @@
     "core-dialogs/custom-tab-exp": "Exporteur tabulaire personnalisé",
     "core-dialogs/select-columns-dialog": "Sélectionner les colonnes",
     "core-facets/regular-dot": "Point de taille normale",
+    "core-facets/facet-current-title": "Le nom actuel de la facette est : $1.\nQue voulez-vous que le nouveau nom soit?",
```
wetneb, line 283: Similarly: "Entrez le nouveau nom :"

main/webapp/modules/core/scripts/facets/list-facet.js
```diff
@@ -161,7 +161,7 @@ ListFacet.prototype._initializeUI = function() {
           '<a href="javascript:{}" class="facet-choice-link" bind="resetButton">'+$.i18n('core-facets/reset')+'</a>' +
           '<a href="javascript:{}" class="facet-choice-link" bind="invertButton">'+$.i18n('core-facets/invert')+'</a>' +
           '<a href="javascript:{}" class="facet-choice-link" bind="changeButton">'+$.i18n('core-facets/change')+'</a>' +
-          '<span bind="titleSpan"></span>' +
+          '<span class="facet-title-span" bind="titleSpan" title="'+$.i18n('core-facets/edit-facet-title')+'"></span>' +
```
wetneb, line 164: I see this was done in the same way above (so it makes sense to imitate it for consistency), but just for the record, this is really bad: the HTML code can be broken easily by a translation which would contain certain characters. We should rather apply the title via JS DOM manipulation, which gets rid of escaping issues (just like this is done in most other places).

antoine2711: Yes, I usually try to follow the coding style present, but I understand in the future you would rather always have JS DOM manipulation, and I agree. So I did it, and 10 other also.

main/webapp/modules/core/styles/project/facets.less
antoine2711: @wetneb: here, I would respectfully disagree. The text cursor must be shown when the edit field is present/accessible. This will happen just after the dialog is opened and the cursor above the edit field.

In the next itineration of this feature, when we will be able to edit directly, then, the text cursor will be the best choice, IMOO.

Does it make sense?

Regards,
   Antoine

antoine2711:
@thadguidry: I've added the request to see the column name when hover the title.
I had to be creative, because, for Flags, Stars, and others, there is no column name. So I improvised with a « source value », like for flags, it's `<internal-flags>`. What do you think of that?

Regards,
   Antoine

image

thadguidry:
@antoine2711 I don't understand this. Sorry.  Perhaps Antonin can comment when he has time.

antoine2711:
> @antoine2711 I don't understand this. Sorry. Perhaps Antonin can comment when he has time.

In I #1011, you asked, when the cursor is over the title, to see the column name of the facet.
image

So, I just coded that more in this PR. Is it like you wanted?

Regards,
   Antoine

thadguidry:
@antoine2711 Not really.  See my comment in #1011

main/webapp/modules/core/styles/project/facets.less
wetneb: If you want to keep a more "clickable" look and feel, then I would add a little pencil icon at the end of the current name when hovering it - that would suggest editing more visually, people would be able to understand it without reading the help text.

wetneb:
I am not sure why you need to add the column in the hover message. Facets can rely on more than one column too (for instance the scatterplot facet or any custom facet with an expression that is relying on multiple columns). I think the simple message you had initially was good.

antoine2711:
> I am not sure why you need to add the column in the hover message. Facets can rely on more than one column too (for instance the scatterplot facet or any custom facet with an expression that is relying on multiple columns). I think the simple message you had initially was good.

Yes, multi-columns or custom expression would have no particular values.

@thadguidry & @wetneb, the reason I tried this was because in #1011, there is this:
image

And I thought it was what you mean, Thad. But no problem, it's easy to remove. It could also stay, it's informative most of the time.

Regards,
   Antoine

main/webapp/modules/core/styles/project/facets.less
antoine2711: > I would add a little pencil icon at the end of the current name when hovering it - that would suggest editing more visually, people would be able to understand it without reading the help text.

Do we currently have this behavior somewhere in OR that I could mimick/copy?

Regards,
   Antoine

thadguidry: @antoine2711 I think he means using a custom image with an Pencil icon and using it for the CSS value for a custom cursor here:  https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

Additionally perhaps with a keyword fallback for `text` or `pointer` where both are typically used to give indication to users that there is a clickable Link.

Something like this:
```css
.custom {
  cursor: url(images/edit_pencil.png), text;
}
```

wetneb:
For me it is trying to put too much information in an `alt` message, which I would normally expect to be a static help text (as a user I would not look for info like this in a help text by hovering an element). Not a hill I will die on though.

antoine2711:
> For me it is trying to put too much information in an `alt` message, which I would normally expect to be a static help text (as a user I would not look for info like this in a help text by hovering an element). Not a hill I will die on though.

Niether for me is it a hill that I'll die for.
Also, I also think this kind of Tooltip, HTML attribute title, is mostly static and must be optional.
But, by enabling the user to change the name of the facet, the user can loose the information of the column. Of course, this information is sometimes not even very indicative from the start. And this PR doesn't cover all the facets that can be created in OR, but still, it's useful information. I just think that we should not consider this as a solution for providing core configuration data about a facet to the user. A little screen snapshot and a line in the docs should be enough for this, and we should continue to improve the facets.

I also think, @wetneb & @thadguidry, that we could mimic the hover use now in OR to change the title of a project, the yellow background. (And we should also change the cursor of the « Edit project name » to the « click-hand » instead of the I-beam that cannot select letters like it should.)
image

What do you think of that? Consistency in the user interface and a more visible and discoverable « Rename facet » functionality?

Regards,
   Antoine

thadguidry:
Let's stick to #1011 needs for now.
1. Allow the user to rename the facet name
2. When hovering over the name of the facet, show hover text of the original name always (not the last custom name given)
3. The hover text should not say anything about "clicking" or whatever... that can be handled visually through a better icon like in 4 below.

4. (optional) Have a nice edit pointer available for hovering over items in the UI that can be renamed/changed/edited.  A Pencil icon is ideal. https://material.io/resources/icons/?search=edit&icon=create&style=baseline

Then we can merge this and close #1011 to the benefit of users more quickly! :-)
(I don't see an enhancement request for Project renaming changes...so discussion of that can go into a new issue or our mailing list. Thanks!)

antoine2711:
> 4. (optional) Have a nice edit pointer available for hovering over items in the UI that can be renamed/changed/edited.  A Pencil icon is ideal. https://material.io/resources/icons/?search=edit&icon=create&style=baseline

Do you mean like wikidata? (But maybe on the right?)
image

> Then we can merge this and close #1011 to the benefit of users more quickly! :-)
> (I don't see an enhancement request for Project renaming changes...so discussion of that can go into a new issue or our mailing list. Thanks!)

It's not that I want to be OT, but since we are talking about the same action, I think OR should be consistent within itself. Editing the name of the project should not be different than editing the name of a facet.

Regards,
   Antoine

antoine2711:
Also, I don't think this « closes » #1011, that's why I didn't write (Closes #1011) in the description, nor did I linked it to that issue. But if you believe I'm wrong, I'll go along.

Regards,
   Antoine

thadguidry:
> 
> 
> This is a partial implementation of I #1011.

This statement?  And also the title of this PR states `(I #1011)`
but yes, I understand that you didn't allude to closing or fixing #1011

thadguidry:
> 
> 
> > 1. (optional) Have a nice edit pointer available for hovering over items in the UI that can be renamed/changed/edited.  A Pencil icon is ideal. https://material.io/resources/icons/?search=edit&icon=create&style=baseline
> 
> Do you mean like wikidata? (But maybe on the right?)
> image
> 

Yes.  Location of icon up to you and how it looks and most importantly usefulness to users.

> It's not that I want to be OT, but since we are talking about the same action, I think OR should be consistent within itself. Editing the name of the project should not be different than editing the name of a facet.

Great! then you are now thinking like me and Antonin!

wetneb:
@antoine2711 concerning consistency, note that on your screenshot above, the cursor is the one I suggested - even if before the user clicks, the field is not editable yet. So I think it would be consistent to have the same in the facets - and perhaps it would be worth highlighting the facet name in a similar fashion. When the user discovers this feature, they do not know that a JS alert will be used - and that should not cause any issue (you want to give the user clues about editability before they click).

wetneb:
I am closing this PR as stale, given the lack of recent activity. For me the blocking points are:
* the use of a JS alert: we should aim for an input field just like for project name editing (in the project view)
* using the same cursor as for project name editing.
Feel free to reopen to push this forward!

antoine2711:
Hi @wetneb : this now does exactly like the project renaming.
I changed the cursor for text although IMOO it's not the best choice.

Regards,
   Antoine

Issue #2771, opened 2020-06-20
Simple preference to control the facet width


antoine2711:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
I want to be able to change the width of the facet panel (left one).
~~`ui.toolPanelWidth=300`~~ or `ui.toolPanelWidth=500`, at the user’s will.
**Changed to `ui.facetsHistoryPanelWidth`**.

**Describe the solution you'd like**
A simple preference that controls the width of the panel.

**Describe alternatives you've considered**
Go into the console and play with JS.

**Screens shapshots**
![image](https://user-images.githubusercontent.com/12217525/82875686-b6474b80-9f05-11ea-98d4-d7522bd7d559.png)

Regards,
   Antoine

thadguidry:
@antoine2711 What does it look like if you click on the `Change` link in the top right of the facet? Does the expression display also stretch horizontally with the facet?

antoine2711:
> @antoine2711 What does it look like if you click on the blue `change` link in the top right of the facet? Does the expression display also stretch horizontally with the facet?

![image](https://user-images.githubusercontent.com/12217525/85210256-714be480-b30c-11ea-8eee-04b4ccc36780.png)

After « cancel »:
![image](https://user-images.githubusercontent.com/12217525/85210264-7e68d380-b30c-11ea-9ea1-fea3aaf74fb1.png)

Firefox:
![image](https://user-images.githubusercontent.com/12217525/85210287-abb58180-b30c-11ea-9ee4-1dfc4ca28595.png)

@thadguidry: as far as I see, everything is normal. Chrome + Firefox. 

Regards,
   Antoine

antoine2711:
> @antoine2711 What does it look like if you click on the blue `change` link in the top right of the facet? Does the expression display also stretch horizontally with the facet?

Yes @thadguidry. It is always full width.

![image](https://user-images.githubusercontent.com/12217525/85210644-b45b8700-b30f-11ea-8ba2-66e21a970176.png)

Regards,
   Antoine

Issue #2768, opened 2020-06-20
Editing a preferences shows the value at loading time, not the current pref's value


antoine2711:
**Describe the bug**
When you edit a pref, the value shown in the prompt will always be the value at the page loaded time, and not the current value.

**To Reproduce**
1. Go to the Preference window
2. Create a pref, let's say « TestPref », and assign it `100`.
3. Reload the page, you can see your new pref with the good value
4. Edit the value of the « TestPref » to `TEXT`
5. Go to edit AGAIN the value of the « TestPref » and see the value `100` proposed, instead of `TEXT`, the current value of this preference.

**Screenshots**
Step 4.
![image](https://user-images.githubusercontent.com/12217525/85208674-45c2fd00-b300-11ea-8a4d-8705e25f5ba1.png)

Step 5.
![image](https://user-images.githubusercontent.com/12217525/85208697-786cf580-b300-11ea-8e49-f068e04c0035.png)

**Desktop:**
 - OS: Mac OS X.14
 - Browser Version: Chrome (current)

**OpenRefine :**
 - Version v3.3. to v3.5

tfmorris:
@antoine2711 Please don't assign people to issues. Let them volunteer themselves. If you need a reviewer for a PR, Github has mechanisms to request that. If you need discussion, you can ping people like:
@thadguidry Did you agree to be assigned to this?

thadguidry:
@tfmorris No sir.

antoine2711:
> @antoine2711 Please don't assign people to issues. Let them volunteer themselves. If you need a reviewer for a PR, Github has mechanisms to request that. If you need discussion, you can ping people like:
> @thadguidry Did you agree to be assigned to this?

Hi @tfmorris: adding you here as an assigne was a mistake. I wanted to ask you to review the code that fixes this, and, of course, that in the PR, which I did after. 

And yes, I should not have added @thadguidry, but rather myself, which I just did. I created this issue but I also already had the code to fix it. So, I'm really looking to get this done from the start to the solution.

Regards,
   Antoine

wetneb:
Sorry for the time it took to review this: I can reproduce the issue.

PR #2772, opened 2020-06-20
(I #2771) Control facet panel width thru a preference

antoine2711:
(Closes I #2771) 

First implementation, functional. Only 2 lines of code.

Very simple. Depends on code merged in v3.5.

Regards,
   Antoine


wetneb:
It clearly does not hurt, but I imagine that it should even not be that hard to make it resizeable with the mouse directly now that resizing the grid is much cheaper?

antoine2711:
> It clearly does not hurt, but I imagine that it should even not be that hard to make it resizeable with the mouse directly now that resizing the grid is much cheaper?

Well, now, the way it's coded, the right pane is « fixed » on the left. So I could do easily a resize but the right pane will not move.

I think we should go incremental on this, commit this and after, make it resizable.

I spent an hour trying, with no big success, and a lot of non-controlled changes. I think it is hard, at least for me. Maybe @lisa761 who have good insight on that, but she must be busy.

Here's the branch on my fork with my explorations: issue-2771-adjust-facet-panel-width-resizable

Regards,
   Antoine

antoine2711:
> It clearly does not hurt, but I imagine that it should even not be that hard to make it resizeable with the mouse directly now that resizing the grid is much cheaper?

Follow-up, after a few other hours of playing, I go something kinda like a prototype, but with known issues and not polished at all. Better than I could ever anticipate (it did'n know flex existed. I was very surprised.).

Still, I wouldn't commit that. (Branch `issue-2771-adjust-facet-panel-width-resizable`)

Regards,
   Antoine

tfmorris:
I agree that direct manipulation would be much better. No one is ever going to find a hidden preference value.

antoine2711:
> I agree that direct manipulation would be much better. No one is ever going to find a hidden preference value.

Hi @tfmorris, yes, preferences are kind of hidden. Maybe we could think of a « help » tab, like in the Expression dialog, to make them more known.

Preferences are perfect to implement a small functionality and not be burden by a lot of UI that take time to code/test. Of course, if a functionality becomes very popular, the next would be to implement better UI for it. So, it can be a step in the elaboration of feature evolution.

Our « Preferences » window is still new with a lot of « bugs ». Yesterday, I created an issue because no language other than EN way working for it. I also did an issue + coded the fix for a situation where you dont see the real current value when editing a preference (Editing a preferences shows the value at loading time, not the current pref's value #2768).

I think preferences enables us to write small features, which mean a better potentiel to providing solutions for our users.

Regards,
   Antoine

main/webapp/modules/core/scripts/project.js
```diff
@@ -63,7 +63,11 @@ Refine.reportException = function(e) {
 };
 
 function resize() {
-  var leftPanelWidth = 300;
+  var leftPanelWidth = JSON.parse(Refine.getPreference("ui.toolPanelWidth", 300));
```
wetneb, line 66: Perhaps the preference key could be improved: I would not call this a "tool panel". Perhaps just `ui.leftPanelWidth`?

antoine2711: Yes, this pref name could be better but what if we have another preference to put tools panel on the right instead of the left?

Regards,
   Antoine

wetneb: What I am worried about with `tools` is that we never refer to this panel as being a tools panel anywhere. `ui.facetsHistoryPanelWidth` if you prefer?

wetneb:
Closing this given the lack of consensus of the preference key and on its discoverability.

main/webapp/modules/core/scripts/project.js
antoine2711: @wetneb : I'm totally open to another name. `ui.facetsHistoryPanelWidth` is good, but `ui.facetsPanelWidth` might be even simplier!

Regards, A.

antoine2711:
`ui.facetsHistoryPanelWidth` is now used.

wetneb (commented):
If we go down that route, then the new preference should also be included in the docs:
https://docs.openrefine.org/manual/running/#preferences

antoine2711:
Looking at other preferences, I think it may be better to call it:
`ui.browsing.facetsHistoryPanelWidth`.

@wetneb?

Regards, A.

wetneb:
Fine with me!

PR #2775, opened 2020-06-21
Fall back to jquery.i18n v1.0.4

tfmorris:
I'm not sure if this is necessary or not. I updated to v1.0.7 as part of the fix for #2700. The speculation about the jQuery version was a red herring, but I'm not sure how closely coupled jquery.18n is to the base jQuery. v1.0.4, which we had before, is a reasonably close match for the vintage of jQuery that we run.

@wetneb Any opinions on the lowest risk path forward? v1.0.4 (our previous version) or v1.0.7 (the current version which lists a dependency of jquery 3.5.1) or perhaps an intermediary version?


wetneb:
Not sure about this! Perhaps it was indeed a bit of a bold move to update a dependency at this stage in the release process of 3.4, but I would be inclined to keep the new version now.

The previous JS file we had for this library was minified, the new one is not. Was this intentional?

I am thinking that minified files have two advantages:
- they are lighter
- they discourage our own contributors from editing them (we should stick to official releases as much as possible)
That being said that is clearly not a hill I will die on. The proper solution would be to have a real package manager to handle our JS dependencies and stop copying them in the repository, I would say.

tfmorris:
OK. Logically there shouldn't really be much/any coupling between the i18n library and the underlying version of jQuery.

Is there a distribution with minified files? I didn't see anything obvious. Or did you minify them by hand?

Doesn't everything get bundled together and minified as part of the packaging? That would be the other way to address this, but I haven't looked into how the packaging is done.

wetneb:
I am not sure where I got the minified version from, I do not remember doing the minification myself.

There is no minification in our packaging process at the moment.

tfmorris:
Closing for now. We can resurrect the branch we run into problems with v1.0.7, but hopefully it'll be fine.

PR #2781, opened 2020-06-21
Fix reference to undefined surfireArgs param

### LINKED ISSUE #2780: Eclipse run/debug launches fail with undefined parameter
Attempting to use any of the launch configurations (Run, Debug, Test) in Eclipse results in an immediate error dialog saying that the `surefireArgs` parameter isn't defined.

tfmorris:
Fixes #2780 

I didn't look into this too deeply, so it might be that defining an empty `surefireArgs` param is a better way to go, but either way it needs to be made consistent. I suspect that Eclipse is being pickier than Maven about missing parameters.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,7 @@
     <jee.port>3333</jee.port>
     <refine.data>/tmp/refine</refine.data>
     <surefire.version>2.22.2</surefire.version>
+    <surefireArgs></surefireArgs>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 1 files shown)


wetneb:
Unfortunately that breaks coverage computation, see the very end of the logs here:
https://travis-ci.org/github/OpenRefine/OpenRefine/jobs/700693457
(and the fact that coveralls has not reported any PR check here)

tfmorris:
Eclipse looks good too. Not sure whether we both fixed the problem in different ways, but left as is because it works.

PR #2783, opened 2020-06-22
Fix Google Sheets export with more than 26 columns

### LINKED ISSUE #2760: Google Sheets uploads are limited to 26 columns (ie A-Z)
**Describe the bug**
From a particular file, if I try to export, I get an empty gSheet, even if I can see 500 rows in OR.
There is no error on the Java/server side, but in the browser, I get an error in the console.

Note that this happens in 25 % of the 12 projects I tried to export.

**To Reproduce**
1. Open the particular project in OR
2. Click on 'Export -> Google Sheet'
3. Open the resulting gSheet, and see no data (empty sheet)

**Current Results**
Empty gSheet.

**Expected behavior**
gSheet with the data of the project.

**Screenshots**
image

**Desktop:**
 - MacOS

**OpenRefine :**
 - Version v3.4b1

**Datasets**
Here's a dataset that you can import as .CSV and then try to export to gSheets, and that will give you an empty sheet.

[DebugRep.csv.zip](https://github.com/OpenRefine/OpenRefine/files/4810544/DebugRep.csv.zip)

tfmorris:
Google Sheets default to just 26 columns (A-Z) and we need to
explicitly add more columns if we need them.

Fixes #2760

### DIFF
--- a/extensions/gdata/src/com/google/refine/extension/gdata/SpreadsheetSerializer.java
+++ b/extensions/gdata/src/com/google/refine/extension/gdata/SpreadsheetSerializer.java
@@ -11,6 +11,7 @@ import org.slf4j.LoggerFactory;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.api.services.sheets.v4.Sheets;
 import com.google.api.services.sheets.v4.model.AppendCellsRequest;
+import com.google.api.services.sheets.v4.model.AppendDimensionRequest;
 import com.google.api.services.sheets.v4.model.BatchUpdateSpreadsheetRequest;
 import com.google.api.services.sheets.v4.model.BatchUpdateSpreadsheetResponse;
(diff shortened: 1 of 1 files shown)


antoine2711:
So I tested the 6 documents that wouldn't have data transfered, and for 5/6, it was good.
Now, I only have one that sends nothing.

@tfmorris: would you like this project full archive to test? I understand that you are looking at a lot of things, if you want my help for testing, maybe some sort of coordination might help. 

Regards,
   Antoine

tfmorris:
@antoine2711 Thanks for the offer, but I think it makes sense to wait until all the Google Sheets PRs are reviewed, merged, and integrated before doing additional testing. One of the things I fixed is the error reporting, so that should hopefully make it easier to test.

tfmorris:
I'm going to merge this with #2784 so that I can restructure things enough that I can do at least a basic unit test.

PR #2784, opened 2020-06-22
Improve Google Sheets upload

tfmorris:
Fixes #2760 along with a variety of problems that I found when testing fixes for it and fixes #2306..

- upload in chunks instead of serializing the entire document at once
- Free up resources as we go
- stop if an error occurs
- reduce batch size to try and stay in 10MB request size limit
  (but need a more dynamic way to do this probably for very wide
   sheets or sheets with large values)
- preserves cell data type on upload - fixes #2785.

I'll open issues for some of the other things. The whole thing is kind of a mess.

PR #2787, opened 2020-06-22
Preserve cell data types

### LINKED ISSUE #2785: Data type not preserved in Google Sheets export
The Google Sheets exporter currently converts all cell values to strings before exporting. It should preserve data types, at least for those that are supported by both OpenRefine and Google Sheets.

tfmorris:
Fixes #2785


tfmorris:
This is just a few lines, so I'm going to merge it into #2784 where I can take advantage of the test infrastructure.

PR #2740, opened 2020-06-16
First pull request to test the contribution workflow

allanaaa:

PR #2799, opened 2020-06-23
begin Docusaurus 2 migration

thadguidry:
**Breaking changes**
- ~~Netlify - its config will need changes and I think we should bite the bullet and try to get Docusarus 2 going quickly for Allana @allanaaa .  And this is really close and nearly there, it builds and runs `yarn run start` on my PC, so just needs a bit of help.~~
  - ~~Fixed!!!  Had to update the build artifact path references and also updated to just use `yarn` to go faster in the Netlify Deploy Settings:~~

- ~~Structure - file changes - the biggest is `/docs` is now preferred to be `website/docs/..` Don't fight this! and it's because of plugin mechanisms and customization which is awesome actually. Docusaurus could power our entire website actually later and not just docs.~~
  - Structure updated for markdown files is `/docs/docs` and works OK! (just have to be careful and remember that website level is 1st /docs and 2nd /docs is the documentation path.
- ~~**Need help** fixing the broken `index` or routing which results in no pages or sidebar currently.  But everything is there in its right place as far as I can tell.~~

  - The above ended up being a necessary evil not in place in `docusaurus.config.js` which was missing top level `themes: []` and `plugins: []` and not just `presets: [...]`  Ugh!  what a headspinner! Added both as empty array and it fixed the content and sidebar displaying!  (Their docs didn't say anything being required there, and so I complained a bit on their Discord channel :-) )
- New Footer changes are breaking and so this needs further customizing footer if we want (/theme/Footer/Footer.js) is not picking up for some reason) See... https://v2.docusaurus.io/docs/migrating-from-v1-to-v2/#footer

**Some of the new stuff available in Docusaurus 2**
- **swizzle** is the way to go now for customizing stuff (search `swizzle` in the Docusaurus 2 docs)
- I did add some /src/css/custom.css for maintaining the coloring based on our colors in OpenRefine logo.
  - Just needs more added to it or tweaks as we go along


wetneb:
(I edited your message to remove the all caps - it would be good to avoid them in general)

Some of the files you moved around appear as deleted and added as new - it would be good to avoid that in general (as it breaks the history of those files).

I am not sure about using `website` as the root directory - it could potentially confuse contributors who are looking for the web app (OpenRefine itself). Perhaps that is a good case for moving the docs to a separate repo.

wetneb:
Also you have changed Netlify's settings, which means that all other PRs now fail since this one has not been merged. I am reverting to the previous settings now.

thadguidry:
> 
> 
> (I edited your message to remove the all caps - it would be good to avoid them in general)
> 
Oops, thanks!

> Some of the files you moved around appear as deleted and added as new - it would be good to avoid that in general (as it breaks the history of those files).
> 

Yes good point, and those few were at the beginning when I started the migration.

> I am not sure about using website as the root directory - it could potentially confuse contributors who are looking for the web app (OpenRefine itself). Perhaps that is a good case for moving the docs to a separate repo.

I thought about that also, for the same reasons.

> Also you have changed Netlify's settings, which means that all other PRs now fail since this one has not been merged. I am reverting to the previous settings now.

I also was going to revert back and got so excited after getting it to work that I actually forgot last night.  So thanks for doing that.

thadguidry:
Working on:
1. fixing `docs` top level instead of `website` and seems to work (just have to be careful with 2 levels of `docs`)
2. fixing the file rename/delete unneeded changes.

Stay tuned.

thadguidry:
@wetneb OK, I think it's ready now.

wetneb:
Nice! I think it is fine to have `docs/docs`, it is not super elegant but we can live with it.

The "User Manual" and "Technical Reference" links you added at the top do not work for me: it would probably be worth fixing this to show how basic navigation will work when they are in place.

@allanaaa: this PR will conflict with yours once merged. Is your PR #2740 still a test PR or does it include changes that you want to merge?

thadguidry:
> The "User Manual" and "Technical Reference" links you added at the top do not work for me: it would probably be worth fixing this to show how basic navigation will work when they are in place.
> 

Those links are in what is called the [Navbar](https://v2.docusaurus.io/docs/theme-classic/#navbar).
You didn't say what is wrong exactly.  Only that "something" needs fixing?  Can you explain?

We have tons of options for the Navbar (dropdown menus, sticky navbar, etc.) if we want:  https://v2.docusaurus.io/docs/theme-classic/#navbar-links

Anyways, the idea is that we can [Theme the Navbar](https://v2.docusaurus.io/docs/using-themes#theme-components) however we want, and it provides Navigation, to external content, additional docs, particular docs path, whatever.

wetneb:
> You didn't say what is wrong exactly. Only that "something" needs fixing? Can you explain?

Right. I thought that was obvious, but there you go:
* As a reader, when I go to the main page and then click "Technical Reference" in the Navbar, my browser takes me to `/docs/tech`, which does not exist, so I see a big "Page Not Found" error. I would prefer to see a page which is not an error.
* Same for the "User Manual" link.

Is that any clearer?

thadguidry:
@wetneb Yes, thanks.  Let me fix that now.

wetneb:
Another small thing: the GitHub links points to Docusaurus' repo, it should point to ours.

thadguidry:
@wetneb Ah, need to fix the link for GitHub in top right to make it go to edit.
Hmm, should we make that more clear with a tweak on the label, instead of just **GitHub**?
```
        {
          href: 'https://github.com/OpenRefine/OpenRefine/edit/master/docs',
          label: 'Edit on GitHub',
          position: 'right',
        },
```

allanaaa:
> @allanaaa: this PR will conflict with yours once merged. Is your PR #2740 still a test PR or does it include changes that you want to merge?

Nothing to merge! Delighted to see this up and running.

wetneb:
@thadguidry ideally the link should point to the page that needs to be edited, directly:
https://github.com/OpenRefine/OpenRefine/blob/master/docs/src/records_mode.md
(once this will be merged, the `src` will be replaced by `docs`)

wetneb:
Can we also remove the all caps in your placeholders?

thadguidry:
> @thadguidry ideally the link should point to the page that needs to be edited, directly:
> https://github.com/OpenRefine/OpenRefine/blob/master/docs/src/records_mode.md
> (once this will be merged, the `src` will be replaced by `docs`)

Ah, each page already has an "Edit Page" link along the bottom.  So, we're good there already.
According to their docs, that top right GitHub link is typically for going to the Project Repo (it can be customized with the GitHub cat logo or text, [just like there own docs site does](https://v2.docusaurus.io/docs/configuration/)).

thadguidry:
@wetneb Hi Antonin, what else is needed before you merge?

PR #2804, opened 2020-06-24
Updated Conflicts-with Scrutinizer tests using Mockito

darecoder1999:
updated test class by creating mocks for ConstraintFetcher

### DIFF
--- a/extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizerTest.java
+++ b/extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizerTest.java
@@ -1,5 +1,6 @@
 package org.openrefine.wikidata.qa.scrutinizers;
 
+import org.openrefine.wikidata.qa.ConstraintFetcher;
 import org.openrefine.wikidata.qa.MockConstraintFetcher;
 import org.openrefine.wikidata.testing.TestingData;
 import org.openrefine.wikidata.updates.ItemUpdate;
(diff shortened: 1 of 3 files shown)


extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizerTest.java
```diff
@@ -33,6 +45,19 @@ public void testTrigger() {
 
         ItemUpdate updateA = new ItemUpdateBuilder(idA).addStatement(statement1).addStatement(statement2).build();
 
+        PropertyIdValue propertyIdValue = Datamodel.makeWikidataPropertyIdValue("P2302");
+        EntityIdValue entityIdValue = Datamodel.makeWikidataItemIdValue("Q21502838");
+        PropertyIdValue property = Datamodel.makeWikidataPropertyIdValue("P2306");
+        Value propertyValue = Datamodel.makeWikidataPropertyIdValue("P31");
+        PropertyIdValue item = Datamodel.makeWikidataPropertyIdValue("P2305");
+        Value itemValue = Datamodel.makeWikidataItemIdValue("Q5");
```
wetneb, line 55: Perhaps it would be worth using more informative variable names here, to emphasize the role they play in the constraint definition. You should also be able to reuse the constants declared at the beginning of `ConflictsWithScrutinizer`.

extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/ScrutinizerTest.java
```diff
@@ -79,4 +92,33 @@ public void assertNoWarningRaised() {
     public Set<String> getWarningTypes() {
         return store.getWarnings().stream().map(w -> w.getType()).collect(Collectors.toSet());
     }
+
+    public void setFetcher(ConstraintFetcher fetcher) {
+        scrutinizer.setFetcher(fetcher);
+    }
+
+    public Stream<Statement> createStatementStream(PropertyIdValue propertyIdValue, EntityIdValue entityIdValue, PropertyIdValue property, Value propertyValue, PropertyIdValue item, Value itemValue) {
```
wetneb, line 100: The first parameter could be removed since it is always going to be P2302. The name of the method could be improved to suggest that those statements are constraints definitions.

extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/ScrutinizerTest.java
```diff
@@ -79,4 +92,34 @@ public void assertNoWarningRaised() {
     public Set<String> getWarningTypes() {
         return store.getWarnings().stream().map(w -> w.getType()).collect(Collectors.toSet());
     }
+
+    public void setFetcher(ConstraintFetcher fetcher) {
+        scrutinizer.setFetcher(fetcher);
+    }
+
+    public Stream<Statement> conflictsWithParamaterStatementStream(EntityIdValue entityIdValue, PropertyIdValue property, Value propertyValue, PropertyIdValue item, Value itemValue) {
```
wetneb, line 100: This method seems to be more generic than its name implies: it can generate any constraint definition which takes two different qualifiers as parameters.

Also I would make the type of the first argument more specific: because it is a value for the P2302 property, it should always be an ItemIdValue.

extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizerTest.java
```diff
@@ -70,6 +95,33 @@ public void testNoValueSnak() {
...
+        ConstraintFetcher fetcher = mock(ConstraintFetcher.class);
+        when(fetcher.getConstraintsByType(conflictsWithPid, "Q21502838")).thenReturn(statementStream);
+
+        setFetcher(fetcher);
+        scrutinize(updateA);
+        assertNoWarningRaised();
+    }
+
+    @Test
+    public void testNoStatement() {
```
wetneb, line 109: Really nice! That is exactly where the usefulness of mocking comes into play. Perhaps you could also add a test where multiple constraint definitions are present on the same property (we had found interesting examples of that on Wikidata).

wetneb:
Can you fix the merge conflicts first?

darecoder1999:
Resolved the merge conflicts :)

PR #2806, opened 2020-06-24
Create Project import improvements

### LINKED ISSUE #2800: File type detection does not recognize JSON in certain cases
**Describe the bug**

Creating a project by fetching JSON from certain URLs does not pick the JSON importer by default.

**To Reproduce**
1. Go to "Create project"
2. "Web addresses"
3. Use "https://query.wikidata.org/sparql?query=SELECT%20%3Fitem%20%3Furl%20WHERE%20%7B%0A%20%20%3Fitem%20p%3AP6269%20%5Bps%3AP6269%20%3Furl%20%3B%0A%20%20%20%20%20%20%20%20pq%3AP2700%20wd%3AQ458022%5D.%0A%7D%20LIMIT%20100&format=json"

**Current Results**

The default importer is "line-based text files"

**Expected behavior**

The JSON importer should be proposed by default.

**Desktop:**
 - OS: Windows, Linux

**OpenRefine :**
 - Version 3.3, still present in master

**Additional context**
Encountered in this video: https://www.twitch.tv/videos/659297903

### LINKED ISSUE #2805: Support fallback to Structured Syntax Name Suffixes in MIME types
We should support the use of [Structured Syntax Name Suffixes](https://tools.ietf.org/html/rfc6838#section-4.2.8) as a fallback for importers.

Currently we support `rdf+xml` and know that it's more specific than the generic `text/xml` so we should try the RDF importer before the XML importer, but if we encounter an unknown type with a syntax suffix, we don't prioritize the appropriate importer.

The idea would be that if we were to encounter `application/3gpdash-qoe-report+xml` or `application/activity+json` (to pick the first ones on the [list](https://www.iana.org/assignments/media-types/media-types.xhtml#application)), we would suggest the XML and JSON importers, respectively, as the first fallback options.

This is something we could implement directly in the importer framework, so perhaps it's orthogonal to the idea that @wetneb had for #2598.

_Originally posted by @tfmorris in https://github.com/OpenRefine/OpenRefine/issues/2598#issuecomment-625901620_

tfmorris:
Fixes #2805. Fixes #2800. This includes three fixes / improvements:

- Fix Content-Type handling to correctly extract the `charset` (which we call "encoding") - no issue
- Implement support for Structured Syntax Name Suffixes on media types so that we'll fall back to `text/json` or `text/xml` for mediatypes which have `+json` or `+xml` suffixes if we don't have an entry for the full type - issue #2800 / #2805
- Switch the default fallback type from `text/line-based` to `text` so that we try the most general guessing strategy. The other will only try to guess TSV, CSV, or Fixed formats, so will never find JSON or XML files with the character frequency guesser. - no issue

### DIFF
--- a/main/src/com/google/refine/importing/ImportingUtilities.java
+++ b/main/src/com/google/refine/importing/ImportingUtilities.java
@@ -44,6 +44,7 @@ import java.io.Reader;
 import java.io.UnsupportedEncodingException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.nio.charset.Charset;
 import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -74,6 +75,7 @@ import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
+import org.apache.http.entity.ContentType;
 import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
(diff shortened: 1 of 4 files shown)


main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -313,17 +315,20 @@ public void update(long bytesRead, long contentLength, int itemCount) {
                                 throw new Exception("No content found in " + url.toString());
                             }
                             InputStream stream2 = entity.getContent();
-                            String encoding = null;
-                            if (entity.getContentEncoding() != null) {
-                                encoding = entity.getContentEncoding().getValue();
-                            }
-                            JSONUtilities.safePut(fileRecord, "declaredEncoding", encoding);
```
wetneb, line 320: Is it never useful to read the encoding from `entity.getContentEncoding()`? Is the `getCharset()` call you replaced it with always preferable (or should there be a fallback mechanism)?

tfmorris: Never useful and not relevant. Content-Encoding specifies compression. In the case of your example, it will be `gzip`, but then get written to `null` when the request handler decompresses the content.

thadguidry: @tfmorris that's useful as a 1 line code comment there I would say. Disambiguating the `Encoding` term on the spot.

tfmorris: This conversation is about deleted code. That's why it's highlighted in red.

thadguidry: Oops, overlooked. Ok, well then not *there*

thadguidry:
@tfmorris Wow, this fixed a big problem for me in my daily work.  Detection is so much better!!!  It's like magic almost!  Thanks Tom!

tfmorris:
Thanks for testing and reporting back.

>

PR #2807, opened 2020-06-24
New implementation for Multivalue Scrutinizer

darecoder1999:
Implemented inner class & improved tests using mocks.

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/MultiValueScrutinizer.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/MultiValueScrutinizer.java
@@ -3,15 +3,36 @@ package org.openrefine.wikidata.qa.scrutinizers;
 import org.openrefine.wikidata.qa.QAWarning;
 import org.openrefine.wikidata.updates.ItemUpdate;
 import org.wikidata.wdtk.datamodel.interfaces.PropertyIdValue;
+import org.wikidata.wdtk.datamodel.interfaces.Snak;
+import org.wikidata.wdtk.datamodel.interfaces.SnakGroup;
 import org.wikidata.wdtk.datamodel.interfaces.Statement;
+import org.wikidata.wdtk.datamodel.interfaces.Value;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 public class MultiValueScrutinizer extends EditScrutinizer {
 
     public static final String new_type = "multi-valued-property-is-required-for-new-item";
     public static final String existing_type = "multi-valued-property-is-required-for-existing-item";
+    public static String MULTI_VALUE_CONSTRAINT_QID = "Q21510857";
+    public static String MULTI_VALUE_CONSTRAINT_STATUS = "P2316";
+
+    class MultivalueConstraint {
+        Value constraintStatus;
+
+        MultivalueConstraint(Statement statement) {
+            List<SnakGroup> snakGroupList = statement.getClaim().getQualifiers();
+            for(SnakGroup group : snakGroupList) {
+                for (Snak snak : group.getSnaks()) {
+                    if (group.getProperty().getId().equals(MULTI_VALUE_CONSTRAINT_STATUS)){
+                        constraintStatus = snak.getValue();
+                    }
+                }
+            }
+        }
+    }
 
     @Override
     public void scrutinize(ItemUpdate update) {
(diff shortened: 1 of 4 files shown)


extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/MultiValueScrutinizerTest.java
```diff
@@ -19,6 +24,9 @@ public void testNoIssue() {
         ItemIdValue idB = TestingData.matchedId;
         ItemUpdate update = new ItemUpdateBuilder(idA).addStatement(TestingData.generateStatement(idA, idB))
                 .addStatement(TestingData.generateStatement(idA, idB)).build();
+        ConstraintFetcher fetcher = mock(ConstraintFetcher.class);
+        when(fetcher.hasMultiValue(anyObject())).thenReturn(true);
```
wetneb, line 28: I think the end goal is to remove any methods that are specific to a given constraint from the `ConstraintFetcher` interface. So I would rather mock a generic method (like `getSingleConstraint`) and make sure the scrunizer calls that.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/MultiValueScrutinizer.java
```diff
@@ -3,15 +3,36 @@
...
 import java.util.Map;
 
 public class MultiValueScrutinizer extends EditScrutinizer {
 
     public static final String new_type = "multi-valued-property-is-required-for-new-item";
     public static final String existing_type = "multi-valued-property-is-required-for-existing-item";
+    public static String MULTI_VALUE_CONSTRAINT_QID = "Q21510857";
+    public static String MULTI_VALUE_CONSTRAINT_STATUS = "P2316";
+
+    class MultivalueConstraint {
```
wetneb, line 22: Is this inner class actually used in the scrutinizer?

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/MultiValueScrutinizer.java
```diff
@@ -3,15 +3,36 @@
...
+        Value constraintStatus;
+
+        MultivalueConstraint(PropertyIdValue propertyIdValue) {
```
wetneb, line 25: I would rather pass a `Statement` as parameter, and do the initial fetching externally. This ensures that an instance of this inner class corresponds to a constraint declaration.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/MultiValueScrutinizer.java
darecoder1999: Actually, no. Because it has just one parameter as constraint status, which we are not currently using to raise warnings. Should we consider that? then only this class makes any sense.

wetneb: I think it the general structure of scrutinizers should be like this:
- fetch all constraint statements of a given type via the fetcher
- map these to constraint definitions (`.map(statement -> new MultivalueConstraint(statement))`)
- for each constraint definition, run the scrutinizing code

So in this case the inner class would still be "used": the scrutinizing code would just not rely on it to fetch parameters for now. In the future, when we add support for exceptions, this class will still be useful :) Or we could replace it simply by the base class that fetches generic constraint parameters, too.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/MultiValueScrutinizer.java
```diff
@@ -29,6 +51,7 @@ public void scrutinize(ItemUpdate update) {
         if (update.isNew()) {
             for (PropertyIdValue pid : propertyCount.keySet()) {
                 if (propertyCount.get(pid) == 1) {
+                    List<MultivalueConstraint> constraints = _fetcher.getConstraintsByType(pid, MULTI_VALUE_CONSTRAINT_QID).map(statement -> statement == null ? null : new MultivalueConstraint(statement)).collect(Collectors.toList());
```
wetneb, line 54: When does it make sense for `getConstraintsByType` to return a null value in its stream? Intuitively I think these should be filtered out upstream, no?

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/MultiValueScrutinizer.java
```diff
@@ -3,15 +3,36 @@
...
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 public class MultiValueScrutinizer extends EditScrutinizer {
 
     public static final String new_type = "multi-valued-property-is-required-for-new-item";
     public static final String existing_type = "multi-valued-property-is-required-for-existing-item";
+    public static String MULTI_VALUE_CONSTRAINT_QID = "Q21510857";
+    public static String MULTI_VALUE_CONSTRAINT_STATUS = "P2316";
```
wetneb, line 20: This "constraint status" property is not specific to the multi value constraint, it is applicable to all other constraint definitions, so it should rather be defined in a generic location. We can leave this as it is right now, but in the future it really ought to be used by a class that all constraint definitions inherit from.

PR #2812, opened 2020-06-25
Strip parentheses from imported filenames when constructing default project name

### LINKED ISSUE #2776: Clipboard imports now get named as "(clipboard)" instead of "clipboard"
**Describe the bug**

One side effect of #2720 is that clipboard imports now get named "(clipboard)" by default instead of "clipboard". Was this intended?

Since this is not something I was expecting I would propose to only include this in 3.5, not 3.4, since there is a risk that there are other unforeseen consequences. Not that I think this particular change is necessarily bad (we might actually want to keep the brackets), just that this is not something I was expecting.

**To Reproduce**
1. Type in anything in the clipboard import area
2. Validate
3. See the default project name in the top right text input

**Current Results**

The project name is "(clipboard)" by default.

**Expected behavior**

So far "clipboard" was the default.

**Additional context**

This new name seems to be generated in the backend, so I was about to propose to localize it but unless we manage to generate it in the frontend instead, it might not be doable at the moment.

wetneb:
Fixes #2776.


wetneb:
(my preferred course of action would be to remove the commit which introduced the change from 3.4, leaving it in master)

wetneb:
I think we all agree that this is not going in the right direction.

PR #2817, opened 2020-06-25
Change pref name for ui.browsing.pageSize

antoine2711:
(Related to I #2624)

Change the preference key name `ui.gridPaginationSize` to `ui.browsing.pageSize`.

@wetneb: per @tfmorris suggestion in I #2520, I would change the preference key.
Since it's the choices, is could also be: `ui.browsing.pageSizeChoices`. But that is probably too long.

Regards,
   Antoine

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
+++ b/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
@@ -34,7 +34,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 function DataTableView(div) {
   this._div = div;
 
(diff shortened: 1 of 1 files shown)


wetneb:
For now this is only documented in the changelog as far as I know.

thadguidry:
I added it to the growing list: 

Ugh, Let me port the hack listing we have in an [issue comment](https://github.com/OpenRefine/OpenRefine/issues/1677#issuecomment-405937473) instead over to a a nice Wiki page!

DONE 
  - Created new page "Preferences" https://github.com/OpenRefine/OpenRefine/wiki/Preferences

And updated our wiki Home - Preferences link to point to that new page.
Now we have a place to begin to document for @allanaaa and the technical writer all the Preferences that can be set.
We or @allanaaa can reorganize that listing on that wiki page at some point into a set of tables (with a column for expected type like Array, String, etc.)

PR #2818, opened 2020-06-26
Implemented SingleValue Scrutinizer tests using mocks

darecoder1999:
Updated test class & added inner class to the scrutinizer

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/SingleValueScrutinizer.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/SingleValueScrutinizer.java
@@ -34,7 +34,6 @@ import org.wikidata.wdtk.datamodel.interfaces.Value;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 /**
  * For now this scrutinizer only checks for uniqueness at the item level (it
@@ -50,6 +49,7 @@ public class SingleValueScrutinizer extends EditScrutinizer {
 
     public static final String type = "single-valued-property-added-more-than-once";
     public static String SINGLE_VALUE_CONSTRAINT_QID = "Q19474404";
+    public static String SINGLE_BEST_VALUE_CONSTRAINT_QID = "Q52060874";
     public static String SINGLE_VALUE_CONSTRAINT_EXCEPTION = "P2303";
 
     class SingleValueConstraint {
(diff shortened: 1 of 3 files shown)


extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/SingleValueScrutinizer.java
```diff
@@ -44,6 +49,24 @@
 public class SingleValueScrutinizer extends EditScrutinizer {
 
     public static final String type = "single-valued-property-added-more-than-once";
+    public static String SINGLE_VALUE_CONSTRAINT_QID = "Q19474404";
+    public static String SINGLE_VALUE_CONSTRAINT_EXCEPTION = "P2303";
+
+    class SingleValueConstraint {
```
wetneb, line 55: I think all exceptions can accept constraints, so how about we pull the parsing code in this class to a common superclass of all the inner classes you have been introducing?

darecoder1999: Sure, we can make a constraint class that will be extended by these inner classes. And check what all parameters can be added over there other than exceptions that are generally common to all property constraints.

wetneb: Great! Constraint status is another one.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/SingleValueScrutinizer.java
```diff
@@ -44,6 +46,17 @@
...
     public static final String type = "single-valued-property-added-more-than-once";
+    public static String SINGLE_VALUE_CONSTRAINT_QID = "Q19474404";
+    public static String SINGLE_BEST_VALUE_CONSTRAINT_QID = "Q52060874";
+
+    class SingleValueConstraint extends Constraint {
+        Set<EntityIdValue> constraintExceptions;
+
+        SingleValueConstraint(Statement statement) {
+            super(statement);
+            constraintExceptions = getConstraintExceptions();
```
wetneb, line 57: You should not need to declare a new field for constraint exceptions again: you can just reuse the one from the super class. So, for all constraints which do not take any parameters, you should not need to declare any inner class: the `Constraint` class itself can be used in the scrutinizer.

Issue #2820, opened 2020-06-26
Include JSON as a possibility to be guessed for .txt files


thadguidry:
Files sometimes have different content that does not match the file extension.  For example, on Windows depending on the software you have installed, saving a file with JSON content might default to saving it with a `.txt` extension instead of the correct `.json`.  On the internet, files might have various extensions or no extensions when downloaded and saved.  This is not OpenRefine's direct problem, but OpenRefine should be able to inspect (guess) what content a file truly has.

### Proposed solution
`.txt` extension and no extension are 2 of the common ways that files are sometimes found with structured data inside them.
When dealing with a `.txt` file or no extension, it would be preferable for OpenRefine to first read a portion of a file's beginning content and using that as the first contextual clue of what kind of file it is.

For example:
OpenRefine does not guess JSON format when opening a `.txt` file on Windows 10 that is 100% JSON content.  It guesses `line-based` and uses that importer instead, incorrectly.  OpenRefine should instead have guessed JSON as the bestFormat in the rankedFormats, by inspecting the beginning of the file chars are JSON content, maybe 50 chars or more?)

Other file extensions normally don't cause context problems in the wild.  But `.txt` is the most common format and the one that software and some OS's sometimes default to saving files with and as such `.txt` should be treated as that, "a generic bunch of text that first needs to be inspected for some kind of structured data" in order to determine the right kind of importer.

**Describe alternatives you've considered**
Opening the file in Notepad and then looking at the content of the file myself to make a judgement call of what format it might contain and then clicking the appropriate importer selection on OpenRefine's UI.

**Test File**
[Testfile.txt](https://github.com/OpenRefine/OpenRefine/files/4845532/Testfile.txt)

**Additional Context**
Breakpoint on L970 in ImportingUtilities.java seems like a good start for debugging from what I saw and where it gave text/line-based as the best format in the ranked list.

tfmorris:
Change this line https://github.com/OpenRefine/OpenRefine/blob/4b146acc6ef8a4f41168824db0967e24d7efb45c/main/webapp/modules/core/MOD-INF/controller.js#L238

to `"text"` instead of `"text/line-based"`. The line-based guesser only handles CSV, TSV, & Fixed.

thadguidry:
@tfmorris tried it, didn't make a difference in the ranking, it still automatically choose line-based in the selection after I choose to load that test file.

[...]

PR #2821, opened 2020-06-26
Update in Controller.js

chetan-v:
In the Reference of #2820 

Fix the typo "text" instead of "text/line-based".


chetan-v:
Can you explain more about how to test it @tfmorris for now I am just clone it
.

tfmorris:
An automated test would be a good addition, but you can start with manual testing. 

- Clone the repo
- apply your fix
- build the fixed OpenRefine
- run the built version
- create a new project using the test file
- confirm that it gets correctly guessed as JSON instead of Line-based text

chetan-v:
Screenshot from 2020-06-28 10-32-08

I add one more line

PR #2828, opened 2020-06-27
Fix update to deprecated Google Drive credential code

tfmorris:
No issue. Restore missing piece of commit 42354c0 so that Builder has the method parameter that it needs.

Not sure what happened here, but the code that I committed isn't the same as I tested with and it doesn't work. This fixes it.


wetneb:
@tfmorris I assume this needs to go to 3.4 as well then?

tfmorris:
The Google Sheets bug fixes don't depend on the Google Drive & Sheets dependency updates. I'd recommend putting the bug fixes in 3.4 and the dependency updates in 3.5. 

This is a patch for the dependency updates (and the gratuitous update of deprecated code that I included), so should go wherever it goes.

I can dig out the actual PR #s later if needed.

Issue #2753, opened 2020-06-18
"Import Project" screen misstatement


allanaaa:
**Describe the bug**
The "Import Project" screen incorrectly states that OpenRefine project exports can be accepted in .zip and .tar, as per https://groups.google.com/forum/#!topic/openrefine-dev/g5UUWpAQMYM 

**Expected behavior**
Should say "Locate an existing Refine project file (.tar.gz):"

**Screenshot**
![unnamed](https://user-images.githubusercontent.com/28012148/85065298-e982a080-b17a-11ea-907d-34e6f1940ee5.png)

Thanks!

wetneb:
Congrats for finding this!

To solve this, we just need to change the UI text. Changing it in English is enough, as Weblate is able to flag outdated translations to contributors (but in many languages it should not be hard to fix it directly I guess).

chetan-v:
I want to work on it

thadguidry:
Sure @chetan-v !

chetan-v:
Hi @thadguidry I am look on this issue but I cant find the source file to fix this issue

thadguidry:
@chetan-v Whenever you see labels (text written) somewhere in OpenRefine, you can do a file content search in the repo for that bit of text written, then it will show which file contains that text.
So do a file search(locally in your IDE or in GitHub) for some of that text: `Locate an existing Refine project file`

allanaaa:
It turns out I was mistaken about this issue: 
When you export a project archive to Google Drive, it creates a .zip. 
You can download that .zip and import it back into OpenRefine successfully. 
So, it accepts .zip and .tar.gz. 

I see that #2797 suggests there is a file-type error here; all I know is that OpenRefine successfully interprets its own exported zips as project archives.

allanaaa:
Whoops, just saw that #2829 will make this note obsolete. Ignore!

tfmorris:
Although named .zip by Google Drive, the contents are actually a .tar.gz. The importer lucks out because the logic tests for != .tar, so anything other than a .tar is assumed to be a .tar.gz. No unzip utility will be able to read those files.

PR #2831, opened 2020-06-28
Fix the typo

chetan-v:
In the reference of #2753

### DIFF
--- a/main/webapp/modules/core/MOD-INF/controller.js
+++ b/main/webapp/modules/core/MOD-INF/controller.js
@@ -236,7 +236,6 @@ function registerImporting() {
    *  Extension to format mappings
    */
   IM.registerExtension(".txt", "text/line-based");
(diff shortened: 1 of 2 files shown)


chetan-v:
> Hi @chetan-v, thank you very much for this!
> 
> The two issues you have tackled in this pull request are unrelated, so it would be better to do one pull request for each.
> 
> Concerning the change in UI text, I think `.tar` archives are supported too, so if that is indeed the case (please check) we should probably keep mentioning that in the UI.

Then what should I do add .tar in it
Or make separate PR

chetan-v:
And I don't know how to separate them 
How to delete one commit ?
When I push it from my local repo it will automatically add in this PR
Not show another create PR option.
That's why I am just Close my Two PR

darecoder1999:
@chetan-v make separate branches for separate issues. Currently, you have made changes to your master branch.

PR #2846, opened 2020-06-30
Ignore events which don't change text input - fixes #1134

### LINKED ISSUE #1134: First pagination action after applying text filter fails

tfmorris:
Fixes #1134.

### DIFF
--- a/main/webapp/modules/core/scripts/facets/text-search-facet.js
+++ b/main/webapp/modules/core/scripts/facets/text-search-facet.js
@@ -149,7 +149,7 @@ TextSearchFacet.prototype._initializeUI = function() {
     this._elmts.input[0].value = this._query;
   }
   
-  elmts.input.bind("keyup change input",function(evt) {
+  this._elmts.input.bind("keyup change input",function(evt) {
     // Ignore events which don't change our input value
     if(this.value === self._query || this.value === '' && !self._query) {
       return;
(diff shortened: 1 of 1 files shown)


main/webapp/modules/core/scripts/facets/text-search-facet.js
```diff
@@ -141,9 +141,9 @@ TextSearchFacet.prototype._initializeUI = function() {
     this._elmts.input[0].value = this._query;
   }
   
-  this._elmts.input.bind("keyup change input",function(evt) {
-    // Ignore non-character keyup changes
-    if(evt.type === "keyup" && (this.value === self._query || this.value === '' && !self._query)) {
+  elmts.input.bind("keyup change input",function(evt) {
```
lisa761, line 144: This line throws an error:



If I switch it back to `this._elmts`, it works just fine and the `next` and `last` buttons work as well.

tfmorris:
@lisa761 Thanks for testing! Feel free to use it, if it's helpful.

main/webapp/modules/core/scripts/facets/text-search-facet.js
tfmorris: Fixed.

Issue #2835, opened 2020-06-28
404 on /fr/fr.json during Project Open


antoine2711:
**Describe the bug**
When loading a project, in the console, if you choose a userLang different than `en`, you get two error lines, `
Failed to load resource: the server responded with a status of 404 (Not Found)` and `:3333/fr/en.json:1 Failed to load resource: the server responded with a status of 404 (Not Found)`

**To Reproduce**
1. Choose another language than english
2. Load a project
3. Open JS Console

**Current Results**
Note that in english, you have just ONE error line for `EN`.

**Screenshots**
![image](https://user-images.githubusercontent.com/12217525/85959399-4ac02600-b96a-11ea-844f-1cdab0b2916a.png)

**Desktop:**
 - OS: MacOS X.14.6
 - Browser Version: Chrome latest
 - `usarLang` = '`fr`'

**OpenRefine :**
 - Version v3.5 (master)

tfmorris:
A cursory look at the call stack indicates this is a bug in the database extension.

wetneb:
Closed by #2847.

PR #2847, opened 2020-06-30
Fix i18n. Fixes #2805

### LINKED ISSUE #2805: Support fallback to Structured Syntax Name Suffixes in MIME types

tfmorris:
Fixes #2805. 
Fix database extensions exporter which is corrupting the dictionary name with the value of the language.

### DIFF
--- a/extensions/database/module/scripts/project/database-exporters.js
+++ b/extensions/database/module/scripts/project/database-exporters.js
@@ -38,7 +38,7 @@ $.ajax({
     },
     success : function(data) {
         dictionary = data['dictionary'];
-        dictionary = data['lang'];
+        lang = data['lang'];
     }
 });
 $.i18n().load(dictionary, lang);


wetneb:
I believe you meant #2835.

PR #2859, opened 2020-07-01
Micro benchmark harness & ToNumber optimizations

tfmorris:
This PR provides a benchmarking subproject which is set up with the Java Microbenchmark Harness along with a benchmark of a performance optimized version of ToNumber which is about 6x faster for floats and less than 10% slower for integers.

Not being a Maven guru, there's still an issue with the shaded JAR not pulling in all the OpenRefine classes, but you can work around that in the short term using the shell script `benchmark/run.sh` which just runs the command:

    java -cp benchmark/target/benchmarks.jar:main/target/openrefine-main.jar org.openjdk.jmh.Main

Here is the output of the benchmark run comparing old and new for both Longs and Doubles:

```
Benchmark                      (iterations)  Mode  Cnt     Score     Error  Units
ToNumberBenchmark.toDoubleNew          1000  avgt   25   316.401 ±  35.192  ns/op
ToNumberBenchmark.toDoubleNew         10000  avgt   25   328.481 ±  51.092  ns/op
ToNumberBenchmark.toDoubleOld          1000  avgt    5  1969.738 ± 474.616  ns/op
ToNumberBenchmark.toDoubleOld         10000  avgt    5  1926.492 ± 577.219  ns/op
ToNumberBenchmark.toLongNew            1000  avgt    5    98.340 ±   8.915  ns/op
ToNumberBenchmark.toLongNew           10000  avgt    5    98.219 ±   7.526  ns/op
ToNumberBenchmark.toLongOld            1000  avgt    5    90.516 ±  10.439  ns/op
ToNumberBenchmark.toLongOld           10000  avgt    5    93.558 ±   7.501  ns/op
```


thadguidry:
@tfmorris So it looks like the basic code optimization is that we now more quickly look for a hint of a pattern of a number in the String and if we see none, then quickly skip extra branching to attempt conversion that is unnecessary.  If there is a hint of a pattern of a number, then we attempt further branching to convert the string to float or double.  We basically now fail faster for conversions compared to before, is that about right?

tfmorris:
> @So it looks like the basic code optimization is that we now more quickly look for a hint of a pattern of a number in the String and if we see none, then quickly skip extra branching to attempt conversion that is unnecessary.

Not exactly, but why do you ask? Perhaps I can answer the underlying question that is motivating that question. The changes are enumerated in this commit message https://github.com/OpenRefine/OpenRefine/pull/2859/commits/423f98e65ce38fcf9dcfe4d967a7d09274e45860.

tfmorris:
FYI - for reviewers - the drop in coverage is due solely to still having the old method so that it can be used for benchmarking. When that is deleted, the coverage should be unchanged.

wetneb:
Nice! I wonder what is the expected workflow around this. Do you intend to remove the old implementation when we merge this PR? In that case, the benchmark module will only run the new implementation, so we will have lost the comparison point - is it still useful to have the benchmark itself in that case?

When should these benchmarks be run? (Is that something you would add to the CI as a test, to make sure we do not degrade the performance of a particular part of the code?)

Also, I am curious how you came to work on this specific part of the code - is it something that was a bottleneck for a certain OR workflow, or was this flagged by a code analysis tool perhaps?

Finally, this adds new dependencies in `main/pom.xml`, which will therefore be included in packaged distributions. Is this necessary? Intuitively, all dependencies related to benchmarking should be required outside this module, so that they are only used when running these benchmarks.

tfmorris:
I'll remove the old code now that we have the performance numbers. 

The setup instructions said to use a mvn generate command with the old project, but when I tried to use the `main` project instead of the top level, it said "not type `pom`" or something similar. I can investigate what it would take fix this (or you know the fix off the top of your head, please feel free to adjust things). Someone also needs to figure out the shading (although perhaps that'll fix itself with the new project structure).

I'll add some of the other questions (and my answers) to the mailing list performance thread.

PR #2862, opened 2020-07-02
Implemented QuantityScrutinizer tests using Mocks

darecoder1999:
Updated test cases and added AllowedUnitsConstraint class.

### DIFF
--- a/extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/QuantityScrutinizerTest.java
+++ b/extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/QuantityScrutinizerTest.java
@@ -1,11 +1,32 @@
 package org.openrefine.wikidata.qa.scrutinizers;
 
-import java.math.BigDecimal;
-
-import org.openrefine.wikidata.qa.MockConstraintFetcher;
+import org.openrefine.wikidata.qa.ConstraintFetcher;
+import org.openrefine.wikidata.testing.TestingData;
+import org.openrefine.wikidata.updates.ItemUpdate;
+import org.openrefine.wikidata.updates.ItemUpdateBuilder;
 import org.testng.annotations.Test;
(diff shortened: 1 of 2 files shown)


extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/QuantityScrutinizerTest.java
```diff
@@ -24,10 +42,17 @@
             new BigDecimal("1.545"));
     
     private QuantityValue wrongUnitValue = Datamodel.makeQuantityValue(
-            new BigDecimal("1.234"), "Q346721");
+            new BigDecimal("2"), "Q346721");
```
wetneb, line 45: Why did you need to change this value? Isn't this changing the meaning of the tests below?

extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/QuantityScrutinizerTest.java
```diff
@@ -36,67 +61,190 @@ public EditScrutinizer getScrutinizer() {
...
     @Test
     public void testBoundsAllowed() {
-        scrutinize(valueWithBounds);
+        ItemIdValue idA = TestingData.existingId;
+        Snak mainSnak = Datamodel.makeSomeValueSnak(propertyIdValue);
+        Statement statement = new StatementImpl("P1083", mainSnak, idA);
+        ItemUpdate update = new ItemUpdateBuilder(idA).addStatement(statement).build();
+
+        ConstraintFetcher fetcher = mock(ConstraintFetcher.class);
+        when(fetcher.getConstraintsByType(propertyIdValue, "Q51723761")).thenReturn(new ArrayList<>());
```
wetneb, line 70: It would be better to reuse the constant declared in the scrutinizer, NO_BOUNDS_CONSTRAINT_QID (and same for all other tests).

PR #2865, opened 2020-07-02
Remove shortest-column-name ordering - fixes #2863

### LINKED ISSUE #2863: XML importer option to maintain element ordering during column creation
The XML importer creates columns for the XML elements but doesn't always maintain the order of those elements against the column index.

With the following XML snippet in the clipboard and imported:

```xml
<records><record><id>1</id><height>11</height><width>111</width></record><record><id>2</id><height>22</height><width>222</width></record><record><id>3</id><height>33</height><width>333</width></record></records>
```

It would be helpful if OpenRefine could have a new checkbox option for the XML importer to try to maintain the Column indexing by next XML element found, for example:
```xml
<record>
  <id>1</id>
  <height>11</height>
  <width>111</width>
</record>
```
1 id
2 height
3 width

instead of losing the ordering during Column creation as shown below and screenshot:
1 record - id
3 record - width
2 record - height

image(2)

Reordering the columns as necessary after project creation.  If many elements, then lots of reordering is needed, taking several minutes using the **Edit column -> Move column** over and over to get the indexing to match the original XML element ordering.

**Additional context**

None

tfmorris:
Fixes #2863
The tree importer sorts columns/column groups by how populated
they are, which is of arguable utility, but the tie-breaker
of ordering by shortest column name is completely silly.

This change removes that and, in conjunction with a stable sort
algorithm, will preserve the original order of the columns.

### DIFF
--- a/main/src/com/google/refine/importers/tree/TreeImportUtilities.java
+++ b/main/src/com/google/refine/importers/tree/TreeImportUtilities.java
@@ -73,8 +73,7 @@ public abstract class TreeImportUtilities {
                     return o1.blankOnFirstRow ? 1 : -1;
                 }
 
-                int c = o2.nonBlankCount - o1.nonBlankCount;
-                return c != 0 ? c : (o1.name.length() - o2.name.length());
+                return o2.nonBlankCount - o1.nonBlankCount;
             }
         });
 
@@ -85,6 +84,7 @@ public abstract class TreeImportUtilities {
             project.columnModel.columns.add(column);
         }
 
+        // The LinkedHashMap iterator will guaranteed that the list is arranged in order found
         List<ImportColumnGroup> subgroups = new ArrayList<ImportColumnGroup>(columnGroup.subgroups.values());
         Collections.sort(subgroups, new Comparator<ImportColumnGroup>() {
             @Override
@@ -96,9 +96,8 @@ public abstract class TreeImportUtilities {
                 // TODO: From a human factors point of view, we probably want
                 // to try to preserve the order that we found things in the XML
                 
-                // Sort by most populated first, then shortest name
-                int c = o2.nonBlankCount - o1.nonBlankCount;
-                return c != 0 ? c : (o1.name.length() - o2.name.length());
+                // Sort by most populated first, but leave order unchanged if they're equal
+                return o2.nonBlankCount - o1.nonBlankCount;
             }
         });
 
--- a/main/tests/server/src/com/google/refine/importers/XmlImporterTests.java
+++ b/main/tests/server/src/com/google/refine/importers/XmlImporterTests.java
@@ -51,6 +51,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.refine.importers.XmlImporter;
 import com.google.refine.importers.tree.TreeImportingParserBase;
 import com.google.refine.importing.ImportingJob;
+import com.google.refine.model.ColumnGroup;
 import com.google.refine.model.Row;
 import com.google.refine.util.JSONUtilities;
 import com.google.refine.util.ParsingUtilities;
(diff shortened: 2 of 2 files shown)


thadguidry:
@tfmorris Thanks Tom.  Do you think we need to add a few more XML tests, or that what's there now covers enough of what kinds of XML structures would be seen in the wild now?

tfmorris:
No idea. I didn't review the tests other than to the test that was affected by the change. I couldn't tell by its structure whether it was written to explicitly test for shortest column name first or just to test the observed behavior when the test was written. I suspect the latter. Feel free to add any tests that you think would be useful.

tfmorris:
> I think the main risk arises in situations where the first column of the project ends up being different, which changes the record boundaries - but I am not sure if this can actually happen with this change?

Yes, it definitely can happen if the columns are populated to the same extent and the column names have different lengths. And, yes, changing record boundaries would be the most dramatic change, but even in purely row mode columns could potentially change position, so any index-based, as opposed to name-based, indexing could break.

> In principle we should probably treat this as a breaking change, since workflows depend on column positions, but we lack a formal policy about that.

"workflow" meaning what? We've always been adamant that the internal RESTful API is just that, internal. The only grey area that I see is any assurances which were given around the portability of extract/apply of Undo history across projects and/or versions. I've always viewed that as "best efforts" only beyond the undo/redo within the last few minutes/hours/days. There are myriad other ways that the subsequent "Apply" operation can (and do) fail.

We can start versioning Undo histories (a good idea regardless) and rejecting mismatched histories, but we should introduce that at the same time as this if we want to be strict about then handling of Undo histories. We can also verify that Undo histories are only being applied against a) the same project and b) with a compatible version of OpenRefine. When we introduce actual scripting support, of course all these restrictions will be relaxed since the DSL / API / etc will be engineered for compatibility in mind.

wetneb:
Yes by "workflow" I mean "Undo history" in your words.

ostephens:
> We can start versioning Undo histories (a good idea regardless) and rejecting mismatched histories, but we should introduce that at the same time as this if we want to be strict about then handling of Undo histories. We can also verify that Undo histories are only being applied against a) the same project and b) with a compatible version of OpenRefine.

- Very much in favour of versioning Undo histories
- Very much against restricting histories to only run on the same project

tfmorris:
Sounds like folks are in favor of encouraging people to shoot themselves in the foot. I don't think I'm in the same camp.

Note also that the current tree importer behavior is data dependent, both before and after this change, so there's no way to predict where any given column will end up without examining the entire file.

ostephens:
>Sounds like folks are in favor of encouraging people to shoot themselves in the foot. I don't think I'm in the same camp.
IMO there are clear use cases for carrying out a repeated set of operations on different files - and I've used this facility as a user and recommended it as a trainer.

I can certainly see an argument for adding warnings in this situation, but I don't think preventing it happening is a good idea

>Note also that the current tree importer behavior is data dependent, both before and after this change, so there's no way to predict where any given column will end up without examining the entire file.
I'm not arguing against this particular PR - I'm happy to see this change applied. I was just commenting on your comment - I do think we need versioning on undo histories to try to avoid the question of breaking changes causing users problems

wetneb:
> Sounds like folks are in favor of encouraging people to shoot themselves in the foot. I don't think I'm in the same camp.

I suspect there is just a misunderstanding here. Personally, the way I describe this feature to newcomers is along the lines of: "the cleaning procedure you have built interactively can be extracted and reapplied automatically when you have a new version of this dataset next year". Importing a new version of the dataset implies creating a new project, so forbidding to apply a workflow ("Undo history") on a different project than the one it was extracted from would remove this use case.

(In fact, as far as I am concerned, I would say it is very rare that I apply a recipe on the same project it was extracted from - it is the exception rather than the norm)

thadguidry:
Agree with Antonin, its the exception rather than the norm.  Most people are not even using Apply History on a monthly basis. Most are doing normalizing, cleanup and changing from one format to another.  https://openrefine.org/blog/2020/02/20/2020-survey-results.html
We would have to check with Martin on the last survey details, but I think there was only 2 people? that asked for improvements for History.  I don't even use Apply History because it not even close to a "workflow" tool like other programs.

So, Tom, I'm not worried really much at all and don't think we'll have many or any users complainging about Apply History uses as a good "workflow" concerning XML/JSON transformations.

tfmorris:
FWIW, my standard description goes along the lines of "There's a half-built feature that lets you extract an internal representation of an operation history and reapply it, but it's got no error handling, versioning, or many other features that it would need to be usable on a production basis. Despite this, some people have built client applications which exploit OpenRefine's internal-only client/server protocol and this operation history to build poor man's automation systems. This is unsupported by the OpenRefine team and could break at any time."

If anyone is telling users something different, you're leading them down the garden path (or committing yourself to an unsustainable maintenance burden).

>  I've used this facility as a user and recommended it as a trainer.

Hopefully this is taught with all the appropriate caveats about it not being reliable, etc and only taught to advanced students who are sophisticated enough to be able to make the judgement call about when to take the risk of undetected problems.

wetneb:
@tfmorris I think you are mixing two unrelated things here: the HTTP API has nothing to do with this. OpenRefine already exposes the ability to reapply JSON histories in its UI, one does not need to reverse-engineer the internal HTTP API to use this. So please do not say that "This is unsupported by the OpenRefine team and could break at any time" - this is a major feature of the tool. I think we all agree that this feature could be improved in many ways, but let us not claim that this feature is deprecated - if it were, then we should remove this from the UI altogether, and I think that would be a very unpopular change.

PR #2867, opened 2020-07-03
Fixed the guessing of JSON for .txt(2820)

### LINKED ISSUE #2820: Include JSON as a possibility to be guessed for .txt files

chetan-v:
Fixes #2820 

In Extension to format mappings :- change "text/line-based" to "text"(238 line)
In Mime type to format mappings :-  change "text/line-based" to "text"(269 line)

### DIFF
--- a/main/webapp/modules/core/MOD-INF/controller.js
+++ b/main/webapp/modules/core/MOD-INF/controller.js
@@ -235,7 +235,7 @@ function registerImporting() {
   /*
    *  Extension to format mappings
    */
-  IM.registerExtension(".txt", "text/line-based");
+  IM.registerExtension(".txt", "text");
   IM.registerExtension(".csv", "text/line-based/*sv");
   IM.registerExtension(".tsv", "text/line-based/*sv");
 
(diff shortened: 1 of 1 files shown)


thadguidry:
@chetan-v Thank you Chetan! Congrats!

PR #2870, opened 2020-07-03
Report HTTP error codes to the user when creating a project from a URL

### LINKED ISSUE #2837: Creating a project from a URL - HTTP errors are ignored
**Describe the bug**
OpenRefine does not show HTTP errors when creating a project from a URL.
Instead it acts like it downloaded successfully but there is nothing shown in the preview data grid.

**To Reproduce**
1. Create new project from URL
`https://download.data.world/file_download/associatedpress/covid-impact-survey-public-data/COVID_W2_toplines.zip`
2. Click on Next>>
3. No error is given
4.  Preview data grid screen shows as if download completed successfully but is blank

**Current Results**
Preview data grid screen is blank. No HTTP error is shown to user

**Expected behavior**
Should not have gone to preview data grid screen because a 401 error was returned in the response.

**Screenshots**

image
image

**Desktop:**
 - OS: Windows 10
 - Browser Version: Firefox
 - JRE or JDK Version: JDK8

**OpenRefine :**
 - Version:  OpenRefine (trunk)

urvashigupta7:
Fixes #2837


urvashigupta7:
@wetneb Please review 😄

thadguidry:
@urvashigupta7 This is great!  

image

However, this doesn't tell the user the right message.  Typically that would come from the server itself.  I think it might be better in the message to the user to just return the Status Code, Message | JSON | HTML as the Exception message?

```
{"code":401,"request":"13a9aff6-7eea-437c-a8e9-590e98cf8e43","type":null,"message":"Unauthorized","responseHeaders":null}
```

You can see an example of this sort of handling we do in other places like here:
https://github.com/OpenRefine/OpenRefine/blob/49fd21759c81cfccd08fce8b6ddf20a427bbf2a0/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java#L434

urvashigupta7:
I have made the changes. And yes @wetneb I would like to write test also.

wetneb:
Great! To write a test I recommend you use `okhttp3.mockwebserver.MockWebServer` as done in this test: https://github.com/OpenRefine/OpenRefine/blob/749704518c330c6f96ec3e784c45b8c055d5c6cd/main/tests/server/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperationTests.java#L198-L202

You should add your test as a new method of `ImportingUtilitiesTest`.
Let me know if you get stuck.

main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -309,11 +310,18 @@ public void update(long bytesRead, long contentLength, int itemCount) {
...
 
                         try {
-                            response.getStatusLine();
+                            StatusLine status = response.getStatusLine();
                             HttpEntity entity = response.getEntity();
                             if (entity == null) {
                                 throw new Exception("No content found in " + url.toString());
                             }
+                            int statusCode = response.getStatusLine().getStatusCode();
+                            if (statusCode / 100 != 2) {
```
tfmorris, line 319: The standard way to do this is to consider anything greater or equal to 400 to be an error. 
```suggestion
                            if (statusCode >= 400) {
```

main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -309,11 +310,18 @@ public void update(long bytesRead, long contentLength, int itemCount) {
                         CloseableHttpResponse response = httpclient.execute(httpGet);
 
                         try {
-                            response.getStatusLine();
+                            StatusLine status = response.getStatusLine();
```
tfmorris, line 313: It would be tidier to put this with the rest of the status processing code below.

urvashigupta7:
@tfmorris okay.. I will make the changes

urvashigupta7:
@wetneb Could you please summaries the steps to create the test. If I am not wrong MockWebServer will be used to return a 401 error code. But, I am not getting how urlimporting request needs to be send. Do I need to call this method? - https://github.com/OpenRefine/OpenRefine/blob/3a6fc90f02559944cacbbe926b19d363db892086/main/src/com/google/refine/importing/ImportingUtilities.java#L176

wetneb:
Yes this is the method that you will need to test. I would recommend to use Mockito to generate mocks for the arguments it expects, and set up the methods of these mocks to return the appropriate values to trigger the code path you have been working on. In particular, you can mock the HttpRequest object, such that you do not have to actually send any real HTTP request.

If you are not familiar with Mockito, I recommend this tutorial: https://www.baeldung.com/mockito-series

urvashigupta7:
Hi @wetneb. I am having trouble with sending Httpservletrequest object to the method. Actually, the request's content-type is multipart/form-data.
Could you please tell me how to send the data.

wetneb:
@urvashigupta7 please update this PR with what you have written so far, otherwise it is hard to help you without seeing the code you are working on.

urvashigupta7:
@wetneb I have pushed the code. Actually I need to send download URL with the request object. Sending URL with getParameter will not work as it need is multipart/form-data

wetneb:
Nice, thank you! It is definitely not an easy method to test - this area of our code would really benefit a good refactoring.
I can see two approaches:
* go deeper down the mocking route: use your debugger to find out how the HttpRequest is read by the file parsing method (`upload.parseRequest(request)`) and update your mocking accordingly
* refactor the `retrieveContentFromPostRequest` by introducing a new method for the specific case of URL download (which could be worth doing for other cases: clipboard, file upload…). Your new method would directly take the supplied URL as parameter, and would be much easier to test.

Also, have you had a look at @tfmorris's suggestions above?
Keep up the great work!

main/tests/server/src/com/google/refine/importing/ImportingUtilitiesTests.java
```diff
@@ -82,4 +95,75 @@ private ObjectNode getNestedOptions(ImportingJob job, TreeImportingParserBase pa
...
+                    JSONUtilities.safePut(progress, "percent", percent);
+                }
+
+                @Override
+                public boolean isCanceled() {
+                    return job.canceled;
+                }
+            });
+        } catch (Exception exception) {
+            Assert.assertEquals(MESSAGE, exception.getMessage());
```
wetneb, line 151: If no exception is thrown, this test will pass because the Assert will not be executed at all. A better way to assert that exceptions are thrown is to remove the try/catch block and to add the exception class in the `@Test` annotation as `@Test(expectedExceptions = Exception.class)`

main/tests/server/src/com/google/refine/importing/ImportingUtilitiesTests.java
```diff
@@ -82,4 +95,75 @@ private ObjectNode getNestedOptions(ImportingJob job, TreeImportingParserBase pa
         JSONUtilities.safePut(options, "recordPath", path);
         return options;
     }
+
+    private static String RESPONSE_BODY = "{\"code\":401," + "\"message\":\"Unauthorised\"" + "}";
+    private final static String MESSAGE = String.format("HTTP error %d : %s | %s", 401,
+            "Client Error", RESPONSE_BODY);
```
wetneb, line 101: Since both of these variables are only used in the test below, I would just define them there. If they really need to be declared as static attributes, then they should go at the beginning of the test class. Also I would remove the `+` concatenation operators and just declare `RESPONSE_BODY` as a single string.

main/tests/server/src/com/google/refine/importing/ImportingUtilitiesTests.java
wetneb: Or if you want to check the message in the exception, make sure you add an `Assert.fail("No exception was thrown")` at the end of the try block. I would also make the try block narrower (it should just surround the call that is going to throw the exception). If there are any other methods throwing exceptions in the test you can add `throws Exception` to the method signature so that you do not have to catch them explicitly.

wetneb: @urvashigupta7 any thoughts about this?

urvashigupta7: Umm.. I think checking exception message would be better Because,  if we write **expectedExceptions = Exception.class **, the test will pass even if gets failed due to any other line which raises exception in retrieveContentFromPostRequest method.

wetneb: Ah sorry I had not seen that you did add the Assert.fail and made the try block narrower. Thank you!

wetneb:
Congratulations and thank you for this PR, welcome to the team!

urvashigupta7:
Thank you @wetneb 😄

tfmorris:
Good work! That's a challenging test for your first PR.

urvashigupta7:
Thank you so much! @tfmorris

PR #2872, opened 2020-07-03
Fix race condition in service manifest fetching

### LINKED ISSUE #2803: Reconciled columns not being draggable to item inputs in Wikidata schema
**Describe the bug**

In some situations, reconciled columns cannot be dragged and dropped to inputs that expect Wikidata entities.

**To Reproduce**
The exact steps are not clear to me, but this can be observed in the following screencast at 40:00:
https://www.twitch.tv/videos/659297903

**Desktop:**
 - OS: Windows
 - Browser Version: Firefox
 - JRE or JDK Version: 

**OpenRefine :**
 - Version 3.3

wetneb:
Fixes #2803

### DIFF
--- a/extensions/wikidata/module/scripts/dialogs/schema-alignment-dialog.js
+++ b/extensions/wikidata/module/scripts/dialogs/schema-alignment-dialog.js
@@ -127,10 +127,7 @@ SchemaAlignmentDialog.setUpTabs = function() {
 
   // Init the column area
   this.updateColumns();
-
-  var url = ReconciliationManager.ensureDefaultServicePresent();
-  SchemaAlignmentDialog._reconService = ReconciliationManager.getServiceFromUrl(url);
-
+ 
   /**
    * Init the issues tab
    */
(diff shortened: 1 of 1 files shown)


extensions/wikidata/module/scripts/dialogs/schema-alignment-dialog.js
```diff
@@ -148,10 +145,16 @@ SchemaAlignmentDialog.setUpTabs = function() {
 
   this._previewPanes = $(".schema-alignment-dialog-preview");
 
-  // Load the existing schema
-  this._reset(theProject.overlayModels.wikibaseSchema);
-  // Perform initial preview
-  this.preview();
+  var url = ReconciliationManager.ensureDefaultServicePresent();
```
tfmorris, line 148: This is depending on Wikidata being the default reconciliation service which is true today (and this patch doesn't change this aspect of the behavior), but it seems like the Wikidata extension should know the address of its own reconciliation service.

Also, using the ensureDefaultServicePresent function to fetch this seems rather roundabout.

wetneb:
Agreed. With this new version, the service manifest is fetched only once, at least in the situations I tested.

wetneb:
Merging this soon to get 3.4 released unless there are any concerns.

PR #2882, opened 2020-07-05
MockConstraintFetcher cleanup part-I

darecoder1999:
Removed its use from different test files.


extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/ConflictsWithScrutinizerTest.java
```diff
@@ -27,12 +26,12 @@
 
 public class ConflictsWithScrutinizerTest extends ScrutinizerTest {
 
-    public static PropertyIdValue conflictsWithPid = MockConstraintFetcher.conflictsWithPid;
-    public static Value conflictsWithValue = MockConstraintFetcher.conflictsWithStatementValue;
-    public static PropertyIdValue propertyWithConflictsPid1 = MockConstraintFetcher.conflictingStatement1Pid;
-    public static Value conflictingValue1  = MockConstraintFetcher.conflictingStatement1Value;
-    public static PropertyIdValue propertyWithConflictsPid2 = MockConstraintFetcher.conflictingStatement2Pid;
-    public static Value conflictingValue2  = MockConstraintFetcher.conflictingStatement2Value;
+    public static PropertyIdValue conflictsWithPid = Datamodel.makeWikidataPropertyIdValue("P2002");
```
wetneb, line 29: Why not use the constant values defined in the corresponding scrutinizer instead of hardcoding the pids in yet another location?

darecoder1999: I can reuse the QIDs and parameter PIDs but rest these properties need to be hardcoded here I guess.

wetneb: Each constraint type should be hard-coded but only in one locations: all classes which use it should import it from there.

darecoder1999: So, I have done that for these two classes. Should I do the changes for all test classes in this PR or open a new PR for these changes?

wetneb: As you prefer!

darecoder1999: So, I'll create another PR for that with an appropriate branch name and commit message to be clear. :) 
What else needed to improve in this one?

Issue #2861, opened 2020-07-02
Stop testing for Oracle JDK in CI


wetneb:
The differences between Oracle JDK and OpenJDK are minor starting from version 11. It is probably wasteful to be running both of them in Travis.

More background on the differences: https://www.baeldung.com/oracle-jdk-vs-openjdk

Over the past few years I do not recall encountering any bug in OpenRefine which would only be present with one of OracleJDK and OpenJDK. Running many builds at the same time increases the chances that we run into spurious build failures and can increase the build time if the builds are not run concurrently. So I propose we remove oraclejdk11 from Travis.

tfmorris:
I agree it doesn't make sense to test multiple JDK 11s. We currently test oraclejdk11 and openjdk 8, 11, and 12. According to that article the Oracle JDK LTS releases are 8 and 11 and the openjdk policy is to drop support of the previous release when a new release comes out. Oracle JDK 8 LTS support officially ended in January, but commercial customers can buy support through at least December 2012.

Based on that, it seems like a better mix might be:
- oraclejdk8 (presuming that we can get it on Travis CI)
- openjdk 11
- openjdk 12

Certainly there doesn't seem to be any value in testing openjdk8 since it's many versions out of support -- and even the oraclejdk8 is on its last legs.

wetneb:
I think setting up oraclejdk8 is harder than openjdk8 - but I would be happy to have the list you propose in principle.

tfmorris:
I was able to find an oraclejd8 on trusty and got rid of the duplicate JDK 11 build, but then I realized that we didn't have JDK 13 or 14 and also didn't have an OS X build, so I ended up adding more builds than I removed. I think it'd be worth adding Early Access Program (ie JDK 16) builds, but jacoco doesn't support it.

You can see an example run here: https://travis-ci.org/github/OpenRefine/OpenRefine/builds/704530492

It currently has 6 jobs, which are, in order:
- openjdk14, bionic
- oraclejdk11, bionic
- bundled java, os x
- oraclejdk 8, trusty
- openjdk 13, bionic
- openjdk 12, xenial

If we can pare it to 5 jobs that will shave a couple of minutes off the build (6.5 min vs 9 min). Dropping the OS X build would get it under 5 minutes.

Having set all this up though, I'm wondering how necessary it is, particularly if we bundle a JRE on all platforms. Perhaps all we need is the Java version that we build our distributions on and future versions (so that we can spot coming compatibility issues).

[...]

PR #2894, opened 2020-07-06
Tweak CI build settings - refs #2861

tfmorris:
- On Mac, sleep for a bit after database server startup before trying to connect
- Use default JDKs on each build platform to avoid having to do Java installs (Xenial Oracle JDK 8, Bionic OpenJDK 11, OS X Oracle JDK 14)
- Make jobs which require installations optional so they don't fail the build (OpenJDK 12 & 13)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ jobs:
         - brew services start mysql
         - brew services start postgresql
         - brew services start mariadb@10.3
-        #- sleep 15 # wait for databases to start up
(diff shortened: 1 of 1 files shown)


tfmorris:
Hopefully these tweaks will stabilize things without investing in moving to an entirely different system.

tfmorris:
@wetneb Do you want me to explicitly request your review on PRs? I tend not to request specific reviewers, but I'm happy to do that if it helps your workflow.

Issue #650, opened 2012-12-17
Extend FingerprintKeyer to decompose ligatures


frodeseverin:
This issue might be related to #409 

fingerprint("Ære") returns "ære"
fingerprint("Øre") returns "ore"
fingerprint("Åre") returns "are"

"Æ" is only lower-cased while "Ø" and "Å" are also exchanged with a different letter. The handling of Æ should be fixed.

I use UTF-8 encoding.

;)Frode

thadguidry:
ICU4J could be used to provide normalize() and transliterate() commands The most recent bikeshedding discussion is here on our mailing list:  https://groups.google.com/d/topic/openrefine/6hY1-GI0KAI/discussion

tfmorris:
The switch to standard text normalization (#2898) has partially addressed this in that "æ" is decomposed into "ae" and "Ø" into "oe", but because "å" is formed with a diacritic, the diacritic stripping process prevents us from seeing "å" to do anything special with it.

We could special case this one letter and handle it before doing the diacritic stripping. Since the vast majority of uses are in Scandinavian languages where the accepted fallback is "aa", that might be the correct thing to do.

Of course these characters can be replaced by hand using the GREL `replace()` function.

tfmorris:
> if Java's text Normalizer is your preferred choice for the normalization, and it proves incapable of performing normalization in this manner, pragmatism wins, and the string 'æøå' normalizes to 'eoa', or whatever Java's Normalizer produces.

Our implementation of Java Normalizer in NFKD mode, followed by stripping of diacritics produces 'æøa' for this example and then we post-process the 'æ' to 'ae' and the 'ø' to 'oe'.

PR #2900, opened 2020-07-07
Use standard text normalization - fixes #2898

### LINKED ISSUE #409: Enhancement FingerprintKeyer - normalize German ß to ss
_Original author: bun...@gmx.net (June 14, 2011 08:43:17)_

Could you please extend the FingerprintKeyer with a normalization for the German Eszett &quot;ß&quot; to &quot;ss&quot;. This would help in common cases like Strasse = Straße.

The code would then look like

case '\u00DF':
  return 'ss';

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=409_

### LINKED ISSUE #2898: Modernize text normalization (ex-asciify())
OpenRefine currently implements its own homegrown diacritic folding in a method called `asciify()` which includes the comment:

>     * NOTE: this function deals only with latin-1 supplement and latin-1 extended code charts

In 2020, we need to do better. We are missing a large number of characters (see below for examples).  We also don't handle any of the combining diacritics and it can only do a 1-for-1 replacement, so can't handle expanding ligatures.

### Describe the solution you'd like
Normalizing to NFD or NFKD, removing all diacritics, and lower casing would probably be a good starting point and then we can consider adding some of the other things described in this recipe:https://stackoverflow.com/a/1453284/167425
(and requested in #409 and #650).

### Additional context

The list below represents just a **portion** of the "A"s that are excluded from the current list:

U+01CD | Ǎ | 461 | Latin Capital Letter A with caron
U+01CE | ǎ | 462 | Latin Small Letter A with caron
U+01DE | Ǟ | 478 | Latin Capital Letter A with diaeresis and macron | 0323 | for Sami
U+01DF | ǟ | 479 | Latin Small Letter A with diaeresis and macron | 0324
U+01E0 | Ǡ | 480 | Latin Capital Letter A with dot above and macron | 0325
U+01E1 | ǡ | 481 | Latin Small Letter A with dot above and macron
U+0200 | Ȁ | 512 | Latin Capital Letter A with double grave
U+0201 | ȁ | 513 | Latin Small Letter A with double grave
U+0202 | Ȃ | 514 | Latin Capital Letter A with inverted breve
U+0203 | ȃ | 515 | Latin Small Letter A with inverted breve
U+0226 | Ȧ | 550 | Latin Capital Letter A with dot above
U+0227 | ȧ | 551 | Latin Small Letter A with dot above
U+023A | Ⱥ | 570 | Latin Capital Letter A with stroke
U+0386 | Ά | 902 | Greek Capital Letter A with acute accent
U+0391 | Α | 913 | Greek Capital Letter Alpha
U+03B1 | α | 945 | Greek Small Letter Alpha
U+0410 | А | Cyrillic Capital Letter A

tfmorris:
Fixes #2898. Fixes #409. Refs #650

Replaces homegrown ISO Latin-1 only character subsitition with standard Java Normalize to NFD, followed by diacritic removal and a few custom character expansions/replacements.

### DIFF
--- a/main/src/com/google/refine/clustering/binning/FingerprintKeyer.java
+++ b/main/src/com/google/refine/clustering/binning/FingerprintKeyer.java
@@ -51,24 +51,28 @@ public class FingerprintKeyer extends Keyer {
     public static final Pattern DIACRITICS_AND_FRIENDS = Pattern
             .compile("[\\p{InCombiningDiacriticalMarks}\\p{IsLm}\\p{IsSk}]+");
 
+    // First part of table based on https://stackoverflow.com/a/1453284/167425 by Andreas Petersson
     private static final ImmutableMap<String, String> NONDIACRITICS = ImmutableMap.<String, String>builder()
             //Replace non-diacritics as their equivalent characters
-            .put("\u0141", "l") // BiaLystock
-            .put("\u0142", "l") // Bialystock
             .put("ß", "ss")
             .put("æ", "ae")
-            .put("ø", "oe") // oe or just o?
-            .put("å", "aa")
-            .put("©", "c")
-            .put("\u00D0", "d") // Icelandic eth All Ð ð from http://de.wikipedia.org/wiki/%C3%90
-            .put("\u00F0", "d")
-            .put("\u0110", "d")
-            .put("\u0111", "d")
-            .put("\u0189", "d")
-            .put("\u0256", "d")
-            .put("\u00DE", "th") // Icelandic thorn Þ
-            .put("\u00FE", "th") // thorn þ
-            .build();
+            .put("ø", "oe")
+            .put("å", "aa") // TODO: We'll never see this after decomposition
+            .put("©", "c") // copyright character
+            .put("\u00F0", "d") // Small letter Icelandic eth
+            .put("\u0111", "d") // Small letter D with stroke
+            .put("\u0256", "d") // Small letter African D
+            .put("\u00FE", "th") // Lower case Icelandic thorn þ
+            // Visually similar replacements from our private former asciify() method
+            // (only need lower case forms since we're already downcased)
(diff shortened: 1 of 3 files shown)


main/src/com/google/refine/clustering/binning/FingerprintKeyer.java
```diff
@@ -69,222 +96,32 @@ public String key(String s, Object... o) {
...
-        }
-        return b.toString();
+        return normalize(s);
     }
-    
-    /**
-     * Translate the given unicode char in the closest ASCII representation
-     * NOTE: this function deals only with latin-1 supplement and latin-1 extended code charts
-     */
-    private char translate(char c) {
```
wetneb, line 85: Random thought: to make sure we are not missing any of these cases by expressing the conversion with your regular expression, would it make sense to just transform this deleted code into a unit test which would check that all the behaviour is preserved for all the cases listed here?

tfmorris: Good idea. I did that and updated the normalization with some substitutions for visually similar characters which are not normally considered the same. I also switched to the more powerful compatibility decomposition (NFKD) from the canonical decomposition (NFD) which is a bigger hammer, but this is intended to be a relatively blunt instrument. When we expose normalization as a function, we can allow more fine grained control.

The current code should handle all characters in the old table the same, except for O with stroke which is now "oe" instead of "o", and provide equivalent semantics for all characters and ligatures which were not included in the old table.

We need to decide (eventually) whether we want to continue the practice of aggressively folding Latin characters to their nearest ASCII equivalent even if the Java normalization code doesn't. The Latin Extended B and Latin Extended C characters (and probably others) would require us to do our own mappings. Latin Basic, Latin-1 Supplement, Latin Extended A and Latin Extended Additional are fully covered.

PR #2901, opened 2020-07-07
Add ItemRequires Constraint

darecoder1999:
Implemented Item requires constraint as part of #2354


darecoder1999:
@wetneb: Can you help me understand why the Travis checks are failing? :(

wetneb:
@darecoder1999 sorry about this, we have had a few issues with Travis lately - it's definitely not related to your PR. We are working on fixing this.

tfmorris:
#2894 is my attempt at improving the build settings. If your CI build succeeds on at least one of the platforms, it's probably safe to assume that it's Travis' fault, not yours. You can restart the build jobs individually if you want to get everything green, but I wouldn't worry too much about it.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ItemRequiresScrutinizer.java
```diff
@@ -0,0 +1,126 @@
...
+                    values = propertyIdValueValueMap.get(pid);
+                } else {
+                    values = new HashSet<>();
+                }
+                values.add(value);
+                propertyIdValueValueMap.put(pid, values);
+            }
+        }
+
+        if (update.isNew()) {
```
wetneb, line 66: Isn't it possible to simplify this code by pulling the `if (update.isNew())` inside, just when the `QAWarning` is generated? I think this only affects the type of warning you emit, so that should get rid of a lot of duplicated code.

extensions/wikidata/module/langs/translation-en.json
```diff
@@ -148,6 +148,10 @@
     "warnings-messages/multi-valued-property-is-required-for-new-item/body": "This property is expected to have more than one statement on each item but it has single statement, for instance on {example_entity}.",
     "warnings-messages/multi-valued-property-is-required-for-existing-item/title": "{property_entity} should have more than one statement on existing items.",
     "warnings-messages/multi-valued-property-is-required-for-existing-item/body": "This property is expected to have more than one statement on each item but it has single statement, for instance on {example_entity}. If the item already has statements with this property in Wikidata, then this warning can be ignored.",
+    "warnings-messages/item-requires-certain-other-statement/title": "{property_entity} requires statement with property {added_property_entity}.",
+    "warnings-messages/item-requires-certain-other-statement/body": "Add the required statement with {property_entity} such as item with property {example_entity}.",
```
wetneb, line 152: The other issue bodies do not tend to give instructions to the user, but rather to indicate in plain terms what the issue is (leaving it to them to decide what to do about it). I think it is quite important not to be too explicit about what the fix should be in the second warning below, since it can be ignored in certain cases.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ItemRequiresScrutinizer.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class ItemRequiresScrutinizer extends EditScrutinizer {
+
+    public static final String newItem_type = "item-requires-certain-other-statement";
```
wetneb, line 20: Let's stick to camelCase and not mix it with underscores.

extensions/wikidata/tests/src/org/openrefine/wikidata/qa/scrutinizers/ItemRequiresScrutinizerTest.java
```diff
@@ -0,0 +1,82 @@
...
+        Snak qualifierSnak2 = Datamodel.makeValueSnak(itemParameterPID, requiredValue);
+        List<SnakGroup> constraintQualifiers = makeSnakGroupList(qualifierSnak1, qualifierSnak2);
+        List<Statement> constraintDefinitions = constraintParameterStatementList(entityIdValue, constraintQualifiers);
+
+        ConstraintFetcher fetcher = mock(ConstraintFetcher.class);
+        when(fetcher.getConstraintsByType(propertyIdValue, ITEM_REQUIRES_CONSTRAINT_QID)).thenReturn(constraintDefinitions);
+        setFetcher(fetcher);
+
+        scrutinize(updateA);
+        assertWarningsRaised(ItemRequiresScrutinizer.existingItem_type);
```
wetneb, line 58: It would be good to have a test for constraint violations on new items, too.

extensions/wikidata/module/langs/translation-en.json
```diff
@@ -148,6 +148,10 @@
     "warnings-messages/multi-valued-property-is-required-for-new-item/body": "This property is expected to have more than one statement on each item but it has single statement, for instance on {example_entity}.",
     "warnings-messages/multi-valued-property-is-required-for-existing-item/title": "{property_entity} should have more than one statement on existing items.",
     "warnings-messages/multi-valued-property-is-required-for-existing-item/body": "This property is expected to have more than one statement on each item but it has single statement, for instance on {example_entity}. If the item already has statements with this property in Wikidata, then this warning can be ignored.",
+    "warnings-messages/item-requires-certain-other-statement/title": "{property_entity} requires statement with property {added_property_entity}.",
+    "warnings-messages/item-requires-certain-other-statement/body": "This property is expected to have another statement with property {added_property_entity} such as item with property {example_entity}.",
```
wetneb, line 152: Thanks! Now in the second part of your message you write `such as item with property {example_entity}` but I do not think `example_entity` is a property here - it is an example item where the conflict was found. So perhaps we could rephrase that as `such as on item {example_entity}`?

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ItemRequiresScrutinizer.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        for (PropertyIdValue propertyId : propertyIdValueValueMap.keySet()) {
+            List<Statement> constraintDefinitions = _fetcher.getConstraintsByType(propertyId, ITEM_REQUIRES_CONSTRAINT_QID);
+            for (Statement statement : constraintDefinitions) {
+                ItemRequiresConstraint constraint = new ItemRequiresConstraint(statement);
+                PropertyIdValue itemRequiresPid = constraint.itemRequiresPid;
+                List<Value> itemList = constraint.itemList;
+                if (!propertyIdValueValueMap.containsKey(itemRequiresPid)) {
+                    QAWarning issue = new QAWarning(update.isNew() ? newItemType : existingItemType, propertyId.getId() + itemRequiresPid.getId(), QAWarning.Severity.WARNING, 1);
```
wetneb, line 73: The severity should also be adapted depending on whether the the `update.isNew()` or not

darecoder1999: Sorry, missed by me.

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/ItemRequiresScrutinizer.java
```diff
@@ -62,27 +63,26 @@ public void scrutinize(ItemUpdate update) {
...
         }
 
-        for(PropertyIdValue propertyId : propertyIdValueValueMap.keySet()){
+        for (PropertyIdValue propertyId : propertyIdValueValueMap.keySet()) {
             List<Statement> constraintDefinitions = _fetcher.getConstraintsByType(propertyId, ITEM_REQUIRES_CONSTRAINT_QID);
             for (Statement statement : constraintDefinitions) {
                 ItemRequiresConstraint constraint = new ItemRequiresConstraint(statement);
                 PropertyIdValue itemRequiresPid = constraint.itemRequiresPid;
                 List<Value> itemList = constraint.itemList;
                 if (!propertyIdValueValueMap.containsKey(itemRequiresPid)) {
```
wetneb, line 72: If I am not wrong, the two branches of this `if { } else if {}` are identical, so you could deduplicate the code with an OR (`||`) in your first `if`. But because of that, I am wondering if the `else if` branch should not be different: we probably want to display a different message for such a violation. If property A requires property B with a list of allowed values, and we have a violation in the `else if` branch here, we should say that although there is indeed a statement with property B, it does not have one of the allowed values, hence the violation. That probably means having 4 different violation messages for this constraint in total!

darecoder1999: I'm not having a clear idea about this, how can we have 4 different violation messages?
Property A requires property B with a set of allowed values, then the violations will be: 
1. It doesn't have property B, then we don't have to check about the values
2. It does have property B but it doesn't have the required value.

Can you specific the other two?

wetneb: Sure:
1. It doesn't have property B, and we don't have to check about the values. And the item where the constraint is found already exists.
2. It does have property B but it doesn't have the required value. And the item where the constraint is found already exists.
3. It doesn't have property B, and we don't have to check about the values. And the item where the constraint is found is new.
4. It does have property B but it doesn't have the required value. And the item where the constraint is found is new.

PR #2908, opened 2020-07-08
Edits to Installing page

allanaaa:
Testing a wide variety of questionable formatting options.


docs/docs/manual/installing.md
```diff
@@ -7,53 +7,269 @@ sidebar_label: Installing
...
 
-*   Java 
-*   Browser
+
+#### Compatible Operating System
+
+OpenRefine is designed to work with Windows, Mac, and Linux operating systems. Our team releases packages for each. 
+
+
+:::tip
```
thadguidry, line 20: I think they are OK.  Perhaps 1 shade darker. Its not so bad with dark mode on to me. Although I am very used to them now.

docs/docs/manual/installing.md
```diff
@@ -7,53 +7,269 @@ sidebar_label: Installing
...
-*   Windows
-*   Mac
-*   Linux
+For extra security you can [export your existing OpenRefine projects](exporting.md#export-a-project). 
 
 
-### Install/Upgrade
+:::caution
 
+Take note of the [extensions](#installing-extensions) you have currently installed. They may not be compatible with the upgraded version of OpenRefine. Installations can be installed in two places, so be sure to check both your workspace directory and the existing installation directory.
```
thadguidry, line 100: Can you add the link for where-is-data-stored for the "workspace directory" in this tip?

docs/docs/manual/installing.md
allanaaa: They're better in dark mode than in light mode, definitely. But I think I will use them more sparingly than they are in this page right now.

docs/docs/manual/installing.md
allanaaa: Definitely!

allanaaa:
Some formatting things to note for discussion and feedback, now or in the future, if you're into that sort of thing:

- Tip box experiments:
-- the use of green "tip" boxes to highlight third-party/external contributions (or potential ones).
-- the use of yellow "caution" boxes to highlight potential problems. 
-- the use of the blue "info" box for troubleshooting instructions (this can probably just be yellow too). 
-- I really dislike the grey "note" box. I think I'll avoid that one. 
-- I wonder if these tip boxes are very accessible; I'll do some testing and see if we need any CSS changes according to the usual tools. 

- there is a Medium post linked (how to set JAVA_HOME on Linux) that I tried to archive in the Wayback Machine but got an error (perhaps based on Medium). It's in a block-quote now, which also looks terrible. I'll keep messing with it.
Normally I'd archive and provide archive links for anything we link to.  Just something I'll keep track of in case it becomes a regular issue.

- using the `code` formatting for:
-- file paths
-- variable names
-- commands
-- file formats (e.g. `.zip`)
-- pretty much everything right now
This will change as I get my style guide more rigourous. Some of these will end up being bold or italic or etc. 

- the Homebrew instructions haven't been moved into this document yet. There may be an argument for leaving the really extraneous stuff like that in separate pages, though I think we're generally against it.

wetneb:
All of the tip, info and other boxes look a bit too bright to me, they really attract the attention to the boxes away from the main text, which seems a bit curious to me. For instance, in the section about compatible operating systems, the first thing that jumps to my eyes is the comment about chromebooks - whereas I would expect to see Windows, Mac and Linux first.

We should have consistent casing for section titles: decide on either title case or normal case (only the first letter is capitalized). At the moment we have "Compatible Operating System" and "System requirements".

I would not quote the guide by Charini Nanayakkara about Ubuntu - we need to have our own docs in here. I think these instructions are outdated.

allanaaa:
My bad on the titles - everything will be sentence case (if there are no objections) but I do not always adhere to my own style guide during early drafting. 

Sounds good on writing our own guide!

wetneb:
Sentence case makes sense to me! Yes it was a bit premature to bring up that sort of issue.

For the Ubuntu guide let me know if you need more info (those sort of things are hard to figure out without having the same system at home!)

allanaaa:
No worries on Linux, I have a VM! I think I have an answer for Mac testing too, but it might be good to have a Max expert on hand to answer some miscellaneous questions.

ostephens:
I can probably answer questions about Mac if necessary (although I have to admit I run the Linux version on my Mac unless I'm specifically testing something to do with the Mac version)

allanaaa:
It's too bad that when I update the files you can't see my approximately one thousand unresolved comments in Google Drive. And I'm leaving images out for the time being. 

Note that in https://deploy-preview-2908--openrefine-docs.netlify.app/manual/installing/#increasing-memory-allocation I have some bolded numbers - this indicates they can and probably should change with further testing.

And yes, I really did test when it would break and found it's at roughly 1550MB in https://deploy-preview-2908--openrefine-docs.netlify.app/manual/installing/#for-openrefineexe - but that could vary from system to system I'm sure. More testing needed. 


@ostephens maybe you (or anyone) can tell me: in the info box at https://deploy-preview-2908--openrefine-docs.netlify.app/manual/installing/#install-via-disk-image
Are those instructions for error messages only relevant to installing from the disk image? Or are they valid for Homebrew installations as well? My guess is they are valid for both but this is where the information was originally located. Thanks!

tfmorris:
- The maximum heap size for a 32-bit JVM is going to vary by host operating system and a number of other factors. You can read a description here: https://stackoverflow.com/a/1434901/167425 Almost everyone should be running 64-bit JVMs and that's what we should recommend.
- I would not expect to see the Mac security warnings with the Homebrew install.
- I think we should try to make this much more streamlined and compact. 
  - One thing that could help with this is separating out the different operating systems. Users are typically only interested in one of them.
  - The extension information could also be split out.
  - "Where is data stored" doesn't seem to be an install topic
  - splitting upgrade from install could streamline things
  - any of the other caveats and corner cases which aren't applicable to the majority of the users can be moved out of line
- Mac installs include Java, so it is not a prerequisite. Windows users have the option of a kit with bundled Java.
- Any place that we discuss the three operating systems, we should do it in a consistent order. Currently the ordering varies. In terms of popularity, the order (as measured by downloads) is Windows, Mac, Linux.
- "vulnerability" should be "security vulnerability" for clarity

I agree with the comments about toning down the prominence of the boxes. Their number, size, and color all combine to make them too dominant.

allanaaa:
Thanks for the comments @tfmorris. I'm making note of the bigger ones for further discussion later; I think all your smaller points have been resolved. 

I'm ready to merge this as-is and continue working on the outstanding sections in other PRs.

wetneb:
I am a bit worried that some of the points raised here are lost track of - did you create issues for the points that you do not want to address in this PR directly?

It seems there are still parts that need changing in pretty important ways:
* We still link to the external guide to set up JAVA_HOME on Ubuntu, which should not be necessary
* We still have instructions to set up JAVA_HOME for Mac whereas this should not be needed either
* The instructions for Windows could be improved too: it should be much easier for users to use a Java installer which should not require them to set up JAVA_HOME manually

ostephens:
>One thing that could help with this is separating out the different operating systems. Users are typically only interested in one of them.

Docusaurus supports a tab system for this kind of situation so you can:
* Setup tabs for each OS
* Use the same tabs each time we split instructions by OS
* Keep the tabs in-sync across the page so that if the user selects "Linux" the first time they see such a tab selection, then they will see the Linux instructions each time we use the same Tab set across the page

I think this is a nice way of streamlining the instructions as Tom suggests while keeping everything together in a single page

ostephens:
>I agree with the comments about toning down the prominence of the boxes
First suggestion is to use the Notes style in many places that you've currently used the Info style - the Notes box is grey which is much less prominent than the Info blue, and I think "Notes" is more appropriate than "Info" for the information being put in these boxes.

If we need to we can tone down the appearance of these boxes by adding the following to our custom.css:

```
    --ifm-color-info
    --ifm-color-warning
    --ifm-color-danger
```

But I'm not sure we are there yet - I think if we use Note primarily then this will feel a little less overwhelming.

For this PR, I'd suggest reviewing the uses of "Info" and seeing which might be re-styled as "Note" and implementing the obvious ones, but create an issue to review the colour scheme for these boxes - we can discuss colours and prominence in that issue

allanaaa:
Yes, I have noted all the comments and I will open issues when I am ready to discuss them. The outstanding pieces on this page are also on my to-do list. Is there some need for my to-do list to be made public? 

Splitting the instructions by OS does create a fair amount of redundancy. @ostephens, can you link us to a Docusaurus tab example so we can see it? 

Earlier in this PR I noted the "note" box is more unreadable than the rest, and certainly uglier. Since we'll be changing the colours anyways there's no reason to change all the "info"s to "note"s now.

wetneb:
If you would like to postpone the discussion of the points raised above, the best way to do this is to just leave this PR open and come back here to continue the discussion and make further changes when you are ready. That should not prevent you from working on other pull requests in the meantime. Is there a specific reason why you want to get this merged immediately and address the comments afterwards?

It is easier to keep track of comments and tasks if they are public: people who comment here should have some assurance that their comments are taken into account. If some of these improvement requests are kicked down the road, we do not want them to be only on your personal todo list: any other contributor should be able to see them and pick them up.

Sorry if the customs of the place seem obscure - I for sure also have a tendency to keep track of some tasks only mentally, and that is a normal behaviour when you feel responsible for a certain project. But in my experience it tends to overload individuals and hinder collaboration.

allanaaa:
Only because we agreed on a workflow that includes merging lots of small
PRs! I suppose our discussions didn't cover the full variety of real-world
scenarios.

ostephens:
>Splitting the instructions by OS does create a fair amount of redundancy. @ostephens, can you link us to a Docusaurus tab example so we can see it?

I don't think it would create any redundancy using the Tab approach. The page I linked to has an example on it  - see https://v2.docusaurus.io/docs/markdown-features/#syncing-tab-choices

wetneb:
> Only because we agreed on a workflow that includes merging lots of small PRs!

Sure, PRs can be as small as needed, but I think it is a good practice to address PR comments in the same PR as much as possible. For instance you could have a PR for a single section of a page - in that case you would only get feedback about what you wrote in that PR, you would not be expected to write up the full page obviously! Perhaps that is something we could try if the feedback you got here was a bit too much?

allanaaa:
Oh, section-specific tabs within pages! That's brilliant, I love that.
If no one has any objections, I can incorporate that into this page today
or tomorrow.

ostephens:
>Earlier in this PR I noted the "note" box is more unreadable than the rest, and certainly uglier. Since we'll be changing the colours anyways there's no reason to change all the "info"s to "note"s now.

Sorry I missed that comment.
I disagree that the Note box is uglier - but I guess that's a matter of opinion
Checking accessibility of the style is definitely something for we should do - for me the least readable is the "Caution" (white text on orange background) 

The Note box is different to the others in that it is allocated the style "--ifm-color-grey" rather than "--imf-color-note" (which would be more consistent with the other boxes). However, we can also control what the colour of the box is via the custom.css.

We essentially have 4 levels of call out here, and I think we should use them all and fix styling issues as we need to

ostephens:
>you could have a PR for a single section of a page 
>that is something we could try if the feedback you got here was a bit too much?

My opinion is that the discussion on this PR is very wide ranging and it would be better to get a baseline commit done for the page before we pick apart specific details - otherwise we are going to bogged down in lots of detail before we get anywhere

At this stage of a documentation project it is challenging to get the size of the PR right - I don't see a problem with getting the basics in place with this PR and having Issues to discuss the outstanding points of discussion/contention - I'd be in favour of this approach tbh.

Ultimately there is always going to be a lot of room for personal views in how we structure documentation - reaching consensus isn't going to be possible (e.g. @allanaaa and I disagreeing about the aesthetics of the Note callout). I think we have to have more lee way than we would with code in terms of getting PRs approved.

@allanaaa I don't think we need issues for everything on your to-do list but if we are going to get this PR merged then I do feel we need to have one or more issues to deal with outstanding feedback. In my view where there is feedback we should either:
* Accept it and make a change
* Reject it and say why (and if necessary - i.e. if clearly no consensus - create an Issue for followup discussion_
* Create an issue for anything you aren't going to do right now but plan to come back to (either for discussion or just later action)

These are just my views of course, but I think we need to find a way of working on the documentation that keeps us moving forward with reasonable speed and lets us iterate over the documentation gradually improving it rather than trying to do that all up front

allanaaa:
@ostephens it's actually five levels of call-out: https://v2.docusaurus.io/docs/markdown-features/#calloutsadmonitions
At this time I am using two: a "troubleshooting" (caution) box, and an "external/third-party/developers" (info) box. 
I don't yet have definitions/needs for all five. 
I don't mind their initial colour palettes, but I would desaturate everything, and convert text back into its natural colour (i.e. black text in light mode, white text in dark mode). I am unsure if this means a light-mode colour and a dark-mode colour for each box - needs testing. 

I've just discovered github email notifications cut off the comments posted on github, sorry I missed your link initially. No more emails. 

Would Wednesday mornings (my time) be a better time for our weekly calls? This flurry of activity feels like it would be a good time to have an IRL conversation if we're all available - yesterday's call felt like we were unprepared to discuss things at this level of detail. Just one anecdata though.

allanaaa:
I think since we now have a potential solution to the operating-systems-display question, the only outstanding discussion is "Install and upgrade: together or separate?" which is something I wanted to put aside and think a bit more on. 

I admit I am always going to prefer to start issues or discussion threads when I am ready to discuss them, not just because someone else brought it up. It would feel unprofessional to open up a discussion on something without having some factors to point out. And as you say, that's a quick way of burning myself out: trying to do too many things at once. So, sure, leave this PR open for longer?

allanaaa:
Although I guess that doesn't really work in principle: any old comment on a PR, whether relevant or irrelevant, would mean leaving that PR open indefinitely until that comment gets addressed somewhere. That presumes a lot of extra work on the part of the PR creator, doesn't it? I suppose you all have been at this longer than I have and can speak to whether this happens in reality.

wetneb:
It depends how important and consensual the comment is! It is common to have disagreements, and not all comments are meant to be "over my dead body". As Owen puts it, it is just necessary to respond on the PR to each point that is raised and decide what to do about it.

ostephens:
>any old comment on a PR, whether relevant or irrelevant, would mean leaving that PR open indefinitely until that comment gets addressed somewhere. That presumes a lot of extra work on the part of the PR creator, doesn't it? I suppose you all have been at this longer than I have and can speak to whether this happens in reality.

That's my concern here. It can happen but often with code the issues are more clear cut (or if it turns out the whole basis of the PR is faulty you want to go back to the drawing board).

With documentation we have to be careful to avoid this IMO

ostephens:
>I admit I am always going to prefer to start issues or discussion threads when I am ready to discuss them, not just because someone else brought it up.

I'd definitely encourage you to not feel this way - an issue can be a way of putting a pin in something that you either can't deal with straightaway or is worthy of wider discussion - just because you create it doesn't mean you have to have a solution or even a strong view on it.

Creating issues can be a way of preventing scope creep on a PR

allanaaa:
Am I not going to get a netlify preview for this? It's been half an hour.

--

I'm not clear where these two lines go:
>import Tabs from '@theme/Tabs';
>import TabItem from '@theme/TabItem';

from https://v2.docusaurus.io/docs/markdown-features/#tabs 

The only place they call it is a JS file in src/plugin that we don't have. 
https://github.com/facebook/docusaurus/search?p=1&q=import+tabs&unscoped_q=import+tabs

allanaaa:
.... Oh, I see. Does this mean we have to convert all our MD files to MDX files?

allanaaa:
I'm going to go back to writing and let someone else take a look at the code requirements.

ostephens:
You should be able to just put them at the top of the file after the metadata:
```
---
id: installing
title: Installing OpenRefine
sidebar_label: Installing
---
import Tabs from '@theme/Tabs';
import TabItem from '@theme/TabItem';
```

ostephens:
I can investigate if we can put them somewhere to work across all files, but definitely putting them at the top of the file works locally for me

Screen Shot 2020-07-15 at 19 57 38

allanaaa:
With or without changing the file to an .mdx?

ostephens:
> With or without changing the file to an .mdx?

Without

allanaaa:
That's enough floundering for me! If someone else can get it working on an example page for me, I would really appreciate it.

wetneb:
@allanaaa you seem to use line breaks very generously, and that is what was confusing Docusaurus. There are places where you have four or five line breaks in a row - is that intentional? Perhaps your editor pushes you to add these? I think you should never need to put more than two line breaks in a row (just to mark paragraph boundaries).

wetneb:
Also, to debug these sorts of things, it really helps if you can set up Docusaurus locally on your machine instead of having to rely on Netlify's preview.

ostephens:
>With or without changing the file to an .mdx?
While I answered "without" above, I can see the docusaurus documentation says:

>Note 1: While both .md and .mdx files are parsed using MDX, some of the syntax are treated slightly differently. For the most accurate parsing and better editor support, we recommend using the .mdx extension for files containing MDX syntax. 

So I should probably have used MDX extension for this file, although in this case just using the tabs it doesn't seem to make any difference

allanaaa:
No, the extra line breaks are "docs to markdown" artifacts that I generally don't bother removing unless it's a final-ish copy. I'm surprised *that's* what the problem was! I wondered if the tab JSX was breaking around markdown formatting, which would have been unfortunate.

thadguidry:
@allanaaa I would file an issue with Docusaurus  as this might be a slight parsing bug in 2.0 After all we are essentially testers for them.

wetneb:
I do not think this is a parsing issue on their side: line breaks have a meaning in Markdown, so I would not expect MDX to work with that syntax.

wetneb:
Sorry that I merged master into this branch, which pulled in unrelated changes. I will restore this now.

allanaaa:
I like the "persistent OS setting" but I hate the way the page-length jumps around if you move from one to another. I suppose it won't bother anyone but me though!

I'd also really like some sort of horizontal line to indicate the end of those OS-specific sections.

ostephens:
> hate the way the page-length jumps around if you move from one to another. I suppose it won't bother anyone but me though!

I dislike that as well, but it only affects you if you are switching between the different OS choices - which most people wouldn't be I think

>I'd also really like some sort of horizontal line to indicate the end of those OS-specific sections.

I think that's a very good idea - you could add those manually right now?

allanaaa:
Line breaks are up now. I like them!

Reading through the comments, the outstanding issues are:
- "We still have instructions to set up JAVA_HOME for Mac whereas this should not be needed either" (@wetneb)
- "splitting upgrade from install could streamline things" (@tfmorris)
- the "Advanced OpenRefine uses" section is not yet written, of course
- images are just quick screenshots for now, will be improved upon later
- finalizing the instructions we want to give for when a user should increase memory

I'm sure I'm missing a few as this PR commentary got quite long. If there are other issues to be created, can you point them out to me? I will make the issues.

PR #2909, opened 2020-07-08
Fix Excel date import - Fixes #1908

### LINKED ISSUE #1908: Excel dates imported as java.util.Date, not OpenRefine date
**Describe the bug**
Date columns from Excel files are imported as text into Refine.

**To Reproduce**
1. Create a new project using an Excel file with dates in it, e.g. the one attached below.
2. The configuration screen for project creation will show the date column in black. Create the project anyway to try to convert it later on.
3. The date is read correctly from the Excel file (December 11th, 2018), but displayed in black, not green, i.e. not recognised as date.
3. Apply Common Transforms > To Date to the "Date" column. The date is still stored as text. See screenshot below.

**Current Results**

**Expected behavior**

**Screenshots**
screenshot 2018-12-15 at 01 15 52

 - OS: macOS Mojave
 - Browser Version: Safari 12.02
 - JRE or JDK Version: 1.8.0_181

 - Version 3.1 [b90e413]

**Datasets**
Minimal file: [Date test.xlsx](https://github.com/OpenRefine/OpenRefine/files/2682052/Date.test.xlsx)

**Additional context**
This same file worked fine in Refine 2.8.

tfmorris:
Uses the OpenRefine 3.0+ date type of OffsetDateTime instead of the original java.util.Date.

Also added date tests which I was too lazy to figure out 8 1/2 years ago when I wrote the tests and added some utility methods to centralize date checking and conversion.

### DIFF
--- a/main/src/com/google/refine/expr/functions/ToDate.java
+++ b/main/src/com/google/refine/expr/functions/ToDate.java
@@ -39,6 +39,7 @@ import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
+import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.List;
@@ -201,5 +202,17 @@ public class ToDate implements Function {
     @Override
     public String getReturns() {
         return "date";
+    }
+
+    public static boolean isDate(Object o) {
+        return o instanceof OffsetDateTime;
+    }
+
+    public static OffsetDateTime toDate(Date date) {
+        return date.toInstant().atOffset(ZoneOffset.UTC);
+    }
+
+    public static OffsetDateTime toDate(Calendar date) {
+        return date.toInstant().atOffset(ZoneOffset.UTC);
     }
 }
--- a/main/src/com/google/refine/expr/functions/ToDate.java
+++ b/main/src/com/google/refine/expr/functions/ToDate.java
@@ -39,7 +39,6 @@ import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
-import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


main/src/com/google/refine/expr/functions/ToDate.java
```diff
@@ -202,4 +203,16 @@ public String getParams() {
...
+        return o instanceof OffsetDateTime;
+    }
+
+    public static OffsetDateTime toDate(Date date) {
+        return date.toInstant().atOffset(ZoneOffset.UTC);
+    }
+
+    public static OffsetDateTime toDate(Calendar date) {
+        return date.toInstant().atOffset(ZoneOffset.UTC);
+    }
```
wetneb, line 217: Should these things live in this class? This is the class for the GREL function, these methods look more like utility methods used throughout the code, so I would put them in a utility class.

tfmorris: The "right" way to do this would probably be to have an OpenRefine `Date` class which encapsulates all the date handling and then the GREL ToDate could call `com.google.refine...Date.parse()` but the date handling is such a mess I didn't want to risk a big refactor when my goal was to get this fixed for 3.4 once I learned you were still considering adding stuff to the release.

The point of putting the methods here was to keep them co-located with at least some of the date handling code, so people couldn't easily change one without the other. The other date handling utility methods are in ParsingUtilities which isn't a great fit either. 

I can create a separate tiny DateUtility class to hold these for the time being.

tfmorris: I decided having them colocated with the rest of the date handling was more important than the inappropriate name, so moved the to `ParsingUtilities`. Investigating the build failures now.

wetneb:
About having an OpenRefine-specific `Date` class: why not! For me the main issue here is the lack of static typing for cell values. We should have a closed list of data types that we can accept as cell and expression values, and enforce that with the compiler. If only Java had union types… Sadly it is pretty ugly to do it with visitor patterns, and there is probably a significant overhead in wrapping values in new classes.

tfmorris:
It is mainly the wrapping overhead that concerns me, but we might be able to finesse that by subclassing `Cell` by data type rather than introducing wrapper types for cell values. Leveraging static typing would require implementing multiple `call()` methods for GREL functions that accept multiple argument types in a single parameter position.

wetneb:
>  Leveraging static typing would require implementing multiple call() methods for GREL functions that accept multiple argument types in a single parameter position.

Yes, in a whole lot of other places: operations, exporters… it would probably be easier to rewrite everything in Scala at this rate!

Issue #1658, opened 2018-06-19
Frustrating UI for the selection of reconciliation services.


wetneb:
#### Steps followed to create the issue:

Select a reconciliation service - the service panel sometimes comes back
https://twitter.com/ruthbrarian/status/1008786606495584257
The UI should be more intuitive there.

#### Version of OpenRefine used:
3.0

tfmorris:
That tweet no longer exists. Is there any more information available on when this happens (a particular reconciliation service?) or why?

thadguidry:
@tfmorris I sent her an email just now:
>Hi Ruth,
>
>We are curious if you still have problems with the Reconciliation Service panel in OpenRefine.We saw that you opened the issue here:https://github.com/OpenRefine/OpenRefine/issues/1658and would like for you to give us a bit more information if you recall if it was Wikidata Recon service or some other custom Recon service that caused the panel to "come back".
>
>Please add the steps into the issue if you recall them (2 years ago!) and we can take a look more deeply.  If the issue has been resolved, then just type in an issue comment that you no longer have this problem or that you are using a newer OpenRefine version that has fixed the issue for you.
>
>Thanks so much!

wetneb:
There you go:
![recon_dialog_ui_bug](https://user-images.githubusercontent.com/309908/84932443-95bd7c00-b0d4-11ea-817f-6af732576e4c.gif)

ruthtillman:
Checking my twitter archive, I found the tweet was "This box will. not. stay. shut. It's like an over-sugared toddler. Any tips? #openrefine ??????????????????? Whyyyy. (I click either services or the bookmark-type icon and it closes and then hey it's back again)"

the above demonstrates it. And I think in the interim I've switched versions and that might've solved it? But maybe not. However if you can still do the above... yeah, that's it. :-/

thadguidry:
@wetneb Is that still a problem in 3.4 version?

wetneb:
Yes it is still current.

tfmorris:
A related issue, which might actually be triggering this, is that the panel doesn't close after adding a new service, obscuring the rest of the panel. I didn't recognize the little tab as a "close" affordance, so ended up dismissing the entire dialog and then starting the reconciliation over again.

thadguidry:
[...]

PR #2915, opened 2020-07-10
fixes service panel toggling

### LINKED ISSUE #2913: Toggle on services dialog after adding standard services does not work
ezgif com-video-to-gif (1)

Before adding any standard service toggling service dialog works fine. But, after adding service it doesn't.

urvashigupta7:
fixes #2913 , #1658 

mee2-online-video-cuttercom-4

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
+++ b/main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
@@ -59,7 +59,16 @@ ReconDialog.prototype._createDialog = function() {
 
   this._level = DialogSystem.showDialog(dialog);
   this._populateDialog();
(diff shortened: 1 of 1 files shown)


urvashigupta7:
https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/scripts/reconciliation/recon-dialog.js#L137

The populateDialog() method gets called more than once, so the click event on this was getting registered multiple times.

urvashigupta7:
@wetneb Please review :smile:

antoine2711:
> @wetneb Please review 😄

If you want a review, just ask in the upper right:
image

You can see it after:
image

Regards,
   Antoine

PR #2923, opened 2020-07-12
Remove informational logging from tests that are passing

tfmorris:
No issue. This removes a bunch of logging messages left over from development by either changing them from `info` to `debug` level or removing them altogether.

I'd like to get rid of the stack traces from the expected exceptions too, but they're logged from the production code, not the test code in most cases.

The end goal is to have the tests be quiet when they're passing so that the offending problem is easily spotted when they're failing.


thadguidry:
Hi @tfmorris just as a general comment, I would love to keep or have more HTTP "debug" logging.  But I think that some of that can be done via Apache environment settings? Could you add that info to our wiki please? Or tell me here or give links and I and @allanaaa will add it.

PR #2924, opened 2020-07-12
Fix the text format guesser so it doesn't inappropriately guess WikiText

### LINKED ISSUE #2850: Allow XLS and XLSX files without a file extension to be detected as Excel instead of Wikitext
Currently, if a XLSX file has no file extension because of someone's preference or OS setup for saving files, then when importing the XLSX file, OpenRefine will not detect it as an Excel file but instead will choose Wikitext importer as the first option.

XLS and XLSX files should be detected as Excel for the importer regardless of their file extension.

Ensuring that files when saved include the .xls or .xlsx extension as well as renaming lots of XLSX files to include the file extension on their filename.

**Additional context**
Windows is an OS that heavily depends on file extensions to navigate and perform actions based on them.  However, when a filename has no file extension, OpenRefine should go through its sets of algorithms to detect the most appropriate importer based on the content of the file.

Unfortunately, there is not file signature (magic bytes) for .xls and .xlsx files.  Although .xlsx files do start with the same magic bytes as .zip files https://en.wikipedia.org/wiki/List_of_file_signatures
So other detection mechanisms will have to be used such as Apache POI.

**Test file**
Extract file and then try to open `Colorado-Municipalities-XLSX` as well as rename it to just `ColoradoMunicipalities` and notice that OpenRefine will detect as Wikitext and not Excel.

[Colorado-Municipalities-XLSX.zip](https://github.com/OpenRefine/OpenRefine/files/4854060/Colorado-Municipalities-XLSX.zip)

tfmorris:
Fixes #2850
- Add simple magic detector for zip & gzip files to keep it from attempting to guess binary files
- Add a counter for C0 controls for the same reason
- Tighten wikitable counters to require marker at beginning of the line, per the specification
- Refactor to use Apache Commons instead of private   counting methods
- Add tests for most TextGuesser formats


tfmorris:
It would require 64KB of binary data with no LF bytes (and there's already an explicit check for Zip and gzip files), which seems a pretty remote possibility, but I've added a BoundedInputStream to the pipeline.

PR #2925, opened 2020-07-13
Add Citation needed Scrutinizer

darecoder1999:
Implemented Citation needed Constraint as part of #2354

### DIFF
--- a/extensions/wikidata/module/langs/translation-en.json
+++ b/extensions/wikidata/module/langs/translation-en.json
@@ -158,6 +158,8 @@
     "warnings-messages/existing-item-requires-certain-other-statement/body": "This property is expected to have another statement with property {added_property_entity} such as on item {example_entity}. If the item already has statements with property {added_property_entity} in Wikidata, then this warning can be ignored.",
     "warnings-messages/values-should-not-be-used-as-qualifier/title": "Invalid value for qualifier {qualifier_entity} on statement {statement_entity}",
     "warnings-messages/values-should-not-be-used-as-qualifier/body": "On items such as {example_entity}, an invalid value for qualifier {qualifier_entity} was supplied on statement {statement_entity}.",
+    "warnings-messages/no-references-provided/title": "{property_entity} requires at least one reference.",
+    "warnings-messages/no-references-provided/body": "The property {property_entity} specifically requires at least one reference for its statements.",
     "warnings-messages/ignored-qualifiers/title": "Some qualifiers were ignored.",
     "warnings-messages/ignored-qualifiers/body": "Qualifier values could not be parsed, so they will not be added to the corresponding statements.",
     "warnings-messages/ignored-references/title": "Some references were ignored.",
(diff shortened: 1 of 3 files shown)


wetneb:
I think we should just merge your new scrutinizer with the existing one, `UnsourcedScrutinizer`.

afkbrb:
Hi @darecoder1999 , thanks for your work on the qa, it does make it easier for me to work on the arbitrary Wikibase support.

One small question (not related to this PR): is it necessary to include the method [`findValues(List<SnakGroup> groups, String pid)`](https://github.com/OpenRefine/OpenRefine/blob/master/extensions/wikidata/src/org/openrefine/wikidata/qa/ConstraintFetcher.java#L61) in the fetcher interface? This method doesn't seem to be related to "fetching", I think it's better to implement it in [EditScrutinizer](https://github.com/OpenRefine/OpenRefine/blob/master/extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/EditScrutinizer.java). Of course, I also hope to hear @wetneb 's opinion on this.

wetneb:
Totally!

extensions/wikidata/src/org/openrefine/wikidata/qa/scrutinizers/UnsourcedScrutinizer.java
```diff
@@ -23,23 +23,44 @@
...
-            warning(type);
+    public void scrutinize(ItemUpdate update) {
+        for (Statement statement : update.getAddedStatements()) {
+            PropertyIdValue pid = statement.getClaim().getMainSnak().getPropertyId();
+            List<Statement> constraintDefinitions = _fetcher.getConstraintsByType(pid, CITATION_NEEDED_QID);
+            List<Reference> referenceList = statement.getReferences();
+
+            if (referenceList.isEmpty()) {
+                if (!constraintDefinitions.isEmpty()) {
+                    QAWarning issue = new QAWarning(update.isNew() ? newItemType : existingItemType, pid.getId(), update.isNew() ? QAWarning.Severity.CRITICAL : QAWarning.Severity.INFO, 1);
```
wetneb, line 56: Why should the severity be"critical" for new items and "info" otherwise? "critical" is really high, it will prevent people from uploading anything with the tool. I do not think we should take into account whether the item is new or not here: there is no point uploading statements if we are sure that they already exist and are sourced in Wikidata.

Currently, we have the "warning" level for all unsourced statements. When there is an explicit "citation needed" constraint, we could perhaps just raise the severity by one level ("important")?

extensions/wikidata/module/langs/translation-en.json
```diff
@@ -148,6 +148,10 @@
     "warnings-messages/multi-valued-property-is-required-for-new-item/body": "This property is expected to have more than one statement on each item but it has single statement, for instance on {example_entity}.",
     "warnings-messages/multi-valued-property-is-required-for-existing-item/title": "{property_entity} should have more than one statement on existing items.",
     "warnings-messages/multi-valued-property-is-required-for-existing-item/body": "This property is expected to have more than one statement on each item but it has single statement, for instance on {example_entity}. If the item already has statements with this property in Wikidata, then this warning can be ignored.",
+    "warnings-messages/no-references-found/title": "{property_entity} requires at least one reference.",
+    "warnings-messages/no-references-found/body": "Statements with property {property_entity} violating citation-needed constraint, it should have at least one reference",
```
wetneb, line 152: I think we can have only one issue type (whether the item is new or not should not matter, see below). The issue body is not very fluid in my opinion, how about this: `The property {property_entity} specifically requires at least one reference for its statements.`

Issue #1266, opened 2017-10-13
Request: Dialog box size change


richyvk:
I don't seem to be able to resize dialog box for doing transforms/facets. Using 2.7-rc.2

I note that another issue discusses css issues with resizing: https://github.com/OpenRefine/OpenRefine/issues/421

So, I'm not sure why I can't resize, unless that feature was removed?

Anyway, I'd like to request that the size of the dialogue box is increased, or ideally resizable. The text area is way too small when doing more complex transforms etc.

Would  PR to increase the standard size be accepted as an interim?

An expression text area height of 500px works well for me, and maybe an overall dialogue box width of 1000px, but that's less critical than the height of the textarea IMO.

wetneb:
It would be great to make the dialogs resizable!

thadguidry:
The Expression input box itself (inside the Transform dialog) is resizeable.

Agreed that Dialogs should be resizeable primarily so that users can see longer value strings previewed by stretching things out.

jackyq2015:
@richyvk PR is welcomed. just make sure you test it out under different browsers. Also you can do a screenshot after your change. :)

jackyq2015:
@thadguidry Are you sure The Expression input box itself (inside the Transform dialog) is resizeable? I tried but seems not.

richyvk:
Done a bit more digging. Expression box is resizable in everything but IE because it is a textarea, and by default there are resizable in none IE browsers, via the resize css property.

I think getting the whole dialog box to resize might be hard, certainly outside of my skills. But, it looks to me like adding css to the expression box to limit resize to vertical would at least offer a possible fix for this: https://github.com/OpenRefine/OpenRefine/issues/421

I will do a PR to do that.

Are there instructions on installing OR from the repo on a Mac? I currently have it installed via homebrew. How would I run it from the cloned repo so I can check my fix works?

thadguidry:
@richyvk Just FYI on the design we're looking towards and in our future plans are to have the expression box a dockable element that can be snapped to the left, right, bottom, etc (same for facets) and that can remain open and allow you to quickly click on a column and then move your mouse to that left, bottom, or right expression box and start transforming with Cloure, Python, GREL, etc.  Even optionally moving windows and panels into a separate browser tab or window if you have 2 or more screens.  We have big plans to make a lot of the edit menus disappear and instead move a lot of operations that you do in OpenRefine into more flow based work.  Think Photoshop for data :)

[...]

Issue #421, opened 2012-10-15
Resizing of Expression Box Causes CSS Overlapping


tfmorris:
_Original author: randy123...@gmail.com (July 19, 2011 19:41:23)_

<b>What steps will reproduce the problem?</b>
1. Open up a text transform on column
2. Resize expression box, so you're not working in a small textbox, I love this feature because working in predefined textbox sizes is so 1985.
3. 3 boxes seem to overlap across outer div, namingly, Preview Tab, the language selection dropdown, and Syntax error box (in grey) &quot;No Syntax error&quot;

<b>What is the expected output? What do you see instead?</b>

The outer div should scale with them or the boxes should be inside of the main div.

What version of Google Refine are you using? 2.1x64 (newest)

What operating system and browser are you using? Firefox - Win7/x64 Ubuntu/x64   

<b>Please provide any additional information below.</b>

Screenshot attached.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=421_

kushthedude:
@wetneb I think this issue has been addressed.

antoine2711:
> @wetneb I think this issue has been addressed.

@kushthedude: 
when I resize the input field, I get this:
![image](https://user-images.githubusercontent.com/12217525/79060922-834c3f80-7c58-11ea-8a73-814d2b4274a8.png)

I think this is what the issue is about.
Now, for me, the real question is: are you going to tackle this with Vue.js?! ;-)

I'm kidding, but I bet you are fierce enough!

Regards, Antoine

elebitzero:
This issue appears to be fixed. I can expand the expression and the dialog adjusts correctly.  I tested this on Brave, Chrome, Firefox, and Edge on Windows 10. This issue should be closed unless there are any objections. Thanks.

PR #2933, opened 2020-07-14
Update to jQuery 1.12.4 and jQuery Migrate 1.4.1 - fixes 2932

### LINKED ISSUE #2932: Update to jQuery 1.12.4 and jQuery Migrate 1.4.1
Our last jQuery update was in 2013 when we updated to jQuery 1.10 (#817). I think we're due for another update.

The first step in getting to jQuery 3.x, which was released in 2016, is to update to the latest 1.x and jQuery Migrate 1.4.1, then fixing all the jQuery Migrate errors/warnings.

The full process, documented [here](http://blog.jquery.com/2016/05/19/jquery-migrate-1-4-1-released-and-the-path-to-jquery-3-0/) is:

* Upgrade to the latest 1.x or 2.x version of jQuery, and the latest 1.x version of jQuery Migrate. (As of today that is jQuery 1.12.4, combined with jQuery Migrate 1.4.1.)
* Fix any problems identified in the Migrate 1.x warning messages you may see on the console.
* Remove the Migrate 1.x plugin and ensure the page still works properly without it loaded.
* Upgrade to the latest jQuery 3.x and latest jQuery Migrate 3.x.
* Fix any problems identified in the Migrate 3.x warning messages you may see on the console.
* Remove the Migrate 3.x plugin and ensure the page still works properly without it loaded.

tfmorris:
Fixes #2932. 

- Refactors module wirings to remove redundancy so I only have one place to udpate
- Updates jQuery to the latest 1.x and jQuery Migrate 1 to the latest.
- Updates Suggest widget to a slightly later version which doesn't use `browser.msie`
- Updates property getting/setting code for `checked`, `disabled`, `value`, `selected` properties to not use `.attr()`, `.removeAttr()` and other obsolete methods.

A cursory check of the various dialogs seems to indicate that that covers the bulk of the issues for this version of jQuery, but we should keep an eye on the Javascript console for a while to see if there are other migration issues to be addressed.


tfmorris:
I cleaned up a few more loose ends and took a pass through all of the dialogs in the UI without finding any new jQuery Migrate warnings, so I think this is as good as I'm going to get it. The jQuery Migrate plugin should protect us from any remaining issues, but it would be useful if developers could run with the Javascript console when possible to spot any additional warnings.

Next steps are:
- remove the jQuery Migrate plugin and make sure everything still works
- Update to jQuery 3.5 and jQuery Migrate 3.3 and start the whole process over again

thadguidry:
@tfmorris I got this one that popped up.
image
let me try to get the context...

thadguidry:
Clicking About link under Preferences , Help causes some problem
image

thadguidry:
This popped up when clicking on one of these in the Extract Operation History dialog:
* individual checkboxes
* Select All
* Unselect All

But cannot get it to trigger again...wtf?  Does it only trigger once for this Jquery Migrate logging? That would be non-helpful?
image

thadguidry:
@tfmorris maybe we should clarify... is a Jquery Migrate warning going to show up in yellow (!) icon in Firefox with the prefix `JQMIGRATE:` ? or does it appear differently? I'm wondering what an INFO versus a WARNING might look like in the context of JQMIGRATE since I want to be helpful here and not waste my time or yours.

thadguidry:
Columnize by Key/Value Columns
`JQMIGRATE: jQuery.fn.attr('selected') may use property instead of attribute`

image

thadguidry:
Facet by choice counts

thadguidry:
This one is weird... 
happened once I clicked on a Star in the History tab on Expression Editor, **I think** after clicking on Reuse links.
image

main/webapp/modules/core/externals/suggest/suggest-4_3a.js
```diff
@@ -724,11 +727,7 @@
       var css = this.options.css;
       var li = $("<li>").addClass(css.item);
       var label = $("<label>").text(data.name);
-      var div = $("<div>").addClass(css.item_name).append(label);
-      if(data.description) {
-          div.append($('<span></span>').text(data.description));
-      }
-      li.append(div);
```
wetneb, line 731: This removes support for descriptions in suggested items.

main/webapp/modules/core/externals/suggest/suggest-4_3a.js
```diff
@@ -1294,23 +1293,6 @@
...
-      if('view_url' in this.options && data.id) {
-        var view_url = this.options.view_url.replace('{{id}}', data.id).replace('${id}', data.id);
-        li.on('mousedown', function(e) {
-           if (e.which == 2) {
-              var win = window.open(view_url, '_blank');
-              win.focus();
-              e.preventDefault();
-           }
-        });
-      }
```
wetneb, line 1313: Those are also changes we made locally that I would keep.

main/webapp/modules/core/externals/suggest/suggest-4_3a.js
```diff
@@ -1508,7 +1490,7 @@
             xhr.getResponseHeader("X-Metaweb-TID"));
           }
         },
-        dataType: o.access === undefined ? "jsonp" : o.access,
```
wetneb, line 1511: This is also necessary for CORS support.

main/webapp/modules/core/externals/suggest/suggest-4_3a.js
```diff
@@ -1475,7 +1457,7 @@
 
       //this.flyoutpane.hide();
       var flyout_id = data.id;
-      var url = this.flyout_url.replace(/\$\{id\}/g, encodeURIComponent(data.id));
```
wetneb, line 1478: That looks like something worth keeping.

tfmorris:
@thadguidry I believe they will all have the `JQMigrate:` prefix with a Warning severity, followed by the traceback as Info severity messages. The traceback is very valuable for figuring out where the error occurs.

@wetneb I'll take a closer look at the Suggest plugin changes. Unfortunately, with Google Code shut down, the SVN history isn't available to extract a patch that I could apply. Perhaps I'll jut fix the msie.browser error and leave the rest alone.

main/webapp/modules/core/externals/suggest/suggest-4_3a.js
tfmorris: Isn't the description rendered in the preview pane (aka flyout)? Does this change cause it to show up twice?

wetneb: This is a description that is rendered in suggest items by default, without needing to hover the item. You are correct that it is redundant with the flyout service - I have introduced it a while ago as a replacement for flyouts because it gives a better user experience (having to hover one item at a time to find the right one is not really workable). Services can decide not to return descriptions and use flyouts, or the converse. (The Wikidata service currently uses both, but the flyouts should just be removed https://github.com/wetneb/openrefine-wikibase/issues/77)

tfmorris: I don't think I agree, but my intent was not remove local changes. It turns out that there were only a couple of changes in SVN, one of which was the one needed for jQuery compatibility, so I just integrated those by diffing against the base version before we applied our local changes.

tfmorris:
@thadguidry I reviewed all of those errors and, as far as I can tell, they all represent things that I've fixed already, except for the last one, which I can't reproduce and doesn't have enough context to be able to figure out what's going on (the entire traceback appears to be jQuery functions). Can you double check that you've got the latest code from the branch and perhaps do a `refine clean` then `refine build` to make sure that everything is up to date?

tfmorris:
> In itself, this change should not break extensions - it should just emit warnings when they are using deprecated APIs, right?

Correct. The Migrate plugin both acts as a shim to restore the deprecated behavior, as well as warning of its use, but only on the Javascript console and only if the non-minified developer version is used. Merging this PR should have no impact on extensions.

> But I wonder how the broader migration should fit with our release process. 
> So, do we need to have one stable release with this step of the migration only, in order to give time to extension maintainers to see the warnings this emits, before migrating to a newer major version of jQuery? Or do you see a way to get this done faster? 

This ties in to the discussion that I tried unsuccessfully to start on the dev list about what interfaces we keep stable and what guarantees we offer related to them. Are there no extensions which have yet repaired the JSON & CSRF breakage, I'd be tempted to just bundled everything together as one giant breaking change. Do we have an inventory of what extensions support what OpenRefine versions?

> I am not aware of a way for extensions to isolate their own jQuery version…

I think you can get namespace isolation using one of the module systems, but I'm not sure whether that would require OpenRefine to move to a module system first. 

> (On that note, if you want to take the responsibility of managing the release process for 3.5, I am not going to stand in the way! But no worries if not)

I'm happy to have that discussion when we know what the timing and content is.

thadguidry:
@tfmorris OK cleaned and built...testing on your branch... I'll start logging these down in comments starting now...(as you fix them with pushed commits, I'll keep pulling and testing/retesting)

When clicking on project tags this is thrown:
`Attribute selector with '#' must be quoted: #projectTags a[href=#records]`

image

thadguidry:
Clicking a Star on GREL History throws...

image

tfmorris:
>  I'll start logging these down in comments starting now...(as you fix them with pushed commits, I'll keep pulling and testing/retesting)

Let's just create new issues for them. Unless we find a functional problem, we should just merge this PR and fix the Migrate warnings later, as we find them. I already created an issue for your second issue yesterday #3013. 

Speaking of functional issues, though, is the expression dialog much wider than normal or is that my imagination?

thadguidry:
@tfmorris Ah, gotcha...you want to address additional stuff after we get this merged...sounds like a better plan and I agree, otherwise it gets confusing for me also!

Yes, Expression dialog is wider but I think only in certain contexts?  We'll figure that out later.

thadguidry:
@tfmorris Found the context on wider Expression Dialog ... it seems only on Edit Cells -> Transform ?
There's this: https://github.com/OpenRefine/OpenRefine/issues/1266
And this: https://github.com/OpenRefine/OpenRefine/issues/421
but maybe it was this?  https://github.com/OpenRefine/OpenRefine/commit/bd07fe0e1aacb2e1c12ecb2082b916f14bf24661

thadguidry:
@tfmorris Should `/benchmark/.factorypath` be added to `.gitignore` ?  I noticed it pulled in.

tfmorris:
@thadguidry Thanks. Good to know that the Expression dialog sizing is a preexisting bug (and yes, it was @tfmorris bd07fe0 for #1339 that broke it).

> Should /benchmark/.factorypath be added to `.gitignore`?

No idea. Doesn't appear in my working tree.

I pushed a fix for the "# must be quoted" bug.

wetneb:
> Are there no extensions which have yet repaired the JSON & CSRF breakage, I'd be tempted to just bundled everything together as one giant breaking change.

There is the rdf extension, which has been updated for Jackson and CSRF changes. I am not aware of others.

tfmorris:
@wetneb Is there anything left to do to get this merged? I think the release strategy discussion affects follow-on steps, not this PR.

PR #2935, opened 2020-07-14
Remove unused scrutinizer specific methods

darecoder1999:
After adding the code for fetching parameters for different scrutinizers in their corresponding classes, we should remove the redundant code from the ConstraintFetcher interface and WikidataConstraintFetcherClass.


darecoder1999:
@wetneb: There were no tests for getConstraintsByType, findValues, etc methods in WikidataConstraintFetcherTests. I'll raise another PR for adding tests regarding that. Would it be fine?

PR #2940, opened 2020-07-14
Added technical reference to docusaurus based on existing wiki pages

ostephens:
This PR is intended to update the navigation and structure and add the initial content for the technical reference.

The content has been migrated from the existing wiki pages pretty much "as is" and the intent of this PR is not to re-write content (which may need doing in addition) but to get the initial migration in place.

Comments on the structure and how this fits with the user reference manual are very welcome.
Comments on the detail of the content are less useful in relation to this PR as it's intention is to simply migrate content from the wiki into the new documentation site


allanaaa:
Seems fine to me.

wetneb:
@ostephens I just realized that this added an image `eclipse-debug-config.png` at the root of the repository - I think this should be in a subdirectory of `docs` instead, no?

ostephens:
Looks like I dropped a duplicate file in the wrong place by accident - I'll do a PR to fix. Apologies

PR #2941, opened 2020-07-15
Move findValues method to EditScrtuinzer

darecoder1999:
As findValues method is not concerned with fetching, therefore moving it to EditScrtuinizer as suggested by @afkbrb in his [comment](https://github.com/OpenRefine/OpenRefine/pull/2925#issuecomment-658496080)


tfmorris:
Looks good. Based on the multiple commits, it seems like perhaps you did the move by hand. If so, you should investigate whether your IDE has a Refactoring operation which would handle this. Many IDEs have a suite of refactoring operations which have gotten more and more powerful over the years and can save tons of time for large scale, mechanical edits like this.

darecoder1999:
@tfmorris: Basically, the extra commits were due to the tests failing in 1 or 2 cases after the refactoring. Sorry for that, I'll squash the commits next time on before requesting for review. :)

tfmorris:
Ok, got it. I guess I should have looked at the commits themselves. :)

PR #2950, opened 2020-07-16
Minor corrections to tech docs to test editing

thadguidry:


thadguidry:
@ostephens @allanaaa @wetneb So, the edit experience from the site itself is flawless.  One thing I did notice is that  the "edit" link is always on the bottom (not ideal when we have those long pages).  It would be nice to have it shown at both bottom and top of pages, or maybe always shown along some extra header we could add in the future?  But if that's too cumbersome to pull off then I guess we'll have to live with it being only along the bottom.  I wonder if others have asked about something like that in the Docusaurus community?  Thoughts?

Issue #1951, opened 2019-02-07
SQLite database connector


psychemedia:
The current database connector supports connections to PostgreSQL, MariaDB and MySQL.

A SQLite3 connector would be useful for several reasons:

- there is lots of tooling around SQLite, so it may be possible to use it as glue between different DBMS;
- SQLite3 provides a convenient way of distributing / sharing data in a way that also makes it amenable to querying.
- SQLIte3 is a file based system so it does not require a connection to a running server.

thadguidry:
@psychemedia Care to hack on this?  Just look at the existing connection code in the extensions database folder and have a look at the tests as well.  Ooo, also just noticed the README.md under /database needs to be touched up a bit also since we use Maven now (don't use Ant anymore)

psychemedia:
@thadguidry I don't have any experience with Java, and don't have a Java environment to work in, so there's a startup cost there that I don't have time to invest in atm... unless maybe you have a VM somewhere I could clone and fire up in an already working state?

tfmorris:
Does #2323 fully resolve this, or is additional documentation required?

wetneb:
Thanks, I forgot to link the PR to the issue.

snussik:
Sorry for posting in closed issue, but how i can connect to sqlite.db file that is locally stored. It seems that OpenRefine is waiting for some local port connection?
_3.4-beta [5597e1c]_

antoine2711:
> Sorry for posting in closed issue, but how i can connect to sqlite.db file that is locally stored. It seems that OpenRefine is waiting for some local port connection?

Hi. Just enter the full path to your file in the last field (`/Folder/File.sqlite3` in this picture):

![image](https://user-images.githubusercontent.com/12217525/82487542-7bc85380-9aac-11ea-9cca-50b4ef3bdad6.png)

@thadguidry: maybe we should consider « Only show the database login fields when necessary #2615 » as a bug fix for question like this one.

Regards,
   Antoine

snussik:
> Hi @snussik. Just enter the full path to your file in the last field, the one named « Database » (`/Folder/File.sqlite3` in this picture):

Thanks a lot! Didn’t think that it’s so simple. By the way, don’t know if it is a bug or feature but if I choose “SQLite” engine and  enter any string to database textfield, “test” button shows “Ok”.

antoine2711:
[...]

PR #2968, opened 2020-07-20
Update starting.md

allanaaa:
I am not at all clear on how well GitHub for Windows will handle multiple PRs pointing to the same work directory, but let's see what happens.

Let me know if this is formatted correctly - e.g., should it be building on base? Or on my other open PR?


allanaaa:
One thing I'd really like some instruction on is the section about connecting to a database:
https://deploy-preview-2968--openrefine-docs.netlify.app/manual/starting#database-sql

I see from discussions on #1951 and etc. that local .db file loading is quite simple. 
I didn't see instructions anywhere else about connecting to one online, so I tested and wrote those instructions from one of my own databases. Just because it worked for me doesn't mean it's the simplest way! So, if anyone with better experience here can offer some more details, I would appreciate it. 
Also I'm not sure how secure this method is - putting any request that comes from my IP, for example, on the "safe" list.

docs/docs/manual/starting.md
```diff
@@ -1,33 +1,23 @@
...
 title: Starting a project
-
 sidebar_label: Starting a project
-
 ---
 
-
 ## Overview
 
 OpenRefine doesn’t allow you to create a dataset from nothing - you can only start a project by importing in some existing data.
```
wetneb, line 9: Nitpicking: I would intuitively start by describing what OpenRefine allows you to do, rather than what it does not (to avoid the slight frustration)

docs/docs/manual/starting.md
```diff
@@ -52,25 +40,33 @@ From these sources, you can load any of the following file formats:
...
-
 #### Web Addresses (URLs)
 
-Type or paste the URL to the data file into the field provided. You can add as many fields as you want. OpenRefine will _____(something I haven’t quite figured out yet)______________________. 
+Type or paste the URL to the data file into the field provided. You can add as many fields as you want. OpenRefine will download the file and preview it for you. 
 
-Do not use this form to load a Google Sheet by its link; use the Google Data form instead. 
+If you supply two or more file URLs, OpenRefine will identify each one and ask you to choose which (or all) to load. 
 
+Do not use this form to load a Google Sheet by its link; use the Google Data form instead.
```
wetneb, line 69: It could be worth adding a link to the relevant section of the page.

docs/docs/manual/starting.md
```diff
@@ -80,24 +76,22 @@ This can be useful if you want to pre-select a specific number of rows from your
 
 This can also be useful if you would like to paste in a list of URLs, which you can use later to fetch the data online and build columns with. 
 
-
 #### Database (SQL)
 
 You may want to pull the latest dataset directly from an online database. This could include an online catalogue, a Wordpress or similar content management system, or a [digital repository or collection management system](https://bits.ashleyblewer.com/blog/2017/08/09/collection-management-system-collection/).
```
wetneb, line 81: It is not clear to me how the external link is relevant to SQL import, it seems to have a much broader scope. Because this functionality only works for SQL databases at the moment, I would keep this paragraph focused on this use case. Because SQL databases are rarely completely public I would not describe them as "online databases" - I would expect people to use this with a database internal to their organization, in general.

docs/docs/manual/starting.md
```diff
@@ -80,24 +76,22 @@ This can be useful if you want to pre-select a specific number of rows from your
...
 
 OpenRefine can connect to PostgreSQL, MySQL, MariaDB, and SQLite database systems. It will automatically populate the “Port” field based on which of these you choose, but you can manually edit this if needed.
 
-To import data directly from a database online, you will need to do two things:
-
+If you have a downloaded database (`.db`) file, you can supply the path to the file on your computer directly in the “Database” field at the bottom of the form. You can leave the rest of the fields blank.
```
wetneb, line 85: It should be clear that this is specific to SQLite: connecting to a SQLite database can only be done by supplying the path to such a file, and SQLite is the only database type where you provide the path to a file.

docs/docs/manual/starting.md
```diff
@@ -80,24 +76,22 @@ This can be useful if you want to pre-select a specific number of rows from your
...
+To import data directly from a database online, you will need to do two things:
 
 *   Add OpenRefine (running from your computer) to an account authorized to access your database
 *   Set up OpenRefine to access that database using that account 
 
 ![A screenshot of connecting to a database](img/databaseconnect.jpg "A screenshot of connecting to a database.")
 
-
-Log in to your hosting provider. Get the database type (such as MySQL), database name, and the URL (either an IP address, such as `127.0.0.1`, or the domain that uses the database, such as _https://openrefine.org/category/blog.html_). Then look at the accounts authorized for access. You may wish to create a new account just for OpenRefine, or add OpenRefine to an existing account. 
+Log in to your hosting provider. Get the database type (such as MySQL), database name, and the URL (either an IP address, such as `127.0.0.1`, or the domain that uses the database). Then look at the accounts authorized for access. You may wish to create a new account just for OpenRefine, or add OpenRefine to an existing account.
```
wetneb, line 94: The database host is not a URL by itself - it only contains the host part of a URL, so a host name or IP address.

docs/docs/manual/starting.md
allanaaa: I think it is worth sharing some definitions - users might have metadata editing access to online collections without knowing anything of the underlying architecture - but yes, this needs more elaboration.

ostephens: I think there is a difference here - the link points to a list of data sources, but they don't seem to (generally) offer access via SQL - so they won't be relevant here as far as I can tell

allanaaa:
In our chat today we talked about simplifying these database-access directions - presuming you have account access to a database, you can probably find your way around this part. 

Just now in searching our user group I found:
https://groups.google.com/g/openrefine/c/GsHYuP1kXV8/m/rSbpK6EpAgAJ

... which brings up two things:
1) One, are there important SQL limitations/options that should be noted in the documentation?
https://github.com/OpenRefine/OpenRefine/blob/master/extensions/database/module/scripts/index/database-source-ui.js 
excludes: DROP, TRUNCATE, DELETE, ROLLBACK, SHUTDOWN, INSERT, ALTER, UPDATE
which all seems fine to me, and not worth mentioning. 
I haven't written any SQL guidance because I assume nobody needs it, but is there anything particular to OpenRefine that we should point out?

2) The user also links to https://docs.rfam.org/en/latest/database.html 
which is a great public database that users can query. The page is pretty explanatory, and it even has example queries, which is great. Any interest in just pointing people to a resource like this? Not necessarily this one - I'll do some shopping around if there's interest.

wetneb:
1. I would not list the statements explicitly - you could just indicate that queries that modify the target database are not allowed (since the purpose of this importer is just to extract data from the database).

2. RFAM is a great example!

docs/docs/manual/starting.md
ostephens: This is marked as resolved but I'm not clear it has been resolved? It feels like we should be specific that it is a SQLite database file that is being referenced here - I don't think just saying `a downloaded database ('.db') file` is specific enough.

Generally I think resolutions need to be clear:
* Change made that directly addresses the comment - I think the change speaks for itself and is fine
* Decision to not make any changes, or change in a different way to what was suggested  - I think worth noting this is the action happening if it isn't going to be obvious to the reviewer
* No decision right now, and will come back for later consideration - again worth noting this action and I'd encourage an issue as a way of tracking although it may not always be necessary

docs/docs/manual/starting.md
```diff
@@ -80,24 +76,22 @@ This can be useful if you want to pre-select a specific number of rows from your
...
 *   Set up OpenRefine to access that database using that account 
 
 ![A screenshot of connecting to a database](img/databaseconnect.jpg "A screenshot of connecting to a database.")
 
-
-Log in to your hosting provider. Get the database type (such as MySQL), database name, and the URL (either an IP address, such as `127.0.0.1`, or the domain that uses the database, such as _https://openrefine.org/category/blog.html_). Then look at the accounts authorized for access. You may wish to create a new account just for OpenRefine, or add OpenRefine to an existing account. 
+Log in to your hosting provider. Get the database type (such as MySQL), database name, and the URL (either an IP address, such as `127.0.0.1`, or the domain that uses the database). Then look at the accounts authorized for access. You may wish to create a new account just for OpenRefine, or add OpenRefine to an existing account. 
 
 Each host will have a slightly different method, but generally speaking: look for “accounts with access” to the database you wish to authorize, and within the settings for that account, look for “allowable hosts” or “access hosts.”
```
ostephens: If this is what the line "Add OpenRefine (running from your computer) to an account authorized to access your database" above refers to it makes a bit more sense to me - but it feels like we are really into details that are outside the scope of OpenRefine here.

I think it's OK to say "Ensure the user you have specified has appropriate permissions to access and query the database" but any more than that feels like detailed troubleshooting (e.g. equivalent to telling the user that to upload a file they must have permission to access the file they are trying to upload - it's true, but not really an OpenRefine problem)

allanaaa: I updated the file to remove these sections: https://github.com/OpenRefine/OpenRefine/pull/2968/commits/4033b279f2c33ce66c46c647c28f1e66aadd55a4

docs/docs/manual/starting.md
```diff
@@ -80,24 +76,22 @@ This can be useful if you want to pre-select a specific number of rows from your
...
 
 OpenRefine can connect to PostgreSQL, MySQL, MariaDB, and SQLite database systems. It will automatically populate the “Port” field based on which of these you choose, but you can manually edit this if needed.
 
-To import data directly from a database online, you will need to do two things:
-
+If you have a downloaded database (`.db`) file, you can supply the path to the file on your computer directly in the “Database” field at the bottom of the form. You can leave the rest of the fields blank.
 
+To import data directly from a database online, you will need to do two things:
 
 *   Add OpenRefine (running from your computer) to an account authorized to access your database
```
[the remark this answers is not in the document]
allanaaa: I think I get where you're going with these comments, but there will be a lot of first-time users who won't understand any of this. So "special" is relative here I think.

What is "appropriate access"? I got errors until I set up the user account to expect calls from my IP; it wasn't just having a username and password (unlike accessing the public Rfam). Is there a way around that? Are there other methods entirely? 

If my experience is just a special case, that's great to know - but it's the only datapoint I have so far.  Ideally I can have some volunteers walk through this process (on PostgreSQL or MariaDB or SQLite) and report back.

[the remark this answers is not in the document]
ostephens: I agree that there will be a lot of users (first time or not) who won't know how to connect to a database using this process. I don't think your experience is a special case or that it is straightforward to do.

What I should have said is I don't think there are any requirements specific to OpenRefine here. The same issues would apply no matter what client software you were using to connect to the database.  It's probably clear that my view would be configuring the database software with an appropriate user is outside the scope of OR documentation - but I'm very happy for you to do what you think is best.

ostephens:
@allanaaa at the moment this PR includes some commits that are already in the master branch. I think if you rebase your docs-starting branch on master this problem should go away

docs/docs/manual/starting.md
```diff
@@ -80,53 +76,27 @@ This can be useful if you want to pre-select a specific number of rows from your
...
-
-``` error:Access denied for user 'yourusername'@'123-45-67-89.yourISP.com' ```
+If you have a downloaded database (`.db`) file from an SQLite database, you can supply the path to the file on your computer directly in the “Database” field at the bottom of the form. You can leave the rest of the fields blank.
 
-Take your IP address from this error message and put that, with periods instead of hyphens, into the “allowable hosts” field on the account you’re trying to use. Add a wildcard to the end of your IP address (“123.45.67.89%”). Save that setting, and then test the connection again with OpenRefine.
+To import data directly from a database online, you will need the database type (such as MySQL), database name, and the hostname (either an IP address, such as `127.0.0.1`, or the domain that uses the database) and the port on the host. You will need an account authorized for access, and you may need to add OpenRefine (connecting from your computer’s IP) to the “allowable hosts” for that account.
```
ostephens, line 89: Sorry I'm being picky but:
"the domain that uses the database" I think should be "or the domain that hosts the database"

And at the risk of re-treading previous discussion:
"you may need to add OpenRefine (connecting from your computer’s IP) to the “allowable hosts” for that account." 
I think would be more accurate to say:
"you may need to add the host or IP that OpenRefine is running on to the “allowable hosts” for that account." 

Even this I don't think is entirely accurate - because I think the allowable host will need to be the host/IP that the database perceives OpenRefine to be connecting from - which won't necessarily be the IP that is allocated to the computer running OpenRefine

docs/docs/manual/starting.md
```diff
@@ -157,7 +126,7 @@ Once OpenRefine is ready to import the data, you will see a screen with “Confi
...
 
-OpenRefine tries to guess how to parse your data based on the file extension. For example, .xml files are going to be parsed as though they are formatted in XML. An unknown file extension (or your clipboard copy-paste) is assumed to be either tab-separated or comma-separated. OpenRefine looks for a tab character; if one is found, it assumes you have imported tab-separated data.
+OpenRefine tries to guess how to parse your data based on the file extension. For example, `.xml` files are going to be parsed as though they are formatted in XML. An unknown file extension (or your clipboard copy-paste) is assumed to be either tab-separated or comma-separated. OpenRefine looks for a tab character; if one is found, it assumes you have imported tab-separated data.
 
 If Google isn’t certain what format you imported, it will provide a list of possibilities under “Parse data as” and some settings. You can specify a custom separator now, or split columns later on in the project interface.
```
ostephens, line 131: "If Google" should be "If OpenRefine"

docs/docs/manual/starting.md
ostephens: Sorry another picky point - "database online" in the first sentence should be "database" - the question of whether the database is online or not doesn't affect the connection (it could be running locally on the users computer)

thadguidry: @ostephens one way to express that is to say "NOTE: Firewall rules and your Database networking rules might require configuration to allow the egress IP of OpenRefine to connect to the ingress IP of your database"

allanaaa:
I'm not sure that that rebase worked - rebasing goes fine, but when I push changes it asks me to pull origin, and then all ... that ... appears. I'm supposed to be rebasing from upstream/master, right?

Capture

It might be worth having a conversation about more general "documentation scope" principles. I feel like we might have a disconnect there. 
The information I've left in the Database section truly feels useful to me, and not out of scope, so I'm going to hold the line for the time being. There will be more opportunities to revisit this of course. (It would be great if there was a generic guide we could link to here! I'm sure someone else has written it better, but I haven't seen it yet.) 
I think a small "user group" of first-timers (2-3) working through database connections would be useful since there seems to be so much attention on this. But I also read elsewhere (https://groups.google.com/g/openrefine-dev/c/Ywj7BxCX72A/m/Eq-U7idMBQAJ) that "only 40 folks will use it" - so it may not be worth this much scrutiny. (Do we keep stats on this?)

Thanks for the other feedback. Are there other comments on the other sections of this page for the time being? I was expecting some feedback on the "Project management" section, tbh.

allanaaa:
I've bumped the hierarchy a bit on this page so you can see all the "methods" in the TOC. It's better for readability I think. 
I also removed "Overview" and then immediately put it back again. For navigation's sake I guess we should always open with an H2.

ostephens:
Yeah - that rebase definitely doesn't look right :( So I'd expect the workflow to be something like:

* Fetch latest changes from upstream
* merge upstream/master into your local master
* rebase docs-starting on local master

With the command-line it would look like:

```
git checkout master
git fetch upstream
git merge upstream/master
git checkout docs-starting
git rebase master
```

It doesn't look like this is quite what has happened here - but I'm not entirely sure what's gone wrong tbh. Trying to rebase now I get a lot of conflicts which seem to be your own edits conflicting with themselves! Maybe we can try to work through this on the call tomorrow

ostephens:
>I think a small "user group" of first-timers (2-3) working through database connections would be useful since there seems to be so much attention on this. But I also read elsewhere (https://groups.google.com/g/openrefine-dev/c/Ywj7BxCX72A/m/Eq-U7idMBQAJ) that "only 40 folks will use it" - so it may not be worth this much scrutiny. (Do we keep stats on this?)

I absolutely feel we've spent more time on it than is worthwhile tbh (although that "40 users" is a figure pulled out of thin air as far as I can tell, but the general point which is this is of interest to a small minority of users is fair).

The only reason its a focus is because its hard to get all this information - which really is specialised - succinctly and usefully into this documentation which is meant to be about getting started with OpenRefine :(

I'm really against spending any more time on this, but if we are going to include information I feel it needs to be accurate - and the comments I've left since last week when I said "I'm very happy for you to do what you think is best" are about accuracy and not scope which I'm leaving to you.

ostephens:
>It might be worth having a conversation about more general "documentation scope" principles. I feel like we might have a disconnect there.

Very happy to do this on the call tomorrow - sorry if this is frustrating - I don't want to get in the way of progress here

ostephens:
> Are there other comments on the other sections of this page for the time being? I was expecting some feedback on the "Project management" section, tbh.

The Project Management section seems OK to me. The only comment I'd really have is that it will need linking up to other areas of the documentation once they are written I think as it seems to contain lots of terms that aren't explained - but I assumed you were going to deal with that as you worked through. (this might point to the need for a glossary for some basic terms - like row/record, facet, filter etc?)

I'm sure we could extend this more, but this is an area where I think the idea of getting some user feedback would be useful - I think what you've already written is a good start and we can work from there

ostephens:
>For navigation's sake I guess we should always open with an H2.
Agreed - I came to the same conclusion with the technical reference even though that sometimes meant an odd heading at the top of the page - but without it the navigation doesn't work so well (no way of jumping back to the top of the page via the righthand menu)

wetneb:
The issue with the rebase is likely due to the fact that you need to to "force push" just after rebasing - you should not pull from origin before that. We can try to fix this during the call today.

docs/docs/manual/starting.md
allanaaa: I wasn't crazy about "ingress/egress" :) but everything else here is incorporated now.

docs/docs/manual/starting.md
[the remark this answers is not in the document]
allanaaa: I did try the list in one draft but found the details uneven. This section is otherwise fixed for the time being, according to the Mayer Stephens Detente of July 2020. :)

wetneb:
@ostephens thanks! if you could use the "Squash" option next time it would be great - that avoids pulling all the intermediate commits into master.

ostephens:
Sure - apologies!

wetneb:
Not a problem at all!

PR #2969, opened 2020-07-20
Start accepting S3 URLs and read them to apply data quality scripts

RD1991:
* Developer needs to define environment variables - AWS_ACCESS_KEY_ID & AWS_SECRET_KEY
* S3 URL will accept one URL at a time to get data to the local system and then read it using configuration shared.


RD1991:
@tfmorris This is not an issue but new functionality. 
* Added new option read from S3 like we have a clipboard, local machine, databases. 
* At this time code can read one file at a time 
* Entire S3 path should be given to read file in the correct way. 

Let me know if there are any questions

wetneb:
Hi @RD1991, thank you so much for this pull request! It is really nice to expand the sources we can pull from.

Here are a few thoughts:
* in this state, it will be a bit hard for the user to discover how to supply AWS credentials. There is no mention of it in the UI. Even then, supplying these credentials as environment variables does not strike me as particularly user friendly. I would expect those fields to be exposed in the UI directly.
* It would be good to be clear on the URL formats that are expected here. Is it `s3://bucket/key` or the HTTPS equivalent?
* do we really need a specific importing panel for S3? Would it not be possible to add support for S3 in the existing URL importer? It could potentially prompt the user for S3 credentials if they are needed to retrieve the resource.
* it would be good to have unit tests for the changes to the backend

RD1991:
Hi @wetneb 

Here are a few comments - 
* in this state, it will be a bit hard for the user to discover how to supply AWS credentials. There is no mention of it in the UI. Even then, supplying these credentials as environment variables does not strike me as particularly user friendly. I would expect those fields to be exposed in the UI directly.
[RD]: Sure that can be done. The idea of not having that part of UI was, AWS key/secret doesn't change too often for a user. 

* It would be good to be clear on the URL formats that are expected here. Is it s3://bucket/key or the HTTPS equivalent?
[RD]: URL format should be "s3://..." 
 
* do we really need a specific importing panel for S3? Would it not be possible to add support for S3 in the existing URL importer? It could potentially prompt the user for S3 credentials if they are needed to retrieve the resource.
[RD]: 
- Though we can add them in the existing URL importer, handling the conditions on the backend side will be tough and code might get messy. 
- Also, s3 is an independent service as compared to URL imports so it should be handled separately is what I felt. 
- if we are working on URLs such as HTTPs equivalent then they will be automatically handled in existing URL importer, but to do it we need to make it public on s3 (private vs public in s3) which is not a safe option.  
Let me know your thoughts. 

* it would be good to have unit tests for the changes to the backend
[RD]: Sure, I will think on a few and add them.

tfmorris:
You can use `>` to quote text (or just select the "Quote Reply" menu option and it'll do it for you). I've edited the last reply to make the formatting a little easier to read.

* I think getting the credentials explicitly from the client is better. It will work in the case where the client and server are on different machines and the user doesn't have the ability to set environment variables. They are typically stored in ~/.aws/credentials, so we could ask the user to pick that file with the file selector, but I'm concerned about sending them in the clear to the server (since we don't use HTTPS).

* I think using the existing URL input field and checking for the s3: protocol prefix is the best approach to keep the UI uncluttered. The backend already has separate handling for http: & https: vs other protocols, so adding another conditional for s3: shouldn't be hard.

RD1991:
> You can use `>` to quote text (or just select the "Quote Reply" menu option and it'll do it for you). I've edited the last reply to make the formatting a little easier to read.
> 
> * I think getting the credentials explicitly from the client is better. It will work in the case where the client and server are on different machines and the user doesn't have the ability to set environment variables. They are typically stored in ~/.aws/credentials, so we could ask the user to pick that file with the file selector, but I'm concerned about sending them in the clear to the server (since we don't use HTTPS).
> * I think using the existing URL input field and checking for the s3: protocol prefix is the best approach to keep the UI uncluttered. The backend already has separate handling for http: & https: vs other protocols, so adding another conditional for s3: shouldn't be hard.

So we are agreeing on - let's update the existing URL input field with 2 more, AWS_CLIENT_KEY and AWS_SECRET_KEY and update backend code accordingly. I will update PR soon. 

Thanks.

RD1991:
@tfmorris I've made changes as per last time discussion. The only thing pending is "Test Case", I tried to write it however when I'm writing one it fails saying "not able to create directory ..." 

Do you have any idea? Otherwise, code is done, please have a look.

tfmorris:
> The only thing pending is "Test Case", I tried to write it however when I'm writing one it fails saying "not able to create directory ..."
> Do you have any idea? 

Can you provide more details on what is generating that error? Can you point to the code that you're stuck on?

> Otherwise, code is done, please have a look.

I'll review it.

tfmorris:
Hi @RD1991 Are you still working on this?

wetneb:
@tfmorris I would propose to close this per the lack of activity. This should normally be superseded by the new architecture which should let the user import from any source supported by HDFS (Hadoop Filesystem).

tfmorris:
Agree with @wetneb. @RD1991 please feel free to reopen (or create a new PR), if you'd like to continue this work.

PR #2976, opened 2020-07-22
Add a importer test for zip archives

tfmorris:
No issue. Adds a test for importing from a zip file. It also verifies that the archive file name and zip entry filename get saved correctly.


tfmorris:
#2978 depends on this, but is also based on it, so if it gets merged first, this will be a no-op. If this gets merged first, there should be no conflicts.

PR #2978, opened 2020-07-22
Refactor importer APIs - Fixes #2963

### LINKED ISSUE #2963: Refactor Importer APIs to support filename and archive name cleanly
In 818e139b430ecb6afe89c3b1c68176d0b4875eef the `NotImplementedException` throwing behavior of the methods in the base class got [removed](https://github.com/OpenRefine/OpenRefine/commit/818e139b430ecb6afe89c3b1c68176d0b4875eef#diff-972882453affea1078cdd88bbc4aef18L148-R151) so that they're now silent no-ops if someone calls them for a subclass where they are not implemented. This is unsafe and misleading behavior that just cost me a good chunk of time to track down.

It looks like TreeImportingParserBase suffers from the same issue: https://github.com/OpenRefine/OpenRefine/commit/818e139b430ecb6afe89c3b1c68176d0b4875eef#diff-fc818390bcd00535a45c7ef1f5dfc8d2L177-R178

### LINKED ISSUE #2981: Import options metadata duplicated
After importing a file, two copies of the options JSON are included in the project metadata. This bug was introduced with the change to support the saving of the archive file name.

### To Reproduce
1. Import a file
2. At the "Open Project" screen, click on "About" next to the file.
3. Look at the bottom of the screen for "Import option metadata (JSON)" 

### Current Results

There are two identical copies of the JSON data.

### Expected Behavior

Only a single copy of the data.

### Versions

 - OpenRefine 3.5 HEAD

tfmorris:
- Restores importer API compatibility for v3.5 which was broken by the archive file name feature (Fixes #2963)
- Rips out the weird control flow that was introduced for the feature that added importer options to the project metadata
- Centralizes all source filename and archive filename handling in a single place, separate from the individual importers
- Adds a test for import options being saved in the project metadata (building on test in #2976) 
- Fixes a couple of bugs in import options saving (Fixes #2981)


tfmorris:
We could create new method signatures and deprecate the one containing the filesource (ie filename) parameter, but:
- that parameter has been in the Importer API for 10 years, since it was first defined
- I'm not thrilled with the general structure of the API, so I'm not sure patching it is worthwhile, rather than designing a new API. I think something defined by Interfaces, rather than Base Classes, would be better, and the mechanism to differentiate between `Reader` & `InputStream` based importers is kind of weird.

I don't want to end up in a situation like the ODF toolkit where they deprecated the old API and then, before we could migrate to the recommended new API, they deprecated the new one as well and are coming up with a third.

We could definitely at least add a note to the documentation to say that subclasses don't necessarily need to do anything with the parameter. Documentation of the (implied) API for importers is pretty weak in general.

wetneb:
Fair enough, it makes sense to try to keep a stable Java API in 3.x and postpone all breaking changes to 4.x. I agree that the importer API is not ideal at the moment, it would be worth making something radically cleaner.

thadguidry:
So in summary @tfmorris and @wetneb we are saying to extension authors that we have decided to keep the API fairly stable in 3.x and we won't break that until 4.x ?  Just want to have clarity here for improving the technical documentation (and roadmaps).

wetneb:
That is the idea!

tfmorris:
I did a little investigation into removing the fileSource parameter and it looks like it's actually used by the ODS and Excel importers. I didn't dig in deeply to see how it is used or how hard it would be to remove the dependency.

I have a couple of other little cleanups to the refactor that I'll open a separate PR for.

@thadguidry I sent an email to the dev list concerning API commitments. Please provide feedback there.

PR #2982, opened 2020-07-22
Update running.md

allanaaa:
So far: starting/exiting, modifications, home screen, preferences page are mostly drafted. 
Soon: languages list, JVM preferences (tomorrow). 
Later: a nicely annotated & explained project screen, maybe some common extension buttons. 

This page has a fair amount of commentary in-text, in italics.
There is also some repetition (I haven't figured out where some details should be just yet).

I'm thinking about moving the memory allocation section from installing.md to this page (technically it's a modification), but I do think it is important enough to mention there. And moving it here would mean sandwiching it awkwardly somewhere inside "Starting with modifications."


allanaaa:
A small note: from https://github.com/OpenRefine/OpenRefine/wiki/History
I took the text at the very bottom:

> There is a limit, if the history is bigger than 100 MB (output file), google chrome will crash and you will not be able to extract the JSON. Firefox can handle around 200 MB(output file), be sure to disable the spellchecker.

OpenRefine generates JSON into a field that you can then copy to the clipboard. So I'm not clear what "output file" means. 100MB worth of text on your clipboard?

wetneb:
It also seems to me that you would need to do a lot of clicking and typing to get a history that big! I am not sure what to make of this comment. The point about the spellchecker is probably not relevant either - if the text area where we display the history JSON is spell-checked, we should disable that in the application directly (the user should not have to disable their spell checker globally just for that input!).

allanaaa:
The only thing outstanding on this page is the "Common extension buttons" section, which I'll probably write near the end.

Does anyone mind if I centre images? The images of the Facet/Filter and the History sidebars are long and skinny, which means they look a bit silly left-aligned. I still haven't looked into the CSS or any stylistic stuff in Docusaurus but I'm sure it's a quick fix.

thadguidry:
@allanaaa centering images should be fine.  Also let's think about sometimes not centering and combining images side by side when it makes sense. Trust you on that.

PR #2984, opened 2020-07-24
Importer refactoring cleanup

tfmorris:
Refs #2963. Followup to #2978. This tidies up a few loose ends from the importer refactoring:
- reverts the archive file name changes to ImportParameters and marks the entire class as deprecated since it was just introduced at the beginning of all this silliness
- removes unused parameters from `TabularImportingParserBase.readTable()` to make it easier to figure out what's used where and marks the old method signature as deprecated
- ditto for `XmlImportUtilities.importTreeData()` and some related methods in that class. Switch them back from using `ImportParameters` and having to guess which of its members they depended on to using specific parameters.
- removes duplicate fileSource and archiveFileName setting code from `XmlImportUtilities`. Likely not a functional problem, but a performance and code duplication issue.
- lowercased method name `XmlImporterUtiltilitiesStub.ProcessSubRecordWrapper()`
- doubled up tests for things that had new forms to test both the old and new forms

### DIFF
--- a/main/src/com/google/refine/importers/FixedWidthImporter.java
+++ b/main/src/com/google/refine/importers/FixedWidthImporter.java
@@ -132,7 +132,7 @@ public class FixedWidthImporter extends TabularImportingParserBase {
             }
         };
         
-        TabularImportingParserBase.readTable(project, metadata, job, dataReader, fileSource, limit, options, exceptions);
+        TabularImportingParserBase.readTable(project, job, dataReader, limit, options, exceptions);
     }
     
     /**
--- a/main/src/com/google/refine/importers/LineBasedImporter.java
+++ b/main/src/com/google/refine/importers/LineBasedImporter.java
@@ -129,6 +129,6 @@ public class LineBasedImporter extends TabularImportingParserBase {
             }
         };
         
-        TabularImportingParserBase.readTable(project, metadata, job, dataReader, fileSource, limit, options, exceptions);
+        TabularImportingParserBase.readTable(project, job, dataReader, limit, options, exceptions);
     }
 }
--- a/main/src/com/google/refine/importers/SeparatorBasedImporter.java
+++ b/main/src/com/google/refine/importers/SeparatorBasedImporter.java
@@ -157,7 +157,7 @@ public class SeparatorBasedImporter extends TabularImportingParserBase {
             }
         };
         
-        TabularImportingParserBase.readTable(project, metadata, job, dataReader, fileSource, limit, options, exceptions);
+        TabularImportingParserBase.readTable(project, job, dataReader, limit, options, exceptions);
     }
     
     static protected ArrayList<Object> getCells(String line, CSVParser parser, LineNumberReader lnReader)
--- a/main/src/com/google/refine/importers/TabularImportingParserBase.java
+++ b/main/src/com/google/refine/importers/TabularImportingParserBase.java
@@ -75,13 +75,36 @@ abstract public class TabularImportingParserBase extends ImportingParserBase {
     protected TabularImportingParserBase(boolean useInputStream) {
         super(useInputStream);
     }
+
+    /**
+     * @param project
+     * @param metadata
+     * @param job
+     * @param reader
+     * @param fileSource
+     * @param limit
+     * @param options
+     * @param exceptions
+     * @deprecated 2020-07-23 Use {@link TabularImportingParserBase#readTable(Project, ImportingJob, TableDataReader, int, ObjectNode, List)}
+     */
+    @Deprecated
+    static public void readTable(
+            Project project,
+            ProjectMetadata metadata,
+            ImportingJob job,
+            TableDataReader reader,
+            String fileSource,
+            int limit,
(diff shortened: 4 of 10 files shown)


wetneb:
Sorry for the delay in reviewing this. Can you explain a bit more why `ImportParameters` should not be used?

I think it would make sense to store importer configurations in classes (rather than `JsonObject` as we are doing now). Each importer could define (and register) its configuration class which would store all the parameters it relies on. This class could extend a base class storing generic configuration parameters used by other importers. Such a base class would look a lot like `ImportParameters` in my opinion, no?

For me the main defects of the current importer architecture are (on top of my head - I will expand on this once I have a proper dive to migrate them to the new architecture):
* The abuse of static methods: most of the importing logic is written in utility classes. This makes it harder to test and forces us to pass a lot of arguments to methods (hence the incentive to introduce classes to bundle them up). The control flow is a lot harder to follow, too.
* The explicit reliance on Jackson classes to store configuration options and other user input. These options should be parsed to POJOs early (similar to the migration we did for facet configurations). This helps detect invalid input early on, instead of failing deep down in the pipeline. With the current situation, we have to resort to boilerplate code like this:
https://github.com/OpenRefine/OpenRefine/blob/5639f1b2f17303b03026629d763dcb6fef98550b/main/src/com/google/refine/importing/ImportingJob.java#L150-L169

tfmorris:
> Can you explain a bit more why ImportParameters should not be used?

As currently implemented it obscures what parameters are needed where. If a utility method needs two parameters, I personally find it much clearer and easier to trace if those two arguments are passed in directly rather than as part of a "kitchen sink" object with 8 different fields. Perhaps refactoring everything would be another solution to this, as you suggest, but a short term stop gap is to back out the obfuscation.

Also, when I see

        super.processRecord(project, parser, rootColumnGroup, trimStrings, storeEmptyStrings, guessDataType);

replaced with 

        super.processRecord(project, parser, rootColumnGroup, 
                new ImportParameters(trimStrings, storeEmptyStrings, guessDataType));

which then requires internal references to be `params.trimStrings` instead of `trimStrings` that seems like a code smell (and silly). This value object isn't used to serialize values at all. It's just a lazy way of having few parameters.

I agree that exposing JSONObjects everywhere is a bad idea, but I'm not sure that Jackson annotated POJOs is much better since they also are tightly tied to our current JSON library. It seems like we could come up with a simple key/value map structure for the importer options which could be implemented generically using standard Java. Since importers have both Javascript & Java components to them, the least common denominator is actually the JSON string form.

Forcing importer writers to learn all about Jackson annotation and deserialization just so they can pass a couple of integers or booleans from the front end to the backend seems very unfriendly to me and I hope we can come up with a better scheme.

Are you also supportive of upper case method names and the other things that I tried to fix here?

tfmorris:
Here's the history of `ImportParameters`: https://github.com/OpenRefine/OpenRefine/commits/0aa3555669753a834174cd6e617fdaaf68106cfe/main/src/com/google/refine/importers/tree/ImportParameters.java
It got introduced as a side effect of "save the file location as a table column"

Some other things that I don't like about it include:
* it's limited to the tree importers (JSON & XML)
* it only includes some of the import parameters (no `recordPath`, etc)
* plus the other stuff that I mentioned above about obscuring data flow, etc.

wetneb:
> I agree that exposing JSONObjects everywhere is a bad idea, but I'm not sure that Jackson annotated POJOs is much better since they also are tightly tied to our current JSON library. It seems like we could come up with a simple key/value map structure for the importer options which could be implemented generically using standard Java.

That sounds like using `Properties` or another `Map<String,Object>` of the same kind - for me this has the same disadvantage as using the `JsonObject` we have at the moment, namely that we do not let the compiler help us keep track of which options are used by each importer, which types they have, and so on.

I know your reluctance to expose Jackson in the API and I am sympathetic to it: I am open to architectures which make this pluggable. But the fact that we do not have this yet should not encourage us to make poor architecture choices in other places. If not having this encourages us to pass JSON as Strings in the core logic of the tool, I think this is pretty sad.

> Are you also supportive of upper case method names and the other things that I tried to fix here?

Yes these changes look good to me.

> Here's the history of ImportParameters: https://github.com/OpenRefine/OpenRefine/commits/0aa3555669753a834174cd6e617fdaaf68106cfe/main/src/com/google/refine/importers/tree/ImportParameters.java
It got introduced as a side effect of "save the file location as a table column"

I agree bundling arguments into a POJO for the sake of reducing the number of parameters of functions is not a very nice move by itself, but I just want to make sure we are not making a precedent here that would prevent us to introduce a proper architecture for full importer configs represented as POJOs, just like we have in many other places (such as FacetConfig or ReconConfig).

tfmorris:
@wetneb Please provide actionable review comments for what needs to change for this to be approved.

wetneb (approved):
Sorry here is a formal approval: I am happy with this PR, I just wanted to make sure we are on the same page about the roadmap for the importing architecture.

tfmorris:
Thanks!

Issue #933, opened 2015-01-17
OpenRefine docker containers


psychemedia:
Would it be useful to make some official releases of OpenRefine available via dockerhub? I've sketched a couple of Dockerfiles at:
- [OpenRefine - latest release](https://hub.docker.com/r/psychemedia/docker-openrefine) (link fixed now)
- [OpenRefine - recent build from github](https://hub.docker.com/r/psychemedia/openrefine) (link fixed now)

that I think do the job, though there may be better ways of doing them?

In respect of _official_ releases, it could be handy to have:
- openrefine/openrefine-latest-build
- openrefine/openrefine-latest-release
- openrefine/openrefine-stable

An example of how to create a dockerfile that adds extensions to any of these base builds might also be useful for cut-and-paste tinkerers such as myself...

Ravenwater:
I like containers, so I want to think with you. Given the fact that openrefine is already a self-contained widget that minimally inserts itself into a machine, the only workflow that I can see an OpenRefine container would enable is dispatching it in the cloud in a fast and reliable way. To do that properly, the dockerfile would need to architect a storage hierarchy, so that there is some persistence and workspace to get non-trivial things done. If you are working with data sets that reside in a public/private cloud it would be interesting to push the OpenRefine functionality to the data and reduce cost, or increase performance, but that would only kick in for data sets that are over a couple GB. That again would make the storage design more important.

What was the use case that you designed the dockerfile for?

psychemedia:
@Ravenwater The particular use case I have been exploring is the assembly of virtual machines for use in distance education by remote students. I am part of a team writing a [distance education course on data management and data analysis for the UK Open University](http://www.open.ac.uk/courses/modules/tm351), and we are supplying students with a virtual machine that contains various databases (mongodb, postgresql), an analysis environment (IPython notebooks with a lot of preinstalled python packages), and OpenRefine.

The original model was to give students a single VM image configured using vagrant and docker. (As well as getting this particular machine built, I was also interested in ways in which we might support workflows for creating new VM configurations, as well as supporting VM configurations for courses that run once or twice a year for up to 5 years). The particular use case we have at the moment is for students to download a VM image and run it using VirtualBox on their own computer; but I was mindful that there might also be a requirement to support students who want to access the VM running in the cloud or on institutional cloud servers. (Another model might be a traditional university where students use computer labs and need to access software packaged and maintained by the institution on pubic access machines).

[...]

PR #2985, opened 2020-07-24
Add Dockerfile for easier exploration

kdungs:
Hi all,

when trying out OpenRefine on my machine, I didn't want to install Java so I wrote a little Dockerfile. Full disclosure: I'm not an expert on Docker and thus the file might not even follow best practices. The reason I am opening this PR is to start a discussion. I'm happy to add documentation, make changes, or even close the PR alltogether if you think it's a bad idea.

With this change, a user would be able to get their own version of OpenRefine running simply through

```bash
docker build -t openrefine:v1 .
docker run -d -it -p3333:3333 openrefine:v1
```

and then point their browser to `http://localhost:3333`.

I can imagine this might be useful to people who don't live in the Java ecosystem but want to give OpenRefine a try anyway. Let me know what you think :) 

Cheers,
Kevin

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM maven:3.6.3-openjdk-8
+
+COPY . /refine
+RUN apt-get update -y \
+    && apt-get install -y curl \
+    && /refine/refine build
+EXPOSE 3333
+WORKDIR /refine
+ENTRYPOINT ["/refine/refine", "-i", "0.0.0.0", "run"]


thadguidry:
Hi @kdungs Kevin, I'm not sure about others opinions, but I think this is something that should be hosted yourself in your own repo towards DockerHub.  The project team doesn't want to be responsible for maintenance of non-official installation mechanisms. However, you could certainly add a link or page to our Wiki for general promotion but as an unofficial alternative installation as @psychemedia already did for example here: https://github.com/OpenRefine/OpenRefine/wiki/Jupyter

In issue #933 we are discussing Docker container maintenance.  So feel free to read through it and give your comments within.

@wetneb I think we should close this given my statements above and look forward to @kdungs future contributions.

tfmorris:
@kdungs Out of curiosity, what made you think you needed to build from source? Are our binary distributions not visible enough or were you running on an operating system that we don't support?

kdungs:
Hi folks,

thanks for looking at this so quickly and for taking the time.

I must admit that I simply overlooked the binary distributions and when straight to running from source. Must have been habit :) That makes thinks a lot easier and removes this "just trying it out via Docker" usecase alltogether.

Since openly maintained Dockerfiles seem to exist already (as discussed in #933), I'll simply close this then.

tfmorris:
Thanks for investing the effort! I hope your investigations of OpenRefine are productive. Please let us know on the mailing list if you have any questions.

PR #2989, opened 2020-07-26
Give the name to button 'Back'

### LINKED ISSUE #2980: Create Project > multiple files > preview: "Back" button not displaying properly
When you start a project by loading an archive file with multiple files inside, there is a "Back" button loaded next to "Start Over" on the preview page. But the "Back" button isn't actually displaying. 

### To Reproduce
1. Create a project with a .zip or etc. 
2. Then, you see a file-selection screen.
3. Finally, the preview screen loads with a malfunctioning "Back" button.

### Screenshots
back-button-not-loading

### Versions
3.4 beta 2
Windows 10
Firefox, and I reproduced it in Chrome

chetan-v:
Fixes #2980

### DIFF
--- a/main/webapp/modules/core/scripts/index/default-importing-controller/parsing-panel.html
+++ b/main/webapp/modules/core/scripts/index/default-importing-controller/parsing-panel.html
@@ -3,7 +3,7 @@
   <table>
   <tr>
   <td width="1px"><button bind="startOverButton" class="button"></button></td>
-  <td width="1px"><button bind="previousButton" class="button"></button></td>
+  <td width="1px"><button bind="previousButton" class="button">Back</button></td>
   <td width="auto" id="or-import-parsopt"></td>
   <td width="100px" style="text-align: right;" id="or-import-projname"></td>
   <td width="200px">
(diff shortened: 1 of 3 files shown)


darecoder1999:
> In the reference of #2980

@chetan-v Instead of this, try to use fixes #2980 or closes #2980.

main/webapp/modules/core/scripts/index/default-importing-controller/parsing-panel.html
```diff
@@ -3,7 +3,7 @@
   <table>
   <tr>
   <td width="1px"><button bind="startOverButton" class="button"></button></td>
-  <td width="1px"><button bind="previousButton" class="button"></button></td>
+  <td width="1px"><button bind="previousButton" class="button">Back</button></td>
```
wetneb, line 6: Good job! But this should be translatable, so it should not be added to this HTML file, but added to the DOM in the corresponding JS file using `i18n`.

chetan-v: got it

PR #2991, opened 2020-07-26
Update EntityCache class

darecoder1999:
Added `getMultipleDocuments` method

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/utils/EntityCache.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/utils/EntityCache.java
@@ -23,8 +23,9 @@
  ******************************************************************************/
 package org.openrefine.wikidata.utils;
 
-import java.util.concurrent.TimeUnit;
-
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import org.wikidata.wdtk.datamodel.helpers.Datamodel;
 import org.wikidata.wdtk.datamodel.interfaces.EntityDocument;
 import org.wikidata.wdtk.datamodel.interfaces.EntityIdValue;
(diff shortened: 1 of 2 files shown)


extensions/wikidata/src/org/openrefine/wikidata/utils/EntityCache.java
```diff
@@ -77,6 +80,13 @@ public static EntityCache getEntityCache() {
         return _entityCache;
     }
 
+    public Stream<EntityDocument> getMultipleDocuments(List<EntityIdValue> entityIds) throws ExecutionException {
```
wetneb, line 83: Why `Stream` and not `List`?

extensions/wikidata/src/org/openrefine/wikidata/utils/EntityCache.java
```diff
@@ -77,6 +80,13 @@ public static EntityCache getEntityCache() {
...
+        List<String> ids = new ArrayList<>();
+        for (EntityIdValue entityId : entityIds) {
+            ids.add(entityId.getId());
+        }
```
wetneb, line 87: You could probably use the streaming API for this instead: `entityIds.stream().map(…).collect(…)`

darecoder1999:
Earlier, I made changes over [here](https://github.com/OpenRefine/OpenRefine/blob/master/extensions/wikidata/src/org/openrefine/wikidata/utils/EntityCache.java#L55) and changed it `new CacheLoader<List<String>, Map<String, EntityDocument>>()` and `load` method as `public Map<String, EntityDocument> load(List<String> entityIds)` and so on the other dependent files. Can't be this a solution?

Then, we don't have to use `getEntityDocument` else we will have to use `getEntityDocuments`

wetneb:
No, this would not work. Consider the following scenario

```
EntityCache cache = new EntityCache();
cache.get(Arrays.asList(idA, idB)); // this triggers one HTTP request to Wikidata: fine
cache.get(Arrays.asList(idC, idD)); // this triggers another HTTP request to Wikidata, since these ids are new: fine again
cache.get(Arrays.asList(idB, idC)); // this will trigger another HTTP request again, because the list "idB, idC" has never been requested before, even if those documents are already in the cache! This is bad.
```

wetneb:
Steps for next week (minutes from our meeting):
* Fix the `loadAll` method signature
* Add a test which loads multiple ids at once in `EntityCacheTests` (feel free to reuse the scenario above)
* Introduce a class to extract the set of `PropertyIdValue` mentioned in a `WikibaseSchema`; add tests for it
* Wire up this property extractor as follows:
   * before inspecting a batch, extract all the property id values that appear in the schema;
   * prefetch these properties with EntityCache
   * run scrutinizers as usual

PR #2996, opened 2020-07-27
Clustering dialog choices limit & performance improvements - fixes #695

### LINKED ISSUE #695: Number of candidates returned for clustering should be capped
I have a file where 31,787 clusters were found and the top clusters have 39 variants in them.

We need to cap the number returned to the browser because otherwise it becomes unresponsive (or crashes).

tfmorris:
Fixes #695
- Caps the total number of choices displayed at 10,000 and warns when over the limit. Users can use facets to tune which clusters are displayed.
- Doubles the performance of the Javascript processing
- Only displays count of rows for a choice if it's > 1 to DOM elements
- Adds internationalization for row count

For 41K clusters containing 118K choices, processing dropped from 3m20s to 1m20s, but with the 10K choice cap total time is ~10sec.


wetneb:
I tried testing this on a dataset (attached) which looks like this:
```
00001b
00001B
00002b
00002B
...
09999b
09999B
10000b
10000B
```
Which should generate at least 10000 clusters. It freezes my browser when I try to cluster that with the default settings (binning, fingerprint).
I have not checked what the clusters concretely look like, but I suspect there could be clusters with many values in it. Perhaps the number of values in a given cluster might be worth capping too? Or the 10000 limit might be still too much for my modest machine.

tfmorris:
That data set generates 10K clusters of size 2, which gets limited to 5K clusters of 2 by the new code. It **is** counting choices, not clusters, but stops on a cluster boundary, so it could theoretically have an issue with giant clusters, but hopefully that's pretty rare (it's statistically unlikely for a cluster to have thousands of choices).

There's another pathological aspect to the test data set, which is that all the clusters have exactly the same characteristics, so there are no facets generated, meaning that there's no way to get to the second 5K clusters, other than to merge the first 5K. Fixing that would require introducing pagination, which is beyond the scope of this project.

I can reduce the limit, but if I do that, I'll probably need add a preference to let users increase it again. What seems like a reasonable limit? (remember, it's total choices across all clusters, not the number of clusters) I think the default text facet choice limit is 2K.

main/webapp/modules/core/scripts/dialogs/clustering-dialog.js
```diff
@@ -256,23 +272,28 @@ ClusteringDialog.prototype._renderTable = function(clusters) {
...
                 .bind("keyup change input",function() {
                     cluster.value = this.value;
                 }).appendTo(tr.insertCell(4));
+
+            return choices.length;
         };
 
-        for (var i = 0; i < clusters.length; i++) {
-            renderCluster(clusters[i]);
+        var maxRenderRows = 10000;
```
tfmorris, line 284: @wetneb If you want to experiment with the number of rows that your machine can handle, the limit is set here

wetneb:
Thanks! For me it takes about 20 seconds to display anything, so I would aim for a lower limit indeed (perhaps 5000?).

The UI text could perhaps be improved slightly:

> Exceeded row limit of 10000 choices
> 5000 clusters filtered from 10000 total

I would not call the cap a "row limit", since users could think we only looked at the first N rows when clustering, whereas we just cap the number of cluster items (or values) that we display, right?

The second line is also a bit confusing to me. The first thing that comes to mind is: if the limit is 10000, why does it only show 5000? Also "filtered" is a bit ambiguous, it could mean the clusters that we actually display or those we have removed.

tfmorris:
How about "Exceeded limit of 5000 total choices"? That was my intent since having both "rows" and "choices" is silly, but I apparently forgot to make the change.

The second line's wording is unchanged. It previously only reported the effect of the facets, but now includes both the facets and the choice cap. I don't mind changing it though. How about "included" instead of "filtered"?

wetneb:
Sounds good to me!

tfmorris:
@wetneb Anything else needed here?

PR #3011, opened 2020-07-29
Add improvement-message for vantage

### LINKED ISSUE #2987: Importing problem when we choose two or more files.
When you choose two files (.xml and .csv)extensions then it will not work properly in importing.
If we checked only one file at a time then it work fine but we click back and checked another extension file then it will not guessed the right file.

### To Reproduce
1. First, go to "This computer" 
2. Then, choose two files at a time [TestFiles](https://drive.google.com/drive/folders/1skaTnzTcTuLJM1HLYC7t6iAh4NN7AB0E?usp=sharing)
3. And press "Configure Parsing Options"
4. and see what happens
### Current Results

(This issue is more understand by these two videos )

ezgif com-video-to-gif

ezgif com-video-to-gif (1)

### Expected Behavior

give an error when we parse two file at a time.

### Versions
 - Operating System:  Ubuntu 18.04 
 - Browser Version: Chrome

chetan-v:
Fixes #2987 

Add message :-
"OpenRefine can only import multiple files into a single project if all selected files have the same number of columns."

### DIFF
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -69,6 +69,7 @@
     "core-index-import/unknown-err": "Unknown error",
     "core-index-import/error": "Error:",
     "core-index-import/select-file": "Select Files to Import",
+    "core-index-import/single-project": "OpenRefine can only import multiple files into a single project if all selected files have the same file format.",
     "core-index-import/several-file": "There are several files available. Please select the ones to import.",
     "core-index-import/sel-by-extension": "Select by Extension",
     "core-index-import/sel-by-regex": "Select by Regex on File Names",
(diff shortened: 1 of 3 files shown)


thadguidry:
@tfmorris I'd like to get your review and agreement with this change for improvement of messaging in the UI for users against #2987 If you think it needs to be slightly reworded that's fine, just let @chetan-v know here.

thadguidry:
@chetan-v Looks good!

> OpenRefine can only import multiple files into a single project if all selected files have the same number of columns.

image

wetneb:
At the moment we do "support" importing multiple files with different numbers of columns (by merging these columns together). It is a bit brittle (column order depends on the order in which files are imported, for instance) but it should work reasonably well as long as all the files are of the same format.

So I think the message should be changed to something like
"OpenRefine can only import multiple files into a single project if all selected files have the same file format" instead.

PR #3019, opened 2020-08-01
Replace Javascript string concatenation with i18n parameters - addresses #1858

tfmorris:
refs #1858
- Remove Javascript string concatentation and use jquery i18n() instead so that translators have the needed context and flexibility to be able to do a good job.
- Also remove code-based plurals conditionalization and replace with i18n.
- Add missing localization in database extension


main/webapp/modules/core/langs/translation-en.json
```diff
@@ -369,21 +369,19 @@
...
+    "core-project/cancel-all": "{{plural:$1|Cancel|Cancel All}}",
     "core-project/canceling": "Canceling…",
-    "core-project/last-op-er": "The last operation encountered some errors",
-    "core-project/continue-remaining": "Continue with the remaining operations",
+    "core-project/last-op-er": "The last operation encountered some errors:",
+    "core-project/continue-remaining": "Continue with the remaining operations?",
     "core-project/undo": "Undo",
     "core-recon/add-std-srv": "Add standard service",
-    "core-recon/access": "Access",
+    "core-recon/access": "» Access",
```
wetneb, line 380: This "» " character feels more like a replacement for an arrow symbol than being really part of the message, so I would leave it outside of i18n intuitively.

Because this is just a part of the `"Access <a href="…">Service API</a>"` message it would be nice if we could merge it with the string below (`core-recon/service-api`) so that these words can be reordered but I am not sure if it is doable without adding HTML to the messages (I am not sure what the risk/benefit balance should be here)

tfmorris: I include the right quote character as a stop gap in case there are locales where the glyph is confusing or inappropriate. (I don't have any way to judge how likely that is.)

There is a whole category of "concatenation by juxtaposition" problems which are not addressed by this PR. That is the (unchecked) item 3 on the checklist in #1858.

The correct solution is probably to have a single phrase which is either all linked or which is followed by a linked "chain link" icon. The latter is probably more attractive visually. I can take a look at this one, but I'm intentionally not addressing this category as a whole.

tfmorris: I fixed this sole example, but don't plan to tackle any other in this category.

PR #3027, opened 2020-08-03
Fix ToDate test failure & inefficiency - fixes #3026

### LINKED ISSUE #3026: ToDate("2012-03-01","XXX") test failure & wasted conversion attempts
I'm investigating a local test failure which I'm unsure whether is specific to OS X or just how timezones are set up on my particular machine.

The failing test is:

      Assert.assertEquals(invoke("toDate", "2012-03-01","XXX"), invoke("toDate", "2012-03-01"));

which returns 2012-03-01T01:00:00 instead of 2012-03-01T00:00:00

This appears to have something to do with the default fallthrough case here, perhaps due to Daylight Savings Time: https://github.com/OpenRefine/OpenRefine/blob/306b541c699e7e87d5e08be1310575b85b15dbf0/main/src/com/google/refine/expr/functions/ToDate.java#L170-L172

While investigating this, I came across this code:

https://github.com/OpenRefine/OpenRefine/blob/306b541c699e7e87d5e08be1310575b85b15dbf0/main/src/com/google/refine/expr/functions/ToDate.java#L135-L145

which I don't completely understand, but in the test case creates an array with 160 copies of the (invalid) code "XXX" which it then uses to attempt to parse the date 160 times, always failing, before falling back to the code above

tfmorris:
Fixes #3026. 

- Use `Instant.truncate(DAYS)` instead of homebrew offset calculation.
- Fix handling of unknown language code as first format string. It was creating 160 copies of it and trying to convert (unsuccessfully) with them all.

The unknown language code used to return an `EvalError` but that behavior was changed and the test that covered it was disabled. Presumably this was intentional, but I'm not sure why. 
@ostephens do you know why?

### DIFF
--- a/main/src/com/google/refine/expr/functions/ToDate.java
+++ b/main/src/com/google/refine/expr/functions/ToDate.java
@@ -62,7 +62,7 @@ public class ToDate implements Function {
     public Object call(Properties bindings, Object[] args) {
         String o1;
         Boolean month_first = null;
-        List<String> formats =  new ArrayList<String>();  
+        List<String> formats =  new ArrayList<>();
         OffsetDateTime date = null;
         
         //Check there is at least one argument
@@ -84,7 +84,11 @@ public class ToDate implements Function {
         }
         
         if(args.length==1) {
-            date = parse(o1,true,formats);
+            try {
+                date = parse(o1, true, formats);
+            } catch (DateFormatException e) {
+                // Should never happen since we're using an empty format list
+            }
         } else if (args.length > 1) {
             if(args[1] instanceof Boolean) {
                 month_first = (Boolean) args[1];
@@ -93,19 +97,21 @@ public class ToDate implements Function {
             } else {
                 return new EvalError("Invalid argument");
             }
-            for(int i=2;i<args.length;i++) {
+            for (int i = 2; i < args.length; i++) {
                 if (!(args[i] instanceof String)) {
-                    // skip formats that aren't strings
-                    continue;
+                    return new EvalError("Invalid non-string format argument " + args[i].toString());
                 }
                 formats.add(StringUtils.trim((String) args[i]));
             }
-            if(month_first != null) {
-                date = parse(o1,month_first,formats);
-            } else {
-                date = parse(o1,formats);
+            try {
+                if (month_first != null) {
+                    date = parse(o1, month_first, formats);
+                } else {
+                    date = parse(o1, formats);
+                }
+            } catch (DateFormatException e) {
+                return new EvalError(e.getMessage());
             }
-            
         }
         if(date != null) {
             return date;
@@ -113,7 +119,7 @@ public class ToDate implements Function {
         return new EvalError("Unable to convert to a date");
     }
     
-    private OffsetDateTime parse(String o1, Boolean month_first, List<String> formats) {
+    private OffsetDateTime parse(String o1, Boolean month_first, List<String> formats) throws DateFormatException {
         if(month_first != null) {
             try {
                return CalendarParser.parseAsOffsetDateTime( o1, (month_first) ? CalendarParser.MM_DD_YY : CalendarParser.DD_MM_YY);
@@ -144,7 +150,7 @@ public class ToDate implements Function {
         return locale;
     }
 
-    private OffsetDateTime parse(String o1, List<String> formats) {
+    private OffsetDateTime parse(String o1, List<String> formats) throws DateFormatException {
         Locale locale = getLocale(formats);
         DateFormat formatter;
         OffsetDateTime date;
@@ -154,8 +160,7 @@ public class ToDate implements Function {
                 try {
                     formatter = new SimpleDateFormat(formats.get(i),locale);
                 } catch (IllegalArgumentException e) {
-                    // TODO: We should pass back error info here to let user know format string is bad
-                    return null;
+                    throw new DateFormatException("Unable to parse date format " + formats.get(i));
                 }
                 date = parse(o1, formatter);
                 if (date != null) {
@@ -202,6 +207,14 @@ public class ToDate implements Function {
     @Override
     public String getReturns() {
         return "date";
+    }
+
+    class DateFormatException extends Exception {
+        private static final long serialVersionUID = -6506736145451835731L;
+
+        public DateFormatException(String string) {
+            super(string);
+        }
     }
 
 }
--- a/main/src/com/google/refine/expr/functions/ToDate.java
+++ b/main/src/com/google/refine/expr/functions/ToDate.java
@@ -41,6 +41,7 @@ import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
+import java.util.IllformedLocaleException;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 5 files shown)


tfmorris:
I found my code from 2013 https://github.com/OpenRefine/OpenRefine/commit/fa072df85c27ce601f867a7f39c5a5d1bf288dec#diff-c7ab88e74f064de9905c27b627771bdd so I think I understand what this is trying to do now - the first format string can optionally be a language tag. I should be able to work out what needs to be done to fix it.

Marking as Draft for now...

tfmorris:
A true fix for the second problem is now included along with a small simplifying refactoring.

main/src/com/google/refine/expr/functions/ToDate.java
```diff
@@ -168,8 +167,7 @@ private OffsetDateTime parse(String o1, Locale locale, List<String> formats) {
         } else {
             try {
                 return javax.xml.bind.DatatypeConverter.parseDateTime(o1).getTime().toInstant()
-                		.plusSeconds(ZonedDateTime.now().getOffset().getTotalSeconds())
-                		.atOffset(ZoneOffset.of("Z"));
+                        .truncatedTo(ChronoUnit.DAYS).atOffset(ZoneOffset.of("Z"));
```
wetneb, line 170: It is not clear to me that this new version is equivalent, for instance of datetime objects with precision down to the second. You are correct that the motivation for this `.plusSeconds()` was to preserve the date part across timezones but I think this code also handles timestamps with a higher precision than day.

tfmorris: Can you add a (failing) test that demonstrates the problem? I'm not sure I follow what you're trying to say.

My understanding of the intent was to make sure that `toDate("1984")` ends up pinned to midnight, rather than offset by the number of hours that local time is from UTC. The old code apparently doesn't account for summer time / Daylight Savings Time so, on my system, it ends up at 1 hour past midnight rather than at midnight. The new code fixes that. I'm open to other solutions, but I'm trying to resist having to learn all of the complexities of the new Java date/time system.

tfmorris: I've got a better understanding of what the original problem was and have a better fix. The issue occurred when the test date to be parsed was on one side of the Daylight Savings Time boundary and the current date was on the opposite side, resulting in the offset computed using `now()` to be incorrect. I switched the offset calculation to use the parsed date.

Somehow, in the mean time, the default timezone for Java switched itself from "EDT" to "UTC" so I also changed the test to use a fixed, non-UTC timezone to hopefully catch any other UTC dependencies in the future (although it won't catch temporal shifts like DST).

tfmorris:
I think this may affect more than just Daylight Savings Time, e.g. `xsd:DateTime`s which include embedded timezone info, so I need to review the test coverage.

Looking at the code which is transforming to UTC, it feels like the definition of an OpenRefine date has shifted at some point and I'm not sure that was intentional. I think it was always, at least implicitly, a local datetime. This, of course, has problems, but migrating to a different definition should be done in a considered manner. It also seems like there was, perhaps, an opportunity missed in updating to the new Java date support to also update parsing code to parse to the new datatypes directly.

tfmorris:
Although I've fixed the timezone offset handling to work as intended, I'm not convinced that the original intention is correct.

I've added a few more tests and one of the tests, that a datetime without a timezone qualifier is interpreted as local, fails because of this code that's casting this one particular case to UTC.

main/tests/server/src/com/google/refine/expr/functions/strings/ToFromConversionTests.java
```diff
@@ -167,12 +170,22 @@ public void testToDate() throws CalendarParserException {
...
+      Assert.assertEquals(invoke("toDate", "01-六月-2013","zh", "MMM-dd-yyyy", "dd-MMM-yyyy"), CalendarParser.parseAsOffsetDateTime("2013-06-01"));
 
       // If a long, convert to string
       Assert.assertEquals(invoke("toDate", (long) 2012), invoke("toDate", "2012-01-01"));
 
       // If already a date, leave it alone
       Assert.assertEquals(invoke("toDate", CalendarParser.parseAsOffsetDateTime("2012-03-01")),CalendarParser.parseAsOffsetDateTime("2012-03-01"));
+
+      // FIXME: Date/times without timezone info should be interpreted as local time, not UTC
+//      Assert.assertEquals(invoke("toDate", "2013-06-01T13:12:11"), CalendarParser.parseAsOffsetDateTime("2013-06-01 13:12:11"));
```
tfmorris, line 182: @wetneb I think this test should pass, but it fails due to UTC timezone offset code that you added. Can you review and see if you agree?

thadguidry: @tfmorris I don't want to poke my eyes or mouth into this between both of you only to say this bit of history:
I recall Jacky in a chat asking me about formatter expectations (GREL, cell values, etc.) because he was indeed having issues with correct parsing behavior based on what we expected or not.  That probably should have been a signal for me to ask Antonin or someone else (who was still newish to our team), but I think that discussion did eventually happen between him and Antonin in the contributing PRs and original issues for Java Date that he worked on.  I recall Instant and Locale being 2 terms he asked me about concerning expectations and I didn't have enough expertise.

The ISO_INSTANT would be UTC, yes? https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/time/format/DateTimeFormatter.html#ISO_INSTANT
But the ISO_LOCAL_DATE_TIME would be local, no? https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/time/format/DateTimeFormatter.html#ISO_LOCAL_DATE_TIME

This is all I recall of the conversations with him, sorry.  Hope it helps!

thadguidry: Speaking of which...which Resolver style do we feel is most useful for regular OpenRefine users? https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/time/format/ResolverStyle.html#STRICT

wetneb: @tfmorris that proposed test looks obviously sensible to me.

Broadly speaking, my user-side requirements for this would be:
- if I parse a date string and then extract some date part from it, the value I get should be consistent with what was in the original string
- the rendered version of the date should be consistent with the source string.

I have not checked but that might require changes to `datePart` and cell rendering, to make sure these work correctly for local date times.

tfmorris:
Thanks both for the feedback. That's useful background. I think fixing everything is too big for this PR, so I'll leave it as is and start a thread on the dev list to figure out what needs to be done with dates in general. It looks like the switch to interpreting things as UTC vs local was made in May 2018, so there are probably compatibility concerns with both restoring the original behavior or leaving it as is.

Since I didn't get an answer on my question about the reasons for the test behavior change, I've restored the previous behavior of treating a bad format string as an error rather than silently ignoring it. I also restored some other previous tests that had been removed, including those for language tags with variants, e.g. zh-CN, although zh-TW and zh-HK which work on my machine and JDK 8 in Travis don't appear to work for later JDKs.

Issue #1998, opened 2019-03-27
Diff function has unpredictable results


thadguidry:
**Describe the bug**
Using Diff and comparing Dates for equality in 2 columns sometimes leads to unexpected results.

**To Reproduce**
1. Have 2 columns of dates to compare, where column `end_date` value is `2900-01-01-T00:00:00Z` and column `begin_date` value is `1923-08-03-T00:00:00Z`
2. Compare dates in 2 columns with this GREL:
`diff(cells['end_date'].value, cells['begin_date'].value, "days")`

**Current Results**
?

**Expected behavior**
Negative number is expected

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Version: Chrome
 - JRE or JDK Version: JRE 1.8.0_181]

**OpenRefine (please complete the following information):**
 - Version OpenRefine 3.1

**Additional context**
See Mailing List https://groups.google.com/d/msg/openrefine/cwgIgA7-d8k/OJmEAc4nBAAJ

This is possibly being caused by an incorrect comparison.
From java.time.OffsetDateTime notes: "This is a value-based class; use of identity-sensitive operations (including reference equality (==), identity hash code, or synchronization) on instances of OffsetDateTime may have unpredictable results and should be avoided. The equals method should be used for comparisons." - https://docs.oracle.com/javase/8/docs/api/java/time/OffsetDateTime.html

Our code is here:
https://github.com/OpenRefine/OpenRefine/blob/ebaad96dfc33f44b1a0904800cd8bdac7c655220/main/src/com/google/refine/expr/functions/strings/Diff.java#L58

wetneb:
I agree that's an important issue but "critical" is probably a bit over the top, no? :)

ostephens:
Is this definitely a problem on 3.1/3.2? I did some limited testing earlier and seemed ok. Happy to test further but would appreciate any definitive examples. 

Note report on mailing list is for LODRefine which is now v outdated & not directly supported by main project

thadguidry:
@wetneb We typically used critical if it affects the reliability of data transforms.  This is borderline, and in those cases in the past, both Tom and David would mark as critical, so I'm following our precedent that we had.  But happy to change that precedent.

wetneb:
One thing that would help is just to isolate the bug as the evaluation result of a single GREL expression, such as `diff(toDate(...), toDate(...), ...)`, and check in which version the behaviour changed (if any).

[...]

PR #1597, opened 2018-05-10
unify the internal date type 

jackyq2015:
Upgrade to Java 8 time API, unify the internal date type to OffsetDateTime. 

Kept java.util.Date and Calendar in some logger classes and utilities which is not related to the internal data structure.

Fix Issue #1588

Issue #1287, opened 2017-10-24
datePart() does not parse out milliseconds correctly


thadguidry:
![datepart_incorrect_milliseconds](https://user-images.githubusercontent.com/986438/31972204-1d068ff2-b8e5-11e7-9ccb-bc031a04c335.PNG)

With value being :

```05/18/2016 11:54:26.5550```

and applying GREL of 

```value.toDate("MM/dd/yyyy HH:mm:ss.S")```
or
```value.toDate("MM/dd/yyyy HH:mm:ss.SSSS")```
or even just
```value.toDate("MM/dd/yyyy HH:mm:ss.")```

it seems that the millliseconds are not parsed out as the SimpleDateFormat docs suggest
```
Date and Time Pattern | Result
-- | --
"yyyy.MM.dd G 'at' HH:mm:ss z" | 2001.07.04 AD at 12:08:56 PDT
"EEE, MMM d, ''yy" | Wed, Jul 4, '01
"h:mm a" | 12:08 PM
"hh 'o''clock' a, zzzz" | 12 o'clock PM, Pacific Daylight Time
"K:mm a, z" | 0:08 PM, PDT
"yyyyy.MMMMM.dd GGG hh:mm aaa" | 02001.July.04 AD 12:08 PM
"EEE, d MMM yyyy HH:mm:ss Z" | Wed, 4 Jul 2001 12:08:56 -0700
"yyMMddHHmmssZ" | 010704120856-0700
"yyyy-MM-dd'T'HH:mm:ss.SSSZ" | 2001-07-04T12:08:56.235-0700
"yyyy-MM-dd'T'HH:mm:ss.SSSXXX" | 2001-07-04T12:08:56.235-07:00
"YYYY-'W'ww-u" | 2001-W27-3
```
https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html

Maybe its actually a java.util.Calendar bug or we missed something within it ?

**Comment by ettorerizza** [comment]
Hi @thadguidry ,

Yep, I noticed the same thing as I tried the formulas you proposed in response to the question on the mailing list. I also get 550 milliseconds. In addition, I have the impression that there is an error in the documentation. It says :

```
> For a date of the form: "1/4/2012 13:30:00" use GREL function:
> 
> toDate(value,"dd/mm/YYYY H:m:s")
```

But using this method, I get a wrong date. 

[...]

PR #3035, opened 2020-08-06
Updating exploring.md

allanaaa:
Images still need some trimming. Haven't explained logs yet.


wetneb:
By "logs" do you mean OpenRefine server logs? I am not sure I would describe those there. It would be great to document how to access them, but probably in Running or some troubleshooting section instead?

allanaaa:
One thing I'd like feedback on:
The TOC sidebar is pretty long, and the facets section could use some breaking-up of its content. As you know, we only get two levels of hierarchy in this menu, so maybe we should move some things out of "Facets" and into their own grouping: "_____ facets" and "Customized facets", e.g. I'm not sure what to name this category. Default facets? Simple facets? Basic facets? Original facets?  

I do want all the facets currently listed to appear, but I can entertain arguments for demoting some of them (e.g. just showing "Customized facets" and then demoting everything under that to an h4 and making them disappear from the TOC). 

sidebar

thadguidry:
@allanaaa Another option is to actually create a GeneralFacets.md file as well as a CustomizedFacets.md file?  Then provide links to them in the exploring.md file?

allanaaa:
Oh yes, thanks! We had discussed splitting this page in two but I forgot to mention it here. 
I am not sure the precise structure you mention will work - I think it would make more sense to have "Exploring data" in the left-hand menu be a category containing multiple pages. Splitting this page into three pages as
- cell data types and record types
- facets and filter
- sort and view

would make the most sense to me, rather than having one page with two more sub-pages tucked inside the facets section. But I can experiment with that and see it's more seamless than I imagine, if there is interest.

thadguidry:
@allanaaa Yes, something like that... sure, let's try your suggestion to see how that feels!

docs/docs/manual/exploring.md
```diff
@@ -1,102 +1,346 @@
...
+OpenRefine understands records based on the content of the first column. Splitting a row into a multi-row record will base all association on the first column in your dataset. If you have more than one column to split out into multiple rows, OpenRefine will keep your data associated with its original record: you can imagine this structure as a tree with many branches, all leading back to the same trunk. 
 
-*   Functions that matter to rows/records distinction
+OpenRefine assigns a unique key behind the scenes, so your records don’t need a unique identifier. You can keep track of which rows are assigned to which record by the record number that appears under the “All” column.
 
+To [split multi-valued cells](transforming-data#split-multi-valued-cells) and apply other operations that take advantage of records mode, see [Transforming data](transforming-data). 
+
+:::caution
+Be careful when in records mode that you do not accidentally delete rows based on being blank in one column where there is a value in another. 
+:::
```
wetneb, line 56: It's not clear to me that this is the single most salient sort of issue we should flag about the records mode - perhaps it could be left in normal text without the caution box (which really jumps to the eyes).

docs/docs/manual/exploring.md
```diff
@@ -1,102 +1,346 @@
...
+
+Facet boxes that appear in the sidebar can be resized and rearranged. You can drag and drop the title bar of each box to reorder them, and drag on the bottom bar of text facet boxes. 
 
 ### Text facet
 
+A text facet can be generated on any column with the “text” data type. Select the column dropdown and go to “Facet” > “Text facet”. The created facet will be sorted alphabetically, and can be sorted by count. 
+
+A text facet is very simple: it takes the total contents of the cells of the column in question and matches them up. It does no guessing about typos or near-matches. 
+
+You can edit any entry that appears in the facet display, by hovering over the facet and clicking the “edit” button that appears. You can also automate the cleanup of facets by using clustering, with the “Cluster” button displayed within the facet window.
```
wetneb, line 135: The editing part might be worth expanding, it's a powerful feature that is not so easy to discover. It would be worth explaining that this will change all the corresponding cell values: it is an effective way to clean up values in a column. You could add a link to the documentation about clustering (to be written) which does that at a larger scale.

thadguidry: We internally call it a "mass edit" link, because it can massively edit from/to all the values in a column as Antonin says.

allanaaa:
@thadguidry mentioned nanosecond support on date formats. I just tested it out and it doesn't seem to be working. 
Assuming this format is correct: 2014-03-14T05:30:04.000789000Z from https://github.com/OpenRefine/OpenRefine/wiki/GREL-Date-Functions

OpenRefine either 
- truncates the .nnnn value, which it did on a toDate column operation, and when I changed the cell data type on "2014-03-14T05:30:04.000789000" (without the Z at the end)
- or it actually misconverts. It changed "2014-03-14T05:30:04.000789000Z" as text to "2014-03-14T01:30:04Z" as a date - pretty sure nanoseconds don't allow us the ability to go back in time, but I'm not a physicist. 

I assume we want neither truncation nor misconversion. 

Also, ISO 8601 states that it slightly prefers a comma instead of a decimal, but the toDate function doesn't support the comma at all (won't convert it). I think I should probably document that, yes?

docs/docs/manual/exploring.md
allanaaa: I added a sentence or two. I can probably say more here so I'll leave this open for now. Current text: 

> You can edit any entry that appears in the facet display, by hovering over the facet and clicking the “edit” button that appears. This will mass-edit every matching cell in your data. This is a great way to fix typos, whitespace, and other issues that may be affecting the way facets appear. You can also automate the cleanup of facets by using [clustering](transforming#cluster-and-edit), with the “Cluster” button displayed within the facet window.

allanaaa:
I've ripped most of the date-format information out of this page, since it's more about the transform operation & the GREL function. I've just expanded a bit on the cell data type with links out. We can deal with Date.parse and SimpleDateFormat later!
My comment above should probably be an issue on its own (assuming I'm not totally off-base with it). 

The last part of the cell-data-types section now reads 

> Converting a cell's data type is not the same operation as transforming its contents. For example, using a column-wide transform such as “Transform” > “Common transforms …” > “to date” may not convert all values successfully, but going to an individual cell, clicking “edit” and changing the data type can successfully convert text to a date.
> To transform data from one type to another, see [Transforming data](transforming#transform) for information on using common tranforms, and see [Expressions](expressions) for information on using `toString()`, `toDate()`, and other functions. 

... which I think is more or less correct based on our chat today, but is probably too vague to be helpful at this point in the documentation. I have written this as a "for example" but perhaps date formatting is the only case in which this is really valid (or important to understand). Thoughts?

tfmorris:
> @thadguidry mentioned nanosecond support on date formats. I just tested it out and it doesn't seem to be working.
> Assuming this format is correct: 2014-03-14T05:30:04.000789000Z from https://github.com/OpenRefine/OpenRefine/wiki/GREL-Date-Functions

A bunch of the examples (e.g. datePart(millis)) on that page are wrong, so I'm not sure I'd trust it. 

> OpenRefine either
> 
> * truncates the .nnnn value, which it did on a toDate column operation, and when I changed the cell data type on "2014-03-14T05:30:04.000789000" (without the Z at the end)
> * or it actually misconverts. It changed "2014-03-14T05:30:04.000789000Z" as text to "2014-03-14T01:30:04Z" as a date - pretty sure nanoseconds don't allow us the ability to go back in time, but I'm not a physicist.
> 
> I assume we want neither truncation nor misconversion.

Correct. Please file bug reports, if you haven't already done so. I don't think nanoseconds are supported, but it should give a reasonable error message. What timezone are you in? The 1:30 vs 5:30 thing sounds suspiciously like a timezone bug in the new date code (which is incredibly messed up).

> Also, ISO 8601 states that it slightly prefers a comma instead of a decimal, but the toDate function doesn't support the comma at all (won't convert it). I think I should probably document that, yes?

Yes, please, but as this comment https://github.com/moment/moment/issues/2580#issuecomment-136226383 says, pretty much everywhere in computer land uses a period decimal separator.

I think I'm missing some nuance of the difference between "convert" and "transform," but the behavior sounds wrong from the description. For cases of wrong behavior, we should: 1) file a bug report, 2) document the current behavior, but with a link to the issue for the correct behavior so that people know when it gets fixed.

thadguidry:
@tfmorris Nanosecond support landed here: PR #1597
Please also read #1287 and also maybe a bit of #1998 
I also went searching through my chat with Jacky at one point in time (which coincides with his [May 10, 2018 comment in #1597](https://github.com/OpenRefine/OpenRefine/pull/1597#issuecomment-388091137))...

> Qi (Jacky) Cui <cuiqi1202@gmail.com>
> Wed, May 9, 2018, 9:24 PM
> will add Nano to inc

tfmorris:
> Nanosecond support landed here: PR #1597

That is a refactoring PR, which, by definition, shouldn't include any functional changes, but it apparently has a half-baked attempt at nanosecond support. What it didn't do, which should have been part of the refactoring, was switch from the use of `SimpleDateFormat` or `GregorianCalendar`. Since `SimpleDateFormat` is used by `ToDate()` and doesn't support nanosecond precision, it's the likely cause of at least some of the bugs that @allanaaa is running into.

allanaaa:
Something I forgot to mention earlier:

Currently a section is titled "Row types: rows vs. records", which is unwieldy. I would love suggestions for a snappier title about records mode.

ostephens:
>Row types: rows vs. records
Why not just "Rows vs. Records"?

docs/docs/manual/exploring.md
```diff
@@ -1,102 +1,60 @@
...
 
-### Sort
+Generally, when you import some data, OpenRefine reads that data in row mode. From there you can convert the project into records mode. OpenRefine remembers this action and will present you with records mode each time you open the project from then on. 
 
+OpenRefine understands records based on the content of the first column, what we call the "key column." Splitting a row into a multi-row record will base all association on the first column in your dataset. If you have more than one column to split out into multiple rows, OpenRefine will keep your data associated with its original record: you can imagine this structure as a tree with many branches, all leading back to the same trunk. 
 
-### View: Collapsing (and expanding)
+For example, your key column may be a family name, with multiple family members identified by their first names, associated to that family name. You may have one or more phone numbers listed for each person. The phone numbers are linked to the first names, which are linked to the last name.
```
ostephens, line 52: I'd suggest that an ID of some kind would be a better example of a key column - the scenario you describe would work in a sense I think a unique ID that links the rows together is a better example of how records work

allanaaa:
Something I forgot to ask about, in the Sort menu:

> *   boolean values by false, then true, or by **core-views/true-false**

coreviews

I don't know what "core-views" means! Can anyone tell me?

thadguidry:
@allanaaa That's a bug... it's a placeholder for translated text in javascript... but looks like someone put in the literal string itself "core-views/true-false" and so it is not getting translated.
File a bug for this please, Thanks!

tfmorris:
> Currently a section is titled "Row types: rows vs. records", which is unwieldy. I would love suggestions for a snappier title about records mode.

Simply "Records vs Rows"? I find the "Row types:" prefix to be confusing. It's like saying "Apple types: Apples vs. Oranges". A record is made up of multiple rows.

Another possibility might be "Display modes" or something similar.

docs/docs/manual/facets.md
```diff
@@ -0,0 +1,284 @@
...
+
+The choices and counts displayed in each facet can be copied as tab-separated values. To do so, click on the "X choices" link near the top left corner of the facet.
+
+## Numeric facet
+
+![A screenshot of an example numeric facet.](/img/numericfacet.png)
+
+Whereas a text facet groups unique text values into groups, a numeric facet sorts numbers by their range - smallest to biggest. This displays visually and allows you to set a custom facet within that range. You can drag the minimum and maximum range markers to set a range. OpenRefine snaps to some basic equal-sized divisions - 19 in the example set above. 
+
+You will be offered the option to include blank, non-numeric, and error values in your numeric visualization; these will appear in the visual range as “0” values.
```
ostephens, line 93: I don't think this is correct. Blank, non-numeric and error values don't appear as zero values - they aren't included in the bar chart - they are included/excluded based on checking the relevant boxes

docs/docs/manual/facets.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Customized facets
+
+Customized facets have been added to expand the number of default facets users can apply with a single click. They represent some common and useful functions you shouldn’t have to work out using an [expression](expressions).
+
+All facets that display in the “Facet/Filter” sidebar can be edited by clicking on the “change” button to the right of the column title. This brings up the expressions window that will allow you to modify and preview the expression being used. 
+
+### Word facet
+
+Word facet is a simple version of a text facet: it splits up the content of the cells based on spaces, and outputs each character string as a facet:
```
ostephens, line 169: I'm not sure I'd say "simple". It's a custom text facet

docs/docs/manual/facets.md
```diff
@@ -0,0 +1,284 @@
...
+
+When you click on your desired square, that two-column comparison will appear in the facets sidebar. From here, you can drag your mouse to draw a rectangle inside the scatterplot, which will narrow down to just the rows matching the points plotted inside that rectangle. This rectangle can be resized by dragging any of the four edges. To draw a new rectangle, simply click and drag your mouse again. To add more scatterplots to the facet sidebar, re-run this process and select a different square. 
+
+If you have multiple facets applied, plotted points in your scatterplot displays will be greyed out if they are not part of the current matching data subset. If the rectangle you have drawn within a scatterplot display only includes grey dots, you will see no matching rows.
+
+If you would like to export a scatterplot, OpenRefine will open a new tab with a generated PNG image that you can save. 
+
+## Custom text facet
```
ostephens, line 115: Is it worth mentioning somewhere that (unlike a Text facet) a Custom text facet does not support editing facet values to update the data in the cells used to create the facet?

docs/docs/manual/facets.md
```diff
@@ -0,0 +1,284 @@
...
+
+This facet creates a numerical chart, which offers you the ability to narrow down to a range of numbers. For example, lowercase characters are numbers 97-122, uppercase characters are numbers 65-90, and numerical digits are numbers 48-57. 
+
+### Facet by error
+
+An error is a data type created by OpenRefine in the process of transforming data. For example, say you had converted a column to the number data type. If one cell had text characters in it, OpenRefine could either output the original text string unchanged or output an error. If you allow errors to be created, you can facet by them later to search for them and fix them. 
+
+![A view of the expressions window with an error converting a string to a number.](/img/error.png)
+
+To output errors, ensure that you have “store error” selected for the “On error” option in the expressions window.
```
ostephens, line 242: I'd suggest "To store errors in cells" rather than "To output errors"

docs/docs/manual/facets.md
```diff
@@ -0,0 +1,284 @@
...
+
+You can also create a text facet on any column with the expression ```row.starred``` or ```row.flagged```.
+
+## Text filter
+
+Filters allow you to narrow down your data based on whether a given column includes a text string. 
+
+When you choose “Text filter” a box appears in the “Facet/Filter” sidebar that allows you to enter in text. Matching rows will narrow dynamically with every character you enter. You can set the search to be case-sensitive or not, and you can use this box to enter in a regular expression. This field supports [Java's regular expression language](http://download.oracle.com/javase/tutorial/essential/regex/).
+
+For example, you can employ a regular expression to view all properly-formatted emails:
```
ostephens, line 276: Is it worth giving a simple text filter example as well as a regular expression version?

docs/docs/manual/facets.md
```diff
@@ -0,0 +1,284 @@
...
+
+```facetCount(value, 'value', '[Column]') > 1```
+
+Duplicates facets are case-sensitive and you may wish to filter out things like leading and trailing whitespace or other hard-to-see issues. You can modify the facet expression, for example, with:
+
+```facetCount(trim(toLowercase(value)), 'trim(toLowercase(value))', 'cityLabel') > 1```
+
+### Numeric log facet 
+
+Logarithmic scales reduce wide-ranging quantities to tiny scopes. A log transformation can be used to make highly skewed distributions less skewed. If your numerical data is unevenly distributed (say, lots of values in one range, and then a long tail extending off into different magnitudes), a numeric log facet can represent that range better than a simple numeric facet. It will break these values down into more navigable segments than the buckets of a numeric facet. This facet can make patterns in your data more visible.
```
ostephens, line 189: I'm not sure "Logarithmic scales reduce wide-ranging quantities to tiny scopes" is very clear. What are "tiny scopes"?

I think the description from wikipedia is better:
"A logarithmic scale (or log scale) is a way of displaying numerical data over a very wide range of values in a compact way"

docs/docs/manual/sortview.md
```diff
@@ -0,0 +1,29 @@
...
+*   text alphabetically or reverse
+*   numbers by largest or smallest
+*   dates by earliest or latest
+*   boolean values by false, then true, or by **core-views/true-false**
+
+You can also choose where to place errors and blank cells in the sorting. Text can be case-sensitive or not: cells that start with lowercase characters will appear ahead of uppercase.
+
+![A screenshot of the Sort window.](/img/sort.png)
+
+After you apply sorting, you can make it permanent, remove it, reverse it, or apply a different **sorting**. You’ll find “Sort” in the project grid header to the right of the rows-display setting, which will show all current sorting settings.
```
ostephens, line 19: I think an explanation of in what sense sorts are usually not "permanent" (i.e. the original row/record order is preserved which can be seen by the numbers in the All column) and what making it "permanent" means (rows/records are renumbered sequentially based on the current sort and the "sort" disappears

Also think we need an explanation of how multiple sorts apply - i.e. if you apply multiple sorts at once, they are ordered and don't override each other (sort by author then title preserves the author orders and sorts titles within a single author).

This links to making sorts "permanent" as to get a sort of books by author then by title within author you'd want to:

Apply an author sort
Apply a title sort
Make the sort permanent

docs/docs/manual/exploring.md
```diff
@@ -1,102 +1,60 @@
...
-### Facet by star or blank or flag
-
-
-## Filtering 
+Unlike spreadsheets, OpenRefine doesn’t store formulas and display the output of those calculations; it only shows the value inside each cell. It doesn’t import any cell colors or text formatting. 
 
+## Cell data types
 
-### Overview
+Data loaded into OpenRefine is assigned a data type. By default every cell will be considered a “string”, but you can have OpenRefine convert cell contents into other data types later.
```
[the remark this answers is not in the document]
allanaaa: On import, OpenRefine considers everything a string at the preview stage, right? That's my experience. Are there specific file formats that import differently? 

I know in preview you can explicitly tell it to parse numbers. But we're no longer in "default" by then.

docs/docs/manual/exploring.md
```diff
@@ -1,102 +1,60 @@
...
+You can see data types in action when you preview a new project: check the box that says “Attempt to parse cell text into numbers” and cells will be converted to the number type based on their contents. You’ll see numbers change from black text to green if they are recognized.
 
+The data type will determine what you can do with the value. For example, if you want to add two values together, they must both be recognized as the “number” data type. 
 
-### Permalink function
+You can check data types at any time by:
+*   clicking “edit” on a single cell (where you can also edit the type)
+*   creating a Custom Text Facet on a column, and inserting `type(value)` into the “Expression” field. This will generate the data type in the preview, and you can facet by data type if you press “OK.”
 
+The data types supported are:
```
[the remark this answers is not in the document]
allanaaa: Okay, we've removed arrays entirely from this section. 
It now reads:

> The data types supported are:
> 
>     string (one or more text characters)
>     number (one or more characters of numbers only)
>     boolean (values of “true” or “false”)
>     date (ISO-8601-compliant extended format with time in UTC: YYYY-MM-DDTHH:MM:SSZ)
> 
> A “date” type is created when a text column is transformed into dates, or when individual cells are set to have the data type “date.”
> 
> OpenRefine recognizes two further data types as a result of its own processes:
> 
>     error
>     null
> 
> An “error” data type is created when the cell is storing an error generated during a transformation in OpenRefine.
> 
> A “null” data type is a special value which basically means “this cell has no value.”...

docs/docs/manual/exploring.md
allanaaa: I'm not sure why you want a unique ID for this. I think this example works well with just the film title, not the ID?

|IMDB ID|Work|Actor|Role|
|---|---|---|---|
|tt0032138|The Wizard of Oz|Judy Garland|Dorothy Gale|
|||Ray Bolger|"Hunk"|
||||The Scarecrow|
|||Jack Haley|"Hickory"|
||||The Tin Man|
|||Bert Lahr|"Zeke"|
||||The Cowardly Lion|
|||Frank Morgan|Professor Marvel|
||||The Gatekeeper|
||||The Carriage Driver|
||||The Guard|
||||The Wizard of Oz|
|||Margaret Hamilton|Miss Almira Gulch|
||||The Wicked Witch of the West|

docs/docs/manual/exploring.md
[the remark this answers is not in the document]
allanaaa: I hope "at first" is an acceptable compromise here. I still think it's a default. :)

docs/docs/manual/exploring.md
allanaaa: I've removed the unique ID for now - maybe the section needs more explication about how you don't actually need a key column of unique IDs (because OR does that work behind the scenes), but of course your data may need to be sorted (or un-sorted) not to confuse the issue. But I think that can wait for a second pass.

allanaaa:
I'm ready to merge this PR if everyone else is!

docs/docs/manual/exploring.md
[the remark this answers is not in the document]
ostephens: No I don't think that's true. Importers will have rules about what they do, and it's definitely possible to have an importer that defaults to all values being treated as strings but that's not a general rule. If you import from Excel behaviour is different to if you import from CSV. I agree that certain importers and the UI has some default behaviour here, but this isn't OpenRefine wide "default" - its specific to those scenarios.

ostephens:
I can't agree with the "default" or "at first" phrasing with regarded to cell data types - I think we need to be clearer here. I've just responded to your comment.

My ideal is that we fix this before the merge, but will be OK to merge as long as there is an issue for this so we don't forget to revisit

docs/docs/manual/exploring.md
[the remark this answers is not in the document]
allanaaa: I haven't seen any evidence of this behaviour but I'm happy to believe you if you say so. 

What about "Most of the time, every cell will be considered a string, but...." or something like that? It does seem to be the vast majority of cases.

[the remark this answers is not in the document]
tfmorris: OpenRefine importers preserve the data types for any format that supports them - Excel (XLS & XLSX), OpenOffice, Google Sheets, JSON, XML, probably others...

wetneb:
Let's fix this before merging then :)

docs/docs/manual/exploring.md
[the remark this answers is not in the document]
allanaaa: > Each piece of information (each cell) in OpenRefine is assigned a data type. Some file formats, when imported, can set data types that are recognized by OpenRefine. Cells without an associated data type on import will be considered a “string” at first, but you can have OpenRefine convert cell contents into other data types later.

🤷‍♀️

ostephens:
Happy with @allanaaa's latest suggestion

ostephens (approved):

PR #3042, opened 2020-08-07
Only 1 refine.ini file for Windows (not either)

thadguidry:


allanaaa:
Is this accompanying a code change? My understanding is that 

> you can modify the way openrefine.exe runs by editing openrefine.l4j.ini; you can modify the way refine.bat runs by editing refine.ini

thadguidry:
@allanaaa Hmm, that I am actually not 100% sure of!  Antonin knows the launcher mechanism better and updated it also.
@wetneb Do we have full control of all settings from within the `openrefine.l4j.ini` file just as we do with `refine.ini` file?

allanaaa:
I hardly ever fire up refine.bat! but I tested it just now to be certain and yes, refine.ini functions for refine.bat and refine.l4j.ini functions for openrefine.exe. 

The syntax for a few things is different (-Xmx1024M vs. REFINE_MEMORY=1024M) but for the JVM preferences it's the same AFAIK. 
Although it doesn't seem to work unless you start with "JAVA_OPTIONS=" in refine.ini - whoops! That's what I get for believing something I read on the internet.

thadguidry:
FYI, I'm also not sure how any of what I saw reported in this mailing list thread
(Known issue with Windows executable and 2 Memory settings?)
https://groups.google.com/g/openrefine/c/3ABCveZ8CXI/m/CJBDACpzBAAJ
is related to this PR or not.

allanaaa:
@thadguidry Definitely agree! I think the documentation so far really shows you how scattered it all seems:
https://docs.openrefine.org/manual/running#starting-with-modifications
https://docs.openrefine.org/manual/running#modifications-set-within-files
https://docs.openrefine.org/manual/running#preferences
(And the "language settings" page could just be a dropdown on the "preferences" page too.)

I recalll @tfmorris had a comment about setting Java-related preferences in the UI, there was some security issue around that perhaps? I can't speak to the technical limitations, but I vote for any and all possible consolidations.

thadguidry:
Issue #3053 is the result of my testing the 2 .ini files and seeing an issue with defaults and `openrefine.l4j.ini`

thadguidry:
Added new issue #3057 to only have 1 .ini file to config OpenRefine on Windows to avoid confusing users, as well as reduce our documentation around the confusing issue!

Closing this PR since we now have a good issue to work towards in #3057 and Antonin approved of on our doc call.

PR #3070, opened 2020-08-12
Changed List return type to ImmutableList in Wikidata schema package classes

### LINKED ISSUE #3064: Make Wikibase schema classes properly immutable
### To Reproduce

The classes that constitute the wikibase schema (in `org.openrefine.wikidata.schema`) are meant to be immutable.
But many of the getters in these classes return lists, which are not immutable in Java.

This means that it is possible to modify these lists after getting them from the schema classes, which should be forbidden. This can lead to subtle bugs, see #3060.

### Current Results
Getters of Wikibase schema classes return modifiable lists.

### Expected Behavior
The returned lists should be unmodifiable. Wrapping with `Collections.immutableList()` should be enough (it is only an unmodifiable wrapper around a list, without actually copying the list).

darecoder:
closes #3064

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/schema/WbItemDocumentExpr.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/schema/WbItemDocumentExpr.java
@@ -96,12 +96,12 @@ public class WbItemDocumentExpr implements WbExpression<ItemUpdate> {
 
     @JsonProperty("nameDescs")
     public List<WbNameDescExpr> getNameDescs() {
-        return nameDescs;
+        return Collections.unmodifiableList(nameDescs);
     }
 
     @JsonProperty("statementGroups")
     public List<WbStatementGroupExpr> getStatementGroups() {
-        return statementGroups;
+        return Collections.unmodifiableList(statementGroups);
     }
 
     @Override
--- a/extensions/wikidata/tests/src/org/openrefine/wikidata/schema/WbItemDocumentExprTest.java
+++ b/extensions/wikidata/tests/src/org/openrefine/wikidata/schema/WbItemDocumentExprTest.java
@@ -24,10 +24,12 @@
 package org.openrefine.wikidata.schema;
 
 import java.util.Collections;
+import java.util.List;
 
 import org.openrefine.wikidata.testing.JacksonSerializationTest;
 import org.openrefine.wikidata.updates.ItemUpdate;
 import org.openrefine.wikidata.updates.ItemUpdateBuilder;
+import org.testng.Assert;
 import org.testng.annotations.Test;
 import org.wikidata.wdtk.datamodel.helpers.Datamodel;
 import org.wikidata.wdtk.datamodel.interfaces.ItemIdValue;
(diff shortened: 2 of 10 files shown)


extensions/wikidata/src/org/openrefine/wikidata/schema/WbItemDocumentExpr.java
```diff
@@ -23,20 +23,20 @@
  ******************************************************************************/
 package org.openrefine.wikidata.schema;
 
-import java.util.Collections;
-import java.util.List;
-
+import com.fasterxml.jackson.annotation.JsonCreator;
```
tfmorris, line 26: Any chance of convincing your IDE to not reorder imports? It makes for a very noisy diff.

extensions/wikidata/src/org/openrefine/wikidata/schema/WbItemDocumentExpr.java
```diff
@@ -96,12 +96,12 @@ public ItemUpdate evaluate(ExpressionContext ctxt)
 
     @JsonProperty("nameDescs")
     public List<WbNameDescExpr> getNameDescs() {
-        return nameDescs;
+        return nameDescs.stream().collect(ImmutableList.toImmutableList());
```
tfmorris, line 99: `Collections.unmodifiableList()` is probably a simpler (and higher performance?) idiom here.
```suggestion
        return Collections.unmodifiableList(nameDescs);
```

extensions/wikidata/tests/src/org/openrefine/wikidata/schema/WbReferenceExprTest.java
```diff
@@ -77,4 +79,10 @@ public void testSerialize()
             throws JsonProcessingException {
         JacksonSerializationTest.canonicalSerialization(WbReferenceExpr.class, expr, jsonRepresentation);
     }
+
+    @Test
+    public void testUnmodifiableList() {
+        List<WbSnakExpr> snakExprs = expr.getSnaks();
+        Assert.assertTrue(snakExprs.getClass().getName().contains("Unmodifiable"));
```
wetneb, line 86: This way of testing immutability might be a bit brittle - the tests could fail if we use other implementations of unmodifiable lists which have a different name (typically, the immutable variant you first used). Perhaps a simpler way to test that would be to try to modify the list and check that it throws an exception?
For instance

```
@Test(expectedExceptions = UnsupportedOperationException.class)
public void testUnmodifiableList() {
        expr.getSnaks().clear();
}
```

extensions/wikidata/tests/src/org/openrefine/wikidata/schema/WbItemDocumentExprTest.java
```diff
@@ -92,11 +92,9 @@ public void testSerialize() {
...
 
-    @Test
+    @Test(expectedExceptions = UnsupportedOperationException.class)
     public void testUnmodifiableList() {
-        List<WbNameDescExpr> nameDescExprs = expr.getNameDescs();
-        Assert.assertTrue(nameDescExprs.getClass().getName().contains("Unmodifiable"));
-        List<WbStatementGroupExpr> statementGroupExprs = expr.getStatementGroups();
-        Assert.assertTrue(statementGroupExprs.getClass().getName().contains("Unmodifiable"));
+        expr.getNameDescs().clear();
+        expr.getStatementGroups().clear();
```
wetneb, line 98: This will only test if `expr.getNameDescs()` is immutable, since that will already throw an exception. The second line will not be executed if the first line throws an exception. So you need to split that into two tests.

PR #3081, opened 2020-08-15
Set Wikidata recon endpoint to https://wikidata.reconci.link/

### LINKED ISSUE #3080: Update Wikidata recon endpoint URL (again!)
I used to maintain a Wikidata reconciliation service at https://wdreconcile.toolforge.org/.

The toolforge is an infrastructure provided by the Wikimedia Foundation for developers to host tools to support the Wikimedia community. It provides a managed ecosystem which enables shared maintenance of tools by multiple contributors, easy access to Wikimedia resources (SQL replicas of the MediaWiki dbs), and other useful things.

Being a managed environment comes with perks (no need to set up SSL yourself, free domain name, no hosting bills) but also downsides (being faced with devops bugs or downtime outside your control). Recently I faced a serious bug which I was unable to fix myself ([HTTP 502 errors being seemingly cached by some upstream component](https://phabricator.wikimedia.org/T257405)). I have therefore decided to host the tool outside this environment, on a server I rent myself and with a domain name I own:

https://wikidata.reconci.link/

Therefore I propose to use this as the default reconciliation service in OpenRefine. This is not benign, since this is a web service I totally control. I have access to all the reconciliation queries submitted to the service, I could decide to shut it down anytime, forget to renew the domain name, or other things like that. That being said, the previous service was not official either (since it was not run by Wikidata itself) and I could also access all the queries run on the service.

Beyond the absence of cached HTTP 502 errors, the new service is also significantly faster because I took the opportunity to rewrite it to use asynchronous parallelism.

I would very much prefer to run this service in the scope of the OpenRefine project but there has been limited appetite from the rest of the team about the idea of running our own server. I have also been trying to encourage the Wikidata team to run the service themselves, but that is an ongoing effort which is not likely to come to fruition immediately.

wetneb:
Closes #3080.

This will need to be updated once #2872 is merged (since it will add another hard-coded URL to be changed).
Apologies for the conflicts with your work, @afkbrb!

### DIFF
--- a/docs/docs/manual/wikidata.md
+++ b/docs/docs/manual/wikidata.md
@@ -28,6 +28,6 @@ sidebar_label: Wikidata
 
 ## More resources
 
-[https://wdreconcile.toolforge.org/](https://wdreconcile.toolforge.org/) 
+[https://wikidata.reconci.link/](https://wikidata.reconci.link/) 
 
 [https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine](https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine) 
--- a/main/src/com/google/refine/importers/WikitextImporter.java
+++ b/main/src/com/google/refine/importers/WikitextImporter.java
@@ -736,7 +736,7 @@ public class WikitextImporter extends TabularImportingParserBase {
             String wikiUrl = JSONUtilities.getString(options, "wikiUrl", null);
             // Wikidata reconciliation endpoint, hardcoded because the user might not have it in its services
             String reconUrl = JSONUtilities.getString(options, "reconService",
-                  "https://wdreconcile.toolforge.org/en/api");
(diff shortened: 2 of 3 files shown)


wetneb:
Merging this soon to get 3.4 released unless there are any concerns.

wetneb:
Since concerns have been brought up on the mailing list, I have held back from merging this. We can just release 3.4 without it.

wetneb:
So, over the last few months I think we have been in a relatively stable situation. The endpoint at https://wdreconcile.toolforge.org/ (that we use by default in OpenRefine) is still deprecated because of stability issues, but these issues seem to have become less common since a big share of the traffic has shifted to https://wikidata.reconci.link/.

So I guess we can keep linking to this deprecated endpoint - there is no official replacement for it anyway. I am continuing the discussion with WMDE and other partners to plan the development of an official service.

PR #3094, opened 2020-08-17
Fix i18n key typo - fixes #3093

### LINKED ISSUE #3039: Speed up git commit extraction in Maven config
Extracting the commit information from the git repository when running `mvn compile` can take some time. There are ways to speed it up:
https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/462

### LINKED ISSUE #3093: Fix appearance of "sort by boolean" options
As per https://github.com/OpenRefine/OpenRefine/pull/3035#issuecomment-674996783

### Current Results
coreviews

### Expected Behavior
The instance of "core-views/true-false" should be "true then false" I believe - with translation. 

(Personally, I would prefer commas too - "true, then false" and "false, then true" but that's just a preference.)

tfmorris:
Fixes #3093. Key was wrong in all translations too.


thadguidry:
@tfmorris that's the wrong fixed issue number.

PR #3096, opened 2020-08-18
Add i18n in formateContainer using Switch statement

### LINKED ISSUE #3069: Importer names not internationalized (i18n)
These strings in controller.js need to be internationalized so that they can be translated:
Screen Shot 2020-08-12 at 12 19 23 PM

https://github.com/OpenRefine/OpenRefine/blob/0ce5c5de13e63ad63c95af08366a4d5a4ba2aadd/main/webapp/modules/core/MOD-INF/controller.js#L207-L229

I haven't looked into how they are being fetched/used, so this may not be completely straightforward.

chetan-v:
Fixes #3069 

Add switch Statement for initializing i18n


wetneb (commented):
Thanks! The problem with this approach is that this only works for importers registered by the main module. Extensions also need to be able to localize the name of their importers, when they register them in their `controller.js` file.
As far as I can tell, to solve this issue we need to introduce a proper localization system in the backend.

chetan-v:
> Thanks! The problem with this approach is that this only works for importers registered by the main module. Extensions also need to be able to localize the name of their importers, when they register them in their `controller.js` file.
> As far as I can tell, to solve this issue we need to introduce a proper localization system in the backend.

Hi @wetneb  I did some change in backend for make it dependable on parsing-panal.js 
(only for labels). may be it will be wrong in some cases 
please check for certain cases.
 
> Extensions also need to be able to localize the name of their importers



` IM.registerExtension(".txt", "text");`
  `IM.registerExtension(".csv", "text/line-based/*sv");`
  `IM.registerExtension(".tsv", "text/line-based/*sv");`
  `IM.registerExtension(".xml", "text/xml");`
  `IM.registerExtension(".atom", "text/xml");`
  `IM.registerExtension(".json", "text/json");`
  `IM.registerExtension(".js", "text/json");`
  `IM.registerExtension(".xls", "binary/text/xml/xls/xlsx");`
  `IM.registerExtension(".xlsx", "binary/text/xml/xls/xlsx");`
  `IM.registerExtension(".ods", "text/xml/ods");`
  `IM.registerExtension(".nt", "text/rdf/nt");`
  `IM.registerExtension(".ntriples", "text/rdf/nt");`
  `IM.registerExtension(".n3", "text/rdf/n3");`
  `IM.registerExtension(".ttl", "text/rdf/ttl");`
  `IM.registerExtension(".jsonld", "text/rdf/ld+json");`
  `IM.registerExtension(".rdf", "text/rdf/xml");`
  `IM.registerExtension(".marc", "text/marc");`
  `IM.registerExtension(".mrc", "text/marc");`
 ` IM.registerExtension(".wiki", "text/wiki"); `

two or more than two extension contain only one formate
thats why I think  naming depend on formate.
In this code I specify name according to formate

wetneb:
Thanks! But there is still the same issue: extensions will not be able to add new translations for the new formats they support, because they will not be able to edit the "switch" block you introduced.

chetan-v:
I am sorry, I don't understand what you are saying about extensions. and switch block is manual. if any new formate come then we have to add a new case in switch block.

wetneb:
> if any new formate come then we have to add a new case in switch block

That's exactly the issue: extensions which define new importers will not be able to modify this file.

chetan-v:
> > if any new formate come then we have to add a new case in switch block
> 
> That's exactly the issue: extensions which define new importers will not be able to modify this file.

Oh! Thats the point 😄. Yes this is bad side of this code.actually I don't think about it.
Ok then I think something else, Thanks.

PR #3103, opened 2020-08-20
Docs > Transforming data

allanaaa:
Cell editing page is up first. Ignore the "overview" page for now.


allanaaa:
A few things to note:

1) https://deploy-preview-3103--openrefine-docs.netlify.app/manual/cellediting#clustering-methods
I've expanded a bit on the different clustering methods (what used to be on [Clustering in Depth](https://github.com/OpenRefine/OpenRefine/wiki/Clustering-In-Depth)) because I think it will be useful for people to understand at least some details on what each of them does, especially the phonetic ones. It's less total text, but all on the same page. I do open with a "You probably don't need to know this" statement, though. But I didn't really include anything from this point on: https://github.com/OpenRefine/OpenRefine/wiki/Clustering-In-Depth#cluster-new-value-suggestions 
I didn't see much there that would help the average user (it's not really anything operable). I think some of it might be useful in the dev documentation, if this is an area where we invite contributions. We can put an infobox on the user manual page to emphasize the "If you'd like to suggest other methods / language coverage / code changes, etc., we're all ears". If this is more of a "feature request" invitation and less of an invitation to develop for us, then we can link to issues instead of the dev documentation.

2) The fetch errors I mentioned above are now here: https://deploy-preview-3103--openrefine-docs.netlify.app/manual/columnediting#common-errors
I'm just looking for general feedback on this - I think it's fine where it is but others may have thoughts about the scope or the way it's presented. There is some Mac-specific stuff but not really enough to warrant using tabs.

Thanks!

allanaaa:
Also on the topic of clustering - do any extensions add clustering methods? I can't tell from the downloads page but y'all know things I don't.

allanaaa:
Aaand the last thing: you can safely ignore the "Add columns from reconciled values" section for now - I'll likely move it out of Column editing and into the Reconcile page; I'm just working top-to-bottom for the time being. I did shamelessly steal [this gif](https://github.com/OpenRefine/OpenRefine/wiki/Reconciliation#fetching-extra-data-from-the-reconciled-database), so thanks to whoever made that!

tfmorris:
No extensions that I know of add clustering algorithms. They certainly could, but I don't think we're responsible for documenting the behavior of extensions.

Agree that there's no value in the clustering "new value" discussion. The only important thing to mention there (which isn't covered by that text) is that clicking on a value in the cluster both enables the checkbox AND updates the new value to the clicked item's text.

The HTTP errors text is going to need to be updated because I fixed the code that was causing the obscure error messages (it was throwing away the "main" error and only reporting the detailed error cause, which made a lot less sense). Now you should get something more like "error: javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure" instead of just "Received fatal alert: handshake_failure."

docs/docs/manual/cellediting.md
```diff
@@ -0,0 +1,162 @@
...
+
+Your data may come from an HTML-formatted source that expresses some characters through references (such as "&amp;nbsp;" for a space, or "%u0107" for a ć) instead of the actual Unicode characters. You can use the “unescape HTML entities” transform to look for these codes and replace them with the characters they represent. 
+
+### Replace smart quotes with ASCII
+
+Smart quotes (or curly quotes) recognize whether they come at the beginning or end of a string, and will generate an “open” quote (“) and a “close” quote (”). These characters are not ASCII-compliant (though they are UTF8-compliant) so you can use this tranform to replace them with a straight double quote character (") instead. 
+
+### Case transforms
+
+You can transform an entire column of text into UPPERCASE, lowercase, or Title Case using these three options. This can be useful if you are planning to do textual analysis and wish to avoid case-sensitivity (which many functions are) causing problems in your analysis.
```
tfmorris, line 42: Note that this isn't "title case" as a librarian or book editor would do it.

docs/docs/manual/cellediting.md
```diff
@@ -0,0 +1,162 @@
...
+
+Smart quotes (or curly quotes) recognize whether they come at the beginning or end of a string, and will generate an “open” quote (“) and a “close” quote (”). These characters are not ASCII-compliant (though they are UTF8-compliant) so you can use this tranform to replace them with a straight double quote character (") instead. 
+
+### Case transforms
+
+You can transform an entire column of text into UPPERCASE, lowercase, or Title Case using these three options. This can be useful if you are planning to do textual analysis and wish to avoid case-sensitivity (which many functions are) causing problems in your analysis. 
+
+### Data-type transforms
+
+As detailed in [Data types](exploring#data-types), OpenRefine recognizes seven data types: string, number, boolean, date, array, error, and null. When you use these transforms, OpenRefine will check to see if the given values can be converted, then both transform the data in the cells (such as “3” as a text string to “3” as a number) and convert the data type on each successfully transformed cell. Cells that cannot be transformed will output the original value and maintain their original data type.
```
tfmorris, line 46:
```suggestion
As detailed in [Data types](exploring#data-types), OpenRefine recognizes five data types: string, number, boolean, date, and array. It also has an internal data type for storing errors. When you use these transforms, OpenRefine will check to see if the given values can be converted, then both transform the data in the cells (such as “3” as a text string to “3” as a number) and convert the data type on each successfully transformed cell. Cells that cannot be transformed will keep the original value and data type. 
```

docs/docs/manual/cellediting.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Fill down and blank down
+
+Fill down and blank down are two functions most frequently used when encountering data organized into [records](exploring#row-types-rows-vs-records) - that is, multiple rows associated with one specific entity. 
+
+If you receive information in rows mode and want to convert it to records mode, the easiest way is to sort your first column by the value that you want to use as a unique records key, then blank down all the duplicates in that column. OpenRefine will retain the first unique value and erase the rest. Then you can switch from “Show as rows” to “Show as records” and OpenRefine will convert the data based on the remaining values in the first column. Be careful that your data is sorted properly before you begin blanking down - not just the first column but other columns you may want to have in a certain order. For example, you may have multiple identical entries in the first column, one with a value in the second column and one with an empty cell in the second column. In this case you want the value to come first, so that you can clean up empty rows later, once you blank down.
```
tfmorris, line 71: A common error here is not using "Make sort permanent," so it's probably worth mentioning.

docs/docs/manual/columnediting.md
```diff
@@ -0,0 +1,128 @@
...
+
+You can join columns by selecting “Edit column” > “Join columns…”. All the columns currently in your dataset will appear in the pop-up window. You can select or un-select all the columns you want to join, and drag columns to put them in the order you want to join them in. You will define a separator character (optional) and define a string to insert into empty cells (nulls). 
+
+The joined data will appear in the column you originally selected, or you can create a new column based on this join and specify a name. You can delete all the columns that were used in this join operation. 
+
+## Add column based on this column
+
+This selection will open up an [expressions](expressions) window where you can transform the data from this column (using `value`) or write a complex expression that takes information from any number of columns or from external reconciliation sources. 
+
+One common use of this function is to join information from two projects together in one place. Just like referring from one table to another table in an entity-relationship database, you need a key that exists in both.
```
tfmorris, line 39: `cross()` is a pretty advanced function. I think various string manipulation functions are likely much more commonly used.

docs/docs/manual/columnediting.md
```diff
@@ -0,0 +1,128 @@
...
+
+```“https://www.wikidata.org/wiki/Special:EntityData/” + value + “.json”```
+
+or whatever metadata format you prefer. [Information about these Wikidata options can be found here](https://www.wikidata.org/wiki/Wikidata:Data_access).
+
+This service is more useful when getting metadata files instead of HTML files, but you may wish to work with a page’s entire HTML contents and then parse out information from that. Be cautioned that the fetching process can take quite some time and that servers may not want to fulfill hundreds or thousands of page requests in seconds. 
+
+Fetching allows you to set a “throttle delay” which determines the amount of time between requests. The default is 5 seconds per row in your dataset (5000 milliseconds), so you can estimate how long it will take to work through the entire column using that number. We recommend leaving this value at 5000 or greater.
```
tfmorris, line 69: 5 seconds is very long for most web services. They should certainly be polite, but reducing this to 1 second or less if fine for the rate limit on most web services.

docs/docs/manual/columnediting.md
```diff
@@ -0,0 +1,128 @@
...
+![A screenshot of the settings window for fetching URLs.](/img/fetchingURLs.png)
+
+Note the following:
+
+* Many systems prevent you from making too many requests per second. To avoid this problem, set the throttle delay, which tells OpenRefine to wait the specified number of milliseconds between URL requests.
+* Before pressing OK, copy/paste a URL or two from the right column in the dialog and test them in another browser tab to make sure they work.
+* In some situations you may need to set[ HTTP request headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers). To set these, click the small “Show” button next to "HTTP headers to be used when fetching URLs" in the settings window. You can set the following request headers:
+  * [User-Agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent)
+  * [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept)
+  * [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization)
```
tfmorris, line 80: We should warn users that the authorization credentials get logged in the operation history in plain text, representing a security concern that they need to be aware of.

docs/docs/manual/columnediting.md
```diff
@@ -0,0 +1,128 @@
...
+
+First, make sure that your fetching operation is storing errors (check “store error”). The run the fetch and look at the error messages. 
+
+**"Received fatal alert: handshake_failure" **can occur when you are trying to retrieve information over HTTPS but the remote site is using an encryption not supported by the Java virtual machine being used by OpenRefine.
+
+You can check which encryption methods are supported by your OpenRefine/Java installation by using a service such as **How's my SSL**. Add the URL `https://www.howsmyssl.com/a/check` to an OpenRefine cell and run "Add column by fetching URLs" on it, which will provide a description of the SSL client being used. 
+
+You can try installing additional encryption supports by installing the [Java Cryptography Extension](https://www.oracle.com/java/technologies/javase-jce8-downloads.html). Note on OpenRefine for Mac, these updated cipher suites need to be dropped into the Java install within the OpenRefine application: something like `/Applications/OpenRefine.app/Contents/PlugIns/jdk1.8.0_60.jdk/Contents/Home/jre/lib/security`.
```
tfmorris, line 92: In addition to Mac, also Windows users who are using the OpenRefine kit with the bundled JRE.

I'm not sure this merits such a prominent mention though. Historically when this happens there is a cluster of problems for a lot of users, but it's a relatively infrequent occurrence.

docs/docs/manual/columnediting.md
```diff
@@ -0,0 +1,128 @@
...
+
+The list of trusted certificates is stored in an encrypted file called `cacerts` in your local Java installation. This can be read and updated by a tool called “keytool.” You can find directions on how to add a security certificate to the list of trusted certificates for a Java installation [here](http://magicmonster.com/kb/prg/java/ssl/pkix_path_building_failed.html) and [here](http://javarevisited.blogspot.co.uk/2012/03/add-list-certficates-java-keystore.html).
+
+Note on OpenRefine for Mac, you need to update the `cacerts` file within the OpenRefine application: something like `/PlugIns/jdk1.8.0_60.jdk/Contents/Home/jre/lib/security/cacerts`.
+
+**"HTTP error 403 : Forbidden" **can be simply down to you not having access to the URL you are trying to use. If you can access the same URL with your browser, the remote site may be blocking OpenRefine because it doesn't recognize its request as valid. Changing the[ User-Agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) request header may help. If you believe you should have access to a site but are “forbidden,” you may wish to contract the administrators.
```
tfmorris, line 100: 4xx errors are going to be the most common, by far, so they should go first, followed by 5xx, then everything else.

docs/docs/manual/columnediting.md
```diff
@@ -0,0 +1,128 @@
...
+
+*   If the reconciliation service supports the[ Data Extension API](https://github.com/OpenRefine/OpenRefine/wiki/Data%20Extension%20API), then you can use the "Add columns from reconciled values" action to augment your OpenRefine project with new columns. For example, if you have a column of chemical elements identified by name, you can fetch categorical information about them such as their atomic number and their element symbol, as the animation shows below.
+
+![A screenshare of elements fetching related information.](/img/reconcileelements.gif)
+
+*   If the reconciliation service does not support this feature, look out for a generic web API for that data source. You can use the "Add column by fetching URLs" operation to call this API with the IDs obtained from the reconciliation process. For instance, if you have used the OpenCorporates service, you can use[ their API](https://api.opencorporates.com/documentation/API-Reference) to fetch JSON-formatted data about the reconciled companies. The GREL expression `"https://api.opencorporates.com/companies/"+cell.recon.match.id` will be translated to URLs that can then be parsed in OpenRefine.
```
tfmorris, line 122: It might be better to use a non-commercial example like VIAF here (plus, we probably have more librarian users)

docs/docs/manual/cellediting.md
```diff
@@ -0,0 +1,162 @@
...
+    *   cologne-phonetic
+    *   Daitch-Mokotoff
+    *   Beider-Morse
+*   nearest neighbor
+    *   levenshtein
+    *   ppm
+
+**Key collisions** are very fast and can process millions of cells in seconds:
+
+**Fingerprinting** is the least likely to produce false positives, so it’s a good place to start. It does the same kind of data-cleaning behind the scenes that you might think to do manually: fix whitespace into single spaces, put all uppercase letters into lowercase, discard punctuation, convert special characters to their ASCII equivalent, split all strings (words) and reorganize them alphabetically (so “Zhenyi, Wang” becomes “Wang Zhenyi”). This makes comparing those types of name values very easy.
```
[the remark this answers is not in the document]
allanaaa: I'm not sure "diacritics" is much better than "ASCII" but you're right; I'll work on it.

docs/docs/manual/cellediting.md
```diff
@@ -0,0 +1,162 @@
...
+
+**N-gram fingerprinting** allows you to set the _n_ value to whatever number you’d like, and will create n-grams of _n_ size (after doing some cleaning), alphabetize them, then join them back together into a _fingerprint_. For example, a 1-gram fingerprint will simply organize all the letters in the cell into alphabetical order - by creating grams with one character length. A 2-gram fingerprint will find all the two-character grams, remove duplicates, alphabetize them, and join them back together. This can help match cells that have typos, or incorrect spaces (such as matching “lookout” and “look out,” which fingerprinting itself won’t identify). The higher the _n_ value, the fewer clusters will be identified. With 1-grams, keep an eye out for mismatched values that are near-anagrams of each other (such as “Wellington” and “Elgin Town”).
```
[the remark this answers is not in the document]
allanaaa: No, it appears to be just a word in my head! 

New version: 

> **N-gram fingerprinting** allows you to set the _n_ value to whatever number you’d like, and will create n-grams of _n_ size (after doing some cleaning), alphabetize them, then join them back together into a _fingerprint_. For example, a 1-gram fingerprint will simply organize all the letters in the cell into alphabetical order - by creating segments one character in length. A 2-gram fingerprint will find all the two-character segments, remove duplicates, alphabetize them, and join them back together (for example, “banana” generates “ba an na an na,” which becomes “anbana”). This can help match cells that have typos, or incorrect spaces (such as matching “lookout” and “look out,” which fingerprinting itself won’t identify). The higher the _n_ value, the fewer clusters will be identified. With 1-grams, keep an eye out for mismatched values that are near-anagrams of each other (such as “Wellington” and “Elgin Town”).

docs/docs/manual/cellediting.md
```diff
@@ -0,0 +1,162 @@
...
+
+**Levenshtein distance** counts the number of operations required to make one value perfectly match another. As in the key collision methods above, it will do things like change uppercase to lowercase, fix whitespace, change special characters, etc. Each character that gets changed counts as 1 “distance.” “New York” and “newyork” have an edit distance value of 3 (“N” to “n”, “Y” to “y,” remove the space). It can do relatively advanced operations, such as understand the distance between “M. Makeba” and “Miriam Makeba” (5), but it may create false positives if these distances are greater than other, simpler transformations (such as the one-character distance to “B. Makeba,” another person entirely).
```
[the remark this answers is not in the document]
allanaaa: "Edits" it is! I don't think it needs a link but if you have one or two to suggest I'd like to see them.

docs/docs/manual/cellediting.md
allanaaa: Apparently we don't want to include "array" in this list either, according to last week's conversation about data types. This whole sentence will be cleaned up.

docs/docs/manual/columnediting.md
allanaaa: Is "leaving this value at 1000 or greater" okay? Or should we provide some guidance on how to calculate this?

Would it also be worth changing the default in the program?

docs/docs/manual/columnediting.md
allanaaa: Yes, I wondered the same - not having anecdotal experience with any of these errors, I'm glad someone piped up. 

Having reordered with the 4xx/5xx errors first, can any of these be eliminated from the list as no longer being "common"?

docs/docs/manual/columnediting.md
allanaaa: In our meeting today we agreed, so there's an issue for this now: https://github.com/OpenRefine/OpenRefine/issues/3127

In terms of guidance we'll simply cut that "recommend" sentence.

docs/docs/manual/columnediting.md
allanaaa: I started with VIAF but I didn't love their metadata formats as a concrete example. I went with the Getty ULAN instead. I would love some feedback on this choice though:

https://deploy-preview-3103--openrefine-docs.netlify.app/manual/reconciling#add-columns-by-fetching-urls

> If the reconciliation service cannot extend data, look for a generic web API for that data source, or a structured URL that points to their dataset entities via unique IDs (such as https://viaf.org/viaf/000000). You can use the "Edit column" → "Add column by fetching URLs" operation to call this API or URL with the IDs obtained from the reconciliation process. This will require using GREL expressions.
> 
> You will likely not want to pull the entire HTML content of the pages at the ends of these URLs, so look to see whether the service offers a metadata endpoint, such as JSON-formatted data. You can either use a column of IDs, or you can pull the ID from each matched cell during the fetching process.
> 
> For example, if you have reconciled artists to the Getty's ULAN, and have their unique ULAN IDs as a column, you can generate a new column of JSON-formatted data by using "Add column by fetching URLs" and entering the GREL expression `"http://vocab.getty.edu/" + value + ".json"` in the window. For this service, the unique IDs are formatted "ulan/000000" and so the generated URLs look like "http://vocab.getty.edu/ulan/000000.json."
> 
> You can use the Add entity identifiers column operation above if you don't already have a column of unique IDs.
> 
> You can alternatively insert the ID directly from the matched column using a GREL expression like `"http://vocab.getty.edu/" + cell.recon.match.id + ".json"` instead.
> 
> Remember to set an appropriate throttle and to refer to the service documentation to ensure your compliance with their terms.

allanaaa:
I plan to return to the Transposing page to even out some of the language inconsistencies between what @wetneb  wrote and my own contributions. I really like what's there, just want to blend it a bit better.

As for Reconciling, I think it's still a bit rough but my eyes have begun to cross when reading it over. 
In particular I'd like to discuss:
* Are there any more places where I can insert reference to the properties in the [testbench table](https://reconciliation-api.github.io/testbench/)? For example, see where I've mentioned the "suggest entities" property: https://deploy-preview-3103--openrefine-docs.netlify.app/manual/reconciling#reconciliation-actions
   * In particular, I'm not sure whether "suggest types" means it will both load a list of types when you select the service, based on sampling your data, and load suggestions when you start typing in the "Reconcile against type" box. Or is it one or the other only? https://deploy-preview-3103--openrefine-docs.netlify.app/manual/reconciling#reconciling-by-type
* I plan to replace the Wikidata example in this section, but I'm not actually having any success getting other services with the data extension API to work. https://deploy-preview-3103--openrefine-docs.netlify.app/manual/reconciling#add-columns-from-reconciled-values (the image below is as far as I get with GND - selecting one of those options brings up an error). I do just need a gif though, so if anyone can get it working and wants to send me one, that would be lovely. 
* Whether the examples I use throughout this page should be switched up a bit more (though it's been largely dependent on getting stuff working so far). Are there any major services I should mention at least once?

docs/docs/manual/columnediting.md
allanaaa: FYI to all, these are the current error messages listed - this is your chance to let me know if anything needs to be removed or added: 

"HTTP error 403 : Forbidden" 
"HTTP error 404 : Not Found" 
"HTTP error 500 : Internal Server Error"
“error: javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure”
"sun.security.validator.ValidatorException: PKIX path building failed"

thadguidry: You mean just for the **Fetch URLs** operation however, right?  because there's also about 250+ error message kinds for Transforming data that could get potentially stored in a cell depending on if you checked the radio button to store the error. ;-)

wetneb:
> In particular, I'm not sure whether "suggest types" means it will both load a list of types when you select the service, based on sampling your data, and load suggestions when you start typing in the "Reconcile against type" box. Or is it one or the other only? https://deploy-preview-3103--openrefine-docs.netlify.app/manual/reconciling#reconciling-by-type

Yes "suggest types" is only used for the "Reconcile against type" input that you have in your screenshot. The list of types above is obtained by running regular reconciliation queries on a sample of the data, so it does not depend on whether the service offers a suggest endpoint for types or not.

> I plan to replace the Wikidata example in this section, but I'm not actually having any success getting other services with the data extension API to work. https://deploy-preview-3103--openrefine-docs.netlify.app/manual/reconciling#add-columns-from-reconciled-values (the image below is as far as I get with GND - selecting one of those options brings up an error). I do just need a gif though, so if anyone can get it working and wants to send me one, that would be lovely.

If data extension does not work for GND, that is a bug, so feel free to open an issue about this (either in this repo or theirs at https://github.com/hbz/lobid-gnd, and depending on what is causing the issue we will move it to the appropriate location).

> Whether the examples I use throughout this page should be switched up a bit more (though it's been largely dependent on getting stuff working so far). Are there any major services I should mention at least once?

I think you got the balance right!

docs/docs/manual/reconciling.md
```diff
@@ -0,0 +1,192 @@
...
+
+You can fill in the “As Property” field with the type of information you are including - when you start typing, potential fields may pop up, such as “birthDate” in the case of ULAN or "Geburtsdatum" in the case of Integrated Authority File (GND). Use the documentation for your chosen service to identify the fields in their terms. Some services will not be able to search with your "As Property" entry, but you can still manually supply the field name. Refer to the service to make sure you enter it correctly. 
+
+![Including a birth-date type.](/img/reconcile-by-type.png)
+
+If your data formatting does not exactly match that of the service you are reconciling, you may be able to use asterisks (&#42;) as wildcards to encourage more matches. [This example from the GND](https://blog.lobid.org/2019/09/30/openrefine-examples.html) uses wildcards to match a column of birth years (YYYY&#42;) to birth dates formatted as YYYY-MM-DD.
```
wetneb, line 138: I would remove this part, since it is specific to this service.

docs/docs/manual/reconciling.md
```diff
@@ -0,0 +1,192 @@
...
+*   [Add entity identifiers column](#add-entity-identifiers-column).
+
+## Reconciling with unique identifiers
+
+Reconciliation services use unique identifiers for their entities. For example, the 14th Dalai Lama has the VIAF ID [38242123](https://viaf.org/viaf/38242123/) and the Wikidata ID [Q17293](https://www.wikidata.org/wiki/Q37349). If you can supply the correct identifiers directly to your chosen reconciliation service, reconciliation will be a fully automated process. 
+
+Select the column with unique identifiers and apply the operation “Reconcile” → “Use values as identifiers.” This will bring up the list of reconciliation services you have already added (to add a new service, open the “Start reconciling…” window first). If you reconcile a column of IDs, other columns cannot be included.
+
+Matching identifiers does not actually "reconcile," that is, query the external service to identify matches. All cells will appear as dark blue "confirmed" matches. You may get false positives, which you will need to hover over or click on to identify:
```
wetneb, line 104:
```suggestion
Matching identifiers does not actually validate identifiers. All cells will appear as dark blue "confirmed" matches. You should check before this operation that the identifiers in the column exist on the target service.
```

allanaaa:
I've popped out Reconciling from under the "Transforming data" category to a standalone entry in the table of contents. 

I've also (tentatively) moved the Wikidata page up, though I know there are arguments for putting it back down at the bottom. I'll be tackling the Wikidata page next so we'll discuss that soon enough!

I've added a section about reconciliation variables/expressions with some examples but now I feel like it's a bit too long:
https://deploy-preview-3103--openrefine-docs.netlify.app/manual/reconciling/#writing-reconciliation-expressions
Maybe it's "good enough for now" and we can cut it down in the second pass.

allanaaa:
This last edit includes, re #3192 : 

> An error that includes **“javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed”** can occur when you try to retrieve information over HTTPS but the remote site is using a certificate not trusted by your local Java installation. You will need to make sure that the certificate, or (more likely) the root certificate, is trusted. 

It also rewords the Columnize by key/value section to flow more consistently with the rest of the docs.

As far as I can tell there are no outstanding change requests on this PR - if that error message looks okay and nobody spots anything else, I think we're good to move on for now.

wetneb (approved):

PR #3104, opened 2020-08-21
internationalised the label using $.i18n

### LINKED ISSUE #3069: Importer names not internationalized (i18n)

chetan-v:
Fixes #3069


main/webapp/modules/core/scripts/index/default-importing-controller/parsing-panel.js
```diff
@@ -141,8 +141,9 @@ Refine.DefaultImportingController.prototype._prepareParsingPanel = function() {
 
   var formats = this._job.config.rankedFormats;
   var createFormatTab = function(format) {
+    var formateNames =Refine.importingConfig.formats[format].label;
```
tfmorris, line 144: There's a typo in `formateNames`, but rather than just fixing that, let's use a variable name which indicates what it contains.
```suggestion
    var formatLabelKey =Refine.importingConfig.formats[format].label;
```
We'll leave `format.label` unchanged for compatibility.

main/webapp/modules/core/scripts/index/default-importing-controller/parsing-panel.js
```diff
@@ -141,8 +141,9 @@ Refine.DefaultImportingController.prototype._prepareParsingPanel = function() {
...
     var tab = $('<div>')
-    .text(Refine.importingConfig.formats[format].label)
+    .text( $.i18n(formateNames))
```
tfmorris, line 146:
```suggestion
    .text( $.i18n(formatLabelKey))
```

tfmorris (dismissed):
Looks good! Just one tiny change and it'll be ready to merge.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -749,5 +749,23 @@
...
+    "core-import-formats/text/rdf/ttl": "RDF/Turtle files",
+    "core-import-formats/text/rdf/xml": "RDF/XML files",
+    "core-import-formats/text/rdf/ld+json": "JSON-LD files",
+    "core-import-formats/text/xml": "XML files",
+    "core-import-formats/binary/text/xml/xls/xlsx": "Excel files",
+    "core-import-formats/text/xml/ods": "Open Document Format spreadsheets (.ods)",
+    "core-import-formats/text/json": "JSON files",
+    "core-import-formats/text/marc": "MARC files",
+    "core-import-formats/text/wiki": "Wikitext",
+    "core-import-formats/text/line-based/pc-axis": "PC-Axis text files",
```
wetneb, line 767: The `text/line-based/pc-axis` format is defined in the `pc-axis` extension, so this translation message should be in that extension, not in the main Butterfly module.

wetneb (changes requested):

main/webapp/modules/core/langs/translation-en.json
chetan-v: Then can I make new folder of name langs in extensions/pc-axis/module/langs

chetan-v: its not a good idea

wetneb: Why is it not a good idea?

chetan-v: yes we can do that

tfmorris:
@chetan-v My apologies for jumping in the middle. I'll let you and @wetneb work out whatever else needs to be done before this gets merged.

main/webapp/modules/core/langs/translation-en.json
chetan-v: Screenshot from 2020-08-28 12-56-22

wetneb: Thanks! You will need to load the translations from the extension. You can just do like other extensions do:

https://github.com/OpenRefine/OpenRefine/blob/master/extensions/wikidata/module/scripts/menu-bar-extension.js#L2-L16

https://github.com/OpenRefine/OpenRefine/blob/master/extensions/database/module/scripts/index/database-import-controller.js#L34-L46

wetneb:
@tfmorris it's totally fine (and actually helpful!)

chetan-v:
@tfmorris apologies for what? your review is helpful for making my code more understandable to others.

chetan-v:
Hi, @wetneb @tfmorris 
Can anyone tell me what else to add.

tfmorris (commented):
I'm going to let @wetneb handle the review for this so that there is consistency.

wetneb:
Sorry for the delay in merging this!
It looks great to me now, except for a small detail: `core-import-formats` does not feel appropriate for an extension such as `pc-axis` since by definition the importer defined there is not part of the "core" package.

chetan-v:
> Sorry for the delay in merging this!

That's fine

> It looks great to me now, except for a small detail: `core-import-formats` does not feel appropriate for an extension such as `pc-axis` since by definition the importer defined there is not part of the "core" package.

Yes, 
So I do something other extensions do

wetneb:
That sounds like a great idea!

wetneb (approved):
Thanks very much again!

PR #3105, opened 2020-08-21
Fix pow function 

### LINKED ISSUE #3062: pow() does not throw error for extra or incorrect arguments like quotient() does
### To Reproduce
1. Open Expression Editor on existing project
2. On any column use Custom Text Transform with `value.pow(2,8)` 
3. Notice the preview result of `null`
4. Change the expression to `value.quotient(2,8)`
5. Notice the preview result of `Error: quotient expects two numbers` 

### Current Results
`null` is shown in preview result, instead of an error (since `value` is passed as one of the arguments with additional 2 arguments `(2,8)` supplied)

### Expected Behavior
`pow()` should check arguments like `quotient()` does and show similar error.

### Screenshots
image

image

image

image

### Versions
 - JRE or JDK Version: JDK8
 - OpenRefine: openrefine-3.4-beta-458-g109aa78

chetan-v:
Fixes #3062

### DIFF
--- a/main/src/com/google/refine/expr/functions/math/Pow.java
+++ b/main/src/com/google/refine/expr/functions/math/Pow.java
@@ -35,19 +35,28 @@ package com.google.refine.expr.functions.math;
 
 import java.util.Properties;
 
+import com.google.refine.expr.EvalError;
+import com.google.refine.grel.ControlFunctionRegistry;
 import com.google.refine.grel.Function;
+
+
+
 
 public class Pow implements Function {
 
     @Override
     public Object call(Properties bindings, Object[] args) {
-        if (args.length == 2 && args[0] instanceof Number && args[1] instanceof Number) {
+        if((args.length != 2)){
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + "  expects two numbers");  
+        }
+        else if(! (args[1] instanceof Number) || ! (args[0] instanceof Number)){
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + "  arguments is not an instance of number");
+        }else{
             return Math.pow(
                 ((Number) args[0]).doubleValue(),
                 ((Number) args[1]).doubleValue()
             );
         }
-        return null;
     }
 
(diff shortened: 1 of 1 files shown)


main/src/com/google/refine/expr/functions/math/Pow.java
```diff
@@ -35,19 +35,22 @@
...
+import com.google.refine.grel.ControlFunctionRegistry;
 import com.google.refine.grel.Function;
 
 public class Pow implements Function {
 
     @Override
     public Object call(Properties bindings, Object[] args) {
-        if (args.length == 2 && args[0] instanceof Number && args[1] instanceof Number) {
+        if (args.length == 2 && args[0] != null && args[0] instanceof Number
+        && args[1] != null && args[1] instanceof Number) {
```
tfmorris, line 47: The null checks are redundant. `null` will never be an `instanceof Number`

chetan-v: Can I use try and catch block for it.

chetan-v: Hi @tfmorris before as you said about inconsistency error handling I did something which is not the proper solution but it give some relevant output
```
  if((args.length != 2)){
        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + "  expects two numbers");  
        }
        else if(! (args[1] instanceof Number) || ! (args[0] instanceof Number)){
        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + "   arguments is not an instance of number");
     }else{
            return Math.pow(
                ((Number) args[0]).doubleValue(),
                ((Number) args[1]).doubleValue()
                 );
        }
```
Screenshot from 2020-08-30 21-16-34
Screenshot from 2020-08-30 21-16-47
Screenshot from 2020-08-30 21-17-07

tfmorris:
There's been no activity on this for a while, so I'm going to close it, but feel free to reopen it or create a new pull request when the tests pass.

PR #3112, opened 2020-08-22
Add test for changed Apache unescape("csv") behavior - Fixes #2997

### LINKED ISSUE #2997: Document and test new behavior of unescape('csv')
#2995 includes a bug fix (https://github.com/apache/commons-text/pull/119) to Apache commons-text which changes the behavior of `unescape('csv')` which apparently is not covered by our test suite since it didn't cause a failure. We should add a test and document the behavior change for the release notes.

tfmorris:
Fixes #2997 

Refactor tests
* Hoist `invoke()` utility to superclass and eliminate redundant copies
* Split up legacy multifunction test modules and distribute their tests to the specific per-function test modules
* Remove unused imports

Adds tests for a number of string functions, including the `unescape("csv")` function reference in the original ticket.


tfmorris:
> Thanks for tackling this duplication of `invoke`, it was embarrassing indeed.

No problem. I wouldn't have invested the effort with emacs, but with modern IDE refactoring tools it was pretty easy.

PR #3159, opened 2020-09-03
Update openrefine-api.md

rajaram5:
Documentations for `Create project` and `Export rows` API calls are modified. When creating a project based on `json input file` it seems like `options` parameter is mandatory. For `Export rows` call both '`project` and `format` parameters should be provided as a request parameters

### DIFF
--- a/docs/docs/technical-reference/openrefine-api.md
+++ b/docs/docs/technical-reference/openrefine-api.md
@@ -17,7 +17,7 @@ multipart form-data:
       'project-file' : file contents
       'project-name' : project name
       'format' : format of data in project-file (e.g. 'text/line-based/*sv') [optional]
-      'options' : json object containing options relevant to the file format [optional]. However for JSON input file this parameter is [Mandatory]
+      'options' : json object containing options relevant to the file format [optional - however, some importers may have required options, such as `recordPath` for the JSON & XML importers].
 
 The formats supported will depend on the version of OpenRefine you are using and any Extensions you have installed. The common formats include:
 
--- a/docs/docs/technical-reference/openrefine-api.md
+++ b/docs/docs/technical-reference/openrefine-api.md
@@ -17,7 +17,7 @@ multipart form-data:
       'project-file' : file contents
       'project-name' : project name
       'format' : format of data in project-file (e.g. 'text/line-based/*sv') [optional]
-      'options' : json object containing options relevant to the file format [optional]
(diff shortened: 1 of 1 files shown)


docs/docs/technical-reference/openrefine-api.md
```diff
@@ -17,7 +17,7 @@ multipart form-data:
       'project-file' : file contents
       'project-name' : project name
       'format' : format of data in project-file (e.g. 'text/line-based/*sv') [optional]
-      'options' : json object containing options relevant to the file format [optional]
+      'options' : json object containing options relevant to the file format [optional]. However for JSON input file this parameter is [Mandatory]
```
tfmorris, line 20:
```suggestion
      'options' : json object containing options relevant to the file format [optional - however, some importers may have required options, such as `recordPath` for the JSON & XML importers].
```

docs/docs/technical-reference/openrefine-api.md
```diff
@@ -159,9 +159,14 @@ On success returns JSON response
...
-      'project' : project id
-      'engine' : JSON string... (e.g. '{"facets":[],"mode":"row-based"}')
+In the parameter
+
+      'project' : project id      
       'format' : format... (e.g 'tsv', 'csv')
+
+In the form data
+
+      'engine' : JSON string... (e.g. '{"facets":[],"mode":"row-based"}')
```
tfmorris, line 169: As far as I know, parameters can be passed either in the query string or in form data and the API doesn't care. By convention, the Engine config data is usually sent as form data because of its size, but that's just a convention.

Did you run into a case where the parameters could only be passed via on mechanism or the other?

tfmorris (changes requested):
Thanks for the contribution! I made a suggestion for generalizing the first edit since it definitely applies to the XML importer as well (and there may be other required parameters for other importers).

For the second proposed edit, I'd like to understand better what caused you to come to this conclusion.

Thanks again for helping to improve the docs! 
(tagging @allanaaa so that this is on the radar for the new docs too)

docs/docs/technical-reference/openrefine-api.md
rajaram5: For me export-rows API call doesn't work if I pass one of these parameters (format, project) via form data. It works only if I pass these parameters as request parameters. I don't have any issues with engine parameter it works for me either way.

wetneb: There are indeed cases where parameters have to be in the URL or in the POST body.

wetneb (approved):

PR #3166, opened 2020-09-06
Adds useful Clojure dependencies

thadguidry:
- Adds clojure.java-time for convenience working with Java 8 Date-Time from Clojure
- Adds tick for advanced date-time manipulation
- Adds cuerdas for String manipulation (similar to GREL functions in Clojure)
- Adds kixi.stats for statistical sampling and transducing

But **what I really want** is somehow to not even bake these Clojure dependencies into OpenRefine's dependencies but instead allow dynamic Clojure dependencies via a `deps.edn` file and whenever I restart OpenRefine it will have pulled down those dependencies?  Or perhaps it can work without restarting OpenRefine, and it downloads the dependencies and adds them to the classpath automatically.

Details: https://clojure.org/guides/deps_and_cli

### DIFF
--- a/main/pom.xml
+++ b/main/pom.xml
@@ -355,7 +355,26 @@
       <artifactId>juniversalchardet</artifactId>
       <version>2.3.2</version>
     </dependency>
-
+    <dependency>
+      <groupId>clojure.java-time</groupId>
+      <artifactId>clojure.java-time</artifactId>
+      <version>0.3.2</version>
+    </dependency>
+    <dependency>
+      <groupId>tick</groupId>
+      <artifactId>tick</artifactId>
+      <version>0.4.26-alpha</version>
+    </dependency>
+    <dependency>
+      <groupId>funcool</groupId>
+      <artifactId>cuerdas</artifactId>
+      <version>2020.03.26-3</version>
+    </dependency>
+    <dependency>
+      <groupId>kixi</groupId>
+      <artifactId>stats</artifactId>
+      <version>0.5.4</version>
+    </dependency>
     <!-- test dependencies -->
 
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -230,6 +230,10 @@
         <enabled>true</enabled>
       </snapshots>
     </repository>
+    <repository>
+      <id>clojars.org</id>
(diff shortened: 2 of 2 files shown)


tfmorris:
What issue does this address? Also, ss there an issue for the related enhancement request that's embedded at the end?

thadguidry:
@tfmorris Hi Tom!  Yes I'm composing the issue... I wanted to just publish this as a thought and we can totally blow this PR away or improve it.  Oh crap, it's not draft!  Sorry, my mistake!  This is how I have to currently do things to get Clojure jars on the classpath.  Yes as mentioned at the end, I'd love to do this dynamically, but don't know if its possible?  That's the issue I am about to send.

wetneb:
I would prefer to come up with a system for users to add dependencies themselves. Adding them as build dependencies of OpenRefine itself will make our packages needlessly bigger and these dependencies will be flagged by analysis tools (such as `mvn dependency:analyze` as unused, so I don't think this is a durable solution. Also, the availability of these packages is not discoverable easily for users.

Overall, we should review the Clojure expression language as a whole for security - if the user is able to call any function from any library on the classpath, it feels pretty dangerous. We should ideally have a controlled environment where we know what we expose to the user. This would also help with stability of expression evaluation (since updating one of our dependencies could break Clojure expressions which rely on it).

thadguidry:
@wetneb agree it would be nice for users to add dependencies themselves. That should be a goal.
It's no more or less dangerous than the exact thing that can be done with Python as well pulling down libraries and using them, which @ettorerizza loves when I found out how to easily do it through[ sys.path](https://github.com/OpenRefine/OpenRefine/wiki/Extending-Jython-with-pypi-modules).

But if we take a step back and look at those few Clojure libraries I mentioned... it's those kinds of functions that are gaps in OpenRefine currently.  We could say that we don't have bandwidth to continue to grow GREL...so instead we help users with easier mechanisms to extend the Expression language.

I'd love to improve the date handling for GREL, in particular helping scientists deal with intervals more (https://cljdoc.org/d/tick/tick/0.3.1/api/tick.interval) was the actual use case for this issue.  I'll focus on that in a new issue, once I collect additional feedback from a few groups.  Stay tuned.

Issue #3163, opened 2020-09-05
Add column containing entity identifiers does not allow ENTER key press


thadguidry:
Pressing ENTER key has no effect on dialog for adding a new column containing entity identifers (from the Reconcile menu)

### To Reproduce
1. Reconcile a few cells in a column  (already done, see attached OpenRefine project to import)
2. On Species column, choose Reconcile -> Add entity identifiers column
3. Type in a new column name into the input box and press ENTER

### Current Results
nothing more happens

### Expected Behavior
I was expecting that the ENTER key was bound to the OK button and would close the dialog and my new column would appear.

### Versions
 - Operating System: Windows
 - Browser Version: Firefox latest
 - JRE or JDK Version: JDK8
 - OpenRefine: openrefine-3.4-beta-514-gaa43445

### Datasets
[SCOR-Data.openrefine.tar.gz](https://github.com/OpenRefine/OpenRefine/files/5178053/SCOR-Data.openrefine.tar.gz)

sjathin:
I would like to work on this issue.

thadguidry:
@sjathin Sure!

PR #3183, opened 2020-09-11
Fix for issue #3163, Calling the okbutton.click() on press of enter

sjathin:
Add a event listener on press of enter key on the column name.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/add-q-column-dialog.html
+++ b/main/webapp/modules/core/scripts/views/data-table/add-q-column-dialog.html
@@ -1,4 +1,5 @@
 <div class="dialog-frame" style="width: 700px;">
+<form action="javascript:okButton;">
   <div class="dialog-border">
     <div class="dialog-header" bind="dialogHeader"></div>
     <div class="dialog-body" bind="dialogBody">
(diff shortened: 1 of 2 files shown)


sjathin:
I have attached a pdf of my testing.
[OpenRefine#3163.pdf](https://github.com/OpenRefine/OpenRefine/files/5209878/OpenRefine.3163.pdf)
Thanks

wetneb (commented):
Thanks! I think it would be cleaner to make sure the form is submitted properly when enter is pressed. You can do so by ensuring that we have an `<input type="submit" />` inside a `<form>`, and that the `<input>` field is in the same `<form>`.

sjathin:
@wetneb , I have a question, currently, there is a bind on the ok button, is there a way if this can be done for a form submit ??

wetneb:
You can have the `bind` attribute and the `type="submit"` on the same button, that should not be a problem.

sjathin:
I did try that, I added bind on type="submit", but the function is not called, the query parameters in the URL gets empty.

wetneb:
You will also need to bind the appropriate Javascript function to the form's submit event:
https://api.jquery.com/submit/

sjathin:
It works !
I had missed the <form action="javascript:okButton;">

main/webapp/modules/core/scripts/views/data-table/add-q-column-dialog.html
```diff
@@ -1,4 +1,5 @@
 <div class="dialog-frame" style="width: 700px;">
+<form action="javascript:okButton;">
```
wetneb, line 2: I'd prefer using jQuery's submit method to bind the event to the form, rather than an explicit action (since that relies on global naming).

sjathin:
@wetneb ,
I'm sorry for these many modification, as I am new to JQuery, is this what you were expecting ?

wetneb (approved):
This is perfect, thank you very much for this first PR!

PR #3203, opened 2020-09-22
Docs > Wikidata

allanaaa:
The first part of this draft is substantially pulled from the [wiki docs on the subject](https://github.com/OpenRefine/OpenRefine/wiki/Reconciliation#wikidata-specific). I believe it can be edited down further.

The second part of this draft is somewhat redundant to what's on the [Wikidata help pages for OpenRefine](https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine). So, I'm sure they can also be edited down, although I am biased and prefer my own version.


docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,251 @@ sidebar_label: Wikidata
 
 ## Overview
 
+OpenRefine includes powerful ways to both pull data from Wikidata and add data to it. 
 
+OpenRefine’s connections to Wikidata were formerly an optional extension, but are now installed automatically with the downloadable package. The Wikidata extension can be removed manually by navigating to your OpenRefine installation folder, and then looking inside `webapp/extensions/` and deleting the `Wikidata` folder inside.
```
ostephens, line 11: The folder should be named lower case `wikidata`

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,251 @@ sidebar_label: Wikidata
...
-*   Former extension, now core [but you can remove it].
-*   Account required
+You do not need a Wikidata account to reconcile your local OpenRefine project to Wikidata. If you wish to upload your cleaned dataset to Wikidata, you will need to register an account, and authorize OpenRefine with that account. 
 
+The best source for information about how OpenRefine works with Wikidata is [on Wikidata itself, under Tools](https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine). This has tutorials, guidelines on editing, and spaces for discussion and help.
 
-## Features
+## Reconciling with Wikidata
 
+Wikidata is, at this time, the most powerful [reconciliation service](reconciling) for OpenRefine. [It supports](https://reconciliation-api.github.io/testbench/):
```
ostephens, line 19: To avoid us having to update this (and have any discussions about what "most powerful" means in this context), can we just state the functions it supports?

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,251 @@ sidebar_label: Wikidata
...
+*   A large number of potential types to reconcile against
+*   Previewing and viewing entities
+*   Suggesting entities, types, and properties
+*   Augmenting your dataset with more information pulled from Wikidata
```
ostephens, line 23: Should we clarify that this means augmenting your dataset within OpenRefine? Or is that obvious from context?

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,251 @@ sidebar_label: Wikidata
...
-*   Issue detection
+For the most part, Wikidata reconciliation behaves the same way other reconciliation processes do, but there are a few processes and features specific to Wikidata. 
 
+### Language settings
 
-## More resources
+You can install a version of the Wikidata reconciliation service that uses your language.
 
-[https://wdreconcile.toolforge.org/](https://wdreconcile.toolforge.org/) 
+First, you need your language code: this is the two-letter code in the domain name of your Wikipedia (for instance "fr" if your Wikipedia is [https://fr.wikipedia.org/wiki/](https://fr.wikipedia.org/wiki/)).
```
ostephens, line 33: Is there any other way of getting the language code? A list would be an easier reference point (and would avoid mentioning Wikipedia here which could be potentially confusing in the context of wikidata)

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,251 @@ sidebar_label: Wikidata
...
+
+### Restricting matches by type
+
+In Wikidata, types are items themselves. For instance, the [university of Ljubljana (Q1377)](https://www.wikidata.org/wiki/Q1377) has type [public university (Q875538)](https://www.wikidata.org/wiki/Q875538), which is witnessed by a statement on [Q1377](https://www.wikidata.org/wiki/Q1377) using the [instance of (P31)](https://www.wikidata.org/wiki/Property:P31) property.
+
+Types are organized into an ontology, by specifying which types are subclasses of the others, using the [subclass of (P279)](https://www.wikidata.org/wiki/Property:P279) property. For instance, [public university (Q875538)](https://www.wikidata.org/wiki/Q875538) is a subclass of [university (Q3918)](https://www.wikidata.org/wiki/Q3918).
+
+One useful tool to visualize the subclasses of a given type is the [Wikidata Graph Builder](https://angryloki.github.io/wikidata-graph-builder/). For instance, here are [all the subclasses of academic institution (Q4671277)](https://angryloki.github.io/wikidata-graph-builder/?property=P279&item=Q4671277&mode=reverse).
```
ostephens, line 45: This seems like quite a lot of detail - do we want to go into this level of detail about wikidata ontology structure?

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,251 @@ sidebar_label: Wikidata
...
+
+Some items are not marked as the instances of anything, because no one has taken the time to add these statements yet. If you restrict the reconciliation to a type, these items will not appear in the results, except when no items of the given type could be found for the particular name. In this case, items with no type can be returned: they will have a low score and will not be matched automatically.
+
+### Reconciling via unique identifiers
+
+You can supply a column of unique identifiers directly to Wikidata in order to pull more data, but these strings will not be "reconciled" against the external dataset. Matching identifiers does not validate them. 
+
+Apply the operation “Reconcile” → “Use values as identifiers” on your column of identifiers. All cells will appear as dark blue "confirmed" matches. You should check before this operation that the identifiers in the column exist on the target service.
+
+You may get false positives, which you will need to hover over or click on to identify.
```
ostephens, line 59: I know we've discussed this already but I think this is lacking a bit of context - I'd suggest that we explain what might happen if you use invalid identifiers rather than saying "false positives" which not everyone will understand

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,251 @@ sidebar_label: Wikidata
...
+*   `Sde` for Sitelink in German
+
+Enter in the three-letter code manually in the <span class="menuItems">Add Property</span> field. You can manually enter as many term/language combinations as you require. 
+
+No language fall-back is performed when retrieving the values.
+
+### Comparing values
+
+By default, data in OpenRefine and entities in Wikidata are compared by string fuzzy-matching. There are some exceptions to this:
+*   If the value is an identifier, then it confirms the exact string without matching.
```
ostephens, line 114: Do we need to be clear here that this is only true when you use the function to use the identifiers (i.e. by passing reconciliation)?

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,251 @@ sidebar_label: Wikidata
...
+*   If the values are integers, exact equality between integers is used.
+*   If the values are floating point numbers, the score is 100 if they are equal and decreases towards 0 as their absolute difference increases.
+*   If the values are coordinates (specified in the "lat,lng" format on OpenRefine's side), then the matching score is 100 when they are equal and decreases as their distance increases. Currently a score of 0 is reached when the points are 1 kilometre away from each other.
+
+Sometimes, we need a more specific matching on sub-parts of these values. It is possible to select these parts for matching by appending a modifier at the end of the property path:
```
ostephens, line 119: This is really useful information but I wonder if it should be in the OpenRefine docs or whether we should have a pointer to external documentation on the Wikidata reconciliation service?

ostephens (changes requested):
I've not reviewed the "uploading data to wikidata" yet but these are my comments on the rest of it

docs/docs/manual/wikidata.md
wetneb: That makes sense, it should be documented externally indeed.

docs/docs/manual/wikidata.md
allanaaa: Instead of explicating this too much here, I think we can just link to the equivalent section on the Reconciliation page? 

> You can supply a column of unique identifiers directly to Wikidata in order to pull more data, but [these strings will not be "reconciled" against the external dataset](reconciling#reconciling-with-unique-identifiers). These identifiers will not be validated. 
>
>Apply the operation “Reconcile” → “Use values as identifiers” on your column of identifiers. All cells will appear as dark blue "confirmed" matches. Some of the "matches" may be errors, which you will need to hover over or click on to identify.

docs/docs/manual/wikidata.md
allanaaa: Do you mean drop the text into someplace like https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine#Wikidata_reconciliation? It doesn't have much space to explicate, at present.

Or somewhere like https://wikidata.reconci.link/? That just points back to our wiki, and to https://github.com/wetneb/openrefine-wikibase - which is another potential place to document it.

allanaaa: Sorry, I should've started with: I can't find this documented anywhere else on the web except in our own wiki.

wetneb: Yes it should be documented in https://github.com/wetneb/openrefine-wikibase, or perhaps on the web service itself, https://wikidata.reconci.link/

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,240 @@ sidebar_label: Wikidata
...
+
+Batches of edits to Wikidata that are created with OpenRefine can be undone. You can test out the uploading process by reconciling to several “sandbox” entities created specifically for drafting edits and learning about Wikidata:
+* https://www.wikidata.org/wiki/Q4115189
+* https://www.wikidata.org/wiki/Q13406268
+* https://www.wikidata.org/wiki/Q15397819
+* https://www.wikidata.org/wiki/Q64768399
+
+If you upload edits that are redundant (that is, all the statements you want to make have already been made), nothing will happen. If you upload edits that conflict with existing information (such as a different birthdate than one already in Wikidata), it will be added as a second statement. OpenRefine produces no warnings as to whether your data replicates or conflicts existing Wikidata elements.
```
tfmorris, line 137:
```suggestion
If you upload edits that are redundant (that is, all the statements you want to make have already been made), nothing will happen. If you upload edits that conflict with existing information (such as a different birthdate than one already in Wikidata), it will be added as a second statement. OpenRefine produces no warnings as to whether your data replicates or conflicts with existing Wikidata elements. 
```

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,240 @@ sidebar_label: Wikidata
...
+You can refer to a term in a specific language by manually entering in a three-letter code. The first letter is the term (“L” for label, “D” for description, “A” for aliases, “S” for sitelink) and the next two are the language code. For example:
+
+*   `Len` for Label in English
+*   `Dfi` for Description in Finnish
+*   `Apt` for Alias in Portuguese
+*   `Sde` for Sitelink in German
+
+Enter in the three-letter code manually in the <span class="menuItems">Add Property</span> field. You can manually enter as many term/language combinations as you require. 
+
+No language fall-back is performed when retrieving the values.
```
tfmorris, line 98: I'm not sure people will know what this means, but since it's something that's not done, does it even need to be mentioned?

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,240 @@ sidebar_label: Wikidata
...
+
+When you select or enter a type for reconciliation, OpenRefine will include that type and all its subtypes. For instance, if you select [university (Q3918)](https://www.wikidata.org/wiki/Q3918), then [university of Ljubljana (Q1377)](https://www.wikidata.org/wiki/Q1377) will be a possible match.
+
+Some items may not yet be set as an instance of anything, because Wikidata is crowdsourced. If you restrict reconciliation to a type, these items will not appear in the results, except as a fallback, and will have a lower score.
+
+### Reconciling via unique identifiers
+
+You can supply a column of unique identifiers directly to Wikidata in order to pull more data, but [these strings will not be “reconciled” against the external dataset](reconciling#reconciling-with-unique-identifiers). Apply the operation <span class="menuItems">Reconcile</span> → <span class="menuItems">Use values as identifiers</span> on your column of identifiers. All cells will appear as dark blue “confirmed” matches. Some of the “matches” may be errors, which you will need to hover over or click on to identify.
```
tfmorris, line 50: This is presumably talking about external identifiers like VIAF, ORCID, etc. It would useful to include that context and also mention that it only works for identifiers which have an associated Wikidata property.

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,240 @@ sidebar_label: Wikidata
...
+
+Sometimes the same external identifier is assigned to different Wikidata items (in which case the conflict will be flagged in Wikidata as a uniqueness constraint violation). In this case, all of the matching items are returned as candidates, but are not automatically matched.
```
tfmorris, line 52:
```suggestion
If the same external identifier is assigned to multiple Wikidata items, all of the items are returned as candidates, with none automatically matched.
```

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,240 @@ sidebar_label: Wikidata
...
+
+For example, [Q5](https://www.wikidata.org/wiki/Q5) has the English label “human” and the aliases “person” and “people,” with a description including “common name of Homo sapiens;” it also has the Italian label “umano” and the alias “persona” with a description “specie a cui appartiene il genere umano.” The English sitelink is “https://en&#46;wikipedia.org/wiki/Human” but Italian Wikipedia has no equivalent article attached to Q5. Not every language will have values, but these terms can still be useful for reconciling. 
+
+After reconciliation, you can extend your data with these terms. You can fetch aliases for a given term, or labels in another languages, by using <span class="menuItems">Edit column</span> → <span class="menuItems">Add columns from reconciled values....</span>. 
+
+You can refer to a term in a specific language by manually entering in a three-letter code. The first letter is the term (“L” for label, “D” for description, “A” for aliases, “S” for sitelink) and the next two are the language code. For example:
```
tfmorris, line 89: This seems very obscure. I know that's not the docs fault, but perhaps we should see if we can improve it.

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,240 @@ sidebar_label: Wikidata
...
+* https://www.wikidata.org/wiki/Q4115189
+* https://www.wikidata.org/wiki/Q13406268
+* https://www.wikidata.org/wiki/Q15397819
+* https://www.wikidata.org/wiki/Q64768399
+
+If you upload edits that are redundant (that is, all the statements you want to make have already been made), nothing will happen. If you upload edits that conflict with existing information (such as a different birthdate than one already in Wikidata), it will be added as a second statement. OpenRefine produces no warnings as to whether your data replicates or conflicts existing Wikidata elements. 
+
+You can use OpenRefine's previews to look at the target Wikidata elements and see what information they already have, and whether the elements' histories have had similar edits reverted in the past. 
+
+Wikidata is unlike Wikipedia in that there are fewer editors supervising consistency. Many entities' discussion pages are not used; rather, discussion happens on individual users' pages and on project pages.
```
tfmorris, line 141:
```suggestion
```

This doesn't have anything to do with OpenRefine

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,240 @@ sidebar_label: Wikidata
...
+|Disgrace|Disgrâce|French|
+||Vergogna|Italian|
+|Wolf Hall|En la corte del lobo|Spanish|
+||ウルフ・ホール|Japanese|
+
+You could upload translated titles to “Label” with the language from “Translation language.” You may wish to fetch the two-letter language code and use that instead for better language matches.
+
+### Manage Wikidata account
+
+To edit Wikidata directly from OpenRefine, you must have a Wikidata account and log in with it in OpenRefine. OpenRefine can only upload edits with Wikidata user accounts that are “[autoconfirmed](https://www.wikidata.org/wiki/Wikidata:Autoconfirmed_users)” - that is, accounts that have more than 50 edits and have existed for longer than four days.
```
tfmorris, line 186: I suspect these criteria are subject to change, so probably best to just point to wherever it's defined on Wikidata.

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,240 @@ sidebar_label: Wikidata
...
+
+### Manage Wikidata account
+
+To edit Wikidata directly from OpenRefine, you must have a Wikidata account and log in with it in OpenRefine. OpenRefine can only upload edits with Wikidata user accounts that are “[autoconfirmed](https://www.wikidata.org/wiki/Wikidata:Autoconfirmed_users)” - that is, accounts that have more than 50 edits and have existed for longer than four days. 
+
+Use the Extensions menu to select <span class="menuItems">Manage Wikidata account</span> and you will be presented with the following window:
+
+![The Wikidata authorization window in OpenRefine.](/img/wikidata-login.png)
+
+For security reasons, you may not wish to use your usual account authorization with OpenRefine. Wikidata allows you to set special passwords to access your account through software. You can find this setting for your account at [https://www.wikidata.org/wiki/Special:BotPasswords](https://www.wikidata.org/wiki/Special:BotPasswords) once logged in. Creating bot access will prompt you for a unique name, and allow you to enable the following required settings:
```
tfmorris, line 192:
```suggestion
For security reasons, it is suggested that you not use your main account authorization with OpenRefine. Wikidata allows you to set special passwords to access your account through software. You can find this setting for your account at [https://www.wikidata.org/wiki/Special:BotPasswords](https://www.wikidata.org/wiki/Special:BotPasswords) once logged in. Creating bot access will prompt you for a unique name, and allow you to enable the following required settings:
```

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,240 @@ sidebar_label: Wikidata
...
+
+Under the <span class="menuItems">Export</span> menu, look for <span class="menuItems">QuickStatements file</span>; under <span class="menuItems">Extensions</span> look for <span class="menuItems">Export to QuickStatements</span>. Exporting your schema from OpenRefine will generated a text file called `statements.txt` by default. Paste the contents of the text file into a new QuickStatements batch using version 1. You can find version 1 of the tool (no longer maintained) [here](https://wikidata-todo.toolforge.org/quick_statements.php). The text commands will be processed into Wikidata edits and previewed for you to review before submitting. 
+
+There are advantages to using QuickStatements rather than uploading your edits directly to Wikidata, including the way QuickStatements resolves duplicates and redundancies. You can learn more on QuickStatements' [Help page](https://www.wikidata.org/wiki/Help:QuickStatements), and on OpenRefine's [Uploading page](https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing/Uploading).
+
```
tfmorris, line 234: All of the QuickStatements specific stuff can be removed. The relevant OpenRefine piece is that we can generate data for it and it's in "v1" format.

tfmorris (changes requested):
We should trim back the redundant Wikidata, QuickStatements, etc info and focus on a) OpenRefine and b) OpenRefine's integration points with Wikidata. The more information of theirs that we repeat, the more likely it is to be wrong, get outdated, and require ongoing maintenance.

docs/docs/manual/wikidata.md
allanaaa: No, this is definitely about Wikidata.

tfmorris: Perhaps then call them "Wikidata identifiers" instead of generically "unique identifiers."

docs/docs/manual/wikidata.md
allanaaa: Not since 2013. But I can add in "as of the time of writing."

docs/docs/manual/wikidata.md
allanaaa: I don't think this is an excessive amount of information so I'd leave it as is.

docs/docs/manual/wikidata.md
allanaaa: I don't think that alone is sufficient criteria to exclude information. 

Although I suppose since someone has to have 50 manual wikidata edits before they can use OpenRefine to edit it, they will likely be familiar with the culture and workflow already.

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,240 @@ sidebar_label: Wikidata
...
+OpenRefine’s connections to Wikidata were formerly an optional extension, but are now installed automatically with the downloadable package. The Wikidata extension can be removed manually by navigating to your OpenRefine installation folder, and then looking inside `webapp/extensions/` and deleting the `wikidata` folder inside. 
 
-*   Former extension, now core [but you can remove it].
-*   Account required
+You do not need a Wikidata account to reconcile your local OpenRefine project to Wikidata. If you wish to upload your cleaned dataset to Wikidata, you will need to register an account, and authorize OpenRefine with that account. 
 
+The best source for information about how OpenRefine works with Wikidata is [on Wikidata itself, under Tools](https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine). This has tutorials, guidelines on editing, and spaces for discussion and help.
 
-## Features
+Everything true about the way OpenRefine works with Wikidata is also true about any other Wikibase. Anything built on the same Wikibase platform will offer a reconciliation endpoint that works the same. To find a Wikibase reconciliation URL,
```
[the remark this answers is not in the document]
allanaaa: Whoops! We left off on our discussion of this because the wikibase connections are in flux right now (https://phabricator.wikimedia.org/T197588). 
I think we may leave this bit off for now.

docs/docs/manual/wikidata.md
allanaaa: @wetneb has put a version of this information here: https://wikidata.reconci.link/#documentation

It looks like it's just copied from the original wiki text. I have an updated version of the same text, but I've now removed it from this draft and saved it elsewhere. If we'd like to update the text on that page, we can take care of it another time.

allanaaa:
@tfmorris Can you let me know if you think there are still redundancies on this page? Thanks!

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,184 @@ sidebar_label: Wikidata
 
 ## Overview
 
+OpenRefine provides powerful ways to both pull data from Wikidata and add data to it. 
 
+OpenRefine’s connections to Wikidata were formerly an optional extension, but are now installed automatically with the downloadable package. The Wikidata extension can be removed manually by navigating to your OpenRefine installation folder, and then looking inside `webapp/extensions/` and deleting the `wikidata` folder inside.
```
wetneb, line 11: I'd simplify that a bit:
```suggestion
OpenRefine’s connections to Wikidata is supplied by an extension that is available by default in OpenRefine. The Wikidata extension can be removed manually by navigating to your OpenRefine installation folder, and then looking inside `webapp/extensions/` and deleting the `wikidata` folder inside. 
```

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,184 @@ sidebar_label: Wikidata
...
+
+There are advantages to using QuickStatements rather than uploading your edits directly to Wikidata, including the way QuickStatements resolves duplicates and redundancies. You can learn more on QuickStatements' [Help page](https://www.wikidata.org/wiki/Help:QuickStatements), and on OpenRefine's [Uploading page](https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing/Uploading).
+
+### Schema alignment
+
+The best resource is the [Schema alignment page](https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing/Schema_alignment) on Wikidata.
```
wetneb, line 175: On the long term I'd prefer to maintain this documentation here rather than on Wikidata, especially since we have now generalized this integration to other Wikibase instances (so it really ought to be documented in OpenRefine's docs directly).

docs/docs/manual/wikidata.md
```diff
@@ -6,28 +6,184 @@ sidebar_label: Wikidata
...
+
+### Issue detection
+
+The best resource is the [Quality assurance page](https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing/Quality_assurance) on Wikidata.
```
wetneb, line 179: Same here (and other places).

wetneb (commented):
It is much better like that indeed, thank you! I think the documentation that you point to in Wikidata should be migrated here eventually. I originally wrote it there because it was more comfortable to do so in a proper wiki, but it really ought to be here instead.

I am not sure if we should make this migration right now. Perhaps it is better to only do it once we can translate our docs, since these docs are already partially translated (so it would be nice to transfer the translations at the same time, to avoid synchronization issues).

allanaaa:
I think we are agreed that more work will happen on this page and the various other areas where information is to be found. Currently this work is scheduled for January, if we don't get to it sooner. So, if we're happy with the draft in this current form, we can go ahead and merge this!

wetneb (approved):

PR #3206, opened 2020-09-23
Clarify refine.ini comments - fixes #3204

### LINKED ISSUE #3204: refine.ini file current states that if I am a user that it is recommended that I use my own credentials for Google Data
In the refine.ini on L29 it mentions that Default OAuth credentials for Google Data have been embedded in the release version of OpenRefine. 
`# Default OAuth credentials for Google Data have been embedded in the release version of OpenRefine`

And on L30 it mentions that if I am a user that I can skip the configuration, but its recommended to use my own credentials.
`# So if you are a user, you can just skip these configurations, but it's recommended to use your own credentials`

This is confusing and conflicting information.  My expectation, as a user, is that I should only be prompted with an authorization login box for Google Data and enter my credentials, and not have to setup special configuration through the use of Google API Console for credential management as a user of OpenRefine.

I think L30 can just be simplified to only say:
`# So if you are a user, you can just skip these configurations.`

### Versions
 - JRE or JDK Version: JDK8
 - OpenRefine: OpenRefine 3.4 with java

tfmorris:
Fixes #3204 

Clarifies comments for:
- Google Data OAuth config
- JVM heap

NOTE: Our current JVM max heap settings are *lowering* the size computed for the default on my machine (4096 MB => 1400 MB), so we may just want to accept the defaults unless the user specifically wants to override them.

### DIFF
--- a/refine.ini
+++ b/refine.ini
@@ -10,11 +10,11 @@ no_proxy="localhost,127.0.0.1"
 #REFINE_HOST=127.0.0.1
 #REFINE_WEBAPP=main\webapp
 
-# Memory and max form size allocations
+# Maximum JVM heap (memory) and max form size allocations
 #REFINE_MAX_FORM_CONTENT_SIZE=1048576
 REFINE_MEMORY=1400M
 
-# Set initial java heap space (default: 256M) for better performance with large datasets
+# Initial (and minimum) size of Java heap
 REFINE_MIN_MEMORY=1400M
 
 # Some sample configurations. These have no defaults.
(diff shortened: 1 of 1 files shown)


refine.ini
```diff
@@ -10,27 +10,25 @@ no_proxy="localhost,127.0.0.1"
...
 
-# OAuth credentials configurations for Google Data
-# Default OAuth credentials for Google Data have been embedded in the release version of OpenRefine
-# So if you are a user, you can just skip these configurations, but it's recommended to use your own credentials
-# If you are a developer, you'll need to acquire them by yourself
-# To get your own credentials, please see the wiki: https://github.com/OpenRefine/OpenRefine/wiki/Google-Extension
-# The wiki will guide you to get a client_id/client_secret pair
-# After getting the client_id and client_secret, put them below
+# Google Data OAuth configuration for developers
+# (NOTE: This are only needed for developers. Users of released versions can ignore this)
```
thadguidry, line 30:
```suggestion
# (NOTE: These GDATA settings are only needed for developers. Users of released versions can ignore this)
```

thadguidry (changes requested):

wetneb:
Periodic note that this release process is flaky: our Google credentials will get deactivated one day because we are not supposed to ship them to users, even in a vaguely hidden way.

wetneb (approved):

refine.ini
tfmorris: Fixed subject verb agreement

Issue #3208, opened 2020-09-23
Unstable Mac CI


wetneb:
The Travis builds regularly fail because of the Mac sub-task:
https://travis-ci.org/github/OpenRefine/OpenRefine/builds/729523826

This task is currently marked as required, but it is less stable than the builds with "allowed failures" that we currently have. So I would mark the Mac build as "allowed failure" and bring the openjdk12 in the mandatory section.

thadguidry:
Perhaps we should just move the Mac build to GitHub Actions?  I'm not a Mac guy so not the best to help there, but I can certainly try.

wetneb:
I would not move a single build to GitHub Actions. If we move to that, we should move all runs there.

tfmorris:
This is the first time I'm hearing that it's been failing. I don't think any of my builds have failed. 

Is the failure mode always the same as the example that you linked (looks like the machine failed to boot/initialize)?

The test setup is a sparse array with that build testing not only Mac, but also Oracle JDK 14.

wetneb:
Here are more examples:
* https://travis-ci.org/github/OpenRefine/OpenRefine/builds/728450315
* https://travis-ci.org/github/OpenRefine/OpenRefine/builds/727971157
* https://travis-ci.org/github/OpenRefine/OpenRefine/builds/725888917
* https://travis-ci.org/github/OpenRefine/OpenRefine/builds/725889319

tfmorris:
Thanks. It looks like the same failure mode for all 5 examples. This might have a simple solution (retries, increased timeouts) so it seems worth spending a minute or two investigating before just dropping testing on one of our three platforms. I'll volunteer to do this.

Although it looks like the `java` command is timing out, but I suspect that's a red herring and that it's actually Homebrew installation of databases that is timing out.

tfmorris:
Travis has finally acknowledged my report, but no progress on fixing it. My current approach is to pursue the migration to travis-ci.com in the hope that it is better behaved.

tfmorris:
After the migration, the problem exists on travis-ci.com as well, so I'll continue to pursue other solutions.

thadguidry:
@tfmorris GitHub Actions, GitHub Actions, nudge nudge nudge... :-)  https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners

tfmorris:
> nudge nudge nudge

[...]

PR #3209, opened 2020-09-23
Mark the Mac build as an allowed failure.

wetneb:
For #3208

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,10 +40,10 @@ jobs:
 #      script: mvn test
 #      after_success: mvn prepare-package -DskipTests=true
   allow_failures:
+    - os: osx
     - jdk: openjdk-ea
     - jdk: oraclejdk-ea
     # JDKs below need to be installed and installation fails frequently
-    - jdk: openjdk12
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -40,10 +40,10 @@ jobs:
 #      script: mvn test
 #      after_success: mvn prepare-package -DskipTests=true
   allow_failures:
+    - os: osx
     - jdk: openjdk-ea
```
tfmorris, line 44: Note to self - the early access versions aren't run, so this "allow failures" has no effect.

.travis.yml
```diff
@@ -40,10 +40,10 @@ jobs:
 #      script: mvn test
 #      after_success: mvn prepare-package -DskipTests=true
   allow_failures:
+    - os: osx
```
tfmorris, line 43: This was our Java 14 test, so we should probably add a replacement to cover that.

tfmorris (approved):
I'm going to merge this as is since the failure rate seems to be even higher now, but we need to figure out both Java 14 and Mac coverage.

PR #3212, opened 2020-09-24
Validate Host header of incoming HTTP requests

### LINKED ISSUE #3288: HTTP server implementation is vulnerable to DNS rebinding attacks
OpenRefine has endpoints with intentional remote code execution without authentication (like running Python or Clojure transformations). While the lack of authentication would not be an issue with normal desktop use when OpenRefine is bound to a loopback interface like 127.0.0.1, the current implementation still exposes users to Remote Code Execution via [DNS rebinding](https://en.wikipedia.org/wiki/DNS_rebinding) attacks.

### To Reproduce

I did not prepare a full Proof of Concept exploit, but the general steps would be:

- Run OpenRefine
- Visit rebindable.com, which loads a malicious script in the browser
- Attacker rebinds rebindable.com to 127.0.0.1
- Malicious script calls OpenRefine endpoints as if they were under rebindable.com

### Current Results

The calls will succeed and an attacker can use them to run arbitrary code on the user's computer. This works because after the domain has been rebound, the script can now call OpenRefine endpoints as if they were same-origin.

### Expected Behavior

The calls should be rejected.

### Additional context

I have sent a pull request (#3212) to address this problem by validating the Host header. See implementation details in the thread associated with the PR. This issue was opened at the request of @tfmorris.

dzsibi:
Fixes #3288 

This patch adds a simple HTTP Host header check to prevent DNS rebinding attacks. As OpenRefine offers endpoints with intentional remote code execution without authentication (like running Python or Clojure transformations), normal desktop use would expose users to Remote Code Execution via DNS rebinding attacks. (Please note that this type of attack bypasses the CSRF protections built into OpenRefine, since `/command/core/get-csrf-token` can be called once cross-origin policies are bypassed.)

As a consequence of this patch, any normal execution of OpenRefine will have Host header checks enabled, except when binding to 0.0.0.0 or if the user explicitly opts out by specifying `-Drefine.host.validate=false` as a command line option. This won't affect desktop users at all who are using the default configuration, or variations of it, but will affect those running OpenRefine on a server or behind a proxy, and there is a mismatch in host names. (Alternatively, the default value of `refine.host.validate` could be altered to make this behavior opt-in, but a secure default value would be better.)

The patch itself should be pretty low impact, but since I am not extremely familiar with Jetty's peculiarities, there could be some issues I missed.


wetneb:
This makes sense, thanks a lot! I will test this out and get back to you.

tfmorris:
Thanks for the PR! Is this addressing an existing issue? If so, could you please reference it in your description with "Fixes #nnn". If not, please create an issue which describes the perceived issue.

Also, if you make your pull request from a feature branch instead of `master`, it will make it less likely that unintentional commits gets mixed in.

dzsibi:
I couldn't find any mention of this problem in existing issues, open or closed. I could open a separate issue and copy over the details from the PR's description, if it helps with the approval process. 

Since I only forked to submit this PR (and will archive / delete the repo once it is merged), I don't expect there to be any additional commits on this branch or any other in my copy, unless changes are requested during review.

wetneb (changes requested):
I think this might still be a bit too strict. For instance, if I start OpenRefine on `127.0.0.1`, I expect accessing it via `localhost` should work. Similarly, I could try to access it via IPv6 with `http://[::ffff:7f00:1]:3333/` and that should probably work too.

I am not sure how to handle these cases. Perhaps we should have a command-line setting to set a list of allowed hosts? Arguably the `localhost`/`127.0.0.1` could perhaps be hard-coded like `0.0.0.0` is.

dzsibi:
Hmmm, instead of relying on `refine.host`, we could allow all loopback IPv4 and IPv6 addresses (127.0.0.0-127.255.255.255, ::1), as well as localhost. Then, add an explicit parameter that would whitelist additional hostnames when remote access or a different network topology is required. Do you think that would work?

wetneb:
Sounds like a very sensible plan!

tfmorris:
I **still** would like to see an issue which describes what problem is being solved here. The fact that it hasn't been reported as a problem would seem to indicate that it's not something that users are clamoring to have fixed.

OpenRefine was designed for use on a single computer, so once you get outside of the environment, you're going to find all kinds of things that need fixing.

dzsibi:
I will to make the changes discussed tomorrow.

@tfmorris Users who have the same assumption you have just mentioned, that OpenRefine was designed for use on a single computer, expect that the web UI is just that: a user interface that just happens to be accessible from a browser. The reality is, however, that they are still running a web server, and even though by default it is bound on a loopback address, this configuration still exposes them to certain remote attacks, like DNS rebinding.

I am not trying to solve a problem that is non-existent for single computer users, I am trying to patch a vulnerability that can be used to run arbitrary code on the user's computer if the user visits a malicious website while running OpenRefine. (Nobody will report this as a "problem" until such an attack is mounted and noticed.)  Also, please note the mitigation is invisible for users who are using OpenRefine as you described, even in its current form. It is only when it is used as a full fledged remote service (either on its own or behind a reverse proxy) that additional configuration might be necessary.

wetneb:
I guess Tom's point is that it would be useful to have an issue to link to when reporting this in the changelog. So if you could just paste what you just explained in an issue, we could then link it to this pull request and it would tick the boxes for our own bureaucracy.

dzsibi:
I have finally gotten around to making the changes. All loopback addresses (IPv4 and IPv6 alike) are now explicitly whitelisted, as is localhost. Any value set for `refine.host` is also allowed, and setting it to 0.0.0.0 or using `refine.host.validate=false` disables the check, as before. Please take a look at the code and the results. I will also open an issue as you and Tom have requested tomorrow.

PS: I just noticed my editor did some housecleaning and trimmed all trailing whitespace. If this is against the coding style, do tell and I will have that fixed.

server/src/com/google/refine/Refine.java
```diff
@@ -435,32 +447,92 @@ static private String fixWindowsUnicodePath(String path) {
...
-    
+
+}
+
+/**
+ * Validate the Host header of the HTTP request to see if it matches either a loopback IP
+ * address, localhost or an explicitly specified hostname. This is required to avoid DNS
+ * rebinding attacks against users running OpenRefine on their desktop computers.
+ */
+class ValidateHostHandler extends HandlerWrapper {
```
wetneb, line 471: I wonder if it would make sense to write one or two simple tests to show which hosts are allowed.
You would probably need to move this class to its own file or making it a nested class of `Refine`.

(I have to admit I wasn't aware of the fact that we had multiple top-level classes in `Refine.java`, it looks a bit confusing to me, and does not seem to be a widely recommended practice.)

dzsibi:
I looked into testability, and found that in the current project configuration, it is extremely hard to add tests to anything related to the `server` project - there is no infrastructure in place right now to run tests there. The `main` servlet project has that in place, but it has no linkage to the `server` project - so I cannot just put parts of the code there to make it testable and then reference it from the `server` project. While I am open to suggestions here, I would rather avoid sweeping architectural changes.

As for the multiple top-level classes, I tried to follow existing conventions, which seemed to place everything into that one file, but I agree that it is confusing and definitely not the recommended practice. I moved it to a separate class. (In the process, I also reverted the whitespace changes in Refine.java, so the diff would look a little cleaner.)

wetneb:
Makes sense, it's not your responsibility to come up with this testing architecture. Testing this interactively when running OpenRefine with default settings, I get:
- `curl -H "Host: example.com" http://localhost:3333/` gives an error as expected, but I am wondering if 404 is the best error code for that? Shouldn't it be a 400 error instead? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400
- http://[::ffff:7f00:1]:3333/ returns a 404 error. It's not clear to me how widespread it is to use IPv6 in URLs, but that should probably be allowed as discussed above

dzsibi:
I don't think that the IPv4-mapped IPv6 range for the 127.0.0.0/8 subnet is used in any realistic scenario, the ::1 native IPv6 loopback address is much more likely to be used in an IPv6 only situation for local access - but I added the additional range to the regular expression and now the address http://[::ffff:7f00:1]:3333/ (and others in the same range) should work correctly. 

I picked the 404 error code since the the problem is not with the request, but with the URL itself. But I understand that this might cause confusion in users who encounter this error, and might make it harder to debug such issues - so I added a more verbose error message to the response, see if this works for you.

dzsibi:
@wetneb Any other changes / additional work needed in your opinion?

wetneb:
Sorry for the delay!

With a fresh look at this, I am wondering why we would disable the host checks if we are listening to 0.0.0.0. As a user, if I want to expose an OpenRefine instance on a specific host, I would do `./refine -i 0.0.0.0 -d refine.host=example.com` and expect that this would enforce host checking.

That's a bit beyond your motivating use case, and OpenRefine is really not designed to be hosted openly like this, but if we agree that this should be the expected behaviour it is not hard to get it right.

wetneb:
@dzsibi any thoughts about my last comment above?

dzsibi:
Sorry, I missed these responses somehow. So right now the host parameter is also used to specify the interface - Jetty's SocketConnector is configured with the value of `refine.host`. Maybe it would be better to split these up, and add a `refine.interface` parameter, which would fall back to `refine.host` if not present (to maintain backwards compatibility). What do you think about this solution?

dzsibi:
I went ahead and implemented it quickly. Current behavior:

- If you only specify `refine.host`, it will be used to select the interface and to validate the Host header
- If you set `refine.host` to 0.0.0.0, there is no Host header validation, since there is no way to know what to expect
- If you specify `refine.host` and `refine.interface`, the server will bind to the interface and the Host will be validated separately
- If you set `refine.host.validate` to `false` and `refine.interface`, `refine.host` will be ignored altogether
- If you only specify `refine.interface`, `refine.host` will fall back to its default value of 127.0.0.1 - likely an impossible combination, but should still work if `refine.interface` was set to a loopback address due to their separate handling

wetneb (commented):
Very nice, thank you! I think we still need to update the `refine` and `./refine.bat` scripts then, because the `-i` option of those scripts passes on its value as `refine.host` property - but it should now be `refine.interface` instead, and we'd need to add a parameter to specify the host.

dzsibi:
To match expectations regarding the `-i` command line argument, I had to switch how the two parameters inherit from each other (now `refine.host` falls back to `refine.interface`, not the other way around). Here is how a full call looks like:
```
./refine -i 0.0.0.0 -h openrefine.internal
```
And this one is how it was called before - here `refine.host` will fall back to 127.0.0.1:
```
./refine -i 127.0.0.1
```
On Windows, we have similar arguments `/i` and `/h`:
```
refine /i 127.0.0.1 /h openrefine.internal
```
To disable Host header validation, pass `*` as the value (on Unix like systems, this would need to be quoted):
```
refine /i 127.0.0.1 /h *
```
I had to cannibalize the `/h` and `-h` help arguments to keep things consistent (on Windows, the default was `/?`, and `/h` wasn't even displayed on the usage screen, while in the bash script, I added a new `--help` switch to replace the old one.

See if these changes work for you.

wetneb:
This is great, thank you so much!
I am a bit torn about the repurposing of the `-h` (or `/h`) options. Perhaps it would be worth making sure that when running `./refine -h` (or `refine.bat /h`) we just give the help text, as before, which would even make sense regardless of the previous behaviour (because using `-h` without supplying a host name is unexpected).
What do @thadguidry and @tfmorris think about the syntax described above?

thadguidry:
@wetneb I think the proposed syntax is fine to use, as long as a failed argument passing through the bash/batch script still tells the user what syntax to use to get the help information always.  So we need to make sure that invalid arguments are still caught, and perhaps tweak the bash/batch scripts.

thadguidry (dismissed):
Update L63 on refine.bat to 
`echo Type 'refine /?' for usage.`

tfmorris:
I'm not keen on changing the behavior of existing switches/flags. Other options might be `-H` or `--host`.

dzsibi:
@tfmorris I changed /h and -h to /H and -H respectively, and restored the old /h and -h arguments to display the usage screen as before.

wetneb (approved):

wetneb:
Thank you so much for your patience and perseverance! For me this seems good to go now.

dzsibi:
I fixed the merge conflict and brought the changes up to date with the new Jetty version. Let me know if you need anything else before this can be merged.

Issue #2265, opened 2019-12-25
SQL Exporter skips null values


colognesdgs:
I have tried to export data as a SQL script and have used tables, that have empty columns:

```
OTS	NAME	STRS	BEZS	KLRID	EXTID	OTNAME	STATUS	STRNAME	BEZNAME	OBJECTID	ORIG_FID	OBJEKTTYP	SONSTIGES	SHAPE_LENG	HYPERLINK1	HYPERLINK2	BEMERKUNGEN	BEMERKUN_1
205	Kalscheurer Weiher	3415	2	12811	 	Zollstock	1	Zollstocker Weg	Rodenkirchen	23	23	10		1472.59572795	http://www.stadt-koeln.de/leben-in-koeln/freizeit-natur-sport/parks/65785/ 			Kiosk, barrierefrei
```

e.g. Column HYPERLINK2, BEMERKUNGEN is empty in the record.

The sql exporter generates a statement like this:

```
INSERT INTO ODC_PARKANLAGEN (OTS,NAME,STRS,BEZS,KLRID,EXTID,OTNAME,STATUS,STRNAME,BEZNAME,OBJECTID,ORIG_FID,OBJEKTTYP,SONSTIGES,SHAPE_LENG,HYPERLINK1,HYPERLINK2,BEMERKUNGEN,BEMERKUN_1) VALUES ( 205,'Kalscheurer Weiher',3415,2,12811,' ','Zollstock',1,'Zollstocker Weg','Rodenkirchen',23,23,10,1472.59572795,'http://www.stadt-koeln.de/leben-in-koeln/freizeit-natur-sport/parks/65785/ ','Kiosk, barrierefrei' ); 
```

Which leads to an error in MariaDB, for example:
```
ERROR 1136 (21S01): Column count doesn't match value count at row 1
```

I have expected the sql exporter to create a statement like this:

```
INSERT INTO ODC_PARKANLAGEN 
[...]

PR #3236, opened 2020-09-30
Maintain ordering for GREL uniques() function

### LINKED ISSUE #3235: GREL uniques() function should preserve array order
Currently the GREL `uniques()` function returns the array elements in, effectively, random order. It would be more useful and predictable if the ordering were stable.

### Proposed solution

Return the array in the same order as the original, but with the duplicates eliminated.

tfmorris:
Fixes #3235. Also adds tests.

Unrelated bonus test additions include:
- enables an array `sort` test that wasn't annotated correctly.
- adds tests for `reverse` and `join`

### DIFF
--- a/main/tests/server/src/com/google/refine/expr/functions/arrays/JoinTests.java
+++ b/main/tests/server/src/com/google/refine/expr/functions/arrays/JoinTests.java
@@ -28,13 +28,35 @@ package com.google.refine.expr.functions.arrays;
 
 import org.testng.annotations.Test;
 
+import com.google.refine.RefineTest;
+import com.google.refine.expr.ParsingException;
 import com.google.refine.util.TestUtils;
 
-public class JoinTests {
+public class JoinTests extends RefineTest {
     @Test
     public void serializeJoin() {
         String json = "{\"description\":\"Returns the string obtained by joining the array a with the separator sep\",\"params\":\"array a, string sep\",\"returns\":\"string\"}";
         TestUtils.isSerializedTo(new Join(), json);
     }
+
+    @Test
+    public void joinArray() throws ParsingException {
+        String[] test = {"[2,1,3].join('|')", "2|1|3"};
+        parseEval(bindings, test);
+
+        // TODO: This is current behavior, but is it what we want?
+        String[] test1 = {"[2,null,3].join(', ')", "2, 3"};
+        parseEval(bindings, test1);
+
+        String[] test2 = {"['z','b','c','a'].join('-')", "z-b-c-a"};
+        parseEval(bindings, test2);
+
+        // TODO: Do we really want the following two tests to return different results?
+        String[] test3 = {"['z', null,'c','a'].join('-')", "z-c-a"};
+        parseEval(bindings, test3);
+        String[] test4 = {"['z', '','c','a'].join('-')", "z--c-a"};
+        parseEval(bindings, test4);
+    }
+
 }
 
--- a/main/tests/server/src/com/google/refine/expr/functions/arrays/UniquesTests.java
+++ b/main/tests/server/src/com/google/refine/expr/functions/arrays/UniquesTests.java
@@ -61,6 +61,9 @@ public class UniquesTests extends RefineTest {
 
         String[] test4 = {"[toDate(2020), '2018-03-02'.toDate(), toDate(2020)].uniques().toString()", "[2020-01-01T00:00Z, 2018-03-02T00:00Z]"};
         parseEval(bindings, test4);
+
+        String[] test5 = {"[null,null,null].uniques().toString()", "[null]"};
+        parseEval(bindings, test5);
     }
 
     @Test
--- a/main/tests/server/src/com/google/refine/expr/functions/arrays/UniquesTests.java
+++ b/main/tests/server/src/com/google/refine/expr/functions/arrays/UniquesTests.java
@@ -39,8 +39,9 @@ public class UniquesTests extends RefineTest {
         TestUtils.isSerializedTo(new Uniques(), json);
     }
 
+    @Test
     public void uniquesJsonArray() throws ParsingException {
-        String[] test = {"'[2,1,1,3]'.parseJson().uniques().toString()", "[2, 1, 3]"};
+        String[] test = {"'{a:[2,1,1,3]}'.parseJson().a.uniques().toString()", "[2, 1, 3]"};
         parseEval(bindings, test);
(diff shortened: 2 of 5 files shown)


main/tests/server/src/com/google/refine/expr/functions/arrays/UniquesTests.java
```diff
@@ -28,13 +28,47 @@
...
+        parseEval(bindings, test1);
+
+        String[] test2 = {"['z','b','c','c','a'].uniques().toString()", "[z, b, c, a]"};
+        parseEval(bindings, test2);
+        String[] test3 = {"['z','z',null,'c','a'].uniques().toString()", "[z, null, c, a]"};
+        parseEval(bindings, test3);
+
+        String[] test4 = {"[toDate(2020), '2018-03-02'.toDate(), toDate(2020)].uniques().toString()", "[2020-01-01T00:00Z, 2018-03-02T00:00Z]"};
+        parseEval(bindings, test4);
+    }
```
thadguidry, line 63: Just in case, I think having a test5 for only nulls (thinking about that problem Owen had one time when updating the GREL null handling parts)
```suggestion
        String[] test5 = {"[null,null,null].uniques().toString()", "[null]"};
        parseEval(bindings, test5);
    }
```

thadguidry (dismissed):

main/tests/server/src/com/google/refine/expr/functions/arrays/UniquesTests.java
tfmorris: I don't think I understand that risk that @ostephens identified, but I've added the test.

As an aside, OpenRefine has a deeply held ambivalence about the equivalency of nulls & empty strings, so it doesn't surprise me that problems were encountered when trying to change null handling.

thadguidry: @tfmorris Hi Tom, we changed that equivalency BECAUSE of the ambivalence documented in #820 and a few other issues, where a majority of the community wanted more precise handling and treating null/empty string separately, which I completely agreed with the majority of the community in that regard.  Hence, since Owen merged that work last year that all landed in [3.0 release](https://github.com/OpenRefine/OpenRefine/issues?q=is%3Aissue+milestone%3A3.0+is%3Aclosed) along with the Database Importer/Exporter which also has null handling (and the metadata fiasco you're now aware of), so we now treat `null` and empty strings entirely separately (well, I hope we treat them separately in all areas of OpenRefine, but that's probably not been 100% verified because of probably a few tests lacking in places)
Like in this issue #2265 on the SQL Exporter where @colognesdgs says when using the JSON importer, and then later using the SQL Exporter.

> The data is interpreted as null, but stored by openrefine as empty strings.

At any rate, you get the idea... we still have some soft bugs floating around regarding null/empty string I think in regards to that previous ambivalence, and so testing around `null` and empty string is very dear to my heart still to finally fully address especially for scientific data being ingested through JSON or SQL importers/exporters (part of the reason we were liked by CZI for the grant).

main/tests/server/src/com/google/refine/expr/functions/arrays/JoinTests.java
```diff
@@ -28,13 +28,35 @@
...
+        parseEval(bindings, test);
+
+        // TODO: This is current behavior, but is it what we want?
+        String[] test1 = {"[2,null,3].join(', ')", "2, 3"};
+        parseEval(bindings, test1);
+
+        String[] test2 = {"['z','b','c','a'].join('-')", "z-b-c-a"};
+        parseEval(bindings, test2);
+
+        // TODO: Do we really want the following two tests to return different results?
```
thadguidry, line 54: Yes, they should return differently.  I think that was agreed in #820 and it's linked issues/pr's

main/tests/server/src/com/google/refine/expr/functions/arrays/JoinTests.java
```diff
@@ -28,13 +28,35 @@
...
+        String[] test3 = {"['z', null,'c','a'].join('-')", "z-c-a"};
+        parseEval(bindings, test3);
+        String[] test4 = {"['z', '','c','a'].join('-')", "z--c-a"};
```
thadguidry, line 57: Yes, they should return differently.  I think that was agreed in #820 and it's linked issues/pr's
We might even consider a combined test for null/empty?
`String[] test4 = {"['z', '', null,'c','a'].join('-')", "z---c-a"};`

thadguidry: As I had mentioned before, other data tools have Null morpheme as a preference for the user for their data.  We might eventually want to do that.  I do feel somewhat disappointed that the whole thing felt rushed, and we really didn't get more opinions come in from the community on this...some just expected things to work well for null/empty strings no matter.  Others expected to have more control given to them and not less taken away.  We seemed to have taken some control away and saying "We shall always treat them differently now, no matter what, because we're too lazy to change all the places in our code to deal with Null morpheme".
Maybe we can make that a goal for 2021/22 ... add a preference for null/empty string treated as equivalent?  But it feels backwards to even say that.

thadguidry: Hmm, Java8 has Optional? https://stackoverflow.com/questions/21936503/get-empty-string-when-null

wetneb (approved):

main/tests/server/src/com/google/refine/expr/functions/arrays/JoinTests.java
ostephens: I'm afraid despite lots of effort and discussion we still don't have consistency in situations where a `null` is forced to a string. I think it's definitely worth another review and attempt to get to a reasonable approach but as a separate issue to this. So pending further discussion & agreement I feel its safer to keep the current behaviour for `join`

tfmorris: @thadguidry Your proposed test would fail. The result would actually be `z--c-a` which is one of the weirdly inconsistent things about this.

For the record, I wasn't actually considering any unrelated changes to `join()` as part of this PR. I was just flagging the inconsistency for future review.

I added the tests because I was working in the neighborhood of completely untested functions. I added the comments when I had to change the tests to match the inconsistent behavior of our current implementation.

thadguidry: @tfmorris Thanks Tom.  I see on other issue comments that you are diving into it all.  We'll let you work it all out :-)  Thanks so much!

PR #3237, opened 2020-10-01
Refactor HTTP code into common module & Improve Fetch URL  - fixes #3129

### LINKED ISSUE #3129: Fix Fetch URL delay calculation
As a side effect of the introduction of caching (782a2f5b48d54dd839398b14b72c878f7b526a10), the delay calculation got changed, reverting the fix (000c0a38a8965a20fbd3a44b6424526609bb64ea) for #721. Mixing multiple different functional changes in a single commit isn't ideal to start with, but I don't agree with this comment: 

https://github.com/OpenRefine/OpenRefine/blob/5ca5f3cb7b6cc1be1813e9e196a3c0f669298a4c/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java#L286-L290

I believe users will most commonly think of this as the inverse of the request rate (ie a delay of 500 msec is equivalent to 2 requests/sec). As a matter of fact, perhaps a rate would be a better way to specify it. Since the requests are issued in a single threaded fashion, there is already natural throttling going on based on the server's response time. Adding an additional delay to that makes it impossible to hit a consistent request rate due to the variability.

Longer term we probably want to let the user specify a rate per second/minute/hour, maximum number of requests outstanding at a time, and handle 429 responses when being throttled, but for now we should fix the delay calculation.

tfmorris:
Fixes #3129. 

Makes the following improvements to the Fetch URL code:
- switches to Apache [HTTP Client 5.0.2](https://www.apache.org/dist/httpcomponents/httpclient/RELEASE_NOTES-5.0.x.txt) including support for HTTP/2 as well as automatic retries and other general improvements
- reuses HTTP client across requests for the entire operation rather than rebuilding it for every request to improve connection reuse and not pay connection setup cost every time (especially bad for HTTPS)
- honors Retry-After header (built-in support), but if none specified uses binary exponential backup (default is single retry at 1 sec.)
- computes interrequest delay using a request interceptor

~~Because the namespace was changed, the new client can coexist with the old client, but we should probably abstract this out into a reusable class that can be reused by project creation, reconciliation, etc.~~

I went ahead and split things out into their own reusable class and extended it to support POSTs as well as GETs so that it could be used for reconciliation too.

### DIFF
--- a/main/src/com/google/refine/model/recon/ReconciledDataExtensionJob.java
+++ b/main/src/com/google/refine/model/recon/ReconciledDataExtensionJob.java
@@ -37,29 +37,14 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package com.google.refine.model.recon;
 
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.StringWriter;
 import java.io.Writer;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
-
-import org.apache.http.Consts;
-import org.apache.http.NameValuePair;
-import org.apache.http.StatusLine;
-import org.apache.http.client.config.RequestConfig;
-import org.apache.http.client.entity.UrlEncodedFormEntity;
-import org.apache.http.client.methods.CloseableHttpResponse;
-import org.apache.http.client.methods.HttpPost;
-import org.apache.http.impl.client.CloseableHttpClient;
-import org.apache.http.impl.client.HttpClientBuilder;
-import org.apache.http.impl.client.HttpClients;
-import org.apache.http.impl.client.LaxRedirectStrategy;
-import org.apache.http.message.BasicNameValuePair;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonInclude;
@@ -69,13 +54,14 @@ import com.fasterxml.jackson.annotation.JsonView;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import com.google.refine.RefineServlet;
 import com.google.refine.expr.functions.ToDate;
 import com.google.refine.model.ReconCandidate;
 import com.google.refine.model.ReconType;
+import com.google.refine.util.HttpClient;
 import com.google.refine.util.JSONUtilities;
 import com.google.refine.util.JsonViews;
 import com.google.refine.util.ParsingUtilities;
+
 
 public class ReconciledDataExtensionJob {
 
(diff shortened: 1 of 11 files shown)


tfmorris:
OK, I've still got some testing to do, but I'd like to get some review feedback in parallel.

Along the way, this kind of morphed from "fix rate limiting" to implement support for Apache HTTP Client 5, HTTP/2 and refactor all our our previous 5 implementations (Create Project, Fetch URL, Guess Types, Reconcile, Extend Data) into a single set of common code.

One thing which probably could be improved is client/connection pool reuse for reconciliation calls, but this is no worse than the current implementation.

main/src/com/google/refine/model/recon/ReconciledDataExtensionJob.java
```diff
@@ -218,46 +203,17 @@ public ReconciledDataExtensionJob(DataExtensionConfig obj, String endpoint) {
         return map;
     }
 
-    /**
-     * @todo this should be refactored to be unified with the HTTP querying code
-     * from StandardReconConfig. We should ideally extract a library to query
-     * reconciliation services and expose it as such for others to reuse.
-     */
```
wetneb, line 225: I would leave this comment since this refactoring does not address this yet.

main/src/com/google/refine/model/recon/StandardReconConfig.java
```diff
@@ -475,51 +463,40 @@ private CloseableHttpClient getHttpClient() {
...
+                    String text = job.text;
+                    String key = "q" + i;
+                    if (o.has(key) && o.get(key) instanceof ObjectNode) {
+                        ObjectNode o2 = (ObjectNode) o.get(key);
+                        if (o2.has("result") && o2.get("result") instanceof ArrayNode) {
+                            ArrayNode results = (ArrayNode) o2.get("result");
+
+                            recon = createReconServiceResults(text, results, historyEntryID);
                         } else {
-                            // TODO: better error reporting
```
wetneb, line 511: I'd also leave this TODO since error reporting should not be in the server logs: users should be notified about it.

wetneb (approved):
Thank you! Upgrading to a newer version of the Apache HTTP client looks sensible.

I am not too sure about introducing our own internal HTTP client:
* it could make it a bit harder to do the other refactoring I was thinking about (described by the comment you removed, see below), since that would make the reconciliation code less independent from the rest of the code base
* it seems quite hard to do it entirely, for other places in the code base where we do HTTP requests (creating a project from a URL, for instance), since those will have their own requirements (such as response streaming), so I suspect this will not fully let us isolate the HTTP library away.
But I am not fiercely opposed to the move either, it does avoid some duplication.

main/src/com/google/refine/model/recon/ReconciledDataExtensionJob.java
tfmorris: My thought here was that the refactoring in this PR was a superset of the suggest recon-only client. Is that not true? What special requirements does reconciliation have? The code that I thought the comment was referring to is just a single line now. 

The Create Project streaming is a slightly different use, but it's already accommodated in the current code.

I can add a version of the comment back if it's still relevant, but perhaps we should update it to reflect the current state of affairs.

main/src/com/google/refine/model/recon/StandardReconConfig.java
tfmorris: I think you are just seeing a diff artifact. The exact same text is a couple of lines down, but I've added a second copy.

main/src/com/google/refine/model/recon/ReconciledDataExtensionJob.java
wetneb: I was thinking about a java library to query reconciliation services. You would be able to construct a reconciliation query with POJOs and send it to a service described by a URL (without having to worry about the JSON serialization of that reconciliation query, for instance). This is more high-level than a HTTP library: it would abstract away the specifics of how reconciliation clients and servers communicate over HTTP.

Such a library could also be the basis of a framework to build reconciliation services.

tfmorris: That sounds like it has very little, if anything to do with OpenRefine, but I'll put the comment back with a clarification that it's talking about a 3rd party library.

tfmorris: I've restored the comment, moved it so it's not on a single line method, and updated it with my limited understanding of what the task is. I've also rebased it to pick up the new CI code and will merge it when the tests pass.

PR #3242, opened 2020-10-01
Update index.vt

shivaligakhar-123:


tfmorris:
Thanks for the contribution, but we don't have any open issues requesting a change in the size of the logo.

You can find open issues that are good first issues to work on here: https://github.com/OpenRefine/OpenRefine/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

PR #3245, opened 2020-10-02
Travis CI build config changes

tfmorris:
Refs #3208

- Enable Java 16 (early access) builds now that we have Jacoco 0.8.6 with support for it
- Allow flaky Travis CI Mac build to fail 
- Add Java 14 Linux build to replace Mac
- Only do full build matrix on pushes to master (mostly PR merges) or release (e.g. v3.5) branches.
- Pull requests against master and pushes to feature branches only get a single Java 8 build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,20 @@ dist: bionic
 
 jobs:
   include:
-    - dist: trusty # Fastest build first
+    - dist: trusty # Fastest build first & for all builds
       jdk: oraclejdk8 # Trusty default
-    - jdk: openjdk11 # Bionic default
-    - jdk: openjdk13
-    - os: osx
+    # Full matrix only for merges to master or anything to do with release branches e.g. v3.5
+    - if: (branch = master AND type != pull_request) OR branch ~= /^v\d\.\d.*/
+      jdk: openjdk11 # Bionic default
+    - if: branch = master AND type != pull_request OR branch ~= /^v\d\.\d.*/
+      jdk: openjdk12
+      dist: xenial # just for a little variety
+    - if: branch = master AND type != pull_request OR branch ~= /^v\d\.\d.*/
+      jdk: openjdk13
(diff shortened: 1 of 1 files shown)


tfmorris:
When merged this should do a full build matrix on all platforms.

wetneb (approved):

.travis.yml
```diff
@@ -6,9 +6,18 @@ jobs:
   include:
     - dist: trusty # Fastest build first
       jdk: oraclejdk8 # Trusty default
-    - jdk: openjdk11 # Bionic default
-    - jdk: openjdk13
-    - os: osx
+    # Whitelist release branches (e.g. v3.5)so they build on all platforms
```
thadguidry, line 9: we should probably stop using colors for allowlist/denylist

wetneb: Which colors and which allow/deny lists?

thadguidry: The comment with `Whitelist` mentioned.
Btw, I've asked CS&S about general concerns on their Slack just so we have some guidance on this from them.  I'm not especially worried because I know human society is slowly changing and becoming "woke" or "aware" of bias in many things.  This is normal, societies improve over time and generally become more caring and as they continue to learn from mistakes from the past.

Anyways, let's not get fully unfocused here.  It's just a comment for now, and let's wait to see what CS&S has to say because it will be helpful guidance for all.

wetneb: Oh right, okay. That's simple enough. @tfmorris could you change that comment to "Allow release branches (e.g. v3.5) to build on all platforms"?

tfmorris: It's a USA thing.

@wetneb You have to scroll back up to see it, but the "Outdated" tag shows that this was already changed.

tfmorris:
I confirmed that the merge got built on all platforms: 
- Oracle JDK 8 on Trusty
- OpenJDK 11,12,13,14, 16 on a mix of Bionic & Xenial
- Oracle JDK 14 on Mac (which failed).

PR #3087, opened 2020-08-17
Update to jQuery 3.6.0

tfmorris:
Fixes #3000 

All the changes should be backward compatible with jQuery 1.19 since they're mostly things that we deprecated ages ago. Some things are just plain broken like missing close tags in HTML which jQuery is now being stricter about.

The only change I knowingly left out was changing $.trim(string) to string.trim() since that was just deprecated in jQuery 3.5, so we've got plenty of time before it actually gets removed.

There are also a number of jqMIGRATE warnings which are being triggered by jQuery UI and other widgets like tablesorter, lavalamp, etc.

PR #3249, opened 2020-10-05
Make it possible to view more than 50 rows

Abbe98:
Adding the options 100, 500, 1000.

**Motivation**

I have in my own OpenRefine had these options for a while and I have found that it saves scrolling and context switching time when one needs to manually go through large amounts of data, by making sure one does not need to scroll up to access the pagination navigation.

**Notes**

I have tested the performance of this on two computers and I don't think even the 1000 option will be problematic on most modern computers.

The only past issue on the subject I'm aware of is #3087

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
+++ b/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
@@ -186,7 +186,7 @@ DataTableView.prototype._renderPagingControls = function(pageSizeControls, pagin
   }
 
   $('<span>'+$.i18n('core-views/show')+': </span>').appendTo(pageSizeControls);
-  var sizes = [ 5, 10, 25, 50 ];
+  var sizes = [ 5, 10, 25, 100, 500, 1000 ];
   var renderPageSize = function(index) {
     var pageSize = sizes[index];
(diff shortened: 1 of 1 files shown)


tfmorris:
Why isn't the solution from #2624 adequate?

thadguidry:
Hi @Abbe98 You can take a look at the listing of preferences that can be set on our wiki page here:  https://github.com/OpenRefine/OpenRefine/wiki/Preferences

Abbe98:
This has a much lower discovery barrier and I believe that broader defaults
would actually be useful for many users and use cases.

The preference page has quite a high discovery barrier, and it's currently
not possible to use it without familiarity or its documentation.



Den mån 5 okt. 2020 kl 20:51 skrev Thad Guidry <notifications@github.com>:

wetneb:
Just noting that I would hope that we get the infinite scrolling PR merged before the next release (#2746), in which case there will not be any page size selection any longer, so this would only benefit people who run OpenRefine from the master branch (or snapshot releases) in the meantime.

Abbe98:
A year later I still think this is a good idea, it provides better and much wanted defaults. The settings are still very inaccessible for most and the proposed defaults can be managed by most computers.

PR #3257, opened 2020-10-11
Narrow permissiveness for fields on null & catch more GREL syntax errors

### LINKED ISSUE #169: No syntax error not given for GREL value.endsWith("ROAD")))))*&(&#37;!#(&)@
_Original author: thadguidry (October 26, 2010 21:00:34)_

Expectation here is to FAIL with syntax error for any additional chars that follow a complete and valid expression defined by 'value' or outer parentheses.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=169_

### LINKED ISSUE #2507: GREL parsing is too lazy and doesn't throw syntax errors
**Describe the bug**
Arrays are not parsed properly for syntax issues.

**To Reproduce**
1. Go to Edit Cells -> Transform
2. Paste the example GREL `[1,['a','b','c']]`
3. Notice no Parsing error is given and output is `[ 1, [ "a", "b", "c" ] ]`
4. Paste the example GREL `[1['a','b','c']]`
5. Notice it evaluates to `[ null ]`

**Expected behavior**
I would expect the GREL `[1['a','b','c']]` would throw a Parsing exception as it currently does for something like `[1,['a','b','c']3]` where the 3rd element `3` in the outer array is missing a comma before it.  Instead, `['a','b','c']` as an index token of the element `[1]` is evaluated to `null` instead of `IndexOutOfBoundsException`

 - OS: Windows 10
 - Browser Version: Firefox
 - JRE or JDK Version: JDK8

 - Version [e.g. OpenRefine 3.3 Beta]

**Additional context**
Interestingly, GREL doesn't seem to fully evaluate all the elements in a continuing expression.  For instance, GREL `['a','b','c']3]`  does not advise the user of a Parsing exception and ignores remaining part of the expression `3]`.

Additionally, 
1. `[a]` evaluates to `[null]` where `a` was not even evaluated properly.
2. `['a']` evaluates to `["a"]` as expected.
3.  `1[]` evaluates to `null` as unexpected.
4. `1['a']` evaluates to `null` as unexpected.
5. `1[0]` evaluates to `1` as expected
6. `1[1]` evaluates to `Error: java.lang.StringIndexOutOfBoundsException: String index out of range: 2` as expected.

tfmorris:
Fixes #169.  Fixes #2507. refs #1036

This change retains `null.value` as a special case which returns `null` but restores previous behavior of evaluating to an error for other field names (e.g. `null.foo`) as well as attempts to access fields on objects that don't support it (e.g. `[1,2].foo`).

Adding tests for the latter piece led to a bit of an odyssey fixing parser errors which were silently ignored (1.foo, 1.1.foo, etc).
Also enabled some commented out GREL tests which failed (both due to exceptions and to silently ignoring errors).

Errors which were silently ignored before, but which will now error include:
- two adjacent numbers or strings without an intervening operator or delimiter (`1 1` or `'a' 'b'`)
- unsupported operators like assignment (`=`) or exponentiation (`^`)
- scientific notation (`1e3`)
- malformed numbers (`1.xyz` would be interpreted as 1 with the rest thrown away)

Also added tests for `get()`, including coverage for the example in #2507, and moved the dot notation tests back to `GrelTests` where they belong (I incorrectly moved them recently). The parameter checking in `get` needed tightening up to fix #2507, but I left the majority of the unusual behavior (e.g. floating point array indices) unchanged and just added tests which match the current behavior.

Finally, added a test case for #169 (which had already been fixed by the above changes)

### DIFF
--- a/main/src/com/google/refine/grel/Parser.java
+++ b/main/src/com/google/refine/grel/Parser.java
@@ -111,6 +111,10 @@ public class Parser {
     protected Evaluable parseSubExpression() throws ParsingException {
         Evaluable sub = parseTerm();
 
+        if (_token != null && _token.type != TokenType.Operator && _token.type != TokenType.Delimiter) {
+            throw makeException("Expecting an operator or delimiter here");
+        }
+
         while (_token != null &&
                 _token.type == TokenType.Operator &&
                 "+-".indexOf(_token.text) >= 0) {
@@ -133,6 +137,10 @@ public class Parser {
      */
     protected Evaluable parseTerm() throws ParsingException {
         Evaluable factor = parseFactor();
+
+        if (_token != null && _token.type != TokenType.Operator && _token.type != TokenType.Delimiter) {
+            throw makeException("Expecting an operator or delimiter here");
+        }
 
         while (_token != null &&
                 _token.type == TokenType.Operator &&
@@ -248,7 +256,9 @@ public class Parser {
         }
 
         while (_token != null) {
-            if (_token.type == TokenType.Operator && _token.text.equals(".")) {
+            if (_token.type == TokenType.Error) {
+                throw makeException("Parse error");
+            } else if (_token.type == TokenType.Operator && _token.text.equals(".")) {
                 next(false); // swallow .
 
                 if (_token == null || _token.type != TokenType.Identifier) {
--- a/main/src/com/google/refine/grel/ast/FieldAccessorExpr.java
+++ b/main/src/com/google/refine/grel/ast/FieldAccessorExpr.java
@@ -37,6 +37,7 @@ import java.util.Properties;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.refine.expr.EvalError;
 import com.google.refine.expr.Evaluable;
 import com.google.refine.expr.ExpressionUtils;
 import com.google.refine.expr.HasFields;
@@ -62,14 +63,17 @@ public class FieldAccessorExpr implements Evaluable {
         if (ExpressionUtils.isError(o)) {
             return o; // bubble the error up
         } else if (o == null) {
-            return null;
+            if ("value".equals(_fieldName)) {
+                return null;
+            }
+            return new EvalError("Cannot retrieve field from null");
         } else if (o instanceof HasFields) {
             return ((HasFields) o).getField(_fieldName, bindings);
         } else if (o instanceof ObjectNode) {
-        	JsonNode value = ((ObjectNode) o).get(_fieldName);
-        	return JsonValueConverter.convert(value);
+            JsonNode value = ((ObjectNode) o).get(_fieldName);
+            return JsonValueConverter.convert(value);
         } else {
-            return null;
+            return new EvalError("No accessible fields available");
         }
     }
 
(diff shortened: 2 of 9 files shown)


main/src/com/google/refine/grel/ast/FieldAccessorExpr.java
```diff
@@ -62,14 +63,17 @@ public Object evaluate(Properties bindings) {
...
         } else if (o instanceof HasFields) {
             return ((HasFields) o).getField(_fieldName, bindings);
         } else if (o instanceof ObjectNode) {
-        	JsonNode value = ((ObjectNode) o).get(_fieldName);
-        	return JsonValueConverter.convert(value);
+            JsonNode value = ((ObjectNode) o).get(_fieldName);
+            return JsonValueConverter.convert(value);
         } else {
-            return null;
+            return new EvalError("No accessible fields available");
```
thadguidry, line 76: Hmm, I wonder if we can make this wording more noob friendly? Something a bit more meaningful to a beginning data journalist or librarian learning GREL.  What does "accessible even mean", what is a "field"?  Our product reference will have to explain this deeply.

main/src/com/google/refine/grel/Parser.java
```diff
@@ -111,6 +111,10 @@ protected Evaluable parseExpression() throws ParsingException {
     protected Evaluable parseSubExpression() throws ParsingException {
         Evaluable sub = parseTerm();
 
+        if (_token != null && _token.type != TokenType.Operator && _token.type != TokenType.Delimiter) {
+            throw makeException("Expecting an operator or delimiter here");
```
thadguidry, line 115: This is friendly. Like it!

main/src/com/google/refine/grel/ast/FieldAccessorExpr.java
tfmorris: Feel free to offer a suggestion based on your experience with users who are just starting out (which I assume is what you meant rather than "noob").

thadguidry: Sorry, was in the middle of a PC game while I was looking at this ... "lingo" within that domain. :-) 
I'm not sure what I would say either.  Hmm... let me think on it.  "field" has been the main thing that made heads spin, and wondering if we could change that term somehow.

thadguidry: Hmm, variable? a modifier? an attribute? https://en.wikipedia.org/wiki/Field_(computer_science)  https://en.wikipedia.org/wiki/Instance_variable

wetneb:
Thanks for this! The parsing improvements seem very welcome, but the change of behaviour on `null.xyz` seems unrelated (and more contentious given the history on this), so I am wondering why do both in the same PR?

main/src/com/google/refine/grel/ast/FieldAccessorExpr.java
thadguidry:
```suggestion
            return new EvalError("You have asked for a non-existent field from the object or tuple, or there is no output for your expression. You might want to double check things, or ensure your working with a valid cell/row object or tuple");
```

I think since this is Evaluable, we should say something about the output of the expression.  I mentioned something like the above suggestion in #1036 and have made more concise.

thadguidry (changes requested):

wetneb:
I am closing this pull request since there has not been recent progress on it, but of course we can reopen it whenever someone wants to pick it up.

PR #3261, opened 2020-10-13
Update to docusaurus-2.0.0-alpha.64 to fix security vuln in dependency

### LINKED ISSUE #3031: Update docusaurus to more recent version to validate internal links
It would be nice if the Docusaurus build could fail when the docs contain broken internal links. This is enabled by default in more recents versions of Docusaurus v2:
https://v2.docusaurus.io/docs/docusaurus.config.js#onbrokenlinks

The feature should initially be disabled while @allanaaa writes up the missing pages.

We could also consider validating external links.

**Additional context**
Discussed during a documentation meeting

tfmorris:
Also update docusaurus config to match new version and fixed some broken links in the documentation, but not sure
if the latter were related to the version change.

Not really much of a risk since it's in only in our documentation build system, but this will let us getting rid of the annoying nag.

Fixes #3031

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -1,4 +1,5 @@
 ---
+slug: /
 id: index
 title: OpenRefine product reference
 sidebar_label: Index
--- a/docs/docs/manual/installing.md
+++ b/docs/docs/manual/installing.md
@@ -70,7 +70,7 @@ Take note of the [extensions](#installing-extensions) you have currently install
 
 ### Install or upgrade Java
 
-[Java Development Kit (JDK)](https://jdk.java.net/) is required to run OpenRefine and should be installed first. [OpenRefine installation packages for Mac and Windows come bundled with JDK]([https://openrefine.org/download.html](https://openrefine.org/download.html)), so you do not need to install it separately if you use those bundles. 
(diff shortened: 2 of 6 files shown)


tfmorris:
Hmmn, looks like my local dev environment isn't as picky as CI, so converting back to draft.

@allanaaa @ostephens I would welcome help migrating the doc sources to be compatible with the new version so that we can check the box on this security vulnerability.

ostephens:
@tfmorris I think most of the Netlify deploy issues are related to image filename case-sensitivity and caused by some of the png files having .PNG extension. I've fixed this (and a couple of incorrect links) in PR #3268 - if that's merged and then you rebase then I think with the changes you've already made in this branch it should build successfully 🤞 

@allanaaa probably just worth noting that image files in /static/img need to have the same case as any references to those files from inside the markdown - I'd suggest we standardise to always using lowercase for the file extensions

wetneb:
I have not checked, but there could also be issues detected by link validation: #3031.

ostephens:
@wetneb you are right, but I think with my corrections in #3268 and the ones that @tfmorris has already fixed in this PR then we have got them all

tfmorris:
That rename operation made my case-preserving, but case-insensitive file system very unhappy. Let's avoid wonky file extensions in the future please!

Thanks for the fixups. I rebased the branch and fixed a couple of other broken links before realizing that #3031 contained the hint as to what changed in the new version. Disabling the default link checking fixed the build. I'll leave it to the pros to repair the broken links.

Should be ready for review now.

wetneb (approved):

PR #3268, opened 2020-10-16
Fix image filenames and links in documentation

ostephens:
Image filenames case sensitive when building with Netlify


wetneb (approved):
Thanks Owen!

PR #3285, opened 2020-10-23
Preventing addition of any empty cells with whitespaces while importing XML data : Resolves #1095

### LINKED ISSUE #1095: Open XML file from URL generates lots of empty lines
In the 2.6RC2 installed on Linux from https://github.com/OpenRefine/OpenRefine/releases/download/2.6-rc.2 I get blank lines displayed when importing XML from a URL.

eg importing XML from  http://api.worldbank.org/countries/all/indicators/SP.POP.TOTL?date=2000:2001

image

Empty lines:

image

If the parsing of this can't be fixed, an `ignore empty rows` setting would be useful?

maheshjindal:
This Pull Request fixes the issue related to the addition of any 'redundant empty cells/records with whitespaces' during preview while importing XML data from a file or any URL: Issue #1095


maheshjindal:
Also, the attached test cases are failing from the latest master branch code during the build.

Screenshot 2020-10-23 at 11 47 27 PM

tfmorris:
As you can see here, the build on master is green: https://travis-ci.com/github/OpenRefine/OpenRefine/branches
so the test failures are specific to this branch and would need to be fixed, but I'm not sure the approach is correct, so I'd suggest going back to the original issue and reviewing the history for what's been done already.

PR #3294, opened 2020-10-31
changed to reflect the function's acceptance of either simple string …

### LINKED ISSUE #3291: Dead code + inaccuracy in the Find function documentation
In #2095 we updated the `find` function to support two types of arguments: regex (`/[a-z]/`) and simple string (`"[a-z]"`).

There are two things we need to clean up there:
* The embedded description of the function does not reflect that:
https://github.com/OpenRefine/OpenRefine/blob/ac7b5a0a19c1ecaefa1ac1d9d0d26ea9baceacee/main/src/com/google/refine/expr/functions/strings/Find.java#L75
* There is a bit of dead code here:
https://github.com/OpenRefine/OpenRefine/blob/ac7b5a0a19c1ecaefa1ac1d9d0d26ea9baceacee/main/src/com/google/refine/expr/functions/strings/Find.java#L59
(at this point the `p` variable cannot be a `String`, we already know it's a `Pattern`).

rubyAnneB:
…or regex

closes #3291

### DIFF
--- a/main/src/com/google/refine/expr/functions/strings/Find.java
+++ b/main/src/com/google/refine/expr/functions/strings/Find.java
@@ -55,7 +55,6 @@ public class Find implements Function {
             }
             
             if (s != null && p != null && p instanceof Pattern) {
-
                 Pattern pattern = (Pattern) p;
                 Matcher matcher = pattern.matcher(s.toString());
                 
@@ -71,7 +70,7 @@ public class Find implements Function {
     
     @Override
     public String getDescription() {
-        return "Returns all the occurances of match given regular expression or simple string";
+        return "Returns all the occurrences of match given regular expression or simple string";
     }
     
     @Override
--- a/main/src/com/google/refine/expr/functions/strings/Find.java
+++ b/main/src/com/google/refine/expr/functions/strings/Find.java
@@ -55,10 +55,8 @@ public class Find implements Function {
             }
             
             if (s != null && p != null && p instanceof Pattern) {
-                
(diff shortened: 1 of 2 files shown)


thadguidry:
Thanks @rubyAnneB as you can see in the LGTM analysis output, you have a build failure:
```
[2020-10-31 02:55:07] [build] [2020-10-31 02:55:07] [autobuild] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project main: Compilation failure
[2020-10-31 02:55:07] [build] [2020-10-31 02:55:07] [autobuild] [ERROR] /opt/src/main/src/com/google/refine/expr/functions/strings/Find.java:[59,36] cannot find symbol
[2020-10-31 02:55:07] [build] [2020-10-31 02:55:07] [autobuild] [ERROR]   symbol:   method matcher(java.lang.String)
[2020-10-31 02:55:07] [build] [2020-10-31 02:55:07] [autobuild] [ERROR]   location: variable p of type java.lang.Object
```

Think you can sort it out and push the necessary changes to fix the code?
(Hint: Does using the type comparison operator `instanceof` convert or type cast things also? Or merely check something is a type?  How are things different between `instanceof` and `isInstance()` method during runtime versus compile time?)

wetneb:
You are missing a cast here: we only know that the object is an instance of this class because we do a runtime check in the `if` condition above, but that's not enough for the compiler to detect that this is type safe.

wetneb (approved):

wetneb:
Thanks a lot @rubyAnneB and congratulations for your first pull request in OpenRefine!

PR #3308, opened 2020-11-03
Fix Reinterpret missing documentation

### LINKED ISSUE #3292: Unit tests for the "reinterpret" GREL function
The `reinterpret` function is supposed to help fix encoding errors, by encoding and re-decoding strings to their correct encoding.

We want to unit tests for this function, demonstrating how it can be useful. Some of these examples should be included in the documentation.

@ostephens reports mixed success with using this function in practice.

The documentation of this function should mention than in most cases, encoding errors should be solved by re-importing the project in the correct encoding. This function should only be used when dealing with inconsistent encoding in a given file.

thadguidry:
Helps fix #3292

### DIFF
--- a/main/src/com/google/refine/expr/functions/strings/Reinterpret.java
+++ b/main/src/com/google/refine/expr/functions/strings/Reinterpret.java
@@ -66,7 +66,7 @@ public class Reinterpret implements Function {
                 return reinterpret(str, decoder, encoder);
             }
         }
-        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects String to reinterpret with a given target encoder and optional source decoder");
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects String to reinterpret with a given target encoding and optional source encoding");
     }
 
     private Object reinterpret(String str, String decoder, String encoder) {
--- a/main/src/com/google/refine/expr/functions/strings/Reinterpret.java
+++ b/main/src/com/google/refine/expr/functions/strings/Reinterpret.java
@@ -102,7 +102,7 @@ public class Reinterpret implements Function {
     
     @Override
     public String getParams() {
-        return "string s, string target encoder, string source decoder";
+        return "string s, string target encoding, string source encoding";
     }
     
     @Override
--- a/main/src/com/google/refine/expr/functions/strings/Reinterpret.java
+++ b/main/src/com/google/refine/expr/functions/strings/Reinterpret.java
@@ -66,7 +66,7 @@ public class Reinterpret implements Function {
                 return reinterpret(str, decoder, encoder);
             }
         }
-        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects 2 or 3 arguments");
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects String to reinterpret with a given target encoder and optional source decoder");
     }
 
     private Object reinterpret(String str, String decoder, String encoder) {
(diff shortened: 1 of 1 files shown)


main/src/com/google/refine/expr/functions/strings/Reinterpret.java
```diff
@@ -66,7 +66,7 @@ public Object call(Properties bindings, Object[] args) {
                 return reinterpret(str, decoder, encoder);
             }
         }
-        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects 2 or 3 arguments");
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects String to reinterpret with a given target encoder and optional source decoder");
```
thadguidry, line 69: Hmm, perhaps we should make this more user friendly... since the [original enhancement on 2.5 says](https://github.com/OpenRefine/OpenRefine/wiki/Changes-for-2.5):
> reinterpret() now supports optionally specifying a source encoding to be used instead of the project encoding - reinterpret("target encoding", "source encoding")

Maybe
```suggestion
        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects String to reinterpret with a given target encoding and optional source encoding");

File: main/src/com/google/refine/expr/functions/strings/Reinterpret.java [hunk 2]
```diff
@@ -89,20 +89,20 @@ private Object reinterpret(String str, String decoder, String encoder) {
...
+            return new EvalError(ControlFunctionRegistry.getFunctionName(this) + ": target encoding '" + encoder + "' is not available or recognized.");
         }
                         
         return result;
     }
     
     @Override
     public String getDescription() {
-        return "Returns s reinterpreted thru the given encoder.";
+        return "Returns s reinterpreted thru the given target encoder and optional source decoder.";
```
thadguidry, line 100: And this is a description of the function, which indeed describes well what it actually does, so keep it worded as this PR, or do we tweak to make friendly as well?

main/src/com/google/refine/expr/functions/strings/Reinterpret.java
```diff
@@ -89,20 +89,20 @@ private Object reinterpret(String str, String decoder, String encoder) {
...
     }
     
     @Override
     public String getParams() {
-        return "string s, string encoder";
+        return "string s, string target encoder, string source decoder";
```
thadguidry, line 105: This might read better for our users as the enhancement was worded, instead of my PR wording?
```suggestion
        return "string s, string target encoding, string source encoding";
```

wetneb:
Yes I'd say encod**ing** rather than encod**er**.

main/src/com/google/refine/expr/functions/strings/Reinterpret.java
tfmorris: Yes, definitely "encoding" not "encoder," or, more specifically, "character encoding"

main/src/com/google/refine/expr/functions/strings/Reinterpret.java
thadguidry: OK sounds like the consensus is to phrase it this way as well, since this is user facing in the Help panel of Expression Editor:
```suggestion
        return "Returns s reinterpreted using a target encoding and optional source encoding.";
```

tfmorris:
The source branch for this should be a branch in your cloned repo, so closing this so it can be recreated with the correct source branch. Please fix the test failures as well.

PR #3315, opened 2020-11-04
add clarity for reinterpret docs

### LINKED ISSUE #3292: Unit tests for the "reinterpret" GREL function

thadguidry:
Helps with #3292

### DIFF
--- a/main/src/com/google/refine/expr/functions/strings/Reinterpret.java
+++ b/main/src/com/google/refine/expr/functions/strings/Reinterpret.java
@@ -66,7 +66,7 @@ public class Reinterpret implements Function {
                 return reinterpret(str, decoder, encoder);
             }
         }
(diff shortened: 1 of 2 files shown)


thadguidry:
@wetneb I guess I should clean up the json in the test now as well to fix the failure?  Or just leave that for later when someone else fixes #3292 ?

wetneb (approved):
Thanks!

Issue #3293, opened 2020-10-30
Reconciliation dialog: clicking on "Reconcile against no particular type" should select this option


wetneb:
We are missing a `<label>` tag properly associated with radio buttons in the reconciliation dialog. We should check that all labels are correctly marked as such on this dialog.

### To Reproduce
1. First, click "Reconcile" -> "Start reconciling" and choose a service
2. Then, click on the text "Reconcile against no particular type" (not the radio button, the text itself)

### Current Results
Nothing happens.

### Expected Behavior
The corresponding radio button should be selected.

### Versions
 - Browser Version: Firefox 68
 - OpenRefine: master branch (3.5 snapshot)

rachittiwari8562:
I want to work on this issue.

wetneb:
Thanks for volunteering! Looking forward to your pull request. Let me know if you get stuck at any point.

rachittiwari8562:
I have created a pull request for this issue.Please review it.If you find anything wrong please let me know.

PR #3327, opened 2020-11-10
fix for issue #3293

rachittiwari8562:
First i have assigned IDS to both the input field according to their respective meaning.
Then we assign to both the input field according to IDS.


wetneb (approved):
Thank you very much, that's perfect. Congratulations for your first PR on OpenRefine and welcome to the team!

PR #3336, opened 2020-11-16
Docs > Exporting

allanaaa:
Very short page. Expanded on the two customizable outputs. May expand on the SQL statements export as well.


docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,76 @@ sidebar_label: Exporting
...
+
+
+### Templating exporter
+
+If you pick <span class="menuItems">Templating…</span> from the <span class="menuItems">Export</span> dropdown menu, you can “roll your own” exporter. This is useful for formats that we don't support natively yet, or won't support. For example, the Templating exporter generates JSON by default. 
+
+The window that appears allows you to set your own separators, prefix, and suffix to create a complete dataset in the language of your choice. In the Row Template section, you can choose which columns to generate from each row by calling them with variables. 
+
+![A screenshot of the Templating exporter generating JSON by default.](/img/templating-exporter.png)
+
```
wetneb, line 66: I think it would be worth expanding a bit here, on the following topics:
* the syntax of the templates (how do you introduce a template bit? how do you escape braces?)
* the evaluation context of the expressions used in the template: what variables can you use there?
* can you use other expression languages than GREL inside the templates?

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,76 @@ sidebar_label: Exporting
...
+With the custom tabular exporter, you can choose which of your data to export, the separator you wish to use, and whether you'd like to download it to your computer or upload it into a Google Sheet. This exporter is especially useful with reconciled data, as you can choose whether you wish to output the cells' original values, the matched values, or the matched IDs. 
+
+You can also choose how to output date-formatted cells, and whether you wish to export all the project's data or just the current view set. 
+
+![A screenshot of the custom tabular file download tab.](/img/custom-tabular-exporter2.png)
+
+With the download options, you can generate a preview of how the first ten rows of your dataset will output. If you do not choose one of the file formats on the right, the "Download" button will generate a text file. 
+
+With the "Option Code" tab, you can copy JSON data that saves your current settings, to reuse on another project, or you can paste in existing JSON settings data to apply to the current project.
```
wetneb, line 46: "Option code" should be displayed as UI text I would say.

docs/docs/manual/exporting.md
allanaaa: Thanks for these suggestions. I'm able to expand on the variables and some syntax, but: _can_ you use other expression languages? I wouldn't know where to begin figuring that out on my own. Is there a jsonize() equivalent in clojure? I can't seem to get any clojure syntax to work but that's hardly an indicator of anything.

allanaaa: My mistake, you've tricked me. How _do_ you escape a curly brace? Our regular approach is useless here.

wetneb: Good points, I don't actually know the answer to any of these questions! I'll investigate in the code and report back.

wetneb: So, looking at the code (I haven't checked that it does work, it'd be useful if you do)
- The characters `\`, `{` and `$` can be escaped with `\` (so, `\\`, `\{` and `\$` respectively produce `\`, `{` and `$`).
- You can use `${myColumn}` to insert values from a given column
- You can use `{{ some expression }}` to insert values obtained by any expression. By default the expression is expected to be written in GREL, but you can use other languages too: `{{jython:return cells["myColumn"].value}}` should work as well (and same for clojure).
It's not clear to me if there is a simple way for users to figure out the exact language identifier (here `jython`) to use - it can be observed in the history when you use these expressions in other operations, but that is not very convenient. Perhaps we could just list the available language identifiers in this part of the documentation for convenience? Or in the general documentation about expressions if this is useful elsewhere.

allanaaa: I don't have any luck escaping curly braces in regex with a backslash. 
`value.match(/.*(\d\d\d\d)/)` works fine but `value.match(/.*(\d\{4\})/)` does not resolve - it just outputs the entire {{}} formula as text.

The statement `value.find(/[,.!{@#$\?]/)` doesn't need the opening curly brace escaped. But it won't accept a CLOSING curly brace, with or without a backslash. 
I don't seem to need to escape a $ at all - the GREL runs just fine with it in. I don't need to cancel a backslash, but I do need to cancel a forward slash with a backslash. (`value.find(/[,.!{@#\/$\?]/)`)
So, I am not sure what this has to do with the code you looked at. Help please!


Thanks for the jython though!

wetneb: I would not be surprised if this escaping code was buggy. Let's debug this together in today's meeting perhaps?

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,97 @@ sidebar_label: Exporting
 
 ## Overview
 
+Once your data is cleaned, you will need to get it out of OpenRefine and into the system of your choice. OpenRefine outputs a number of file formats, can upload your data directly into Google Sheets, and can export directly to Wikidata. 
+
```
ostephens, line 10: Would "use it to create or update Wikidata items" be a better phrasing than "export directly to Wikidata"?

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,97 @@ sidebar_label: Exporting
...
+You can also export your full project data so that it can be opened by someone else using OpenRefine (or yourself, on another computer).
 
 ## Export data
 
 Note you will only export data in the current view - that is, with current filters and facets applied.
```
ostephens, line 15: This isn't completely accurate. Some of the exports have the option to ignore current filters and facets

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,97 @@ sidebar_label: Exporting
...
+*   [SQL statement exporter](#sql-statement-exporter)
+*   [Templating exporter](#templating-exporter)
 
-*   Custom tabular export
-*   SQL
-*   Templating export \
+You can also [export reconciled data to Wikidata](wikidata#editing-wikidata-with-openrefine), or save your schema for future use with other OpenRefine projects:
 
 *   Upload edits to Wikidata
 *   Export to QuickStatement
```
ostephens, line 31: I think this should be plural `Quickstatements`. Do we also need to mention which version of Quickstatements is supported?

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,97 @@ sidebar_label: Exporting
...
 
-
-*   TSV/CSV
-*   HTML table
+*   Tab-separated value (TSV) or Comma-separated value (CSV)
+*   HTML-formatted table
 *   Excel (XLS or XLSX)
 *   ODF spreadsheet
-*   Google Sheets \
+*   Upload to Google Sheets (requires Google account authorization)
```
ostephens, line 23: How does the Google account authorization work?

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,97 @@ sidebar_label: Exporting
...
+
+![A screenshot of the custom tabular content tab.](/img/custom-tabular-exporter.png)
+
+With the custom tabular exporter, you can choose which of your data to export, the separator you wish to use, and whether you'd like to download it to your computer or upload it into a Google Sheet. 
+
+You can drag and drop the columns appearing in the column list to reorder the output. The options for reconciled and date data are applied to each column individually. 
+
+This exporter is especially useful with reconciled data, as you can choose whether you wish to output the cells' original values, the matched values, or the matched IDs. Ouputting “match entity's name”, “matched entity's ID”, or “cell's content” will output, respectively, the contents of `cell.recon.match.name`, `cell.recon.match.id`, and `cell.value`. 
+
+“Output nothing for unmatched cells” will export empty cells for both newly-created matches and cells with no chosen matches. “Link to matched entity's page” will produce hyperlinked text in an HTML table output, but have no change in other formats.
```
ostephens, line 44: Do we need to say that for "Output nothing..." option what happens for matched cells?

In second sentence I think "have no effect" rather than "have no change" would be better phrasing

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,97 @@ sidebar_label: Exporting
...
+
+![A screenshot of the SQL statement creator window.](/img/sql-exporter.png)
+
+A window with two tabs will pop up: one to define what data to output, and another to modify the statement table. This allows you to craft a complete SQL command, including DROP and IF EXISTS if you require them. 
+
+You can set a default value for empty cells after unchecking “Allow null” in one or more columns. Trimming column names will remove their whitespace characters. 
+
+![A screenshot of the SQL statement download window.](/img/sql-exporter2.png)
+
+“Include schema” means that you will start your statement with the creation of a table. Without that, you will only have an INSERT statement.
```
ostephens, line 66: I feel this would be easier to read with some more structured layout - maybe a table? I also feel we should probably list all the options together (at the moment some options are explained above the graphic and some under the graphic - I think we should keep them all together

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,97 @@ sidebar_label: Exporting
...
+
+The window that appears allows you to set your own separators, prefix, and suffix to create a complete dataset in the language of your choice. In the <span class="menuItems">Row Template</span> section, you can choose which columns to generate from each row by calling them with variables. 
+
+This can be used to:
+* output reconciliation data (`cells["column name"].recon.match.name`, `.recon.match.id`, and `.recon.best.name`, for example) instead of cell values
+* create multiple columns of output from different member fields of a single project column
+* employ GREL expressions to modify cell data for output (for example, `cells["column name"].value.toUppercase()`). 
+
+Anything that appears inside curly braces ({{}}) is treated as a GREL expression; anything outside is generated as straight text. You can include [regular expressions](expressions#regular-expressions) as usual (inside forward slashes, with any GREL function that accepts them), with the exception that you cannot use curly braces (for example, to denote a `{min,max}` capture range).
```
ostephens, line 82: I think the inclusion of curly braces in the templates is more general than just the instance of RegEx in GREL. Need to check this but we may need a more blanket statement about this

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,97 @@ sidebar_label: Exporting
...
+
+We have recipes on using the Templating exporter to [produce several different formats](https://github.com/OpenRefine/OpenRefine/wiki/Recipes#12-templating-exporter).
 
 ## Export a project
 
+You can share a project in progress with another computer, a colleague, or with someone who wants to check your history. This can be useful for showing that your data cleanup didn’t distort or manipulate the information in any way. Once you have exported a project, another OpenRefine installation can [import it as a new project](starting#import-a-project). 
+
+From the <span class="menuItems">Export</span> dropdown, select <span class="menuItems">OpenRefine project archive to file</span>. OpenRefine exports your full project with all of its history. It does not export any current views or applied facets. Any reconciliation information will be preserved, but the importing installation will need to add the same reconciliation services to keep working with that data.
```
ostephens, line 96: Is it worth us emphasising here that the fact it is the entire project means that any recipient can step the project back to its original state - so, for example, if there was confidential information in the original data import that you don't want to share, you shouldn't share the project file directly even if you removed that data while working with the project?

thadguidry: Yes, absolutely emphasize.  That point of "OpenRefine Projects contain confidential data from previous steps which is hidden but still accessible to anyone who has the OpenRefine Project file" is something that I emphasize when I give trainings.  "When you don't want to expose any history or hidden, removed, confidential data, but only the current selected data as sharable, then export in any other format besides an OpenRefine Project".

docs/docs/manual/exporting.md
wetneb: It produces QuickStatements commands in version 1.

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,97 @@ sidebar_label: Exporting
...
-*   Google Sheets \
+*   Upload to Google Sheets (requires Google account authorization)
+*   [Custom tabular exporter](#custom-tabular-exporter)
+*   [SQL statement exporter](#sql-statement-exporter)
+*   [Templating exporter](#templating-exporter)
 
-*   Custom tabular export
-*   SQL
-*   Templating export \
+You can also [export reconciled data to Wikidata](wikidata#editing-wikidata-with-openrefine), or save your schema for future use with other OpenRefine projects:
```
ostephens, line 28: The option here is to export the schema rather than "save" (only because we use "save" for the schema to mean saving it without exporting it

docs/docs/manual/exporting.md
allanaaa: "Output nothing" will do whatever you've chosen above it for matched cells. Or do you mean something else?

docs/docs/manual/exporting.md
allanaaa: Now that I think of it, the reasons why some thing are included on "Content" and some on "Download" are not clearly defined to me. The options on the "Download" tab include the option to exclude everything you decided on the "Content" tab; to me, it makes sense to present those checkboxes at the outset.
I wonder if these options should be moved around, or if I'm just missing why someone would want the workflow to be this way.

docs/docs/manual/exporting.md
allanaaa: > Anything that appears inside curly braces ({{}}) is treated as a GREL expression; anything outside is generated as straight text. You can use Jython or Clojure by declaring it at the start: for example, {{jython:return cells["Author"].value}} will run a Jython expression.
> 
> caution
> 
> Note that some syntax is different in this tool than elsewhere in OpenRefine: a forward slash must be escaped with a backslash, while other characters do not need escaping. You cannot, at this time, include a closing curly brace (}) anywhere in your expression, or it will cause it to malfunction.

Here's what we've got for now.

wetneb: Is it not worth mentioning the `${columnName}` syntax too?

allanaaa: It is, I totally forgot. Thanks!

docs/docs/manual/exporting.md
allanaaa: I think this is the only unresolved bit on this PR. If everyone could give this section a re-read, I can use some guidance on a good way to present this information. 

https://deploy-preview-3336--openrefine-docs.netlify.app/manual/exporting/#sql-exporter

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,115 @@ sidebar_label: Exporting
...
+
+![A screenshot of the custom tabular file download tab.](/img/custom-tabular-exporter2.png)
+
+On the <span class="tabLabels">Download</span> tab, you can generate a preview of how the first ten rows of your dataset will output. If you do not choose one of the file formats on the right, the <span class="buttonLabels">Download</span> button will generate a text file. On the <span class="tabLabels">Upload</span> tab, you can create a new Google Sheet. 
+
+With the <span class="tabLabels">Option Code</span> tab, you can copy JSON of your current settings to reuse on another project, or you can paste in existing JSON settings to apply to the current project. 
+
+### SQL exporter
+
+The SQL exporter creates a SQL statement containing the data you’ve exported, which you can use to overwrite or add to an existing database. A window with two tabs will pop up: one to define what data to output, and another to modify the statement table.
```
ostephens, line 57: The sentence "A window with two tabs will pop up: one to define what data to output, and another to modify the statement table." doesn't make sense to me. What is "the statement table."?
Maybe "A window with two tabs will pop up: one to define what data to output, and another to modify other aspects of the SQL statement with options to preview and download the statement"

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,115 @@ sidebar_label: Exporting
...
+
+With the <span class="tabLabels">Option Code</span> tab, you can copy JSON of your current settings to reuse on another project, or you can paste in existing JSON settings to apply to the current project. 
+
+### SQL exporter
+
+The SQL exporter creates a SQL statement containing the data you’ve exported, which you can use to overwrite or add to an existing database. A window with two tabs will pop up: one to define what data to output, and another to modify the statement table. 
+
+![A screenshot of the SQL statement creator window.](/img/sql-exporter.png)
+
+The <span class="tabLabels">Content</span> tab allows you to craft a complete SQL statement, including DROP and IF EXISTS if you require them.
```
ostephens, line 61: The options to include DROP and IF EXISTS are on the Download tab, not the Content tab

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,115 @@ sidebar_label: Exporting
...
 
-*   Custom tabular export
-*   SQL
-*   Templating export \
+![A screenshot of the SQL statement download window.](/img/sql-exporter2.png)
 
-*   Upload edits to Wikidata
-*   Export to QuickStatement
-*   Export Wikidata schema
+The <span class="tabLabels">Download</span> tab allows you to finalize your statement, name it, and preview or download it.
```
ostephens, line 67: "tab allows you to finalize your statement, name it, and preview or download it."
This suggests that you can "name your statement" but in face you can set a name for the table which the statement will refer to

ostephens (changes requested):

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,117 @@ sidebar_label: Exporting
...
-*   TSV/CSV
-*   HTML table
+*   Tab-separated value (TSV) or Comma-separated value (CSV)
+*   HTML-formatted table
 *   Excel (XLS or XLSX)
 *   ODF spreadsheet
-*   Google Sheets \
+*   Upload to Google Sheets (requires [Google account authorization](starting#google-sheet-from-drive))
+*   [Custom tabular exporter](#custom-tabular-exporter)
+*   [SQL statement exporter](#sql-statement-exporter)
```
ostephens, line 25: There's a mismatch between this link (#sql-statement-exporter) and the anchor further down (#sql-exporter) meaning that the link doesn't work

ostephens (approved):

ostephens:
There are a couple of very minor issues that need fixing, but I'm approving this so we can get it merged and then fix up those minor issues separately (the outstanding issues are not substantive and not about the content)

ostephens:
@wetneb unless you have any other outstanding issues, I think we should go ahead and merge

docs/docs/manual/exporting.md
```diff
@@ -6,34 +6,117 @@ sidebar_label: Exporting
...
+![A screenshot of the SQL statement download window.](/img/sql-exporter2.png)
 
-*   Upload edits to Wikidata
-*   Export to QuickStatement
-*   Export Wikidata schema
+The <span class="tabLabels">Download</span> tab allows you to finalize your complete SQL statement. 
 
+<span class="fieldLabels">Include schema</span> means that you will start your statement with the creation of a table. Without that, you will only have an INSERT statement. 
+
+<span class="fieldLabels">Include content</span> means the INSERT statement with data from your project. Without that, you will only create empty columns.
```
ostephens, line 71: In my local preview there is no line break between these two lines, but looking at the code I'm not sure why not

Issue #733, opened 2013-06-12
Implement front end browser-based UI tests


tfmorris:
The testing tool that we use for our (relatively limited) front end tests is basically abandoned, so we should switch to something better supported such as Selenium 2.  One possibility would seem to be to use a free open source plan from Sauce Labs in conjunction with our Travis CI build/test setup.

http://sauceio.com/index.php/2013/03/run-your-selenium-tests-completely-in-the-cloud-using-travis-ci-and-sauce-labs/

tfmorris:
So it looks like Windmill got disabled in 2017 https://github.com/OpenRefine/OpenRefine/commit/9bc2c3078544b0898649bfb3fd3ab81d8d438616 and hasn't been replaced with anything.

Going from minimal UI tests to no UI tests is probably not the direction that we want to be trending...

wetneb:
(Just to make it clear: the tests that were disabled by this commit were not running at the time - Windmill was already broken and unmaintained)

wetneb:
I think we can close this, based on the awesome work of @fgiroud and @kushthedude!

PR #3340, opened 2020-11-18
Implemented front end browser-based UI tests #733

fgiroud:
This PR introduces Cypress as a browsed-based testing framework. Implements #733 
It includes
* The setup for Cypress
* Developer documentation
* Tests that cover basic but broad OpenRefine interface


wetneb (commented):
Thanks a lot, it looks great to me and it is fantastic to see the documentation being updated in the same pull request.

fgiroud:
Hi, thx !
I have added the entry in the sidebar, passed the markdown to autoformat, and linked the installation from the build-test-run page.
Also added a proper license to the package.json

Regarding the build-test-run, the command will be replaced by a simple call to `./refine ui_test`, but for this I need to setup Travis first, otherwise it will break the build. Coming soon.

Best,
- Florian

wetneb (approved):
Thanks again. I am happy to merge this in this current state.

As discussed in the last testing meeting, the tests currently rely on the "production" instance of OpenRefine (the one running on the standard port 3333, which opens by default on the "official" workspace for the machine).  For `./refine ui_test`, I would expect that the script starts an instance of OpenRefine on a different port (say 3334 for instance), and on a temporary workspace directory, so that the test workspace remains isolated from any other projects the developer might have on their machine.

Currently, running the tests on my machine give 4 failures (with Chromium, Linux), but I am sure we will get occasions to debug them down the line, when we set them up in Travis.

Issue #3341, opened 2020-11-18
GitHub CI uses set-env command which is a vulnerability


wetneb:
Our GitHub CI workflow runs have started failing because they rely on the `set-env` command, which has been deprecated and is now disabled by default:
https://github.com/OpenRefine/OpenRefine/actions/runs/370648289

```
The `set-env` command is disabled.
Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`.
For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```
We need to figure out if we could use Environment Files instead for this purpose.

wetneb:
We do not use the `set-env` command anymore, the structure of our CI has evolved a lot since this ticket.

PR #3342, opened 2020-11-18
Temporary workaround set-env deprecation

wetneb:
This is an attempt to temporarily re-enable the `set-env` command, to make sure our CI stays green, before we figure out how to use the new system.

For #3341


wetneb:
Merging to master since GitHub workflows only run there (per our configuration).

Issue #3330, opened 2020-11-12
Phonetic GREL function should raise an error 


wetneb:
The `phonetic` GREL function does not validate its second argument, the phonetic encoding, and silently falls back to metaphone.

For instance, `"open refine".phonetic("metaphone3")` evaluates to `"APNRFN"` and so does `"open refine".phonetic(undefined_value)` or `"open refine".phonetic(null)`.

In these last two cases, we should instead throw an error, similar to the one we get with `"open refine".phonetic("invalid_encoding")` (`Error: phonetic doesn't know how to handle the 'invalid encoding' encoding.`).

rachittiwari8562:
I think the error should be "phonetic expects second argument to be a string" because if we make a call to "open refine".photenic(soundex) the soundex argument is translated to null object  so the error should be to that a string argument is expected.

rachittiwari8562:
I have looked into this issue and found that if second argument is a string or character without " " then the object corresponding is treated as null and in file OpenRefine/main/src/com/google/refine/expr/functions/strings/Phonetic.java there is no else case for handling this.

wetneb:
That sounds like you identified the bug here!

rachittiwari8562:
The first was accidentaly done so please ignore it.

PR #3345, opened 2020-11-20
Fix for issue #3330 grel phonetic-function

rachittiwari8562:
Fix:
There is one test case missing in the file in the phonetic.java due to which it fails to validate the second argument when it is not string or number and hence falls back to default behaviour of considering "metaphone3" as its second argument.
Adding a if-else case for checking second argument solves the issue.

### DIFF
--- a/main/src/com/google/refine/expr/functions/strings/Phonetic.java
+++ b/main/src/com/google/refine/expr/functions/strings/Phonetic.java
@@ -72,7 +72,10 @@ public class Phonetic implements Function {
                     return new EvalError(ControlFunctionRegistry.getFunctionName(this)
                             + " expects a string for the second argument");
                 }
-            }
+            } else {
+                return new EvalError(ControlFunctionRegistry.getFunctionName(this)
+                        + " expects a string for the second argument.Make sure to pass the argument in proper string format");
+                }     
         }
         if (args.length < 3) {
             if ("doublemetaphone".equalsIgnoreCase(encoding)) {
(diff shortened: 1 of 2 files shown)


main/src/com/google/refine/expr/functions/strings/Phonetic.java
```diff
@@ -72,7 +72,10 @@ public Object call(Properties bindings, Object[] args) {
                     return new EvalError(ControlFunctionRegistry.getFunctionName(this)
                             + " expects a string for the second argument");
                 }
-            }
+            } else {
+                return new EvalError(ControlFunctionRegistry.getFunctionName(this)
+                        + " expects a string for the second argument.Make sure to pass the argument in proper string format");
+                }
```
wetneb, line 78: The indentation seems a bit off: this "}" should be at the same level as the first "}" of "} else {".

main/src/com/google/refine/expr/functions/strings/Phonetic.java
```diff
@@ -72,7 +72,10 @@ public Object call(Properties bindings, Object[] args) {
                     return new EvalError(ControlFunctionRegistry.getFunctionName(this)
                             + " expects a string for the second argument");
                 }
-            }
+            } else {
+                return new EvalError(ControlFunctionRegistry.getFunctionName(this)
+                        + " expects a string for the second argument.Make sure to pass the argument in proper string format");
```
wetneb, line 77:
```suggestion
                        + " expects a string for the second argument, the phonetic encoding to use.");
```

wetneb (commented):
Thanks a lot! Do you think you could add a unit test to demonstrate the fix? You could add it in the `PhoneticTests` class (see `FingerprintTests` for an example of test which actually calls the function).

rachittiwari8562:
I have added some tests please check if they are correct or do they need some improvment.

wetneb (approved):

PR #3349, opened 2020-11-22
Add Wynn to fingerprint to support Old English texts

### LINKED ISSUE #3348: Wynn and Thorn in Old English text does not get fingerprinted correctly
Using `fingerprint` and `ngramFingerprint` should correctly transliterate (fold) the Wynn `Ƿƿ` and thorn `þ` grapheme's for modern composition.

### Current Results

`ngramFingerprint('ƿiges ƿarne')`
```
aeginrsƿ
```

`fingerprint('ƿiges ƿarne')`

```
ƿarne ƿiges
```

`fingerprint('Ƿenne bruceþ, þe can ƿeana lyt')`
```
bruceþ can lyt þe ƿeana ƿenne
```

### Expected Behavior
`ngramFingerprint('ƿiges ƿarne')`
```
aeginrsw
```

`fingerprint('ƿiges ƿarne')`

```
warne wiges
```

`fingerprint('Ƿenne bruceþ, þe can ƿeana lyt')`
```
bruceth can lyt the weana wenne
```

### Versions
 - Browser Version: Firefox
 - JRE or JDK Version: JDK8
 - OpenRefine: 3.4.1

### Additional context
https://en.wikipedia.org/wiki/Wynn
https://en.wikipedia.org/wiki/Thorn_(letter)
(Thad spent his weekend playing with Gensim (TF-IDF) and playing with Vector Space Models on some Old English texts and wanting further analysis and tried a few experiments in OpenRefine additionally and hit upon this issue)

thadguidry:
Fixes #3348

### DIFF
--- a/main/src/com/google/refine/clustering/binning/FingerprintKeyer.java
+++ b/main/src/com/google/refine/clustering/binning/FingerprintKeyer.java
@@ -72,6 +72,7 @@ public class FingerprintKeyer extends Keyer {
             .put("\u0111", "d") // Small letter D with stroke
             .put("\u0256", "d") // Small letter African D
             .put("\u00FE", "th") // Lower case Icelandic thorn þ
+            .put("ƿ","w") // Lower case Wynn from Old English modernly transliterated to w
             // Visually similar replacements from our private former asciify() method
             // (only need lower case forms since we're already downcased)
             .put("\u0127", "h") // small H with stroke
--- a/main/tests/server/src/com/google/refine/clustering/binning/KeyerTests.java
+++ b/main/tests/server/src/com/google/refine/clustering/binning/KeyerTests.java
@@ -55,7 +55,7 @@ public class KeyerTests extends RefineTest {
 //        {"å","aa"}, // Requested by issue #650, but conflicts with diacritic folding
         {"æø","aeoe"}, // Norwegian replacements from #650
         {"©ß","css"}, // issue #409 esszet
-        {"\u00D0\u00DE", "dth"}, // Icelandic eth and thorn 
+        {"\u00D0\u00F0\u00DE\u00FEǷƿ", "ddththww"}, // eth, thorn, & wynn for Icelandic / Olde English
         {"ﬀﬁﬂﬃﬅﬆ", "fffiflffistst"}, // ligatures
         // Test legacy replacements
         {"\u00C0\u00C1\u00C2\u00C3\u00C4\u00C5\u00E0\u00E1\u00E2\u00E3\u00E4\u00E5\u0100\u0101\u0102\u0103\u0104\u0105", "aaaaaaaaaaaaaaaaaa"},


wetneb (commented):
Thanks! I don't think it hurts, but we should have a more principled approach to building a map like this, rather than adding the characters we encounter in our own data cleaning workflows. Perhaps there is a Java library out there which already does this for a large class of unicode characters? In Python I am aware of the `unidecode` package which does this pretty well:

https://pypi.org/project/Unidecode/

main/src/com/google/refine/clustering/binning/FingerprintKeyer.java
```diff
@@ -85,6 +85,8 @@
             .put("œ", "oe")
             .put("ẜ", "s") // more long S forms
             .put("ẝ", "s")
+            .put("Ƿ","w") // Wynn from Old English modernly transliterated to w
```
tfmorris, line 88: Is case folding not working correctly? We shouldn't need two entries.

As long as we're doing this, should we map the rest of the wynn forms? http://www.fileformat.info/info/unicode/char/search.htm?q=wynn&preview=entity
All runes? http://www.fileformat.info/info/unicode/char/search.htm?q=runic&preview=entity

main/tests/server/src/com/google/refine/clustering/binning/KeyerTests.java
```diff
@@ -100,6 +100,9 @@
 //         "bbddddhiikkllmnnooooopprssssttttuuuuyyzz                      aaiioouuuuuuuuuux"},
 //        {"ǞǟǠǡǢǣǤǥǦǧǨǩǪǫǬǭǮǯǰǱǲǳǴǵǶǷǸǹǺǻǼǽǾǿȀȁȂȃȄȅȆȇȈȉȊȋȌȍȎȏȐȑȒȓȔȕȖȗȘșȚțȜȝȞȟȠȡȢȣȤȥȦȧȨȩȪȫȬȭȮȯȰȱȲȳȴȵȶȷȸȹȺȻȼȽȾȿɀɁɂɃɄɅɆɇɈɉɊɋɌɍɎɏ",
 //         "aaaaaeaeggggkkqqqqqssjdzdzggpnnaaaeaeooaaaeeeeeiiiioooorrrruuuusstt    zzaaeeooooooooyy"},
+        // WIP: These we know should be folded based on research (Wikipedia,etc.), and modern expectations (transliterations) by researchers (philologists)
+        {"Ƿƿ",
+         "ww"},
```
tfmorris, line 105:
```suggestion
        {"Ƿƿ", "ww"}, // Old English letter wynn
```
Let's also relocate this up below the test for eth and thorn since they're similar.

tfmorris (changes requested):
Expanding the scope is optional, but please make the other changes (or let me know that case folding isn't working here)

main/src/com/google/refine/clustering/binning/FingerprintKeyer.java
thadguidry: You can take this PR over Tom.  I've got what I needed out of of it unblock me.  If it's useful for others or needs more work to be more useful...have at it.

thadguidry: The mapping that I was following was from the Unicode spec itself.  In particular the `scf` and `slc`, which I still find a bit weird on some glyphs.

thadguidry: and where I was trying to get `fingerprint` and `ngramFingerprint` to follow the rules of `scf` and `slc` by default.  If that's reading too much into it, then I goofed and sorry, but felt they were sensible most of the time for Old English texts.

tfmorris: The string is lowercased before it gets to this point, so only the lowercase substitution is needed.

tfmorris (approved):

PR #3357, opened 2020-11-25
Fix: Preventing addition of any empty cells with whitespaces while importing Xml Data with Tests #1095

maheshjindal:
* Fixes the issue while importing XML data containing whitespace characters generating empty cells in the table.
* Contains Unit Tests for the Fix

### DIFF
--- a/main/src/com/google/refine/importers/XmlImporter.java
+++ b/main/src/com/google/refine/importers/XmlImporter.java
@@ -209,7 +209,9 @@ public class XmlImporter extends TreeImportingParserBase {
     
     static public class XmlParser implements TreeReader {
         final protected XMLStreamReader parser;
-        
+        static final String WHITESPACES_PATTERN = "^[\\n\\r\\t].*$";
+        static final int WHITESPACE_CHARACTERS_TOKEN = 15;
+
         public XmlParser(InputStream inputStream) throws XMLStreamException, IOException {
             parser = createXMLStreamReader(inputStream);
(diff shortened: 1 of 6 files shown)


main/src/com/google/refine/importers/XmlImporter.java
```diff
@@ -209,7 +209,9 @@ public void parseOneFile(Project project, ProjectMetadata metadata,
     
     static public class XmlParser implements TreeReader {
         final protected XMLStreamReader parser;
-        
+        static final String WHITESPACES_PATTERN = "^[\\n\\r\\t].*$";
```
wetneb, line 212: Instead of introducing our own regex here, perhaps we could use `CharMatcher.whitespace()` just like in other places in the code (see for instance `main/src/com/google/refine/expr/functions/strings/Trim.java`).

wetneb (approved):
This looks really great to me, thank you! The tests are very nice and this should make a pretty big difference to users. I just noticed a tiny detail below.

main/src/com/google/refine/importers/XmlImporter.java
maheshjindal: @wetneb Thanks for the suggestion. I have updated the PR with new changes. Please let me know if anything else we can enhance.

wetneb:
Merging without Travis' checks because of #3377. Appveyor is still running and there are no risks of platform dependency for this change.

PR #3178, opened 2020-09-09
Documentation style changes

sjathin:
Fixes for Issue #3059 

1. Used span and blue back-ground style to refer to UI text.
2. Infobox styled in the spirit of Vector's
3. More space, above the section headers.
4. More space between the sections in the sidebar
5. Bigger header sizes
6. Images centered 

Attached a pdf document which shows the changes made to the design
[OpenRefine#3059.pdf](https://github.com/OpenRefine/OpenRefine/files/5197641/OpenRefine.3059.pdf)

PR #3365, opened 2020-11-26
Style modifications in docusaurus docs to fix #3329

### LINKED ISSUE #3329: Style issue in Docusaurus docs
We have too much padding on menu entries in the left-hand side bar of the docs:

image

This was probably introduced by mistake in #3178.

We should restore this padding to Docusaurus' defaults - the area that needed more padding is on the right-hand side (see PR discussion).

Mentioned by @allanaaa in today's documentation meeting.

leoshrey:
Fixes #3329 

- Padding is now restored to Docusaurus defaults.( Corrected the mistake in #3178 as mentioned by @wetneb )

check it out

Screenshot 2020-11-26 at 11 22 31 PM


docs/src/css/custom.css
```diff
@@ -38,8 +38,8 @@
 
   .menu__link--sublist {
     margin-top: 10px;
-    padding-bottom: 25px;
-    padding-top: 25px;
+    padding-bottom: 6px;
+    padding-top: 6px;
```
wetneb, line 42: If these are the Docusaurus defaults, could we simply remove these lines?

wetneb (commented):
Thanks a lot for taking care of this!

docs/src/css/custom.css
leoshrey: Actually, this is a great idea. I have checked it too and it is working fine.

leoshrey: I think we can remove.

leoshrey:
Since this is tested, do you want me to raise another PR or edit this one?

wetneb:
Please edit this PR directly, you should be able to do it by committing and pushing your changes on the existing branch.

wetneb (approved):
Thank you very much!

Issue #3377, opened 2020-12-01
Travis build credits


wetneb:
Travis currently says:

"Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting."

https://travis-ci.com/github/OpenRefine/OpenRefine

I think we need to shrink our build matrix, especially if we want to add new jobs for the UX tests. I would start by removing builds marked as allowed failures.

Concerning the Java versions, I would only test with the oldest and newest versions we support.

We could also consider migrating to another provider, or requesting a raise of our usage limit, but I think it generally makes sense to try to keep the build matrix minimal, for the sake of getting a quicker feedback on PRs and saving resources.

tfmorris:
I can't access either of the links in that warning. Do we know whether the
credits lasted most of November or expired on the first day of December or
something else?

I suspect that we're going to need to migrate away from Travis. That seems
to be what many other OSS organizations are doing, but I'm not sure what
are spend is relative to their budget or how free they are about increasing
their allowances.

The build matrix for PRs is already pretty minimal -- just Java 8, as I
recall. It's the merges that trigger the full build matrix, but we can
certainly pare that back and try to remember to manually run whatever
matrix we decide on before releases.

Tom

wetneb:
Both links in the warning also give errors for me.

tfmorris:
I logged in again via Github and was able to get to https://travis-ci.com/organizations/OpenRefine/plan
which says "-2410 available credits (purchase date: December 01, 2020) You have used 12410 of 10000 credits"

Implying that we used 12k credits in a few hours (or perhaps that we used that amount since Nov. 2 when they announced the pricing change). Either way, the credit was a one-time thing, not a monthly allotment, so they are gone. We can apply for more, but that would also be a one-time thing, at least by default. I know another org got 50K relatively easily (although they assigned it to the requestors account not the OSS org), so perhaps we should make the request to bridge us over while we figure out where to move.

The charge a premium for Mac OS X, so we should remove those from the matrix.

wetneb:
Marking as high priority since this is impeding our reviewing of other pull requests.

wetneb:
Closing as we migrated to GitHub actions. If we want to go back we can open a specific issue about it.

PR #3366, opened 2020-11-27
Issue 3352 travis setup

fgiroud:
This PR adds the UX tests to ./refine ui_test command, closes #3352

* Added Travis Matrix
* Fixed flaky tests
* Refactored/Fixed  some tests
* Tried integration with Cypress dashboard

PR #3378, opened 2020-12-01
Migrate from Travis to GitHub Actions

wetneb:
This is a drastic simplification of our build matrix, for #3377.

There are relatively few differences between versions of Java, so by testing only with the oldest and the newest we should cover both issues due to deprecation of old APIs (with the newest version, 14) and the use of APIs that are too recent (with the oldest version, 8).

The differences between OpenJDK and OracleJDK are pretty negligible nowadays so it is not worth duplicating tests for both vendors.

The mac OS build had to be removed because it costs more than Linux / Windows builds (although it is not exactly clear how much).

I'll request new credits for the organization to resume builds.


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,57 @@
...
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+    - uses: actions/checkout@v2.3.4
+      with:
+        fetch-depth: 0 # This is wasteful, but needed for git describe
```
thadguidry, line 26: @wetneb Wait, I'm confused here on this `fetch-depth` line.  I don't see the git describe need. Where at?  I see `git describe` used in the `refine` script for `mac_dist()` which says no longer used in a comment, and then also for `broker_build()` and `broker_appengine_build()`.  Are those even run or needed now since we now do "mvn package" ?  For our PR workflow, I think we just care about what the last tag was on master, no?  Or even just 'heads/master' ? 
```
E:\GitHubRepos\OpenRefine>git describe --tags --first-parent --all
heads/master
```
BTW, we also have access to the environment variable in the runner at any time with **GITHUB_REF** which represents the PR merge branch `refs/pull/:prNumber/merge` as shown in docs https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

wetneb: Good point, I just took that from the other workflow, it's not needed here indeed.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,66 @@
+name: Java CI
+
+on: [pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
tfmorris, line 8: Do we want to pin this? Based on the [update to Ubuntu 20](https://github.com/actions/virtual-environments/issues/1816), it looks like they're happy to include updates of installed versions which have breaking changes.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,66 @@
...
+
+    steps:
+    - name: Create test MySQL database
+      id: create_test_db
+      uses: mirromutth/mysql-action@v1.1
+      with:
+        host port: 3306
+        container port: 3306
+        mysql database: 'test_db'
+        mysql root password: 'root'
```
tfmorris, line 31: I think we can probably run MySQL as a service too, along the lines of https://firefart.at/post/using-mysql-service-with-github-actions/, including the "wait for availability" check

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+
+    - name: Configure connections to databases
+      id: configure_db_connections
+      run: cp extensions/database/tests/conf/github_actions_tests.xml extensions/database/tests/conf/tests.xml 
+
+    - name: Populate databases with test data
```
tfmorris, line 52: Perhaps precede this with a "wait for databases to be available" check?

extensions/database/tests/conf/github_actions_tests.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suite name="DatabaseExtensionUnitTestSuite">
+    <parameter name = "mySqlDbName" value="test_db"/> 
+    <parameter name = "mySqlDbHost" value="127.0.0.1"/> 
+    <parameter name = "mySqlDbPort" value="3306"/>
```
tfmorris, line 5: I think these will need to be adjusted for DB as a service.

extensions/database/tests/conf/github_actions_tests.xml
```diff
@@ -0,0 +1,55 @@
...
+    <parameter name = "mySqlDbName" value="test_db"/> 
+    <parameter name = "mySqlDbHost" value="127.0.0.1"/> 
+    <parameter name = "mySqlDbPort" value="3306"/> 
+    <parameter name = "mySqlDbUser" value="root"/>
+    <parameter name = "mySqlDbPassword" value="root"/>
+    <parameter name = "mySqlTestTable" value="test_table"/>
+    
+    <parameter name = "pgSqlDbName" value="test_db"/> 
+    <parameter name = "pgSqlDbHost" value="postgres"/> 
+    <parameter name = "pgSqlDbPort" value="5432"/>
```
tfmorris, line 12: Ditto comment above

extensions/database/tests/conf/github_actions_tests.xml
```diff
@@ -0,0 +1,55 @@
...
+    <parameter name = "pgSqlDbUser" value="postgres"/>
+    <parameter name = "pgSqlDbPassword" value="postgres"/>
+    <parameter name = "pgSqlTestTable" value="test_table"/>
+    
+    <parameter name = "mariaDbName" value="test_db"/> 
+    <parameter name = "mariaDbHost" value="127.0.0.1"/> 
+    <parameter name = "mariadbDbPort" value="3306"/> 
+    <parameter name = "mariaDbUser" value="root"/>
+    <parameter name = "mariaDbPassword" value="root"/>
+    <parameter name = "mariaDbTestTable" value="test_table"/>
```
tfmorris, line 22: I don't see this database being started anywhere.

tfmorris (commented):
I suspect this is still a work in progress, but here are some of the things I noticed

tfmorris:
> No, you are doing it wrong... 

@thadguidry There are probably more constructive ways to phrase your suggestion.

While we could start the MySQL server as a process on the host, per your suggestion, it might be more consistent to run all the databases as services like PostgreSQL is.

thadguidry:
@tfmorris Ah, rereading that and I did come off rude-like huh?  I did not intend to do that, sorry!  Yes, for integration testing, GitHub recommends using Service containers with Docker and the service containers will be destroyed after the jobs complete.

https://docs.github.com/en/free-pro-team@latest/actions/guides/about-service-containers

My point was around this:  A reason for not using Service containers is when your application has specific needs, non-Linux runners for instance: 

> Note: If your workflows use Docker container actions or service containers, then you must use a Linux runner:
> 
>     If you are using GitHub-hosted runners, you must use an Ubuntu runner.
>     If you are using self-hosted runners, you must use a Linux machine as your runner and Docker must be installed.


so GitHub provides all in one virtual environment images with most tooling that developers need, if they cannot easily breakup there workflow jobs.

In our case, yes, we can easily use Service containers (DB, etc.) and take advantage of them for our PR and UI testing cycles.

thadguidry:
Another thing that is helpful (I used a similar feature in Gitlab) is the use of Grouping log lines for various parts of testing, perhaps Extensions, or whatever we decide is helpful to collapse and group the log lines https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#grouping-log-lines
but maybe we don't care so much.

If we ever want to add some specific workflow jobs for Extension integration testing, we can make Docker images something like this, which I made for a client today in fact: https://gist.github.com/thadguidry/149790397ee88f08a8fcfc162b362817

extensions/database/tests/conf/github_actions_tests.xml
wetneb: Yes indeed, I realized that so far in Travis we actually tested the MariaDB connector with the MySQL instance (since they both connected to port 3306). It could be worth spinning a real MariaDB instance instead.

wetneb:
Thanks both, I'll give a try to MySQL as a `service`. I did not realize this was available, it looks much cleaner indeed.

.github/workflows/pull_request.yml
wetneb: I would say that if we become incompatible with new versions of Ubuntu, it's good to be aware of it as soon as possible, so that's why `ubuntu-latest` makes intuitive sense to me.

wetneb:
Since I have no idea how long it will take for Travis to unblock our build, I propose that we switch to GitHub Actions for now.
Otherwise this is going to block @fgiroud's work on integrating UI tests in the CI.

@fgiroud are you okay with switching to GitHub Actions? That would involve redoing #3366, I am sorry about that!

If there is agreement on this, I can delete `.travis.yml` in this PR.

thadguidry:
@wetneb I think we should switch to GitHub Actions.  I only see benefits there and generous hands from Microsoft towards Open Source projects such as ours.  There is also an enormous community now that can help us maintain this CI infrastucture where some also maintain component Actions on the marketplace as well!  Whenever we need some changes, we could just open an issue an mark as "good first issue" and we'd probably have it closed and fixed by the growing community in a few hours.

So +1 for moving to GitHub Actions.  (I wish I could help more hands-on, but I'm buried in my own work with my client these past weeks and going into 2021, sorry!)

tfmorris:
I don't have good visibility into how much more work is needed, but it seems like the bulk of the work has been done already. I'm fine with switching to Github Actions. I think it's the path of least resistance globally, even if it causes some short term pain.

.github/workflows/pull_request.yml
wetneb: Done when adding MySQL as a service.

.github/workflows/pull_request.yml
tfmorris: They apparently generate artificial build "warnings" when they're getting ready to change what "latest" points to, so I suppose we can use that as a prompt to proactively test. Having avoidable unscheduled build breakages isn't really the best (as we've found out).

tfmorris (approved):
LGTM as a starting point. Now I guess we can test the "merge" builds. ;)

PR #3380, opened 2020-12-02
Docs > formatting

allanaaa:
This is a sort of grab-bag PR, sorry. Changes and tweaks across all the docs based on conversations and things forgotten have been accumulating. 

Especially note:
- the "Advanced OpenRefine uses" section at the bottom of the Running page
- a "Dates" subsection in Data Types on the Exploring page (and I'll be consolidating date descriptions from various places)
- some small changes to the Troubleshooting page.

This is in draft because I am working on getting the formatting consistent, and updating links within the docs. As I understand it, we format the following in their respective ways:
```
- a single folder name: quotes ('your "Extensions" folder')
- a folder path: `D:\Program Files\OpenRefine`
- a single file name: `blah.exe`
- a line of code in a file: ``` (with line breaks, for a full-width code block) ```
- a quote of code from a file: `-Xmx1024M`
- a full GREL formula: `value.toString()`
- a GREL function name: toString()
- a variable: `value`
- an OpenRefine menu item: <span class="menuItems">
- an OpenRefine field name or checkbox: quotes ("Allow null")
- an OS or other non-OR button or menu item: bold 
```

Happy to be corrected or change any of these.


wetneb:
I would also put GREL function names in backquotes: 
```
`cross`
````

**Comment by allanaaa** [comment]
Our consensus from the meeting today was to create a span for each UX element in OpenRefine. Then they can be styled (or not) later.
So far I have the following classes, though I may add more:
- menuItems
- fieldLabels
- buttonLabels

And we'll do the same for function names (functionNames).

**Comment by allanaaa** [comment]
As I work through, I'm seeing some placeholders left. One in particular is on running.md: 

|Command|Use|Syntax example|
|---|---|---|
|-x|Additional configuration parameters||

Understandably, googling for examples of "./refine -x" used in forum posts and tutorials doesn't work too well. Does anyone have an example of what an additional parameter would be and how to use it here? If this exists in theory but there are no use cases we should probably remove it from this table.

**Comment by thadguidry** [comment]
@allanaaa It's here:  https://github.com/OpenRefine/OpenRefine/blob/99e7d0d3faf41dd14228d1037cfed3f4b78051d9/refine.bat#L43

**Comment by allanaaa** [comment]
> 
> 
> @allanaaa It's here:
> 
> https://github.com/OpenRefine/OpenRefine/blob/99e7d0d3faf41dd14228d1037cfed3f4b78051d9/refine.bat#L43
> 
> Which then does this:
> 
> https://github.com/OpenRefine/OpenRefine/blob/99e7d0d3faf41dd14228d1037cfed3f4b78051d9/refine.bat#L133

That's the /x for Windows; I'm looking for the Mac and Linux startup command -x (you can do JMX monitoring with --jmx I believe). 
The table I drafted is here: https://deploy-preview-3380--openrefine-docs.netlify.app/manual/running#starting-with-modifications

I fired up the VM and the Linux version (when you run ./refine -h) just says:

> -x <name=value> additional configuration parameters to pass to OpenRefine default: [none]

**Comment by thadguidry** [comment]
On Linux and Mac, it's treated more generically, and basically the argument for passing ANY extra Java Options to the `java` command.
Take a look at https://github.com/OpenRefine/OpenRefine/blob/99e7d0d3faf41dd14228d1037cfed3f4b78051d9/refine#L813
```
    -x) shift; REFINE_EXTRA_OPTS="$1"; shift; continue;;
    --debug) shift; add_option '-Xdebug' '-Xrunjdwp:transport=dt_socket,address=8000,server=y,suspend=n'; continue;;
    --jmx) shift; add_option '-Dcom.sun.management.jmxremote'; continue;;
    -*) fail "Invalid option: $1";;
```
@wetneb should probably explain in more detail.

**Comment by allanaaa** [comment]
> 
> 
> On Linux and Mac, it's (our `refine` script) treats it more generically, and basically the argument for passing ANY extra Java Options to the `java` command.
> Take a look at
> 
> https://github.com/OpenRefine/OpenRefine/blob/99e7d0d3faf41dd14228d1037cfed3f4b78051d9/refine#L813
> 
> ```
>     -x) shift; REFINE_EXTRA_OPTS="$1"; shift; continue;;
>     --debug) shift; add_option '-Xdebug' '-Xrunjdwp:transport=dt_socket,address=8000,server=y,suspend=n'; continue;;
>     --jmx) shift; add_option '-Dcom.sun.management.jmxremote'; continue;;
>     -*) fail "Invalid option: $1";;
> ```
> 
> @wetneb should probably explain in more detail.

Okay, I think I get the idea. It may be sufficient to just say "See Java documentation" and not bother supplying an example. If anyone does actually use this regularly and has a helpful tip to share, let me know - otherwise, we can leave it blank. 

> |-x|Additional Java configuration parameters (see Java documentation)||

File: docs/docs/manual/exploring.md [hunk 1]
```diff
@@ -1,54 +1,92 @@
...
+
+To transform data from one type to another, see [Transforming data](cellediting#data-type-transforms) for information on using common tranforms, and see [Expressions](expressions) for information on using [toString()](grelfunctions#tostringo-string-format-optional), [toDate()](grelfunctions#todateo-b-monthfirst-s-format1-s-format2-), and other functions. 
+
 
-Converting a cell's data type is not the same operation as transforming its contents. For example, using a column-wide transform such as “Transform” → “Common transforms …” → “to date” may not convert all values successfully, but going to an individual cell, clicking “edit” and changing the data type can successfully convert text to a date. These operations use different underlying code. 
+### Dates
```
allanaaa, line 44: The date subsection in the Data Types section is new and needs review.

docs/docs/manual/running.md
```diff
@@ -387,22 +433,74 @@ When you click on <span class="menuItems">Undo / Redo</span> in the sidebar of a
...
+To reuse one or more operations, first extract it from the project where it was first applied. Click to the <span class="tabLabels">Undo/Redo</span> tab and click <span class="menuItems">Extract…</span>. This brings up a box that lists all operations up to the current state (it does not show undone operations). Select the operation or operations you want to extract using the checkboxes on the left, and they will be encoded as JSON on the right. Copy that JSON to the clipboard. 
 
-Move to the second project, go to the <span class="menuItems">Undo/Redo</span> tab, click <span class="menuItems">Apply…</span> and paste in that JSON.
+Move to the second project, go to the <span class="tabLabels">Undo/Redo</span> tab, click <span class="menuItems">Apply…</span> and paste in that JSON.
 
 Not all operations can be extracted. Edits to a single cell, for example, can’t be replicated.
 
-### Common extension buttons
+## Advanced OpenRefine uses
```
allanaaa, line 454: Everything from here to the end of the page is new and requires review.

allanaaa:
I've tried to mark the major changes that will interest reviewers; I'll go back and mark a few more later.

Generally speaking, if you want to check up on this PR, the earliest pages (installing, running, starting a project) had the most changes to bring them in line with decisions made throughout the drafting process.
Reconciling also got a respectable number of edits on this round. 
So, reading one of any of those would be a great place to start if you have time.

docs/docs/manual/running.md
```diff
@@ -387,22 +433,74 @@ When you click on <span class="menuItems">Undo / Redo</span> in the sidebar of a
...
+
+### Running as a server
+
+:::caution
+Please note that if your machine has an external IP (is exposed to the Internet), you should not do this, or should protect it behind a proxy or firewall, such as nginx. Proceed at your own risk.
```
wetneb, line 459:
```suggestion
Please note that exposing an OpenRefine instance to the Internet is dangerous, as it gives anyone the ability to read and modify your projects and run arbitrary code on your computer. OpenRefine should at least be protected by an authenticating proxy.
```

docs/docs/manual/exploring.md
```diff
@@ -1,54 +1,92 @@
...
 
-To transform data from one type to another, see [Transforming data](transforming#transform) for information on using common tranforms, and see [Expressions](expressions) for information on using `toString()`, `toDate()`, and other functions. 
+A “date” type is created when a column is [transformed into dates](transforming#to-date), when an expression is used to [convert cells to dates](grelfunctions#todateo-b-monthfirst-s-format1-s-format2-) or when individual cells are set to have the data type “date.”
```
wetneb, line 46:
```suggestion
A “date” type is created when a column is [transformed into dates](transforming#to-date), when an expression is used to [convert cells to dates](grelfunctions#todateo-b-monthfirst-s-format1-s-format2-) or when individual cells are set to have the data type “date”. 
```

wetneb (approved):

allanaaa:
These changes to the Mac running instructions invalidate the "Starting with modifications" section for Mac as well (since that is itself a mirror of the modifications from Linux). Should this section simply be empty, or can you modify the way OpenRefine runs when you open it with Terminal? 
(I will attempt to test this myself when I get this VM cooperating, but I imagine someone can answer quicker.)

ostephens:
> These changes to the Mac running instructions invalidate the "Starting with modifications" section for Mac as well (since that is itself a mirror of the modifications from Linux). Should this section simply be empty, or can you modify the way OpenRefine runs when you open it with Terminal?

Good catch @allanaaa. As far as I know you can't use command line options when starting the Mac packaged bundle (and I've tested the obvious options and done some googling for this). So I think the Mac section should just state this and point the user and the "Modifications set within files" section if they need to modify any aspects of how the application runs

ostephens:
@allanaaa I think you've removed the Mac section completely. I think better to have a Mac section that says "You can't do this" and points the user to the file based instructions, otherwise the tab switching in the documentation gets confusing and the user won't be clear why there are no Mac instructions here

allanaaa:
Sorry, I was thinking over the options, shouldn't've uploaded yet. I think removing the tab is the best idea, and putting the line about not being able to in the paragraph above.

docs/docs/manual/running.md
```diff
@@ -92,12 +106,13 @@ If you are having problems connecting to OpenRefine with your browser, [check ou
...
 
+You cannot start the Mac version with modifications using Terminal, but you can modify the way the application starts with [settings within files](#modifications-set-within-files).
+
 <Tabs
   groupId="operating-systems"
   defaultValue="win"
   values={[
     {label: 'Windows', value: 'win'},
-    {label: 'Mac', value: 'mac'},
     {label: 'Linux', value: 'linux'}
```
ostephens, line 116: Really sorry - I think we need to re-instate the Mac section here - otherwise when the user is switching to the "Mac" tab, this can display instructions from Windows/Linux instead - leading to a mix of operating specific instructions on a single screen

allanaaa: Oh that's a fair point. I guess it doesn't just leave a blank.

ostephens: So - that is to say I don't agree removing the tab is the best idea because Docusaurus tab switching can get mucked up :(

allanaaa: Yep, it's restored now - take a look and let me know if it looks good to you?

ostephens (approved):

ostephens:
All looks good to me

allanaaa:
Great! If there are no further thoughts we can merge this and start sharing it around.

PR #3393, opened 2020-12-08
Run UI tests in continuous integration

fgiroud:
* Fixed flaky tests
* Refactored ui_test command
* Integration with Github actions


fgiroud:
Hello, this is a new PR that replace the previous one, with copy-pasting my work.
I have an issue running Cypress tests, it doesn't run OpenRefine
I've made an attempt with a copy of packaging/test_pom.xml into packaging/pom.xml but it doesn't work.
And I am not familiar with maven, can someone point me to the right direction ?

Here is what OR says at run time

``` 
2020-12-08T13:33:08.4046142Z ##[endgroup]
2020-12-08T13:33:08.4303915Z Using refine.ini for configuration
2020-12-08T13:33:08.5029099Z You have 6954M of free memory.
2020-12-08T13:33:08.5030049Z Your current configuration is set to use 1400M of memory.
2020-12-08T13:33:08.5030950Z OpenRefine can run better when given more memory. Read our FAQ on how to allocate more memory here:
2020-12-08T13:33:08.5032723Z https://github.com/OpenRefine/OpenRefine/wiki/FAQ-Allocate-More-Memory
2020-12-08T13:33:08.5051372Z fatal: No names found, cannot describe anything.
2020-12-08T13:33:08.5204745Z Waiting for OpenRefine to load...
2020-12-08T13:33:08.5908667Z Error: Could not find or load main class com.google.refine.Refine

``` 


Also please note that `on: [push]` inside the workflow is for testing purpose only, will be replaced by `on: [pull_request_target]`

wetneb:
That seems to be because you are calling `mvn process-resources` only: that does not compile OpenRefine itself. You should do `mvn process-resources build` or just call `./refine build`.

refine
```diff
@@ -498,23 +514,37 @@ ui_test() {
...
-        "$WINDMILL" firefox firebug loglevel=WARN http://${REFINE_HOST}:${REFINE_PORT}/ jsdir=$REFINE_TEST_DIR/client/src exit
+
+    echo "Starting Cypress..."
+    CYPRESS_RUN_CMD="yarn --cwd ./main/tests/cypress run cypress run --browser $BROWSER --headless --quiet --reporter list --env OPENREFINE_URL=http://$REFINE_HOST:$REFINE_PORT"
+    if [ "$CYPRESS_RECORD" = "1" ] ; then
+        # if tests are recorded, project id is added to env vars, and --record flag is added to the cmd-line
+        export CYPRESS_PROJECT_ID=$CYPRESS_PROJECT_ID
+        CYPRESS_RUN_CMD="$CYPRESS_RUN_CMD --record --key $CYPRESS_RECORD_KEY --tag $BROWSER,$REVISION"
+    fi
+    export MOZ_FORCE_DISABLE_E10S=1
```
thadguidry, line 525: what is this for or does?

fgiroud:
Thanks to @thadguidry I was able to fix the CI build
Questions: Why cypress test are not triggered? is it because we have two workflows with `on: [pull_request_target]` ?
But if we merge into `pull_request.yml`, we don't want the full java matrix for each browser, any thoughts?

I've created #3401 to add edge, that will be a separate PR later down the road

wetneb:
@fgiroud because you are using `pull_request_target`, the checks will not run until the PR is merged (and even then, they will only run on the next PR, assuming the branch the PR is made from branched off master after your commit). You could temporarily switch to `pull_request` (and then revert to `pull_request_target` before merging) but that would not solve the issue because you are doing the PR from a fork so secrets will not be available from there.

In any case it's of course fine to have follow-up PRs if things don't work as expected. I agree we should not duplicate the browser tests for each java version.

fgiroud:
this remains unclear on my side, I've changed the workflow for 'pull request', so we can expect that the Cypress tests runs for each and every PR ?
Then why 'pull-request-target' was triggered for my last push ?

wetneb:
The docs probably explain it better than I do:
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target

fgiroud:
@wetneb Can you please add CYPRESS_RECORD_KEY in the github secrets, the build will fail until then.

Here is my plan 
- Push a commit switching the workflow to "pull_request", then we cypress tests will be triggered
- Revert to pull_request_target
This way we can ensure once that the workflow is running, that should be good for subsequent PRs

Also: I've merged cypress workflows into the Java CI, si it's all packed into one single workflow, but with different jobs
I also added the cypress tests as a separate job on the snapshot_release

thadguidry:
@fgiroud FYI, Jobs run in parallel.  If it makes sense, we can always add another Job (have 2 Jobs in the workflow file). Just thought I'd point that out.

wetneb:
@fgiroud secrets are not available in builds from forks triggered by the `pull_request` event. They are only available in `pull_request_target` events.

This is a security measure to prevent people from uncovering secrets by submitting a pull request to a repository which changes the build script to print out the secret.

The plan you outlined above would only work if your branch was inside the official repository, not in your fork.

fgiroud:
Ok I understands @wetneb 
So the only way I see this working is that if you merge the PR as is (I reverted to pull_request_target)
The UI tests will not runs for this PR, and we are hoping that this will not fail the build for subsequent PRs, adjust if necessary

thadguidry:
@fgiroud can be added as a collaborator in our GitHub org... like we did Tony before.  That might help ease this hurdle.

wetneb:
Oops, we should have done that earlier indeed - invite sent. Let's merge this then!

wetneb (approved):

thadguidry:
@wetneb and with him as a collaborator with write access, he can create branches to work with, submit PR's against them, he can later delete those branches after we've merged his work.  And we have protected "master" branch for force pushing (except Admins) so we should be good.  Essentially he becomes part of the team, permanent or temporary, however we decide depending on his quality and cooking skills :-)

PR #3399, opened 2020-12-11
Add Zenodo DOI badge

tfmorris:


thadguidry:
@tfmorris Nice Tom!  I noticed on Zenodo this note on the right hand side:

> Cite all versions? You can cite all versions by using the DOI 10.5281/zenodo.595996. This DOI represents all versions, and will always resolve to the latest one. Read more. 

1. Or maybe that's not the intent of the that Badge DOI itself?  But I thought it was, where there's a Badge to represent **all** versions of OpenRefine, and other DOI's would represent other specific versions of OpenRefine?
It seems maybe that's more useful for a README Badge to point to the Zenodo DOI for all versions?  Literature can cite whatever DOI (version of OpenRefine) they wish?

2. Will our CI be changed a bit for the release action template to automatically update the badge, or again is that not the intent.
Thanks for advance for your upcoming explanation!

tfmorris:
My assumption was that it was using the "all versions" DOI. If that's not the case, I'll change it.

wetneb (approved):
Using the all versions DOI here makes sense to me.

thadguidry:
@tfmorris Hi Tom, someone in another channel said that this GitHub guide worked for them for Zenodo.
https://guides.github.com/activities/citable-code/

However, I do see that in our  Settings -> Webhooks -> Zenodo url -> Edit ... that it seems to be working, but on Zenodo site, I don't see it updating and creating DOI's for our newest releases? 
image

tfmorris:
In my cursory check last night, it didn't look like it was possible to use get a badge for the "all versions" DOI, but I'll look again.

I was getting errors from Github and had to reauthenticate Zenodo to it. It looks like I set this up for 2.7 and it broke sometime after 2.8. I'm guessing either a) it'll never create DOIs for the missing versions or b) it'll only notice that they're missing the next time we create a release at which point it will backfill. Given that there have only been <500 views over the last several years, I'm not sure it's super high priority to try and figure out.

PR #3405, opened 2020-12-12
Add Security Policy on how we handle reports

thadguidry:


SECURITY.md
```diff
@@ -0,0 +1,18 @@
+# OpenRefine Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 3.4.x   | :white_check_mark: |
+| < 3.3   | :x:                |
```
wetneb, line 8:
```suggestion
| <= 3.3   | :x:                |
```

wetneb (changes requested):
Thanks for looking into this! In general, projects try to keep security vulnerabilities secret until a new version is released with a fix. So I think we should not encourage people to report these on GitHub directly, but rather to a restricted mailing list (openrefine-coredev, if posting from non-subscribers is allowed? Otherwise, we could create a new mailing list for that purpose).

thadguidry:
@wetneb Yes, GitHub does have that security option for secret reporting.  I'll change that and point it to the openrefine-coredev mailing list.  I also had the same question and concerned but thought I'd start with this simple template as they suggest and then get the coredev team's feedback.  Let me update this...

tfmorris (approved):

wetneb (approved):

PR #3406, opened 2020-12-13
Add feature to trigger a button on hitting Enter to fix #3137

### LINKED ISSUE #3137: Pressing "Enter" in the project name field at the preview stage should create the project

When previewing an importer's results, there is a text field to enter the project name. When pressing Enter in this field, I would expect the project to be created (without having to click the "Create Project" button).

image

leoshrey:
Fixes #3137 

- Pressing "Enter" in the project name field at the preview stage will now create the project.

check it out

sample123 (1)


wetneb (changes requested):
Thanks a lot, much cleaner!

I think it would be better to make sure the field and the button are both part of the same `<form>` element, with the event listener being bound to the form submission event rather than the click on the button. This would directly ensure that the form gets submitted when the user presses "Enter", and would probably be more accessible too.

leoshrey:
Agreed and I think it’s a great idea as this event of triggering a button makes less sense than directly submitting the form by pressing Enter only, it will probably increase the accessibility as well as speed.
I am working on it.

leoshrey:
Hi @wetneb , I tried to bind the event listener with the form submission event rather than clicking on the button.
But as I found that this way of submitting the form requires form id attribute value, to which either I am not able to locate through or there isn't any. I also tried creating the <form> tag by providing useful attributes there and wrote the JS code for event listener using those attributes, but after doing so whenever I press Enter to Project Name input field it redirects me to the previous page (where we import the project). 
Can you please guide me to the form id value or should we proceed using .click() method instead ?

wetneb:
If the input and button are not yet enclosed in a `<form>`, then you should be able to add one, and add an `id` so that you are able to locate it in the Javascript code.

leoshrey:
I tried enclosing:
1. Input and button in <form>
2. All html elements in <form>
and writing JS for the same for submit event with event listener.
But both of the times whenever I press Enter key I am redirecting to the previous page (project import page).

In case 1. I am only redirecting to previous page without any extra warning log (checked in console).
In case 2. I am redirecting to the previous page as well as I am getting one extra warning as "Form submission canceled because the form is not connected".
Looks like .submit() event is not working because of browser version.
What are your views ?

wetneb:
Does your submit event handler cancel the default action with `e.preventDefault()`?

leoshrey:
No, I am not using this method.
Although I tried it both ways but the result is same.

wetneb:
You could try to follow this example:
https://github.com/OpenRefine/OpenRefine/blob/79eac0b8113f295c48b7152461e373beb0433e43/extensions/wikidata/module/scripts/dialogs/manage-account-dialog.js#L174
https://github.com/OpenRefine/OpenRefine/blob/79eac0b8113f295c48b7152461e373beb0433e43/extensions/wikidata/module/scripts/dialogs/password-login-dialog.html#L15

leoshrey:
Hey @wetneb, changes made as requested https://github.com/OpenRefine/OpenRefine/pull/3406#pullrequestreview-550943567. 

- Now the same event is bound as with createButton to keep it standard.

main/webapp/modules/core/scripts/index/default-importing-controller/parsing-panel.js
```diff
@@ -99,6 +99,13 @@ Refine.DefaultImportingController.prototype._prepareParsingPanel = function() {
   this._parsingPanelElmts.previousButton.html($.i18n('core-buttons/previous'));
   this._parsingPanelElmts.startOverButton.html($.i18n('core-buttons/startover'));
   this._parsingPanelElmts.nextButton.html($.i18n('core-buttons/create-project'));
+
+  // Submit the form directly by hitting Enter key
+  document.getElementById("projectNameTextField").addEventListener("keyup", function (event) {
```
wetneb, line 104: This still uses a keyup event listener, we should use a submit event instead.

wetneb:
Hi @leoshrey, do you still intend to work on this?

leoshrey:
https://github.com/OpenRefine/OpenRefine/pull/3406#issuecomment-768137725 Yes, I am working on this. I was just busy with something for the past few days. Now, back in flow.

wetneb:
@leoshrey I am closing this PR to clear up our backlog. If you want to keep working on this do not hesitate to open a new PR. Thanks!

Issue #3401, opened 2020-12-11
Having UX tests running on Edge


fgiroud:
Cypress tests should also runs on Edge

### Proposed solution
Add edge to the build matrix

### Additional context
Edge requires additional packages, if we can find a solution to install them only for the edge build we can save some build time

fgiroud:
Closed with pr #3411

PR #3411, opened 2020-12-16
Added edge to the CI build, #3401

fgiroud:


wetneb:
Merging since this is a change of workflow that can only be tested on master.

Issue #3363, opened 2020-11-26
Normalize casing in UI text in English


wetneb:
We do not use a consistent casing throughout the UI.
For instance, in the "Export" menu:
* "Custom **t**abular **e**xporter" versus "SQL **E**xporter"

In the "SQL Exporter" dialog:
* "Output **e**mpty **row** (i.e. all cells null)" vs "Trim **C**olumn **N**ames"

### Expected Behavior

We should consistently use lowercase since that is what is used in most places so far.

### To solve this issue

Simply go through the existing translation files (in English) and normalize the phrases with incorrect casing.

allanaaa:
![sql](https://user-images.githubusercontent.com/28012148/100376553-b839b680-2fdd-11eb-8ca5-8e8d0afe92e8.PNG)

Thanks for this. Just want to mention an example: "Drop Statement" should become "DROP statement"

swkasica:
This seems like a good issue to tackle as I get familiar with the UI code. I could probably resolve it sometime this week.

thadguidry:
Great Steve! Go for it! Appreciate all the help we can get on our issues and roadmap!

leoshrey:
I would like to tackle it, if @swkasica not proceeding...

swkasica:
@leoshrey I am proceeding.

leoshrey:
Sure thing, please go ahead.

swkasica:
**Quick update**: I'm about halfway through. End of term stuff took longer than I anticipated. Grammar is kind of tricky in UI because in addition to the normal rules of grammar and usage, there's UX specific conventions. I'm adapting [these guidelines from Firefox UX](https://design.firefox.com/photon/copy/capitalization.html) on when to use different capitalization styles: title case, sentence, case, all caps, lowercase, camel case, etc...

SmritiSharmaBG:
Hello, I am new to Open Source. I want to know if this bug is fixed? If it is fixed and merged then why is it still showing up in Issues section? If not, then can I start to work on it? Any help/feedback will be appreciated. Thankyou :)

wetneb:
We should ask @swkasica first, to see if he is still working on this.

swkasica:
Hi all, 

Sorry I haven't had time to work on this issue this month. Academic stuff has consumed all my time. To boot, my computer died a week ago and I don't have a way of running OpenRefine 😢. But I will have more free time by February 4th (and hopefully a real computer too 🤞).

[...]

PR #3436, opened 2020-12-23
Added confirm dialog for starred expression remove links #501

### LINKED ISSUE #501: Add a Confirm dialog for Starred expression Remove links
_Original author: thadguidry (December 01, 2011 15:32:49)_

Using Expression Editor and the Starred expression tab.

It would be nice to have an extra &quot;Confirm ?&quot; popup dialog with REMOVE CANCEL buttons when a user clicks Remove link on a Starred expression.

With the slip of a mouse button, I just lost a very long and nice expression that took me 15 mins to pull together with some research, and probably should have just stored that in a Notepad document and uploaded to Google Docs for safekeeping. :)

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=501_

CameronBedard:
Fixes #501 

Used the built in dialog system to create the small popup asking for confirmation.
Chose the 'Remove Expression?' and the 'OK' and 'Cancel' for the text as they were already in the translation files.

### DIFF
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -244,6 +244,7 @@
     "core-dialogs/sql-exporter": "SQL Exporter",
     "core-dialogs/custom-tab-exp": "Custom Tabular Exporter",
     "core-dialogs/select-columns-dialog": "Select columns",
+    "core-dialogs/unstar-expression": "Unstar Expression?",
     "core-dialogs/content": "Content",
     "core-dialogs/download": "Download",
     "core-dialogs/upload": "Upload",
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -244,7 +244,7 @@
     "core-dialogs/sql-exporter": "SQL Exporter",
     "core-dialogs/custom-tab-exp": "Custom Tabular Exporter",
     "core-dialogs/select-columns-dialog": "Select columns",
-    "core-dialogs/unstar-expression": "Unstar Expression?",
+    "core-dialogs/unstar-expression": "Unstar expression?",
     "core-dialogs/content": "Content",
     "core-dialogs/download": "Download",
     "core-dialogs/upload": "Upload",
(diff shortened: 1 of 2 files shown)


main/webapp/modules/core/scripts/dialogs/expression-preview-dialog.js
```diff
@@ -350,15 +350,30 @@ ExpressionPreviewDialog.Widget.prototype._renderStarredExpressions = function(da
...
-                { expression: entry.code, returnList: true },
-                function(data) {
-                    self._renderStarredExpressions(data);
-                    self._renderExpressionHistoryTab();
-                },
-                "json"
-            );
+            var removeExpression = DialogSystem.createDialog();
+                removeExpression.width("250px");
+            var removeExpressionHead = $('<div></div>').addClass("dialog-header").text($.i18n('core-dialogs/remove') + " " + $.i18n('core-dialogs/expression') + "?")
```
wetneb, line 355: Thanks for your attempt to reuse existing translations! However, we cannot do something like `$.i18n('core-dialogs/remove') + " " + $.i18n('core-dialogs/expression') + "?"` because that assumes that word order and punctuation remain identical in all languages. For instance in German you want to have the noun before the verb in this context. Or in French you want to have a space before the `?` and also an article before "expression". So you should introduce a new bit of UI text in this context. You can add it to the English file only, and translators will take care of adding it to other languages.

wetneb (changes requested):
Thank you! I haven't tried it out yet but I noticed one problem below.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -244,6 +244,7 @@
     "core-dialogs/sql-exporter": "SQL Exporter",
     "core-dialogs/custom-tab-exp": "Custom Tabular Exporter",
     "core-dialogs/select-columns-dialog": "Select columns",
+    "core-dialogs/unstar-expression": "Unstar Expression?",
```
wetneb, line 247:
```suggestion
    "core-dialogs/unstar-expression": "Unstar expression?",
```
We are in the process of normalizing the case of all text in the UI (#3363), so let's get this one right from the start.

wetneb (approved):
Thanks, it works great! One tiny last change below.

PR #3439, opened 2020-12-30
chore: align test-suite name with npm standards

kushthedude:

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,4 +1,4 @@
-name: Java CI
+name: OpenRefine
 
 on: [pull_request_target]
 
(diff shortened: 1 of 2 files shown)


kushthedude:
@wetneb I have also renamed ~Java CI~ OpenRefine, since we have cypress tests as well as server tests and build checks in the CI.

wetneb:
"OpenRefine" seems a bit generic, no?

kushthedude:
> "OpenRefine" seems a bit generic, no?

We can add a CI prefix to the name, and make the jobs name more specific like `server_build`, `server_tests` & `ui_tests`

wetneb:
Yes, I'd say let's change "Java CI" to just "Continuous integration", and then change the "build" job into "server_tests". Using "ui_tests" instead of "cypress_tests" also makes sense, it's more transparent.

wetneb (approved):

PR #3443, opened 2021-01-03
fix broken jdatapath dependency not found

thadguidry:
On the `new-architecture` branch, I couldn't find the old codeberry groupId but see on Maven Central that we have it contained in the `org.openrefine.dependencies` for now, so let's update that pom to make the clean/build seamless for folks like me wanting to help test later.


wetneb (approved):
Yes thanks!

PR #3454, opened 2021-01-07
fix: cache all UI paths for cypress

kushthedude:
- Looks like pull_request_target doesn't build fresh changes in PR, hence replaced it with `on: pull_request`, please ignore the failed check for `pull_request_target` as the same test is getting passed on `pull_request`.
- As well as I have disabled CI on changes only involving docs subdirectory as they are not needed when we do changes in the docs-website.
- I have fixed the cache path to cache the cypress binaries properly.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,9 +1,7 @@
 name: Continuous Integration
 
 on:
-  pull_request:
-    branches:
-      - 'master'
+  pull_request_target:
     paths-ignore:
       - 'docs/**'
 
@@ -123,6 +121,7 @@ jobs:
     - name: Install test dependencies
       run: |
           cd ./main/tests/cypress
+          npm cache clean --force
           npm i -g yarn
           yarn install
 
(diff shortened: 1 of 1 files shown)


wetneb (commented):
It's okay to change from `pull_request_target` to `pull_request` for testing purposes during the PR, to check that the new workflows work, but we need to revert to `pull_request_target` before merging, otherwise builds will break for PRs made from forks because secrets will not be available there.

kushthedude:
@wetneb If we want to use `pull_request_target`, then this has to be 2 PR. In first I will clear the corrupt cache and re-create it and in the second I will remove the line which cleared the cache.
Does this sounds good?

wetneb:
Fine with me then, let me know when this one is ready to be merged.

kushthedude:
> Fine with me then, let me know when this one is ready to be merged.

this is good to go now 😄

fgiroud (approved):

tfmorris:
We need to be careful with the use of `pull_request_target` - https://securitylab.github.com/research/github-actions-preventing-pwn-requests

PR #3467, opened 2021-01-13
Add support for API keys in authentication services

tfmorris:
If indicated as being required by the manifest of the reconciliation service, this PR will send API keys, with the correct parameter name, as advertised in the manifest, with requests to the reconciliation service. It does not support username/password authentication and if it finds a manifest requiring that, it'll inform the user.

This also includes the original implementation for the preferences based authentication support which was my first plan, before I decided it was simpler to just keep everything local to the reconciliation service. I can pull that out into a separate branch (where we could potentially use it for authenticated access for Fetch URL, etc) if desired.


wetneb (commented):
This PR exposes a field for the API key regardless of whether the service requires one or not:
image
I think it would be cleaner to keep this dialog as it was before, with a field for the endpoint URL only, and prompt the user for an API key in a second dialog if the service requires one.
The approach of doing everything in the same dialog does not generalize well if we consider adding support for other authentication methods (as https://github.com/reconciliation-api/specs/pull/41 suggests).

Also, as mentioned on the mailing list, I would avoid introducing a clever URL-matching algorithm since it adds complexity with little benefit to the uninformed user. I think we can simply prompt the user for an API key every time they add a service which requires one, and associate that API to the exact endpoint supplied.

wetneb:
Closing this per inactivity, as there are many merge conflicts.

PR #3471, opened 2021-01-14
Adjust the menuItem class to avoid colliding borders

wetneb:
As suggested by @ostephens orally, this attemps to shrink the blue boxes around menu items:

**Before:**
image
**After:**
image

### DIFF
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -72,8 +72,8 @@
     display: inline-block;
     border: 2px solid #4dc6e1;
     border-radius: var(--ifm-code-border-radius);
-    margin: 0;
-    padding: 1px var(--ifm-code-padding-horizontal);
+    margin: 1px 0px;
+    padding: 0px 4px;
   }
 
   .markdown img {


docs/src/css/custom.css
```diff
@@ -72,8 +72,8 @@
     display: inline-block;
     border: 2px solid #4dc6e1;
     border-radius: var(--ifm-code-border-radius);
-    margin: 0;
-    padding: 1px var(--ifm-code-padding-horizontal);
+    margin: 1px 0px;
```
kushthedude, line 75: Can we use `em` here in place of pixels? Since em is aware of screen resolutions and px is not relative.

ostephens (approved):
LGTM

docs/src/css/custom.css
wetneb: In principle yes, but would you do that even here? If we define the padding and margin in `em` I assume we should do that for the border width as well, no? How would you find the appropriate `em` values?

kushthedude: That would require a certain hit and trial approach but seeing the stylesheets, many pixels are being used, so I guess this all can be handled in a separate PR?

PR #3474, opened 2021-01-14
tests: add test for project_metadata

kushthedude:

Test metadata dialogue and edit


kushthedude:
@wetneb Does it even makes sense to have a PR workflow with `pull_request_target`? As using `pull_request_target` limits the CI to not check the changes made in the PR it is running too which seems only overhaul to me as it doesn't help or let the maintainer know about any failures until & unless that PR is merged into master.

wetneb:
@kushthedude I think you are misunderstanding what `pull_request_target` does. It triggers builds where:
* the build configuration (workflow in Yaml) is taken from the base branch (master)
* the rest of the files (code, tests…) is taken from the branch the PR is made from
So the changes you make in your PR will be checked.

The reason we use `pull_request_target` is because we would not be able to use workflow secrets otherwise (secrets are not made available to pull request build jobs when the PR is made from a fork).

Let me know if you have any other questions.

kushthedude:
> @kushthedude I think you are misunderstanding what `pull_request_target` does. It triggers builds where:
> 
> * the build configuration (workflow in Yaml) is taken from the base branch (master)
> * the rest of the files (code, tests…) is taken from the branch the PR is made from
> 
> So the changes you make in your PR will be checked.
> 
> The reason we use `pull_request_target` is because we would not be able to use workflow secrets otherwise (secrets are not made available to pull request build jobs when the PR is made from a fork).
> 
> Let me know if you have any other questions.

No this is not the current scenario, if you would see the ui_tests in this PR, you wont find the latest tests I have added in this PR.

wetneb:
Ah, good point, there seems to be a problem with our workflow. I would say we should adapt the `checkout` action to fetch the correct commit (although this is still a vulnerability, technically), or use the new `workflow_run` event recently introduced: https://securitylab.github.com/research/github-actions-preventing-pwn-requests

kushthedude:
Strange, the checkout action is checking out the correct latest ref of the commit `a25f3739db5b49c6d719b2a5eb8632b3571754e0` yet the latest tests added in the PR are not being shown in the ui_tests 😕

kushthedude:
@fgiroud , I had to write a different test for testing of each metadata because we use `window.prompt` as an input method in metadata edit dialogue and had to use the stub to pre-populate the window.prompt.

floriangiroudrefinepro:
Tests are looking good !
One thing : I think we should avoid as much as possible to add custom `data-cy` to OpenRefine core codebase, except in a last resort scenario.
In this case, you could simply target the metadata body using `.dialog-body`

Also, do you mind to format your code ;-) ? Jsprettier is a good solution to autoformat JS code. We should have a .prettierrc file to enforce consistent formatting across contributors
Best,
Florian

fgiroud:
Tests are looking good !
One thing : I think we should avoid as much as possible to add custom data-cy to OpenRefine core codebase, except in a last resort scenario.
In this case, you could simply target the metadata body using .dialog-body

Also, do you mind to format your code ;-) ? Jsprettier is a good solution to autoformat JS code. We should have a .prettierrc file to enforce consistent formatting across contributors
Best,
Florian

kushthedude:
@fgiroud I have added prettier and initialised a configuration file. Also, I have run the linter on local to format all outstanding files. Also have removed the data-cy tag from the UI files.

fgiroud (approved):

main/tests/cypress/.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
+    "trailingComma": "es5",
+    "tabWidth": 4,
+    "semi": false,
+    "singleQuote": true
```
tfmorris, line 5: We should get these reviewed by the team before making large-scale changes based on them. 

I think our existing Javascript mostly uses 2 space indenting and trailing semicolons.

main/tests/cypress/cypress/integration/create-project/create_project.spec.js
```diff
@@ -1,78 +1,114 @@
...
-		cy.visitOpenRefine();
-		cy.navigateTo('Create Project');
-		cy.get('#create-project-ui-source-selection-tabs > div').contains('This Computer').click();
-		cy.get('.create-project-ui-source-selection-tab-body.selected').contains('Locate one or more files on your computer to upload');
+    it('Test the create project page', function () {
+        cy.visitOpenRefine()
+        cy.navigateTo('Create Project')
+        cy.get('#create-project-ui-source-selection-tabs > div')
+            .contains('This Computer')
+            .click()
```
tfmorris, line 7: I'm not sure splitting this into multiple lines makes it more readable (which should be the primary goal).

What do others think?

kushthedude: This is done by the Prettier, automated javascript beautifier, splitting into multiple lines makes you clear of the multiple invocations done on the grabbed element.

main/tests/cypress/.prettierrc.json
kushthedude: This is the standard prettier configuration for cypress based project. But if the team has other preferences, I can make those edits accordingly.

tfmorris: I think all of our Javascript should conform to a common style, whether it's test related or not. I don't have a strong preference of 2 vs 4 spaces, but if we're going to switch to 4 spaces, we should do it all at once and put tooling in place to enforce it. Of course, this will introduce a lot of noise across the entire repo, so we should make sure it's worth the cost.

wetneb:
Concerning linting, I don't have a strong opinion about which style to adopt. For me it is okay to have slightly different conventions between tests and UI code since they are clearly distinct code bases, and to start enforcing the linting only on the tests for now (since we are creating this from scratch). I would also prefer that any automated linting is done in an isolated PR, not mixed up with addition of new content.

main/tests/cypress/cypress/integration/create-project/create_project.spec.js
wetneb: The version on multiple lines seems okay to me, I wouldn't actively fight it if it's done by a reasonably standard bit of tooling. For me it's mostly a discussion up to @kushthedude and @fgiroud who will be the ones writing most content in this style in the coming months, so if both of them agree on a style, it works for me.

kushthedude:
> Concerning linting, I don't have a strong opinion about which style to adopt. For me it is okay to have slightly different conventions between tests and UI code since they are clearly distinct code bases, and to start enforcing the linting only on the tests for now (since we are creating this from scratch). But if we can make small config changes in the linter to match the current code style (such as tab width), why not!
> 
> I would also prefer that any automated linting is done in an isolated PR, not mixed up with addition of new content.

Agreed, separating the last commit in a separate PR

kushthedude:
@wetneb Reverted the formatting changes and create #3494

fgiroud:
@kushthedude can you also remove the change in `open-project-ui.js` ?
Thx

main/tests/cypress/cypress/integration/create-project/create_project.spec.js
fgiroud: Agreed, whatever options, it works for me about formatting

kushthedude:
@fgiroud It's good to be merged as well as I resolved all conflicts

main/tests/cypress/cypress/integration/project_management/project_metadata.spec.js
```diff
@@ -122,4 +122,142 @@ describe(__filename, function () {
 		cy.get('#metadata-body tbody>tr').eq(13).contains('Homepage');
 		cy.get('#metadata-body tbody>tr').eq(13).contains('openrefine.org');
 	})
+=======
```
wetneb: Leftover merge conflict marker.

fgiroud (approved):

fgiroud:
Changes related to the tests of project metadata have actually been introduced in [PR 3494](https://github.com/OpenRefine/OpenRefine/pull/3494/files#diff-4e9d29b392f499258ade055876a5bc63d6bc5f4d6edbec26cabdb1a73e27cb09)
Merging this PR would be confusing, as it only contains a code formatting of a few files
It's best if we close this one without merging

kushthedude:
> Changes related to the tests of project metadata have actually been introduced in [PR 3494](https://github.com/OpenRefine/OpenRefine/pull/3494/files#diff-4e9d29b392f499258ade055876a5bc63d6bc5f4d6edbec26cabdb1a73e27cb09)
> Merging this PR would be confusing, as it only contains a code formatting of a few files
> It's best if we close this one without merging

makes sense!

PR #3478, opened 2021-01-16
fix: checkout pull_request head always

kushthedude:

Reference https://github.com/OpenRefine/OpenRefine/pull/3474#issuecomment-761526641


kushthedude:
I hope this should fix the issue of workflows not checking the changes in the PR

wetneb (approved):
This will likely be flagged as a security vulnerability, but let's see.

Issue #3479, opened 2021-01-16
Not all test projects created during ui_testing are getting scraped


kushthedude:
We already have an afterEach hook that deletes all the projects created during the test but due to some bug, not all projects are being deleted.
### To Reproduce
1. Open Cypress application
2. Run a Spec
3. See a few outstanding projects remaining

### Current Results

Few projects remains after successful afterEach hook.
### Expected Behavior

All projects should be flushed out

kushthedude:
This is the error being generated when cypress tries to delete some of those projects
```sh
22:43:38.157 [                   refine] POST /command/core/delete-project (93ms)
22:43:38.157 [                  command] Exception caught (0ms)
java.lang.NullPointerException
	at com.google.refine.commands.project.DeleteProjectCommand.doPost(DeleteProjectCommand.java:64)
	at com.google.refine.RefineServlet.service(RefineServlet.java:187)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:820)
	at org.mortbay.jetty.servlet.ServletHolder.handle(ServletHolder.java:511)
	at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1221)
	at org.mortbay.servlet.UserAgentFilter.doFilter(UserAgentFilter.java:78)
	at org.mortbay.servlet.GzipFilter.doFilter(GzipFilter.java:131)
	at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1212)
	at org.mortbay.jetty.servlet.ServletHandler.handle(ServletHandler.java:399)
	at org.mortbay.jetty.security.SecurityHandler.handle(SecurityHandler.java:216)
	at org.mortbay.jetty.servlet.SessionHandler.handle(SessionHandler.java:182)
	at org.mortbay.jetty.handler.ContextHandler.handle(ContextHandler.java:766)
	at org.mortbay.jetty.webapp.WebAppContext.handle(WebAppContext.java:450)
	at org.mortbay.jetty.handler.HandlerWrapper.handle(HandlerWrapper.java:152)
	at org.mortbay.jetty.Server.handle(Server.java:326)
	at org.mortbay.jetty.HttpConnection.handleRequest(HttpConnection.java:542)
	at org.mortbay.jetty.HttpConnection$RequestHandler.content(HttpConnection.java:945)
	at org.mortbay.jetty.HttpParser.parseNext(HttpParser.java:756)
[...]

PR #3480, opened 2021-01-16
tests: UX Test for project list deletion, sort, filter

### LINKED ISSUE #3413: Add UX Test for the projects list
Improve the tests we have on the projects list page (project management)

Screen Shot 2020-12-16 at 13 08 48

#### Test the "About" panel

* Editing of metadata
* test the panel

#### Test the Delete behavior

Test project deletion and ensure it's been really deleted, not only removed from the list.

#### Test the sorting of tagging list
To be tested

* Sort by tags
* Filter by tags

kushthedude:

Fixes #3413


kushthedude:
The failure here is due to #3479 since few projects from past tests still remained which caused the failure

kushthedude:
Looks like I made a mistake while linking the PR 😅

fgiroud:
@kushthedude , the test is technically valid.
However there might be a more simple way to compare indexes, see below
```javascript
cy.get('a.project-name').contains('projectA').then((projectALink) => {
    cy.get('a.project-name').contains('projectZ').then((projectZLink) => {
        cy.log('Index for project A is ' + projectALink.parent().parent().index() + ', Index for project Z is ' + projectZLink.parent().parent().index());
        // assertion can be placed here
    });
})
```

[This article contains useful resours about variables](https://docs.cypress.io/guides/core-concepts/variables-and-aliases.html#Return-Values)
Basically they recommend to use their own wrapper `.as()`, but you found an interesting way to get the row index with `(rows.index(element))`, does it means it was a closure and you get the result immediately?


Also, do you think it's best to add a unique suffix (Date) to project names to ensure the test is running with fresh projects ? Just because we already have issues with the deletion method, and I wonder if it would fail the test if the project is not deleted properly

kushthedude:
> @kushthedude , the test is technically valid.
> However there might be a more simple way to compare indexes, see below
> 
> ```js
> cy.get('a.project-name').contains('projectA').then((projectALink) => {
>     cy.get('a.project-name').contains('projectZ').then((projectZLink) => {
>         cy.log('Index for project A is ' + projectALink.parent().parent().index() + ', Index for project Z is ' + projectZLink.parent().parent().index());
>         // assertion can be placed here
>     });
> })
> ```
> 
> [This article contains useful resours about variables](https://docs.cypress.io/guides/core-concepts/variables-and-aliases.html#Return-Values)
> Basically they recommend to use their own wrapper `.as()`, but you found an interesting way to get the row index with `(rows.index(element))`, does it means it was a closure and you get the result immediately?

Yes; rows.Index returns the returns immediately. We can use the above code too if it's required, anyway works with me.

>Also, do you think it's best to add a unique suffix (Date) to project names to ensure the test is running with fresh projects ? Just because we already have issues with the deletion method, and I wonder if it would fail the test if the project is not deleted properly

About the test names, I guess we can rather have names as `test${Date.Now()}` rather than `Date.Now()` as it may create hindrance in testing specs for projects. How this sounds to you?

fgiroud:
> About the test names, I guess we can rather have names as `test${Date.Now()}` rather than `Date.Now()` as it may create hindrance in testing specs for projects. How this sounds to you?

Good idea !

kushthedude:
@fgiroud Modified the code here to according to latest changes, please review and merge

fgiroud (approved):

PR #3482, opened 2021-01-16
fix: unwanted characted in version metadata

kushthedude:
fixes #3480 

Thought, there could be some error handling issue causing this symbol appearance but it was a pretty easy fix.

### DIFF
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -3,7 +3,7 @@
     "core-index/slogan": "A power tool for working with messy data",
     "core-index/help": "Help",
     "core-index/about": "About",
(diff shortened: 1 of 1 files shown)


wetneb (approved):

tfmorris (commented):
This doesn't look correct to me. It removes the i18n parameter substitution (ie "weird symbol) which is needed here: https://github.com/OpenRefine/OpenRefine/blob/af1cf375f5d4fe64fbf885529c043691369130da/main/webapp/modules/core/scripts/index.js#L136

If the two different locations where this is referenced have different needs, than we'll need to create different strings.

kushthedude:
@tfmorris Yes, you are correct. I guess we need to create two different strings to accommodate two different locations

Issue #2601, opened 2020-04-26
Excel: support writing very big sheets


pjfanning:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
xlsx files are zip files and things get messy when you have more than 4GB of uncompressed data in file - eg each sheet in a workbook has its file in the zip

**Describe the solution you'd like**
Uptake the solution described in https://github.com/apache/poi/pull/154

**Additional context**
Based on a recent review of the excel support

Shubhangi2901:
Hi @pjfanning I would like to work on this issue

pjfanning:
Thanks @Shubhangi2901 - looking at the Apache POI code again, it may not be necessary to make any extra code changes.

I guess the main issue is to test with enough data to generate an xlsx file with a sheet that is big enough to potentially cause problems.

https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/exporters/XlsExporter.java#L76

The SXSSFWorkbook has a `public void setZip64Mode(Zip64Mode zip64Mode)` but this now defaults to Zip64Mode.AsNeeded. In theory, this should mean that the code automatically switches to a zip format that means that small and big xlsx files should work.

If you have time to try out the latest dev version of OpenRefine to see if it indeed works, that would be great.

PR #3486, opened 2021-01-17
POI 5.0.0 release

pjfanning:
useful for https://github.com/OpenRefine/OpenRefine/issues/2601

Release notes - https://www.apache.org/dyn/closer.lua/poi/release/RELEASE-NOTES.txt
Change log - https://poi.apache.org/changes.html#5.0.0


tfmorris:
Thanks for the contribution. It would be useful to include a link to the release notes as well as a description of how much testing you've done.

Given that this is a major version upgrade for POI, we probably want to be relatively conservative in doing the upgrade and not rush to do it immediately before a little cooling off period to see how stable the release ends up being.

wetneb:
@pjfanning thank you so much for proactively proposing the update! And for your work on POI in general. We just need to decide how this should fit in our release schedule but this should definitely make it to the users eventually.

tfmorris:
Looks like there's already one fix targeted for a 5.0.1 in April. We probably want to continue to monitor this.
https://poi.apache.org/changes.html#5.0.1

wetneb:
This fix looks important: https://github.com/apache/poi/pull/221
I don't know exactly which bug it fixes but by the looks of it it seems to be something we might rely on.
I propose we wait for 5.0.1. The update to that version should be proposed automatically by dependabot so we can close this PR. Or if @pjfanning wants to increment their commit count in OpenRefine, manual PRs are totally welcome too. Thank you again!

pjfanning:
https://github.com/apache/poi/pull/221 affects Powerpoint tables

pjfanning:
next POI release is likely to be v6.0.0

PR #3495, opened 2021-01-18
Broken counts in file selection dialog #3460

### LINKED ISSUE #3460: Broken counts in file selection dialog
The counts of files by extension in the file selection dialog (when importing an archive) are broken, we have "$1" in the UI instead of the actual number.

### To Reproduce
1. First, import the file `https://ndownloader.figshare.com/files/25791104` in OpenRefine
2. Then, see the selection dialog

### Current Results

image

### Expected Behavior

Actual counts should be shown.

### Versions
 - Operating System: Debian
 - Browser Version: Firefox 78
 - JRE or JDK Version: 11
 - OpenRefine: Version 3.5-beta [ff6220a]

S-Harshit:
"Fixes #3460"

Fix for - $1" in the UI instead of the actual number.

Have attached the screenshot for reference. 

image

Please let me know if it solves your purpose as I have recently joined this community.

### DIFF
--- a/main/webapp/modules/core/scripts/index/default-importing-controller/file-selection-panel.js
+++ b/main/webapp/modules/core/scripts/index/default-importing-controller/file-selection-panel.js
@@ -201,7 +201,7 @@ Refine.DefaultImportingController.prototype._renderFileSelectionPanelControlPane
   var renderExtension = function(extension) {
     var tr = table.insertRow(table.rows.length);
     $('<td>').text(extension.extension).appendTo(tr);
-    $('<td>').text($.i18n('core-index-import/file-count'), extension.count).appendTo(tr);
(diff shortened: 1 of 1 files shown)


wetneb:
Hi @S-Harshit, welcome to the project and thank you for this first pull request!

First, a small detail: you have put the issue number in the title of the pull request, but that does not link it formally to the issue. You should write "Closes #3460" or "Fixes #3460" in the description (body text) of the pull request to ensure that if we merge this, the issue will be closed.

Second, the intention of whoever wrote this code was clearly that the UI would show "11 files" and not just "11", which is less understandable as such. So you need to keep the message and fix the way the number is inserted into it. For this, you should only have to move the bracket after the identifier of the UI text after the number.

S-Harshit:
Moving the bracket Has fixed the issue. I have Edited the file and description 

image

wetneb (approved):

PR #3497, opened 2021-01-18
Added 'Store blank columns' importing feature 

### LINKED ISSUE #1750: Remove Blank Columns upon project creation/import

Creating/importing a project allows for the removal of blank rows. It would be great to also be able to remove blank columns.

Have a similar functionality to `Store blank rows` with a `Store blank columns` option. 

**Additional context**
This was discussed on the google group 9/19/2018.

CameronBedard:
Fixes #1750 
5 different importing formats in the main WebApp (not including extensions folder) support the 'store blank rows' feature during the import process. 
Excel, fixed-width, line-based, separator and Wikitext.
These formats all have specific importing classes on the backend but share a 'TabularImportingParserBase', where shared features such as 'store blank rows', 'header rows', 'store blank cells as null' are handled in the importing process. 
Because of the similarities with 'store blank rows' I found it fit to implement the 'store blank columns' here and apply it to all of the tabular importing formats.

It only includes data lines, not header lines, when determining if a column is blank.
In the TabularImportingParserBase class there is a boolean list that represents whether each column has data. At the end of the Tabular Importing Process, if store blank columns is false, this list is iterated through in ImportingUtilities.java and a ColumnRemovalChange is called directly on each column.


wetneb (commented):
Thanks for the contribution!

I think this would be a useful option to have. It would deserve to be exposed not just in the Excel, Fixed-width and Wikitext parser, but also in others (ODS, CSV, JSON, XML…).

It would be good to have tests demonstrating the effect of this option. Ideally they should be testing `TabularImportingParserBase` directly, but if that is not so easy it would also be okay to add the tests to some downstream importers (such as CSV).

wetneb:
Hi @CameronBedard, I am wondering if you intend to finish this up? If you need help to write tests for this let me know.

CameronBedard:
Hi, sorry for leaving this for so long. I do intend to finish. 

Currently the branch supports the feature for all tabular importers (including ODS, CSV). 

But does not support JSON and XML as these fall under the separate TreeImportingParserBase, which is a more complicated piece of code which also didn't have the sister feature 'store blank rows'. This is kind of what I was stuck on. Is it useful to add 'store blank rows' to this section as well?

wetneb:
No worries, there is no rush at all :)

I'd say the option would be less useful for the tree-based importers since they should only create columns when they encounter the corresponding markup / attributes. I am not saying it is impossible for them to create blank columns, but I think the problem is likely to be less widespread than in tabular formats.

wetneb:
I feel bad about closing this because it is a really nice pull request, but it seems to have stalled on the testing side. @CameronBedard we can totally reopen this if you feel like working on it again. Thank you again for your initiative!

wsmmxmm:
https://github.com/wsmmxmm/OpenRefine-master/commit/49982bb693659a118f7cdd74a9acc80987155a03
I found that I made a mistake. I tried and tried but I still couldn't push this branch to the right repo successfully.

antoine2711:
> I tried and tried but I still couldn't push this branch to the right repo successfully.

When I started, I did have problem a few times. At first, I repaired it by extrating the modified pages (13 in your case), and putting them in a new branch. Maybe you can try that?

Regards, Antoine

PR #3508, opened 2021-01-21
Updating codebase links

allanaaa:
Changing obsolete links to the github wiki to the new docs.

I'm sure I haven't caught them all yet, and there is no rush on this just yet.

### DIFF
--- a/.github/workflows/release_manager.sh
+++ b/.github/workflows/release_manager.sh
@@ -2,7 +2,7 @@
 
 
 # Build Release Message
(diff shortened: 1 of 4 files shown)


allanaaa:
I should note that the three wiki pages here that are technically deprecated should be retained because older versions of OpenRefine (& older changlog pages) will continue to link to them. (History, Back Up Your Data, & the Documentation for Users homepage).

Those pages can be blanked + turned into manual redirects if you'd like, or they can stay the same with a warning+link at the top.

wetneb (approved):

PR #3510, opened 2021-01-22
Reduced cypress config numTestsKeptInMemory

fgiroud:
Reduced cypress config numTestsKeptInMemory to 1 as an attempt to fix stability issues on CI


fgiroud:
Since we've reverted Cypress, it looks like our stability issues are fixed, and this PR is not necessary anymore

PR #3506, opened 2021-01-21
test: Improved the way we handle fixtures, #3505

fgiroud:
Refactored the method to load projects, it nows accept either a referenced fixture set (using a string identifier).
@kushthedude, now you can use loadAndVisitProject like this

```javascript
const pdata = [
            ['NDB_No', 'Shrt_Desc', 'Water', 'Energ_Kcal'],
            ['01001', 'BUTTER,WITH SALT', '15.87', '717'],
            ['01002', 'BUTTER,WHIPPED,WITH SALT', '15.87', '717'],
        ]

cy.loadAndVisitProject(pdata)
```

or Like this

```
cy.loadAndVisitProject('food.mini')
```

PR #3519, opened 2021-01-25
feat: add tests for sorting column

### LINKED ISSUE #3418: Add UX tests for the sorting
Enrich and improve the tests we have on the sorting
Test all sorting capabilities of OpenRefine, with a given project

Screen Shot 2020-12-16 at 13 27 47

#### Test for sorting from the column dropdown

Test sorting behaviors from the column dropdown

* test the sorting of the Water column, using text sort, number sort, the sort should be different
* Cast a column to a boolean, test the boolean sort
* Test the text sort a-z, z-a, and with case-sensitive values
* With a given sort, test the different combinations for Blanks, Valid values, Errors

#### Test the sorting from the grid header

Once a sort is added from the column dropdown, sorting controls appear in the grid header

* Ensure the "Sort/remove sort" button is working
* Ensure the "Reorder permanently" is working by refreshing the page
* Test the reverse sort
* Remove the sort

kushthedude:
fixes #3418


fgiroud:
@kushthedude, Since [PR 3506](https://github.com/OpenRefine/OpenRefine/pull/3506), there's a new way to load fixtures
Do you mind to rebase master, then use javascript fixtures inside your test instead of adding csv files ?
Sorry for the trouble
Thx

kushthedude:
Sure the PR is in draft now; will add more tests related to same

kushthedude:
@fgiroud updated the method to load fixtures; please review

fgiroud:
Hi, overall I think that should be better to split the tests in order to test one and only one behavior at the time. It results in more tests, but smaller, and therefore more easy to understands
For example, you're testing the sort from the columns controls and from the viewpanel header at the same time, those should be splitted into
* some tests in grid/column.sort.spec.js
* some tests in grid/viewpanel-header/sort.spec.js

It may make sense to add a new method into support/commands to cast a whole column into the given type, that would make the tests more readable, what do you think?

kushthedude:
> Hi, overall I think that should be better to split the tests in order to test one and only one behavior at the time. It results in more tests, but smaller, and therefore more easy to understands
> For example, you're testing the sort from the columns controls and from the viewpanel header at the same time, those should be splitted into
> 
> * some tests in grid/column.sort.spec.js
> * some tests in grid/viewpanel-header/sort.spec.js
> 
> It may make sense to add a new method into support/commands to cast a whole column into the given type, that would make the tests more readable, what do you think?

Agreed! Making the refactoring as you suggested

kushthedude:
@fgiroud Reordered and rearranged the tests; please review

fgiroud:
Thx @kushthedude , Looks very good to me except one thing: The file should be in in column/sort.spec.js, because sorting is not a submenu of sort-cells

kushthedude:
> Thx @kushthedude , Looks very good to me except one thing: The file should be in in column/sort.spec.js, because sorting is not a submenu of sort-cells

fixed

fgiroud (approved):

PR #3521, opened 2021-01-26
tests: add test for filtering project through tags

kushthedude:

### DIFF
--- a/main/tests/cypress/cypress/integration/open-project/filter_projects.spec.js
+++ b/main/tests/cypress/cypress/integration/open-project/filter_projects.spec.js
@@ -1,0 +1,43 @@
+describe(__filename, function () {
+    it('List an existing project tag, ensure a newly created project tag is created and displayed', function () {
+        const projectName = Date.now()
+        cy.loadProject('food.mini', projectName, 'TestTag')
+        cy.visitOpenRefine()
+        cy.navigateTo('Open Project')
+        cy.get('#projects-list table').contains(projectName)
+        cy.get('#projects-list table').contains('TestTag')
+    })
+    it('Ensure all project tags are created and displayed as filter', function () {
+        const project1 = 'Project A'
+        const project2 = 'Project B'
+        cy.loadProject('food.mini', project1, 'TestTagOne')
+        cy.loadProject('food.mini', project2, 'TestTagTwo')
+        cy.visitOpenRefine()
+        cy.navigateTo('Open Project')
+        cy.get('#projects-list table').contains(project1)
+        cy.get('#projects-list table').contains('TestTagOne')
+        cy.get('#projectTags ul').children().should('contain', 'TestTagOne')
+        cy.get('#projectTags ul').children().should('contain', 'TestTagTwo')
+    })
+
+    it('Ensure projects are being filtered through tags', function () {
+        const project1 = 'Project A'
+        const project2 = 'Project B'
+        cy.loadProject('food.mini', project1, 'TestTagOne')
+        cy.loadProject('food.mini', project2, 'TestTagTwo')
+        cy.visitOpenRefine()
+        cy.navigateTo('Open Project')
+        cy.get('#projects-list table').contains(project1)
+        cy.get('#projects-list table').contains('TestTagOne')
+        cy.get('#projectTags ul').children().contains('TestTagOne').click()
+        cy.get('#projects-list table')
+            .contains(project2)
+            .should('not.be.visible')
+        cy.get('#projects-list table').contains(project1).should('be.visible')
+        cy.get('#projectTags ul').children().contains('TestTagTwo').click()
+        cy.get('#projects-list table').contains(project2).should('be.visible')
+        cy.get('#projects-list table')
+            .contains(project1)
+            .should('not.be.visible')
+    })
+})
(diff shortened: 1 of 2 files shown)


fgiroud (changes requested):
Hi,
There's an issue with the `loadProject` method
When adding a tag, it's not loading the project properly, it's not an issue with your test because you don't check the grid but it will be an issue for other tests

Looks like adding a tag is breaking the import process, see the project in the grid, all columns merged together

My guess is that when adding a tag as an option, you must also pass other parameters like the csv options

If it makes the method loadProject too heavy, you can also create a project, edit a tag manually within the test, and only then do the test

kushthedude:
> Hi,
> There's an issue with the `loadProject` method
> When adding a tag, it's not loading the project properly, it's not an issue with your test because you don't check the grid but it will be an issue for other tests
> 
> Looks like adding a tag is breaking the import process, see the project in the grid, all columns merged together
> 
> My guess is that when adding a tag as an option, you must also pass other parameters like the csv options
> 
> If it makes the method loadProject too heavy, you can also create a project, edit a tag manually within the test, and only then do the test

Let me see into it; however using the second method of creating tag manually will again require UI, I am thinking to get the same result from server for consistency in creation

kushthedude:
@fgiroud Please take a look and review; we need to explicitly describe the importer format when we need to create a project with tags. I know this is a workaround that needs to be changed but it should work for this PR

thadguidry:
@kushthedude @fgiroud 
> Let me see into it; however using the second method of creating tag manually will again require UI, I am thinking to get the same result from server for consistency in creation

Sorry I have not been closely following the testing development as much as I wanted, but the above caught my eye.

Wait, are both of you saying you are not always testing from the UI, but sometimes taking javascript or other shortcuts?  I hope that isn't the case, we need tests to always test from the UI, emulating and mimicking a human driver.  If that isn't always possible, then I'd like to know which areas that is the case.

fgiroud:
@thadguidry 
Thanks for following-up on this, that's an extremely important point.
We are always testing from the UI, using only the same controls than a real user. We never use (at least so far) javascript or any shortcut, that would be a bad idea.

But each test needs to create the initial conditions to be run, usually  by creating a fresh project. This is one of the only two cases where something is done through the OpenRefine API (The other one being setting-up a preference)
And this is this utility method that we are talking about here.
The present test needs to test the projects filtering (by tags), and we are reviewing if we could add this option to the method to create a project (to have them tagged)
Tagging a project would be a separate test, that would need to be done with and only with the UI.

(Actually, to ensure there should be no abuse of OpenRefine API for convenience, those commands are isolated in a dedicated file, `commands/openrefien_api.js`)
Best,
Florian

thadguidry:
> @thadguidry
> Thanks for following-up on this, that's an extremely important point.
> We are always testing from the UI, using only the same controls than a real user. We never use (at least so far) javascript or any shortcut, that would be a bad idea.
> 

Great! Glad we agree!

> But each test needs to create the initial conditions to be run, usually by creating a fresh project. This is one of the only two cases where something is done through the OpenRefine API (The other one being setting-up a preference)

Right understood.  Your THEN: needs some initial setup and scripting, etc. and that is perfectly fine, as long as, we are mimicking those exact steps that a normal human would have done to create those conditions.  Remember that we want to always automate the human, not short-circuiting him/her.

> And this is this utility method that we are talking about here.
> The present test needs to test the projects filtering (by tags), and we are reviewing if we could add this option to the method to create a project (to have them tagged)
> Tagging a project would be a separate test, that would need to be done with and only with the UI.
> 
> (Actually, to ensure there should be no abuse of OpenRefine API for convenience, those commands are isolated in a dedicated file, `commands/openrefine_api.js`)

Right understood.  So here, we are talking a bit about NON-UI testing.  Let's make sure to actually partition those NON-UI tests into other folder structures/scripts.  We don't want to mix NON-UI testing with UI testing together.  Keep things very small, tiny, isolated. The idea is the smallest test possible.  Don't worry about speed or performance, we can always optimize and parallel batch things later on to blaze through tons of tiny small tests.

kushthedude:
To add a little bit more to what @fgiroud said, 
- We are only utilising a small portion of openRefine to create these condition (mostly the create-project-from-upload endpoint), everything else is just manipulation of the query parameters passed to the request data as of which here is ProjectTags.
- All the portion which we have created through openRefine API is also done through UI tests; so there would be no leftover part which is being utilised in our cypress suite but is not being tested.

> Remember that we want to always automate the human, not short-circuiting him/her.

About this we are only utilising the openRefine API to create initial conditions like a project being already present; tags being already added at the time of project creation.

thadguidry:
Thanks @kushthedude @fgiroud and regarding the tiny tests, you DO want to add multiple assertions and not worry about it.
Let's follow those Cypress Best Practices around that: https://docs.cypress.io/guides/references/best-practices.html#Creating-%E2%80%9Ctiny%E2%80%9D-tests-with-a-single-assertion
and actually as much of Cypress's best practices in that doc as we can.
Thanks!

kushthedude:
Hi @fgiroud, There was a small mistake in the queryData which I have fixed now. Please review and merge

fgiroud (approved):

PR #3527, opened 2021-01-27
Prevent snapshot release when only docs push

thadguidry:
I "think" this small change is the right way to prevent unnecessary snapshot release workflow run such as this one:
https://github.com/OpenRefine/OpenRefine/actions/runs/515619980
when the only files changed were in the /docs folders in a PR that was merged into master branch.

Example from: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-positive-and-negative-patterns-1

### DIFF
--- a/.github/workflows/snapshot_release.yml
+++ b/.github/workflows/snapshot_release.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+    paths:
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thanks, let's try it out!

Issue #3509, opened 2021-01-21
Cypress tests are randomly failing


fgiroud:
The CI Cypress test suite is overall flaky, sometimes it passes, sometimes not

Example PRs
* [3508](https://github.com/OpenRefine/OpenRefine/pull/3508/checks?check_run_id=1744850062)
* [3507]()

Most of the time with this weird error message

```
Timed out waiting for the browser to connect. Retrying...
Warning: Cypress failed to connect to Chrome via stdio after 1 minute, 0 seconds. Falling back to TCP...
Connecting to Chrome via TCP was successful, continuing with tests.
```
Coming from Cypress itself ? Introduced in Cypress 6.3 ? To be investigated.
Looks like there are several different issues.
The Cypress dashboard will help to troubleshoot.

skauffmann:
Link to the following cypress issue : https://github.com/cypress-io/cypress/issues/14701

fgiroud:
See Also this issue [Error: Could not process 'take:screenshot'](https://github.com/cypress-io/cypress/issues/14663)
We will revert to Cypress 6.2 if we don't see any progress there

tfmorris:
I would suggest falling back to Cypress 6.2.1 now, unless there's a critical 6.3 feature that we need. No sense in letting another project's bug cause us extra work.

wetneb:
Or temporarily running only tests on Edge (and possibly Firefox), since most tests should not exhibit browser-dependent behaviour.

kushthedude:
I would also suggest to fallback to 6.2.1 and let dependabot ping us on the next minor release. Not sure if edge would give different results as it is based on chromium.

rooby:
Release 6.4 has an item relating to this:

https://github.com/cypress-io/cypress/releases/tag/v6.4.0
https://github.com/cypress-io/cypress/pull/14819

PR #3531, opened 2021-01-28
fix: Reverted Cypress to 6.2.1, #3509

fgiroud:
Reverted Cypress, too much instability with the latest version
#3509


kushthedude:
@fgiroud we need to update yarn.lock too, I think you missed it by mistake. I have made this PR to fix this https://github.com/OpenRefine/OpenRefine/pull/3533/files

PR #3534, opened 2021-01-28
test: Refactored the facet test, added more cases

fgiroud:
Added as many cases as possible to test all facets controls and behavior, in a generic manner
This is not specific to one facet
@kushthedude I am especially interested to know your views on the tests readability and quality. The goal is to have tests that are easy to understand, even without having Cypress running


main/tests/cypress/cypress/integration/project/grid/column/facet/facets.spec.js
```diff
@@ -1,184 +1,417 @@
...
+			.find('.facet-choice')
+			.contains('ALLSPICE,GROUND')
+			.parent()
+			.find('.facet-choice-toggle')
+			.click()
+		cy.getCell(0, 'Shrt_Desc').should('to.contain', 'ALLSPICE,GROUND')
+		cy.get('#tool-panel').contains('1 matching rows')
+
+		// now invert, expect 198
+		cy.getFacetContainer('Shrt_Desc').find('a[bind="invertButton"]').click()
```
thadguidry, line 288:
```suggestion
		cy.getFacetContainer('Shrt_Desc').get('a[bind="invertButton"]').as('invertButton')
		cy.get('@invertButton').click()
```

Hmm, save yourself lots of typing throughout tests by using more aliases `@` syntax, such as above.
find() gets descendents but you cannot alias it like with get()
Please read:
https://docs.cypress.io/api/commands/get.html#Alias
https://docs.cypress.io/guides/core-concepts/variables-and-aliases.html#Aliases

thadguidry: I wonder if Cypress has some linting help to spot things like all those same find()'s and suggest to use alias?  Kinda like Sonarlint does with Java where there's sometimes repeated statements and then asking to refactor to simplify?  Maybe sometime wrote a plugin for Cypress to catch things like that, dunno.

thadguidry: It might also make sense, or perhaps not, to use shorthand notation via this plugin: https://github.com/FlorianGoussin/cypress-cy-select  which I found while browsing through all the various plugins here: https://docs.cypress.io/plugins/

main/tests/cypress/cypress/integration/project/grid/column/facet/facets.spec.js
```diff
@@ -1,184 +1,417 @@
...
+
+	it('Create a simple text facet, check various elements in a Facet', function () {
+		cy.loadAndVisitProject('food.small')
+		cy.columnActionClick('Water', ['Facet', 'Text facet'])
+		cy.get('#refine-tabs-facets').should('exist')
+
+		cy.getFacetContainer('Water').should('exist')
+		cy.getFacetContainer('Water').contains('182 choices')
+		cy.getFacetContainer('Water').contains('Sort by')
+		cy.getFacetContainer('Water').contains('Cluster')
```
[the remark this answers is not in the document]
fgiroud: @thadguidry this syntax will not work, because `get` only accepts a selector.
Moreover, `cy.get` always starts from the entire document, [explained here](https://docs.cypress.io/api/commands/get.html#Get-in-within) 
You could use `within`, but actually if you just want to refine the selector and ensure it's a button, you could simply replace the line by `cy.getFacetContainer('Water').find('button').contains('Cluster')` (tested and working)

main/tests/cypress/cypress/integration/project/grid/column/facet/facets.spec.js
fgiroud: Regarding use of `as`, yes it will work in that case (tested), but it's actually a double edge sword.
As soon as an element is detached from the DOM, Cypress will loose the reference, and we never know, and can't know when does OpenRefine attach/detach/replace entire portions of the dom.
So it would make the tests difficult to troubleshoot, especially for newcomers. 
I would rather keep this verbose syntax, for consistency across tests and scenarios.

Regarding adding `data-cy`, well, I am not sure it's a good idea or not. For now I've recommenced to never touch OpenRefine codebase, so it makes PRs easy to merge, and allows newcomers to write some tests without worrying to break anything.
What could be quite usefull actually, is to have a better syntax to select DOM elements by the `bind` attribute, because OR heavily rely on this to structure the DOM, not sure yet what's the best way to implement this

fgiroud: > I wonder if Cypress has some linting help to spot things like all those same find()'s and suggest to use alias? Kinda like Sonarlint does with Java where there's sometimes repeated statements and then asking to refactor to simplify? Maybe sometime wrote a plugin for Cypress to catch things like that, dunno.

There issue #3536 where we can discuss linting

main/tests/cypress/cypress/integration/project/grid/column/facet/facets.spec.js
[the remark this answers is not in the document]
thadguidry: Ah, OK.

main/tests/cypress/cypress/integration/project/grid/column/facet/facets.spec.js
thadguidry: OK thanks for explaining.
I thought I saw some plugins to help you select for `bind` attributes?

thadguidry:
@fgiroud I don't have anything else to add for a review here. It's a good enough start on our facet testing I think for the community to continue as needed.

PR #3538, opened 2021-01-28
skip building with Appveyor for only docs

thadguidry:


appveyor.yml
```diff
@@ -9,6 +9,11 @@ init:
 
 clone_depth: 5
 skip_branch_with_pr: true
+skip_commits:
+  files:
+    - docs/
+    - '**/*.md'
+    - '**/*.html'
```
wetneb, line 16: `docs/` looks safe enough, but I am not too sure about the rest. In theory we could break tests by modifying Markdown or HTML files in the application itself. Which files where you thinking of?

thadguidry: oof. You are right.  Let's limit to just the `docs/` folder then.

wetneb (approved):

Issue #3423, opened 2020-12-16
Add UX Test for column / Edit cells


fgiroud:
To test every behavior under Column / Edit Cells
 
 
![Screen Shot 2020-12-16 at 13 50 24](https://user-images.githubusercontent.com/6267288/102350838-cf990d80-3fa5-11eb-9f45-16bf1e739766.png)

### Proposed solution

Add multiple tests

#### Transform

Add a simple test to append a string to a value `value+"test"`
Ensure the modification is in the grid 

#### Common Transforms

Those transforms are very straightforward and do not need to be specified.   
Simplify add a test for each one of them and ensure the values are properly modified   

#### Fill down, Blank Down

Use "Sugar Tot" on food.small dataset and 
* ensure lines 8,10 have the value of the previous cell when using "Fill down"
* ensure line 2 is blank when using "Fill blank"

#### Split multi-valued cells

Ensure several rows are created when using "Split multi-valued cells" on "Shrt_Desc"
Ensure no new rows are created when using "Split multi-valued cells" on "Water" (or any column that does not contains ",")

#### Join multi-valued cells

How does this feature works?

#### Cluster and edit

See #3422 

#### Replace

Replace a simple string in column "Shrt_Desc", such as "BUTTER" replaced by "TEST" and ensure the modification is in the grid.

fgiroud:
I am curious about the expected behavior of split & join
See the PR

### Split

#### Splitting by length 

When splitting a cell by field length, given one single number as input, the function doesn't do anything
Bug or expected behavior ?

#### Splitting by transition

Split by transition from lowercase to uppercase
Given a grid with one single cell `BuTtEr` 
I am expecting to have one row per letter, if my understanding is correct, each time there's a transition, there should be a new cell createdright ?
So I am expecting
``` 
B
u
T
t
E
r
```

Instead I've got
```
Bu
Tt
Er
```

Bug or expected behavior ?

#### Splitting by transition

Using the examples given by the UX, the function doesn't produce any results
Tried for example
Two cells `11Abc`, `Def22` then `split by transition from lowercase to uppercase`, but it doesn't do anything
Is that the example that is not good, or OpenRefine having a bug ?

Same with `11` and `AbcDef22`. Clicking on `Test a split by transition from numbers to letters` doesn't produce any results

fgiroud:
Looks like all of those are expected behaviors of OpenRefine
Splitting by transition works when a part of the strings is in lowercase, followed by uppercase

[...]

PR #3541, opened 2021-01-29
test: Added test cases for edit Cells, #3423

fgiroud:
I think this PR gives enough coverage about Edit cells to let the community implement the remaining cases, it's relatively simple
Will split issue #3423 into 2,3 separate issues once this is merged


kushthedude (commented):
Tests look great! But I had a doubt since all of the tests are just suboptions for common-transforms, do you think we should just club them all under common-transforms.spec.js 🤔 ?

fgiroud:
> Tests look great! But I had a doubt since all of the tests are just suboptions for common-transforms, do you think we should just club them all under common-transforms.spec.js 🤔 ?

No, because we want to adhere to the organisation of menus in OR, hence we have one file per menu entry.

kushthedude (approved):

PR #3542, opened 2021-01-30
fixex #2519 hyperlink issue

singhakshita:
Screenshot 2021-01-30 at 4 13 11 PM

hyperlink issue fixed 
please reveiw and let me know if any change is required

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
@@ -58,6 +58,16 @@ DataTableCellUI.previewMatchedCells = true;
   });
 })();
 
+function validURL(str) {
+  var pattern = new RegExp('^(https?:\\/\\/)?'+ // protocol
+    '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|'+ // domain name
+    '((\\d{1,3}\\.){3}\\d{1,3}))'+ // OR ip (v4) address
+    '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*'+ // port and path
+    '(\\?[;&a-z\\d%_.~+=-]*)?'+ // query string
+    '(\\#[-a-z\\d_]*)?$','i'); // fragment locator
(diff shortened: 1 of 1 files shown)


thadguidry:
Hi @singhakshita instead of using regex directly, can you instead use the URL detection as mentioned in my comment on the issue?  Previous contributor did nearly exactly what you have done here: https://github.com/OpenRefine/OpenRefine/pull/2540  but our project agreed that perhaps we can just use the browsers built-in handling now (and not worry about Internet Explorer).

singhakshita:
yeah sure , sorry didn't go through the previous commits . i m working on it .
thanks for the response

singhakshita:
can we do this by importing a validator library of npm??

thadguidry:
No need, read my comment on the issue: https://github.com/OpenRefine/OpenRefine/issues/2519#issuecomment-612645821

singhakshita:
do you want to remove the the validation check if the type is chosen as "text".?

singhakshita:
are we supposed to remove the any such validation if the type = "text" ?
because cleaning up the code and keeping only that what you mentioned is not showing valid for even valid links .

thadguidry:
The [URL interface](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) and its functions should be used in order to check the entire cell value in OpenRefine.  If you use my reproducing steps in the original issue, and use OpenRefine, you will see how it works (or doesn't work).

Regarding OpenRefine cell types:  Later on we plan to introduce various rules (Data Validation) for different OpenRefine cell types, but in current code we only have some lightweight scripting in place as you see in parts.  So for now, only do URL validation when the cell type = 'text'.

You can start OpenRefine and use our Clipboard importer from Create Project to reproduce using a simple column containing these 4 values where only the 4th row should be represented as a valid URL link and the first 3 rows should be strings (however all 4 rows in OpenRefine are treated by default as Text for the cell type).

You can use the Column menu to change the column type to test how it works with Edit cells -> Common transforms -> To Number|To Date|To Text

```
abc
123
https://www.google.com
```

PR #3543, opened 2021-01-31
Pressing "Enter" in the project name field at the preview stage will now create the project

2001lyshis:
This does not use the submit event, because there is no `<form>` in it

#3137 

test


wetneb:
Hi @2001lyshis, thanks for this pull request!
However, @leoshrey has already submitted a similar pull request for this issue: #3406.
Please only start working on issues for which no one is assigned yet and there is no open pull request. Thank you!

PR #3544, opened 2021-01-31
fix for issue #2519

singhakshita:
Screenshot 2021-01-31 at 1 06 39 PM

plese reveiw it once and let me know it any changes are required

### DIFF
--- a/main/webapp/modules/core/scripts/util/url.js
+++ b/main/webapp/modules/core/scripts/util/url.js
@@ -102,3 +102,5 @@ URL.looksLikeUrl = function(s) {
   return false;
 };
 
+
+
--- a/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
@@ -98,30 +98,33 @@ DataTableCellUI.prototype._render = function() {
       nonstringSpan.className = 'data-table-value-nonstring';
       nonstringSpan.textContent = cell.v;
       divContent.appendChild(nonstringSpan);
-    } else if (URL.looksLikeUrl(cell.v)) {
+    } else {
       var arr = cell.v.split(" ");
-      console.log(arr.length);
+      var spanArr =[];
       for( var i=0;i<arr.length;i++){
         if(URL.looksLikeUrl(arr[i])){
-          console.log(arr[i]+"url");
+          if(spanArr.length !=0 ){
+            var span = document.createElement('span');
+            span.textContent = spanArr.join(" ");
+            divContent.appendChild(span).appendChild(document.createTextNode('\u00A0'));
+            spanArr = [];
+          }
           var url = document.createElement('a');
           url.textContent = arr[i];
           url.setAttribute('href', arr[i]);
           url.setAttribute('target', '_blank');
-          divContent.appendChild(url);
+          divContent.appendChild(url).appendChild(document.createTextNode('\u00A0'));
         }
         else{
-          console.log(arr[i]+"span")
-          var span = document.createElement('span');
-          span.textContent = arr[i];
-          divContent.appendChild(span);
+          spanArr.push(arr[i]);
         }
       }
-    } else {
-      var span = document.createElement('span');
-      span.textContent = cell.v;
-      divContent.appendChild(span);
-    }
+      if(spanArr.length !=0){
+        var span = document.createElement('span');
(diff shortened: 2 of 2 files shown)


thadguidry:
OK, checking it out now...stay tuned...let me test a few things.
1 thing to note, I don't they you need to keep the console logging anymore right? You probably only needed that for testing yourself?

singhakshita:
yeah that console ,i just forgot to remove

singhakshita:
@thadguidry once u check it ,i will together do all the changes  ,suggested by you. 
yeah that console was just for testing purpose ,i just forgot to remove

thadguidry:
@singhakshita Tested and the cell's original value is not kept intact.  For instance, using my test string:
`https://www.wikidata.org/wiki/Property:P670 https://www.wikidata.org/wiki/Property:P669 are now mapped to https://schema.org/streetAddress via https://www.wikidata.org/wiki/Property:P2235`

The `are now mapped to` is changed to `arenowmappedto` where whitespace has been collapsed and not kept.
This in turn also causes extra <span> elements where it should ideally only be a single `<span>` or even none at all actually.
```html
<a class="data-table-cell-edit" title="Edit this cell" href="javascript:{}" style="visibility: hidden;">&nbsp;</a><a href="https://www.wikidata.org/wiki/Property:P670" target="_blank">https://www.wikidata.org/wiki/Property:P670</a><a href="https://www.wikidata.org/wiki/Property:P669" target="_blank">https://www.wikidata.org/wiki/Property:P669</a><span>are</span><span>now</span><span>mapped</span><span>to</span><a href="https://schema.org/streetAddress" target="_blank">https://schema.org/streetAddress</a><span>via</span><a href="https://www.wikidata.org/wiki/Property:P2235" target="_blank">https://www.wikidata.org/wiki/Property:P2235</a>
```
image

I expected it to look something like `...</a>are now mapped to<a href="...`
and the same for the `via` as well where whitespace was not kept but collapsed.

If you were not already aware, it's an important feature in OpenRefine that users always have complete control over cell value changes, and we never change those values until they ask us to perform some functions, transforms, scripting, etc.

singhakshita:
yeah i got your point completely .try to fix this as soon as possible

singhakshita:
@thadguidry  sorry i previously ignored few problems

this is completely tested from my side, this should work completely please review its once

thadguidry:
@singhakshita I think you've introduced somehow a LF or CR and extra whitespace?

See the LEFT which is AFTER your change and compare to RIGHT which is BEFORE your change.

image

singhakshita:
Screenshot 2021-02-02 at 9 35 47 AM

there wasn't any extra space in between .
yes but at the end of line one space was being introduced ,that has been checked 

also it may seem like it is introducing number of spaces in the "are now map to" but as inspecting in html its just one span and one space .

thadguidry:
@singhakshita I think this looks great now!  It would be great to get some UI tests added for this.  Do you think you could do that?  Work with @fgiroud and @kushthedude to add them?
I'd probably just add my test string which gives good example and probably 1 more with a date without a space between that should not be a url:
`2021-01-31https://www.google.com`

We're so close to merging this now! Congrats @singhakshita !

singhakshita:
I m completely new to this  so i have no clue of how to add test , is there any documentation that i can refer to regarding testing

fgiroud:
@singhakshita , we are using the [Cypress framework](https://docs.cypress.io/guides/overview/why-cypress.html#In-a-nutshell)
If you want to add some tests (which I recommend) you should first get familiar with Cypress testing.
Then you can read the [OpenRefine documentation section about testing](https://docs.openrefine.org/technical-reference/functional-tests)
From there you can run the test suite locally to review what's already implemented, and add your new tests

singhakshita:
[tests for hyperlink #2519] i have added the tests please review it once

singhakshita:
https://github.com/OpenRefine/OpenRefine/pull/3588 i had to open a different pull request for this , i have added some tests please review it once @fgiroud

singhakshita:
@thadguidry i had to open another pull request (due to some technical issues with my IDE)  .i have added the tests and also pushed the initial changes here https://github.com/OpenRefine/OpenRefine/pull/3588 please review it once

thadguidry:
@singhakshita OK, then go ahead and close this PR if it's not needed and all your work is in the PR #3588 .  I have started review of PR #3588

PR #3550, opened 2021-02-01
feat: add tests for pagination and pagesize

### LINKED ISSUE #3417: Add UX Test for the grid header / pagination
Add tests for the header block that sits on top of the grid (pagination)

Screen Shot 2020-12-16 at 13 21 58

### Proposed solution

Add a pagination test to

* Ensure 5,10,25 rows to be shown when changing pagination
* Test the first, previous, next, last buttons
* Test entering an arbitrary page number
* Test entering a page number that does not exist

kushthedude:
fixes #3417


fgiroud:
@kushthedude for the test about the pagination size, looks like there is a lot of duplicated code
Could you simplify the test with one single assertion that just count the number of rows ? We don't necessarily need to validate the content of the cells in that case, but if you think that's absolutely necessary you could use the utility method `assertCellEquals`

For the first,next,previous,last, I don't think that's necessary to duplicate the test for each pagination size, but I like the way you've done it, separating the first,previous,next,last into separate tests.
Looks like the `previous` button is not tested

kushthedude:
> Could you simplify the test with one single assertion that just count the number of rows? 

I was sceptical here as if the actual transition is being done or not; so just for greater confidence I asserted cell value too but I don't think it's absolutely necessary.

> I don't think that's necessary to duplicate the test for each pagination size

I think it's better to have better coverage of each functionality including different options too, right?

>Looks like the previous button is not tested

Sorry, I missed it adding it/

kushthedude:
Done!

fgiroud (approved):

PR #3556, opened 2021-02-02
Added Links

risav-sarkar:
Added links of Jdk and Apache Maven for ease of use

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ If you have cloned this repository to your computer, you can run OpenRefine with
 * `./refine` on Mac OS and Linux
 * `refine.bat` on Windows
 
-This requires JDK 8 and Apache Maven.
+This requires [JDK 8](https://www.oracle.com/in/java/technologies/javase-downloads.html) and [Apache Maven](https://maven.apache.org/).
 
 Documentation and Videos
 -------------------------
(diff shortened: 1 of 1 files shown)


thadguidry:
@risav-sarkar I'm not sure we can truly say that to **Run from source** as a developer you need JDK 8 any longer.  It builds and runs fine with JDK 15 (backward compatibility) from [https://jdk.java.net/](https://jdk.java.net/) where that generic JDK download link is probably a better link to use.  I'm not sure we even want to keep that section in our Readme, since we'll be changing many things about our developer docs and improving them.  For instance, we already have lots of detail and that needs updating on this Wiki page: https://github.com/OpenRefine/OpenRefine/wiki/Building-OpenRefine-From-Source

risav-sarkar:
I had given the link of the latest JDK(https://www.oracle.com/in/java/technologies/javase-downloads.html). Should I change it to (https://jdk.java.net/) ?

wetneb:
I think it's a good idea to add links here! For the JDK, I would point people to https://adoptopenjdk.net/ instead, no?

thadguidry:
@wetneb It's changing however what we already have on the User manual.  OpenJDK page points to the same link https://jdk.java.net from https://openjdk.java.net/

wetneb (approved):

PR #3557, opened 2021-02-02
Update Java version

thadguidry:
since the older JDK 14 version page says its been superseded with the new version

### DIFF
--- a/docs/docs/manual/installing.md
+++ b/docs/docs/manual/installing.md
@@ -4,19 +4,23 @@ title: Installing OpenRefine
 sidebar_label: Installing
 ---
 
-## System requirements 
+## System requirements
 
-OpenRefine does not require internet access to run its basic functions. Once you download and install it, it runs as a small web server on your own computer, and you access that local web server by using your browser. It only requires an internet connection to import data from the web, reconcile data using a web service, or export data to the web. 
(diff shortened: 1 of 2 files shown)


wetneb (commented):
Hmm, actually users should not need a JDK, only a JRE (Java Runtime Environment), so we should update this page to replace all mentions of "JDK" by "JRE".

thadguidry:
@wetneb The **Run from source** section was originally put on the Readme prior to our Snapshot releases, and also to give devs, not users, a quick glance at how to run OpenRefine.  I don't think we need that section at all anymore on the Readme.  Instead, I'd remove it since we already have a links further down for Contribution (Developers Guide & Architecture).

thadguidry:
@wetneb In that case actually...we should probably just tell them about https://www.java.com to download JRE and the rest of all those technical details actually should go into the Technical docs, I would think?  I mean, for User docs, do we really need to go into all the details for "Install or upgrade Java"?

allanaaa:
Oh! How interesting. Is JDK required for anything? Development?

wetneb:
Yes JDK is required for development only.

thadguidry:
@allanaaa Thanks this looks much better and makes more sense.

allanaaa:
If we're good with this, let's merge it - there are people testing the docs right now. :)

wetneb (approved):

PR #3570, opened 2021-02-04
Let's move the Readme for UX tests into the main/tests folder

thadguidry:
this will make it easier to also see from GitHub perspective if news devs are just browsing.


wetneb (commented):
Since it applies to the UX tests, I think we should keep it in the folder for UX tests.

thadguidry:
@wetneb renamed the title so it's more clear.  Perhaps you missed the **Files changed** however?
** main/tests/cypress/Readme.md → main/tests/Readme.md **

wetneb:
Thad,

The directory `main/tests` currently contains both backend tests written in Java and integration tests with Cypress. The README file you moved is specific to the integration tests, with Cypress. Therefore it makes sense to keep this file in a directory which is specific to Cypress tests. Which is its current location. So I would close this PR without merging it.

Is this clear?

thadguidry:
Ah! Sorry, didn't catch the backend tests.  Yeah, closing.

PR #3576, opened 2021-02-04
Enable docusaurus link checker.

### LINKED ISSUE #3529: Add link checker to documentation CI
The dead links from #3524 are something that we should be able to catch by checking the links when running CI on the docs build.

W3C has an online checker here: http://validator.w3.org/checklink?uri=https%3A%2F%2Fdocs.openrefine.org%2F&hide_type=all&depth=&check=Check
but there are local utilities which can be used as well.

wetneb:
Closes #3529.


docs/docusaurus.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  onBrokenLinks: 'warn',
+  onBrokenLinks: 'error',
```
tfmorris, line 2: This line probably can be removed. I think I added it because when we upgraded docusaurus versions, they changed the defaults and I didn't want to be blocked on fixing broken links. If all the links have been fixed, removing the line should return to the default behavior of `error`.

wetneb: Don't we also want to be protected against changes of defaults on Docusaurus' side? If this is a feature we explicitly want to rely on, perhaps it's useful to state it as such? (No strong feelings of course…)

tfmorris: That sounds fine.

PR #3585, opened 2021-02-05
Attempt to run the CI on the test merge instead of the branch

### LINKED ISSUE #3571: Pull request workflow should test the merged pull request, not the branch
It seems that our GitHub Actions workflow to test pull requests is checking out the branch that the PR intends to merge, without generating the merge commit (which includes further changes on master).

We should fix this, using one of these options:
* fix the `checkout` action to use the correct commit id
* go back to the `pull_request` event and the default checkout options for it (requires not relying on secrets)
* configure the GitHub project to only make PRs mergeable when they are up to date with master (this adds a button on the PRs to generate these merge commits automatically).

wetneb:
This hopefully closes #3571, we'll need to try it out.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -42,7 +42,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2.3.4
       with:
-        ref: ${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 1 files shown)


wetneb:
Since the PR uses the existing workflow, we need to merge this to try it out. I will do it now and stand ready to revert it if it does not work.

PR #3592, opened 2021-02-07
Fix #3089 issue

### LINKED ISSUE #3089: Missing internationalization (i18n) on About page
None of the text on the About page is internationalized so that it can be translated into languages other than English.

The individual HTML elements should be given unique ids which can be used with the `$.i18n()` calls to set the text for the element based on the user's current language. The license text and copyright statement should **not* be translated.

These elements should be translated and their text moved into the English translations file, referenced by the element IDs (the first two can be combined for consistency and to minimized translation):
https://github.com/OpenRefine/OpenRefine/blob/fc21d58ed11af8e512be5fd5852a281c88791fb7/main/webapp/modules/core/about.html#L38
https://github.com/OpenRefine/OpenRefine/blob/fc21d58ed11af8e512be5fd5852a281c88791fb7/main/webapp/modules/core/about.html#L67
https://github.com/OpenRefine/OpenRefine/blob/fc21d58ed11af8e512be5fd5852a281c88791fb7/main/webapp/modules/core/about.html#L88
https://github.com/OpenRefine/OpenRefine/blob/fc21d58ed11af8e512be5fd5852a281c88791fb7/main/webapp/modules/core/about.html#L110

The two main paragraphs of text can translated independently to minimize the HTML markup that they include and the links and anchor text for Metaweb and Google (e.g. `<a href="http://www.metaweb.com/">Metaweb Technologies, Inc.</a>`) can be substituted in as parameters (ie `$1`, `$2`) to the `$.i18n()` call since they don't need to be translated.

singhakshita:
Screenshot 2021-02-07 at 9 30 10 PM

internationalisation added on about page


wetneb (changes requested):
Thank you very much!

It looks like you have used machine translation to translate the text in many languages. It would be better not to do it, since there is no quality guarantee for these translations. (For instance, "OpenRefine" is translated to "OpenSify" in many languages - that is a big problem and there could be more like this).

More comments inline.

main/webapp/modules/core/about.html
```diff
@@ -62,30 +62,32 @@
...
         sources. OpenRefine is not a web service but a desktop app that runs on your own computer,
         so you can process sensitive data with privacy.
       </p>
 
-      <p>
-        OpenRefine was originally developed as "Freebase Gridworks" by
-        <a href="http://www.metaweb.com/">Metaweb Technologies, Inc.</a>. Metaweb was acquired by
-        <a href="http://www.google.com/">Google</a> in July 2010 and they renamed the product
+      <p id="history-openrefine">
+        <!-- <p id="history-1"  style="display: inline-block;">OpenRefine was originally developed as "Freebase Gridworks" by</p>
```
wetneb, line 79: Markup that is commented out should be removed.

main/webapp/modules/core/about.js
```diff
@@ -0,0 +1,19 @@
...
+let link_1 = document.createElement('a');
+link_1.setAttribute('href' ,"http://www.metaweb.com/")
+link_1.textContent ="http://www.metaweb.com/";
+const history_2 = document.createTextNode($.i18n('core-index/history-2'));
+let link_2 = document.createElement('a');
+link_2.setAttribute('href' ,"http://www.google.com/")
+link_2.textContent ="http://www.google.com/";
+const history_3 = document.createTextNode($.i18n('core-index/history-3'));
+ 
+$("#history-openrefine").append(history_1 ," " ,link_1, " " ,"Metaweb Technologies, Inc."," ",history_2," ",link_2," ","Google"," " ,history_3);
```
wetneb, line 15: I appreciate your effort to keep HTML markup out of the translatable strings but I am worried that this will give incomplete strings to translate, such as `". Metaweb was acquired by"`.

It would be better to include the whole history text as a single string in which the links are introduced as parameters (or leave all of the HTML markup in the string to translate).

singhakshita:
@wetneb  if i pass links as parameter it is not displayed as links its just displayed as normal text  . we have a fix for this in react but i am not sure how to do this in jquery and json ,html , 
i am looking for it . 
i just found one approach to due it to break it as i did ,
can you please guide me to any reference or anything as such to find this solution.

main/webapp/modules/core/about.html
```diff
@@ -62,30 +62,32 @@
...
     <div id="body-info">
-      <h1>About OpenRefine</h1>
+      <h1 id="about-openrefine">About OpenRefine</h1>
       <!-- The following two IDs must match those in index.vt/index.js -->
       <h2 id="openrefine-version"></h2>
       <h3 id="java-runtime-version"></h3>
       
-      <p>
+      <p id="definition">
         OpenRefine is a power tool for working with messy data. Use it to improve data consistency,
```
wetneb, line 71: Text content that is covered by i18n should not be in the HTML template anymore (otherwise people will be tempted to make edits to the HTML file directly but that will not have any effect since the text will be overridden by i18n).

wetneb (changes requested):
Thanks! Please also remove all the translations except the English ones (if you master some other language yourself it is also fine to leave this language). We do not want any automated translation.

wetneb:
Another comment: please use an informative title for your pull request, and add the "Fix #3089" in the description (body text) of the pull request instead, so that GitHub can properly link the pull request to the issue.

PR #3595, opened 2021-02-07
Attempts to fix Flaky tests, Issue 3594

fgiroud:
Increased retries
Improved method to click on column menus, now waiting for Or to not be updating anything

### DIFF
--- a/main/tests/cypress/cypress.json
+++ b/main/tests/cypress/cypress.json
@@ -1,11 +1,11 @@
 {
-  "integrationFolder": "./cypress/integration",
-  "nodeVersion": "system",
-  "retries": {
-    "runMode": 2,
-    "openMode": 1
-  },
-  "env": {
-    "OPENREFINE_URL": "http://localhost:3333"
-  }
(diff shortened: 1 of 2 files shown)


main/tests/cypress/cypress.json
```diff
@@ -2,8 +2,8 @@
   "integrationFolder": "./cypress/integration",
   "nodeVersion": "system",
   "retries": {
-    "runMode": 2,
-    "openMode": 1
+    "runMode": 3,
+    "openMode": 0
```
kushthedude, line 6: @fgiroud Why is there need to reduce retry in openMode? It helps in debugging the test through cypress binary.

fgiroud: @kushthedude I found it not practical to have local retries, if some retries are necessary locally, it means there is probably an issue.
When a test fail, the dev needs to investigate quickly and retry manually after fix, local retries make the process cumbersome
Retries should be kept for CI/CD

kushthedude (approved):
Looks good, triggered the ci twice all green

PR #3591, opened 2021-02-07
fix-test-and-hyperlinks

singhakshita:
for the PR #3588 for the issue #2519

1.initially  "https://www.wikidata.org/wiki/Property:P670 https://www.wikidata.org/wiki/Property:P669 are now mapped to https://schema.org/streetAddress via https://www.wikidata.org/wiki/Property:P2235 " was not treated as a link ,now each text component inside it is treated as text and each link is uniquely identified.
2. tests added to check if links inside of text snippets are treated as hrefs.

PR #3588, opened 2021-02-05
fixed href tag creation for URLs in cell text

singhakshita:
fixes #2519
- added <a>href tag creation for URLs in cell text.
- added tests.

PR #3597, opened 2021-02-08
 hyperlink issue  fixed and  tests added

### LINKED ISSUE #2519: Hyperlink is created for cell value even when invalid and text datatype is choosen

singhakshita:
fixes #2519  
due to some conflict issues and due to shown changes in huge number of files (which i never did) i have to send another pull request


fgiroud:
Is this ready for review ?
Your code should be linted, and the file should probably go under `grid/misc/cell_display.spec.js`, as this is not a transform I think

singhakshita:
yes this is up for review

fgiroud (approved):

fgiroud:
Good on the testing side, @thadguidry do you mind to review the changes ?

thadguidry:
@fgiroud I'd rather not.

main/tests/cypress/cypress/integration/project/grid/misc/proper-display.js
```diff
@@ -0,0 +1,23 @@
+describe(__filename, function () {
+  it('proper tag assingnment to each text snippet', function () {
```
kushthedude, line 2: typo

main/tests/cypress/cypress/integration/project/grid/misc/proper-display.js
```diff
@@ -0,0 +1,23 @@
...
+    const fixture = [
+      ['tests'],
+      ['2021-01-31https://www.google.com'],
+      [
+        'https://www.wikidata.org/wiki/Property:P670 https://www.wikidata.org/wiki/Property:P669 are now mapped to https://schema.org/streetAddress via https://www.wikidata.org/wiki/Property:P2235',
+      ],
+      ['vhjhjjj https://github.com/OpenRefine/OpenRefine/issues/2519'],
```
kushthedude, line 9: Can you please change this column name to something more appropriate?

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -96,17 +96,35 @@ DataTableCellUI.prototype._render = function() {
...
-      url.textContent = cell.v;
-      url.setAttribute('href', cell.v);
-      url.setAttribute('target', '_blank');
-      divContent.appendChild(url);
     } else {
-      var span = document.createElement('span');
-      span.textContent = cell.v;
-      divContent.appendChild(span);
-    }
+      var arr = cell.v.split(" ");
```
kushthedude, line 100: > The cell's value should be shown as text and not as a valid hyperlink.
(I think what is happening is that our current code just looks for a string beginning with http:// or https:// and doesn't care about validation or remaining characters such as whitespace after that, but should)
I think this is mentioned elsewhere in another related issue but couldn't find it.
Browsers now have native support for URL detection and we should take advantage of that.

Can you please see the desired behaviours, if I have the text as **https://www.wikidata.org/wiki/Property:P670 https://www.wikidata.org/wiki/Property:P669 are now mapped to https://schema.org/streetAddress via https://www.wikidata.org/wiki/Property:P2235** it should be displayed as a `span`, not a `hyperlink`? What the logic does over here is first splits the complete text by space and then checks the each as URL function and then appends them to a new array and other non-matching elements to another array bucket which in itself is quite complex logic. Also, the current logic can also be improved further rather than having a for loop to iterate over an array, the changes should be done in looksLikeURL function to not do the judgement on the basis of protocol prefix.
I would still rely on @wetneb and @thadguidry feedback on this behaviour.

main/tests/cypress/cypress/integration/project/grid/misc/proper-display.js
```diff
@@ -0,0 +1,23 @@
...
+    ];
+    cy.loadAndVisitProject(fixture);
+
+    cy.getCell(0, 'tests').contains('2021-01-31https://www.google.com');
+    cy.getCell(1, 'tests')
```
kushthedude, line 14: Can this test also check if there is `span` too as this is only checking the presence of `href`?

kushthedude (changes requested):

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
tfmorris: The behavior should match the ticket, but I think the current implementation is better than what's proposed on the ticket, so I'd suggest that we update the ticket instead of the code. I think the main thrust of the bug report is that multiple URLs or a URL followed by extra text shouldn't be hyperlinked as a single HREF, which this satisfies.

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -96,17 +96,35 @@ DataTableCellUI.prototype._render = function() {
...
+            url.setAttribute('target', '_blank');
+            divContent.appendChild(url);
+            if (i == arr.length-1) {
+             divContent.appendChild(url)
+            } else {
+            divContent.appendChild(url).appendChild(document.createTextNode('\u00A0'));
+            }
+          } else {
+          spanArr.push(arr[i]);
+          }
```
tfmorris, line 121: Indentation needs cleanup - 2 space indents should be used for each level. Any good IDE should be able to do this for you automatically.

tfmorris (changes requested):
Thanks for including the test. Things are getting closer, but still a couple of things to clean up. I would hold off on changing things functionally until we have consensus on the desired behavior, but please go ahead and clean up the other stuff that @kushthedude and I mentioned. 

Also, please edit the your original description to include the phrase "Fixes #2519" so that Github links the PR with the issue that it fixes (and will automatically close the issue when we merge the PR).

For reference, this is a follow-on #3591, which I reviewed before, and #3588, which got reverted.

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
singhakshita: "  https://www.wikidata.org/wiki/Property:P670 https://www.wikidata.org/wiki/Property:P669 are now mapped to https://schema.org/streetAddress via https://www.wikidata.org/wiki/Property:P2235 " 
i think the desired behaviour is ,it should be able to uniquely identify each text component and each url ,and it does so .
Screenshot 2021-02-10 at 7 40 42 AM

also i think if the text contain both url and normal tag , should it entirely be in <a> or <span> tag ,i guess  each component of it should be assigned a tag depending upon what it is ,whether a text or an url .

if i am getting anything wrong please let me know . @kushthedude 
Screenshot 2021-02-10 at 7 48 56 AM

it may seem like its adding extra spaces or manipulating the data but please look to this

main/tests/cypress/cypress/integration/project/grid/misc/proper-display.js
singhakshita: okay

main/tests/cypress/cypress/integration/project/grid/misc/proper-display.js
singhakshita: yes

main/tests/cypress/cypress/integration/project/grid/misc/proper-display.js
singhakshita: yes

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
singhakshita: yes working on it

thadguidry: @tfmorris @kushthedude @singhakshita - I have updated https://github.com/OpenRefine/OpenRefine/issues/2519#issue-595223760 to give some Additional Context into the use case.  Please take a look.

singhakshita: i have updated span test , typos fixed ,indentation fixed.  i guess we now can recognise all separate urls as mentioned by @thadguidry  , if any thing else is supposed to be done please let me know

wetneb:
@singhakshita you have included "fixes #2519" in the title of your pull request, but it should not be there. It should be in the description (the first comment in the discussion), in other words the body text (the one you type in a `<textarea>`, where you wrote `issue #2519`).

singhakshita:
@wetneb  sorry and thanks for your patience

wetneb:
No worries, this GitHub feature is not very well discoverable and confuses most of our new contributors at the moment.

tfmorris (approved):
LGTM. Thanks for the contribution!

PR #3600, opened 2021-02-08
fixes #3089 missing (i18n) internationalisation on about page 

### LINKED ISSUE #3089: Missing internationalization (i18n) on About page

singhakshita:
fixes internationationlization on about page


main/webapp/modules/core/langs/translation-en.json
```diff
@@ -1,6 +1,10 @@
 {
     "name": "English",
     "core-index/slogan": "A power tool for working with messy data",
+    "core-index/definition" : "OpenRefine is a power tool for working with messy data. Use it to improve data consistency,link it to data registries like Wikidata, augment it with data from other sources, transform it into different formats for other tools to consume, and contribute it to back to the original sources. OpenRefine is not a web service but a desktop app that runs on your own computer,so you can process sensitive data with privacy.",
+    "core-index/contributor" : "Contributors (in chronological order)",
+    "core-index/history" : "OpenRefine was originally developed as 'Freebase Gridworks' by <a href='$1'>$1</a>  Metaweb Technologies, Inc. Metaweb was acquired by <a href='$2'>$2</a> Google in July 2010 and they renamed the product Google Refine. In October, 2012, the product was renamed OpenRefine as it transitioned to  a community supported project.",
```
wetneb, line 6: `<a href='$1'>$1</a>` will not work, we want different values for the link URL and for the text, so that we match the previous rendering.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -1,6 +1,10 @@
 {
     "name": "English",
     "core-index/slogan": "A power tool for working with messy data",
+    "core-index/definition" : "OpenRefine is a power tool for working with messy data. Use it to improve data consistency,link it to data registries like Wikidata, augment it with data from other sources, transform it into different formats for other tools to consume, and contribute it to back to the original sources. OpenRefine is not a web service but a desktop app that runs on your own computer,so you can process sensitive data with privacy.",
```
wetneb, line 4: Perhaps the prefix shouldn't be `core-index` but rather `core-about` since this is not on the index page but the about page?

main/webapp/modules/core/about.html
```diff
@@ -62,30 +62,20 @@
...
-        <a href="http://www.google.com/">Google</a> in July 2010 and they renamed the product
-        Google Refine. In October, 2012, the product was renamed OpenRefine as it transitioned to
-        a community supported project.
+      <p id="history-openrefine">
       </p>
       
       <table cellspacing="0" cellpadding="0"><tr>
         <td width="300">
-          <h2>Contributors (in chronological order)</h2>
+          <h2 id="contributor"></h2>
```
wetneb, line 78: The `id` should be `contributors` I would say, not `contributor`.

wetneb (changes requested):
Thanks! Note that the pull request is still not linked to the issue: for this to work, you need to add "Fixes #3089" in the description (body text) of your pull request. You should be able to edit this yourself without having to open a new pull request.

singhakshita (commented):
i m made all the above mentioned changes

wetneb (approved):
Thank you very much!

PR #3601, opened 2021-02-08
Add UX Test for controls at the row level 

### LINKED ISSUE #3414: Add UX Test for controls at the row level
This tests intent to ensure the behaviors of the features at the row-level

Screen Shot 2020-12-16 at 13 12 56

### Proposed solution

Add test for controls that sit at the very left of each row in the grid.

### Flag

Flag, unflag one and several rows. Ensure that the flag icon is colored in yellow.

### Star

Star, unstar one and several rows. Ensure that the star icon is colored in yellow.

S-Harshit:
fixes #3414

### DIFF
--- a/main/tests/cypress/cypress/integration/project/grid/all-column/flag.spec.js
+++ b/main/tests/cypress/cypress/integration/project/grid/all-column/flag.spec.js
@@ -1,34 +1,24 @@
 describe(__filename, function () {
   it('Ensure flag is visible and toggle on/off', function () {
     cy.loadAndVisitProject('food.mini');
-    cy.get('.odd > :nth-child(2) > .data-table-flag-off')
-      .should('exist')
+    cy.get('.data-table tr:nth-child(1) td:nth-child(2) a')
+      .should('have.class', 'data-table-flag-off')
       .click();
-    cy.get('#notification > div:nth-child(2) > span:nth-child(1)').should(
-      'have.text',
-      'Flag row 1'
-    );
-    cy.get('.even > :nth-child(2) > .data-table-flag-off')
-      .should('exist')
+    cy.assertNotificationContainingText('Flag row 1');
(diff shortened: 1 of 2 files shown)


main/tests/cypress/cypress/integration/project/grid/all-column/flag.spec.js
```diff
@@ -0,0 +1,34 @@
+describe(__filename, function () {
+  it('Ensure flag is visible and toggle on/off', function () {
+    cy.loadAndVisitProject('food.mini');
+    cy.get('.odd > :nth-child(2) > .data-table-flag-off')
+      .should('exist')
+      .click();
+    cy.get('#notification > div:nth-child(2) > span:nth-child(1)').should(
```
fgiroud, line 7: @S-Harshit there's a command to check OR notifications
` cy.assertNotificationContainingText('Flag row 1');`

fgiroud: @S-Harshit the way you've done the test is technically working, but the selectors you've used could be improved.
Using odd/even can make one think you are really testing the behavior of odd/even rows, and that's not the case here

What you can do is to select rows by their index, columns a well.
example
``` 
cy.get('.data-table tr:nth-child(1) td:nth-child(2) a').should('have.class', 'data-table-flag-off').click();
cy.get('.data-table tr:nth-child(1) td:nth-child(2) a').should('have.class', 'data-table-flag-on');
```
This way the selector is meaningful, in human language that would says: Give the first row, then give me the second column, then click

S-Harshit: I have changed the selectors and the command to check notifications as suggested

fgiroud (approved):

PR #3602, opened 2021-02-09
CI: add eslint workflow in the CI

kushthedude:

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -133,7 +133,7 @@ jobs:
     - name: Lint with Prettier
       run: |
           cd ./main/tests/cypress
-          yarn check-lint
+          yarn lint
 
     - name: Test with Cypress on ${{ matrix.browser }}
       run: | 
--- a/main/tests/cypress/package.json
+++ b/main/tests/cypress/package.json
@@ -7,8 +7,8 @@
     "private": true,
     "scripts": {
         "test": "cypress run --browser electron --headless --quiet",
-        "lint": "prettier --write .",
-        "check-lint": "prettier --check ."
+        "fix-lint": "prettier --write . && eslint --fix .",
+        "lint": "prettier --check . && eslint ."
       },
     "dependencies": {
         "cypress": "6.4.0",


kushthedude:
The failure is due to the case that the Pull_Request_Target doesn't honour the latest changes in the PR regarding the workflow. As the check-lint is removed from the code but not from the master's CI config hence the failure

wetneb (approved):

fgiroud:
Are you sure your commands are working ?
Running `yarn lint` does not run eslint on my computer (macos, bash)
Also, when running `eslint .`, I've got the following error message: `Error: .eslintrc.json: Environment key "es2021" is unknown`

```
eslint --version
v6.8.0
```

Is that ok to use es2021 ? ? Is it an absolute necessity?
I just don't want all the developers to update their local tools if it is not really needed

kushthedude:
You are on outdated version of eslint we have used v7.8 in the
package.json, also for best practice I would recommend running eslint as a
global binary rather it should be like `yarn run eslint .`, which would
reference eslint from node_modules of that specific project as having
multiple projects with different version having a single binary may create
issues. Can you please run `yarn install` again and try running `yarn run
eslint .` or yarn lint?

fgiroud:
`yarn run eslint` -> working
` yarn run eslint .` -> working
`yarn run lint` -> not working,  it runs only the prettier
```
yarn run lint
yarn run v1.22.4
$ prettier --check . && eslint .
Checking formatting...
All matched files use Prettier code style!
✨  Done in 3.14s.
```

kushthedude:
Oh, eslint . will not give you any output unless and until there is an
error.

fgiroud:
Alright, got it @kushthedude 
A question: I've runned eslint on the master, shouldn't I see some warnings, because of my usage of `wait()` ?

kushthedude:
I ignored the three waits condition by putting a comment of those three
lines as exception in the earlier eslint pr. If there will be any more wait
added by PRs then eslint would definitely complain.

fgiroud (approved):

fgiroud:
All good for me
Thx for the work @kushthedude

PR #3603, opened 2021-02-10
feat: add tests for create project

### LINKED ISSUE #3428: Add UX tests for all methods to create a project
Creating a project is a major feature of OpenRefine, and should be extensively tested

Screen Shot 2020-12-16 at 14 26 17

### Proposed solution

One single test file per creation method is probably the best way to organize things.

#### Create from a local file

Add a test to ensure an alert is triggered when uploading an invalid file (jpg for example)

#### Web addresses

To test this, we would have first a PR that adds a file in the repository, then use it for testing upload from a web address.
Test with multiple URLs.

What is the expected behavior here? Dataset to be combined into one single project?

#### Clipboard

Test with copy-pasting
* CSV data
* JSON data

... and ensure this goes to the next step

#### Database

To keep simple here, only SQLite should be tested at first.
The SQLite file can be saved in the fixtures folder

#### Google Data

Is that possible to have a google spreadsheet stored in the repository? (No)

### Additional context

Generally speaking, those tests should only test the expected behaviors on the initial project creation page.
Different options for CSV and so on is a different test.

kushthedude:
fixes #3428


kushthedude:
@wetneb @fgiroud, regarding gsheet create project test, should we create a public viewable google sheet which we may export in here?

fgiroud:
@kushthedude I would rather not the tests to rely on something external, like google storage
Any way to have a spreadsheet locally stored ?

kushthedude:
Then how do we plan to manage test the google sheet importer?

fgiroud:
I don't know @kushthedude ;-), maybe that's a test we should postpone for now

wetneb:
There does not seem to be any easy way to mock out this external service! Sadly I don't have a solution to propose. Using a public spreadsheet or leaving the feature untested seem both reasonable to me.

fgiroud:
Is there a public/demo spreadsheet that we can use ?
IF you choose to use a spreadsheet on google, be sure it's no tied up to your gmail account, and that anyone, even without a gmail account, can access the spreadsheet

kushthedude:
I can create a spreadsheet and make the link accessible to anyone on
internet. Except the editing rights should be on with the maintainers.

kushthedude:
@fgiroud I need to increase the cypress timeout as openrefine needs time while downloading and uploading the files, therefore the wait conditions have been removed now as the timeout is increased

fgiroud:
@kushthedude I am not really ok with increasing the timeout globally, it think tests should runs with the default cypress timeout
Can you find a workaround with the `timeout` option that you can pass to any cypress command ?

kushthedude:
> @kushthedude I am not really ok with increasing the timeout globally, it think tests should runs with the default cypress timeout
> Can you find a workaround with the `timeout` option that you can pass to any cypress command ?

Done, used a local override to increase timeout/

fgiroud:
@kushthedude that's interesting what you've done with the cypress setting directly in the test, didn't knew this trick.
However, couldn't you just use `cy.get('.whatever', {timeout:1000000})` ?

kushthedude:
> https://docs.cypress.io/api/commands/get.html#Arguments

YEs this makes more sense rather than increasing the complete spec timeout, done the change

kushthedude:
@fgiroud , DB cant be tested due to https://github.com/OpenRefine/OpenRefine/issues/3605 , Please review the two importers and file changes

main/tests/cypress/cypress/integration/create-project/create_project.spec.js
```diff
@@ -26,89 +26,168 @@ describe(__filename, function () {
...
+      'Paste data from clipboard here:'
+    );
+    // add file
+    const csvFile = `Username; Identifier;First name;Last name
+    booker12;9012;Rachel;Booker
+    grey07;2070;Laura;Grey
+    johnson81;4081;Craig;Johnson
+    jenkins46;9346;Mary;Jenkins
+    smith79;5079;Jamie;Smith`;
+    cy.get('textarea').invoke('val', csvFile);
```
fgiroud, line 82: Why not using `cy.get('.selector').type(csvFile);` ?

main/tests/cypress/cypress/integration/create-project/create_project.spec.js
```diff
@@ -26,89 +26,168 @@ describe(__filename, function () {
...
     // add file
-    const csvFile = {
-      filePath: 'food.mini.csv',
-      mimeType: 'application/csv',
-    };
-    cy.get(
-      '.create-project-ui-source-selection-tab-body.selected input[type="file"]'
-    ).attachFile(csvFile);
+    const csvURL =
+      'https://people.sc.fsu.edu/~jburkardt/data/csv/addresses.csv';
```
fgiroud, line 141: @kushthedude what is this domain ?
I don't think it's a good idea to rely on external domains for ou test suite
Is it possible to find a workaround using files stored directly on this Git repository ? Or something owned by OpenRefine ?
Alternative: Possible to use a local url ? Using cypress mock capabilities  ?

main/tests/cypress/cypress/fixtures/param.tsv
```diff
@@ -0,0 +1,2 @@
+Some parameter	Other parameter	Last parameter
+CONST	123456	12.45
```
fgiroud, line 2: Can we have a better name for this file ? something more meaningful than `param.tsv` ?

main/tests/cypress/cypress/integration/create-project/create_project.spec.js
kushthedude: Invoke is more like copy-paste rather than the type which is individually typing each character. That's why I used invoke to have tested more resembling to copy-paste in clipboard.

main/tests/cypress/cypress/integration/create-project/create_project.spec.js
kushthedude: Let's not make it a local URL, since we need to test openrefine fetch and upload API. I agree to put csv in the openrefine but the csv which is attached is public license, we can copy its contents or either we can provide here the link to one of our cypress fixture.

main/tests/cypress/cypress/fixtures/param.tsv
kushthedude: done!

main/tests/cypress/cypress/integration/create-project/create_project.spec.js
fgiroud: For the wikimedia extension, I am testing like that
```
 cy.get('.add-wikibase-dialog input[bind="manifestURLInput"]').type(
      'https://raw.githubusercontent.com/OpenRefine/wikibase-manifests/master/openrefine-wikibase-test-manifest.json'
    );
```
maybe you can use github

main/tests/cypress/cypress/integration/create-project/create_project.spec.js
fgiroud: Agree, good with it

main/tests/cypress/cypress/integration/create-project/create_project.spec.js
tfmorris: Yes, there are lots of URLs that we control. Github, openrefine.org, etc. Let's pick one that we control.

kushthedude: done!

kushthedude:
@fgiroud @tfmorris Done the requested changes, please take a look

fgiroud (approved):

fgiroud (approved):

fgiroud (approved):

fgiroud:
Looks good to me.
@tfmorris can we merge this?

kushthedude:
Ping @tfmorris, if this can be merged or is it LGTM from your side?

tfmorris:
Sorry for the delay. I'll be able to review in a few hours.

tfmorris (approved):
I'm not sure I know enough about Cypress to be on the critical path blocking merging this, but looks good to me except for one minor cosmetic issue.

I had 3 test failures locally (timeouts), but the tests pass in CI, so I'm going to assume that that's just a problem with my local environment.

main/tests/cypress/cypress/fixtures/shop.mini.tsv
```diff
@@ -0,0 +1,6 @@
+storm	instant	amount	choice	pool	married
+keep	fifty	memory	behavior	courage	troops
+hello	mass	industry	cannot	complete	practical
+add	shop	surrounded	time	highway	myself
+first	task	fine	steep	should	great
+has	shot	bet	became	damage	share
```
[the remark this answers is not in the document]
wetneb: Ah, sorry I saw so much green that I merged a bit quickly without leaving the opportunity to address this.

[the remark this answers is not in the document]
tfmorris: No worries. It was a "nice to have" which is why I didn't make it a "Changes requested"

PR #3606, opened 2021-02-10
feat: add tests for rows-records mode

### LINKED ISSUE #3416: Add UX Test for the grid header / show as records-rows
Add test for the feature that shows records as rows, as records

Screen Shot 2020-12-16 at 13 21 58

### Proposed solution

Add a test to show as rows, records and
* load a project of CSV data, show as row/records, ensure nothing changes
* load JSON data, and test the rendering as rows, as records

Exact expected behavior of this feature to be reviewed

kushthedude:

Fixes #3416


fgiroud:
Generally speaking @kushthedude, I would like to keep fixtures "as files" for datasets used in multiple tests
In your case, maybe that's better to keep sweets.json right inside the test folder

I noticed that when switching to records, there's a red background on the grid header, do you think that's a behavior that should be tested?
Overall, is there more to be tested on the record/row behavior?

Unrelated, but do you understand why I am not able to link issue #3416 to this PR?

kushthedude:
> I noticed that when switching to records, there's a red background on the grid header, do you think that's a behavior that should be tested?

That test isn't exclusive to rows-records rather it's a JSON importer feature to show the parent-child relation which we should not include in row-records imo.

fgiroud:
> > I noticed that when switching to records, there's a red background on the grid header, do you think that's a behavior that should be tested?
> 
> That test isn't exclusive to rows-records rather it's a JSON importer feature to show the parent-child relation which we should not include in row-records imo.
> 
> > Unrelated, but do you understand why I am not able to link issue #3416 to this PR?
> 
> You have already linked the issue 😄 , I am unsure why there wasn't any event shown in the PR but you can see it in #3416

Okay, I guess I need some sleep :-)
Alright for the red background

kushthedude:
> I would like to keep fixtures "as files" for datasets used in multiple tests
In your case, maybe that's better to keep sweets.json right inside the test folder

About this, I guess it's good to have one JSON fixture since my next test involves this more. That's why I did put the sweets.json in fixtures. Also, I have added a data-cy element in this PR because I spend hours grabbing the elements of JSON-importer-ui but I had no luck. Therefore as a last resort I had to add data-cy tag in the ui.

fgiroud:
@kushthedude unless the fixture is used by many test scenarios, I would rather have it close t other test for now
Ok to move it to the global fixtures once there is a real need to share sweet.json across multiple tests
Just checking, but would it be possible to use the API to create a "record" project (json), because right now all the steps to create the project are displayed in the runner, which make it difficult to understand
Also, using the API would be consistent with the way we create projects from CSV data
Possible ?

kushthedude:
> Also, using the API would be consistent with the way we create projects from CSV data
Possible ?

You cant import JSON projects through the only API as it involves a major intermediary step of choosing the opening and closing nodes which is only possible through importer-UI from what I tried and experimented with.

kushthedude:
@fgiroud Please review and merge, removed the global fixture and defined a const in the spec file as is

fgiroud (approved):

fgiroud:
Looks like this test is failing from time to time, detached element from the dom/

- https://dashboard.cypress.io/projects/s5du3k/runs/713/overview
- https://dashboard.cypress.io/projects/s5du3k/runs/702/overview

PR #3622, opened 2021-02-12
common transform toBoolean added

### LINKED ISSUE #3518: Common transforms for columns should contain `to boolean`
There are a variety of options to cast a column into text, number or date. But there is no common cast option to change a column to a boolean type.
### Proposed solution

Common transforms for columns should contain `to boolean`
### Alternatives considered

None
### Additional context

Why wasn't this implemented, or are there concerns on this option?

Screenshot 2021-01-26 at 12 07 51 AM

singhakshita:
fixes #3518  , changes all the cell data that doesn't match with "true" as false . similar to what the "edit" shown on each cell does when "to boolean" option is chosen.
Screenshot 2021-02-12 at 1 41 26 PM
Screenshot 2021-02-12 at 1 41 41 PM

### DIFF
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -645,6 +645,7 @@
     "core-views/to-number": "To number",
     "core-views/to-date": "To date",
     "core-views/to-text": "To text",
+    "core-views/to-boolean": "To boolean",
     "core-views/blank-out": "To null",
     "core-views/blank-out-empty": "To empty string",
     "core-views/replace": "Replace",
--- a/main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js
+++ b/main/webapp/modules/core/scripts/views/data-table/menu-edit-cells.js
@@ -454,7 +454,7 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
         {
           id: "core/to-boolean",
           label: $.i18n('core-views/to-boolean'),
-          click: function() { doTextTransform("if(value != 'true' , 'false' , 'true')", "keep-original", false, ""); }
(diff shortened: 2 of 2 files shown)


wetneb (changes requested):
I don't think `if(value != 'true' , 'false' , 'true')` is the expression we want here. It should return a boolean, not a string. I would expect something such as `value == 'true'` (without the surrounding `if` statement). Perhaps with an additional tweak to handle null values, depending on what we want to do with them.

singhakshita:
@wetneb  yeah of course it had to be boolean data ,sorry my bad ! updated

wetneb (approved):
That works for me, but I would like to leave the opportunity to @ostephens to comment about how nulls should be handled here: should this convert nulls to `false` or keep them as null values?

singhakshita:
Screenshot 2021-03-01 at 4 25 30 PM
Screenshot 2021-03-01 at 4 24 54 PM

i have added a submenu to choose the language ,for now javascript sql and python , (now the data is being transformed to true or false based truthy values as documented in that particular language.
please let me know if it works well

wetneb (commented):
I like it! I think this adds more value than the previous proposal. A few things:
* I would expect the menu labels to be translatable (so, using `$.i18n`)
* In English we should keep consistent casing: use "Python **t**ruthy **v**alues", not "Python Truthy Values".

It would still be nice to have more comments from others about how useful these transforms would be (@thadguidry perhaps?), but since this has been open for a while I think we can also merge this if no one else chimes in.

wetneb:
Ah sorry I didn't see the discussion on the linked issue. It looks like people have legitimate concerns about adding those.

wetneb:
@singhakshita unfortunately I think we need to close this PR - sorry for the work you invested into it! We should have checked the discussion on the related issue more carefully.

PR #3627, opened 2021-02-13
Fix Google OAuth redirect_uri behind an HTTPs reverse proxy

matiasgarciaisaia:
If OpenRefine runs behind an HTTPs reverse proxy, Jetty 6 doesn't set the correct client's schema on the HttpServletRequest object, causing the generated `redirect_uri` for OAuth to always be plain HTTP - which Google forbids for external OAuth applications.

This commits hardcodes the check, effectively allowing the OAuth flow to work.

Jetty 7 and newer support a correct way to do this - but Jetty 6 doesn't.


matiasgarciaisaia:
I do know that running OpenRefine on the cloud is not a first-class use case.

I'm currently setting up an instance behind a proxy that implements access control and HTTPs termination for a user with limited hardware & network resources. The only thing forbidding OpenRefine to work so far has been this issue.

tfmorris:
> running OpenRefine on the cloud is not a first-class use case.

A more accurate way to phrase this would be "running OpenRefine any place but the user's local computer is completely unsupported."

wetneb:
Related to the need to migrate to a newer version of Jetty, being discussed on the mailing list:
https://groups.google.com/g/openrefine-dev/c/hXwX_SqjXbg/m/E2LAWbJlBQAJ

wetneb:
Thanks a lot again @matiasgarciaisaia for this contribution and sorry it still hasn't been merged. In the meantime we have upgraded to Jetty 9… Is this PR still needed?

matiasgarciaisaia:
[From my comment](https://github.com/OpenRefine/OpenRefine/pull/3627/files#diff-21acd55474e1111cf1f1c944d7fd52bc35c7217d01138a42d4f20631284608d1R94-R97) back in February, this commit as-is won't work, but we have to enable _something_ to address forwarded requests anyway.

I guess with [the new `ServerConnector`](https://github.com/OpenRefine/OpenRefine/blob/1654cd31b1b967bc23a5df921b65b2c24dce8395/server/src/com/google/refine/Refine.java#L177-L180) we should use [`HttpConfiguration.addCustomizer`](https://www.javadoc.io/static/org.eclipse.jetty/jetty-server/9.4.41.v20210516/org/eclipse/jetty/server/HttpConfiguration.html#addCustomizer(org.eclipse.jetty.server.HttpConfiguration.Customizer)) to add a [`ForwardedRequestCustomizer`](https://www.javadoc.io/doc/org.eclipse.jetty/jetty-server/9.4.41.v20210516/org/eclipse/jetty/server/ForwardedRequestCustomizer.html) - but I know nothing about all of this.

I'll try to test it whenever I can, but if someone else wants to try it - be my guest :)

PR #3634, opened 2021-02-14
Fix database extension tests configuration.

### LINKED ISSUE #3633: Database extension tests are not run
The backend tests for the database extension (support for SQL import/export) are not currently run, when running tests locally. In the CI, they are run (because we use a specific configuration file in that context).

It is good that tests requiring a database engine are not run by default (we do not want OpenRefine developers to have to run multiple database engines just to run the tests, especially if they are working on an independent aspect of the tool). But we should at least run the tests which do not rely on any such external database.

### To Reproduce
Run `./refine test` and observe that 0 tests are run for the database extension.

This is due to an incorrect configuration file for the test suite, which I seem to be responsible for. Oops.

wetneb:
Closes #3633.

### DIFF
--- a/extensions/database/tests/conf/tests.xml
+++ b/extensions/database/tests/conf/tests.xml
@@ -44,25 +44,9 @@
                 <exclude name="requiresMariaDB" />
                 <exclude name="requiresSQLite" />
             </run>
-        </groups> <!--
-        <classes>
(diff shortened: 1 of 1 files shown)


wetneb:
@kushthedude I am puzzled by the linting failure, which seems unrelated. Any idea?

tfmorris (approved):
I'm assuming that the test failure is going to get sorted out separately

PR #3626, opened 2021-02-12
Translations update from Weblate

weblate:
Translations update from [Weblate](https://hosted.weblate.org/projects/openrefine/translations/) for OpenRefine/Translations.

Current translation status:

Weblate translation status

PR #3635, opened 2021-02-14
fix: accidental non-linted files in repo

kushthedude:

refer https://github.com/OpenRefine/OpenRefine/pull/3626#issuecomment-778817086

- The issue was caused due to the `pull_request_target` as well as the branch which was being merged was not in sync with upstream/master


wetneb (approved):
Thanks! Yes our `pull_request_target` workflow is a real problem. But I think what confused me more is that the build on master passes - because linting is only enforced in pull requests, not in master. Perhaps that's also something worth fixing?

kushthedude:
> Thanks! Yes our `pull_request_target` workflow is a real problem. But I think what confused me more is that the build on master passes - because linting is only enforced in pull requests, not in master. Perhaps that's also something worth fixing?

Yes I will plug that additional step into snapshot_release workflow.

wetneb (approved):

PR #3647, opened 2021-02-17
Add link to GREL reference in Help tab of expression editor

### LINKED ISSUE #3580: Add a link to the GREL reference in the expressions editor window
Currently the expressions editor window supplies an in-tool Help tab, but it would also be nice to refer people to the GREL functions reference: https://docs.openrefine.org/manual/grelfunctions

helpdisplay

It seems likely that the manual will usually be more up-to-date than the help text inside whatever version of OpenRefine people are using. You may wish to provide the link to the appropriate version fork of the docs as you go forward, or supply a simple link and warn users to select their respective version from the dropdown. 

### Proposed solution
This link could go inside the Help tab itself, the first line at the very top (presuming #3579 is possible and completed), or it can appear as a button elsewhere inside the expressions editor window (perhaps to the right of the tabs). Maybe a button could appear to the right of the language dropdown when GREL is selected. Wherever it makes sense to take up some space.

singhakshita:
fixes #3580  ,added grel reference
Screenshot 2021-02-17 at 4 56 19 PM

the link if added on the left side ,it would seem related to "variable" text  ,
 so i added it on top so that it seems global .

- also when we change the language(grel,python,clojure) the text content of help tab doesn't change , so just removing the grel reference link ,didnt seem worth .(isn't the  content of "help" tab suppose to change for change of language ,just want to know its not a bug)


ostephens:
I think we shouldn't use all caps for this. Also suggest the text should be slightly more explanatory - e.g.:

"View the <link>GREL reference in the OpenRefine User Manual</link> for more detail."

@allanaaa what do you think?

allanaaa:
That works for me, Owen!

singhakshita:
please let me know if any changes are required
Screenshot 2021-02-17 at 11 22 08 PM

ostephens:
This looks great @singhakshita but I think the text needs to go in as a piece of translatable text - you'll need to add a key to at least the translation-en.json file (probably under `core-dialogs`) and then use code like `$.i18n('core-dialogs/newkeyname')` to bring it into the display

main/webapp/modules/core/scripts/dialogs/expression-preview-dialog.js
```diff
@@ -187,7 +187,11 @@ ExpressionPreviewDialog.Widget.prototype._renderHelpTab = function() {
 
 ExpressionPreviewDialog.Widget.prototype._renderHelp = function(data) {
     var elmt = this._elmts.expressionPreviewHelpTabBody.empty().width(this._tabContentWidth);
-    
+    $("<a />", {
+      href: "https://docs.openrefine.org/manual/grelfunctions",
+      text: "View the GREL reference in the OpenRefine User Manual for further details",
```
ostephens, line 192: I think this text should come from a translation file to support multiple languages

ostephens (changes requested):

main/webapp/modules/core/scripts/dialogs/expression-preview-dialog.js
singhakshita: yeah done ,this just slipped off my mind

ostephens (approved):

singhakshita:
@wetneb  am i supposed to make  any change in it ,to resolve conflicting issues?

wetneb:
It would be great if you could do that actually!

singhakshita:
@wetneb  please let me know if any other change is required

wetneb (approved):

Issue #3579, opened 2021-02-04
Enable hyperlinks in in-tool Help field


allanaaa:
The expressions editor window includes a Help tab that has some plain-text information about each variable and function.

![display3](https://user-images.githubusercontent.com/28012148/106948555-d6b4e000-66f9-11eb-8b06-bf4c5b181b0f.PNG)

I am editing this text right now to match the user manual. I'd like the fields to render with links (they currently just have full URLS that users can copy-paste). 

If you look into the code, you'll find that the function files themselves support the Help entries, for example for `cross()`:

>     @Override
>     public String getDescription() {
>         return "Looks up the given value in the target column of the target project, returns an array of matched rows. Two values match if and only if they have the same string representation. " +
>                 "The first argument will be interpreted as cell.value if set to cell. " +
>                 "The second argument will be interpreted as the current project name if omitted or set to \"\". " +
>                 "The third argument will be interpreted as the index (starts from 0) column if omitted or set to \"\"";
>     }
> 
>     @Override
>     public String getParams() {
>         return "cell or value, string projectName (optional), string columnName (optional)";
>     }
> 
>     @Override
>     public String getReturns() {
>         return "array";
>     }

### Proposed solution
Render these text entries with hyperlinks, somehow! Whether it's simple HTML or markdown or what-have-you, I can work with it.

sarahelshabrawy:
I would like to help with this issue

wetneb:
Thanks! Note that there will be a few design decisions to do on this issue, so perhaps it's worth discussing your approach here before opening a pull request, to make sure we agree on the direction.

sarahelshabrawy:
So my approach is to delete the String URL reference from all translation .json files and add an anchor tag in expression-preview-dialog class to stand for all languages, Another change is to replace .text() with .html() in order to render the html content of the String of the table cell
Below is a screenshot of the link rendered
![image](https://user-images.githubusercontent.com/58959185/107526072-2266fe00-6bc0-11eb-9a4f-4bb16a6f6296.png)
Is that okay ?
As for the other link I have a question about it.

wetneb:
I think allowing HTML in translations is a sensible way to go. Since we review the translation updates as pull requests we can consider these strings as trusted. Allowing HTML markup could be an opportunity to style code samples more nicely (such as `row.index` instead of 'row.index').

sarahelshabrawy:
I agree, I will go with that approach then and create a pull request.

[...]

PR #3648, opened 2021-02-17
Enable hyperlinks in in-tool Help field #3579

sarahelshabrawy:
I allowed rendering HTML in the expression-preview-dialog.js file and added an anchor tag for all the URLs
I also changed any tag symbol '<' to '<' in the translations files in order not to be rendered as an HTML tag and appear as '<' as it was before


wetneb:
@sarahelshabrawy thank you very much for that! It looks very good to me. Unfortunately this conflicts with another change, could you solve the merge conflicts? Thank you!

sarahelshabrawy:
@wetneb Done

wetneb (approved):

Issue #3426, opened 2020-12-16
Add UX Test for column / Transpose


fgiroud:
Testing all the possibilities for "Transpose" is complex because there are many combinations and options possible.
Community feedback is welcome to specify some tests to explore all possible combinations of options.

It's also possible that such intensive testing is a backend concern only and should not be implemented

This issue only specifies an MVP test for transpose.

![Screen Shot 2020-12-16 at 14 08 19](https://user-images.githubusercontent.com/6267288/102352582-361f2b00-3fa8-11eb-93b9-49ae13fcaf29.png)

![Screen Shot 2020-12-16 at 14 08 12](https://user-images.githubusercontent.com/6267288/102352590-38818500-3fa8-11eb-82c1-fc2408e1596f.png)

![Screen Shot 2020-12-16 at 14 08 03](https://user-images.githubusercontent.com/6267288/102352593-3a4b4880-3fa8-11eb-9fd3-b503880c04c8.png)

### Proposed solution

Using food.mini.csv, try the general behavior, with
* Key column: A
* Value Column: B

Ensure all the cells in the grid match the expected outcome (Example below)

![Screen Shot 2020-12-16 at 10 16 17](https://user-images.githubusercontent.com/6267288/102352711-6666c980-3fa8-11eb-91af-ee930f724ed0.png)

PR #3650, opened 2021-02-18
test: Added tests for transpose functionalities, #3426

fgiroud:
Added tests for the transpose functionalities of OR
Basically, it follows [tutorials found on the doc](https://docs.openrefine.org/manual/transposing/)

Does this needs to be improved ?
Any missing or critical points that are not covered by the present tests ?

@kushthedude please have a look at the new command available, `cy.assertGridEquals`, it let us test the whole grid in one single assertion


kushthedude (approved):
Looks good to me!

PR #3680, opened 2021-02-28
Conditionalize steps that require secrets

### LINKED ISSUE #3677: Cypress builds fail in forked repo
I'm getting the following errors from the "Test with Cypress on Chrome" step when Github Actions attempts to run in my fork of OpenRefine:

```
Run echo REFINE_MIN_MEMORY=1400M >> ./refine.ini
You have 6954M of free memory.
Your current configuration is set to use 4096M of memory.
https://github.com/OpenRefine/OpenRefine/wiki/FAQ-Allocate-More-Memory
fatal: No names found, cannot describe anything.
ERROR: Found a Cypress project id but no record key
Type './refine -h' for usage information.
```

@fgiroud @kushthedude Do you know what's going on here and how I can fix it?

tfmorris:
Fixes #3677

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -75,7 +75,7 @@ jobs:
       run: mvn jacoco:prepare-agent test
 
     - name: Submit test coverage to Coveralls
-      if: ${{ secrets.COVERALLS_TOKEN }} != ""
+      if: ${{ secrets.COVERALLS_TOKEN }}
       run: |
         mvn prepare-package -DskipTests=true
         mvn jacoco:report coveralls:report -DrepoToken=${{ secrets.COVERALLS_TOKEN }} -DpullRequest=${{ github.event.number }}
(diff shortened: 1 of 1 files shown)


kushthedude (commented):
This PR will amend that any of the PR which are made against openrefine master will not get a recorded run in the cypress dashboard. We can't conditionalism the secrets as they are essential assets to record and upload the cypress recorded artefacts to the dashboard. 

To run the action on the fork, we can change the script to not upload the videos to the dashboard if the `RECORD_KEY` is not passed, which seems like the best workaround according to our present conditions.

wetneb:
I think this is still a step in the right direction though. I think we can aim for a two-stage process:
- a workflow on the `pull_request` event which does not use any secrets (so, no coverage tracking or report on the cypress dashboard)
- a workflow on the `pull_request_target` event which only runs if a certain label is present on the PR, with access to the secrets. So when we see a legit PR for which we want to see the coverage analysis or the Cypress dashboard, we can just add the label to the PR manually (and come back some time after that).

wetneb:
This is no longer required, we now use the `pull_request` hook only, as the credentials we need for those workflows are not sensitive.

PR #3689, opened 2021-03-02
Quote SQL column names

### LINKED ISSUE #1940: SQL exporter doesn't create compliant column names
**Describe the bug**
The SQL exporter will not create SQL-compliant column names. If a column has spaces, they are replaced with "-" by default, and replaced with nothing if the "Trim Column Names" checkbox is unchecked. Hyphens are not an acceptable character in a SQL column name. 

**To Reproduce**
1. Open a dataset that has column names with spaces in them. 
2. Click Export>SQL Exporter. 
3. Click "Download" tab. 
4. Click Preview
5. Attempt to create a table in an SQL database with the SQL exported by OpenRefine. It will not create, and SQL will print an error like:
````
ERROR:  syntax error at or near "-"
LINE 11: Line1-Street TEXT NULL,
              ^
````

**Current Results**

**Expected behavior**
The SQL Exporter should, by default, convert column names to an SQL-compliant name, replacing spaces and other illegal characters with "_" (underscore). Column names can also not begin with numbers. They cannot have single quotes or punctuation. 

**Screenshots**
image
image
image

 - OS: Mac 10.14.2
 - Browser Version: Safari 12.0.2
 - JRE or JDK Version: 1.8.0_121

 - Version 3.1 [b90e413]

### PR DESCRIPTION
**Pull request by NikitaEmberi**
Fixes #1940 and #2716 

- 
- 
-

### DIFF
--- a/main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
+++ b/main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
@@ -45,6 +45,7 @@ public class SqlCreateBuilder {
     private String table;
     private List<String> columns;
     private JsonNode options;
+    public static String dbType="";
 
 
     public SqlCreateBuilder(String table, List<String> columns, JsonNode sqlOptions) {
@@ -82,13 +83,21 @@ public class SqlCreateBuilder {
       
                 
                 if (name != null) {
-                    if(trimColNames) {
+                	if(trimColNames) {
                         String trimmedCol = name.replaceAll("\\s", "");
-                        createSB.append( trimmedCol + " ");
-                    }else{
-                        createSB.append(name + " ");
-                    }
-                   
+                        if(dbType.equals("PostgreSQL")) {
+                    	       createSB.append("\"" +trimmedCol.replaceAll("'","''")+ "\" ");
+                    	   }else {
+                    		   createSB.append("`"+trimmedCol+"`" + " ");
+                    	   }
+                     }else{
+                     	if(dbType.equals("PostgreSQL")) {
+                     	    createSB.append("\"" +name.replaceAll("'","''")+ "\" ");
+                     	}else {
+                     		createSB.append("`"+name+"` ");	
+                     	}
+                     }
+                	
                     if (type.equals(SqlData.SQL_TYPE_VARCHAR)) {
                         if (size.isEmpty()) {
                             size = "255";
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java [hunk 1]
```diff
@@ -82,13 +83,21 @@ public String getCreateSQL() {
       
                 
                 if (name != null) {
-                    if(trimColNames) {
+                	if(trimColNames) {
```
tfmorris, line 86: There seem to be whitespace changes mixed in with functional changes. Please don't reformat existing code and use 4 space indents for new code (to match the existing code in the module).

main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
```diff
@@ -82,13 +83,21 @@ public String getCreateSQL() {
...
                         String trimmedCol = name.replaceAll("\\s", "");
-                        createSB.append( trimmedCol + " ");
-                    }else{
-                        createSB.append(name + " ");
-                    }
-                   
+                        if(dbType.equals("PostgreSQL")) {
+                    	       createSB.append("\"" +trimmedCol.replaceAll("'","''")+ "\" ");
+                    	   }else {
+                    		   createSB.append("`"+trimmedCol+"`" + " ");
```
tfmorris, line 91: There are several copies of what's effectively the same code. Let's create a convenience method like `String quoteName(String name, quoteChar)` to encapsulate this.

I'm not familiar with SQL's quoting rules. Do single quotes have to be escaped (ie doubled) even though they are enclosed in double quotes?

main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
```diff
@@ -82,13 +83,21 @@ public String getCreateSQL() {
...
+                    	   }
+                     }else{
+                     	if(dbType.equals("PostgreSQL")) {
+                     	    createSB.append("\"" +name.replaceAll("'","''")+ "\" ");
+                     	}else {
+                     		createSB.append("`"+name+"` ");	
+                     	}
+                     }
```
tfmorris, line 99: Indentation looks off here. Please make sure all indents are 4 spaces.

tfmorris (changes requested):

main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
NikitaEmberi: Yes sir single quotes have to be escaped even though they are enclosed in PostgreSQL.

main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
NikitaEmberi: Yes sir I will work on indentation and try to encapsulate repeating code in a method.

NikitaEmberi: I have declared function like this:
image
does this look fine now?
image

PR #3702, opened 2021-03-07
#3186 : Wikibase login form shows Wikibase instance logo

### LINKED ISSUE #3186: Display the logo of each Wikibase instance in the login form and Wikibase selection dialog
The Wikibase login form now shows the Wikibase logo on the left hand side:

image

It would be nice if we could have the logo of the selected Wikibase instance instead. This would be immediately recognizable to the user, who would therefore be less likely to enter credentials for another instance.

These logos could also be shown in the UI where we select which instance to work against.

image

### Proposed solution

The URL of the logo used by the Wikibase instance can be fetched via the API:
https://daap.bannerrepeater.org/w/api.php?action=query&meta=siteinfo&siprop=general%7Cnamespaces%7Cnamespacealiases%7Cstatistics
Which gives a logo URL: https://daap.bannerrepeater.org/w/resources/assets/wiki.png

This would give something like this:
image

### Alternatives considered
The logo could be added to the manifest, but if we can fetch it from the MediaWiki instance directly I think it simplifies everyone's life if we do that.

### Additional context

We need to think about whether fetching these images from the frontend directly from the remote site could cause security or privacy issues. It seems that it is not more dangerous than visiting the Wikibase instance itself (since that loads the logo too), so I think is is acceptable?

sjathin:
Fixes #3186 

- Wikibase login form should show the selected Wikibase instance logo
- Logos shown in the UI during selection of instance to work against

Screenshot 2021-03-07 at 23 40 13

Screenshot 2021-03-08 at 17 05 57


sjathin:
Some suggestions:
1. In the UI to select the instance, is the name required as the logo is displayed?
2. If the name is not displayed, it would be nice to see the logos in a grid-like structure.

Also, I would like to test this using a project where there can be multiple wikibase instances. I am not sure as to how I can test this thoroughly. Any suggestions for testing here?

wetneb:
Hi @sjathin,

Thank you so much for working on this! I would say:
- let's keep the name alongside the logo, because some Wikibase instances might not have set up their own distinctive logo but use a stock one, so it could be difficult to distinguish them from others because of that.
- In the list of instances, I would aim to keep the logos small (with the CSS properties `max-height` and `max-width` set). The Wikidata logo looks very big in your screenshot above, intuitively it should be about 3 times as small (roughly).

sjathin:
**Testing**:

**Case 1**: When API does not give an error
Screenshot 2021-03-12 at 22 17 20

**Case 2**: When API gives an error (Tested the functionality without internet connectivity)
Screenshot 2021-03-12 at 22 17 59
Screenshot 2021-03-12 at 22 17 43

wetneb (commented):
I wonder if it would make sense to cache the logo URL so that it is not re-fetched from the Wikibase instance every time we need to display it. For instance, if I open the dialog to select the Wikibase instance multiple times, this will trigger one HTTP request per Wikibase instance every time I open the dialog.

I think we could at least cache it client-side (in Javascript) so that it is fetched at most once per session. I don't think it is worth the trouble of persisting it server-side (in Java) since it is more complicated and is likely to go out of sync with the Wikibase instance.

wetneb:
@sjathin what do you think about my comment above? It should really be pretty easy to implement, I don't have strong feelings about how this caching should look like in JS. Let me know if I can help in any way :)

sjathin:
@wetneb Yes, I agree, caching the logo URL will be much more efficient. Working on this.

sjathin:
**Testing**:

**Case 1:**

sjathin:
**Testing:**

**Case 1:** When API does not give an error, only the first time the API is called.

Screenshot 2021-04-17 at 12 56 42

Now the logo URL is cached at the client side. API is not called again.
Screenshot 2021-04-17 at 12 57 10


**Case 2:** When API gives an error (Tested the functionality without internet connectivity)
Screenshot 2021-04-17 at 13 00 20

@wetneb  Please let me know if you recommend any changes, I am all ears :)

extensions/wikidata/module/scripts/dialogs/manage-account-dialog.js
```diff
@@ -79,6 +79,11 @@ ManageAccountDialog.displayPasswordLogin = function (onSuccess) {
   const frame = $(DOM.loadHTML("wikidata", "scripts/dialogs/password-login-dialog.html"));
   const elmts = DOM.bind(frame);
   ManageAccountDialog.initCommon(elmts);
+  WikibaseManager.getSelectedWikibaseLogoURL(function(data) {
+	elmts.wikibaseLogoImg.attr("src", data);
+  });
+  elmts.wikibaseMainPage.attr("href", WikibaseManager.getSelectedWikibaseMainPage());
+  elmts.wikibaseLogoImg.attr("alt", WikibaseManager.getSelectedWikibaseName() + " Logo");
```
wetneb, line 86: Minor detail: we could translate the " logo" string since it is language-specific. To make it possible for translators to reorder the words, we can use a template such as "$1 logo" as translation string.

extensions/wikidata/module/scripts/wikibase-manager.js
```diff
@@ -188,3 +188,61 @@ WikibaseManager.fetchManifestFromURL = function (manifestURL, onSuccess, onError
...
+  };
+  let wikibase = (wikibaseName) ? WikibaseManager.wikibases[wikibaseName] : WikibaseManager.getSelectedWikibase();
+  const url = wikibase.mediawiki.api;
+  return $.ajax({
+    url: url,
+    data: params,
+    dataType: "jsonp",
+	cache: true,
+    timeout: 1000,
+    cache: true,
```
wetneb, line 220: LGTM warns us of a duplicate property `cache: true` here, it would make sense to remove it.

wetneb (approved):
Thanks a lot, it works very well. I just noticed two minor things that could be worth fixing.

sjathin:
@wetneb Not exactly sure why the checks failed. Any suggestion regarding how I could fix this?

wetneb (approved):
Sorry about this spurious failure, it is unrelated to your changes. Thank you again very much for this very nice contribution!

PR #3710, opened 2021-03-10
Property suggestion for data extension should try CORS first.  

### LINKED ISSUE #3672: Data extension service not using CORS
### To Reproduce
1. Implement a data reconciliation service with the data extension service specified in the service manifest.
2. Open the browser web console.
3. Finally, try to "Add columns from reconciled values"

### Current Results
Currently you get back a blank form with no properties for selection.  I think given the bug from this issue it probably should have shown an error to the user.  But when you look in the web console, you can see that the request was using JSONP.

``` sh
GET http://127.0.0.1:5000/properties?type=&callback=jQuery1111022504064285953862_1613364041041&_=1613364041042
```

### Expected Behavior

Similar to reconciliation, I would have expected it to try to use CORS first and fallback to JSONP.

### Versions
 - Operating System: 
 - Browser Version: 
 - JRE or JDK Version: 
 - OpenRefine: 

### Additional context

I can provide a sample service demonstrating this bug if you like.  Please let me know.  I think the code making this request is [here](https://github.com/OpenRefine/OpenRefine/blob/6c47c81e5913d64ea58952454a1d67750fff2ff0/main/webapp/modules/core/scripts/dialogs/extend-data-preview-dialog.js#L98).  It looks like recon [calls ajax directly](https://github.com/OpenRefine/OpenRefine/blob/82b1cf0bfaf3c8ab57a12e0f2004bdeb71432383/main/webapp/modules/core/scripts/reconciliation/recon-manager.js#L93) rather than with the [jQuery shorthand](https://api.jquery.com/jquery.getjson/).

For clarity, this is using OpenRefine 3.4.1 on a Mac with FF 84.0.1.

[EDIT: You can see the use of JSONP from the web console using the Wikidata reconciliation service, so no need to implement your own service.]

gitonthescene:
Also fixed a small bug that doesn't account for the optional "suggest" field of the [service manifest](https://reconciliation-api.github.io/specs/latest/#service-manifest).

Fixes #3672

- Allow reconciliation services to drop support for JSONP.

NOTE:  The issue contains a question about whether data extension should be provided through the "reconcile" endpoint.  I think the [Reconiliation API](https://reconciliation-api.github.io/specs/latest/) should be clearer on that.  I'll open up an issue there.


wetneb (approved):
Thank you very much, that was an oversight on my part indeed!

PR #3716, opened 2021-03-13
Show left panel when a facet is created

### LINKED ISSUE #2891: Facet UI remains hidden when left panel set to be hidden
When using facets with the data-table set to full width (added by #2616), the left panel does not reappear to show the facet UI. The facet happens, and the user will need to toggle the left panel back to visible to be able to see it. 

Since we are not telling the user that it has happened in anyway (except for the table going back to the first row), the left panel should automatically reappear to show the results. We could also just send a message that tells the user to toggle the panel back to see it. Although, the former is more user-friendly.

antoine2711:
Fixes #2891

Added the Refine.showLeftPanel() function.

When a facet is created, make sure the left panel is being shown.

### DIFF
--- a/main/webapp/modules/core/scripts/facets/facet.js
+++ b/main/webapp/modules/core/scripts/facets/facet.js
@@ -37,6 +37,8 @@ class Facet {
   	this._config = config;
   	this._options = options || {};
   	this._minimizeState = false;
+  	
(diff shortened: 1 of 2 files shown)


wetneb (approved):
Thank you very much Antoine, and welcome back!

PR #3717, opened 2021-03-14
Proposition for a more controlled facet mechanism

antoine2711:
Proposition for a more controlled facet mecanism.

Does just like « project renaming », but for facets, with lots of code to be able to recreate the original name. 

A more simplified PR will be submitted.

Regards,
   Antoine


main/webapp/modules/core/scripts/facets/facet.js
```diff
@@ -35,6 +35,13 @@ class Facet {
   constructor(div, config, options) {
   	this._div = div;
   	this._config = config;
+  	
+  	if(this._config.nameAtCreation == "" || this._config.nameAtCreation == null)
+  	  this._config.nameAtCreation = config.name;
+  	  
+  	if(this._config.source == "" || this._config.source == null)
+  	  this._config.source = config.columnName;
+
```
wetneb, line 44: I am a bit confused by these new fields - I am not sure why it is necessary to introduce them? If we add them, doesn't that mean we should also add support for them in the backend (for instance so that those configuration fields appear in the JSON version of the history)?

antoine2711: @wetneb : if the user modifies the name of the facet, and then wants to get back the original name, we must save it. The source is an indicator to help the user to know what this facet is about. 

I know this add quite some characters to the permalink, and that there is a limit on the URL length, but still, I believe it's worth.

As far as I know, nothing get sent from facets to the Java server.

Regards,
   Antoine

wetneb: I am not sure why it is useful to let the user restore the original name to be honest. I'd just keep it simple and only store the current name. For instance, for projects, we don't store the original name either, and it works just fine.

antoine2711: @wetneb : I did that (add the name in the JSON of facets) because it's the way that the permalink will get it. It I want to retain both the changed name and the original name, I had to do this (add it to the JSON).

But, I will change that to just a ID, numerical, and I will be able to restore both new info I added. It will be clean and short.

Regards,
   Antoine

wetneb: Thanks! I think we should be able to implement this without adding any new field to the JSON representation of facets.

wetneb (changes requested):
I am a bit worried by the turn this is taking - this seems to be very far away from what I would expect. It would be worth discussing more before you make complicated changes like this.

I don't see why we could rely on numerical ids like this - note that we need to support custom facets with arbitrary expressions. Introducing numerical ids creates a whole lot of issues (how do we ensure stability of those ids? How can extensions define facets with new ids without conflicting with existing ones?).

We really should have a proper conversation asap before you invest more energy into something that will not be mergeable at all.

antoine2711 (commented):
PR changed to draft, this PR will be kept as a proof of concept, but as no future.

Regards, A.

antoine2711:
> I am a bit worried by the turn this is taking - this seems to be very far away from what I would expect. It would be worth discussing more before you make complicated changes like this.
> 
> I don't see why we could rely on numerical ids like this - note that we need to support custom facets with arbitrary expressions. Introducing numerical ids creates a whole lot of issues (how do we ensure stability of those ids? How can extensions define facets with new ids without conflicting with existing ones?).
> 
> We really should have a proper conversation asap before you invest more energy into something that will not be mergeable at all.

@wetneb : yes, I was aware that it was going in the wrong direction. I still had to code it, and show it to you. But I will simplify all this very much, and putting aside the ID concept for facet. I wanted to go that far… just for the pleasure.

Expressions are taken in account. For the stability of ID, it would be a matter of just adding them with a new incremental ID. For extensions, I was thinking of a 4 bytes hash, CR16, * 256, so this gives a numerical space of 250 facets for the extension, with little chance of colliding with other extensions.

Don't worry, I'm not wasting anyone’s time with this branch anymore, and I am coming back with something you’ll like, much simplier. I already knew before committing that it was not mergeable material. I still believe the exercice has some merits.

Regards,
   Antoine

wetneb:
If this PR isn't meant to be merged perhaps we could close it?

PR #3718, opened 2021-03-14
Follow-up fix for facet panel resizing when using permalinks to restore facets

### LINKED ISSUE #2891: Facet UI remains hidden when left panel set to be hidden

antoine2711:
Fixes #2891

- If loaded thru a permalink, the new facets will not try to resize the display that is not yet loaded (and though crashes).

This fix a bug introduced in code merged to master yesterday.

Regards,
   Antoine

### DIFF
--- a/main/webapp/modules/core/scripts/project.js
+++ b/main/webapp/modules/core/scripts/project.js
@@ -184,6 +184,7 @@ Refine._showHideLeftPanel = function() {
 
 Refine.showLeftPanel = function() {
   $('div#body').removeClass("hide-left-panel");
+  if(ui.browsingEngine == undefined || ui.browsingEngine.resize == undefined) return;
   resizeAll();
 };
 


antoine2711:
@wetneb : FYI.

main/webapp/modules/core/scripts/project.js
```diff
@@ -184,6 +184,7 @@ Refine._showHideLeftPanel = function() {
 
 Refine.showLeftPanel = function() {
   $('div#body').removeClass("hide-left-panel");
+  if(ui.browsingEngine == undefined || ui.browsingEngine.resize == undefined) return;
```
antoine2711, line 187: @wetneb : I know this is a cheap fix, if you have suggestions, I'm all ears…

Regards, A.

wetneb (approved):

PR #3729, opened 2021-03-16
Only show candidates in range of best score facet settings (fixes #3673)

### LINKED ISSUE #3673: Only show candidates in range for best candidate facet
After reconciling you're presented with a numerical facet called "best candidate's score" which you can use to adjust the range of which rows to show.  But currently it only limits which _rows_ to show and not which _candidates_ to show.

### Proposed solution

It would be nice if this facet limited _both_ which rows and which candidates to show so that only rows which have some candidates within this range (the current behavior) but also only show those candidates in this range as well.

### Additional context

This is mostly a "nice to have".  I think it reduces cognitive dissonance just a hair and also makes it a tiny bit easier to aim the mouse at the choice you want by removing clutter.

I'm guessing this is mostly a change on the client, though I'm not sure how accessible the range is when the rows are being rendered.

gitonthescene:
Fixes #3673

- When using best candidate score facet, don't only limit which rows are shown, but limit which candidates are shown in those rows.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
@@ -210,7 +210,7 @@ DataTableCellUI.prototype._render = function() {
             $('<span></span>').addClass("data-table-recon-score").text("(" + score + ")").appendTo(liSpan);
           };
 
-          bestScoreFacets = ui.browsingEngine.getJSON().facets.filter( facet => facet.expression == 'cell.recon.best.score' );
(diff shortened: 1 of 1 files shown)


wetneb:
I am reluctant but it hasn't got much to do with the PR so I have commented in #3673 instead.

gitonthescene:
@wetneb No worries.  I'm happy to discuss.

wetneb:
Following the discussion in #3673 I propose to close this - I'd say we need a more generic design to let users filter recon candidates and work on them independently.

gitonthescene:
Sounds good to me.

PR #3731, opened 2021-03-16
Fix URI encoding for View and Preview services

### LINKED ISSUE #3730: Preview & View services amputate ids with a hash character "#"
### To Reproduce
Configure a reconciliation service that returns URIs as IDs. 

Hovering over reconciliation suggestions,  automatically matched values and manually mapped values will fail to retrieve the preview information as ID contained in the callback is truncated.

matthieuj:
Fixes #3730

- encode reconciliation item IDs using encoreURIComponent so special characters like # used in URIs are properly encoded.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
@@ -149,7 +149,7 @@ DataTableCellUI.prototype._render = function() {
       .appendTo(divContentRecon);
 
       if (service && (service.view) && (service.view.url)) {
-        a.attr("href", encodeURI(service.view.url.replace("{{id}}", match.id)));
+        a.attr("href", service.view.url.replace("{{id}}", encodeURIComponent(match.id)));
       }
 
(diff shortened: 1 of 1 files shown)


wetneb:
That was quick! I just realized it's a bit trickier than I thought (see https://github.com/OpenRefine/OpenRefine/issues/3730#issuecomment-800158508). Would you be keen to also clarify the specs in the same go?

fsteeg:
We've updated the specs for the behavior implemented in this PR for view and preview: https://github.com/reconciliation-api/specs/pull/71. For flyout, [it's already implemented](https://github.com/OpenRefine/OpenRefine/blob/4c7ba73c992a407f8d7c746dfa4cc1a977ca07c3/main/webapp/modules/core/externals/suggest/suggest-4_3a.js#L1480) with `encodeURIComponent`.

wetneb:
The specs have been updated so this should now be good to go.

wetneb:
I think it might be worth rethinking about this, as this PR rendered the OpenCorporates reconciliation service incompatible with the upcoming version of OpenRefine (3.6).

PR #3741, opened 2021-03-17
(I# 1011) Simple implementation of facet renaming

### LINKED ISSUE #1011: Custom facet names

antoine2711:
First simple implementation of facet naming. If new name is empty, do nothing.

Fixes #1011

- Add an :hover over the facet's title
- On click on the title, bring a dialog to edit name, just like project name
- Display a ToolTip with the column name and the expression if there is any

Very simple implementation.

Regards,
   Antoine


main/webapp/modules/core/langs/translation-en.json
```diff
@@ -300,6 +300,11 @@
     "core-dialogs/for-null-cell-value-to-empty-str-label": "Convert null value to NULL in INSERT",
     "core-dialogs/choose-export-destination": "Please choose the destination for project export",
     "core-dialogs/export-to-local": "OpenRefine project archive to file",
+    "core-facets/column-tooltip": "Column: $1",
+    "core-facets/expr-tooltip": "Expression: $1",
+    "core-facets/source-tooltip": "Source: $1",
+    "core-facets/facet-current-title": "The current name of the facet is: $1.\n Enter the new facet's name",
```
wetneb, line 306: I would perhaps go for something simpler, without giving the current name, just like when editing the project name. Something like `Enter the new facet name`?

wetneb (commented):
Very nice, thank you! Perhaps it could be nice to add a UI test in the same go, just to demonstrate the feature on a facet?

main/webapp/modules/core/langs/translation-en.json
antoine2711: @wetneb : I respectfully think that this additional information is worth.

When making the decision to change a facet name, the old name might be important.
Also, one side advantage of doing this is making the dialog bigger. Since it can appear far from the facet, bigger maker is more easier to see for the user.

Anyway, if I understand $i18n correctly, changing this string will not break the code, so this is very easy to change in the future.

Regards, A.

gitonthescene: Could you have the best of both words by having the shorter label but have the old name pre-filled in the dialog?  Maybe also selected to make deletion easier.

antoine2711: > Could you have the best of both words by having the shorter label but have the old name pre-filled in the dialog? Maybe also selected to make deletion easier.

Sure, « Current facet name: XYZ » could do. But is it really better?
For the dialog, no luck, it's a JS prompt, and we have little control over that.

Regards,
   Antoine

gitonthescene: No, I actually meant having the label “Enter facet name:” but putting the old name in the text entry area, highlighted to make it clear it’s changing.  It should be obvious that that is the old name and it makes it easier to modify if you only need to add a suffix, say.

gitonthescene: Or maybe just “Rename facet $1:” as a label.

antoine2711: @gitonthescene : hey, why not… At this point, I guess it's a matter of preference. 

I did your last suggestion, but with « ' » to delimit the name.
I now it's pretty short! ;-)

Regards,
   Antoine

antoine2711:
@wetneb : thanks for merging this in v3.6.

Regards,
   Antoine

wetneb:
yes, that's my intention! I will try writing a Cypress for it myself.

antoine2711:
I’ve just been too busy, I still plan to write the test cases, like I
committed.

Regards,
   Antoine


Le mer. 24 nov. 2021 à 08:59, Antonin Delpeuch ***@***.***> a
écrit :

wetneb:
Ok great! Then I leave it to you. You might have seen that there are some problems with the stability of some Cypress tests: https://github.com/OpenRefine/OpenRefine/issues/4302. So you should ignore those and just make sure the tests you write do pass :)

antoine2711:
> Ok great! Then I leave it to you. You might have seen that there are some problems with the stability of some Cypress tests: #4302. So you should ignore those and just make sure the tests you write do pass :)

@wetneb: I have written the functional test. I need to squash my commit… But for the rest, it should be good.

Regards, Antoine

wetneb:
It looks like a relevant Cypress test is still failing.

antoine2711:
> It looks like a relevant Cypress test is still failing.

Yes @wetneb. This was triggered by a change that I made to have all the facet with standardized HTML structure.
It should be fixed now…

Regards, Antoine

wetneb:
I would close this as we now have a lot of merge conflicts, and the "File changed" tab shows unrelated changes. But this feature is very much needed!

PR #3746, opened 2021-03-17
Sanitize cookie keys in Wikibase extension

### LINKED ISSUE #3745: Invalid characters in cookies set by the Wikibase extension
As part of the update of Jetty I am also upgrade the javax.servlet-api, from
```
<dependency>
      <groupId>javax.servlet</groupId>
      <artifactId>servlet-api</artifactId>
      <version>2.5</version>
</dependency>
```
to
```
<dependency>
      <groupId>javax.servlet</groupId>
      <artifactId>javax.servlet-api</artifactId>
      <version>3.1.0</version>
</dependency>
```
And that introduces new checks on cookie keys (in the `Cookie` constructor). We seem to use cookies such as "https://www.wikidata.org/w/api.php-wb-consumer-token"

This raises exceptions: `java.lang.IllegalArgumentException: Cookie name "https://www.wikidata.org/w/api.php-wb-consumer-token" is a reserved token`

It seems to be because characters like `:` are not allowed in cookie names.
https://stackoverflow.com/questions/1969232/what-are-allowed-characters-in-cookies

We should transform the wikibase endpoint to remove special characters before including it in the cookie name.

wetneb:
Fixes #3745.

- Replace some special characters in cookie keys so that they are valid even by strict standards

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/commands/LoginCommand.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/commands/LoginCommand.java
@@ -40,6 +40,8 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static org.apache.commons.lang.StringUtils.isBlank;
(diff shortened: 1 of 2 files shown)


afkbrb (approved):

afkbrb:
LGTM :)

PR #3755, opened 2021-03-18
Attempt to fix flaky integration tests

fgiroud:
This PR is an attempt to fix some tests that are flaky.

- apply.spec.js
- add_columns_based_on_this_colum.spec.js

Also added some assertions to ensure OpenRefine is properly loaded in method `loadAndVisitProject`, I believe that should also help to lower the flakyness rate

### DIFF
--- a/main/tests/cypress/cypress/integration/project/grid/column/edit-column/add_columns_based_on_this_column.spec.js
+++ b/main/tests/cypress/cypress/integration/project/grid/column/edit-column/add_columns_based_on_this_column.spec.js
@@ -12,7 +12,7 @@ describe(__filename, function () {
     cy.typeExpression('value.toLowercase()');
     cy.get(
       '.expression-preview-table-wrapper tr:nth-child(2) td:last-child'
-    ).contains('butter,with salt');
+    ).should('to.contain', 'butter,with salt');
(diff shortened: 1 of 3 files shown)


main/tests/cypress/cypress/integration/project/grid/column/edit-column/add_columns_based_on_this_column.spec.js
```diff
@@ -12,7 +12,7 @@ describe(__filename, function () {
     cy.typeExpression('value.toLowercase()');
     cy.get(
       '.expression-preview-table-wrapper tr:nth-child(2) td:last-child'
-    ).contains('butter,with salt');
+    ).should('to.contain', 'butter,with salt');
```
kushthedude, line 15: How is this fixing the text 😕 ?

fgiroud: I am not sure, but it's working :-), when running 10x times locally I can have some flaky tests, even locally
With the new version I couldn't reproduce the flaky test

kushthedude (approved):
If it works then let's get into the master 🙂

Issue #3664, opened 2021-02-23
Every other reconciliation batch fails in 3.5


tfmorris:
With the current development version, every other reconciliation call fails with a `NoHttpResponseException`, apparently because it's attempting to reuse a closed connection.

### To Reproduce

Reconcile a project with more than 20-30 rows and observe that the console logs failures on every other call.

### Current Results
Every other batch fails with an HTTP error.

### Expected Behavior
All batches get reconciled

### Versions
 - OpenRefine: current development version

gitonthescene:
I think there is a related bug with using Data Extension.  It might be exactly the same bug, but I thought it worth pointing out.  Getting `127.0.0.1:5000 failed to respond` messages for every other batch.

gitonthescene:
Oho!!  This looks like it might be a issue with the server.  I found [this post](https://stackoverflow.com/questions/10523879/how-to-make-flask-keep-ajax-http-connection-alive) which enables [Keep-Alive](https://www.imperva.com/learn/performance/http-keep-alive/) on my flask server and I'm not seeing issues with every other batch failing.

This raises the question about whether to consider this a server problem or a client problem.  Should OpenRefine be able to handle servers which close connections after each request?  Or maybe we should require that `Keep-Alive` be enabled for reconciliation servers.

In any event, I believe this to be the source of the problem.  OpenRefine is [pooling connections](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/util/HttpClient.java#L57) but they're being closed on the server side when `Keep-Alive` is not enabled.

gitonthescene:
On the positive side, this was probably exposed because reconciliation is so fast on OpenRefine 3.5!!  :D

tfmorris:
> Should OpenRefine be able to handle servers which close connections after each request? 
> Or maybe we should require that Keep-Alive be enabled for reconciliation servers.

Yes, we need to handle this. Persistent connections are a performance optimization, but the system should still be functional without it.

Our HTTP implementation has changed completely since last release, so some of the behavior changes may be associated with the switch to Apache HTTP Client 5.

I should have noted what reconciliation service I was using on the original ticket, but I suspect it was your CSV reconciliation service, so the comment about this being specific to servers that don't support Keep-Alive makes sense.

gitonthescene:
[...]

PR #3770, opened 2021-03-24
Add exponential backoff retries for reconciliation calls.

### LINKED ISSUE #3369: Parts of reconciliation batches can fail silently; and parts of reconciled column lookup batches can fail silently
This may be a duplicate of an existing bug or bugs, but I find it incredibly frustrating: when items are 'unmatched' it appears to be impossible to distinguish whether there is really no match for the item, or whether the reconciliation process simply failed to complete for some sub-batch of entries.

A couple of days ago, I was running a reconciliation of a set of authors with LoC authority IDs against Wikidata using the "reconcili.link" service.  I ran it once, and about 80% reconciled.  I then filtered for the unmatched authors, reconciled again, and a further 15% of the original set reconciled.

I presume that what happened was that during the first reconciliation, one or more sub-batches of contiguous records timed out, or otherwise failed to return.  (The problem did appear to be affecting contiguous records).   That, I guess, is something that can always happen.  The problem is that OpenRefine then marked those records' reconciliation status as "unmatched" rather than "unknown", not differentiating between items for which the reconcilation had failed to complete from those for which the reconcilation had concluded but found nothing.

Similarly, when I then added a column looking up LoC IDs based on the reconciled values (which I find is a necessary thing to have to do to sanity check reconcilations), only about 80% of the IDs got added, with more added when I re-ran the lookup for those without values.

This is frustrating enough for me, aware that this can happen, knowing that I may need to re-run reconciliations or augmentations two or three or even more times to be sure.  But the even worse impact is with unaware users, who as a result may then add new duplicate items to Wikidata for their unreconciled entries, without realising that the unreconciled status cannot be depended on.

wetneb:
Fixes #3369.

This does not touch the existing (broken) retry mechanism to retry individual queries in a given batch. We just retry the HTTP call up to 4 times, with exponential backoff.

Overall this code is a big mess - this was already refactored quite a lot in the new architecture, so I am only making a minimal change here (hoping to only fix a bug and not introduce others). I hope to find the time to do more cleaning up in the new architecture in the coming months.


main/tests/server/src/com/google/refine/model/recon/StandardReconConfigTests.java
```diff
@@ -337,9 +337,8 @@ public void reconTest() throws Exception {
         try (MockWebServer server = new MockWebServer()) {
             server.start();
             HttpUrl url = server.url("/openrefine-wikidata/en/api");
-            // FIXME: Retry doesn't currently work, but should be tested
-//            server.enqueue(new MockResponse().setResponseCode(503)); // service overloaded
-            server.enqueue(new MockResponse().setBody(reconResponse));
+            server.enqueue(new MockResponse().setResponseCode(503)); // service initially overloaded
```
afkbrb, line 340: Do we expect the recon request to fail on the first try, so there should be something like "Retrying recon queries:..." on the console? I ran `StandardReconConfigTests.reconTest()` locally, but there's no such information 

It's worth mentioning that duplicating `server.enqueue(new MockResponse().setResponseCode(503));` for several times here did cause "Retrying recon queries:..." to be printed, not sure why.

afkbrb: > It's worth mentioning that duplicating server.enqueue(new MockResponse().setResponseCode(503)); for several times here did cause "Retrying recon queries:..." to be printed, not sure why.

All right, after taking a look at [HttpClient.java](https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/util/HttpClient.java), I found that a request will be retried up to 3 times if the server returns 503, since we do use the [ExponentialBackoffRetryStrategy](https://github.com/OpenRefine/OpenRefine/blob/68b148d7755c829cf42fadba5b821c1ba193d050/main/src/com/google/refine/util/HttpClient.java#L185) when building the http client.

The change to this testcase doesn't cause the newly added retry logic to be executed as I understand and tested, but the logic does work as I tested. Anyway, since this is a minimal change and a temporary fix, it should be ok to merge it into the master branch :)

afkbrb (approved):

gitonthescene:
Linking issue potentially addressed by this.  https://github.com/OpenRefine/OpenRefine/issues/3664

main/tests/server/src/com/google/refine/model/recon/StandardReconConfigTests.java
wetneb: Ah ok then I need to have a second look at it - I think the ExponentialBackoffRetryStrategy is not used for POST requests, so I thought the change to the testcase did test my changes. If that's not the case then I need to revise my PR.

tfmorris:
Perhaps this is a good indication that we should be using GET, not POST.

You might want to take a look at the Apache retry handler for the types of errors that it doesn't attempt to retry. It looks like the current code retries everything.

wetneb:
Actually I take that back! @afkbrb is correct that ExponentialBackoffRetryStrategy is used here too. So #3369 was actually fixed by @tfmorris' refactoring. Brilliant! This PR just updates the test to witness that.

afkbrb:
The http client used in `com.google.refine.model.recon.StandardReconConfig` is instantiated with the  `_delay` field of `com.google.refine.util.HttpClient` set to 0. And in the following code, the retry interval is calculated as `pow(2, execCount) * defaultInterval`, where the defaultInterval is actually equal to _delay. Since _delay is set to 0, the interval will always be 0! In that case, we will actually always retry immediately after getting a 503 response.

https://github.com/OpenRefine/OpenRefine/blob/2de5929350b32e08cb52eeab7e4ab3203235fd76/main/src/com/google/refine/util/HttpClient.java#L194-L206

To fix this, we could simply set _dealy to, for example, 1000 (ms) when instantiating the http client. But since the interval between  
 two contiguous requests is also set to _delay, that will slow down the recon rate.

afkbrb:
That makes more sense!

main/tests/server/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperationTests.java
```diff
@@ -337,7 +337,7 @@ public void testExponentialRetries() throws Exception {
             }
 
             // Use 503 Server Unavailable with no Retry-After header this time
-            for (int i = 0; i < 5; i++) {
+            for (int i = 0; i < 4; i++) {
```
afkbrb, line 340: I don't think this should be changed, and changing this will cause inconsistency with the comment below in the code.

```java
...
// 1st row fails after 4 tries (3 retries), 2nd row tries twice and gets value, 3rd row is hard error
...
```

afkbrb (changes requested):

main/tests/server/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperationTests.java
wetneb: I changed it because I thought that the previous version was inconsistent with the comment actually. This adds four 503 requests: 4 tries (3 retries). The previous code added 5 of them - do you understand why?

afkbrb: Seems you ignored `2nd row tries twice and gets value`. Besides the four 503 responses consumed by the first row, the second row is also supposed to get a 503 response on the first try,

afkbrb: That's why there's another 200ms at the end of the following comment:

```
// 6 requests (4 retries 200, 400, 800, 200 msec) + final response
```

wetneb: Oops, thanks indeed! sorry about that.

afkbrb (approved):

main/src/com/google/refine/util/HttpClient.java
```diff
@@ -70,7 +76,7 @@ public HttpClient(int delay) {
                 .setDefaultRequestConfig(defaultRequestConfig)
                 .setConnectionManager(connManager)
                 // Default Apache HC retry is 1x @1 sec (or the value in Retry-Header)
-                .setRetryStrategy(new ExponentialBackoffRetryStrategy(3, TimeValue.ofMilliseconds(_delay)))
+                .setRetryStrategy(new ExponentialBackoffRetryStrategy(3, TimeValue.ofMilliseconds(_retryInterval)))
 //                .setRedirectStrategy(new LaxRedirectStrategy()) // TODO: No longer needed since default doesn't exclude POST?
 //               .setConnectionBackoffStrategy(ConnectionBackoffStrategy)
```
tfmorris, line 81: This TODO can probably be deleted based on the investigation that was done here confirming the comment to be correct.

tfmorris (commented):
LGTM. One potential future cleanup.

PR #3777, opened 2021-03-30
Reconciliation tests

fgiroud:
Added tests for the reconciliation features of OpenRefine


wetneb (commented):
Very nice! As I mentioned before, I think it would be interesting to use a pre-reconciled project archive as fixture to avoid having to run the reconciliation operation as a first step to many tests. I think that should help speed the tests up (although I haven't actually checked how long those steps currently take). But it's not a blocker.

wetneb (approved):

Issue #3109, opened 2020-08-22
Validate Wikibase manifest on the server


tfmorris:
Currently the backend depends on the manifest validation which is done in the web client. We should also validate on the backend, so that we're not relying on well-behaved clients, as discussed here: https://github.com/OpenRefine/OpenRefine/pull/2810#discussion_r475101438

### Current Results
A manifest without a version field will currently likely cause the backend to get a NullPointerException when it tries to process the version.

### Expected Behavior
Server works correctly regardless of the data that it receives.

wetneb:
Given that Wikibase manifests are stored in the preference store, it is not clear to me how to do this: the preference store can store arbitrary values and there is no clear way to set up custom validation for one of those fields. Also, the preference store is a core component, whereas the Wikibase integration is an extension.

PR #3779, opened 2021-04-01
Remove EditGroups URL template from Wikibase schema

### LINKED ISSUE #3656: Wikibase schema serialization issues
In the migration to arbitrary Wikibase support, some changes to the Wikibase schema serialization were introduced.
The following fields were added:
* `mediaWikiApiEndpoint`
* `editGroupsURLSchema`

I can see two problems with this:

### Lack of serialization of `editGroupsURLSchema`

This field is currently marked as `@JsonIgnore` so it is not serialized to the schema. This means that when re-opening a project, the schema will be restored without this parameter, causing OpenRefine to edit without EditGroups link.

### Potential lack of backwards-compatibility with schemas serialized with previous versions
It would be nice to still support schemas serialized by previous versions. We should ensure that we can still import the schema without these two parameters. Namely, these could be retrieved from the list of installed manifests (since we have the `siteIri` to identify which instance we are talking about). Or even simpler: the absence of the `mediaWikiApiEndpoint` parameter indicates that the schema was made by a version of OpenRefine which only supported Wikidata, so we can directly prefill those parameters with constants in that case.

wetneb:
Instead it is fetched from the manifest and stored in operation metadata. I propose to keep the MediaWiki API endpoint, because it could be considered a better identifier for the Wikibase instance than the siteIRI, so it might be useful to keep it around. Also, that field was properly exposed during serialization so far (no `@JsonIgnore` annotation) so schemas created with a snapshot of the master branch should still work after this change (as well as the ones created with 3.4, which are of course still valid).

Closes #3656.


extensions/wikidata/src/org/openrefine/wikidata/manifests/ManifestV1.java
```diff
@@ -25,6 +28,9 @@ public ManifestV1(JsonNode manifest) {
         JsonNode mediawiki = manifest.path("mediawiki");
         name = mediawiki.path("name").textValue();
         mediaWikiApiEndpoint = mediawiki.path("api").textValue();
+        if (mediaWikiApiEndpoint == null || mediaWikiApiEndpoint.isEmpty()) {
```
afkbrb, line 31: If this happens, it means that the manifest provided is invalid, so perhaps we should throw some Exception? This is kind of a validation to the manifest, it could be better to do this in a fix to #3109.

wetneb: Ah good point, I'll remove it.

afkbrb (approved):
Looks great!

PR #3789, opened 2021-04-06
Common transforms on multiple columns

### LINKED ISSUE #1843: Add option to use the "Common Transforms" when applying transformations from the "All" dropdown menu
When applying transformations to All columns (from the "All" dropdown) you have to write the transformations by hand - there is no option to select the pre-set transforms that are in the Common transforms menu when you apply a transformation to a single column

Add a 'Common Transfroms' option to the All drop down menu. After selecting a transform from the list you would be prompted to select the columns that the transformation will be applied to (in the same way as the "Transform" option from the All menu

Originally requested on  #1724 by @jenyoung  - splitting out here to keep the feature requests separate

singhakshita:
Fixes #1843 }

- added common transform menu
- choose any if the transform and then select columns and apply those transforms
Screenshot 2021-04-06 at 12 10 47 PM
Screenshot 2021-04-06 at 12 10 58 PM

- 
-


magdmartin:
I like how we have more option to batch edit all column at once. Thank you for the PR!

As the `All` menu expand, I suggest we group this function under a `Edit all columns`. 
The All menu will be organized like this: 
* `Edit all cells`
  *  `Transform`
  * `Trim leading ...`
  * `Collapse consecutive ...`
  * `Unescape HTML entities` 
  * ...
* `Facet`
* `Edit rows`
* `Edit columns`
* `View`

singhakshita:
is there anything else that i should change

martinrefinepro:
thanks for the update @singhakshita 
Your pull request will be technically reviewed and merged by one of OpenRefine senior developer (@wetneb or @tfmorris )

magdmartin:
thanks for the update @singhakshita
Your pull request will be technically reviewed and merged by one of OpenRefine senior developer (@wetneb or @tfmorris )

wetneb (commented):
Hi @singhakshita, thank you very much for all this work and sorry I did not review this earlier, I hadn't seen that you took Martin's feedback into account already.

It looks very good to me. Perhaps one tiny detail I would bring up is the fact that the dialog does not mention anywhere which of the common transforms is getting applied. I know the user just clicked on the corresponding entry, but I feel like it could still be useful to remind it there. Perhaps the entry in the menu could actually be used as dialog title? What do you think?

singhakshita:
yeah ,adding that would be better .I am on it

singhakshita:
updated the dialog header with elaborative header i.e name of the transform
Screenshot 2021-04-21 at 9 32 50 PM

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -660,66 +660,66 @@ DataTableView.prototype._createMenuForAllColumns = function(elmt) {
...
           id: "core/to-blank",
           label: $.i18n('core-views/blank-out'),
-          click: function() { new commonTransformDialog("null", "keep-original", false, ""); }
+          click: function() { new commonTransformDialog("null", "core-views/blank-out"); }
         },
         {
           id: "core/to-empty",
           label: $.i18n('core-views/blank-out-empty'),
-          click: function() { new commonTransformDialog("\"\"", "keep-original", false, ""); }
+          click: function() { new commonTransformDialog("\"\"","ore-views/blank-out-empty" ); }
```
wetneb, line 722:
```suggestion
          click: function() { new commonTransformDialog("\"\"","core-views/blank-out-empty" ); }
```

singhakshita: sorry for the typo , fixed

singhakshita:
@wetneb  please let me know if this works

wetneb:
I think the new UI is a lot better but it seems that the styling of the buttons has gone wrong a bit, the font size looks bigger than usual to me:
image

main/webapp/modules/core/scripts/dialogs/common-transform-dialog.html
```diff
@@ -0,0 +1,12 @@
+<div class="dialog-frame" style="width: 400px;">
+    <div class="dialog-header" bind="dialogHeader"></div>
+    <div class="dialog-body" bind="dialogBody"><div class="grid-layout grid-layout-for-ui layout-normal layout-full"> 
+    <td width="100%"><div id="checkedColumn" class="column-reordering-dialog-column-container" bind="columnContainer"></div></td>
+    <div  class="all-column-transform-dialog-container dialog-footer" bind="dialogFooter" style="font-size: small;" >
```
wetneb, line 5: I think we should avoid explicitly setting `font-size: small` here, but rather rely on the existing CSS classes. It seems to me that the two divs opened above (`<div class="dialog-body" bind="dialogBody"><div class="grid-layout grid-layout-for-ui layout-normal layout-full">`) are not closed anywhere, which might be the cause of the styling issue.

singhakshita: I changed it ,please see once

wetneb (approved):
Thank you very much, it works great!

PR #3795, opened 2021-04-08
fix: all flaky tests related to create error

kushthedude:

- The flaky tests were due to exception raised by the cypress due to the window failure, this here ignored that only one exception


kushthedude:
@fgiroud , I think this is going to fix all the flaky tests related to create-project

fgiroud (approved):

PR #3810, opened 2021-04-12
Fixes #3809. Enables cross-origin property type-ahead.

### LINKED ISSUE #3809: Type-ahead property selection fails with CORS exception in data-extension panel
### To Reproduce
Configure OpenRefine with a reconciliation service that provides a property suggestion service. 

The manifest for such service includes:

	"suggest": {
		...
		"property": {
			"service_path": "/suggest/property",
			"service_url": "http://dock05:8080",
			"flyout_service_path": "/flyout/property?id=${id}"
		},
		...
		

Reconcile against that reconciliation server, then on the reconciled column menu select "Edit column" then "Add columns from reconciled values".

Start typing in the "Add property" text box

### Current Results
Instead of receiving suggestions as expected the browser consoles shows: "Sorry, something went wrong. Please try again later"

At the same time the browser console shows:

    project-bundle.js:10439 Cross-Origin Read Blocking (CORB) blocked cross-origin response http://dock05:8080/suggest/property?callback=jQuery112406009935240625213_1618220980169&prefix=d&spell=always&exact=false&scoring=schema&prefixed=true with MIME type application/json. See https://www.chromestatus.com/feature/5629709824032768 for more details.

matthieuj:
Fixes #3809. Enables cross-origin property type-ahead.

### DIFF
--- a/main/webapp/modules/core/scripts/dialogs/extend-data-preview-dialog.js
+++ b/main/webapp/modules/core/scripts/dialogs/extend-data-preview-dialog.js
@@ -171,6 +171,12 @@ ExtendReconciledDataPreviewDialog.prototype._show = function(properties) {
   suggestConfig.key = null;
   suggestConfig.query_param_name = "prefix";
 
+  // CORS/JSONP support
+  if (this._serviceMetadata.ui && this._serviceMetadata.ui.access) {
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thank you very much!

PR #3820, opened 2021-04-16
Broken link in reconciling.md

padefla:
The link at line 20 is broken.
It should be:  [clean and cluster](https://docs.openrefine.org/manual/cellediting)

Fixes #{issue number here}

- 
- 
-

### DIFF
--- a/docs/docs/manual/reconciling.md
+++ b/docs/docs/manual/reconciling.md
@@ -17,7 +17,7 @@ You may wish to reconcile in order to:
 * add to an editable platform such as [Wikidata](https://www.wikidata.org)
 * or see whether entities in your project appear in some specific list, such as the [Panama Papers](https://aleph.occrp.org/datasets/734).
 
-Reconciliation is semi-automated: OpenRefine matches your cell values to the reconciliation information as best it can, but human judgment is required to review and approve the results. Reconciling happens by default through string searching, so typos, whitespace, and extraneous characters will have an effect on the results. You may wish to [clean and cluster](cellediting) your data before reconciliaton.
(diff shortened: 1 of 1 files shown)


wetneb:
Hi @padefla, thanks for this PR!

I cannot reproduce the issue on my side: when I go to https://docs.openrefine.org/manual/reconciling and click "clean and cluster", I am directed to https://docs.openrefine.org/manual/cellediting as expected.

Which browser and operating system is this happening with?

wetneb:
Ah, hang on, I can actually reproduce that too. It seems to be due to the presence (or absence) of a trailing slash at the end of the current page URL.

For instance, if you go to https://docs.openrefine.org/manual/reconciling, the links there will work, but if you go to https://docs.openrefine.org/manual/reconciling/, the links will not work.

padefla:
Hi @wetneb,
I am using Firefox 87.0 on macos 10.14.6 (18G8022)
when I click "clean and cluster", I am directed to https://docs.openrefine.org/manual/reconciling/cellediting which returns a "Page Not Found"

padefla:
I just noticed that when I use https://docs.openrefine.org/manual/reconciling to get the page, after loading the URL displayed is actually https://docs.openrefine.org/manual/reconciling/

wetneb:
@padefla yes I also noticed that. I have just changed deploy settings and I think this has solved the issue, can you check on your side? When you click on https://docs.openrefine.org/manual/reconciling, no trailing slash should get added to the URL.

We still have the issue that if someone does visit a URL with a trailing slash, the links will be broken there. I am looking into adding a bit of JS to automatically remove those trailing slashes.

padefla:
Now it works fine.

Thank you for the prompt reaction.

Have a nice day!

PR #3825, opened 2021-04-18
Fix date parsing for XLS and ODS files to avoid timezone-dependency.

### LINKED ISSUE #3740: Timezone-dependent test fails in CET
One of our tests fails when it is run in Central European Time:

```
OdsImporterTests.readOds:113 expected [2012-03-28] but found [2012-03-27]
```

### To Reproduce
1. First, change your timezone to CET
2. Then, `./refine test`

### Versions
 - Operating System: Debian
 - JRE or JDK Version: 11
 - OpenRefine: master

wetneb:
Closes #3740.

- Avoid drift in days because of inconsistency between date parsing (timezone-dependent) and representation (in UTC).

This was brought up by a test which was failing if run in the CET (or CEST) time zone. I don't know if there is a good way to override the default system timezone when running tests, so that this sort of problem is also detected by the CI. Perhaps we should generally set non-standard locales / timezones in our testing pipelines.


wetneb:
Merging this in the coming days unless anyone has concerns.

PR #3839, opened 2021-04-22
Conditionalize CI steps requiring secrets

### LINKED ISSUE #3677: Cypress builds fail in forked repo

wetneb:
Fixes #3677.

Follow-up to #3680 which this PR builds on.

The problem with #3680 (I think?) is that it disables secrets (so, coverage reporting and the Cypress dashboard) even for PRs made by org members / from within the repository. We want a way to still enable those as easily as possible when we see well-meaning PRs. One option is to add a label on the PR to mark it as trusted. I would also like to check if the author of the PR is an org member (in which case we would not have to label the PR manually) but I haven't found the syntax for that yet.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,7 +1,7 @@
 name: Continuous Integration
 
 on:
-  pull_request_target:
+  pull_request:
     paths-ignore:
       - 'docs/**'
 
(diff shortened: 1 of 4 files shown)


wetneb:
I am merging this without review as this change on the CI cannot be tested without being merged. Happy to revert if there are concerns.

wetneb:
It looks like the syntax I tried to introduce conditions is incorrect so I reverted this change.

PR #3840, opened 2021-04-22
Use environment variables for default workspace location on Windows, 

### LINKED ISSUE #2961: Replace jdatapath dependency with Java 8 user home directory finding
We currently use an obscure dependency, `jdatapath`, to find the user's home directory on Windows. Since Java 8 was taught how to do that correctly itself, we should switch to that and drop the dependency.

`System.getProperty("user.home")` or `org.apache.commons.lang3.SystemUtils.getUserHome()`

https://stackoverflow.com/questions/585534/what-is-the-best-way-to-find-the-users-home-directory-in-java

wetneb:
Fixes #2961.

According to code comments the invocation of the jdatapath library fails on 64bit systems, which should represent the overwhelming majority of our user base by now. We might be able to tweak the logic further using the standard way to detect the user directory (fixed in Java 8) as mentioned in #2961, but I am not sure if there is a standard equivalent for the "APPDATA" directory.


wetneb:
@thadguidry I have merged this and added it to the 3.5 branch. It would be useful if you could double-check that this does not change the default workspace directory on Windows (at least with the machine you have access to!)

Ideally it would be good to check the behaviour of this on a 32 bits Windows machine, as this is likely to be the case where there could be a difference, but maybe it's not so easy to find such machines these days!

thadguidry:
I tested the snapshot just now and works well.  However, read on...

32 and 64 bit versions won't matter.  What matters is User Profile context and data locality in Enterprise controlled user environments (where some of that we cannot control, but only use smart defaults)

There's no standard default for Application Data on any of the 3 OS's.  All 3 are tied to a User Profile and it changes if the system in use is an Enterprise system (Roaming, LDAP, etc.) on any of the 3 OS's.
For Windows 10 there are standard environment variables:  https://docs.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables

One thing that we do which is advertised incorrectly is that we say "OpenRefine always stores data locally".
That's **not entirely true** , since APPDATA (and what jdatapath point to) never had the LOCAL context, so a user in a corporate setting might be unaware that their data was being stored on the company network against their user profile (Roaming) and not on the local machine.
We probably should change the smart default to LOCALAPPDATA environment variable instead? (Then we'd need to inform users of that change of default from previous versions of OpenRefine).  Windows (and enterprise group policy) seamlessly handles the rest.  The variable will work, but the path might not be writable (group policy) and that's fine... at any time a path on any OS can be not writable.

wetneb:
Makes sense, it's probably worth opening an issue about LOCALAPPDATA vs APPDATA.

thadguidry:
@wetneb OK, I'll open a new issue about LOCALAPPDATA.

PR #3843, opened 2021-04-23
Support BCE dates in Wikibase upload

loulankxh:
Fixes #{3816}

- one test for BCE year added in WbDateConstantTest.java
- make change in method parse() of WbDateConstant.java to support BCE year(for example, regard a BCE year -1100
- as 1100 when call other read-only methods and finally change the year from positive 1100 to -1100 when return the TimeValue for parse())
-

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
@@ -108,11 +108,6 @@ public class WbDateConstant implements WbExpression<TimeValue> {
         String calendarIri = TimeValue.CM_GREGORIAN_PRO; // Gregorian calendar is assumed by default
         
         String trimmedDatestamp = datestamp.trim();
-
-        if(trimmedDatestamp.startsWith("-")){
-            trimmedDatestamp = trimmedDatestamp.replace("-", "");
-            bceFlag = true;
-        }
         
         if("TODAY".equals(trimmedDatestamp)) {
 	        Calendar calendar = Calendar.getInstance();
@@ -123,7 +118,11 @@ public class WbDateConstant implements WbExpression<TimeValue> {
 	    			(byte)0, (byte)0, (byte)0, (byte)11, 0,0,0, TimeValue.CM_GREGORIAN_PRO);
 	    	return todaysDate;
         }
-        
+
+        if(trimmedDatestamp.startsWith("-")){
+            trimmedDatestamp = trimmedDatestamp.substring(1);
+            bceFlag = true;
+        }
     	
         for (Entry<SimpleDateFormat, Integer> entry : acceptedFormats.entrySet()) {
             ParsePosition position = new ParsePosition(0);
--- a/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
@@ -104,7 +104,7 @@ public class WbDateConstant implements WbExpression<TimeValue> {
         Date bestDate = null;
         int precision = 0; // default precision (will be overridden if successfully parsed)
         int maxLength = 0; // the maximum length parsed
-        Boolean bceFlag = false; // judge whether this is a BCE year
+        boolean bceFlag = false; // judge whether this is a BCE year
         String calendarIri = TimeValue.CM_GREGORIAN_PRO; // Gregorian calendar is assumed by default
         
         String trimmedDatestamp = datestamp.trim();
--- a/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
@@ -104,9 +104,15 @@ public class WbDateConstant implements WbExpression<TimeValue> {
         Date bestDate = null;
         int precision = 0; // default precision (will be overridden if successfully parsed)
         int maxLength = 0; // the maximum length parsed
+        Boolean bceFlag = false; // judge whether this is a BCE year
         String calendarIri = TimeValue.CM_GREGORIAN_PRO; // Gregorian calendar is assumed by default
         
         String trimmedDatestamp = datestamp.trim();
+
+        if(trimmedDatestamp.startsWith("-")){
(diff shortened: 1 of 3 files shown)


extensions/wikidata/pom.xml
```diff
@@ -174,20 +174,25 @@
...
       <artifactId>mockwebserver</artifactId>
       <version>4.9.1</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.openrefine</groupId>
+      <artifactId>main</artifactId>
+      <version>3.5-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
```
wetneb, line 195: We should not need to add this here, it is already included above (line 117).

pom.xml
```diff
@@ -219,6 +219,11 @@
     <!-- dependencies are stored in the relevant submodules:
           see main/pom.xml, server/pom.xml and each extension.
     -->
+      <dependency>
+          <groupId>org.apache.commons</groupId>
+          <artifactId>commons-lang3</artifactId>
+          <version>3.4</version>
+      </dependency>
```
wetneb, line 226: We should not need to add dependencies here either.

extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
```diff
@@ -104,46 +104,51 @@ public static TimeValue parse(String datestamp)
         Date bestDate = null;
         int precision = 0; // default precision (will be overridden if successfully parsed)
         int maxLength = 0; // the maximum length parsed
+        Boolean bceFlag = Boolean.FALSE; // judge whether this is a BCE year
```
wetneb, line 107: How about using the primitive data type instead?
```suggestion
        boolean bceFlag = false; // judge whether this is a BCE year
```

extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
```diff
@@ -104,46 +104,51 @@ public static TimeValue parse(String datestamp)
...
         String calendarIri = TimeValue.CM_GREGORIAN_PRO; // Gregorian calendar is assumed by default
-        
+
         String trimmedDatestamp = datestamp.trim();
-        
+
+        if(trimmedDatestamp.startsWith("-")){
+            trimmedDatestamp = trimmedDatestamp.replace("-", "");
+            bceFlag = Boolean.TRUE;
```
wetneb, line 114:
```suggestion
            bceFlag = true;
```

extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
```diff
@@ -155,7 +160,10 @@ public static TimeValue parse(String datestamp)
             Calendar calendar = Calendar.getInstance();
             calendar = Calendar.getInstance();
             calendar.setTime(bestDate);
-            return Datamodel.makeTimeValue(calendar.get(Calendar.YEAR), (byte) (calendar.get(Calendar.MONTH) + 1), 
+            long year = calendar.get(Calendar.YEAR);
+            if(bceFlag == Boolean.TRUE)
```
wetneb, line 164:
```suggestion
            if (bceFlag)
```

wetneb (commented):
Hi @loulankxh, thanks a lot for this PR! That was fast!

It looks very good to me, but I can see a few things that we need to improved. First, your PR includes whitespace changes to some unrelated files. You can see this here: https://github.com/OpenRefine/OpenRefine/pull/3843/files. It would be better to keep these files unchanged. I have also added a few comments below.

pom.xml
loulankxh: Thanks, I will try to remove this unnecessary changes then

loulankxh:
The code is modified as required. Thanks!

extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
```diff
@@ -104,9 +104,15 @@ public static TimeValue parse(String datestamp)
         Date bestDate = null;
         int precision = 0; // default precision (will be overridden if successfully parsed)
         int maxLength = 0; // the maximum length parsed
+        boolean bceFlag = false; // judge whether this is a BCE year
         String calendarIri = TimeValue.CM_GREGORIAN_PRO; // Gregorian calendar is assumed by default
         
         String trimmedDatestamp = datestamp.trim();
+
+        if(trimmedDatestamp.startsWith("-")){
+            trimmedDatestamp = trimmedDatestamp.replace("-", "");
```
wetneb, line 113: Replacing every occurrence of `-` in the string is likely to cause problems if we want greater precision than years.
For instance, it seems that we would not be able to parse `-1234-10` (October 1234 BCE). Perhaps you could add a test case for this and use another method to remove the leading minus sign?

extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
```diff
@@ -104,9 +104,15 @@ public static TimeValue parse(String datestamp)
...
+            bceFlag = true;
+        }
         
         if("TODAY".equals(trimmedDatestamp)) {
```
wetneb, line 117: Since you added your handling of the leading minus sign before the handling of `TODAY`, it means that we would parse `-TODAY` as some BCE date, which does not seem ideal. How about moving your BCE-handling block after the `TODAY` one? In this way, `-TODAY` would be rejected (perhaps it is worth adding a test about that).

wetneb (commented):
Thanks! I noticed a few other things below. Almost there!

wetneb:
Ah, and one last thing: we should also update the frontend code which validates date constants client-side:
https://github.com/OpenRefine/OpenRefine/blob/791ae1f06be752d3f6e4a70347d50aa157325452/extensions/wikidata/module/scripts/schema-alignment.js#L1024

extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
loulankxh: Yes, that's sure

wetneb (approved):
Perfect, thank you very much!

Issue #3143, opened 2020-08-29
Enhance randomNumber() function


tfmorris:
The `randomNumber()` function currently takes two Number arguments, converts them to integers and returns a random number in the open interval.

### Proposed solution

I think the following changes would improve the function:
- shorten the name to just `random()`
- create a 0-arg version which returns a Double in the interval 0.0 - 1.0
- provide two alternatives for the 2-arg version which return either an Long or a Double, depending on type of the parameters which are passed in

### Alternatives considered
- the 2-arg integer version could have a different name like `randint` a la Python
- a variant that accepts a seed value could be useful for those who want reproducible results

SustechcsKing:
Hi, sir, I just add a function. First, it generates a random number by random(). Second, it can generate a random Long or random double by argument, but it still needs the typecasting when getting the number, because its return type is Object. I think this may meet your standard. Thank you.

elroykanye:
@wetneb , I noticed this was taken but not completed and pr closed. Please can I take and work on it?

PR #3845, opened 2021-04-24
Add files via upload

SustechcsKing:
Fixes #{3143}

- 
- 
-

### DIFF
--- a/main/src/com/google/refine/util/RandomNumber.java
+++ b/main/src/com/google/refine/util/RandomNumber.java
@@ -1,0 +1,28 @@
+
+public class RandomNumber {
+	
+	// This function will return a value from 0.0-1.0 (in close interval, for 1 decimal)
+	public static double random() {
+		double a = Math.random();
+		a = a * 11;
+		a = Math.floor(a);
+		return a/10;
+	}
+	
+	/* According to the requirement, first convert a, b to Long type, then 
+	   get a number in the open interval, return type is depend on the
+	   third argument: isLongType
+	   If the return type is Double, it is for 2 decimal */
+	public static Object randomNumber(double a, double b, boolean isLongType) {
+		long low = (long)a;
+		long high = (long)b;
+		long diff = high - low;
+		if(isLongType) {
+			return low + 1 + (long)(Math.random()*(diff-1));
+		} else {
+			double ans = low + (Math.random()*diff); 
+			return 1.0 * Math.round(100 * ans) / 100;
+		}
+	}
+
+}


wetneb (commented):
Hi @SustechcsKing, welcome to the project and thanks for this contribution!

This pull request adds a single class that is not used from anywhere else in the code base. So it will not have any influence on OpenRefine's behaviour as far as our users are concerned. Did you perhaps forget to commit changes to other files?

The goal of issue #3143 is to make improvements to the `randomNumber` GREL function, which is defined in `com.google.refine.expr.functions.math.RandomNumber`.

SustechcsKing:
> Hi @SustechcsKing, welcome to the project and thanks for this contribution!
> 
> This pull request adds a single class that is not used from anywhere else in the code base. So it will not have any influence on OpenRefine's behaviour as far as our users are concerned. Did you perhaps forget to commit changes to other files?
> 
> The goal of issue #3143 is to make improvements to the `randomNumber` GREL function, which is defined in `com.google.refine.expr.functions.math.RandomNumber`.

Thank you so much. I should add some functions in the existed com.google.refine.expr.functions.math.RandomNumber rather than create a new file and add some functions. I will modify them, thank you for your advising.

SustechcsKing:
Hi @wetneb , I want to ask a question. The newly added function is a new function, but no other code will invoke it. I did what tfmorris said, but I don't know where the function will be used, can I ask you for the suggestions? Thank you.

ostephens:
@SustechcsKing if you make the changes to `com.google.refine.expr.functions.math.RandomNumber` as suggested by @wetneb including the ability to interpret with no arguments or two arguments (as specified in #3143) then you can edit `com.google.refine.grel.ControlFunctionRegistry` to control how it is called in a GREL expression - see https://github.com/OpenRefine/OpenRefine/blob/577caabec1cce968df56156d7561a1ffde4b01c4/main/src/com/google/refine/grel/ControlFunctionRegistry.java#L305

SustechcsKing:
> The issue is about a GREL function, which is implemented in `com.google.refine.expr.functions.math.RandomNumber` as mentioned above. This is the function that could be improved.

Thank you very much.

SustechcsKing:
> @SustechcsKing if you make the changes to `com.google.refine.expr.functions.math.RandomNumber` as suggested by @wetneb including the ability to interpret with no arguments or two arguments (as specified in #3143) then you can edit `com.google.refine.grel.ControlFunctionRegistry` to control how it is called in a GREL expression - see
> 
> https://github.com/OpenRefine/OpenRefine/blob/577caabec1cce968df56156d7561a1ffde4b01c4/main/src/com/google/refine/grel/ControlFunctionRegistry.java#L305

Thank you very much.

SustechcsKing:
Hi, I made some changes in com.google.refine.expr.functions.math.RandomNumber.java, when parsing 3 parameters, which the third is the type, then the function can return a value of a certain type.

main/src/com/google/refine/expr/functions/math/RandomNumber.java
```diff
@@ -45,12 +45,28 @@
...
+            int high = ((Number) args[1]).intValue();
+            int randomNum = ThreadLocalRandom.current().nextInt(low, high);
+            if(args.length == 2){
+                return randomNum;
+            } else if (args.length == 3){
+                Object o3 = args[2];
+                if (o3 != null && o3 instanceof String) {
+                    String type = ((String ) o3).toLowerCase();
+                    if(type == "long"){
+                        return (long)randomNum;
```
wetneb, line 60: Here we are casting an int to a long, so the random value cannot actually go beyond the range allowed by `int`, right? Intuitively we should generate a `long` in the first place, I think.

main/src/com/google/refine/expr/functions/math/RandomNumber.java
```diff
@@ -45,12 +45,28 @@
...
+                    } else if(type == "double"){
+                        double randomDoubleNum = 1.0 * Math.round(100 * (low + Math.random()*(high - low))) / 100;
```
wetneb, line 62: So this will generate random floating-point numbers with not a lot of fractional digits… Perhaps we could use other random-number generating functions in Java to generate random values spread more uniformly in the desired interval?

wetneb (commented):
That's great! Thanks a lot, it seems functional already! I think we might be able to improve it a bit by picking the appropriate random number generator depending on the type, see below.

Also, it would be useful to update the documentation about the function. You can find the source for the docs at `docs/docs/manual/grelfunctions.md` (look for `randomNumber` there).

wetneb:
Hi @SustechcsKing, are you still interested in working on this?

PR #3860, opened 2021-04-29
Update suggest-api.md

VDK:
API endpoint moved from https://openrefine-wikidata.toolforge.org/ to https://wikidata.reconci.link/en/


wetneb (approved):
Thank you!

PR #3863, opened 2021-05-01
Cleanup technical reference, incorporating changes made on the wiki

wetneb:
(No corresponding issue, documentation PR)

Although we had migrated some content from the wiki to the technical reference section of the docusaurus docs, the wiki pages had not been marked as being replaced by the Docusaurus site, so they kept evolving independently. I have now blanked them with a pointer to the corresponding Docusaurus page.

I have also removed the reconciliation API docs from Docusaurus, with pointers to the W3C CG specs, since those are more up to date and there is no point keeping this obsolete content around.


wetneb:
Merging this too, because the wiki currently points to empty pages in the docs.

PR #3865, opened 2021-05-01
Adds windows-1252 to the list of common encodings

### LINKED ISSUE #3862: Add 'Windows-1252" to the Common Encodings List
It might sound a bit pretentious but I would suggest adding **windows-1252** to the list of common character encodings as well.
In my experience most of the documents I deal with are **UTF8** or **windows-1252**.

Also according to wikipedia "[It is the most-used single-byte character encoding in the world](https://en.wikipedia.org/wiki/Windows-1252)" :)

image

gitonthescene:
Fixes #3862

- Adds windows-1252 to the list of common encodings

### DIFF
--- a/main/webapp/modules/core/scripts/util/encoding.js
+++ b/main/webapp/modules/core/scripts/util/encoding.js
@@ -74,7 +74,7 @@ Encoding.selectEncoding = function(input, onDone) {
     });
   };
   generateEncodingList(elmts.commonList,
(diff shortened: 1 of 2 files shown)


wetneb (approved):
According to https://en.wikipedia.org/wiki/Windows-1252 it seems like it should be fairly widespread indeed! Fine with me.

PR #3887, opened 2021-05-08
Fix wikibase date export format

### LINKED ISSUE #3694: timestamp for Wikidata export: +2021-00-00 versus +2021-01-01
When exporting dates with a precision smaller than days, OpenRefine (Version 3.4.1) seems to export the year 1883 (as a random example) as +1883-01-01T00:00:00Z both with the export tool and QuickStatements.

However, if I understand those links correctly, it should be +1883-00-00T00:00:00Z (even though this isn’t ISO style):

https://www.wikidata.org/wiki/Help:Dates#Time_datatype
https://www.wikidata.org/wiki/Wikidata:Project_chat/Archive/2021/02#Dates_with_%3C11_precision
https://www.mediawiki.org/wiki/Wikibase/DataModel#Dates_and_times

While an argument could be made that Wikibase should normalize those entries, this isn’t the case at the moment, so I wondered if you could adjust OpenRefine’s behaviour. Thank you!

loulankxh:
Fixes #3694

- Adjust "extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java" and add a variable "segment" to judge whether the date has a precision smaller than year, month or day
- Adjust "extensions/wikidata/tests/src/org/openrefine/wikidata/schema/WbDateConstantTest.java" accordingly to test correctness.
- Note that the basic classes such as "Date" and "Calendar" are not modified since the authority is not assigned and I also think there isn't necessity to modify them because time like "1883-00-00T00:00:00Z" is actually not a real valid date.

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
@@ -124,8 +124,6 @@ public class WbDateConstant implements WbExpression<TimeValue> {
             bceFlag = true;
         }
 
-        int segment = trimmedDatestamp.split("-").length;  // judge whether the precision is smaller than year, month or day
-
         for (Entry<SimpleDateFormat, Integer> entry : acceptedFormats.entrySet()) {
             ParsePosition position = new ParsePosition(0);
             Date date = entry.getKey().parse(trimmedDatestamp, position);
@@ -163,8 +161,8 @@ public class WbDateConstant implements WbExpression<TimeValue> {
             calendar = Calendar.getInstance();
             calendar.setTime(bestDate);
             long year = calendar.get(Calendar.YEAR);
-            int month = segment > 1 ? calendar.get(Calendar.MONTH) + 1: 0;
-            int day_of_month = segment > 2 ? calendar.get(Calendar.DAY_OF_MONTH) : 0;
+            int month = precision < 10 ? 0 : calendar.get(Calendar.MONTH) + 1;
+            int day_of_month = precision < 11 ? 0 : calendar.get(Calendar.DAY_OF_MONTH);
             if(bceFlag)
                 year = -1*year;
             return Datamodel.makeTimeValue(year, (byte) month,
--- a/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
@@ -123,7 +123,9 @@ public class WbDateConstant implements WbExpression<TimeValue> {
             trimmedDatestamp = trimmedDatestamp.substring(1);
             bceFlag = true;
         }
-    	
+
+        int segment = trimmedDatestamp.split("-").length;  // judge whether the precision is smaller than year, month or day
+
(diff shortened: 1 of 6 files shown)


loulankxh:
Totally 5 test files in module wikidata are changed to adapt to the modification according to date precision

extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
```diff
@@ -123,7 +123,9 @@ public static TimeValue parse(String datestamp)
             trimmedDatestamp = trimmedDatestamp.substring(1);
             bceFlag = true;
         }
-    	
+
+        int segment = trimmedDatestamp.split("-").length;  // judge whether the precision is smaller than year, month or day
```
wetneb, line 127: I don't think we need to introduce this variable, we should be able to rely on the `precision` variable further down. That should be more reliable (because we could potentially introduce new date formats with more dashes in them).

extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
```diff
@@ -161,10 +163,12 @@ public static TimeValue parse(String datestamp)
             calendar = Calendar.getInstance();
             calendar.setTime(bestDate);
             long year = calendar.get(Calendar.YEAR);
+            int month = segment > 1 ? calendar.get(Calendar.MONTH) + 1: 0;
```
wetneb, line 166:
```suggestion
            int month = precision < 10 ? calendar.get(Calendar.MONTH) + 1: 0;
```

extensions/wikidata/src/org/openrefine/wikidata/schema/WbDateConstant.java
```diff
@@ -161,10 +163,12 @@ public static TimeValue parse(String datestamp)
...
+            int day_of_month = segment > 2 ? calendar.get(Calendar.DAY_OF_MONTH) : 0;
```
wetneb, line 167:
```suggestion
            int day_of_month = precision < 11 ? calendar.get(Calendar.DAY_OF_MONTH) : 0;
```

extensions/wikidata/tests/src/org/openrefine/wikidata/schema/WikibaseSchemaTest.java
```diff
@@ -158,6 +158,7 @@ public void testEvaluateRespectsFacets()
         List<ItemUpdate> expected = new ArrayList<>();
         ItemUpdate update1 = new ItemUpdateBuilder(qid1).addStatement(statement1).build();
         expected.add(update1);
+        System.out.println(expected);
```
wetneb, line 161: We can probably remove this `System.out.println` call.

wetneb (commented):
Thank you very much! It looks great to me. I just have a couple of minor comments below.

extensions/wikidata/tests/src/org/openrefine/wikidata/schema/WikibaseSchemaTest.java
loulankxh: Sorry， I forget to delete this

wetneb (approved):
Thank you very much!

loulankxh:
Thanks really!

PR #3900, opened 2021-05-11
Update Wikidata Toolkit to 0.12.0

wetneb:
This update is done manually rather than via dependabot because we were previously using an unofficial snapshot of this library.


wetneb:
Let's wait for the next release, I fixed the handling of units which is problematic for us in this new version.

PR #3911, opened 2021-05-13
(I#2104) HTTP headers when creating projects

### LINKED ISSUE #2104: Allow users to adjust HTTP headers (Authentication, User-Agent, Accept) when creating projects from Web Addresses (URLs)
When creating a project by getting the data from "Web Addresses (URLs)" it is currently not possible to change the nature of the request by altering the HTTP headers for Authentication, User-Agent, Accept. 
OpenRefine_Get_data_from_Web_Addresses

If there were such an option, it would be possible to load datasets directly from remote protected APIs or request data in a format that can be ingested by OpenRefine, instead of having to download it manually and importing the file.

In #218 the support of custom HTTP headers was added to the "add column by fetching URLs" dialog. The same fields should be available when loading a dataset directly from an URL.  
Open_Refine_Add_column_by_fetching_URLs

The alternative would be to download the datasets manually, store them as file and import them as "Get data from ... This Computer". For my specific usecase described below, it would also be possible to set up a proxy server that adds the Accept header to the request and returns the result.

**Additional context**
My specific use case for this feature is the possibility to load data directly from the Wikidata Query Service. For instance [this query](https://query.wikidata.org/sparql?query=SELECT%20DISTINCT%20%3Fcity%20%3FcityLabel%20%3Fcountry%20%3FcountryLabel%20%3Flocation%20%3Fpopulation%20%3Farea%20%3Finception%20%3Fisocode%20%3Felevation%0AWHERE%20%0A%7B%0A%20%20%3Fcity%20p%3AP31%2Fwdt%3AP279*%20%3Fstatement.%0A%20%20%3Fstatement%20ps%3AP31%20wd%3AQ5119.%0A%20%20FILTER%20NOT%20EXISTS%20%7B%20%3Fstatement%20pq%3AP582%20%3Fx%20%7D%0A%20%20%3Fcity%20wdt%3AP1376%20%3Fcountry.%0A%20%20%3Fcity%20wdt%3AP17%20%3Fcountry.%0A%20%20%3Fcity%20wdt%3AP625%20%3Flocation.%0A%20%20%3Fcity%20wdt%3AP1082%20%3Fpopulation.%0A%20%20%3Fcity%20wdt%3AP2046%20%3Farea.%0A%20%20OPTIONAL%7B%3Fcity%20wdt%3AP571%20%3Finception.%7D%0A%20%20OPTIONAL%7B%3Fcity%20wdt%3AP300%20%3Fisocode.%7D%0A%20%20OPTIONAL%7B%3Fcity%20wdt%3AP2044%20%3Felevation.%7D%0A%20%20SERVICE%20wikibase%3Alabel%20%7B%20bd%3AserviceParam%20wikibase%3Alanguage%20%22en%2Cen%22.%20%7D%0A%7DORDER%20BY%20DESC(%3Fpopulation) ) retrieves information about the worlds capitols. When requested as such in the browser (or via OpenRefine) it delivers the result set as XML. But when the "Accept" header is set to "text/csv" the data comes in a format the can be directly imported by OpenRefine. With the ever increasing interconnectedness between OpenRefine and Wikidata, I think that many others would profit from such a change as well.

sjathin:
Fixes #2104 

**This is a draft PR**

Would like to hear suggestions, is this the right way for implementing this? 

**UI:**
Screenshot 2021-05-13 at 17 33 02

**Case 1**
Screenshot 2021-05-13 at 17 33 14

Screenshot 2021-05-13 at 17 33 28

**Case 2**

Screenshot 2021-05-13 at 17 33 41

Screenshot 2021-05-13 at 17 33 55


main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -327,7 +329,11 @@ public String handleResponse(final ClassicHttpResponse response) throws IOExcept
                         };
 
                         HttpClient httpClient = new HttpClient();
-                        if (httpClient.getResponse(urlString, null, responseHandler) != null) {
+                        Header acceptHeader = new BasicHeader(HttpHeaders.ACCEPT, parameters.get("accept"));
```
ostephens, line 332: You shouldn't be assuming a pre-determined set of headers here. You need to rely either on what's been submitted from the UI or the supported headers currently registered (from HttpHeadersSupport)

ostephens:
Thanks @sjathin . I've not had time to fully review but one thing I did notice I've commented on above.

main/src/com/google/refine/commands/GetHttpHeadersCommand.java
```diff
@@ -0,0 +1,74 @@
+/*
+
+Copyright 2021, sjathin
```
kushthedude, line 3: OpenRefine

sjathin:
Made the necessary changes

Screenshot 2021-05-16 at 11 52 05
Screenshot 2021-05-16 at 11 52 20
Screenshot 2021-05-16 at 11 52 32
Screenshot 2021-05-16 at 11 52 47

main/src/com/google/refine/importing/ImportingUtilities.java
ostephens: This is still only using three specific headers (accept, user-agent and authorization). However, OpenRefine isn't limited to these three headers. Instead OpenRefine should support any headers that are registered via HttpHeadersSupport.

Therefor this could should either be relying on what's being submitted from the UI or work from the supported headers currently registered. So to do this you need to find all the headers that have been submitted/are supported here and then check if there is a value for them.

You can see how this works in https://github.com/OpenRefine/OpenRefine/blob/577caabec1cce968df56156d7561a1ffde4b01c4/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java#L142

sjathin: Please correct me if I am wrong, I think this is still the old commits, I push other commits as well, whether I have updated this - https://github.com/OpenRefine/OpenRefine/pull/3911/files

ostephens: @sjathin - ah - thank you. Can see the latest changes address this. Will re-review when I get a chance

sjathin:
Any updates on this PR?

wetneb:
@sjathin sorry for the delay, I did not see it because it is marked as a draft. When you want a PR to be reviewed, please press the "Ready for review" button to mark it as such.

wetneb:
I feel very bad for having left this pull request lingering for so long. Just looking at the screenshots, I wonder if we really want to duplicate the headers UI for each URL that the user adds? It feels a bit much to me, I would rather add it only once and apply those headers to all of the files we download. Those are very likely to come from the same source since they will end up in the same project.

wetneb:
Closing as stale. Sorry again that it fell through the cracks earlier!

PR #3916, opened 2021-05-17
Update the README.md for database extension

### LINKED ISSUE #3915: README.md for database extension is out of date
We need to update the database extension README.md to our more current architecture and documentation moves.

thadguidry:
Fixes #3915

- Updated README.md for database extension with current Maven architecture and documentation links.

### DIFF
--- a/extensions/database/README.md
+++ b/extensions/database/README.md
@@ -1,20 +1,9 @@
 This project is an OpenRefine extension for importing database data using JDBC.
 For exporting to a database, other code can be found under folder `OpenRefine/main/src/com/google/refine/exporters/sql`
 
-INSTALL
+## Adding support for other database vendors
 
-1. Before installing this extension download OpenRefine code from https://github.com/OpenRefine/OpenRefine 
-
-2. Pull this extension's code into folder `/database` under folder `/extensions`. 
-For more information on how to write OpenRefine extensions and where to put the files see https://docs.openrefine.org/technical-reference/writing-extensions#introduction
-
-The folder structure should resemble this:
-OpenRefine/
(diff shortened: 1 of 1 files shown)


wetneb (commented):
Thanks, it was indeed very outdated! Actually I think we can simplify those instructions even more since the extension is already maintained in the same repository, you don't need to pull both code bases separately or to edit any build file. So perhaps this file could just contain some description of the extension and instructions to add support for a new DB vendor? I don't think it is the right place to document how to install third-party extensions (since it is not one).

thadguidry:
@wetneb sure, simplify away to your hearts content.  or someone else can come back to this PR and improve it.

This is what I got from @tcbuzor back in the day...
1. You'll want to register an additional Database Service:  https://github.com/OpenRefine/OpenRefine/blob/master/extensions/database/src/com/google/refine/extension/database/DatabaseService.java
2. Provide the connection and service classes, look at the PostgreSQL one or MySQL one as examples: https://github.com/OpenRefine/OpenRefine/tree/master/extensions/database/src/com/google/refine/extension/database
3. Then wire up the interface with defaults as desired: https://github.com/OpenRefine/OpenRefine/blob/master/extensions/database/module/scripts/index/database-source-ui.js#L93
4. Add drivers manually to the classpath, or update the pom file to provide them as dependencies as other DB libraries are done: https://github.com/OpenRefine/OpenRefine/blob/master/extensions/database/pom.xml

wetneb:
Thanks, I just pasted that then!

thadguidry (commented):
Great, looks good!

PR #3920, opened 2021-05-20
Fix links in documentation

LaiWang2020:
Fixes #{3885}

Hi, I also encountered this problem and try to solve the problem according to my understanding. However, as I read the comments, I don't know if this problem is closed now. If it is closed, I'm sorry to waste your time to review and remind you that this problem should be closed.

### DIFF
--- a/docs/docs/manual/expressions.md
+++ b/docs/docs/manual/expressions.md
@@ -92,12 +92,12 @@ Do not use slashes to wrap regular expressions outside of a GREL expression.
 
 On the [GREL functions](grelfunctions) page, functions that support regex will indicate that with a “p” for “pattern.” The GREL functions that support regex are:
 *   [contains](grelfunctions#containss-sub-or-p)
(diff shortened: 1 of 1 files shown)


extensions/database/src/com/google/refine/extension/database/DatabaseConfiguration.java
```diff
@@ -42,8 +42,8 @@
     
     //optional parameters
     private boolean useSSL;
-    
-    
+
+
```
wetneb, line 46: Could you perhaps cancel this spurious and unrelated change?

wetneb (approved):
Thank you very much!

wetneb:
This replaces #3885

extensions/database/src/com/google/refine/extension/database/DatabaseConfiguration.java
LaiWang2020: Sure, sorry for that.

PR #3921, opened 2021-05-20
Support for "Store file source" in XML and JSON importers

### LINKED ISSUE #3462: XML and JSON importers and using "Store file source" doesn't make a column
### To Reproduce
1. Create new project from a file or URL  (either JSON or XML)
2. Choose JSON or XML as appropriate for file.
3. Then make any record row selection from an element in the JSON or XML with a mouse selection
4. Click on **Store file source** check box
5. Notice the preview refresh

### Current Results
Preview refreshes but doesn't add the extra column **File** as it does if you simply select another importer like Line based or CSV.

### Expected Behavior
**File** column should be created when clicking **Store file source** checkbox option on importers

### Versions
 - JRE or JDK Version: JDK8
 - OpenRefine: 3.5-beta [8eabec1] (master)

### Datasets
https://github.com/PerseusDL/treebank_data/raw/master/v2.1/Greek/texts/tlg0012.tlg001.perseus-grc1.tb.xml

Warpeas:
Fixes #3462

- Add file column when parse file with option "includeFileSources" true
- Change private function addArchiveColumn to protected
- Test on json and xml files

I'm not sure if the second change is allowable. Please leave comments.


wetneb (approved):
Thank you!

PR #3922, opened 2021-05-21
Cypress parallel manual

fgiroud:
Implemented parallelism for Cypress runs
A job execute before and generate a matrix of specs files, the cypress job is executed in parallel according to the matrix
Tests are not recorded if a secret is not provided


wetneb:
We are merging this as such because the workflow files have been modified, so we cannot test this without merging.

PR #3926, opened 2021-05-22
fixe rotation parsing in scatterplot facet

### LINKED ISSUE #3344: Rotate buttons are not working in scatterplot facet and dialog.
The rotate buttons in the scatterplot dialog and facet does not work.

### To Reproduce
1. First, create a create any scatterplot using scatterplotfacet.
Screenshot from 2020-11-19 22-28-50

2. Then, click on the rotate buttons given below the plot.

### Current Results
After clicking the rotate button the graph does not rotate.

Image with no rotation:
Screenshot from 2020-11-19 22-32-27

Image with rotation:
Screenshot from 2020-11-19 22-33-11

Same happens in the scatterplotfacet.

### Expected Behavior
The plot should rotate.

### Versions
 - Operating System: Ubuntu 20.04.1 
 - Browser Version: Chrome 19
 - OpenRefine:  OpenRefine 3.4.1

### Additional context
The problem is somewhere in the server side either when converting the client-end configurations of the plot into plotting parameters or somewhere else in between the client side and plotting of the graph.

Warpeas:
Fixes #3344

- Parse the rotation as string values
- Add two test to see if the rotation is set correctly

### DIFF
--- a/main/src/com/google/refine/commands/browsing/GetScatterplotCommand.java
+++ b/main/src/com/google/refine/commands/browsing/GetScatterplotCommand.java
@@ -65,110 +65,106 @@ import com.google.refine.util.ParsingUtilities;
 public class GetScatterplotCommand extends Command {
 
     final static Logger logger = LoggerFactory.getLogger("get-scatterplot_command");
-    
+
     @Override
     public void doGet(HttpServletRequest request, HttpServletResponse response)
             throws ServletException, IOException {
-        
+
         try {
             long start = System.currentTimeMillis();
-            
+
             Project project = getProject(request);
             Engine engine = getEngine(request, project);
             PlotterConfig conf = ParsingUtilities.mapper.readValue(
-            		request.getParameter("plotter"),
-            		PlotterConfig.class);
-            
+                    request.getParameter("plotter"),
+                    PlotterConfig.class);
+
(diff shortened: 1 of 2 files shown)


main/src/com/google/refine/commands/browsing/GetScatterplotCommand.java
```diff
@@ -145,7 +145,7 @@ public void setDimY(String dim) {
         @JsonProperty(ScatterplotFacet.ROTATION)
         public void setRotation(Object rotation) {
         	try {
-        		this.rotation = Integer.parseInt(rotation.toString());
+        		this.rotation = ScatterplotFacet.ScatterplotFacetConfig.getRotation(rotation.toString());
         	} catch(NumberFormatException e) {
```
wetneb, line 149: It looks like this NumberFormatException can no longer be thrown, so perhaps we could remove this `try`/`catch` block?

wetneb (approved):
Thank you very much! Just a tiny detail and it's good to go :)

main/src/com/google/refine/commands/browsing/GetScatterplotCommand.java
Warpeas: Sure. Then the process will be handled by getRotation().

PR #3927, opened 2021-05-22
Fix templating exporter parser to handle braces in expressions correctly

### LINKED ISSUE #3381: Broken escaping in templating export
Curly braces are used in the templating exporter to delimit expressions from the rest of the template text. Therefore if we want to use curly braces in the expressions they need to be escaped. According to the code, we are supposed to do this with `\{`, but that does not actually work.

### To Reproduce
1. First, open any project
2. Then, open the templating export dialog
3. Finally, input the following template:
```
{{"\}\}"}}
```

This is supposed to evaluate to just `}}` for each row.

### Current Results
We get `{{"\}\}"}}` instead.

### Expected Behavior
We should get `}}`

### Versions
 - Operating System: Linux
 - Browser Version: irrelevant
 - JRE or JDK Version: 11
 - OpenRefine: 3.4.1

### Additional context

Brought up while trying to document the templating exporter.

Warpeas:
Fixes #3381 

- When searching for the closeBrace, we match "}}" at once to avoid internal '}'
- A test to reproduce


main/src/com/google/refine/templating/Parser.java
```diff
@@ -81,7 +81,7 @@ static public Template parse(String s) throws ParsingException {
                     continue;
                 }
             } else if (c == '{' && c2 == '{') {
-                int closeBrace = s.indexOf('}', current + 2);
+                int closeBrace = s.indexOf("}}", current + 2);
                 if (closeBrace > current + 1 && closeBrace < s.length() - 1 && s.charAt(closeBrace + 1) == '}') {
```
wetneb, line 85: It looks like the last condition is no longer necessary:
```suggestion
                if (closeBrace > current + 1 && closeBrace < s.length() - 1) {
```

main/tests/server/src/com/google/refine/exporters/TemplatingExporterTests.java
```diff
@@ -125,6 +125,29 @@ public void exportEmptyTemplate(){
 
         Assert.assertEquals(writer.toString(), prefix + suffix);
     }
+
+    @Test
+    public void exportCurlyBracesTemplate(){
+        CreateGrid(1,1);
+        String template = "{{\"\\}\\}\"}}";
+//        template="{{jsonize(cells[\"_ - version\"].value)}}";
```
wetneb, line 133: I would remove this commented out code (and same below).

wetneb (commented):
Thank you very much, it works perfectly!

I added some minor comments on your changes.

In addition, I think it would be worth updating the documentation (in [docs/docs/manual/exporting.md](https://github.com/OpenRefine/OpenRefine/blob/master/docs/docs/manual/exporting.md)), which currently says:

> Note that some syntax is different in this tool than elsewhere in OpenRefine: a forward slash must be escaped with a backslash, while other characters do not need escaping. You cannot, at this time, include a closing curly brace (}) anywhere in your expression, or it will cause it to malfunction.

It would be nice to update this to explain how the escaping of expressions work.

main/src/com/google/refine/templating/Parser.java
Warpeas: Will the second one becomes not necessary too? As if there's no "}}", the closeBrace will have value -1. Otherwise, the closeBrace will always be s.length() - 1

wetneb: Good point!

docs/docs/manual/exporting.md
```diff
@@ -95,7 +95,7 @@ Anything that appears inside doubled curly braces ({{ }}) is treated as a GREL e
 ```
 
 :::caution
-Note that some syntax is different in this tool than elsewhere in OpenRefine: a forward slash must be escaped with a backslash, while other characters do not need escaping. You cannot, at this time, include a closing curly brace (}) anywhere in your expression, or it will cause it to malfunction.
+Note that some syntax is different in this tool than elsewhere in OpenRefine: a forward slash must be escaped with a backslash, while other characters do not need escaping. You cannot, at this time, include doubled closing curly brace (}}) anywhere in your expression, or it will cause it to malfunction.
```
wetneb, line 98: Now it should be possible to include two closing braces by escaping them, right? Using `\}\}` should work

Warpeas: Yes, but I'm not sure how to explain it 😂. Are there any other expressions that may be related to this problem?

wetneb: Intuitively I think it should only be braces that need escaping, I cannot think of anything else. But do check with the code :)

wetneb (commented):
I have had another look at this PR and I am not convinced we have got the code right. At the moment, we support escaping *outside* of dynamic fragments, not inside, so I am even confused as to why the test passes in the first place (perhaps because GREL accepts to translate `\{` to `{` in its expressions? but we should not pass it `\{` in the first place, just `{`).

Warpeas:
Sure. I will close this PR. Another shows a way to fix it. https://github.com/OpenRefine/OpenRefine/pull/4355 You may have a look.

PR #3928, opened 2021-05-22
Sustechcs king epoch to date

SustechcsKing:
Fixes #{608}

- Add a new function EpochToDate().
- Can parse one parameter, can change a number into the Date
- Can parse two parameters, the first is the number, the second is the type, such as second, millisecond, and microsecond.


wetneb (commented):
Hey, thanks for this pull request! It looks very good to me. I haven't tested it yet but I had a question just by having a brief look at the code. How should we handle the time zone in which the timestamp should be interpreted?

main/src/com/google/refine/expr/functions/EpochToDate.java
```diff
@@ -0,0 +1,114 @@
...
+import com.google.refine.grel.Function;
+import com.google.refine.util.ParsingUtilities;
+
+public class EpochToDate implements Function{
+
+    @Override
+    public Object call(Properties bindings, Object[] args) {
+        if(args.length >= 1 && args[0] instanceof Number) {
+            long epoch = (long) args[0];
+            ZoneId zoneId = ZoneId.of("America/Chicago");
```
wetneb, line 67: Intuitively I don't think this should be hard-coded :)

PR #3936, opened 2021-05-26
Add test case for my sql

duke326:
Fixes #{add some test case for the Mysql}

- I add some test case for MySQLServices and MySQLManagement
- Improve the test method coverage of both class to 100%

### DIFF
--- a/extensions/database/tests/src/com/google/refine/extension/database/DBExtensionTests.java
+++ b/extensions/database/tests/src/com/google/refine/extension/database/DBExtensionTests.java
@@ -43,7 +43,7 @@ public class DBExtensionTests {
     protected final String DEFAULT_MYSQL_HOST = "127.0.0.1";
     protected final String DEFAULT_MYSQL_PORT = "3306";
     protected final String DEFAULT_MYSQL_USER = "root";
-    protected final String DEFAULT_MYSQL_PASSWORD = "secret";
+    protected final String DEFAULT_MYSQL_PASSWORD = "as2188988";
     protected final String DEFAULT_MYSQL_DB_NAME = "testdb";
     
     protected final String PGSQL_DB_NAME = "postgresql";
--- a/extensions/database/tests/src/com/google/refine/extension/database/mysql/MySQLDatabaseServiceTest.java
+++ b/extensions/database/tests/src/com/google/refine/extension/database/mysql/MySQLDatabaseServiceTest.java
@@ -85,7 +85,7 @@ public class MySQLDatabaseServiceTest extends DBExtensionTests{
      }
      
      @Test
-     public void testExecuteQuery() throws DatabaseServiceException {
+     public void testTestExecuteQuery() throws DatabaseServiceException {
 
          MySQLDatabaseService pgSqlService = (MySQLDatabaseService) DatabaseService
                  .get(MySQLDatabaseService.DB_NAME);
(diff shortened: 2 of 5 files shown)


wetneb (commented):

It looks like you have duplicated the `testExecuteQuery` test case. I am not sure why we should do that?

duke326:
> Thank you for this pull request!
> 
> It looks like you have duplicated the `testExecuteQuery` test case. I am not sure why we should do that?

Hi Wetneb:

From my perspective, the old one is test method is `testExecuteQuery`. But actually test the `testQuery` method. The name of it makes me confused for a while. Then I separate them, to write 'testExecuteQuery' and 'testTestQuery' for both 'executeQuery' and 'testQuery' in the 'MySQLDatabaseServices'. Besides, I also write a test case for 'getType' method in the 'MySQLConnectionManager'. Now whole MySQL part's method coverage is 100%. I think this explanation can help

Duke

extensions/database/tests/src/com/google/refine/extension/database/mysql/MySQLDatabaseServiceTest.java
```diff
@@ -85,7 +85,7 @@ public void testConnect() throws DatabaseServiceException {
      }
      
      @Test
-     public void testExecuteQuery() throws DatabaseServiceException {
+     public void testTestExecuteQuery() throws DatabaseServiceException {
```
wetneb, line 88:
```suggestion
     public void testTestQuery() throws DatabaseServiceException {
```

extensions/database/tests/src/com/google/refine/extension/database/DBExtensionTests.java
```diff
@@ -43,7 +43,7 @@
     protected final String DEFAULT_MYSQL_HOST = "127.0.0.1";
     protected final String DEFAULT_MYSQL_PORT = "3306";
     protected final String DEFAULT_MYSQL_USER = "root";
-    protected final String DEFAULT_MYSQL_PASSWORD = "secret";
+    protected final String DEFAULT_MYSQL_PASSWORD = "as2188988";
```
wetneb, line 46: I don't think we should change this. You should be able to change the db password in the xml settings: `extensions/database/tests/conf/tests.xml` (and not commit that file).

wetneb (commented):
Ah ok, that makes sense! How about changing the name slightly then? See below.

extensions/database/tests/src/com/google/refine/extension/database/DBExtensionTests.java
duke326: Thank you ! I will revise it a little bit

duke326: > I don't think we should change this. You should be able to change the db password in the xml settings: `extensions/database/tests/conf/tests.xml` (and not commit that file).

By the way, during I write the test case, I find that change configuration in the `extensions/database/tests/conf/tests.xml` cannot finish the MySQL configuration correctly. I am not sure about that. I will do more test later

extensions/database/tests/src/com/google/refine/extension/database/DBExtensionTests.java
```diff
@@ -43,7 +43,7 @@
     protected final String DEFAULT_MYSQL_HOST = "127.0.0.1";
     protected final String DEFAULT_MYSQL_PORT = "3306";
     protected final String DEFAULT_MYSQL_USER = "root";
-    protected final String DEFAULT_MYSQL_PASSWORD = "secret";
+    protected final String DEFAULT_MYSQL_PASSWORD = "";
```
wetneb, line 46: We should probably keep this and use the XML files to change the password when needed.

PR #3937, opened 2021-05-26
Jython test improve

LaiWang2020:
Fixes #{add new test case for Jython Model}

- add new test case for unwarp method in JythonEvaluable Class
- improve tested methods coverage from 55% to 66%


wetneb (commented):

I had a quick look and it is not really clear to me why you want to do here. It looks like you have duplicated the `evaluate` method and I am not sure why. Can you tell a bit more about the motivation?

LaiWang2020:
> Thank you for this pull request!
> 
> I had a quick look and it is not really clear to me why you want to do here. It looks like you have duplicated the `evaluate` method and I am not sure why. Can you tell a bit more about the motivation?

Hi wetneb,
From my perspective, in the original version there is two unwarp method in JythonEvaluable Class. One returns the Object and anthor one returns PyObject. Unwarp method that returns PyObject is never tested before and I think it might be used in the future. In that case, I wrote the test case called pyObjectTest to test this method and evaluatePyobjectTest method in JythonEvaluable Encapsulated this method. Hope this explanation would help.
Lai

wetneb:
Ok, I think I understand your approach better now, thanks!

As a general principle, if we want to improve the test coverage, we should write test cases which execute uncovered areas, ideally without artificially changing the classes being tested for the sake of testing. So:
1. Either the `unwrap` method you want to test is currently reachable in some use cases (so, in some calls to the existing `evaluate` method), in which case we should just write a test case which reproduces this situation;
2. Or the `unwrap` method you want to test is never actually used. In that case, it is dead code and should be removed.

In both cases there should not be a need to introduce this new method `evaluatePyobjectTest`.
Can you see in which of the two scenarios above we are?

LaiWang2020:
> Ok, I think I understand your approach better now, thanks!
> 
> As a general principle, if we want to improve the test coverage, we should write test cases which execute uncovered areas, ideally without artificially changing the classes being tested for the sake of testing. So:
> 
> 1. Either the `unwrap` method you want to test is currently reachable in some use cases (so, in some calls to the existing `evaluate` method), in which case we should just write a test case which reproduces this situation;
> 2. Or the `unwrap` method you want to test is never actually used. In that case, it is dead code and should be removed.
> 
> In both cases there should not be a need to introduce this new method `evaluatePyobjectTest`.
> Can you see in which of the two scenarios above we are?

Hi wetneb,
I think unwrap method I want test is currently reachable in some calls to the existing evaluate method and I agree with what you said about artificially changing in test class. I will try to move this method and reproduces this situation with a new test case.

wetneb (commented):
Hi @LaiWang2020,
That is better, but I think it should be possible to avoid re-implementing all the logic in a new method `evaluatePyobjectTest`. Could you try to write a test case which only calls the existing `evaluate` method?
For instance you could try using a Python evaluable which would return a nested value containing various datatypes. I haven't checked but something like `return [true, 13, {"a":3}]` might work?

LaiWang2020:
> Python evaluable

Hi wetneb,
I think I'm kind confused about the Python evaluable. I try to produce the situation and only call the existing evaluate method with `return [true, 13, {"a":3}]`, but it didn't go the PyObject breanch.

wetneb:
Perhaps you actually need to return an instance of a class:

```python
class Foo(object):
      pass

return Foo()
```

LaiWang2020:
> Perhaps you actually need to return an instance of a class:
> 
> ```python
> class Foo(object):
>       pass
> 
> return Foo()
> ```

Yeah, it works, thank you. I did have a detour at first submission. Sorry for that.

extensions/jython/tests/src/com/google/refine/jython/JythonEvaluableTest.java
```diff
@@ -16,6 +21,31 @@
...
+    props.put("false", "false");
+    props.put("rowIndex", "0");
+    props.put("value", 1);
+    props.put("project", project);
+    props.put("call", "number");
+    props.put("PI", "3.141592654");
+    props.put("cells", new CellTuple(project, row));
+    String funcExpression = "class Foo(object):\n      pass\n\nreturn Foo()";
+    JythonEvaluable eval1 = new JythonEvaluable(funcExpression);
+    eval1.evaluate(props);
```
wetneb, line 46: Perhaps it would be worth doing an assertion on the result of the evaluation? So that we make sure we don't return just *anything*?
For instance, you could check the value of a field on the object:
```python
class Foo(object):
    bar = 1

return Foo()
```
and then check that you can access the `foo` attribute via the `HasAttr` interface (not sure about the name, only from memory).

LaiWang2020: Yeah, I think that would be cool

wetneb (approved):
Thank you!

PR #3939, opened 2021-05-26
fixed options to add to run OpenRefine in headless mode

Shatakshi0805:
Fixes #{3818}

- modified the option to add to run OpenRefine in headless mode from -Drefine.headless=true to -x refine.headless=true in refine folder.
- 
-

### DIFF
--- a/refine
+++ b/refine
@@ -511,7 +511,7 @@ ui_test() {
 
     REFINE_DATA_DIR="${TMPDIR:=/tmp}/openrefine-tests"
     
-    add_option "-Drefine.headless=true"
+    add_option "-x refine.headless=true"
     add_option "-Drefine.autoreload=false"
     add_option "-Dbutterfly.autoreload=false"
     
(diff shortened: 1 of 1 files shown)


wetneb:
Thanks for this pull request! Have you tested it? I don't think it will work :)

Shatakshi0805:
Apologies for the late reply. Also, I have not tested it. I have been trying to set up my environment for Open Refine but every time I build open refine in the Command line by typing `./refine clean` or `./refine build` then I get -
`'.' is not recognized as an internal or external command,` 
`operable program or batch file.`
I also checked multiple times if I had set the environment variable wrong for JAVA_HOME and also tried to change the JAVA_HOME to point to different location as I was able to observe that there are different Java-jdk locations but nothing worked. Could you please help?@wetneb

wetneb:
The message `'.' is not recognized as an internal or external command, operable program or batch file.` seems to indicate that you are on Windows, and so you should use `refine.bat build` instead of `./refine build`.

Shatakshi0805:
Thank you for helping. I am able to build and run the tests to.

Issue #3929, opened 2021-05-22
splitByLengths does not return the array of substrings split but only first substring


thadguidry:
The description of the function `splitByLengths` states clearly that:
>Returns the array of strings obtained by splitting s into substrings with the given lengths
but it never returns split substrings in an array, but instead always a single element array of the first split substring.

### To Reproduce
1. Use custom text transform on a column of text
2. Paste in the GREL test:  `"University of Washington".splitByLengths(4)`
3. Notice that only the first substring at length 4 is returned in the array, and no other substrings from the rest of the max length of string

### Current Results
`"University of Washington".splitByLengths(4)`
 outputs
`[ "Univ" ]`

### Expected Behavior
`[ "Univ", "ersi", "ty o", "f Wa", "shin", "gton"]`

### Screenshots
![image](https://user-images.githubusercontent.com/986438/119238977-88421a80-bb0b-11eb-9bdc-28d50b2cb9a5.png)

### Versions
 - Operating System: Windows 10
 - Browser Version: Firefox
 - JRE or JDK Version: JDK8
 - OpenRefine: 3.4.1 win-with-java

nfazlay:
If you look at the definition [here](https://github.com/OpenRefine/OpenRefine/blob/master/docs/docs/manual/grelfunctions.md#splitbylengthss-n1-n2-), and the example that follows, my understanding is that the number of items in the array would be equal to the number of parameters you pass. Hence, to get your desired output, your function call should look like this:

```java
"University of Washington".splitByLengths(4, 4, 4, 4, 4, 4)
```

Calling `"University of Washington".splitByLengths(4)` splits the first 4 characters from the string and discards the rest.

## Examples

---

  ### input
 `"University of Washington".splitByLengths(4)`
 ### output
 `[ "Univ" ]`

---


  ### input
  `"University of Washington".splitByLengths(4, 6)`
  ### output
  `[ "Univ", "ersity"]`

---

 ### input
 `"University of Washington".splitByLengths(4, 4, 4)`
 ### output
 `[ "Univ", "ersi", "ty o"]`

thadguidry:
@nfazlay Thanks, then let's improve the documentation descriptions in our code which will automatically show in the HELP tab in the Expression editor dialog.  We'll also need to update our docs section for the function.

thadguidry:
I think a more useful enhancement then to meet my expectations would be to:
[...]

PR #3940, opened 2021-05-26
Added splitByLength() between indices function to grelfunctions documentation

### LINKED ISSUE #3934: Update user manual and in-app documentation for splitByLength()
The documentation, specifically for splitByLength() does not inform the user that they have the possibility to only return values within a given range of indices, Iv since updated the document but wanted to make a bug report on it for submitting the PR.

ex: 
`"123456789012345678901234567890".splitByLengths(3,4,5)[0,2]`
`returns [ "123",  "4567" ]`

More about this issue can be seen in this thread: https://github.com/OpenRefine/OpenRefine/issues/3929

mgutierrez98:
Fixes #3934}

Added splitByLength() between indices function to grelfunctions documentation 
- 
- 
-

### DIFF
--- a/docs/docs/manual/grelfunctions.md
+++ b/docs/docs/manual/grelfunctions.md
@@ -174,9 +174,9 @@ Returns a string converted to a number. Will attempt to convert other formats in
 
 Returns the array of strings obtained by splitting s by sep. The separator can be either a string or a regex pattern. For example, `"fire, water, earth, air".split(",")` returns an array of 4 strings: [ "fire", " water", " earth", " air" ]. Note that the space characters are retained but the separator is removed. If you include “true” for the preserveTokens boolean, empty segments are preserved.
 
-###### splitByLengths(s, n1, n2, ...)
+###### splitByLengths(s, n1, n2, ...)[i1,i2] 
 
-Returns the array of strings obtained by splitting s into substrings with the given lengths. For example, `"internationalization".splitByLengths(5, 6, 3)` returns an array of 3 strings: [ "inter", "nation", "ali" ]. Excess characters are discarded.
+Returns the array of strings, between i1 and (i2 - 1), obtained by splitting s into substrings with the given lengths. For example, `"internationalization".splitByLengths(5, 6, 3)[0,2]` returns an array of 2 strings: [ "inter", "nation" ]. "ali" is desgraded as it falls outside of the indices. Excess characters are also disregarded.
 
 ###### smartSplit(s, s or p sep (optional))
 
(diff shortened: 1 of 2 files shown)


wetneb:
Thank you! You will first need to merge your changes with master.

mgutierrez98:
@wetneb I should probably add to the original splitByLength document as opposed to a new line as such: 
splitByLengths(3,4,5)[0,2]
as for functionality, the program does currently split if indices are specified which is lacking in the documentation. 
I will however attempt to toss in what you stated as well for functionality when the source string is shorter than the slice parameters

docs/docs/manual/grelfunctions.md
```diff
@@ -174,9 +174,9 @@ Returns a string converted to a number. Will attempt to convert other formats in
 
 Returns the array of strings obtained by splitting s by sep. The separator can be either a string or a regex pattern. For example, `"fire, water, earth, air".split(",")` returns an array of 4 strings: [ "fire", " water", " earth", " air" ]. Note that the space characters are retained but the separator is removed. If you include “true” for the preserveTokens boolean, empty segments are preserved.
 
-###### splitByLengths(s, n1, n2, ...)
+###### splitByLengths(s, n1, n2, ...)[i1,i2] 
 
-Returns the array of strings obtained by splitting s into substrings with the given lengths. For example, `"internationalization".splitByLengths(5, 6, 3)` returns an array of 3 strings: [ "inter", "nation", "ali" ]. Excess characters are discarded.
+Returns the array of strings, between i1 and (i2 - 1), obtained by splitting s into substrings with the given lengths. For example, `"internationalization".splitByLengths(5, 6, 3)[0,2]` returns an array of 2 strings: [ "inter", "nation" ]. "ali" is desgraded as it falls outside of the indices. Excess characters are also disregarded.
```
ostephens, line 179: I think the original misunderstanding reported in #3929 was that the user was expecting something like:

"University of Washington".splitByLengths(4)

to return
["Univ","ersity of Washington"]

but it actually returns
["Univ"]

I think what is needed here is just to make this even more explicit - so expanding on the existing comment "Excess characters are discarded" to make it extra clear - so taking what you have already written something like:

"The remaining characters (in this case "zation") are discarded and not included in the array output"

ostephens (changes requested):
@thadguidry  had proposed improving the documentation that appears in the OpenRefine in-app "Help" - which is text that comes from the code rather than the documentation pages. You'll need to look in https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/expr/functions/strings/SplitByLengths.java

It currently reads: "Returns the array of strings obtained by splitting s into substrings with the given lengths"

PR #3942, opened 2021-05-27
Fix one test case in wikidata

samaritanhu:
Fixes #{one failed test case in wikidata}

When running the test cases, I found there was one test case that failed and 342 passed.

I went to the test case and tried to get access to the website: http://not.exist/w/api.php . When getting access to this website, the information returned is null as well. Therefore, this test case failed for a reason.

-  To make it right, I changed the “de” to “null”, which would pass this test case.

### DIFF
--- a/extensions/wikidata/tests/src/org/openrefine/wikidata/schema/WbLanguageConstantTest.java
+++ b/extensions/wikidata/tests/src/org/openrefine/wikidata/schema/WbLanguageConstantTest.java
@@ -57,6 +57,6 @@ public class WbLanguageConstantTest extends WbExpressionTest<String> {
 
     @Test
     public void testFallbackLangCodes() {
-        assertEquals("de", WbLanguageConstant.normalizeLanguageCode("de", "http://not.exist/w/api.php"));
+        assertEquals(null, WbLanguageConstant.normalizeLanguageCode("de", "http://not.exist/w/api.php"));
     }
 }


wetneb:
Hi @samaritanhu, thank you for this contribution!
I cannot see the image you posted (it looks like you added as a link instead of an image).
I am not sure why this test is failing on your side. The new version of the test seems to be failing on the continuous integration:
https://github.com/OpenRefine/OpenRefine/pull/3942/checks?check_run_id=2682238500

This might be caused by the specifics of your own internet access (proxy? DNS resolution?). It might be worth debugging on your side to see if there is anything we should fix in OpenRefine.

nalzok:
> I went to the test case and tried to get access to the website: http://not.exist/w/api.php . When getting access to this website, the information returned is null as well. Therefore, this test case failed for a reason.

I think the URL `http://not.exist/w/api.php` is intended to be inaccessible, and the test in question is verifying if a fallback language code is available in such cases. It looks like that test failed on your machine for some other reason, but we need more details to be certain. Could you re-post the screenshot, or even better, copy & paste the error message for us?

samaritanhu:
I don't know if you can see the pictures from here: 

When running the test cases, I found there was one test case that failed and 342 passed.
https://sm.ms/image/W5VA6agJ8OY1Hcs

When getting access to this website, the information returned is null as well.
https://sm.ms/image/JREqGjzxmT9FQyU

Thanks for answering my pr so quickly!

nalzok:
Cool, I can see the images now. To transcribe, the error you are getting is

```
java.lang.AssertionError:
Expected :null
Actual   :de
```

with the offending line being

https://github.com/OpenRefine/OpenRefine/blob/73fc36de727c521831f64def3120798050e708d3/extensions/wikidata/tests/src/org/openrefine/wikidata/schema/WbLanguageConstantTest.java#L60

The first parameter to `java.lang.AssertionError` is named `expected`, and the error message indicates you are expecting `null`. Are you sure you are passing `"de"` as the first parameter to `assertEquals` when getting the error? In other words, have you changed the code while running the tests?

Also, it might help to rebuild the project and run the tests again. IDEA occasionally have cache invalidation problems.

nalzok:
You may also want to post the entire test log for all tests (something like [this one](https://github.com/OpenRefine/OpenRefine/pull/3942/checks?check_run_id=2682238500)). Logs contain much more information than a screenshot. For example, we can tell `http://not.exist` is deliberately a dead link simply by look at [this line](
https://github.com/OpenRefine/OpenRefine/pull/3942/checks?check_run_id=2682238500#step:8:699).

wetneb:
I am closing this PR since it seems to come from a problem in the local development setup rather than an issue with the test's assertions.

PR #3945, opened 2021-05-27
Templating exporter should have clear error reporting

### LINKED ISSUE #3382: Templating exporter should have clear error reporting
Currently when your code in the Templating exporter has a problem, the Preview window will just show "Updating..." forever. 
It would be good to have a timeout after a few seconds and report back the line/character causing the problem. If not that, then just a generic error message would be an improvement.

When your code (inside double curly braces) isn't resolving correctly (e.g. because of an extra closing curly brace) the entire expression including the curly braces just outputs as a string instead. 
This should have its own error message, instead of the string output. 

templating

sjathin:
Fixes #3382

- Adding error reporting
- Adding unit tests for the Templating exporter 
- Resolve all json errors

**Testing:
TC1:**
Screenshot 2021-05-27 at 21 29 25

**TC2:**
Screenshot 2021-05-27 at 21 29 35

**TC3:**
 
Screenshot 2021-05-27 at 21 29 44

**TC4:**

Screenshot 2021-05-27 at 21 29 59

**TC5:**
Screenshot 2021-05-27 at 21 30 13


sjathin:
@allanaaa - It would help me, if you could provide me some more syntaxes that can be used in the templating exporter. I think adding a few more test cases will make sure this doesn't break.

wetneb (commented):
Hey, this is really great, thank you for all these test cases!

main/src/com/google/refine/templating/Parser.java
```diff
@@ -86,13 +95,21 @@ static public Template parse(String s) throws ParsingException {
...
+                    	}
                         fragments.add(new StaticFragment(s.substring(start, current)));
                     }
                     start = current = closeBrace + 2;
 
+                    Evaluable exprEval = MetaParser.parse(expression);
+                    
+                    if (exprEval.evaluate(bindings) == null) {
+                    	throw new ParsingException("Missing or bad template");
+                    }
```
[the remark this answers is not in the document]
sjathin: Oh that's right, my bad. Yes ```MetaParser.parse``` checks the invalid functions. Will remove the null check

main/src/com/google/refine/exporters/TemplatingExporter.java
```diff
@@ -82,6 +84,7 @@ protected TemplateConfig(
 
     @Override
     public void export(Project project, Properties options, Engine engine, Writer writer) throws IOException {
+        Properties bindings = ExpressionUtils.createBindings(project);
```
wetneb, line 87: This should not be needed anymore.

main/src/com/google/refine/exporters/TemplatingExporter.java
```diff
@@ -94,7 +97,7 @@ public void export(Project project, Properties options, Engine engine, Writer wr
         
         Template template;
         try {
-            template = Parser.parse(templateString);
+            template = Parser.parse(templateString, bindings);
```
wetneb, line 100: (same here)

main/src/com/google/refine/templating/Parser.java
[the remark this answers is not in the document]
sjathin: I just checked that why I added this statement, the ```MetaParser.parse``` gives an exception for 
1.  ```jso}nize(cells["URN"].value)```
2. ```jsonnize(cells["URN"].value)``` 

But does **not** give an exception for **```jso{nize(cells["URN"].value)```**. I suspect this is an error with the Grel parser, as it considers just ```jso``` as the expression.

main/src/com/google/refine/grel/Parser.java
```diff
@@ -248,7 +248,9 @@ protected Evaluable parseFactor() throws ParsingException {
         }
 
         while (_token != null) {
-            if (_token.type == TokenType.Operator && _token.text.equals(".")) {
+        	if (_token.type == TokenType.Error) {
```
sjathin, line 251: This change is for checking the error: ```jso{nize(cells["URN"].value)```

wetneb: Perhaps it is worth adding a test for this in `GrelTests`, since it is independent from the templating exporter?

main/src/com/google/refine/templating/Parser.java
```diff
@@ -86,6 +92,9 @@ static public Template parse(String s) throws ParsingException {
                     String expression = s.substring(current + 2, closeBrace);
 
                     if (current > start) {
+                    	if (!s.substring(start, current).matches("\\s*,?\\s*\\\"*[a-zA-Z0-9() ]*\\\"\\s*:\\s*")) {
```
wetneb, line 95: It is not clear to me which case this is catching, could you expand a bit, maybe with a comment?

sjathin: I see, this is to test the keys of the json, i.e, ```"URL"```, ```"LA (name)"```. Will add a comment here :)

wetneb: I am not sure what JSON has to do with the context here - the templating exporter can be used to generate JSON, but it can be used to generate anything else as well, so we should definitely not build in anything specific to JSON in this parser as far as I can tell. But maybe I missed something?

sjathin: I see, by json I mean the template keys. For instance the keys - ```"URL" :``` and ```"LA (name)" :```. Currently, I have added a regex expression such that it would allow only alphabets and parentheses between the double-quotes. My question is should we allow ```"URL{" :``` this, i.e, curly brackets in between the double-quotes. And should we restrict only to alphabets and parentheses?

wetneb: There is nothing special about "URL" or "LA (name)" in the test cases, they are just some substrings in the static parts of the template. There is no reason to treat them specially.
For instance, a template such as `{ "UR{N": {{jsonize(cells["URN"].value)}}, "foo:"bar }` would be totally legitimate. It would not generate valid JSON but the templating exporter does not know anything about the expected format of what it is generating. All it knows is that a template should be a series of static and dynamic fragments. The static fragments can be arbitrary strings (once the braces have been unescaped). The dynamic fragments must be valid expressions but those expressions can generate arbitrary strings too.

sjathin:
@wetneb Based on the discussion, I have added a regex to test the keys, which allows ```{ "UR{N": {{jsonize(cells["URN"].value)}}, "foo:"bar }```. And throws an error for all invalid syntaxes.

main/src/com/google/refine/templating/Parser.java
```diff
@@ -86,6 +92,10 @@ static public Template parse(String s) throws ParsingException {
                     String expression = s.substring(current + 2, closeBrace);
 
                     if (current > start) {
+                    	// Regex to match invalid keys of the template exporter 
+                    	if (!s.substring(start, current).matches("^\\s*,?\\s*\".*\"\\s*:\\s*$")) {
+                    		throw new ParsingException("Missing or bad template");
+                    	}
```
wetneb, line 98: I still don't understand what you mean by "keys". There is simply no notion of keys in the templating exporter. I think this `if` statement should be removed entirely.

sjathin: I guess ```StaticFragment``` is a better way, this If statement ensures that the ```StaticFragment``` is valid, a ```StaticFragment``` here which is a ```substring(start, current)```, the valid values can be:
1. ``` { "foo":```
2. ``` { "fo}o":```
3. ```, { "foo":```

This if statement tests for the following invalid values:
1. ``` { {"foo":```
2. ``` { "foo" { :```
3. ``` bar "foo"  : ```

wetneb: The three values that you consider invalid are actually valid.

sjathin: Oh I see, I might have misunderstood the concept, will remove the regex test, then it would add the ```StaticFragment``` as it is. Althought bit curious, if for instance, let choose:
{ ABC"URN": {{jsonize(cells["URN"].value)}} }

This will generate a JSON:
```
{
    "rows" : [
        {
              ABC"URN": 23434
        },
        {
              ABC"URN": 1111
        }
    ]
}
```

which is an incorrect JSON.

wetneb:
@sjathin I don't know how I could explain the issue better, really. The templating exporter can be used for *any* text-based format.
For instance you could use it to generate a CSV, some XML, some BibTeX files, or anything else really!
See for instance this template which generates some MODS file: https://gist.github.com/sallain/7604ffb0c155294fcfaf
Or this example for another XML format: https://blogs.bl.uk/digital-scholarship/2015/12/using-open-refine-to-create-xml-records-for-wikimedia-batch-upload-tool.html
Or this example for Bibtex files: https://github.com/caltechlibrary/open_refine_bibtex/blob/master/row_template.txt

For this reason, we should not assume anything about the format of the text in static fragments. Is that clearer?

sjathin:
Yes, I get it now, much more clearer now, thank you for explaining.

wetneb:
In the interest of getting this merged, I have removed your changes to the templating parser itself, since they were based on a misunderstanding of the sort of templates that were acceptable or not.

wetneb (approved):

sjathin:
Thanks, @wetneb for completing the PR.

PR #3946, opened 2021-05-27
Wiki create process test

chupengrocky:
Noticing that there are no test cases for the "PerformEditsProcess" class in the wiki extension, I try to create a simple test case to test the constructor of "PerformEditsProcess" by calling the "createProcess" method in PerformWikibaseEditsOperation. I plan to further test the process with threading simulation if this approach is acceptable. 

add "testCreateProcess" method in PerformWikibaseEditsOperationTest.java

### DIFF
--- a/extensions/wikidata/tests/src/org/openrefine/wikidata/operations/PerformWikibaseEditsOperationTest.java
+++ b/extensions/wikidata/tests/src/org/openrefine/wikidata/operations/PerformWikibaseEditsOperationTest.java
@@ -28,6 +28,7 @@ import static org.testng.Assert.assertEquals;
 import java.io.LineNumberReader;
 import java.util.Properties;
 
+import com.google.refine.process.Process;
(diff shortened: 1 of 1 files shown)


wetneb:
Hi @chupengrocky, thank you very much for this pull request!

First, you can safely ignore the `prepare_ui_test_matrix` failing check, it is unrelated (we are working on it).

Concerning your new test, it tackles an area that is indeed not so easy to test (especially the `run` method)!
I wonder if it could be possible to add some assertions on the `Process` instance returned by `createProcess`? For instance, you could check the value of some fields (since they are `protected` they should be visible from this test class).

wetneb:
Hi @chupengrocky, do you still intend to work on this?

chupengrocky:
Yes, I am still working on this. Sorry about the late update:
https://github.com/OpenRefine/OpenRefine/pull/4221

PR #3949, opened 2021-05-31
Update Homebrew documentation

joewiz:
As of Homebrew 2.6.0:

> All brew cask commands have been deprecated in favour of brew commands (with --cask) when necessary. 

See https://brew.sh/2020/12/01/homebrew-2.6.0/.

Also, removed mention of openrefine-dev cask as this facility does not appear to be in use.

### DIFF
--- a/docs/versioned_docs/version-3.4/manual/installing.md
+++ b/docs/versioned_docs/version-3.4/manual/installing.md
@@ -111,30 +111,29 @@ Once you have downloaded the `.dmg` file, open it and drag the OpenRefine icon o
 The quick version:
 
 1. Install [Homebrew](http://brew.sh)
-2. In Terminal enter ` brew cask install openrefine`
+2. In Terminal enter `brew install openrefine`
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thank you @joewiz!
This reminds me that we should migrate https://github.com/OpenRefine/OpenRefine/wiki/Maintaining-OpenRefine's-Homebrew-Cask to https://docs.openrefine.org/technical-reference/homebrew-cask-process, I don't know if you would feel like it…

joewiz:
@wetneb I can take a crack at it. I'm a little [out-of-practice](https://github.com/Homebrew/homebrew-cask/commits/c40b4b69f08f3453eab6973dcfd0f07fe1e03383/Casks/openrefine.rb), my last update being in 2018 - but I'll look into what may need to be updated. Would you like those updates in this same PR?

wetneb:
As you see fit! This is already a documentation PR so I don't think it would inappropriate to add further documentation changes in the same go :)

PR #3394, opened 2020-12-08
Add search/filter to project list page

mayralgr:
Closes #2749 

- Add searchbar
- Modify styles to match
- Icon from https://www.iconfinder.com/search/?q=search&price=free

Current state

openrefine

PR #3966, opened 2021-06-07
Issue 2749 add search/filter to project list page

duke326:
Fixes #{2749}

- Inspired from mayralgr's  PR (https://github.com/OpenRefine/OpenRefine/pull/3394)
- Find and Fix bugs in mayralgr's code. In `show`/`hide`/`filterSearch`/`filterSearchList` function. doesn't unbind/off the function. which will cause method call exponential after 10-15 times execution. And lead to the "Maximum call stack size exceeded" error.
- Fix code use `off()` before `on()` for both `hide`/`show` function. And use `one()` for` filterSearch`/`filterSearchList` function. Solve the problem.
- Use `keyup` to do the dynamic search without press the `enter`


wetneb:
Hi @duke326, thanks for this PR!

It seems that your PR also contains unrelated changes, so you probably need to rework your git setup to avoid those.
Also, it could be nice to start off the branch that #3394 proposed, so that the authorship of those commits is still represented in the git history (but I think that is not formally required).

duke326:
> Hi @duke326, thanks for this PR!
> 
> It seems that your PR also contains unrelated changes, so you probably need to rework your git setup to avoid those.
> Also, it could be nice to start off the branch that #3394 proposed, so that the authorship of those commits is still represented in the git history (but I think that is not formally required).

Sorry seems there really are some problems in my git settings. 

I see you add some comment for the https://github.com/OpenRefine/OpenRefine/pull/3394 said you want to achieve the dynamic search. And during I implement `keyup()` I find out that bug. Seems no one noticed that and I think it might be a very serious problem. So I decide to fix that and create this new PR to help revise this function. The reason why I create new one because I am not sure if mayralgr still work for it. If I start off at that branch on her PR it might cause some unnecessary problem. But if you want me to do that. I can do that later. 

Thank you for your reply! I am really happy I can contribute to this project.

duke326:
> Hi @duke326, thanks for this PR!
> 
> It seems that your PR also contains unrelated changes, so you probably need to rework your git setup to avoid those.
> Also, it could be nice to start off the branch that #3394 proposed, so that the authorship of those commits is still represented in the git history (but I think that is not formally required).

Finally I finish erase unnecessary database file. Now you are free to check the new code!

wetneb:
Sincere apologies for having left this lingering for so long. It is not clear to me why I have not looked at it earlier (perhaps the failing build status?).

Here are the outstanding problems I can see on this PR:
- trying it out interactively, I cannot actually figure out how to search in my projects list, I do not recognize where I should type in my query
- OpenRefine opens in the "Open project" tab by default, whereas it normally opens at the "Create project" by default
- there are some leftover `console.log` and commented out code that should probably go

Also there are some minor unrelated changes but those are clearly fixes so I have no problem with leaving them in this PR.

wetneb:
I would now close this PR because of the problems. Again, many apologies for having left it without a review for so long. I am clearing the backlog of PRs to prevent this from happening again in the future.

PR #3974, opened 2021-06-10
Remove lavalamp js dependency, non-free.

### LINKED ISSUE #3957: jquery.lavalamp.min.js is non-free
[Posted by Markus Koschany on the openrefine-dev mailing list](https://groups.google.com/g/openrefine-dev/c/eJbPPYnqJiw):
> 
> Lavalamp is licensed under BY-NC 3.0 which makes it technically non-free and it
> cannot be included in Debian main because of the non-commercial clause. If I
> remove the file obviously some of the functionality of the main startpage is
> removed. I believe it should be possible to work around it or replace the
> plugin with a similar effect. In any case I had to remove the file from the
> sources for now.

This is a blocker for OpenRefine's packaging for Debian. Therefore we should investigate as soon as possible how to get rid of this dependency.

wetneb:
Closes #3957.

As a user this only removes the animation in project tags, but the same selection capabilities are still provided.


wetneb:
@thadguidry do you want to check that the tags panel is still functional like that?

thadguidry:
@wetneb Tested in Firefox and Chrome, works fine.
image

PR #4005, opened 2021-06-22
Respect spaces in clustering dialog.

### LINKED ISSUE #4004: Clustering dialog does not show duplicate whitespace clearly
When clustering values which contain duplicate whitespace characters, these characters are not clearly shown in the clustering dialog, giving the impression that we are clustering identical values.

### To Reproduce
1. First, import the following file (as a lines-based or CSV file for instance):
```
foo
b ar
b  ar
b   ar
```
2. Then, open the clustering dialog on that column.

### Current Results
All options look the same:
image

### Expected Behavior
Space differences should be respected:
image

### Versions
 - Operating System: Linux
 - Browser Version: Firefox 78
 - JRE or JDK Version: 11
 - OpenRefine: 3.5 beta1

wetneb:
Fixes #4004

- replace spaces by non-breakable spaces in clustering values to avoid them looking identical when they only differ by spaces


thadguidry:
This will be useful, and this can help later with visibility if/when we add in for support for showing hidden characters as in #1286

Issue #3890, opened 2021-05-09
Make the user manual translatable


wetneb:
We want to make it possible for translators to translate our English documentation into other languages.

### Proposed solution

Docusaurus has a i18n feature which can be used for that. It can be used with various providers. We could try it out with Crowdin, as they are often used together.

### Alternatives considered

* Re-using Weblate (which we already use to translate the interface), but that requires inserting a complicated workflow based on po4a to split the markdown files in translatable strings. This probably gives a not so great translator experience since translators will translate small paragraphs in isolation, without the context.
* Using GitLocalize, but that seems not so stable and reliable.

### Additional context

This was discussed on the mailing list: https://groups.google.com/g/openrefine-dev/c/aesx9AiswIY.
Issue #3888 is a prerequisite for this.

thadguidry:
@wetneb From the Crowdin.com FAQ at the bottom of their [Pricing page](https://crowdin.com/pricing#annual), it seems you as the Admin (or maybe even @joehand ) first need to do this, Open Source Request
> Do you have plans for open-source projects and educational institutions?
>
> Yes. If you want to use Crowdin for an Open Source project, sign up for a free account, set up your project and send us a request. Apply for an Academic License if your project has educational purposes. Each granted license will include an unlimited number of projects, strings, and members.
>
> [Open Source Request](https://crowdin.com/page/open-source-project-setup-request)
> [Academic License Request](https://crowdin.com/page/academic-license-project-setup-request) 

In the meantime, I'm working on my forked branch in PR #4018 against my Trial license for CrowdIn over next 14 days.  Just to get comfortable with their platform, the `crowdin.yml` file, and ensure I have all our Docusaurus config worked out properly for CrowdIn. (Since Kathy my wife is Chinese, I will be trying it out with a few beginning sentences on a few of our pages)

Especially important I just found out today is understanding about "hiding duplicate strings" feature documented here: https://docusaurus.io/docs/i18n/crowdin#docs-versioning

PR #4018, opened 2021-06-28
Support for versioning nav and translations

### LINKED ISSUE #2288: Create branches with granular versions of OpenRefine we support
We will need to support both 4.x and 3.x versions of OpenRefine in the future and be able to handle Security Patches and perhaps using GitHub's Security Policy doc as I mentioned in PR #2048 (now closed) with granular semantic versions (major.minor.patch)

@wetneb had suggested previously in that PR that we should probably do the following:
> Given our current workflows we are basically not able to release fixes for any previous version. We would need to use branches for the major versions we want to support on the long term. That's definitely something we can consider. Each security patch would be merged in all supported major release branches to create the corresponding versions. That would probably mean using more granular release numbers (such as 3.2.0).

thadguidry:
Fixes #2288 #3890

- Add nav handling for Versions
- Add nav handling for Translations
- Need to decide on Archived Versions `versionsArchived.json` and how to handle them, perhaps like Docusaurus docs themselves through Netlify? Or do we actually store the older versions in our doc repo path so that anyone checking out can see the older docs that we still support?
- Need to decide on Crowdin for translations (I think YES), and then wire up our own project path on Crowdin so that the "Help Us Translate" link on the i18N nav dropdown works correctly.  Docusaurus currently makes that link go right to https://github.com/facebook/docusaurus/issues/3526


thadguidry:
Discussed with @wetneb and we'll work against what was merged in #4029 and add onto it in other PR's instead of this one.

PR #4024, opened 2021-06-29
Update info on Docs README about Next version

thadguidry:
@wetneb Does this look OK?  I just tested and seems that this is how we can actually work and see the "next version" of docs and we didn't really document this anywhere as far as I can tell.  So this README section is a good place?

Fixes #{issue number here}

- Update Docs README about Next version
- 
-

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -26,7 +26,14 @@ yarn
 yarn start
 ```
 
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Yep it looks good, I agree the versioning made that workflow a bit more confusing (you need to remember to add the `/next` prefix)

PR #4025, opened 2021-06-30
Rough draft for use of official accounts (Twitter)

thadguidry:
We'll probably need several passes on this until we come to something that is not so wordy and effectively shows or mimics how @magdmartin has so generously been handling our official tweeting and his practices!
My section that I think hits to the core of the issue that @wetneb brought up on list and provoked adding language into our Governance is in section `### Use of official OpenRefine accounts`

Jess @ CS&S provided their section `#### Tweeting`, which I have quickly, roughly cleaned up just so we could glance over it here in review and pick what we like, and remove what we don't need.

- [ ] It will be good to come to some agreement on a few points that Jess @ CS&S had put into theirs (like avoiding politics, etc.)
- [ ] The rest of her section kinda flows in and out of "how to properly tweet with OpenRefine's account", which I think can be pulled out, or just cleaned up, or alternatively put into another doc that is not Governance related.
- [ ] Maybe we want others like Sandra or Alicia or whomever from Steering committee to also help us a bit here?

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -127,3 +127,73 @@ Combined with CS&S, the Steering and Advisory committees have experience applyin
 ### Other assets 
 
 CS&S owns and manages the domain openrefine.org 
+
+### Use of official OpenRefine accounts
+
+1. Use of any official OpenRefine accounts (social media, funding, banking, code repository, domain, etc.) for the purpose of promotion of another entity must always be discussed prior with the Advisory committee or optionally Steering committee.
+2. "promotion" includes any and all types such as funding, crowdfunding, partnerships, internships.
+
+Any language that gives the impression that the OpenRefine organization approves, authorizes or otherwise supports a project, person or company is not permissible without written authorization from the Advisory Committee or optionally Steering committee.
+
+### Tweeting
+
+General guidelines for OpenRefine admins using our official Twitter account and looking to represent an organization or project ("the entity"). 
+
+#### Setup
+
+- Add a photo, ideally a logo of the entity that is readable.
+- Add a background photo or pick a color that looks good with the logo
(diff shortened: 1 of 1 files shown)


wetneb (commented):
Thanks! To me it feels a little bit over-complicated - it is not clear to me we need such a strong-handed policy for this.

As a follow up to this Gitcoin episode I would rather add just a few sentences, which would not be focusing on Twitter, but rather emphasizing that starting new crowdfunding campaigns, funding applications or other partnerships should be coordinated in the advisory committee first.

thadguidry:
@wetneb Reading and thinking through my paragraph drafted and your feedback thus far.
Hmm, I think anyone in the community, if they want to start a crowdfunding campaign to help OpenRefine could do so, no? I mean, we can't really stop them or stop them from impersonating our official org.  Criminals will be criminals.

So this Governance is about those with access of our official accounts and using them for ANY funding campaign, yes?  I mean, the only bounds that I overstepped in your mind was that I didn't communicate enough with the whole advisory committee on using our official Twitter, yes?

wetneb:
Ok my comment was not so clear: I would say that "starting new crowdfunding campaigns, funding applications or other partnerships **in the name of the project** should be coordinated in the advisory committee first."

Using the official Twitter account is one way of doing things "in the name of the project".

thadguidry:
@magdmartin What are your thoughts also now after this last revision I made?  Looking better? Good enough for now?

thadguidry:
@magdmartin Hi Martin, I had a followup reminder... asking again for your thoughts on the changes now and if we have agreement and further thoughts on full majority needed or not?

wetneb:
Closing since we agree we can make progress on this topic in other ways.

PR #4029, opened 2021-07-02
3890 crowdin

### LINKED ISSUE #3890: Make the user manual translatable

wetneb:
Fixes #3890.

This wires up Crowdin to make the docs translatable.
As with most PRs which change deployment / CI, I expect we'll need some follow-up tweaks to this PR.

For now I have enabled Japanese and French as first locales, so that @isaomatsunami and I can try out the integration before trying to recruit translators in other languages.

At the moment this will only translate the current development version, not the latest stable release, so the translated pages will be deployed at URLs such as https://docs.openrefine.org/fr/next/ and not just https://docs.openrefine.org/fr/. When we cut out the next stable version (3.5) this would release the translations at that point.


thadguidry:
@wetneb Should I delete my #4018 draft, I guess? probably right?

wetneb:
It seems to be doing things that I didn't do, but I cannot tell exactly what! If you are happy with the current result, then I guess another PR is not needed…

thadguidry:
@wetneb yeah, let's build on what you have merged now. If we need some of the custom variable declarations (that I borrowed from Docusaurus's own CI/CD), we can always add them in as necessary.  I'll delete my PR.

PR #4042, opened 2021-07-07
Speed-up testing time for Cypress in CI/CD

fgiroud:
Added an environment variable to disable projects cleanup on CI/CD.
This way we save an extra API call after each test run.
It will also fix [a flaky scenario](https://dashboard.cypress.io/projects/s5du3k/runs/1369/test-results/7f66c317-4292-4347-a6de-eed8e0397d4a) where the delete token wasn't found


fgiroud:
Well, the gain might not be that big after all, but it's still a good parameter to have
It's also fixing the flaky test mentionned, so let's keep it even if it doesn't really change the run duration

wetneb (approved):

kushthedude (approved):

PR #4062, opened 2021-07-17
Add mention of adding Row numbers into column

thadguidry:


wetneb (commented):
This is sort of a "recipe", so maybe it should rather go to the wiki?
Otherwise, I would rather add this sort of comment in the docs about facetting/browsing: as a user I would likely look for that there (how to select a consecutive set of rows).

thadguidry:
Yes, I felt similarly, It's important info to know.  Hmm, yeah Faceting might be a better location for it.

thadguidry:
@wetneb OK, I think this small info tip now under Custom Numeric Facet is good enough.

Issue #4106, opened 2021-08-17
Open Refine opens with logo only when running on Java 16


P3rCo:
### To Reproduce
1. Click on open refine Application

### Current Results
2. This window appears:
3. 
![image](https://user-images.githubusercontent.com/89087309/129763348-5ea6f617-03d8-43fc-889d-2b9cec64d4b1.png)
4. This page appears
5. 
![image](https://user-images.githubusercontent.com/89087309/129763413-9f71ea36-5a20-4a3e-b079-eb1fe1934e4e.png)
6.I have installed the embedded Java version:
![image](https://user-images.githubusercontent.com/89087309/129763476-9d5707ff-98e0-4f98-af53-fb295a7247cb.png)
7.I have run it as adminstrator same results.

8. Tried to run the batch file
9. Same results: 
![image](https://user-images.githubusercontent.com/89087309/129764156-be788373-3b43-4ff6-bff2-afed0f7151ce.png)
10. Tried firefox
11. 
![image](https://user-images.githubusercontent.com/89087309/129764321-5b18991d-5330-4e2b-a56f-fc32072fa1e1.png)

### Expected Behavior
Open Refine page appears

### Versions
 - Operating System: Edition	Windows 10 Pro
Version	20H2
Installed on	‎17/‎03/‎2021
OS build	19042.1165
Experience	Windows Feature Experience Pack 120.2212.3530.0

 - Browser Version: 
![image](https://user-images.githubusercontent.com/89087309/129764189-7a54e277-0aed-4ea9-b043-52e03b0faa47.png)

 - JRE or JDK Version: 
 - Using embedded but:
 - 
![image](https://user-images.githubusercontent.com/89087309/129764400-6afd092a-659b-4e4f-a1d2-1df0e1c1f1d7.png)

 - OpenRefine: 3.4.1

P3rCo:
Thanks a lot for any help guys, really stumped.

orothad:
Do you still have an environment variable in Windows set for `JAVA_HOME` ?
If so, can you try temporarily removing that and trying to launch OpenRefine again ?

P3rCo:
![image](https://user-images.githubusercontent.com/89087309/129797839-e0348bf5-2a16-41e6-8776-7615f057c37d.png)
Same result
![image](https://user-images.githubusercontent.com/89087309/129797873-a45487a5-0f53-4da4-9fe6-0a5111debf0a.png)
System variables (there was a java home, deleted it)
[...]

PR #4135, opened 2021-08-31
mention it requires java 14

VladimirAlexiev:
... and doesn't work with Java 16 or later versions (see issue #4106).

This PR is a duplicate of #4134: I wasn't sure whether I should edit `docs` or `versioned_docs`, so edited both.


wetneb (approved):
Thank you very much! It does make sense to edit both. We can adjust the text for version 3.4 once 3.5 is released.

PR #4149, opened 2021-09-15
Fix closing of the project metadata dialog for large project metadata values

### LINKED ISSUE #4119: Add scrollable div layer to Project metadata view to prevent cutoff of CLOSE button
Currently if values in the Project metadata editor are too large then the CLOSE button is not seen and clickable to users.

### Current Results
If the length of values is too long in any of the keys, then sometimes the CLOSE button will be cutoff and hidden from a user preventing them from being able to close the Project metadata panel.

### Expected Behavior
Provide 2 things:
1. a scrollable div layer that can hold and scroll through all the key:value rows.
2. while scrolling, keep the CLOSE button at the bottom, and the Key and Value header labeling at the top.

### Screenshots
image

### Versions
 - Browser Version: Firefox
 - JRE or JDK Version: JDK15
 - OpenRefine: 3.5 beta1

### Additional context
Using [Cypress support for Visual testing](https://docs.cypress.io/guides/tooling/visual-testing#Functional-vs-visual-testing), ensure to add 1 Cypress test that can 
1. click edit on Title row and add lots of text into value input box and click OK.
2. using a previously cropped baseline image of the lower left corner that includes the CLOSE button.
3. compare that cropped baseline image with a clip region of the screen during test, such as example (adjust numbers as necessary)
```cy.get('.element')
  .toMatchImageSnapshot({
    clip: { x: 125, y: 870, width: 75, height: 50 },
  });
  ```
4. assert that the CLOSE button is still displayed and seen

rachittiwari8562:
Fixes #4119

1.  Added a div element around table to restrict the height of table to 70 view-port height .
2. Made header of table "sticky" using CSS--"position:sticky".
3. Previously the Dialog-Box was centred using height but the value once set would not change when data is changed is table . So to overcome that used the dynamic positioning using percentage positions and transform .
4. Added "cypress-image-snapshot" visual test for the checking close button .


rachittiwari8562:
The visual test is failing by 6-7% image difference . In my local system there is no error .  Is it possible to look what is the difference in pictures . I am attaching original image and screenshot of test .
Screenshot from 2021-09-15 20-39-41

close-button snap

wetneb:
Thanks a lot @rachittiwari8562! I think visual tests tend to be quite unreliable indeed, I would rather avoid them for cases like this. Perhaps it would be sufficient to check that the overall height of the dialog does not exceed some reasonable threshold?

rachittiwari8562:
So should I comment out/ delete the visual test and commit ?

wetneb:
I would remove the test indeed, don't worry about your commits in this branch, I will squash them into one when merging this PR.
Also, since you are starting to work on multiple issues, I would recommend you don't open a PR from the master branch of your fork, but instead create one branch per issue you are working on (it should make your life a bit easier).

wetneb:
@rachittiwari8562 do you plan to work on this further?

wetneb:
Closing this per inactivity, but happy to reopen it if you feel like pushing further on this.

wetneb:
Ah sorry, that is a misunderstanding: I generally do not edit pull requests myself before merging them.
Somehow I cannot reopen this PR because your fork no longer exists, apparently.

rachittiwari8562:
Okay .....may be sometime in future I will solve this issue , I will contact when I will be ready to contribute..

PR #4156, opened 2021-09-16
Removed Microsoft Edge for CI/CD tests, #4150

### LINKED ISSUE #4150: Cypress build stability on PRs / master branch
It seems to me that Cypress tests are much more stable when run from a pull request compared to when run on the master branch. I have the impression that very often, the build succeeds on the pull request, and then when I merge it in master, the build fails.

One notable difference between the two workflows is that in pull requests, we test for Chrome only, and on master we test for both Chrome and Edge. So, it's natural that the master build is more likely to fail.

With those quasi-systematic failures I have learned to disregard failures on the master branch and only pay attention to build statuses on pull requests, which is not ideal. So I would be interested in bringing the CI on the master in line with that for pull requests, so disabling Edge there.

I would be curious to see if it solves the problem or if there are more differences between the two environments that could explain this pattern. Perhaps it's also a bias on my side (when a test fails on a pull request, it's easy to trigger a rebuild by merging master into the branch, which I cannot do on master).

fgiroud:
Fixes #4150
Removed Microsoft Edge for CI/CD tests

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -132,16 +132,6 @@ jobs:
         with:
           node-version: '12'
 
-      - name: Install Edge
-        if: matrix.browser == 'edge'
-        run: |
-          sudo curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg
-          sudo install -o root -g root -m 644 microsoft.gpg /etc/apt/trusted.gpg.d/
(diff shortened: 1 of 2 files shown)


fgiroud:
Damn! not even passing on chrome, I think it's completely unrelated, and doesn't prevent merging, disabling edge tests can only improve the failure rate

wetneb (approved):

wetneb:
Let's see what happens!

fgiroud:
Created #4157 to investigate

PR #4163, opened 2021-09-17
Improved stability of Cypress tests

### LINKED ISSUE #4157: Cypress tests for undo/redo are unstable
See the following PR : https://github.com/OpenRefine/OpenRefine/runs/3629571950?check_suite_focus=true
The undo redo tests quite often fails
Need to check if this is a regression or something else

fgiroud:
Fixes #4157
I am also extending the scope of this PR to reduce the flakiness rate of the "pagination" tests
@kushthedude I've refactored some tests you've written about pagination, because it seems like duplicated code to me. 
But I've refactored them mostly because they were flaky, and there is a re-rendering in the UI that makes it almost impossible to test properly, hence me adding a 'wait' time.
Let me know if you're ok with that.


main/tests/cypress/cypress.json
```diff
@@ -2,12 +2,11 @@
   "integrationFolder": "./cypress/integration",
   "nodeVersion": "system",
   "retries": {
-    "runMode": 3,
+    "runMode": 4,
```
kushthedude, line 5: Isnt this too much to retry like a test is buggy if it is failing for 3 times, dont you think so 😅 ?

main/tests/cypress/cypress/integration/project/grid/viewpanel-header/pagination.spec.js
```diff
@@ -1,111 +1,45 @@
...
   });
-  it('Ensure it redirects to next-page when pagesize is 50', function () {
+
+  it('Test the "next" button', function () {
     cy.loadAndVisitProject('food.small');
 
     cy.get('.viewpanel-pagesize').find('a').contains('50').click();
+    cy.get('.viewpanel-header').should('be.visible');
+    cy.get('.data-table-container').should('be.visible');
+    cy.wait(250); // eslint-disable-line
```
kushthedude, line 29: Can we not increase the timeout rather having a explicit `cy.wait` 🤔 ?

main/tests/cypress/cypress/integration/project/undo_redo/apply.spec.js
```diff
@@ -2,14 +2,18 @@ describe(__filename, function () {
...
-      .click();
-
+    // Check some columns before the test
     cy.get('table.data-table thead th[title="Shrt_Desc"]').should('exist');
     cy.get('table.data-table thead th[title="Water"]').should('exist');
 
+
+    // find the "apply" button
+    cy.get('#or-proj-undoRedo').click();
+    cy.wait(500); // eslint-disable-line
```
kushthedude, line 12: Same as above for the explicit wait

kushthedude (commented):
Some general comments @fgiroud , regarding the duplication yes they were actually just checking the proper display size with each button in case if someone breaks the display for 25 size, but I guess its okay to remove.

fgiroud:
@kushthedude I agree to all of your comments, but here is the thing
The pagination test can't be done properly, I have tried many times. Reason is that OpenRefine re-render ".data-table-container", so in many cases, the element gets detached from the dom. And I can't find anything else in the dom to be checked.
Adding wait + retry was the only way for me to prevent this test to create false positives
So, it's either we disable this test to avoid this ugly 'wait' + increase of cypress retry. Or we want to have this test but I am a bit out of options here 😅 

If you want to join the meeting tomorrow we can discuss

fgiroud:
Hi @kushthedude I think you're right, if a test is too flaky raising retries is a bad idea
Then I suggest to remove the pagination tests as we can't make them stable
Undo/redo can be kept, as I think that's an important feature
What's your opinion ?

kushthedude:
>Hi @kushthedude I think you're right, if a test is too flaky raising retries is a bad idea
Then I suggest to remove the pagination tests as we can't make them stable
Undo/redo can be kept, as I think that's an important feature
What's your opinion ?

I think we can follow this, rather than removing I suggest to comment them out with a code comment to explain the issue, what do you say?

main/tests/cypress/cypress/integration/project/undo_redo/apply.spec.js
fgiroud: We do want to keep this one

fgiroud:
I think I found an acceptable compromise @kushthedude .
If testing a different number of results per page makes the tests flaky, then I removed changing the page number from all the tests. So next,previous,first,last tests just use the default pagination number (10), and it works fine
We were able to keep most of the tests you've written
And not a single flaky test in 5 runs
Pr is good to go IMO

kushthedude (approved):

PR #4172, opened 2021-09-23
normalize casing in UI text in English

riteshsoni123:
Fixes #{3363}

- Make some modification in  previous commits.
- Normalize the phrases that was suggested by wetneb.
- Normalize phrases that was missed earlier. eg: line 643 "Replace smart quotes with ascii" to "Replace smart quotes with ASCII".


riteshsoni123:
@wetneb can you review the changes that I have made?? I don't know why some of the checks are not successful. Please help me solve it if you can.

wetneb:
Hi @riteshsoni123, thank you very much for this fantastic PR! I had a look at your changes and I agree with every single of them, yay!
The tests are indeed failing, but that will not be hard to fix. If you click on the "Details" link in the failing test, you get to https://github.com/OpenRefine/OpenRefine/pull/4172/checks?check_run_id=3690188285, where you can see that a test fails because it was looking for a "Create Project" button which never appeared. That is normal, because you have changed it to "Create project". So we just need to update this test (and potentially others in similar ways). You can just search for the "Create Project" string in the `main/tests/cypress` directory and normalize the case of all occurrences you find there.
To test your changes you should be able to run the UI tests locally, see https://docs.openrefine.org/technical-reference/functional-tests

riteshsoni123:
@wetneb thanks for guiding me on solving the check errors that I was getting. I have made the changes that you asked for and it has passed all the checks. Can you review it??

wetneb (approved):

wetneb:
Thank you very much @riteshsoni123!

PR #4173, opened 2021-09-23
eclipse launch file path fix

tejasrbhat:


tejasrbhat:
@wetneb I was trying to debug from eclipse and couldn't launch. I did these changes and able to launch now. I raised PR for the same. If this is intentionally left or if there is a better approach, let me know. I will close the PR.

wetneb:
Huh, I didn't know we still had this file lying around here! I think it should ideally be deleted altogether since we just recommend using the Maven integration of IDEs instead (import the project as a Maven project). What do you think? If using Maven would not have been possible for you and this file was useful, then I am also happy with keeping it.

tejasrbhat:
@wetneb It requires VM arguments right ? I tried to run it directly, but looks like I need to set somethings manually. This file will help for sure.

wetneb:
Ok! It would still be preferable to just use `mvn exec:java` I think, but if this file was useful to you, fine with me!

PR #4184, opened 2021-09-28
Adds parent() GREL function for jsoup

### LINKED ISSUE #4181: Add new GREL function to support Jsoup parent() function to complement our select()
There is no easy way to navigate up the parse tree provided by jsoup selector syntax.  We currently can navigate down very easily with Jsoup's selector syntax `value.parseHtml().select("div > p")` and also `value.parseHtml().select("div:has(@) ~ div")` but 
this does not directly allow DOM traversal itself.  However, Jsoup conveniently provides DOM traversal for finding a https://jsoup.org/apidocs/org/jsoup/nodes/Element.html#parent()
which mimics the behavior from BeautifulSoup's https://www.crummy.com/software/BeautifulSoup/bs4/doc/#parent

### Proposed solution
Add a new GREL function for the jsoup method `parent()` and for simplicity sake just call the GREL function `parent()`

### Alternatives considered
A difficult path of custom installed Jython 2.7 with BeautifulSoup

### Additional context
Existing code for `select()` functionality is here:
https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/expr/functions/xml/SelectXml.java
and function registration to be added to is here:
https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/grel/ControlFunctionRegistry.java#L249

thadguidry:
fixes #4181

- Adds new GREL function `parent()` for jsoup for DOM traversal of parent nodes.  Example:  `value.parseHtml().select("p[type*=none]")[0].parent()`


thadguidry:
@wetneb Could use your help on the test assertEquals.  For some reason it kept failing for me even locally and I'm thinking it's some LF or CRLF issue in the expected formatting of the string?  I tried several variations  and reduced it to the simplest test I could think of but still the string shows differently slightly and don't know how to fix in the test.

tejasrbhat:
@thadguidry I think `Jsoup.parse(h).select("p[type*=child]").first())`**`.toString()`** works.

As I am new here, I have a question. In cases like these, is it good to comment the solution and let the creator of the PR do the changes, or if we know the fix, can we go ahead and do the changes in the same PR ourselves ?

wetneb:
@tejasrbhat thanks a lot for chiming in!
I think it is polite to ask the PR author whether they are happy with you adding commits to their branch (in general they are fine with it). Failing that it makes sense to do another PR. Looking forward to your collaboration!

tejasrbhat:
@wetneb Got it. Thanks for the clarification.

thadguidry:
@tejasrbhat Perfect, yes `toString` is what I was missing.

PR #4188, opened 2021-09-30
Added the test for labels in columns

### LINKED ISSUE #3427: Add UX tests for column / View
Test the Column / View dropdowns

Screen Shot 2020-12-16 at 14 13 59

### Proposed solution

Add four tests,
* collapse.spec.js
* collapse_all.spec.js
* collapse_left.spec.js
* collapse_right.spec.js

The test should basically collapse one or more columns, and ensure that the columns labels are empty (or possibly just hidden).

### Additional context

See also #3415, with some very similar tests about collapsing
When collapsing, column labels are not hidden, they are removed from the DOM.

riteshsoni123:
Fixes #3427

- Add four tests (collapse.spec.js, collapse_all.spec.js, collapse_left.spec.js, collapse_right.spec.js).
- Check if the labels in columns are empty empty when collapsed or not.


riteshsoni123:
@fgiroud my pull request is still not reviewed yet. Is there any problem with it? If yes then please inform me.

wetneb:
Hi @riteshsoni123, we do our best to review PRs swiftly but it generally takes longer than a few days unfortunately. Thanks for your PR and your patience :)

riteshsoni123:
yes I understand and sorry for being impatient.

main/tests/cypress/cypress/integration/project/grid/all-column/collapse.spec.js
```diff
@@ -0,0 +1,12 @@
+describe(__filename, function () {
+  it('it collapses all columns', function () {
+    cy.loadAndVisitProject('food.mini');
+    cy.columnActionClick('NDB_No', ['View', 'Collapse this column']);
+
+    cy.get('.odd td:nth-child(4)').should('to.contain', '');
+
+    cy.get('.even td:nth-child(4)').should('to.contain', '');
+
+    cy.get('tr').children().first().next().should('to.contain', '');
```
fgiroud, line 10: What is the purpose of this assertion ?

fgiroud:
Hi @riteshsoni123 , one question, overall, what is the purpose of all the assertions with the `next()` functions ?

Few comments 
- collapse_all contains copy-pasted code from view.spec.js, or it looks like they are doing the same thing
It's a good idea to move them to a dedicated file, but then you also want to create expand.spec.js to eventually completely delete view.spec.js

- `all-column` is the folder that contains tests related to the special column "All" on the left of the grid header.
Your tests should live in `column/view`

But except those few minor things, the tests looks good to me

riteshsoni123:
Hello @fgiroud thanks for pointing out things that can be improved in my PR.

These are the changes that I think will solve the problem.

- As I am new to cypress so I used the first thing that I found (next() function) to select. Instead of it what I can do is to use statement like "cy.get('[title="NDB_No"]').should('to.contain', '');" which is much cleaner and readable.
- I totally agree that performing the same tests in more than one file is not a good idea. So what I can do is to remove the extra tests from the new file that has already been present previously.
- As you mentioned that all-column folder contains tests only related to special column. So I can move all the test files that are created by me to the column\view.

I think these changes will solve the previous problem. If you have any more suggestions then do tell. I will be making all these changes after getting green signal from you.

fgiroud:
Yes, totally make sense to check for the column header to contains ''

Something you can do to improve your test is to load a project with a dynamic fixture.
Explanation: food.mini is a small dataset for testing purpose, but it still has a cost in terms of mem/cpu. What you can do instead is to specify the dataset you would like to have as a pure json

Example : 
```javascript
const fixture = [
      ['a', 'b', 'c'],

      ['0a', '0b', '0c'],
      ['1a', '1b', '1c']
    ];

    cy.loadAndVisitProject(fixture);
 ```
 This will create a project, with 3 columns (a,b,c), and two rows.
 
 Regarding point 2,  so you can simply delete collapse_all, as the code is already in view.spec.js
 
 yes, pls move the test that perform tests on a specific column to grid/column/view folder

fgiroud (approved):

fgiroud:
Merged ! Thank you @riteshsoni123

PR #4192, opened 2021-10-01
Add GREL scriptText() function for HTML/XML elements

### LINKED ISSUE #4189: Add new GREL jsoup function "data()" to support HTML <script> and <style> tag data extraction
Data does not only exist in HTML or XML **TextNode**'s but also in `<script>` and `<style>` nodes.
Useful JSON data is also sometimes contained in `<script type="application/json">` as well as JSON-LD data in `<script type="application/ld+json">`
Jsoup provides a different method `data()` for working with Elements that are a DataNode for handling the contents of style, script tags etc, where contents of those tags do not show with **TextNode**'s `text()` - our GREL `htmlText()`

### Proposed solution
1. Add support for jsoup's `data()` functionality for **Element**'s as defined at https://jsoup.org/apidocs/org/jsoup/nodes/Element.html#data()
3. `data()` will operate similarly to our existing `htmlText()` for **TextNode**'s but instead work with [DataNodes](https://jsoup.org/apidocs/org/jsoup/nodes/DataNode.html)

### Alternatives considered
Manually parsing based on root element and splitting the entire document as a single string

### Additional context
1. Example tests of `data()` are [here](https://github.com/jhy/jsoup/blob/master/src/test/java/org/jsoup/nodes/ElementTest.java#L38).
2. `:containsData()` is already available currently from within our GREL `select()` function.
3. `parseJson()` should be immediately applicable to parse the JSON data string.
Ex.  `value.parseHtml().select("script")[2].data().parseJson().sitecore.context`

thadguidry:
Fixes #4189

- Adds new GREL ~~`data()`~~ `scriptText()` function to extract contents from HTML/XML script, style, comment tags
- Adds ability to directly extract content and JSON and JSON-LD objects from a `<script>`  eg. `value.parseHtml().select("script").scriptText().parseJson().sitecore.context`

image


wetneb:
@thadguidry sorry that I have not found the time to look at this earlier.

I wonder if it would make sense to use a function name a tiny bit more specific than `data()`. It surely is not a problem to introduce such a method in an object-oriented programming language, but since GREL only has functions, I feel like this could conflict with other uses… How about something like `innerText` instead?
Also, there are now some merge conflicts.

thadguidry:
@wetneb Yeah that makes sense.  However, we want to avoid `inner` prefixing since that always means Text **and** Elements in Jsoup.  In this case it's just Text (precisely it returns a String always, and never Elements to further process with Jsoup functions).
So maybe `scriptText()` is a good name since that will be 99% of users needs, and we just mention in description that it also works with HTML comment `<! --- --->` and `<style>` tags.  Let me work on this today.

wetneb (approved):

PR #4196, opened 2021-10-06
added test for replace and transform

riteshsoni123:
Fixes #{3554}

- Create new file "replace.js" and add test for "Replace".
- Create new file "transform.js" and add test for "Transform" (still incomplete).


fgiroud:
Thanks @riteshsoni123 !

PR #4212, opened 2021-10-13
added test for header elements

### LINKED ISSUE #3412: Add UX Test for project header elements
Header elements should be tested.

Screen Shot 2020-12-16 at 13 04 28

#### Test for header left components

* Test project renaming
* Ensure that the permalink is working

Move export_project.spec.js into this folder

#### Tests for export formats 

* Add a test for each export format.   
For now, we can not ensure that the downloaded file exists, but it will allow detecting when the server triggers an HTTP error on exporting.
* Add a test for the help button

riteshsoni123:
Fixes #3412 

- Create new file "permalink.spec.js" that contains the test for permalink.
- Create new file "help.spec.js" that contains the test for help button.
- Create new file "project_rename.spec.js" that tests the renaming of project.
- Add test for first ten ways to export project in file "export_project.spec.js"

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -17,7 +17,7 @@
   "dependencies": {
     "@crowdin/cli": "3",
     "@docusaurus/core": "^2.0.0-beta.6",
-    "@docusaurus/preset-classic": "^2.0.0-beta.6",
+    "@docusaurus/preset-classic": "^2.0.0-beta.7",
     "clsx": "^1.1.1",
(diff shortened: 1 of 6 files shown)


riteshsoni123:
Hello @fgiroud I found a way through which we can check if the contents of the file is empty or not through "cy.readFile(`cypress/downloads/${name}.tsv`).should('not.be.empty');" . Will it be enough or you want me search for anything else?

PR #4217, opened 2021-10-16
Fix Recon id generation.

### LINKED ISSUE #3785: "Use values as identifiers" via Wikidata QIDs adds erroneous items
I have a list of Wikidata QIDs and I want to use them in an import. Thus, I proceed to simple reconciliation via identifier values ("Use values as identifiers"). However, As many as 1 % of QIDs get erroneously paired to a completely different ID. This is hard to spot and leads to nonsense getting imported into Wikidata.

### To Reproduce
1. Download a list of QIDs, eg: https://w.wiki/3ACf (CSV) and import as a new OpenRefine project
2. First column--> Use values as identifiers. 
3. Extract reconciled QID from 1st column via "cell.recon.match.id".
4. compare 1st column values with values in the column created in (3).

### Expected Behavior
I'd like all QIDs to be reconciled without mistake.

### Screenshots

Bez názvu

### Versions
 - Operating System: Windows 10,
 - Browser Version: Newest Chrome
 - OpenRefine: 3.4.1.

### Datasets
Linked from above. But this has happened to me on numerous occassions.

wetneb:
The previous method was prone to creating collisions when a lot of recon ids
were created around the same time.

Closes #3785.

- use fully random Recon ids that span the entire range of Java `long`s instead of only using a range of 1000000.

### DIFF
--- a/main/src/com/google/refine/model/Recon.java
+++ b/main/src/com/google/refine/model/Recon.java
@@ -39,6 +39,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Random;
 
 import com.fasterxml.jackson.annotation.JsonFilter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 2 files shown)


wetneb:
Seems to work well, merging.

PR #4220, opened 2021-10-17
Allow changing the path to `refine.ini` via refine.sh/bat on command line

### LINKED ISSUE #4113: Allow changing the autosave period via refine.sh/bat on command line
I have some large projects that require launching where I really need the autosave period to be closer to 60 minutes.  I use a few Windows shortcuts on my desktop for some common large OpenRefine projects that I sometimes cross with, etc.  The problem is that I cannot easily use the Windows shortcuts to launch these particular large OpenRefine projects because I cannot also save the shortcuts with a command line parameter to control the autosaving.  I can only control autosaving period from the `refine.ini` file.

a custom `refine.ini` path also is not a parameter that can be passed to the refine.sh/bat which probably would be the best solution all around.

### Proposed solution
allow passing a custom `refine.ini` path as a parameter for the `refine.sh/bat` launcher.
We already have in the scripts a "endConfigReading" section that could be re-used and parameterized for consuming a custom `refine.ini` path such as 
```sh
refine /c C:\Users\thadg\bigfile_refine.ini
```

tejasrbhat:
Fixes #4113


tejasrbhat:
@wetneb @thadguidry sorry for fixing this so late. Was busy with personal work. I have fixed the issue with double quotes and tested. @thadguidry you can test it once and let me know if anything else needs to be added.

refine.bat
```diff
@@ -141,11 +131,25 @@ goto shift2loop
...
 shift
 goto loop
 
+:readIniFile
+
+rem --- Read ini file -----------------------------------------------
+
+if "%REFINE_INI_PATH%" == "" set REFINE_INI_PATH=refine.ini
+echo Using %REFINE_INI_PATH% for configuration
+for /f "tokens=1,* delims==" %%a in (%REFINE_INI_PATH%) do (
```
thadguidry, line 149: You'll need to handle a **file not found**.  Perhaps quickly fail here and echo the error `The system cannot find the file C:\Users\thadg\.sonarlint\refine.ini.`?

My invalid path fail test `refine /c C:\Users\thadg\.sonarlint\refine.ini` still progressed and tried to launch Refine:
```
E:\GitHubRepos\OpenRefine>goto loop

E:\GitHubRepos\OpenRefine>if """" == """" goto readIniFile

E:\GitHubRepos\OpenRefine>rem --- Read ini file -----------------------------------------------

E:\GitHubRepos\OpenRefine>if "C:\Users\thadg\.sonarlint\refine.ini" == "" set REFINE_INI_PATH=refine.ini

E:\GitHubRepos\OpenRefine>echo Using C:\Users\thadg\.sonarlint\refine.ini for configuration
Using C:\Users\thadg\.sonarlint\refine.ini for configuration

E:\GitHubRepos\OpenRefine>for /F "tokens=1,* delims==" %a in (C:\Users\thadg\.sonarlint\refine.ini) do (set %a=%b )
The system cannot find the file C:\Users\thadg\.sonarlint\refine.ini.

E:\GitHubRepos\OpenRefine>rem --- Fold in Environment Vars --------------------------------------------

E:\GitHubRepos\OpenRefine>if not "" == "" goto gotJavaOptions

E:\GitHubRepos\OpenRefine>set JAVA_OPTIONS=

E:\GitHubRepos\OpenRefine>set OPTS=

E:\GitHubRepos\OpenRefine>if not "" == "" goto gotMemory

E:\GitHubRepos\OpenRefine>set REFINE_MEMORY=1024M

E:\GitHubRepos\OpenRefine>if not "" == "" goto gotMemory
```

thadguidry (changes requested):

refine.bat
thadguidry: Something? https://ss64.com/nt/syntax-error.html
Or? https://stackoverflow.com/questions/53654565/batch-exception-handling-using-a-for-loop-with-multiline-in-running-the-commands

thadguidry: Looks like from reading around that `in` is special within for loops and a launches cmd.exe separate process outside of batch-file context.  See the comment in 2nd link https://stackoverflow.com/a/53658484

thadguidry: and https://superuser.com/a/615141

tejasrbhat: @thadguidry I just added "if not exist" and it is working fine. So the flow now is, if /**c** is not given it will take refine.ini from the root folder, if given it will take that. And if any of these files does not exist it will throw exception and exit.

thadguidry: @tejasrbhat tested and works great now!

thadguidry (approved):
@tejasrbhat tested and now works great! Thanks so much for adding this command flag! (and pushing yourself to help get it finished! We thank you!)

PR #4221, opened 2021-10-18
Fix #3946. Add testcase for testing the process instance.

chupengrocky:
Fixes #{3946}
Sorry for late update. I was busy on dealing with academic last few months. Not sure if I should test the runnable part in a process. Thanks 
- Add testCase to test the process generated by the PerformWikibaseEditsOperation class


chupengrocky:
Thanks for replying. Consideration to improve the code coverage is my initial purpose of this pull request. And I also agree that we need to think about the trade-off of testing time. I could take a look on other open issues if it does not require extra test case here.

wetneb:
Ok, yes I think I would rather not merge this as things stand, for the reasons mentioned above. Code coverage is an important measure of the quality of a test suite but it can easily be gamed and we should not encourage tests which only focus on this metric but rather make significant assertions on the behaviour of the program.

PR #4252, opened 2021-10-29
Eval error

### LINKED ISSUE #4193: Improve EvalError output in Expression preview for HTML/XML GREL functions

### Proposed solution

Example:

```java
```

nikhilp3:
Fixes #4193 

- Improved Eval Error messages

### DIFF
--- a/main/src/com/google/refine/expr/functions/Coalesce.java
+++ b/main/src/com/google/refine/expr/functions/Coalesce.java
@@ -53,7 +53,7 @@ public class Coalesce implements Function {
             }
             return null;
         }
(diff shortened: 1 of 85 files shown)


wetneb:
Hi @nikhilp3, do you intend to work on this further or is this ready for review? I just merged https://github.com/OpenRefine/OpenRefine/pull/4255 which addresses some of that but maybe it makes sense to do it systematically for all functions as you have done here.

wetneb:
Since this pull request does not seem active and #4255 is merged already, I am closing this. Happy to reopen if you want to work on this further!

PR #4255, opened 2021-10-31
Eval error test fix

### LINKED ISSUE #4193: Improve EvalError output in Expression preview for HTML/XML GREL functions

nmamoon:
Fixes #4193 

Update error logging and tests


wetneb (commented):
Thanks! I would rather test this with Java unit tests, because using Cypress integration tests is quite expensive and does not add much in this case. Do you think you could give it a try?

nmamoon:
> Thanks! I would rather test this with Java unit tests, because using Cypress integration tests is quite expensive and does not add much in this case. Do you think you could give it a try?

Yep, I'll make this update.

wetneb (approved):
Perfect, thank you so much!

Issue #388, opened 2012-10-15
bad error message when giving -m less than 256M


tfmorris:
_Original author: jeff.al...@gmail.com (May 22, 2011 22:48:28)_

<b>What steps will reproduce the problem?</b>
1. ./refine -m 128M
2. Get error &quot;incompatible min and max heap size&quot; or something.

<b>What is the expected output? What do you see instead?</b>

I expected it to just work, even on my little machine. It didn't. I discovered the -m argument. I expected it to work, it didn't. I read the shell script and realized I had to hack it to lower the default 256M to 128M.

What version of Google Refine are you using?  r1836

What operating system and browser are you using? Linux, Firefox

<b>Is this problem specific to the type of browser you're using or it happens</b>
in all the browsers you tried? No

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=388_

aakankasha028:
@thadguidry I would like to take up this issue so as to make the first contribution. I propose the following changes 

1. Update the usage message for -m by adding the line "values less than 256M are not allowed"
2. While parsing the command line arguments, if the value for -m is less than 256 ( if [ $1 -lt 256  ) then, show them the usage and exit (or show them an error message and exit)
3. Add a comment in the wiki page

wetneb:
Sounds like a great plan to me!

aakankasha028:
@wetneb I have made the changes to the shell script and the PR is at https://github.com/OpenRefine/OpenRefine/pull/2489. The electricty went off just as I pushed my changes so had to rebase them all over again and then create the PR, hence the delay between the push and the creation of PR

Please review the same. Post that, I will update the wiki page as well.

wetneb:
Ok, I am a bit confused by the result. Could you perhaps re-submit the PR as a single commit? Start a new branch from master, cherry-pick your commit from the old branch, push the new branch and submit it as a PR?

aakankasha028:
Sure, I'll do that!

PR #4266, opened 2021-11-04
Fix CLI setting for memory size (-m and /m)

### LINKED ISSUE #4265: Memory setting on Linux CLI is broken
The parameter to define the memory used by OpenRefine is broken in the `./refine` script that we use on Linux.

### To Reproduce
1. Run `./refine -m 2048M`

### Current Results
`./refine: line 852: [: 2048M: integer expression expected`

### Expected Behavior
No error, the memory settings are properly set.

### Versions
 - Operating System: Debian
 - Browser Version: Irrelevant
 - JRE or JDK Version: 11
 - OpenRefine: 3.5-beta2

### Additional context
Brought up here: https://groups.google.com/g/openrefine/c/2nfrI8eteCU

wetneb:
Fixes #4265.

- Revert the faulty fix for #388.
- Let -m (on UNIX) and /m (on Windows) control both the initial heap size and the maximum one, so that we cannot run into the situation where the initial heap size is larger than the maximum one.

### DIFF
--- a/refine
+++ b/refine
@@ -47,8 +47,7 @@ where [options] include:
      default: OS dependent
 
   -m <memory> max memory heap size to use
-     default: 1024M
-     Values less than 256M not allowed
+     default: 1400M
 
   -v <level> verbosity level [from low to high: error,warn,info,debug,trace]
      default: info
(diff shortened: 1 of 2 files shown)


wetneb:
Hmm, by reverting the original commit and not the merge one I have a pretty interesting commit history here! Let's make sure we squash this PR (as usual).

wetneb:
Oh the revert actually made unintended changes, I will redo it.

PR #4279, opened 2021-11-07
Drop support for Java 8.

### LINKED ISSUE #4278: Drop support for Java 8
As discussed on the mailing list, supporting Java 8 prevents us from doing important dependency updates.
We want to shift from supporting Java 8 to 15 to supporting Java 11 to higher versions of Java (at least 16).

Discussion: https://groups.google.com/g/openrefine-dev/c/K2J_61oDQsU

wetneb:
I have attempted to update all the relevant configuration parameters but might have missed some, let's see!

I have also tried to control all of these parameters by two Maven properties in the root artifact, to make their updating a bit easier.

Closes #4278.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -11,7 +11,7 @@ jobs:
   server_tests:
     strategy:
       matrix:
-        java: [ 8, 14 ]
+        java: [ 11, 14 ]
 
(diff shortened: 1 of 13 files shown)


wetneb:
The failure should due to the discrepancy between the new and old config files, so I will merge this to test this further.

PR #4280, opened 2021-11-07
Updated build-test-run.md no JDK later that 14.0.2

mattzeeee:
Added a recommendation about JDK version to use for building and running OpenRefine.
Build-test-run.md sugguested to install the latest JDK which let's you build OpenRefine, but running it does not work see: 
https://github.com/OpenRefine/OpenRefine/issues/4106
This might be frustrating for beginners not aware of this issue.

Related to #{4106}


wetneb:
Thanks! That being said I am curious about how you determined that JDKs > 14.0.2 do not work - have you tried with Java 15? At the moment we pretend we are compatible with Java 15, so if that is not the case I would also update that.

wetneb:
(Sorry about the spurious build failures, I am responsible for them…)

mattzeeee:
> 
> 
> Thanks! That being said I am curious about how you determined that JDKs > 14.0.2 do not work - have you tried with Java 15? At the moment we pretend we are compatible with Java 15, so if that is not the case I would also update that.

My bad, I thought I read that Java 14 was the last version that worked properly. I haven't tried Java 15 to be honest.

wetneb:
Absolutely, you are very welcome to add further commits to this branch and the pull request will be updated accordingly.

wetneb (approved):

PR #4281, opened 2021-11-08
Add Homebrew docs to both 3.5 and master docs

wetneb:
Adds back the text contributed by @joewiz which got lost in the versioning of the docs.


wetneb:
For #4285.

joewiz:
It appears this that one of the checks for this PR has stalled. Would it be possible to restart it?

wetneb:
Those checks are not normally run for documentation-only PRs (I agree with you that the UI is confusing, we need to fix this). Let me add back the missing changes and finally merge this.

joewiz:
@wetneb Thank you!

PR #4260, opened 2021-11-01
PPMd Zip Compression Error

nmamoon:
Fixes #4112

Changes proposed in this pull request:
- Throw exception if trying to import PPMd-compressed zip file

PR #4286, opened 2021-11-10
Display error for unsupported compression file type

nikhilp3:
Fixes #4260 

- Display on the UI when an invalid compression format is used


main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -539,13 +542,15 @@ static public long saveStreamToFile(InputStream stream, File file, SavingUpdate
                 }
             }
             return length;
+        } catch (ZipException e) {
+            throw new IOException("Compression format not supported, " + e.getMessage());
```
wetneb, line 546: Are we sure that all `ZipException`s are related to the fact that the compression format is not supported? Intuitively there could be other cases, where the zip archive is corrupted, for instance?

main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -136,7 +138,7 @@ public boolean isCanceled() {
             JSONUtilities.safePut(config, "state", "error");
             JSONUtilities.safePut(config, "error", "Error uploading data");
             JSONUtilities.safePut(config, "errorDetails", e.getLocalizedMessage());
-            return;
+            throw new IOException(e.getMessage());
```
wetneb, line 141: I have not checked thoroughly, but it seems curious that we need to throw an exception again after handling it above. Is it a problem with the architecture of the current code?

wetneb (commented):
Thank you so much! I have not tried it yet but had a few questions just looking at your changes.

main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -691,8 +696,8 @@ static public boolean explodeArchive(
                     }
                 }
             } catch (IOException e) {
-                // TODO: what to do?
                 e.printStackTrace();
+                throw new IOException(e.getMessage());
```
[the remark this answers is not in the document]
nikhilp3: That would probably be a good idea, I can remove the try/catch

main/src/com/google/refine/importing/ImportingUtilities.java
nikhilp3: Would have to do some more research here, but that makes sense. Perhaps we can catch Exception and forward along the error message rather than specifying the compression format is not supported.

main/src/com/google/refine/importing/ImportingUtilities.java
nikhilp3: To display on the UI we need to throw an exception here, we can not catch anything as an alternative. However, in general I think this file needs a major refactor but that would be for another ticket.

wetneb: Yes the state of this file is pretty terrible. Thanks for confirming.

nikhilp3: After looking more, I think this is fine for now and the only way to handle without changing the logic - an issue can be opened about cleaning up this file

main/src/com/google/refine/importing/ImportingUtilities.java
nikhilp3: Looking into this - I think this current way makes sense. Looking at microsoft docs it is throws "when the file or stream is not a valid zip file."

While that could mean a corrupted file I think that can be bucketed into compression format not supported since it is not a valid compression in the first place. Any other non ZipException will be throws as IOException and still propagated up to the UI but without the additional text added for ZipException.

main/src/com/google/refine/importing/ImportingUtilities.java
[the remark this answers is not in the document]
nikhilp3: Removed in the latest commit

wetneb (approved):
Thank you very much! It works as expected.

PR #4305, opened 2021-11-17
Make WbItemDocumentExpr extend EntityIdValue instead of ItemIdValue

### LINKED ISSUE #4268: Make Wikibase extension compatible with other entity types than Items

### Proposed solution

### Additional context

j-sal:
Fixes #4268 

- Make `WbItemDocumentExpr` extend `EntityIdValue` instead of `ItemIdValue`.
- Modify corresponding tests.


wetneb:
@j-sal amazing! It looks like this pull request also includes your previous changes, which have already been merged into master. It is worth making sure you update your master branch locally (`git checkout master ; git pull`) before you start working on a new pull request. For now we can solve this simply by pushing the "Update branch" button which will simply merge the master branch in your pull request branch.

wetneb:
Another small point: it is helpful if you give an easy to read title to your pull request. By default, GitHub draws it from your branch name (if you are pushing multiple commits), so it currently is "4268 edit media info entity": you could make it more descriptive such as "Make WbItemDocumentExpr extend EntityIdValue instead of ItemIdValue"

wetneb (approved):
Apart from the two small points above it looks great to me! I can merge as soon as you update the title (which will become the commit message).

PR #4307, opened 2021-11-17
Fix #4194 toString() and parseJson() are not compatible with JSON spec

### LINKED ISSUE #4194: toString() and parseJson() are not compatible with JSON spec
Per JSON Spec...

> A value can be a string in double quotes, or a number, or true or false or null, or an object or an array. These structures can be nested.

But we throw away (or don't handle properly for some reason) any double quotes in a JSON array when we convert `toString()`.

- Double Quotes retained correctly when encapsulated as expected by single quotes
```
'[ "dentsu mcgarrybowen Regents Place 10 Triton Street", "London NW1 3BF", "+44 (0)20 3640 5592", "For new business enquiries please contact our Marketing and New Business Director, Oli Richards", "For everything else please email us", "© 2020 mb - uk", "22 Torrington Place London, WC1E 7HJ 0203 640 5592" ]'.toString()
```
### To Reproduce
Import the below provided Dataset `test.openrefine.tar.gz` project file as a test case.

1. Try to Add a column based on FETCH column with the following GREL expression to convert the parsed out array to a String:

```
forEach(value.parseHtml().select("p"),v,v.htmlText()).toString()
```

2. Notice that in Preview and after clicking OK in the column that double quotes were **not retained** and useful structure that was needed by retaining the quotes is now lost completely.

### Expected Behavior
I would expect this String value representation with Double Quotes retained for JSON structure to be stored in the cell as such:
`[ "dentsu mcgarrybowen Regents Place 10 Triton Street", "London NW1 3BF", "+44 (0)20 3640 5592", "For new business enquiries please contact our Marketing and New Business Director, Oli Richards", "For everything else please email us", "© 2020 mb - uk", "22 Torrington Place London, WC1E 7HJ 0203 640 5592" ]`

### Screenshots
Good JSON array structure I want to keep as a String and push into a cell
image

But the type is an `array`

image

Shown below we see `toString()` mangles and loses that good structure where by default it should keep it. If necessary perhaps a parameter option added IF we wanted to throw away that structure or manipulate the formatting of the String ( #816 concerns numbers not JSON arrays as strings.)

image

### Versions
 - JRE or JDK Version: JDK15
 - OpenRefine: OpenRefine 3.5 Beta1

### Datasets

[test.openrefine.tar.gz](https://github.com/OpenRefine/OpenRefine/files/7271903/test.openrefine.tar.gz)

### Additional context

Looking at the array possibilities on https://www.json.org/json-en.html

I think our test cases could also probably be expanded and improved for not only `parseJson()` but also `toString()` to include JSON values that are **string in double quotes**, along with **another array**,  and not only **number, true, false, null** as we currently test.

Mengfan-Zhu:
Fixes #4194

-  Add a new optional parameter on toString(), if set this parameter to true, the output string will follow JSON array format
-  Add test cases for the new feature


wetneb:
Hi @Mengfan-Zhu, thank you very much for this contribution!

I just added [a comment](https://github.com/OpenRefine/OpenRefine/issues/4194#issuecomment-971936455) on the issue this pull request is meant to fix, can you have a look?

Mengfan-Zhu:
> Hi @Mengfan-Zhu, thank you very much for this contribution!
> 
> I just added [a comment](https://github.com/OpenRefine/OpenRefine/issues/4194#issuecomment-971936455) on the issue this pull request is meant to fix, can you have a look?

Yes, I reply to your comment

main/src/com/google/refine/util/StringUtils.java
```diff
@@ -48,4 +48,35 @@ public static String toString(Object o) {
             return o.toString();
         }
     }
+
+    /**
+     * String formatting method for the array of objects which keep the json array format
+     * @param o an array of objects to be converted to a string
+     * @return string representing array of objects
+     */
+    public static String JSONtoString(Object[] o){
```
wetneb, line 57: I think we could directly use the Jackson library for this (`ParsingUtilities.mapper.writeValueAsString` if I remember correctly).

Mengfan-Zhu: > I think we could directly use the Jackson library for this (`ParsingUtilities.mapper.writeValueAsString` if I remember correctly).

I think it's different. I just search for this function, it will return the JSON object to string like:
{key1: value 1,
key2: value 2}
What my function does is to return a JSON array to string like:
["value1", "value2", "value3"]

wetneb: I think the Jackson function will work both for objects (where it will indeed return things like `{"key1":"value1", "key2":"value2"}`) and arrays (where it will return things like `["value1", "value2", "value3"]`).

Mengfan-Zhu: > I think the Jackson function will work both for objects (where it will indeed return things like `{"key1":"value1", "key2":"value2"}`) and arrays (where it will return things like `["value1", "value2", "value3"]`).

I did more search but still did not find any information that this function can handle the array case. (I do find the ObjectMapper can read from JSON array but no information about write) I would really appreciate it if you can reference a document or tutorial to show how to do that.

wetneb: How about you just try it out?
```java
Object[] myArray = new Object[] { "foo", true, 32 };
System.out.println(ParsingUtilities.mapper.writeValueAsString(myArray));
```

Mengfan-Zhu: Thanks for your hint. Yes, it works. I will modify it later.

main/src/com/google/refine/expr/functions/ToJson.java
```diff
@@ -0,0 +1,113 @@
...
+                if (element instanceof OffsetDateTime && format != null) {
+                    OffsetDateTime odt = (OffsetDateTime) element;
+                    element = (Object) odt.format(DateTimeFormatter.ofPattern(format));
+                    formatFlag[i] = true;
+                }
+                if (element instanceof Number && format != null) {
+                    element = String.format(format, (Number) element);
+                    formatFlag[i] = true;
+                }
+                array[i] = element;
```
wetneb, line 81: Here, you are modifying the argument of the function, which is generally not recommended since it can lead to unpredictable behaviours.

wetneb (commented):
Thank you for following up on this!
I think we are on the right track. I appreciate that you added this second argument to specify a formatting string, but I can see a few problems with this:
* first, the same format string is used both for dates and for numbers, so that is not very flexible: if you happen to have dates and numbers in the same array, it will be difficult to find a format string which makes sense for both
* second, it will not be supported for nested objects: for instance, if you want to serialize something like `[{"date":"2018-03-02"},{"date":"2011-04-05"}]`, it will not work because the elements of the array are not dates themselves, but objects which contain dates
* third, it artificially restricts the scope of this function to arrays, whereas it could also work on objects or even basic datatypes (boolean, number, string without an enclosing array or object).

For those reasons, I think I would recommend to remove this logic altogether and let the user do those conversions themselves before calling `toJson`. Basically, the `toJson` function would become a simple wrapper over the `ParsingUtilities.mapper.writeValueAsString` function, so it would be much simpler. What do you think?

main/src/com/google/refine/expr/functions/ToJson.java
Mengfan-Zhu: Thanks for pointing out the hidden problems. I think your previous feature of the formatting option is really good and I want to keep it. But if you think just including the basic functionality of wrapper over the ``ParsingUtilities.mapper.writeValueAsString``  is ok, I can modify it to just keep that.

Mengfan-Zhu:
I simplify the function @wetneb

main/src/com/google/refine/expr/functions/ToJson.java
```diff
@@ -0,0 +1,87 @@
...
+    public Object call(Properties bindings, Object[] args) {
+        String error_message = " accepts an array of objects";
+        Object result = new EvalError(ControlFunctionRegistry.getFunctionName(this) + error_message);
+        if (args.length == 1 && args[0] instanceof Object[]) {
+            Object[] array = (Object[]) args[0];
+            try{
+                result = ParsingUtilities.mapper.writeValueAsString(array);
+                String result_str = (String) result;
+                // add whitespace based on formatting
+                result_str = result_str.replace(",",", ").replace("[","[ ").replace("]"," ]").replace("[  ]","[ ]");
```
wetneb, line 64: I think those `replace` calls are unfortunately incorrect in general.
For instance, if you do `toJson(["[,]"])`, you will get `"[\"[ ,  ]\"]"`, which is incorrect. I would remove all of those calls to `replace`. If you want to have a pretty-printed JSON output (which should not be necessary) then Jackson's native features for that should be used.

main/src/com/google/refine/expr/functions/ToJson.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Override
+    /**
+     * Transfor an array of objects into a json array format string
+     * @param args the user input
+     *        args[0] is an array of objects
+     * @return string representing array of objects in json array format
+     */
+    public Object call(Properties bindings, Object[] args) {
+        String error_message = " accepts an array of objects";
```
wetneb, line 56: With this new version, we are actually compatible with more datatypes. I think the simplest way to say it is `an object` (not necessarily an array).

wetneb (commented):
Thank you! Nearly there :)

main/src/com/google/refine/expr/functions/ToJson.java
Mengfan-Zhu: Thanks for pointing out that, I didn't think of that case. I did that to make the function can do what the Expected Behavior shows in issue 4194. I will change it to use Jackson's features. I'm not familiar with that, thanks for your reminder.

main/src/com/google/refine/expr/functions/ToJson.java
Mengfan-Zhu: Yes, I will change it

Mengfan-Zhu:
Finish the changes. @wetneb

main/src/com/google/refine/expr/functions/ToJson.java
wetneb: I think it is fine not to follow the exact formatting given in #4194, the differences are only cosmetic.

wetneb (approved):
Thank you very much!

wetneb:
I tried to update the documentation to mention this new function there, and realized that we already have a `jsonize` function which does exactly that… That is unfortunate! I am sorry about this, we should have closed the issue without any change in the first place. To still acknowledge your work on this, I will merge your change into the `Jsonize` function and merge this pull request.

Issue #4009, opened 2021-06-24
Preview update with no effect is initiated after changing unused parameters


jquartel:
After loading a new file and coming to the Configure Parsing Options page, you have the ability to modify the parsing rules, which I do pretty frequently with the kinds of (mostly tsv-like) files that I get. Most frequently I need to change the "Ignore first" and "Discard initial" parameters and since I usually need both of these and since my files are large and complex enough for the preview update to take some time, I try to save time by changing the parameters before they are checked/enabled. However, I find that even with the parameters left unenabled, changing their value results in an "Updating preview" pause even though the preview isn't changed (as it shouldn't be because the parameter isn't enabled).

### To Reproduce
1. Open a data file (a csv, say) and continue to the "Configure Parsing Options" page
2. Change the "Ignore first" value without checking the enable box 
3. Observe the pause while "Updating preview", even though nothing about the preview is changed (until you enable the "Ignore first" value and then there is another pause and a correct preview update).

### Expected Behavior

I don't expect there to be any preview update when changing values of unchecked (unenabled) parsing parameters. Either that or the parameter should be enabled when you change the value, and then the preview updated accordingly. It might be good to have some way to hold off all preview updates until you explicitly press the Update Preview button, since I usually want to to change a few parameters and waiting for the update after each one is time consuming.

### Screenshots
<img width="582" alt="Screenshot 2021-06-24 at 14 28 27" src="https://user-images.githubusercontent.com/36742596/123272756-f65a7280-d4f9-11eb-9c0b-c5e6648abf51.png">

### Versions
 - Operating System: MacOS 10.14.6
 - Browser Version: Chrome 91.0.4472.114
 - JRE or JDK Version: Unknown
 - OpenRefine: Version 3.5-beta1 [1b9907e]

johnvl2:
@wetneb Hello, I have been working this issue. Could you go ahead and assign me?

wetneb:
Sure, thanks for working on this!

johnvl2:
@wetneb I'm trying to manually test this out for Google Data and Database parsing but I'm unable to get either type working (on my branch and even on master). I'm trying to use the google doc mentioned in #3604 and the public database specified here: https://docs.rfam.org/en/latest/database.html (this db is mentioned in the manual). Google data gets stuck on the "Preparing..." screen and Database gives me a "error:Unauthorized" message. if you have any tips or could point me in the right direction, I would really appreciate it.

wetneb:
[...]

PR #4308, opened 2021-11-18
Issue 4009 preview UI

jinouy2:
Fixes #{issue number here}

- 
- 
-


wetneb:
Hi @jinouy2, it looks like this pull request intends to tackle #4009, which has already been solved by #4309, so I suggest we close this, unless you want to improve on the solution proposed in #4009?

wetneb:
Closing per the comment above.

PR #4309, opened 2021-11-19
Add ability to disable the refreshing of import previews

### LINKED ISSUE #4009: Preview update with no effect is initiated after changing unused parameters

johnvl2:
Fixes #4009 

- Added a checkbox to disable automatic preview updates
- Implemented the feature for all parser types, google data, and databases
- Added a test to preview_project.spec.js to test the feature on a csv file. The test passes (confirmed on cypress)

This pull request addresses issue 4009. This provides a way to hold off all preview updates when desired so the user can change all parameters without having to wait for the update between each change.

Note: I was able to test the feature on all file types except for google data. I was not able to successfully import data from a google doc (on my branch or on master)


main/tests/cypress/cypress/integration/create-project/preview_project.spec.js
```diff
@@ -219,4 +219,46 @@ describe(__filename, function () {
...
+    cy.get('table.data-table tr').eq(1).should('to.contain', '717');
+
+    cy.get('input[bind="ignoreCheckbox"]').uncheck();
+    cy.waitForImportUpdate();
+
+    // **Testing ignore feature with auto preview disabled** //
+    cy.get('input[bind="disableAutoPreviewCheckbox"]').check();
+    // Verify no auto update
+    cy.get('input[bind="ignoreCheckbox"]').check();
+    cy.wait(5000); // 5 second wait to be safe
```
wetneb, line 248: I would normally avoid using `wait` but I understand that here it is not so easy to do, since we are not waiting for anything to happen in the DOM. We are just waiting to make sure nothing changed even after a while. I am happy with it as it stands but I just wonder if @fgiroud has something else to propose?

extensions/database/module/langs/translation-en.json
```diff
@@ -47,6 +47,7 @@
     "database-parsing/worksheet": "Worksheets",
     "database-parsing/option": "Options",
     "database-parsing/preview-button": "Update&nbsp;Preview",
```
wetneb, line 49: This is not strictly speaking related to your changes, but this button seems to have escaped our normalization of casing (#3363), so let me suggest this change anyway:
```suggestion
    "database-parsing/preview-button": "Update&nbsp;preview",
```

wetneb (approved):
Thank you very much, it looks really great to me! I tried it out and it works as it should.

I was a bit sad that you had to go through so many importers to do this change. Ideally this functionality ought to be independent from the importer. But that is a problem with the current architecture of this part of the code, not of your pull request.

I will leave this open for a bit to give @fgiroud a chance to chime in on the `wait(5000)`, but merging soon if no other approach comes to mind.

extensions/database/module/langs/translation-en.json
fgiroud: @johnvl2 Hello and sorry for the late reply
One thing I have noticed is that the button "Ignore first" has no effect alone, you must also enter a number of rows to ignore.
But yes, to have a real test, you have no other choice than to "wait", because the dom is not rendered, and that's maybe something you can add as a comment next to it
Also, you can probably keep only one single assertion, about one single cell of the grid to improve the test readbility

Thx,

main/tests/cypress/cypress/integration/create-project/preview_project.spec.js
fgiroud: @johnvl2 Hello and sorry for the late reply
One thing I have noticed is that the button "Ignore first" has no effect alone, you must also enter a number of rows to ignore.
But yes, to have a real test, you have no other choice than to "wait", because the dom is not rendered, and that's maybe something you can add as a comment next to it
Also, you can probably keep only one single assertion, about one single cell of the grid to improve the test readbility

Thx,

johnvl2:
@wetneb I thought the same thing about going through all of the different importers, but I wanted to stick with the design of the code. Maybe a refactorization down the road to make things like this easier would be a good idea. Going to resolve comments now...

main/tests/cypress/cypress/integration/create-project/preview_project.spec.js
johnvl2: @fgiroud Yes, I noticed this as well about "ignore first", but with the disable checkbox we can avoid the unnecessary preview update for this as well. Added a comment to the wait call with your explanation above. Also removed unnecessary assertions. Kept just the first two which checks an unchanged cell, and a cell we expect to change to verify proper functionality.

extensions/database/module/langs/translation-en.json
johnvl2: Updated to adhere to casing standards.

johnvl2:
Resolved comments from review.

wetneb (approved):

PR #4336, opened 2021-12-01
Make schema editor serialize WbEntityVariable

### LINKED ISSUE #4268: Make Wikibase extension compatible with other entity types than Items

j-sal:
Fixes #4268 

- Make schema editor serialize WbEntityVariable instead of WbItemVariable

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang.NotImplementedException;
 import org.openrefine.wikidata.schema.entityvalues.ReconEntityIdValue;
 import org.openrefine.wikidata.schema.exceptions.NewItemNotCreatedYetException;
 import org.openrefine.wikidata.updates.TermedStatementEntityUpdate;
@@ -38,8 +39,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.wikidata.wdtk.datamodel.helpers.Datamodel;
 import org.wikidata.wdtk.datamodel.interfaces.EntityDocument;
+import org.wikidata.wdtk.datamodel.interfaces.EntityIdValue;
 import org.wikidata.wdtk.datamodel.interfaces.ItemDocument;
 import org.wikidata.wdtk.datamodel.interfaces.ItemIdValue;
+import org.wikidata.wdtk.datamodel.interfaces.MediaInfoIdValue;
 import org.wikidata.wdtk.datamodel.interfaces.MonolingualTextValue;
 import org.wikidata.wdtk.wikibaseapi.WikibaseDataEditor;
 import org.wikidata.wdtk.wikibaseapi.WikibaseDataFetcher;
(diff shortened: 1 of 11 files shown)


extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
```diff
@@ -149,33 +152,42 @@ public void performEdit()
...
+	                editor.updateTermsStatements(currentDocument,
+                            labels,
+	                        descriptions,
+	                        update.getAliases().stream().collect(Collectors.toList()),
+	                        new ArrayList<MonolingualTextValue>(),
+	                        update.getAddedStatements().stream().collect(Collectors.toList()),
+	                        update.getDeletedStatements().stream().collect(Collectors.toList()),
+	                        summary, tags);
+                } else if (newCell instanceof MediaInfoIdValue) {
+                    throw new NotImplementedException();
```
wetneb, line 189: Nice! Now we just need to replace this exception by something that works :)

extensions/wikidata/src/org/openrefine/wikidata/schema/entityvalues/ReconItemIdValue.java
```diff
@@ -33,6 +33,11 @@ public ReconItemIdValue(Recon recon, String cellValue) {
         super(recon, cellValue);
     }
 
+	@Override
+	public boolean isPlaceholder() {
+		return true;
```
wetneb, line 38: I would say we should only return `true` when the `ReconItemIdValue` is associated with a `Recon` object with a `New` judgment. If the entity exists, then this is not a placeholder.

extensions/wikidata/src/org/openrefine/wikidata/schema/entityvalues/ReconItemIdValue.java
```diff
@@ -33,6 +33,15 @@ public ReconItemIdValue(Recon recon, String cellValue) {
         super(recon, cellValue);
     }
 
+	@Override
+	public boolean isPlaceholder() {
+		if (isNew()) {
+            return true;
+		} else {
+            return false;
+		}
```
wetneb, line 42: Here you can just say `return isNew()`. Also, the same change should be done for the others: `ReconPropertyIdValue` and `ReconMediaInfoIdValue`.

extensions/wikidata/src/org/openrefine/wikidata/schema/entityvalues/SuggestedItemIdValue.java
```diff
@@ -31,6 +31,11 @@ public SuggestedItemIdValue(String id, String siteIRI, String label) {
         super(id, siteIRI, label);
     }
 
+	@Override
+	public boolean isPlaceholder() {
+		return true;
```
wetneb, line 36: For items coming from a suggest service, we know they always exist, so this should be `return false` (and similarly for other `Suggested*IdValue`).

extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
```diff
@@ -149,33 +154,47 @@ public void performEdit()
...
+	                        new ArrayList<MonolingualTextValue>(),
+	                        update.getAddedStatements().stream().collect(Collectors.toList()),
+	                        update.getDeletedStatements().stream().collect(Collectors.toList()),
+	                        summary, tags);
+                } else if (newCell instanceof MediaInfoIdValue) {
+                    MediaInfoDocument currentDocument = (MediaInfoDocument) currentDocs.get(update.getItemId().getId());
+	                List<MonolingualTextValue> labels = update.getLabels().stream().collect(Collectors.toList());
+	                labels.addAll(update.getLabelsIfNew().stream()
+	                      .filter(label -> !currentDocument.getLabels().containsKey(label.getLanguageCode())).collect(Collectors.toList()));
+	                MediaInfoUpdate updatesCollection = Datamodel.makeMediaInfoUpdate((MediaInfoIdValue) update.getItemId(), currentDocument.getRevisionId(), Datamodel.makeTermUpdate(labels, tags), Datamodel.makeStatementUpdate(update.getAddedStatements(), update.getDeletedStatements(), tags));
```
wetneb, line 195: This line is too long: you should introduce intermediate variables for the arguments. For instance, by writing on the line above:
`TermUpdate labelUpdate = Datamodel…`

Also, it seems that you have put too much trust in your IDE's auto-completion mechanism. When you use the IDE to guess the arguments of a method, it will try to find objects in the context with matching types. Sometimes that works well, in other cases that is nonsense. For instance, the variable `tags` has type `List<String>` and it looks like your IDE has used it for all the places where a `List<String>` is expected. But those `tags` are just tags that are applied when the edit is made (they are not part of the content that we are pushing), so it does not make sense to use `tags` in any part of the `MediaInfoUpdate`.

So I would encourage you to have a critical eye on the suggestions made by the IDE. When using a method, you can look at the documentation of that method to understand what the required arguments are supposed to mean, and check if the suggested variables make sense for that.

wetneb (approved):

PR #4342, opened 2021-12-06
Issue 4193 evalerror output

### LINKED ISSUE #4193: Improve EvalError output in Expression preview for HTML/XML GREL functions

jinouy2:
Fixes #4193

- 
- 
-


wetneb:
The corresponding issue was already closed.

PR #4347, opened 2021-12-07
Mention UniversalLanguageSelector requirement

lucaswerkmeister:
ULS provides the [`languagesearch` API](https://www.mediawiki.org/wiki/Special:MyLanguage/API:Languagesearch) used to select term languages.

- Add ULS to the 3.5 docs
- (no other docs because IIUC arbitrary Wikibase support is new in 3.5)

### DIFF
--- a/docs/docs/manual/wikibase/configuration.md
+++ b/docs/docs/manual/wikibase/configuration.md
@@ -19,7 +19,7 @@ To let your users contribute to your Wikibase instance with OpenRefine, you will
 
 ### Requirements {#requirements}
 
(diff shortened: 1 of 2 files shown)


lucaswerkmeister:
Hm, I guess I should also add this to the [non-versioned docs](https://github.com/OpenRefine/OpenRefine/blob/de0112d0f5084705e69a64bd319f415249642b5b/docs/docs/manual/wikibase/configuration.md)?

wetneb:
Thank you! Yes indeed it would be worth adding it to the development docs. (I would like to set something up to automatically port those changes, but maybe it's overkill, I don't know…)

lucaswerkmeister:
Alright, updated.

wetneb (approved):
Thanks again!

PR #4352, opened 2021-12-08
Fix issue #3955 Missing Row Separators in Templating Export

### LINKED ISSUE #3955: Missing Row Separators in Templating Export
### To Reproduce
1. First, select the "Templating..." option under Export menu.
2. Use the default values for the json output.
3. Finally, observe that a number of "rows" (really records) at the start, are missing the "Row separator".

### Current Results
The first two exported items are missing a row separator:

```
{
  "rows" : [
    {
      "One" : "a",
      "Two" : "1",
      "Three" : "2"
    }    {
      "One" : null,
      "Two" : "3",
      "Three" : "4"
    },
    {
      "One" : "c",
      "Two" : "5",
      "Three" : "6"
    }
  ]
}
```

Also this data set also gets displayed with 2 (not 3) rows because the second row, with no value in the first column, doesn't get numbered.

### Expected Behavior

My data set should have three rows and the export should have all the row separators.
### Screenshots

Screen Shot 2021-06-02 at 15 36 00

### Versions
 - Operating System: MacOS 10.15.7
 - Browser Version: Safari 14.1.1, FireFox 88.01
 - JRE or JDK Version: build 1.8.0_111-b14
 - OpenRefine: 3.4.1

### Datasets
A small reproducible example:

```
"One","Two","Three"
"a",1,2
"",3,4
"c",5,6
```

Mengfan-Zhu:
Fixes #3955

- Modify main/src/com/google/refine/templating/Template.java
- Add test

### DIFF
--- a/main/src/com/google/refine/templating/Template.java
+++ b/main/src/com/google/refine/templating/Template.java
@@ -87,6 +87,7 @@ public class Template {
         protected Properties bindings;
 
         public int total;
+        public int totalRows;
 
         public RowWritingVisitor(Writer writer, int limit) {
             this.limit = limit;
@@ -117,32 +118,22 @@ public class Template {
             }
         }
 
-        /**
-         * This method is modified for issue 3955
-         * Issue link: https://github.com/OpenRefine/OpenRefine/issues/3955
-         * The modification is to remove the self-increasing of total 
-         * and move it to method internalVisit
-         */
         @Override
         public boolean visit(Project project, int rowIndex, Row row) {
             if (limit <= 0 || total < limit) {
                 internalVisit(project, rowIndex, row);
             }
+            total++;
 
             return limit > 0 && total >= limit;
         }
 
-        /**
-         * This method is modified for issue 3955
-         * Issue link: https://github.com/OpenRefine/OpenRefine/issues/3955
-         * The modification is to remove the self-increasing of total 
-         * and move it to method internalVisit
-         */
         @Override
         public boolean visit(Project project, Record record) {
             if (limit <= 0 || total < limit) {
                 internalVisit(project, record);
             }
+            total++;
 
             return limit > 0 && total >= limit;
         }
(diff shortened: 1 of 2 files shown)


Mengfan-Zhu:
A little description of my changes:
The problem will only happen in the record mode and the missing value is in the first record. 
I can see from the code in Template.java that currently for the record mode, it just transfers it to row and handle it as the same as the row mode. But the problem is, in record mode, if the second row has the missing value it will think that the first and second rows are the same record, and the ``total`` will not self-increasing after processing the first row. Therefore, I change the position for ``total++`` to avoid it.

wetneb (approved):
Thank you! I like the simple fix and the accompanying test.

I guess by doing this change, you are slightly changing the meaning of the `limit` parameter when run in records mode. Before, it was the maximum number of records to output, and now it is the maximum number of rows to output (which might not correspond to a record boundary).
I think it would be cleaner to preserve this meaning, for instance by introducing a new variable to determine if a row has been written already.

That being said, as mentioned in the issue, we do not actually support the records mode in any interesting way in this exporter, so it is pretty unlikely that people use this exporter in records mode with a genuine intention to influence its behaviour. So I would be happy to merge that in this state if you prefer not to introduce a new variable.

Mengfan-Zhu:
> Thank you! I like the simple fix and the accompanying test.
> 
> I guess by doing this change, you are slightly changing the meaning of the `limit` parameter when run in records mode. Before, it was the maximum number of records to output, and now it is the maximum number of rows to output (which might not correspond to a record boundary). I think it would be cleaner to preserve this meaning, for instance by introducing a new variable to determine if a row has been written already.
> 
> That being said, as mentioned in the issue, we do not actually support the records mode in any interesting way in this exporter, so it is pretty unlikely that people use this exporter in records mode with a genuine intention to influence its behaviour. So I would be happy to merge that in this state if you prefer not to introduce a new variable.

Thanks for pointing out that. I agree that introducing a new variable is a better idea . If someone wants to add new features for the record mode in the future, they may need the total to count the number of records.  
I introduce the new variable ``totalRows``.  Please check my new commit.

wetneb (approved):
Looks great to me, thank you!

PR #4354, opened 2021-12-08
Fix spurious escaping for TSV exporter

### LINKED ISSUE #2071: TSV export escapes double quotes (") unnecessarily

### To Reproduce

### Current Results

### Expected behavior

### Screenshots
In my project
image

In my tsv
image

**Desktop**
 - OS: win10
 - Browser Version: Chrome

**OpenRefine**
 - Version : OR 3.1

JohnMcGrane:
Fixes #2071

I made some errors with Git so the commit history may look a little wonky, but my work does effectively solve this issue. 

This fix creates a tsv exporter that functions according to the specifications outlined in the original issue.


wetneb:
Thanks a lot! I tried looking at [your changes](https://github.com/OpenRefine/OpenRefine/pull/4354/files) but there are quite a lot of unrelated formatting changes. Could you try to revert those? For instance, `main/src/com/google/refine/commands/expr/PreviewExpressionCommand.java` seems to only have formatting changes, so you could restore this file to its original version entirely.

main/src/com/google/refine/exporters/CsvExporter.java
```diff
@@ -74,12 +72,12 @@ public CsvExporter(char separator) {
     @Override
     public void export(Project project, Properties params, Engine engine, final Writer writer)
             throws IOException {
-        
+
         String optionsString = (params == null) ? null : params.getProperty("options");
-        Configuration options = new Configuration();
+        com.google.refine.exporters.CsvExporter.Configuration options = new com.google.refine.exporters.CsvExporter.Configuration();
```
wetneb, line 77: If this import conflicts with another one, then I would rather import our own class, and keep the fully qualified name for the other class (the external one).

wetneb (approved):
Thanks! It is a bit sad that we have to resort to this, ideally this should be handled by the CSV library we use. But unfortunately we depend on a patched version of that library which has not been updated for ages. So even if that became (or is already) supported upstream, we would not benefit it >_<. Merging soon unless anyone has concerns…

JohnMcGrane:
Hey, @wetneb this is my first pull request so I'm not completely familiar with this process. Please let me know if there is anything left for me to do with this issue. If there is, please be specific because I'm still a newbie. Thanks!

wetneb:
@JohnMcGrane you could ideally add a further commit which would address the import issue mentioned above. You could push this commit to your `issue#2071` and it will update this pull request.

JohnMcGrane:
@wetneb Okay I reverted the file to what it was before. I believe I had made that change because PMD said it was violating some codestyle but I'm happy to change it back as clearly it was working fine beforehand.

cdapp:
unsubscribe

JohnMcGrane:
Hey @wetneb, is there anything else I can do to help this get merged? Thanks for your help this far with everything!

main/tests/server/src/com/google/refine/exporters/TsvExporterTests.java
```diff
@@ -170,14 +169,31 @@ public void exportTsvWithQuote(){
...
+        try {
+            SUT.export(project, options, engine, writer);
+        } catch (IOException e) {
+            Assert.fail();
+        }
+
+        Assert.assertEquals(writer.toString(), "column0\tcolumn1\tcolumn2\n" +
+                "row0cell0\trow0cell1\trow0cell2\n" +
+                "row1cell0\t\"whole line is quoted\"\trow1cell2\n" +
+                "row2cell0\trow2cell1\trow2cell2\n");
```
wetneb, line 191: This test case worries me a bit. If we have a cell whose string is already surrounded by quotes and export it like you specify above, then I bet most parsers (including ours) will remove those leading and trailing quotes at parsing time, by understanding them as being part of the format?

main/src/com/google/refine/exporters/CsvExporter.java
```diff
@@ -116,21 +113,53 @@ public void addRow(List<CellData> cells, boolean isHeader) {
...
-                            (cellData != null && cellData.text != null) ?
-                            cellData.text :
-                            "";
+                        // If file is tsv and cell text contains internal tabs, then manually escape tabs
+                        if (separator.charAt(0) == '\t' && cellData != null && cellData.text != null && cellData.text.contains("\t")) {
+                            final String tabString = cellData.text.replace("\t", "\\t");
+                            strings[i] = tabString;
+                        }
+                        // If file is tsv & cell contains internal newlines, manually escape newlines
+                        else if (separator.charAt(0) == '\t' && cellData != null && cellData.text != null && cellData.text.contains("\n")) {
```
wetneb, line 122: Why is this second case specific to TSV files? I suspect the same reasoning could also apply to CSV, or with any other choice of separator, no?

wetneb (commented):
Sorry for the delay!
I am still a bit torn on this actually. I have added my concerns again below.

I feel like the existing behaviour is also a legitimate one. The behaviour you propose is also very sensible, and has the benefit of making sure each line of the CSV/TSV file corresponds to a single row (which is great for parsing with Spark), but I feel like we should probably give the user a choice between those, no? Of course, there is currently no UI to specify this, so we would need to introduce one… Or does that mean we are actually doing something redundant with the custom tabular exporter? My brain hurts…

@msaby, any thoughts?

msaby:
sorry I'm quite far from Openrefine these days

JohnMcGrane:
@wetneb It's been a while since I worked on this problem. I was mainly just following the specifications from the issue page. I don't know the ins and outs of csv vs tsv import and exporting other than what I read while working on this issue.

wetneb:
Ok, I think this is not ready for prime time then, we need to step back a bit and think beyond those particular use cases, to make sure we are not breaking anything.

PR #4355, opened 2021-12-08
Fix templating exporter to accept closing braces in expressions

### LINKED ISSUE #3381: Broken escaping in templating export

JohnMcGrane:
Fixes #3381

Templating is fixed according to the issue specification.


main/src/com/google/refine/templating/Parser.java
```diff
@@ -33,15 +33,39 @@
...
+     * Finds the first true ending brace in a cell, ignoring escaped end braces
+     *
+     * @param expression string that will be interrogated for first true end brace
+     * @param lowIndex   index from which the interrogation will start
+     * @return index of the first true end brace
+     * CS427 Issue Link: https://github.com/OpenRefine/OpenRefine/issues/3381
+     */
+    private static int findEndBrace(final String expression, final int lowIndex) {
+        int closeBrace = expression.indexOf('}', lowIndex);
+        if (expression.charAt(closeBrace - 1) == '\\') {
```
wetneb, line 60: What happens if no `}` character could be found? In the case of an unfinished template, we should make sure we fail graciously.

wetneb (commented):
Besides the spurious reformatting problems (which I cannot really blame you for, we should have a CI-enforced linting), this PR looks very good to me. I am just wondering about a side case, see below.

main/src/com/google/refine/templating/Parser.java
JohnMcGrane: Hey @wetneb I just added a parsing exception in case no end brace can be found. Is that what you meant by fail graciously? If not I can go back and implement whatever you were thinking.

wetneb:
Looks like we have some merge conflicts that should be solved before the CI can kick in.

JohnMcGrane:
@wetneb Okay, I think this should be good to go, let me know what you think!

JohnMcGrane:
@wetneb Okay I added back the test that I had taken out. (It was causing some build failure but I just needed to format it correctly).

wetneb (approved):
Thanks a lot!

PR #4356, opened 2021-12-08
Unify number rendering in expression preview and data grid

### LINKED ISSUE #3102: Inconsistent display of scientific notation between data grid and expression preview
### To Reproduce

Using GREL in the Expression Editor with `value.toNumber()` to transform a String representation of scientific E notation 9.07185e-5 to a real number 0.0000907185, the preview only capitalizes the E (9.07185E-5) instead of showing the real number in the transform preview.  The transform however does happen correctly when applied.

### Current Results
the preview only capitalizes the E (9.07185E-5) instead of showing the real number in the transform preview. 

### Expected Behavior
We should be consistent in terms of what we display in the preview and what is displayed in the cells in the data grid. So I think if we show "0.0000907185" in the data grid we should show that in the preview as well.

At the moment in the data grid it looks like we switch to scientific notation at 1e-7 (going down) and 1e+21 (going up)
I think the preview should work in the same way

### Versions
 - JRE or JDK Version: JDK8
 - OpenRefine:  Openrefine 3.3

JohnMcGrane:
Fixes #3102

Preview cells will now display the numbers in the format that they will ultimately be displayed in the data grid.

### DIFF
--- a/main/src/com/google/refine/commands/expr/PreviewExpressionCommand.java
+++ b/main/src/com/google/refine/commands/expr/PreviewExpressionCommand.java
@@ -23,8 +23,8 @@ LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
 A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
 OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
 SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
-LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,           
-DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY           
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
 THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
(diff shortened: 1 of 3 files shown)


main/src/com/google/refine/commands/expr/PreviewExpressionCommand.java
```diff
@@ -244,24 +277,35 @@ static protected void writeValue(StringBuffer sb, Object v, boolean quote) {
...
                         sb.append((String) v);
                     }
                 } else if (v instanceof Double || v instanceof Float) {
                     Number n = (Number) v;
+                    //CS427 Issue Link: https://github.com/OpenRefine/OpenRefine/issues/3102
+                    final double minNonSciNotation = 0.000_001;
+                    final double maxNonSciNotation = 1_000_000_000_000_000_000_000.0;
+                    final DecimalFormat decimalFormat = new DecimalFormat("0"); //, DecimalFormatSymbols.getInstance(Locale.ENGLISH));
+                    decimalFormat.setMaximumFractionDigits(15);
+                    decimalFormat.setMaximumIntegerDigits(22);
```
wetneb, line 299: Those constants are hard-coded here, but if the goal is to match the rendering provided in the data table itself, would it not make sense to have this logic stored only in one place instead? That would avoid those renderings getting out of sync? I have not checked if that makes sense though :)

wetneb:
Hi @JohnMcGrane, do you plan to go back to this pull request?

main/src/com/google/refine/commands/expr/PreviewExpressionCommand.java
JohnMcGrane: @wetneb Yes I think that definitely makes sense. Do you know where I can access those values in the code without hardcoding them? Thanks!

wetneb: I do not know it on top of my head, perhaps you could find the relevant place by debugging? I suspect it could happen in the frontend, actually. If that is the case, then we cannot refer directly to it from the Java code. We have multiple options:
* shift the rendering of previewed values to the frontend too, so that we can unify rendering there
* introduce a Java class dedicated to value rendering, which replicates what the frontend does, and that can be used outside of the `PreviewExpressionCommand`

wetneb:
Closing per inactivity, but happy to reopen if you are interested in pushing this further.

JohnMcGrane:
Hey @wetneb This is definitely still on my to do list, but currently it is not at the top. So while I am interested in moving this forward it may not be for a few weeks yet.

wetneb:
Ok, great! Let's reopen it when you get back to it :)

thadguidry:
@JohnMcGrane still on your todo list?

JohnMcGrane:
Unfortunately I don't think I'll get around to this. Thanks for checking!

PR #4357, opened 2021-12-09
Use OSSRH snapshot repository to host snapshot releases.

### LINKED ISSUE #4346: Use Maven Central to host snapshot releases instead of a custom GitHub repository
We currently upload our snapshot releases to the releases section of a dedicated GitHub repository.
That works okay, but the process is a bit convoluted (it requires some scripts and a dedicated GitHub account to do the upload in the CI). Also, we want to extend this to cover not just the `master` branch, but also the `4.0` branch, and this is not so simple.

### Proposed solution

We could do like Gephi does: use the Maven Central snapshot repository, where we can upload not just jars, but also released artifacts like our `.dmg`, `.zip` and `.tar.gz`. This lets us store one snapshot per candidate version (we would be able to access the latest master snapshot and the latest 4.0 snapshot).

### Alternatives considered
We could keep the existing system put both of those snapshot releases in the same repository, but then the latest release would not always correspond to the same branch, giving unpredictable results for the users.
We could also create another dummy repository to hold the snapshot releases of 4.0, but that is a bit confusing.

### Additional context

Solution discovered at the Gephi code retreat.

wetneb:
Closes #4346.

- integrate snapshot uploading into the Maven build process (for the standard `deploy` goal)
- remove scripts to upload to GitHub releases

This might need merging before it can be tested.

### DIFF
--- a/.github/workflows/release_manager.sh
+++ b/.github/workflows/release_manager.sh
@@ -1,8 +1,0 @@
-#!/usr/bin/env bash
-
-
-# Build Release Message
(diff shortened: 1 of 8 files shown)


wetneb:
Merging to try it out.

PR #4364, opened 2021-12-10
Grammatical Error

shriyasankhyan:
Facets are essential aspects or angles of data variance in a given column. Instead of essentially, essential should be used.

Fixes #{issue number here}

- 
- 
-


wetneb:
I think the original version is correct actually!

shriyasankhyan:
> I think the original version is correct actually!

Perhaps, I may be wrong.

wetneb:
@shriyasankhyan thanks for the pull request in any case!

PR #4365, opened 2021-12-10
One of the instead of one of

shriyasankhyan:
Fixes #{issue number here}

- 
- 
-


wetneb:
Thanks @shriyasankhyan! I think the original wording is correct in this case too, actually :)

shriyasankhyan:
> Thanks @shriyasankhyan! I think the original wording is correct in this case too, actually :)

Thanks a lot for your kind responses.

PR #4374, opened 2021-12-17
Migrate to log4j 2.16.0

wetneb:
Even though our version of log4j is apparently not vulnerable to the recent Log4Shell vulnerability that made the headlines, the version we rely on is not supported anymore and there are reports of people not being able to use OpenRefine anymore because it relies on this old version.

This migrates to Log4j 2.16.0.


wetneb:
Build is currently failing because the new version of butterfly is not available in the repository yet.

PR #4387, opened 2021-12-22
update Maven GAVC for correct Commons-IO

thadguidry:
Updates correct groupId for commons-io for refine-model

- 
- 
-

### DIFF
--- a/refine-model/pom.xml
+++ b/refine-model/pom.xml
@@ -161,11 +161,10 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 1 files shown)


wetneb:
@thadguidry do you plan to work on this further?

thadguidry:
Yeah let me work on updating this now.

thadguidry:
@wetneb Ah! There is no need for this I found out.  It was a bad artifact in my M2 local repo actually from cross-pollination with another GitHub project.
We have the correct GAVC as shown officially in build distribution from Apache:  https://commons.apache.org/proper/commons-io/summary.html

So closing.

Issue #2313, opened 2020-02-14
spark-prototype: temp files not being deleted on Windows


thadguidry:
**Describe the bug**
temp files on Windows are not deleted and cause SparkException.

**To Reproduce**
1. Load csv into OpenRefine Spark
2. Do a few cell edits
3. Wait for spark.ContextCleaner to clean the accumulators
4. See error when temp files try to get removed.

**Current Results**
```
Caused by: org.apache.spark.SparkException: Job aborted due to stage failure: 
  Task 0 in stage 13.0 failed 1 times, most recent failure:
    Lost task 0.0 in stage 13.0 (TID 13, localhost, executor driver):
      java.io.IOException:
       (null) entry in command string:
         null chmod 0644 C:\Users\thadg\AppData\Roaming\OpenRefine\2096101862730.project\initial\grid\_temporary\0\_temporary\attempt_20200214082621_0034_m_000000_0\part-00000
        at org.apache.hadoop.util.Shell$ShellCommandExecutor.execute(Shell.java:762)
        at org.apache.hadoop.util.Shell.execCommand(Shell.java:859)
        at org.apache.hadoop.util.Shell.execCommand(Shell.java:842)
<snip>
```

**Expected behavior**
On Windows, Spark should be able to delete the temp files.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Version: Firefox 
 - JRE or JDK Version: OpenJDK 11

**OpenRefine (please complete the following information):**
 - Version spark-prototype

**Additional context**
We might need to properly configure Spark for Windows: https://spark.apache.org/docs/latest/configuration.html
We might have to configure Spark for Windows correctly, through trial and error, or research with Spark community.
We might instead use Python API for Spark (supposedly it didn't have issues according to Kingsley Jones)
Possible Resolutions? (I found it interesting what Kingsley Jones had to say ): https://issues.apache.org/jira/browse/SPARK-12216?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=16134527#comment-16134527

thadguidry:
Ah!   We might need to fix the currently missing \bin\winutils.exe to really fix this according to this: 
https://stackoverflow.com/a/48012285/1100717

Here's my startup logs currently today ( built from commit : https://github.com/OpenRefine/OpenRefine/commit/f947be6e36ee4a9f55fa060084bd7655e7bf2f24 )
```
[...]

PR #4392, opened 2021-12-28
Remove Hadoop winutils, too brittle (see #2313)

wetneb:
For #2313.

Let's see if it runs okay on Appveyor and interactively as a user.


wetneb:
Hmmm, it appears that Winutils is actually not optional on Windows.
Let me think of another solution.

PR #4394, opened 2021-12-29
Remove hadoop dependency in local runner

wetneb:
Although Spark was only a dependency of the Spark runner (and therefore OpenRefine could run without it), we still depended on Hadoop for its filesystem, even with the local runner.
Concretely, we used it for its ability to read text files in a partitioned way (which lets us parallelize computations run on a single file easily).

This was not really ideal for a few reasons:
- this is a big dependency, which pulls in a lot of other things we do not use. This also depends on some libraries we already used, and therefore forces us to stick to the versions of those libraries that Hadoop depends itself on (for instance, Jetty).
- the interface that Hadoop offers to control the partitioning strategy was not really ideal for us, leading to some inelegant code to run the partitioning heuristics multiple times until we get the desired result
-  the feature of Hadoop that we use (partitioning text files) can actually be implemented fairly easily without it.
- and Hadoop is a pain to run on Windows (https://issues.apache.org/jira/browse/HADOOP-13223).

Therefore this removes the dependency to Hadoop in the local runner. Installing Hadoop's winutils on Windows is still a requirement to use the Spark runner, which we cannot avoid because that is a requirement in Spark itself anyway.


wetneb:
Unfortunately our UI tests do not support re-running inside the CI apparently, so I am just going to merge this and look into any integration test failures later if they appear to persist.

PR #4395, opened 2021-12-30
Fixed sample extension pom.xml

JanC:
hey,

the sample extension does not compile because of the main maven dependency

```
cd extensions/sample
mvn compile

..
 Could not resolve dependencies for project org.openrefine:sample:jar:3.0-SNAPSHOT: org.openrefine:main:jar:3.0-SNAPSHOT was not found in https://oss.sonatype.org/content/repositories/snapshots/ during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of snapshots has elapsed or updates are forced
```

This PR should address it


wetneb (approved):
Thanks!

PR #4408, opened 2022-01-05
fix #4407 down-arrow image path broken on custom installation

### LINKED ISSUE #4407: down-arrow image path broken on custom installation
I setup a docker image with OpenRefine 3.5.1 and found out that the down-arrow image path is broken. It would be ok if the image would have the same path as the logo.

### To Reproduce
1. Deploy OpenRefine in a Docker container
2. Then, import some dummy data into a project
3. Finally, the down-arrow button is broken

### Screenshots
screenshot

### Versions
 - Operating System: Linux/Docker
 - Browser Version: Latest Chrome
 - JRE or JDK Version: 11
 - OpenRefine: 3.5.1

patrickwyler:
Fixes #4407


wetneb (approved):
Oh I see! Then it was definitely a bug in the first place, the leading `..` were just ignored because there was no folder to go up to. Interesting… thanks a lot for the fix!

PR #4416, opened 2022-01-10
Refactor editor api for existing Items

### LINKED ISSUE #4268: Make Wikibase extension compatible with other entity types than Items

j-sal:
Fixes #4268

- Factor out deprecated editor api in `EditBatchProcessor` for existing Item updates
- Update `EditBatchProcessorTest` class

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
@@ -29,6 +29,8 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang.NotImplementedException;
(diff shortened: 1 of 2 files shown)


extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
```diff
@@ -182,14 +184,31 @@ public void performEdit()
...
+                        aliasUpdate = aliasMap.get(a.getLanguageCode());
+	                }
+	                /* solution 2
+	                 * Getting same error as with previous attempts:
+	                 * Type mismatch: cannot convert from Map<Object,Object> to Map<String,AliasUpdate>
+	                 *
+	                Set<MonolingualTextValue> aliases = update.getAliases();//set of pairs of strings we're going to add
+	                Map<String, List<MonolingualTextValue>> aliasesMap = aliases.stream().collect(Collectors.groupingBy(MonolingualTextValue::getLanguageCode));
+	                Map<String, List<MonolingualTextValue>> map = new HashMap<>();
+	                Map<String, AliasUpdate> aliasMap = map.entrySet().stream().collect(Collectors.toMap(Entry::getKey, e -> String.valueOf(e.getLanguageCode())));
```
wetneb, line 201: Solution 2 is indeed invalid as things stand because your expression `String.valueOf(e.getLanguageCode())` does not produce an `AliasUpdate` but a `String`.

Given the variable `e` of type `List<MonolingualTextValue>`, you need to produce a corresponding `AliasUpdate` object.
This can be done with `Datamodel.makeAliasUpdate(e, Collections.emptyList())`.

wetneb, line 201:
```suggestion
	                Map<String, AliasUpdate> aliasMap = map.entrySet().stream().collect(Collectors.toMap(Entry::getKey, e -> Datamodel.makeAliasUpdate(e, Collections.emptyList())));
```

wetneb (approved):
Amazing! Next step: writing a similar unit test for MediaInfo

PR #4430, opened 2022-01-13
WIP: Add MediaInfo entity update test-case

### LINKED ISSUE #4268: Make Wikibase extension compatible with other entity types than Items

j-sal:
Fixes #4268 

- Add a test-case for MediaInfo entity updates

### DIFF
--- a/extensions/wikidata/tests/src/org/openrefine/wikidata/editing/EditBatchProcessorTest.java
+++ b/extensions/wikidata/tests/src/org/openrefine/wikidata/editing/EditBatchProcessorTest.java
@@ -173,8 +173,10 @@ public class EditBatchProcessorTest extends WikidataRefineTest {
             throws MediaWikiApiErrorException, InterruptedException, IOException {
         // Prepare test data
         MonolingualTextValue label = Datamodel.makeMonolingualTextValue("village in Nepal", "en");
+        List<MonolingualTextValue> labels = Collections.singletonList(label);
         StatementGroup statement = Datamodel.makeStatementGroup(Collections.emptyList());
-        TermUpdate labels = Datamodel.makeTermUpdate(null, Collections.emptyList());
+        List<StatementGroup> statements = Collections.singletonList(statement);
+        TermUpdate labelsUpdate = Datamodel.makeTermUpdate(null, Collections.emptyList());
         List<String> ids = new ArrayList<>();
         for (int i = 124; i < 190; i++) {
             ids.add("Q" + String.valueOf(i));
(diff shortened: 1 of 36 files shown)


extensions/wikidata/tests/src/org/openrefine/wikidata/editing/EditBatchProcessorTest.java
```diff
@@ -161,11 +167,69 @@ public void testMultipleBatches()
...
+            ids.add("Q" + String.valueOf(i));
+        }
+        List<MediaInfoIdValue> mids = ids.stream().map(e -> Datamodel.makeWikimediaCommonsMediaInfoIdValue(e))
+                .collect(Collectors.toList());
+        List<TermedStatementEntityUpdate> batch = mids.stream()
+                .map(mid -> new TermedStatementEntityUpdateBuilder(mid).addLabel(label, false).build())
+                .collect(Collectors.toList());
+
+        int batchSize = 50;
+        List<MediaInfoUpdate> fullBatch = mids.stream()
```
wetneb, line 187: This should be a `List<MediaInfoDocument>` instead.

wetneb (commented):
The problem you have here is that you are mixing up two concepts:
* `MediaInfoDocument` represents the state of a MediaInfo entity at a given time;
* `MediaInfoUpdate` represents an edit on a MediaInfo entity (in other words, a change, or the difference between two MediaInfoDocument.

You cannot use the two interchangeably.

extensions/wikidata/tests/src/org/openrefine/wikidata/updates/ItemUpdateTest.java
```diff
@@ -78,36 +78,36 @@ public ItemUpdateTest() {
```
wetneb, line 78: This test class (`ItemUpdateTest`) should also be renamed to `TermedStatementEntityUpdateTest`.

extensions/wikidata/tests/src/org/openrefine/wikidata/editing/EditBatchProcessorTest.java
```diff
@@ -161,11 +168,70 @@ public void testMultipleBatches()
...
+        List<MediaInfoIdValue> mids = ids.stream().map(e -> Datamodel.makeWikimediaCommonsMediaInfoIdValue(e))
+                .collect(Collectors.toList());
+        List<TermedStatementEntityUpdate> batch = mids.stream()
+                .map(mid -> new TermedStatementEntityUpdateBuilder(mid).addLabel(label, false).build())
+                .collect(Collectors.toList());
+
+        int batchSize = 50;
+        List<MediaInfoDocument> fullBatch = mids.stream()
+                .map(mid -> Datamodel.makeMediaInfoDocument(mid, labels, statements).findLabel(label.toString())
+                        .build()).collect(Collectors.toList());
```
wetneb, line 193: I am not sure why you are using the `findLabel` method here. Why not just `Datamodel.makeMediaInfoDocument(mid, labels, statements)` (without the `.build()` call, since this is not a builder class?)

wetneb (approved):
Amazing!

PR #4441, opened 2022-01-18
Update translation-fr.json

belett:
ortho and typo

- "nouvel" and not "nouveau"
- "élément" and not "Élément"


wetneb:
Thanks! I assume you also know about our Weblate interface when you can do those changes? It is also fine to submit PRs like this one of course.
https://hosted.weblate.org/engage/openrefine/

PR #4446, opened 2022-01-19
Does this release process docs need to be updated to say Java 11 now?

thadguidry:
@wetneb I'm also wondering what else within the release process docs needs to be updated to make it "current practice for now" ?


wetneb:
The file you edited is for 3.5, for which Java 8 still works… so it is the current docs that we need to update :)

wetneb (approved):

PR #4448, opened 2022-01-19
Fix warning about license during packaging

thadguidry:
Fix license download warning during packaging
- Updates `<name>` tag to a SPDX identifier https://spdx.org/licenses/ as recommended by https://maven.apache.org/pom.html#licenses
- Updates `<url>` tag to use raw content for easier preview with tools

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,8 +16,8 @@
   </organization>
   <licenses>
     <license>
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thanks!

PR #4450, opened 2022-01-20
Update facets example documentaion

iainelder:
You need this step to complete the numeric faceting example.

Without it, the numeric facet shows "no numeric value present".

### DIFF
--- a/docs/versioned_docs/version-3.5/manual/facets.md
+++ b/docs/versioned_docs/version-3.5/manual/facets.md
@@ -17,7 +17,7 @@ Typically, you create a facet on a particular column. That facet selection appea
 
 ### An example {#an-example}
 
-You can learn about facets and filtering with the following example. You can copy the following table and paste it using the <span class="menuItems">Clipboard</span> method of starting a project if you would like to try it yourself.
+You can learn about facets and filtering with the following example. You can copy the following table and paste it using the <span class="menuItems">Clipboard</span> method of starting a project if you would like to try it yourself. Check the "Attempt to parse cell text into numbers" option so that we can use numeric faceting.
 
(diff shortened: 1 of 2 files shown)


docs/versioned_docs/version-3.5/manual/facets.md
```diff
@@ -17,7 +17,7 @@ Typically, you create a facet on a particular column. That facet selection appea
 
 ### An example {#an-example}
 
-You can learn about facets and filtering with the following example. You can copy the following table and paste it using the <span class="menuItems">Clipboard</span> method of starting a project if you would like to try it yourself.
+You can learn about facets and filtering with the following example. You can copy the following table and paste it using the <span class="menuItems">Clipboard</span> method of starting a project if you would like to try it yourself. Check the "Attempt to parse cell text into numbers" option so that we can use numeric faceting.
```
wetneb, line 20:
```suggestion
You can learn about facets and filtering with the following example. You can copy the following table and paste it using the <span class="menuItems">Clipboard</span> method of starting a project if you would like to try it yourself. Check the "Attempt to parse cell text into numbers" option so that you can use numeric faceting.
```

wetneb: That's a minor point - it feels more natural to me, in the given context ("We" is used later on as referring to the author of the docs, so it does not feel suitable to use "we" here)

iainelder: Fine by me. It's consistent with the first sentence in the paragraph.

The whole page text seems to use "you" and "we" interchangably.

wetneb (approved):

Issue #2191, opened 2019-10-14
macOS Catalina cannot verify that OpenRefine is free from malware?


hughmarera:
I have just installed OpenRefine under macOS Catalina and i am getting the two message below.

** “OpenRefine” cannot be opened because the developer cannot be verified.
** macOS cannot verify that this app is free from malware.

I tend to ignore the first message but I am particularly concerned about the second one.

wetneb:
Hi @hughmarera,

This message is shown because we do not have a paid Apple Developer Program account.
https://developer.apple.com/support/compare-memberships/

If anyone wants to pay $100 a year for this message to disappear, they should get in touch on this ticket or at openrefine-dev@googlegroups.com.

thadguidry:
Added a small footnote to this issue on the [macOS installation section of our wiki](https://github.com/OpenRefine/OpenRefine/wiki/Installation-Instructions#macos)

hughmarera:
Fine, just to say that the previous version of macOS Mojave mentioned only this first message (i.e. “OpenRefine” cannot be opened because the developer cannot be verified.) and did not complain about potential malware issues.

ostephens:
Thanks @hughmarera - I've further clarified that in the wiki entry - it's good for us to be aware of the messaging that Apple is using here, even though despite the change in wording, the cause (lack of paid Apple Developer Program account means we aren't 'verified') and solution is the same as previously (use the "Open Anyway" option)

stevevance:
The warning that I get is different and doesn't offer the "open anyway" button. I see only "move to trash" and "cancel". 

<img width="420" alt="image" src="https://user-images.githubusercontent.com/1330459/74694970-e02afc80-51b7-11ea-9c1d-709f934b9396.png">

coopr:
This remains an issue due to discoverability - https://openrefine.org/download.html has a brief tip for the Windows kit, but includes no similar tip for the Mac kit - and we now know that EVERY user of the Mac kit (at least those with not-ancient MacOS versions) will encounter this issue.

I created https://github.com/OpenRefine/openrefine.github.com/pull/68 to address the solution discoverability issue.

alexladerman-nav:
Open Anyway only gives me the option to _Move to Bin_ or _Cancel_ (MacOS Big Sur 11.4)

wetneb:
[...]

PR #4458, opened 2022-01-24
Update brew installation

Lisanna:
"All brew cask commands have been deprecated in favour of brew commands (with --cask) when necessary" from brew.sh/2020/12/01/homebrew-2.6.0

Related to issue #2191 in case users prefer the Brew option to the `.dmg` download. 

- Update commands from `brew cask install` to `brew install --cask` as suggested by brew.sh/2020/12/01/homebrew-2.6.0


wetneb (approved):
Thanks a lot!

PR #4476, opened 2022-01-29
Add linting check in CI

### LINKED ISSUE #2338: Code style improvements

wetneb:
Fixes #2338

This adds support for Java linting in CI. The build will fail if the code is not correctly linted, which can be fixed by running `mvn formatter:format`.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -64,7 +64,7 @@ jobs:
         distribution: 'adopt'
         java-version: ${{ matrix.java }}
 
-    - name: Check Java linting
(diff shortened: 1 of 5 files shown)


wetneb:
Merging to try it out on other PRs, will revert if the setup reveals inconvenient.

Issue #3221, opened 2020-09-25
Allow easier running, stopping, logging of OpenRefine


thadguidry:
We want to make running and stopping easier for OpenRefine on Windows and Linux (we assume MacOS is good enough for now).

#### Windows
On Windows, allow running OpenRefine as a native application, possibly with tray icon support (without displaying the console by default).
- JAVA - The current supported and recommended community practice in **Java** world is to code a small **JavaFX** application.
- RUST - But we might also instead use an actual systems programming language like **Rust** with **Druid** or **Relm** ((see Alternatives below).

The MVP for the application would be:
1. The application window UI would have the following elements:
   - scrollable console text output (configurable line history display buffer with default at 3000)
   - File menu with options
     - **Exit** to ask a standard "Are you sure - OK/Cancel" and then gracefully shutdown OpenRefine and then the GUI app.
   - Close button (red X) on the GUI app would offer a confirmation dialog **Are you sure you wish to exit OpenRefine?**
   - Minimize/Maximize buttons
2. OPTIONAL - 
   - **Save Log** to allow saving the entire log output since beginning of application startup to a file
   - System Tray icon support
   - Right click on the System Tray gives only a single option **Exit OpenRefine**.
   - Double-click System Tray icon opens the application UI.
   - Toast notifications on Windows 10/11
    
#### Linux
**TODO:** On Linux, allow running OpenRefine as / with ???  

#### MacOS
Nothing to do ?

### Alternatives considering
1. @dante9992 also mentioned that [Tauri](https://github.com/tauri-apps/tauri) also supports [system tray icon on Windows](https://tauri.studio/en/docs/usage/guides/visual/system-tray) and can do app signing, self updating, toast notifications, etc.  It also allows the app to be packaged as a .deb (Debian package) file.  @thadguidry asked and paid for the Tauri team (see comments below) to put together a prototype with Tauri and Rust with code now stored here:  https://github.com/OpenRefine/openrefine-tauri

2. @timClicks said to me on Discord that we could just use Rust and [Druid](https://github.com/linebender/druid/tree/master/druid) or maybe [Relm](https://github.com/antoyo/relm)

3. some very hacky Launch4j opensource wrappers that are not 100% ideally compatible with Windows 10. (they were mostly coded and made for pre Win8.1 OS and no longer sufficient)

4. Electron - has some non-free issues (see [comment](https://github.com/OpenRefine/OpenRefine/issues/3221#issuecomment-867067089) below)

[...]

PR #4484, opened 2022-01-30
Add snap package

kristbaum:
Adds snap package as #{1445} suggests. 

- Create snapcraft.yaml
- add .snap packages to .gitignore to prevent accidental uploading of locally build packages

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ refine-dev.ini
 
 # Ignore Node modules that might inadvertently install at current path instead of user or configured env var NODE_PATH 
 node_modules/
+
+
+*.snap
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -24,8 +24,7 @@ parts:
     source: https://github.com/OpenRefine/OpenRefine/releases/download/$SNAPCRAFT_PROJECT_VERSION/openrefine-linux-$SNAPCRAFT_PROJECT_VERSION.tar.gz
     stage-packages:
       - default-jre
-      - wget
-      - ca-certificates-java 
+      - wget 
     prime:
       - -usr/lib/jvm/java-*/lib/security/cacerts
       - -usr/lib/jvm/java-*/jre/lib/security/cacerts
@@ -34,11 +33,10 @@ apps:
   openrefine:
     command: ./refine
     environment:
-      JAVA_HOME: $SNAP/usr/lib/jvm/java-11-openjdk-amd64
+      JAVA_HOME: $SNAP/usr/lib/jvm/default-java
       PATH: $JAVA_HOME/jre/bin:$PATH
       _JAVA_OPTIONS: -Duser.home="${SNAP_USER_DATA}"
     plugs:
-      - home
       - network
       - network-bind
       - x11
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -26,6 +26,9 @@ parts:
       - default-jre
       - wget
       - ca-certificates-java 
+    prime:
+      - -usr/lib/jvm/java-*/lib/security/cacerts
+      - -usr/lib/jvm/java-*/jre/lib/security/cacerts
 
 apps:
   openrefine:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,40 @@
+name: openrefine
+base: core20
+version: '3.5.2'
+summary: A free, open source, powerful tool for working with messy data
+description: |
+  OpenRefine (previously Google Refine) is a powerful tool for working with messy data:
+  cleaning it; transforming it from one format into another; and extending it with web services and external data.
+  OpenRefine always keeps your data private on your own computer until YOU want to share or collaborate. Your private data never leaves your computer unless you want it to. (It works by running a small server on your computer and you use your web browser to interact with it)
+
+grade: stable
+confinement: strict
+compression: lzo
+
+architectures:
+  - build-on: amd64
+  
+layout:
+  /usr/bin/free:
+    bind-file: $SNAP/usr/bin/free
+
+parts:
+  openrefine:
+    plugin: dump
+    source: https://github.com/OpenRefine/OpenRefine/releases/download/$SNAPCRAFT_PROJECT_VERSION/openrefine-linux-$SNAPCRAFT_PROJECT_VERSION.tar.gz
+
+    stage-packages:
+      - default-jre
+      - wget
+
+apps:
+  openrefine:
+    command: ./refine
+    environment:
+      JAVA_HOME: $SNAP/usr/lib/jvm/java-11-openjdk-amd64
+      PATH: $JAVA_HOME/jre/bin:$PATH
+    plugs:
+      - home
+      - network
+      - network-bind
+      - x11
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -25,9 +25,7 @@ parts:
     stage-packages:
       - default-jre
       - wget
-    prime:
-      - -usr/lib/jvm/java-*/lib/security/cacerts
-      - -usr/lib/jvm/java-*/jre/lib/security/cacerts
+      - ca-certificates-java 
 
 apps:
   openrefine:
@@ -35,6 +33,7 @@ apps:
     environment:
       JAVA_HOME: $SNAP/usr/lib/jvm/java-11-openjdk-amd64
       PATH: $JAVA_HOME/jre/bin:$PATH
+      _JAVA_OPTIONS: -Duser.home="${SNAP_USER_DATA}"
     plugs:
       - home
       - network
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -22,10 +22,12 @@ parts:
   openrefine:
     plugin: dump
     source: https://github.com/OpenRefine/OpenRefine/releases/download/$SNAPCRAFT_PROJECT_VERSION/openrefine-linux-$SNAPCRAFT_PROJECT_VERSION.tar.gz
-
     stage-packages:
       - default-jre
       - wget
+    prime:
+      - -usr/lib/jvm/java-*/lib/security/cacerts
+      - -usr/lib/jvm/java-*/jre/lib/security/cacerts
 
 apps:
   openrefine:


thadguidry:
@kristbaum I wouldn't be the ideal reviewer for this.  I did test it in my dev environment, but I'm a Windows user with a highly configured Docker Ubuntu (and WSL2 sessions with it primarily) that I use for dev work.  Let's get someone who actually is an Ubuntu user ideally.

wetneb (commented):
Thanks a lot!
I have not tested it yet, but I wonder how to work around the problems you mentioned:
> * the command to start is available from PATH and is called "openrefine" not "refine". We could request an alias for that in the future

I expect people will just launch the app graphically, from the list of applications installed on their computer, so this does not seem like a big deal.

>     * The browser doesn't automatically launch, because of the missing java-desktop integration. (This seems to be solvable but complicated)

I wonder if this could be solved by opening the browser from the `openrefine` script rather than from the Java backend (for instance with `xdg-open`).
This issue seems quite important to me: otherwise, we need to make it clear to users how they can access the tool (they cannot really discover by themselves which URL to visit, as things stand).

> 
>     * The location of $HOME/.local/share/openrefine is $HOME/snap/openrefine/versionnr/.local/share/openrefine

That's interesting! Can we control which version number is being used? If it is the full version number (`3.5.2`) then that means users will effectively see their workspace being cleared every time they upgrade. It would be better if we could rely only on the major version (`3`).

kristbaum:
> I expect people will just launch the app graphically, from the list of applications installed on their computer, so this does not seem like a big deal.

There isn't an desktop icon yet, just to keep the user experience close to the original zip package. There should be no problem adding one, but then users can't additionally configure cmd options when launching. For most users this shouln't be an issue though.

> I wonder if this could be solved by opening the browser from the openrefine script rather than from the Java backend (for instance with xdg-open).
This issue seems quite important to me: otherwise, we need to make it clear to users how they can access the tool (they cannot really discover by themselves which URL to visit, as things stand).

Launching the browser from java-desktop as it is happening now, seems quite complicated, but maybe doable. I've asked about it [here](https://forum.snapcraft.io/t/remapping-local-share-applicationname-to-inside-the-snap/28521/4). Using xdg-open in the .desktop file is a solution, but would ignore any custom port setting of the users.

> That's interesting! Can we control which version number is being used? If it is the full version number (3.5.2) then that means users will effectively see their workspace being cleared every time they upgrade. It would be better if we could rely only on the major version (3).

Sorry I should have referred to the revision number of the snap here, no this is not really controlable. There are also other options to store user files listed [here](https://snapcraft.io/docs/data-locations) and they should be copied over afaik on an upgrade, to allow for rollbacks, but this should be tested as well.

wetneb:
I might be wrong, but I would expect that someone who installs OpenRefine through the snap store would also be happy to have a fully graphical launching experience (just like what we have on Mac: an OpenRefine logo, which lauches the server and opens the browser for you). For those who are familiar with the command line, it is not so hard to download the zip and run it from there, no?
So generally I would support having a non-configurable fully graphical experience. Of course it would be better to also have the option to configure graphically (I think we have the issue somewhere to write a launcher, similarly to #3221).

wetneb:
I would take inspiration (as always) from the Gephi project, where they have their Snapcraft config file within the `.github/workflows` subdirectory, which sounds very nice:
https://github.com/gephi/gephi/blob/master/.github/workflows/snapcraft/snap/snapcraft.yaml

kristbaum:
You're right, although I wanted to prevent the additional baggage that comes with including the gnome extension. I tried adding a desktop launcher without it, but so far could not get it working. Will report back if I find a solution, or just give in and include the gnome extension :smile:

wetneb:
Ah, apparently Gephi also moved back to having a `snap` folder at the root of the repository: https://github.com/gephi/gephi/commit/531e81194ee309e1674dd43af72d5c6e995e2be2

antoine2711:
> * the command to start is available from PATH and is called "openrefine" not "refine". We could request an alias for that in the future

Yes @kristbaum, I think that would be great. openrefine is kind of weird when you use OR, and longer to write.

When OR launch, the refine script shows no number for the RAM… (i.e. you have M available…) Why is that, and can we fix that?

Also, I installed snap on a ChromeBook, and although I can run OR from a custom install, snap-OR doesn't work. I think it can't even open a TCP socket (not sure).

Can you help me debug that?

Regards,
   Antoine

kristbaum:
@antoine2711 I think i fixed the RAM issue, still working on the automatic opening of the browser! It's already in the released snap, you can try it out! 

> And although I can run OR from a custom install, snap-OR doesn't work. I think it can't even open a TCP socket (not sure).
> 
> Can you help me debug that?

Sure, what are you referencing with a "custom install"?

antoine2711:
> > And although I can run OR from a custom install, snap-OR doesn't work. I think it can't even open a TCP socket (not sure).
> > Can you help me debug that?
> 
> Sure, what are you referencing with a "custom install"?

We I run from source… In this Linux extension mode of a ChromeBook, you can have all the tools needed, and it works, mostly… But not the snap install (also from the Linux extension mode of ChromeBook).

Regards, Antoine

kristbaum:
I think I found a much better solution to the "browser not opening" problem. The fallback option using xdg-open should now trigger when it detects it runs inside a snap with this line in the starting script: 
`!(System.getenv("SNAP") == null`
As this is a source code change it will not appear in the published snap until it's picked up in the next release blob, but I'm quite sure it should work. 
As a next step I would ask about adding the "refine" alias in the snapcraft forums, although it may be a bit more contentious than I originally thought, as the debian package "argyll" seems to use that command.

wetneb:
Note that OpenRefine is now available via APT in Ubuntu/Debian. I suspect this could make this snap file a lot simpler, no? https://launchpad.net/ubuntu/+source/openrefine/3.5.2-1

antoine2711:
> @antoine2711 I think i fixed the RAM issue, still working on the automatic opening of the browser! It's already in the released snap, you can try it out!

Yes @kristbaum, it works!

```
antoine@penguin:~$ sudo snap run openrefine
You have 2781M of free memory.
https://docs.openrefine.org/manual/installing#increasing-memory-allocation
```

But I still can't run it on my ChromeOS.

```
Picked up _JAVA_OPTIONS: -Duser.home="/root/snap/openrefine/10"
15:31:15.213 [            refine_server] Starting Server bound to '127.0.0.1:3333' (0ms)
15:31:15.231 [            refine_server] refine.memory size: 1400M JVM Max heap: 1468006400 (18ms)
15:31:15.585 [            refine_server] Initializing context: '/' from '/snap/openrefine/10/webapp' (354ms)
15:31:20.748 [            refine_server] Creating new workspace directory /root/snap/openrefine/10/.local/share/openrefine (5163ms)
No protocol specified
Exception in thread "main" java.awt.AWTError: Can't connect to X11 window server using ':0' as the value of the DISPLAY variable.
        at java.desktop/sun.awt.X11GraphicsEnvironment.initDisplay(Native Method)
        at java.desktop/sun.awt.X11GraphicsEnvironment$1.run(X11GraphicsEnvironment.java:102)
        at java.base/java.security.AccessController.doPrivileged(Native Method)
        at java.desktop/sun.awt.X11GraphicsEnvironment.<clinit>(X11GraphicsEnvironment.java:61)
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:315)
        at java.desktop/java.awt.GraphicsEnvironment$LocalGE.createGE(GraphicsEnvironment.java:101)
        at java.desktop/java.awt.GraphicsEnvironment$LocalGE.<clinit>(GraphicsEnvironment.java:83)
        at java.desktop/java.awt.GraphicsEnvironment.getLocalGraphicsEnvironment(GraphicsEnvironment.java:129)
        at java.desktop/java.awt.Window.initGC(Window.java:487)
        at java.desktop/java.awt.Window.init(Window.java:507)
        at java.desktop/java.awt.Window.<init>(Window.java:549)
        at java.desktop/java.awt.Frame.<init>(Frame.java:423)
        at java.desktop/java.awt.Frame.<init>(Frame.java:388)
        at java.desktop/javax.swing.JFrame.<init>(JFrame.java:180)
        at com.google.refine.RefineClient.<init>(Refine.java:473)
        at com.google.refine.Refine.init(Refine.java:124)
        at com.google.refine.Refine.main(Refine.java:110)
```

Got a clue why?

Regards, Antoine

kristbaum:
The "refine" alias has been [granted](https://forum.snapcraft.io/t/refine-alias-for-openrefine/30104). 

> Note that OpenRefine is now available via APT in Ubuntu/Debian. I suspect this could make this snap file a lot simpler, no? https://launchpad.net/ubuntu/+source/openrefine/3.5.2-1

That would work, but would make the release cycle dependent on the package release, which could be slower than the zip files for linux.

kristbaum:
> But I still can't run it on my ChromeOS.
> Got a clue why?
> 
> Regards, Antoine

I'm still not sure, but have found multiple reports of problems with snaps and the Linux environment of ChromeOS. There seem to be some [guides](https://forum.snapcraft.io/t/how-to-install-snap-on-any-chromebook-chromeos-with-crostini-fully-working-guide/15734) on the topic but I haven't found anything relating to your particular problem.

My guess would be that the Linux environment on ChromeOS is normally headless, and confused by the request to use the x11 connection, but sadly this is needed for the desktop support. Are there other snaps you have tried before?

antoine2711:
> My guess would be that the Linux environment on ChromeOS is normally headless, and confused by the request to use the x11 connection, but sadly this is needed for the desktop support. Are there other snaps you have tried before?

K @kristbaum. I will take a look at that. But like I said earlier, OpeRefine runs fine in the Linux environment on ChromeOS, when I launch it through the command line.

Regards, Antoine

kristbaum:
> K @kristbaum. I will take a look at that. But like I said earlier, OpeRefine runs fine in the Linux environment on ChromeOS, when I launch it through the command line.
> 
> Regards, Antoine

AFAIK ChromeOS Linux mode is not a supported environment for snaps. The primary supported ones are listed on the [store listing](https://snapcraft.io/openrefine), and some others in the [docs](https://snapcraft.io/docs/installing-snapd). Maybe you could try [these](https://chromeunboxed.com/install-snap-packages-chromebook-crostini-linux-how-to/) steps too, but I'm not certain it will work. 
In conclusion it may not be possible to offer support to Linux mode ChromeOS users with the snap package, but for the rest it could be an improvement.

kristbaum:
Would it be possible to merge this small change, to improve the usability of the snap? Atm it has about 200 users and the launching of the browser still is manual for them :) I know there are still issues some issues, but the launcher fix would help a lot.

wetneb:
Hi @kristbaum, I hope the release of 3.7.6 fixed the launching problem for Snap users. I just created a repository which is meant to host packages like this one (AppImage, Flatpak, Docker…), could you add your `.snap` config there instead? I'm closing this PR accordingly.

PR #4491, opened 2022-02-01
Update Cell.java

mvipe:
making this file bit more cleaner

Fixes #{issue number here}

- 
- 
-

### DIFF
--- a/main/src/com/google/refine/model/Cell.java
+++ b/main/src/com/google/refine/model/Cell.java
@@ -60,7 +60,7 @@ public class Cell implements HasFields {
     @JsonIgnore
     final public Serializable   value;
     @JsonIgnore
-    final public Recon          recon;
+    final public Recon recon;
     
     public Cell(Serializable value, Recon recon) {
         this.value = value;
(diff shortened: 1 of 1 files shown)


main/src/com/google/refine/model/Cell.java
```diff
@@ -60,7 +60,7 @@
     @JsonIgnore
     final public Serializable   value;
     @JsonIgnore
-    final public Recon          recon;
```
wetneb, line 63: This spacing is intentional I think (so that all the field names are aligned as a column).

main/src/com/google/refine/model/Cell.java
```diff
@@ -124,12 +124,11 @@ public Object getValue() {
...
-            } else {
-                return value.toString();
-            }
-        } else {
-           return null;
+            } 
+            return value.toString();   
         }
+        return null;
+
```
wetneb, line 131: Those changes are valid but it is really a matter of preference, so unless we have a documented guideline around this, I would just keep it as it is, for simplicity.

wetneb (commented):
Thanks for the contribution! However, I am not sure it is really worth doing PRs with cosmetic changes like those. If we want to enforce certain guidelines, I would make sure those are documented and ideally integrated in the IDEs and the CI, otherwise there is no guarantee those changes stay in place.

wetneb:
For the reasons above I think we can close this PR. Your contributions on code style are totally welcome but I would rather integrate them in the form of improvements to our linting setup (which for now is limited to this:
https://docs.openrefine.org/technical-reference/build-test-run#code-style-in-eclipse

PR #4501, opened 2022-02-07
Improve CI structure by migrating to pull_request event

### LINKED ISSUE #3944: New structure for GitHub Actions
We currently build pull requests using the `pull_request_target` event for the sake of being able to embed secrets (to post coverage statistics and ui test results), but this is not great for various reasons:
- this exposes a vulnerability, meaning that our secrets can actually be read by a malicious GitHub user
- if we make changes to the testing infrastructure which affect both the code base and the workflow definitions, then pull requests which are open at that time might start failing spuriously
- the reliance 

Instead, I propose we use the normal `pull_request`. The secrets we need for building pull requests (Cypress project key, code coverage secret) are not actually that sensitive and could be displayed publicly with a bit of obfuscation.

wetneb:
Fixes #3944.

- migrate from `pull_request_target` to `pull_request` event in CI


wetneb:
The duplication in checks will disappear once this is merged.

Issue #3705, opened 2021-03-08
Documentation pull requests cannot be merged without admin privileges


wetneb:
For documentation pull requests we do not run the Java CI jobs, which is good because they are irrelevant. However, since one of them is marked as required to merge the pull request, GitHub does not recognize that the pull request can be merged after having been approved:

![image](https://user-images.githubusercontent.com/309908/110358573-dbdcb600-803c-11eb-8605-c6044272d840.png)

(from #3703 )

This can still be overridden by admin privileges but it would be better not to have to resort to that.
Perhaps this issue could be escalated to GitHub itself.

wetneb:
Since [there is no simple workaround at the moment](https://github.community/t/feature-request-conditional-required-checks/16761/21), I am removing the "Required" marker from this check now. We can trust people who merge PRs to look at those checks.

wetneb:
Actually, this move is not ideal either, because GitHub needs at least one required check to enforce that PRs are up to date with master before merging.

tfmorris:
My thought was to have the documentation PRs still run a mandatory step, but have that step know how to be a no-op for the necessary cases and just succeed without doing anything, but I haven't had a chance to investigate whether there are problems with doing it this way.

wetneb:
Now that we have migrated away from `pull_request_target` to `pull_request`, this should be easily solvable: we no longer need manual checks for pull requests being up to date with master, so we can mark all steps as not being mandatory. If any of the checks fail, it will still mark the pull request as failing CI, which is enough feedback for the maintainers.

wetneb:
I have checked is #4504 that this is solved.

PR #4504, opened 2022-02-08
Remove empty technical reference index

wetneb:
Test PR to check that #3705 is solved.


wetneb:
The build fails, but for a good reason, and the "required" Java 11 check no longer appears, so #3705 is indeed solved!

PR #4505, opened 2022-02-08
Download Windows and Mac JREs automatically during packaging

### LINKED ISSUE #4333: Download JRE on the fly during the packaging process
We currently rely on the `settings.xml` file for the developer to specify where they have downloaded Mac and Windows JREs, to package them into the `.dmg` and `.zip` packages.

Instead, we could download those JREs on the fly in the packaging process. This is [what Gephi does](https://github.com/gephi/gephi/blob/master/modules/application/pom.xml) (using an `antrun` maven goal).

The advantage would be that the developer does not have to care about this step (simplifying our release process greatly).
That does not solve the fact that the JRE URLs must still be regularly updated to make sure we are bundling a fresh JRE at each release.

wetneb:
Fixes #4333.

- integrate inside Maven configuration the steps to download JREs to be embedded in Windows and Mac releases
- remove the `settings.xml` file at the root of the repository which was used to trigger Windows with JRE builds (now done automatically alongside the other artifacts)
- remove one step in the documentation of how to cut out a release (only 17 left, yay!)


wetneb:
Merging, in the hope that snapshots with embedded JRE will now be published too. Will revert if failures occur.

Issue #4078, opened 2021-07-30
There is a vulnerability in  Apache Velocity Engine 1.5,upgrade recommended


QiAnXinCodeSafe:
https://github.com/OpenRefine/OpenRefine/blob/409cf2c30097ae7a5981e941d4bbee49896cd590/main/pom.xml#L236

CVE-2020-13936

Recommended upgrade version：2.3

tfmorris:
@QiAnXinCodeSafe Thanks for reporting this and apologies for taking so long to address it.

PR #4508, opened 2022-02-08
Update Velocity version to patch CVE-2020-13936

mazen160:
Update Velocity version to patch - Sandbox Bypass in Apache Velocity Engine (CVE-2020-13936)

Fixes: https://github.com/OpenRefine/OpenRefine/issues/4078

This has been overlooked for the past year, and it seems to be a valid vulnerability. This PR should resolve the issue by upgrading Velocity to a patched version.

### About CVE-2020-13936 (from NVD DB)
An attacker that is able to modify Velocity templates may execute arbitrary Java code or run arbitrary system commands with the same privileges as the account running the Servlet container. This applies to applications that allow untrusted users to upload/modify velocity templates running Apache Velocity Engine versions up to 2.2.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@
     <!-- v1.15 of commons-codec is needed for Jena 3.14+ - copied from org.apache.jena:jena:pom -->
     <commons-codec.version>1.15</commons-codec.version>
     <commons-compress.version>1.21</commons-compress.version>
-    <velocity.version>1.5</velocity.version>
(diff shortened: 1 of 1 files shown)


wetneb:
The CI seems to indicate that the corresponding artifact cannot be found on Maven Central.

wetneb:
Closing because of the failing CI, feel free to reopen when the problem is fixed.

PR #4510, opened 2022-02-08
Added support for .zst compressed files

### LINKED ISSUE #4058: Add support for importing "zst", "zstd" files
I would like to import a `.zst` or `.zstd` file but currently that file type is not recognized to be imported by OpenRefine.

### Proposed solution
Perhaps using Apache Commons Compress (we already have usage in `ImportingUtilities.java`)
https://commons.apache.org/proper/commons-compress/examples.html#Zstandard

### Additional context
RFC standard: `RFC 8478`
mime type: `application/zstd`
official website: https://facebook.github.io/zstd/
example file:  https://github.com/facebook/zstd/blob/dev/tests/golden-decompression/rle-first-block.zst
example code Apache Druid: https://github.com/apache/druid/blob/8296123d895db7d06bc4517db5e767afb7862b83/core/src/test/java/org/apache/druid/java/util/common/CompressionUtilsTest.java#L366

rachittiwari8562:
Fixes #4058 

- Added `ZstdCompressorInputStream` from `org.apache.commons.compress.compressors.zstandard.ZstdCompressorInputStream` .
- Added a new dependency [zstd-jni](https://github.com/luben/zstd-jni) .

### DIFF
--- a/main/pom.xml
+++ b/main/pom.xml
@@ -145,6 +145,12 @@
   </build>
 
   <dependencies>
+
+    <dependency>
+      <groupId>com.github.luben</groupId>
+      <artifactId>zstd-jni</artifactId>
+      <version>${zstd-jni.version}</version>
+    </dependency>
     <dependency>
       <groupId>javax.servlet</groupId>
       <artifactId>javax.servlet-api</artifactId>
--- a/main/src/com/google/refine/importing/ImportingUtilities.java
+++ b/main/src/com/google/refine/importing/ImportingUtilities.java
@@ -67,6 +67,7 @@ import javax.servlet.http.HttpServletResponse;
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream;
+import org.apache.commons.compress.compressors.zstandard.ZstdCompressorInputStream;
 import org.apache.commons.fileupload.FileItem;
 import org.apache.commons.fileupload.FileUploadException;
 import org.apache.commons.fileupload.ProgressListener;
@@ -723,6 +724,11 @@ public class ImportingUtilities {
                     is.reset();
                 }
                 return new BZip2CompressorInputStream(is);
+            }
+            else if (fileName.endsWith(".zstd") || fileName.endsWith(".zst")
+                ||"application/zstd".equals(mimeType)){
+                InputStream inputStream = new FileInputStream(file);
+                return new ZstdCompressorInputStream(inputStream);
             }
         } catch (IOException e) {
             logger.warn("Something that looked like a compressed file gave an error on open: "+file,e);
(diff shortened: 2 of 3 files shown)


wetneb:
Thanks! I think we currently do not depend on any JNI-based library, so we should probably consider the pros and cons carefully.
The jar itself weighs 5.7 MB, which is non-negligible. I guess if we keep it like this, we are effectively shipping the binaries for all platforms on all platforms. I do not know if Maven offers a way to be more selective.

I will also ask Markus, who maintains our Debian package, if packaging this Java library is doable.

If the addition of this new dependency is not a problem, then we will need some tests to make sure the feature works as expected (which will be especially interesting to run in our CI because of the platform-dependency).

wetneb:
So I guess a natural question would be: is it imaginable to do this with a pure Java dependency instead? I have not checked if that exists. I assume that if the Apache Commons library relies on this one, it must be fairly trustworthy.

rachittiwari8562:
> 

As far as my initial knowledge zstd-jni is the implementaion of zstd algorithm on java language as mentioned here https://facebook.github.io/zstd/ .

wetneb:
One implication of adding JNI-based libraries is a more complicated signing process on Mac (since we need to extract the JAR, sign the dynamic library and put the JAR back together). So I am not really sure this is worth the effort and additional bundle weight, given that ZSTD is not so widespread.
I would be happy to add support for this with a more lightweight pure-Java solution though.

thadguidry:
As I [mentioned in our dev mailing list](https://groups.google.com/g/openrefine-dev/c/j3sTfjJvROo/m/hNpwjZRYBAAJ) ,
Since we are only concerned with uncompressing during import time... would refactoring the PR to use native directly help?
Looks like there is an alternative (fewer features, but still it can decompress?)

https://stackoverflow.com/questions/60974402/lz4-and-zstd-for-java

wetneb:
If the dependency has a compatible license and its jar is not too heavy, then I cannot think of other problems.

thadguidry:
The native one "aircompressor" is Apache-2.0

rachittiwari8562:
I have looked into 'aircompressor' and they are taking bytebuffer as input and providing the same as the output .While as far as I can see , we are using InputStream . Is it a good idea to open a file as byte[] and convert it into bytebuffer then pass it on to the decompressor?.....

rachittiwari8562:
I have tried using the 'aircompressor' and tested it out on a 200mb zst file(600mb csv file) . It is working . Should I create a new branch for that implementation for comparision ... or should I overwrite in this branch only.

wetneb:
As you prefer!

rachittiwari8562:
I have added a new pr from a new branch  #4522.

wetneb:
Thank you!

luben:
zstd-jni also publishes architecture classified jars, e.g. [zstd-jni-1.5.2-1-linux_amd64.jar](https://repo1.maven.org/maven2/com/github/luben/zstd-jni/1.5.2-1/zstd-jni-1.5.2-1-linux_amd64.jar).

PR #4512, opened 2022-02-09
Mac packaging in the CI

### LINKED ISSUE #4507: Run MacOS packaging in CI, on MacOS
With our current setup, we can generate the `.dmg` artifact for MacOS on Linux, but it is not compressed, and the only way to compress it (as far as I know) is to do that inside a MacOS VM. Furthermore, we want to sign the MacOS build, and that can also not be done on another platform.

### Proposed solution

Our MacOS build should be generated inside the CI, by a runner running MacOS. We should take inspiration from [Gephi](https://github.com/gephi/gephi) for that.

### Alternatives considered
None

wetneb:
Fixes #4507.

For now this is an initial draft without code signing, to try out the Mac-based runner.

### DIFF
--- a/.github/workflows/snapshot_release.yml
+++ b/.github/workflows/snapshot_release.yml
@@ -169,9 +169,30 @@ jobs:
     - name: Generate dist files
       run: mvn package -DskipTests=true
 
-    - name: Upload snapshot releases to OSSRH
+  mac_server_tests:
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/checkout@v2.3.4
+
+    - name: Restore dependency cache
+      uses: actions/cache@v2.1.7
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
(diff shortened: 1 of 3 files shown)


wetneb:
Merging to try out the snapshot release signing.

.github/workflows/snapshot_release.yml
```diff
@@ -37,6 +37,11 @@ jobs:
           ref: ${{ github.event.pull_request.head.ref }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
 
+      - name: Set up secrets
+        run: |
+          echo "CYPRESS_RECORD_KEY=$(echo YzE3ZDU4OGItZTBkOC00ZjJmLTg4NjYtNzJmNmFmYmRhNGQxCg== | base64 -d)" >> $GITHUB_ENV
```
thadguidry, line 42: @wetneb Shouldn't we also use GitHub encrypted secrets for Cypress key and Coveralls token?

wetneb: no, because then they would not be available to pull requests, where we need them.

PR #4515, opened 2022-02-10
Update build-test-run.md

trnstlntk:
- Update two Java JDK links to general download page which showcases current version (17 at this moment)

Note: in general, references to the Java JDK seem outdated in this document. It's probably not feasible to keep these up to date at all times so a bit of rewriting to make this less 'expirable' may be good, but the specifics are beyond my knowledge unfortunately.


wetneb:
We could also consider changing the JDK download links to https://adoptium.net/.

PR #4520, opened 2022-02-11
Configurable statement merging in Wikibase extension

### LINKED ISSUE #2116: Ability to mark statement as "update if pre-existing" in the Wikidata schema editor
WIkidata schema editing and upload.

Ability to flag a statement in the schema editor as "statement to add or update". When uploading edits to Wikidta, OpenRefine would first search for an existing statement with the same value, and if an existing statement exists, add qualifiers and sources to the existing statement rather than creating a new/duplicate statement.

Pywikibot or QuickStatements.

### LINKED ISSUE #3383: Ability to update/replace statements
I am looking for a tool to do mass edits to a Wikibase. It would be useful if the same tool could do both adding missing data and updating existing data. Based on my understand, currently OpenRefine is only able to add new items and new statements.

### Proposed solution

**Brute force solution**
In Wikibase schema editor, have a checkbox to replace existing statements for the given property. Internally this would be implemented as deletion of all statements for the subject using that property. This would be sufficient for our case.

**Ideal solution**
"Reconcile" statements. Ideally the Query Service can be used to output statement identifiers that uniquely identify the statement. Otherwise a manual reconciliation is needed (but limited to the statements of the already reconciled item). Then the value (in another column) could be updated in OpenRefine and it would automatically update the existing statement instead of creating a new one.

### Alternatives considered
I think I saw a suggestion to export to quickstatements and duplicate lines so that they are preceded by deletions. But this is rather difficult to do manually.

I think I could write my own tool that processes either the quickstatements v1 format, but it wouldn't know which columns should be overwritten. I could also write my own tool that just takes CSV (from Query Service, then modified) that does this, but for the end user there are already many tools to learn for the data update process (Wikibase, Query Service, OpenRefine, QuickStatements, this new tool?).

### Additional context
https://nimiarkisto.nikerabb.it/w/index.php?title=Item:Q5106574&diff=6759469&oldid=6759468 example of an attempted edit creating a new statement instead of updating it.

#2116 is similar to this request, but per my understanding it is about updating qualifiers rather than the value of the statement.

https://github.com/OpenRefine/OpenRefine/issues/2999#issue-667012515 mentions that deleting statements is currently not possible.

### LINKED ISSUE #4130: Some qualifiers in Wikidata are non-discriminatory and new statements added by OpenRefine should be merged into existing ones with the same value
Whenever OpenRefine tries to import a statement into an item but sees an existing statement of the same property and with the same value, it will merge the statements together, leading to eg. one statement with two references. Situation changes when OpenRefine sees the same value, but qualified with a qualifier. It will then create a new statement, because qualifiers are usually discriminatory - they specify the circumstances under which a statement is true.

However, some Wikidata qualifiers are non-discriminatory, they merely specify some aspects of the value itself. For example, reason for preferred rank (https://www.wikidata.org/wiki/Property:P7452) or reason for deprecation (https://www.wikidata.org/wiki/Property:P2241). These should be always kept but new statements with the same value should be imported into them. 

Example of erroneous behaviour:
https://www.wikidata.org/w/index.php?title=Q2634996&type=revision&diff=1486438858&oldid=1465342427
Fixed like this: https://www.wikidata.org/w/index.php?title=Q2634996&type=revision&diff=1489850300&oldid=1487541248

wetneb:
Fixes #3383. Fixes #2116. Fixes #4130.

- [x] introduce backend-side capabilities for configurable statement merging
- [x] add a dialog to expose this feature in the frontend
- [x] add support for more datatypes for lax value matching
- [x] update constraint checks to react well to deleted statements
- [x] update statement rendering to visually distinguish between added and removed statements
- [x] add user and dev documentation

The UI is arguably quite ugly and I welcome any suggestions to make it nicer.
## Schema editor view
image
## Statement configuration dialog
image


extensions/wikidata/src/org/openrefine/wikidata/commands/PreviewResults.java
```diff
@@ -4,7 +4,7 @@
 
 import org.openrefine.wikidata.qa.QAWarning;
 import org.openrefine.wikidata.qa.QAWarning.Severity;
-import org.openrefine.wikidata.updates.TermedStatementEntityUpdate;
+import org.openrefine.wikidata.updates.TermedStatementEntityEdit;
```
wetneb, line 7: @j-sal in this PR I am renaming (again!) this class to avoid name clashes between our classes (now `Edit`s, before `Update`s) and Wikidata-Toolkit's clases (still `Update`s).
I will try to draw a diagram to explain the relationship between all those layers, because the distinctions are subtle!

wetneb: Here is the diagram: https://deploy-preview-4520--openrefine-docs.netlify.app/next/technical-reference/wikibase/architecture

antoine2711:
> Hi @antoine2711, I would be curious to know what you think of this! Is the UI intuitive enough, for instance?

Hi @wetneb : This is a good start!

image

So, it's a first way to implement it. I guess the first downside I would see is that if there is existing statements in WD, the my OR project might have line with item with qualifiers and some without, so I would have to push 2 times, one with `property-value` and one with `property-value-qualifier`. That logic is also true with `property` and `property-value`.

The second problem is to take this decision, I must know in advance if a qualifier already exist in WD. And, if I can query value thru OR, I can't query qualifiers, and as such, I would have to create a query in WDQS, import it in a new project, and get the qualifiers data thru a well made `cross()` call.

Is it a usefull comment? What should I discuss more?

Regards,
   Antoine

wetneb:
Thanks!

> if there is existing statements in WD, the my OR project might have line with item with qualifiers and some without, so I would have to push 2 times, one with property-value and one with property-value-qualifier

I am not sure I understand the problem. Why cannot you push just once, with the property-value setting? Just to be clear, with the property-value strategy there is nothing preventing you from pushing qualifiers in your statements (they will be added to the matching statement on Wikidata, just like QuickStatements does).

> I can't query qualifiers

I know you that it is a popular request to add support for this. Let me ping you on the corresponding issue to highlight the design questions there.

wetneb:
Merging for the sake of letting people try it out via the snapshot releases. The feature can be revisited once we have more feedback.

PR #4522, opened 2022-02-12
.zst file support via java library(aircompressor)

### LINKED ISSUE #4058: Add support for importing "zst", "zstd" files

rachittiwari8562:
Fixes #4058

As discussed over #4510  , In this PR a native java library [aircompressor](https://github.com/airlift/aircompressor/tree/master/src/main/java/io/airlift/compress/zstd) is used for adding support of the .zst files .


wetneb (commented):
Thanks a lot for this new version!
With this approach, we are loading the whole file in memory (in the byte buffer). I wonder if there would be a way to use this library, but still read the file in streaming mode? That would avoid running out of memory for large files.
Also, we should have some accompanying tests to demonstrate the feature.

rachittiwari8562:
>Also, we should have some accompanying tests to demonstrate the feature. 

For the tests can you guide me to some previous tests for files such as (.gz) . Currently I don't know what type of test I am expected to write .

rachittiwari8562:
> I would recommend to just test the tryOpenAsCompressedFile method, by providing it a very small file compressed in ZSTD, and checking that the returned stream contains the decompressed contents.

Do I have to write a server test(TestNG) or a client test(Cypress) ?

wetneb:
I would write a server test, it is more lightweight, but it is true that a Cypress test could do too :)

wetneb:
Ah, it is very interesting that the test passes on Linux but fails on Windows. (It is rare that we pick up any OS-specific bug with the CI, so that makes me happy!)

rachittiwari8562:
> Ah, it is very interesting that the test passes on Linux but fails on Windows. (It is rare that we pick up any OS-specific bug with the CI, so that makes me happy!)

I have looked into this ...it seems quite intersting . In windows apparently the uncompressed byte stream returned by the zstd decompressor is 115bytes long  which is different from ubuntu(111 bytes) .

thadguidry:
> > Ah, it is very interesting that the test passes on Linux but fails on Windows. (It is rare that we pick up any OS-specific bug with the CI, so that makes me happy!)
> 
> I have looked into this ...it seems quite intersting . In both the OS (windows and ubuntu) zstd is returning 111 bytes long data ..In windows the uncompressed zst_test.csv file is read as 115 bytes . Don't know what causing this odd behaviour.

Inspect the bytes themselves...use your eyes...when looking at the decompressed stream returned, which bytes are missing as comparing it to the raw original csv file? What bytes changed or are missing? Perhaps `\n\r` or `CR/LF` (2 bytes) or LF (1 byte)? Remember that in 8-bit encodings and UTF-8 that `\n\r`  is 2 bytes.  A `\r` is 1 byte.  Line endings are different by default between Windows and Unix systems, but default line endings can be changed and thus Windows can store files with only `\r` and still interpret them.  Often when writing or storing a file, it's typically best to write in binary.  I would assume aircompressor does this, but does your method of compressing do this? :-)  https://stackoverflow.com/questions/3776923/how-can-i-normalize-the-eol-character-in-java

That should give you some quick hints and context to think about ;-)  
You can use a text editor that displays differences as bytes. Or an IDE plugin.

wetneb:
Let's close this and reconsider when (if?) a library with streaming support is available.

PR #4523, opened 2022-02-12
Extraction of the Operation History list as a downloadable JSON file

### LINKED ISSUE #4498: Exporting 'Extract operation history' as JSON
Currently if I want to export the Extract Operation History list of actions, I manually copy paste the JSON and save as a .json file. Would it be possible to add an Export button with a traditional 'save as' dialog?

nicktz1408:
Closes #4498 

- Enabled the extraction of the Operation History list of actions, which is a JSON object, by downloading as a JSON file, called "history.json"
- On the Extract operation history dialog, added a button that can be used to initiate the above task, and placed it on the same level as the "Select all", "Unselect all" buttons, with a right float (to be placed close to the JSON textarea).

For the purpose of creating the downloadable file and actually downloading it on the user's computer, I used a small utility function from the following StackOverflow answer: https://stackoverflow.com/a/18197341/5564816
This approach works with HTML5, however, it could be prone to potential compatibility issues with old browsers.

If you have a better approach to accomplish this task, please share them with me in order to adapt and use them on the code! Also, as this is my first PR, please forgive me if I have missed anything else and I would really appreciate your feedback!

### DIFF
--- a/main/webapp/modules/core/scripts/project/history-panel.js
+++ b/main/webapp/modules/core/scripts/project/history-panel.js
@@ -199,6 +199,7 @@ HistoryPanel.prototype._showExtractOperationsDialog = function(json) {
   elmts.or_proj_extractSave.html($.i18n('core-project/extract-save'));
   elmts.selectAllButton.html($.i18n('core-buttons/select-all'));
   elmts.unselectAllButton.html($.i18n('core-buttons/unselect-all'));
+  elmts.saveJsonAsFileButton.html($.i18n('core-buttons/export'))
   elmts.closeButton.html($.i18n('core-buttons/close'));
 
   var entryTable = elmts.entryTable[0];
@@ -254,6 +255,23 @@ HistoryPanel.prototype._showExtractOperationsDialog = function(json) {
     frame.find('input[type="checkbox"]').prop('checked', false);
     updateJson();
   });
+  elmts.saveJsonAsFileButton.click(function() {
+    var historyJson = elmts.textarea[0].value;
+
+    downloadFile('history.json', historyJson);
+  });
+
+  var downloadFile = function(filename, content) {
+    var element = document.createElement('a');
+    element.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(content));
+    element.setAttribute('download', filename);
+
+    element.style.display = 'none';
+    document.body.appendChild(element);
+
+    element.click();
+    document.body.removeChild(element);
+  }
 
   var level = DialogSystem.showDialog(frame);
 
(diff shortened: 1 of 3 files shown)


main/webapp/modules/core/scripts/project/history-panel.js
```diff
@@ -254,6 +255,23 @@ HistoryPanel.prototype._showExtractOperationsDialog = function(json) {
     frame.find('input[type="checkbox"]').prop('checked', false);
     updateJson();
   });
+  elmts.saveJsonAsFileButton.click(function() {
+    var historyJson = elmts.textarea[0].value;
+
+    downloadFile('history.json', historyJson);
+  });
+
+  var downloadFile = function(filename, content) {
```
wetneb, line 264: It could be worth adding a comment to attribute the author (StackOverflow content is CC-BY-SA).

wetneb (approved):
Very nice, that is a simple solution.

I guess the addition of a phantom DOM element is maybe not so elegant but if we wanted to turn the button into an `a` element so that it plays this role instead, we would have to update its `href` every time the JSON contents change, which is perhaps a bit heavy.

Just a small nitpick below.

wetneb:
Ah, also, if you feel like it, you could add a UI test for this feature:
https://docs.openrefine.org/technical-reference/functional-tests

main/webapp/modules/core/scripts/project/history-panel.js
wetneb: Absolutely!

nicktz1408: Done!

I have a question about commits. What is a better way to write commits? Can I use the format: closes #1234 - (short description of the commit) for each commit I produce related to this issue? Also, if we need to update the commit messages, shall I use "git push -f"?
Thanks! :)

wetneb: It is a good practice to include "closes #4498" in your commit message, but preferably after some informative description of the commit. You can also put "closes #4498" further down the commit message (not necessarily on the first line).
For instance:
```
Add download button for JSON history.

Closes #4498.
```

You can use `git push -f` on your PR branches but it is not necessary, we squash commits when merging pull requests. (Personally I even like it when people just do more commits and not force-push, that helps me review the changes since my last review)

wetneb (approved):
Amazing, thanks a lot for the test!
The failing check in the CI is clearly unrelated, sorry about that instability.

main/webapp/modules/core/scripts/project/history-panel.js
nicktz1408: Thanks so much for your detailed explanation! I will keep this noted! :)

PR #4528, opened 2022-02-14
Turn column name parsing options of CSV/TSV importer into radio buttons

### LINKED ISSUE #4299: CSV importer: confusing behaviour of checkboxes
In the configuration options for the CSV/TSV importer, two checkboxes are linked as if one of them had to be checked.

This is a bad pattern: if those checkboxes are really meant to be grouped, then they should be radio buttons and be positioned next to each other.

### To Reproduce
1. First, open a CSV/TSV file
2. Then, click on the "Parse next 1 lines as column headers"

### Current Results

import_options

### Expected Behavior

In this case I was actually trying to get both checkboxes to be not checked (so that my project gets created with `Column 1`, `Column 2`… as column headers). I can achieve this by keeping the `Parse next…` checkbox on and setting the number of lines to 0 but that is a bit counter-intuitive.

### Versions
 - Operating System: Debian
 - Browser Version: Firefox
 - JRE or JDK Version: 11
 - OpenRefine: master (3.6 SNAPSHOT)

rachittiwari8562:
Fixes #4299 

- Changed "parse next x lines as column header" checkbox to radio button and placing it with custom column names .
- minor replacement of the sections .
- reformated html file .

### DIFF
--- a/main/tests/cypress/cypress/integration/create-project/preview_project.spec.js
+++ b/main/tests/cypress/cypress/integration/create-project/preview_project.spec.js
@@ -171,7 +171,7 @@ describe(__filename, function () {
   });
   it('Tests parse-next of parsing options', function () {
     navigateToProjectPreview();
-    cy.get('input[bind="headerLinesCheckbox"]').uncheck();
+    cy.get('input[bind="columnNamesCheckbox"]').check();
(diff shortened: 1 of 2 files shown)


wetneb (approved):
Thank you very much!

PR #4545, opened 2022-02-18
Documentation command line -x examples

### LINKED ISSUE #3818: Command line options documentation are inconsistent
In https://docs.openrefine.org/manual/running we indicate `-Drefine.headless=true` as the option to add to run OpenRefine in headless mode. But trying it with `./refine -Drefine.headless=true` does not work. Instead `./refine -x refine.headless=true` does.

There seems to be a discrepancy between the syntax on the command line (`-x`) and in `refine.ini` (`-D`). The documentation should make that clear. Ideally we should probably avoid this difference of syntaxes.

### To Reproduce
Run `./refine -Drefine.headless=true`

### Current Results
The syntax is rejected.

### Versions
 - Operating System: Linux
 - OpenRefine: master branch

elebitzero:
Fixes #3818

- Added -x examples in documentation files

Rich diff:
image


wetneb (approved):
Thank you!

PR #4550, opened 2022-02-18
Help and record variable documentation

### LINKED ISSUE #3578: Fix help and documentation for record variables
The current help for the `record` variable in the web client points to https://github.com/OpenRefine/OpenRefine/wiki/Variables#record, but that's just a stub that includes a generic link to the top level of the user manual.

It could be updated to point to https://docs.openrefine.org/manual/expressions#record, but it would be better to just include the documentation in the help, like everything else except the `record` variable.

Also, the manual's documentation could use improvement. For `row.record.cells` it says "The cells of the row" when it's a slice of cells from the record for the given column.

elebitzero:
Fixes #3578

- Updated help page for the record variable in the web client to point to: https://docs.openrefine.org/manual/expressions#record


wetneb:
Nice, thank you! Ideally this URL should point to a versioned page of the documentation, so that the user lands on the page that is relevant to the version they are using, but that can be an improvement for later (we do not really have the infrastructure ready for that).

wetneb (approved):

PR #4557, opened 2022-02-22
Add and Remove All buttons for multiple column selection

### LINKED ISSUE #4282: Allow multiple column selection in "Re-order / Remove Columns" modal
As mentioned on #3289 , the  "Re-order / Remove Columns" modal (accessed from the All > Edit Columns menu) could benefit from enhancement to help with datasets that have a large number of columns that one might not be interested in. In particular, I'd love to see a way to move all of the column names to the "Drop columns here to remove" area on the right so that one can then drag back only the columns one wants. Ideally this would be by allowing selection of multiple column names from the left (including a select-all with usual keyboard shortcuts, like cmd-A or ctrl-A), but if that is too complicated then perhaps just a button to "Remove all".

### Proposed solution
Allow the column names on the left (the "Drag columns to re-order" section) to be bulk-selected before the drag action. This would require some form of highlighting in order to show which columns are selected. The selection mechanism could be shift-click for a range from the first click, and ctrl-click (cmd-click on Mac) for selecting additional columns, as is normal for applications which allow selection of multiple rows (such as file browsers, iTunes, etc). A Ctrl-A (cmd-A on Mac) could select all columns, though this cold be ambiguous if we allowed selection from both the 'reorder' and 'delete' sides.

### Alternatives considered
A simple "Remove All" button so that one drags the desired column names back from the right rather than individually deleting columns from the left. In fact this could be included in the proposed solution if there is no good way to unambiguously select-all with a keyboard shortcut.

elebitzero:
Fixes #4282

- Implemented Remove and Add All buttons for moving multiple columns in the All > Edit Columns menu
all_edit_columns_menu


main/webapp/modules/core/langs/translation-en.json
```diff
@@ -205,6 +205,8 @@
     "core-dialogs/reorder-column": "Re-order / Remove columns",
     "core-dialogs/drag-column": "Drag columns to re-order",
     "core-dialogs/drop-column": "Drop columns here to remove",
+    "core-dialogs/add-all": "Add All",
+    "core-dialogs/remove-all": "Remove All",
```
wetneb, line 209: To keep the capitalization consistent with the rest of the tool, those should be `Add all` and `Remove all` (second word not capitalized).

wetneb (commented):
Thank you! It looks great to me :)

wetneb (approved):
Merging soon unless anyone wants to suggest improvements.

PR #4561, opened 2022-02-23
Grammar fixes in `core-about/definition` string

hughlilly:
- Fixes some grammatical and syntactical errors in the `core-about/definition` string.

### DIFF
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -1,7 +1,7 @@
 {
     "name": "English",
     "core-index/slogan": "A power tool for working with messy data",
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thanks! I am not actually sure where this is displayed, we should run a check for any translations that are not actually used anywhere in the UI…

hughlilly:
> I am not actually sure where this is displayed

I believe it’s in `main/webapp/modules/core/about.html`:
https://github.com/OpenRefine/OpenRefine/blob/8a2eee75a9c0d68cd3b7ddedc06542c4c01acce1/main/webapp/modules/core/about.html#L70-L74

PR #4563, opened 2022-02-25
Java redirect documentation

### LINKED ISSUE #4511: Running OpenRefine on Windows without Java points to Oracle website
When running OpenRefine without Java on Windows, users are redirected to https://java.com/en/download/manual.jsp, which proposes to download Java 8. That is a bad idea, since OpenRefine 3.6 will not be compatible with Java 8 anymore (only Java 11+).

We should consider updating this link (if it is possible) and changing it to a user-friendly alternative.

elebitzero:
Fixes #4511

- Changed URL that Launch4j opens when the user starts OpenRefine without Java to go to https://docs.openrefine.org/manual/installing#java
- Updated documentation at https://docs.openrefine.org/manual/installing#java


docs/docs/manual/installing.md
```diff
@@ -16,11 +16,11 @@ OpenRefine is designed to work with **Windows**, **Mac**, and **Linux** operatin
 
 #### Java {#java}
 
-[Java](https://java.com/en/download/) must be installed and configured on your computer to run OpenRefine. The Mac version of OpenRefine includes Java; new in OpenRefine 3.4, there is also a Windows package with Java included.
+Java must be installed and configured on your computer to run OpenRefine. The Mac version of OpenRefine includes Java; new in OpenRefine 3.4, there is also a Windows package with Java included.
 
-If you install and start OpenRefine on a Windows computer without Java, it will automatically open up a browser window to the [Java downloads page](https://java.com/en/download/), and you can simply follow the instructions there.
+If you want to install Java yourself, you can install a pre-built Java Runtime Environment (JRE) from  [Adoptium.net](https://adoptium.net/releases.html). Or, you can install the Java Development Kit (JDK) from [Java.net](https://jdk.java.net/). Please note that OpenRefine works with Java 8 to Java 17 for OpenRefine 3.5.
```
wetneb, line 21: I would remove the mention of `Java.net`, the page looks really uninviting for users, and a JDK should only be useful for development purposes anyway.
You are editing the page that will become the documentation for OpenRefine 3.6 so you can already mention that it only works with Java 11 to 17.

wetneb (commented):
Thank you! I am not completely sure about redirecting people to the docs rather than Adoptium directly. I guess both have pros and cons.

docs/docs/manual/installing.md
elebitzero: Removed the mention of Java.net and updated the Java version compatibility part

elebitzero:
The redirect to the docs is because users might not know that Adoptium Temerium is Java. Another reason is that the user will get confirmation that they don't have Java installed if they read the docs.

wetneb (approved):

PR #4597, opened 2022-03-08
Rename column popup

### LINKED ISSUE #4590: Rename column should use a normal dialog
The "Rename this column" option in column->Edit Column is using the browser popup:
image

### Proposed Solution
Make a new popup window for "Rename this column" similar to this one:
image

elebitzero:
Fixes #4590 

- Replaced browser popup for rename column with a regular dialog:

image


wetneb (commented):
Very nice, thank you! In addition, I think it would be nice if the text field could be focused when the dialog is set up, so that the user can directly edit the column name without having to click.

Also, since you have been very active on the project recently, I wonder if you would be interested in joining our monthly dev meet-up? The details of the event are here: https://groups.google.com/g/openrefine-dev/c/bYOT50_c3HA (I should advertise it better on GitHub)

elebitzero:
> Very nice, thank you! In addition, I think it would be nice if the text field could be focused when the dialog is set up, so that the user can directly edit the column name without having to click.

I added autofocus to the text field, now it auto focuses to the right of the old column name.

wetneb (approved):

PR #4609, opened 2022-03-11
Add URL remove button

### LINKED ISSUE #4606: Create project -> Web addresses -> add URL needs a remove button
The add URL menu under Create Project doesn't let users remove added URL rows

### Proposed solution
Add a trashcan button to added rows which allows users to remove rows:

image (1)

elebitzero:
Fixes #4606

- Added trashcan button to remove URL rows after the first.


main/webapp/modules/core/scripts/index/default-importing-sources/sources.js
```diff
@@ -100,7 +100,14 @@ UrlImportingSourceUI.prototype.attachUI = function(bodyDiv) {
     }
   });
   this._elmts.addButton.click(function(evt) {
-    self._elmts.buttons.before(self._elmts.urlRow.clone());
+    let newRow = self._elmts.urlRow.clone();
+    let trashButton = $('<a href="javascript:{}" title="Remove this row" bind="removeUrlButton"><span class="ui-icon ui-icon-trash"></span></a>');
```
wetneb, line 104: The `title` could ideally be localized and the `bind=` is maybe not necessary since you are directly binding the event handler below.

wetneb, line 104: The `title` could ideally be localized and the `bind=` is maybe not necessary since you are directly binding the event handler below.

elebitzero: I made the title localized and removed the bind.

Issue #2670, opened 2020-06-02
Check that all i18n messages are used in the interface, in the CI


wetneb:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
It would be nice if the CI could check that all messages ids stored in the translations are actually used in the interface. When changing a part of the UI, it is easy to forget to remove any unused messages. Having this check in place would ensure translators are not wasting their efforts on messages that are never shown.

The script could also check for duplicate message ids in the translation files (#2668).

**Describe the solution you'd like**
A basic prototype for this would just search for each message id in the JS source. There might be cases where we are dynamically constructing the message id, in which case this would falsely report that the message is not used. This could be fixed by adding as a comment all the possible values of the constructed message ids.

**Describe alternatives you've considered**
There might be off-the-shelf tooling for this.

**Additional context**

thadguidry:
Is this the only source path to conduct that search? or other paths as well?
`/main/webapp/modules/core/scripts`

thadguidry:
Is it also the case of keeping keys in sync across languages?  Such as this library for i18next : https://www.npmjs.com/package/i18next-json-sync

wetneb:
For the `main` module the path above looks appropriate (although I haven't checked), but we should also do this for extensions.

Keeping keys in sync across languages is done by Weblate directly.

tfmorris:
Something like this could be used as a starting point:

    find main/webapp/modules/core/scripts -name '*.js' -exec grep -E -o \\\$\\\.i18n\\\(\\\'\(\[\^\\\'\]\+\?\)\\\' {} \; | cut -f 2 -d '(' | sort | uniq -c | sort -r -n | head

Remove one level of escaping for non-shell use (e.g. Python). The regex has a capture group for the key, but the grep utility doesn't support capture groups.

Babi-B:
Hello @wetneb @antoine2711 @tfmorris @thadguidry 

About this issue. If I understand correctly you wish to write a test that ensures that all the keys used for localization are actually being used?

>The script could also check for duplicate message ids in the translation files

I was wondering if this could be implemented in java so that a test is run whenever the `Refine.java` class is run? Or do you wish to implement tests which run when the application is deployed?

wetneb:
> If I understand correctly you wish to write a test that ensures that all the keys used for localization are actually being used?

Yes! If that could be part of the CI, it would be nice.

[...]

PR #4613, opened 2022-03-12
Edit project tags tooltip

### LINKED ISSUE #3979: Edit project tags does not work
When hovering on tags on projects in the Project table there is a hover tip, "Edit project tags".
Clicking on a tag does nothing.

### To Reproduce
1. Click Open Project
2. Then hover on some tags in the Tags column on list of projects
3. Try to click on some tags to edit

### Current Results
Clicking is not allowed.  Not actually editable, but the hover text says so?

### Expected Behavior
However I don't know if it truly is supposed to be clickable in that area to edit.
It might have been an oversight in PR #1357 or a change in Javascript or just now instead should say "Project tags"?
Good news is that clicking on "About" link and editing a projects set of tags within the metadata editor does still work that way.

### Screenshots
image

### Versions
 - JRE or JDK Version: JDK13
 - OpenRefine: OpenRefine 3.5 Beta1

elebitzero:
Fixes #3979

- Removed "Edit project tags" tooltip because it wasn't implemented.


wetneb (approved):
Thank you!

wetneb:
I guess we could remove the associated message from the translations files, so that translators do not translate content that is not shown in the tool. Let us do this as part of #2670.

Issue #3958, opened 2021-06-03
Minified Javascript files in the repository


wetneb:
[Posted by Markus Koschany on the openrefine-dev mailing list](https://groups.google.com/g/openrefine-dev/c/eJbPPYnqJiw):
> The webapp includes several minified Javascript files without providing the
> original non-minified source files. This is acceptable and in accordance with
> the license of those files but it is a reason to reject a package for Debian's
> ftp team because Debian requires the original sources. I suggest to include the
> corresponding non-minified Javascript files as well. Actually I believe
> minified JS files don't improve the performance that much since OpenRefine is
> run on the user's local system. You could remove them completely and use the
> non-minified versions instead. There are some files like date.js which appear
> to be minified Javascript as well despite the missing min.js suffix.

We should therefore reevaluate whether we can use non-minified files instead.
The real problem here is our lack of a proper dependency management system for our frontend, #2423.

thadguidry:
We should be able to use non-minified files since we support local use and don't run an OpenRefine service for the world.
Historical context:  They were minified long ago with the idea that Metaweb/Google "might" provide an online data cleaning service based on OpenRefine, but after 1 year of headbanging they gave up. Instead they began to partner with Trifacta.

Issue #2423, opened 2020-03-17
Improve dependency management for the frontend


wetneb:
**Problem**

We currently ship frontend javascript libraries in the repository itself. These libraries have not been updated for a while and they sometimes get patched in various ways. It would be cleaner to use a proper dependency management system for this, just like we have moved from Ant to Maven for the backend.

**Possibilities**

* npm is the standard javascript package manager. We could try to use this. This would mean that dependencies are configured in a `package.json` and are pulled in at compile time using npm. The PR #2418 demonstrates some of that. This means that in addition to requiring Maven for build, we would also need npm. We should look into downloading npm on the fly in the `refine(.bat)` wrappers, which currently do this for Maven itself.
* webjars offer a way to use Maven's existing dependency management system to fetch Javascript libraries. This would let us use a single dependency management system, but the selection and freshness of packages is a bit more restricted.
* anything else?

thadguidry:
What is the scope here?
1. Users of OpenRefine
2. Developers of OpenRefine

wetneb:
I am not sure what you mean?

thadguidry:
Else's:
https://nodejs.org/en/download/package-manager  <-- Windows users rarely use Chocolatey, only corp's use it for automation.
https://github.com/nodejs/help/wiki/Installation  <-- Installing Node.js via binary archive

However's:
npm strongly suggests to [use a Node version manager rather than a Node installer](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#using-a-node-version-manager-to-install-node-js-and-npm)
I personally like and use [nvm](https://github.com/coreybutler/nvm-windows) on Windows.

thadguidry:
I mean "dependency"... for what?  Running OpenRefine or Building OpenRefine?  Your Problem statement doesn't really state a problem well enough...it assumes the reader of this issue knows.

Problem:
OpenRefine frontend development and testing is currently cumbersome and brittle with keeping up with Javascript library versions.

"I think" what you were trying to explain is that:
1. We would like to make testing OpenRefine easier for frontend development and ensure that it can be more easily tested with different versions of npm.
2. We would like to remove some Javascript dependencies from being stored in our repository?
3. We would like to use a package manager for Javascript for managing dependencies better for frontend development.
4. npm recognizes this and strongly suggests the use of a `Node version manager` instead of a `Node installer` when configuring for a development environment, rather than a regular user environment.

fpompermaier:
[...]

PR #4621, opened 2022-03-17
Frontend dependency management part II

elebitzero:
Part II for #2423.
Partial fix for #3958 (Only suggest package is using minified JavaScript now).

- Added postinstall script.
- Updated package.json.
- Removed unneeded files in externals.
- Changed file paths from externals to 3rdparty.
- Updated refine.bat and refine scripts to call npm install for the build.
- Updated developer documentation
- imgareaselect, suggest (4.3a), and jquery-ui remain in externals.

### Note
- Verified package contents of Windows zip files and confirmed refine.bat run works without errors.
- Linux tested and confirmed to work.
- Mac installer confirmed to build on Linux but not tested.


wetneb (commented):
Very nice, thank you!

We should update our technical documentation to indicate that `npm` is now required to build the project. This could be done in the same PR since the docs are also stored in this repository (for now): this could go in `docs/docs/technical-reference/build-test-run.md`.

Also, concerning the `dependencies.json` file, I did not anticipate that it would be required, it is a tiny bit sad because it will not be very convenient to maintain that alongside the `package.json`, but I guess it is a necessary hurdle as long as we are not using a proper pipeline with (for instance) webpack. So it feels fine for now :)

elebitzero:
> Also, concerning the `dependencies.json` file, I did not anticipate that it would be required, it is a tiny bit sad because it will not be very convenient to maintain that alongside the `package.json`,

Yes, webpack would be great, but that is a bigger change, and it could have some issues with this Jquery app that is allowing UI extensions. Webpack works great with SPA apps like Angular and React.  I will take a look at what is involved with using webpack after the webapp is using package.json for dependencies.

wetneb:
Fabulous, thank you!
Maybe I am missing something but it looks like the `refine` and `refine.bat` scripts has not been updated yet to call `npm`, right? If you intend to do this in a follow-up PR then the documentation update was perhaps premature (but fine to leave it like this of course).

elebitzero:
> Maybe I am missing something but it looks like the `refine` and `refine.bat` scripts has not been updated yet to call `npm`, right?

Yes, I was going to do a part III, but I just went ahead and added the remaining changes to this PR.

elebitzero:
The installer now works for Windows, but still needs to be tested on Linux and Mac.

elebitzero:
The installer is now tested on Linux and confirmed to work, just Mac remains.

refine
```diff
@@ -585,6 +590,7 @@ run() {
         IS_JAR=`ls $REFINE_LIB_DIR | grep openrefine`
         if [ -z "$IS_JAR" ] ; then
             mvn_prepare
+            build_prepare
```
wetneb, line 593: I have added this to make sure we can run `./refine` just after cloning the repository (implicitly building before launching) as this is supported on Linux.

wetneb (approved):

PR #4630, opened 2022-03-23
Wikibase dialog buttons

### LINKED ISSUE #4628: Wikibase dialog buttons should be in the dialog footer
The buttons in the following Wikibase dialogs should be in the footer while the content that is currently in it should be either moved to the top or right above the footer:

image
image
image

elebitzero:
Fixes #4628
- Move login note to be under the "Remember me" checkbox.
- Move dialog buttons to the dialog footer of Wikibase login and logged-in dialogs:

image
image
image

### DIFF
--- a/extensions/wikidata/module/scripts/dialogs/manage-account-dialog.js
+++ b/extensions/wikidata/module/scripts/dialogs/manage-account-dialog.js
@@ -116,7 +116,7 @@ ManageAccountDialog.displayPasswordLogin = function (onSuccess) {
     ManageAccountDialog.displayOwnerOnlyConsumerLogin(onSuccess);
   });
 
-  elmts.loginForm.submit(function (e) {
+  elmts.loginButton.click(function (e) {
     frame.hide();
     let formArr = elmts.loginForm.serializeArray();
     let username = elmts.usernameInput.val();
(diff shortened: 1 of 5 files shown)


extensions/wikidata/module/scripts/dialogs/manage-account-dialog.js
```diff
@@ -116,7 +116,7 @@ ManageAccountDialog.displayPasswordLogin = function (onSuccess) {
     ManageAccountDialog.displayOwnerOnlyConsumerLogin(onSuccess);
   });
 
-  elmts.loginForm.submit(function (e) {
+  elmts.loginButton.click(function (e) {
```
wetneb, line 119: I think it is preferable to bind to the `<form>`'s submit event rather than the login button click event, for accessibility reasons. For instance, if you press "Enter" in one of the input fields, it will trigger a form submit event, but not a button click. So I would rather keep the event binding as it was, and make the `<form>` encompass the buttons in their new location.

elebitzero: Changed the event back to submit.

wetneb (approved):

PR #4635, opened 2022-03-24
Redirect to wiki for Google Export options

### LINKED ISSUE #4631: Export to Google Drive and Sheets is broken
Export to Google Drive and Export to Google Sheets options give the following error when clicked when it  should recognize if a user is logged in, or at least give the option to log in. (Seen on Brave, Chrome, and Edge):

OpenrefineGoogledriveError

elebitzero:
Fixes #4631

- Added redirect to [wiki](https://github.com/OpenRefine/OpenRefine/wiki/Google-Extension#missing-credentials) for when the Google export to Drive/Sheets options are clicked while the Google API credentials are undefined.

### DIFF
--- a/extensions/gdata/src/com/google/refine/extension/gdata/GoogleAPIExtension.java
+++ b/extensions/gdata/src/com/google/refine/extension/gdata/GoogleAPIExtension.java
@@ -60,10 +60,8 @@ abstract public class GoogleAPIExtension {
     static public String getAuthorizationUrl(ButterflyModule module, HttpServletRequest request)
             throws MalformedURLException {
         String host = request.getHeader("Host");
-        if (host.startsWith("127.0.0.1")) {
-            if (CLIENT_ID.equals("") || CLIENT_SECRET.equals("") || API_KEY.equals("")) {
-                return "https://github.com/OpenRefine/OpenRefine/wiki/Google-Extension#missing-credentials";
-            }
+        if (CLIENT_ID.equals("") || CLIENT_SECRET.equals("")) {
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thank you! I wonder why the host check is necessary - intuitively it makes sense to redirect whatever the host is, no?

elebitzero:
> I wonder why the host check is necessary - intuitively it makes sense to redirect whatever the host is, no?

I scoped the logic to http://127.0.0.1/ host only, because you mentioned that the instructions were for the development version only.  So, I thought we should only redirect to the GitHub wiki for the development version.

Is there another way to detect the development version, in case they have changed the host? And how does it work for non-development?

wetneb:
It's up to you, but I guess the check could fail if accessing via `localhost` or `[::1]` in IPv6, or some other URL if working with Docker… I do not think it hurts to redirect users there either, if credentials are not supplied: it is still better than redirecting them to a Google error.

elebitzero:
I installed the official OpenRefine 3.5.2 release on Windows 10 and ran "refine.bat run" which opens the browser to http://127.0.0.1:3333/, which is the same as the development version.

But the Google Export now prompts for my Google account without any error, and without having to set client_id, client_secret, or api_key.

Right now, I don't feel comfortable checking this change in without understanding how it works for the production version.  Since the production software can also use http://127.0.0.1:3333/,  my check may redirect to the wiki when it would work.

wetneb:
Yes, in the packaged software, CLIENT_ID and CLIENT_SECRET are inserted. So the check above will work: it will not be triggered for official releases which have been properly set up.

elebitzero:
Looks good, should be ready to merge.

PR #4642, opened 2022-03-27
modify default return of call() for ToString to state for…

### LINKED ISSUE #3133: Improve accuracy of returned error message for toString()
`toString()` can also be used for Number formatting (not only Date formatting) as evidenced by @tfmorris on the mailing list today.

We should improve the `toString()` error message output here:
https://github.com/OpenRefine/OpenRefine/blob/ded6ca07d1fb1f88f1b7f9bae7e1de6d3c22e311/main/src/com/google/refine/expr/functions/ToString.java#L67

replacing the line with this would help more:

```java
return new EvalError(ControlFunctionRegistry.getFunctionName(this)  +
    " accepts an object and an optional second String argument containing a date format or number format");
```

elroykanye:
… Number format

Fixes #3133

- The default EvalError constructor arg has been modified to state that the ToString call() function's second optional parameter can be a Date or Number format string.

### DIFF
--- a/main/src/com/google/refine/expr/functions/ToString.java
+++ b/main/src/com/google/refine/expr/functions/ToString.java
@@ -64,7 +64,7 @@ public class ToString implements Function {
                 } 
             }
         }
-        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " accepts an object and an optional second argument containing a date format string");
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " accepts an object and an optional second argument containing a Date or Number format string");
     }
 
     


wetneb:
Thanks! It would be great if you could move the "fixes #3133" from the title of your pull request to the body text (instead of `Fixes #{issue number here}` which is part of the pull request template)

elroykanye:
> Thanks! It would be great if you could move the "fixes #3133" from the title of your pull request to the body text (instead of `Fixes #{issue number here}` which is part of the pull request template)

I totally didn't see that 😄, my apologies

wetneb (approved):

elroykanye:
Oh! thanks, I was just about to edit the name 😁

elroykanye:
I noticed something I wish to ask. These pull request rules are necessary because some action auto links the PR to an issue so it closes it right? (I am still quite new to github actions)

PR #4651, opened 2022-03-28
added language detection

### LINKED ISSUE #642: Add language detection
A function to detect the language of a piece of text would be useful.  There's a high performance (99% accuracy), Apache licensed implementation in Java here: https://github.com/shuyo/language-detection (originally http://code.google.com/p/language-detection/)

elroykanye:
Fixes #642

- Changes to pom.xml in the `main` module to include dependencies for [language detector](https://github.com/optimaize/language-detector)
- Utils class to perform detection
- DetectLanguage Function implementation for detecting and return language code.


elroykanye:
I didn't have much of a reason for picking this. 
But among the ones I was able to find following the tree created from this [project](https://github.com/shuyo/language-detection), this was the first best suited 😄. 
I faced problems with the others before this because they had no maven artefacts. Working with them gave a lot of trouble when running the tests or loading the classes during runtime (NoSuchClassDef issues). 

As for memory concerns, reading through [this](https://github.com/optimaize/language-detector/wiki/Memory-Consumption) actually made me realise how important it is to consider how resources an application should be designed to follow. Before now, I never really watched out for this and only just wrote code that works 😬

I will look at the alternatives you have suggested and those stated in [this](https://javarepos.com/lib/pemistahl-lingua#2--why-does-this-library-exist-top-), including the memory footprints and report back to you with updates here.

elroykanye:
Hello @wetneb , 

My reading through Apache Tika gave me this:

> `Apache Tika detects only 18 languages as there are 184 standard languages standardized by ISO 639-1.`

I did tests with tika, lingua and optimaize in a separate empty project. Currently, I have difficulty finding a way to include the Apache TIka Lang Detect library using maven. What I did then was that I got a 188KB jar of tika-langdetect-tika-2.0.0. 
Comparing with other libraries, I think the accuracy and scope of tika is quite low. 

I thought and pondered too, that if it is to be a dependency, will all of the tika library be added as a dependency or just the this langdetect module?

From what I see concerning this case of language detection, optimaize is a good option.

main/src/com/google/refine/util/LangDetectUtils.java
```diff
@@ -0,0 +1,38 @@
...
+import com.optimaize.langdetect.text.TextObject;
+import com.optimaize.langdetect.text.TextObjectFactory;
+
+import java.io.IOException;
+import java.util.List;
+
+public class LangDetectUtils {
+
+    public static Optional<LdLocale> detect(String text) throws IOException {
+        List<LanguageProfile> languageProfiles = new LanguageProfileReader().readAllBuiltIn();
```
wetneb, line 21: Here you are loading all the languages, which I expect can take a bit of time. Since the language detection function will typically be called many times (for instance once per row in the dataset), it would be good to do that just once, the first time the function is called.

main/tests/server/src/com/google/refine/util/LangDetectUtilsTest.java
```diff
@@ -0,0 +1,63 @@
...
+        );
+
+        sampleTexts.forEach(sample-> {
+            Optional<LdLocale> lang;
+            try {
+                lang = LangDetectUtils.detect(sample.getText());
+                if(lang.isPresent()) {
+                    assertEquals(lang.get(), sample.getLdLocale());
+                    // print out that the language was detected
+                    System.out.println(sample.getText() + " was detected as " + lang.get().getLanguage());
```
wetneb, line 28: Those `System.out.println` calls should not be needed in tests.

main/tests/server/src/com/google/refine/util/LangDetectUtilsTest.java
```diff
@@ -0,0 +1,63 @@
...
+                } else {
+                    System.out.println(sample.getText() + " was not detected");
+                }
+            } catch (IOException e) {
+                e.printStackTrace();
```
wetneb, line 33: Make sure the test fails in that situation (printing a stack trace will not mark the test as having failed).

main/tests/server/src/com/google/refine/util/LangDetectUtilsTest.java
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+
+        });
+    }
+}
+
+class SampleTexts {
```
wetneb, line 41: We generally try to have only one class per file, so you could add this as a static inner class of your main test class.

wetneb (commented):
Ok! Then here are some comments on your changes.

main/pom.xml
```diff
@@ -348,6 +348,24 @@
...
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
```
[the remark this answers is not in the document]
elroykanye: I initially didnt't have those, but without them I had issues running the project. However, I realised a workaround without those dependencies while working on another issue and I am working to fix that :)

main/src/com/google/refine/util/LangDetectUtils.java
elroykanye: I see. Thanks for this!

main/tests/server/src/com/google/refine/util/LangDetectUtilsTest.java
elroykanye: 😬 Noted, thanks. :)

main/tests/server/src/com/google/refine/util/LangDetectUtilsTest.java
elroykanye: Very true! 😁 Noted

main/tests/server/src/com/google/refine/util/LangDetectUtilsTest.java
elroykanye: Noted!

elroykanye:
Thanks very much @wetneb . Will do all necessary arrangements :)

main/src/com/google/refine/util/DetectLanguageUtils.java
```diff
@@ -15,11 +14,18 @@
...
 
-public class LangDetectUtils {
+public class DetectLanguageUtils {
+    private static List<LanguageProfile> languageProfiles;
 
-    public static Optional<LdLocale> detect(String text) throws IOException {
-        List<LanguageProfile> languageProfiles = new LanguageProfileReader().readAllBuiltIn();
+    static {
+        try {
+            languageProfiles = new LanguageProfileReader().readAllBuiltIn();
```
wetneb, line 22: Having this in a static block will indeed load the language profiles only once, which is better, but I think it would be even nicer if that could only happen the first time the GREL function is used.

elroykanye: 🤔 ah! I get it so clearly now!
I'll work something out 😃

wetneb (commented):
Nice, thank you! I think we are nearly there. How about adding documentation about this new function in the user manual? You could do this in this pull request already.

elroykanye:
Alright, will do.

wetneb:
@tfmorris are you happy if we merge this once documented, or do you feel like we need a more formal review of other libraries before proceeding?

docs/docs/manual/grelfunctions.md
```diff
@@ -252,6 +252,12 @@ Returns an array of strings describing each character of s in their full unicode
 
 Returns an array of strings describing each character of s by their unicode type. For example, `"Bernice Rubens".unicodeType()` outputs [ "uppercase letter", "lowercase letter", "lowercase letter", "lowercase letter", "lowercase letter", "lowercase letter", "lowercase letter", "space separator", "uppercase letter", "lowercase letter", "lowercase letter", "lowercase letter", "lowercase letter", "lowercase letter" ].
 
+### Translating
+
+###### detectLanguage(s) {#detectlanguages}
+
+Returns a string locale describing the language of s, with support for 71 languages. For example, `"Hello, world!".detectLanguage()` outputs "en".
```
wetneb, line 259: Brilliant! Perhaps a link to the list of the supported languages and their identifiers would be nice (does the library provide that?)

elroykanye: It does. Coming up in a few :)

wetneb (approved):
I think this is good to go! Leaving it open a bit more to leave the opportunity for others to chime in, but as far as I am concerned this could be merged as is.

antoine2711:
Good work @elroykanye. Thank you. 

Regards, Antoine

tfmorris:
> @tfmorris are you happy if we merge this once documented, or do you feel like we need a more formal review of other libraries before proceeding?

Sorry I missed this ping. My recommendation was here: https://github.com/OpenRefine/OpenRefine/issues/642#issuecomment-1095797737 specifically: 

> I would suggest drawing up a list of evaluation criteria by which to measure implementations and then scoring all the current reasonable alternatives against that list of criteria. Criteria might include # of languages supported, model size, run time performance, implementation language, etc. You might also want to solicit input from the user community, since it's a relatively important implementation decision.

Given that the implementation was/is done, something is better than nothing, but I still think it would be useful to evaluate what the best language identification implementation is for OpenRefine's purposes. They all have different strengths/weaknesses.

tfmorris:
Actually, scrolling back up, @wetneb raises most of the concerns that I would have raised, although I'd also add that the chosen library appears to have been abandoned with no updates since 2015.

The statement about Tika only supporting 18 languages isn't correct. As best I can tell from [this CSV](https://github.com/pemistahl/lingua/blob/main/accuracy-reports/aggregated-accuracy-values.csv) the language counts and accuracy scores for the various languages are as follows. I also added a couple from the [other evaluation](https://modelpredict.com/language-identification-survey) that I previous linked to. The fasttext scores are actually for the fasttext-compressed variant since the size of the uncompressed version is 136 MB which seems excessive.

|                   | Optimaize | OpenNLP | Tika | Lingua Low | Lingua Hi | fasttext | CLD2 | CLD3|
|-----------|-----------|-----------|-----|-------------|----------|---------|-------|------|
|Languages|56 | 68 | 56 | 76 | 76| 176 | 83 | 107 |
| Sentence Accuracy| 93.4 | 94.9 | 96.1 | 93.3 | 95.9 | 97.9 | 87.1 | 87.0 |
| Word Pair Accuracy| 60.8 | 74.1 | 80.8 | 78.5 | 89.2 |

boxplot-average

One of the things that Lingua is optimized for is short texts (e.g. tweets) which is why that evaluation includes scores for both single words and word pairs, where it excels due to the special focus.

I think it would be worth having a discussion about what the evaluation criteria should be and what's the best fit for OpenRefine. The linked evaluation closes with: 

>The benchmark shows that different libraries offer different tradeoffs in terms of language coverage, accuracy, speed, and memory footprint.
>
>Surprisingly, langdetect, the oldest model in the benchmark, performs surprisingly well in terms of accuracy, but is so slow (~1100x slower than pycld2) that it serves better as an accuracy benchmark than a tool used in production.
>
>Pycld2 is the fastest one with a tiny memory footprint, but also performs at the lower end of accuracy.
>
>Fasttext sits in the middle. It’s very fast (~120k sentences/s on MBP M1), it covers the biggest number of languages, and has the highest accuracy on all three datasets. It also provides two different models so you can trade off smaller memory footprint for a tiny accuracy hit.
>
>Overall, fasttext seems to be a great default choice for the language identification task whenever you don’t have time or a labeled dataset to benchmark it on.

PR #4655, opened 2022-03-29
fixes #3172 Reconcile > Use values as identifiers" does not reconcile

### LINKED ISSUE #3172: "Reconcile > Use values as identifiers" does not reconcile
### To Reproduce
Run "Reconcile" > "Use values as identifiers" on any column, whether it contains properly-formatted unique IDs for the given service or not. You will see that it seems not to reconcile at all - it only says, whatever your cell contents are, it's 100% matched and generates a real-looking URL, whether or not that URL will resolve. 

### Expected Behavior
Reconciliation seems to be missing the part where it actually validates these unique IDs to see if they match up with existing entities. I expect failures to say so. 

### Screenshots
error1

I tested on Wikidata and VIAF. VIAF gives no hover-information, the matches just send you to 404s, e.g. http://viaf.org/viaf/Q17291. None of the content in that VIAF column should be matching (Q### is an invalid format) except that first value, which resolves correctly to https://viaf.org/viaf/38242123/.  

There are two things happening on Wikidata that I thought I'd mention - IDs that don't exist give the error in the above screenshot. IDs that seem to exist but have yet to be assigned (something that fits the Q### format) look a bit different (so that may be a second bug to work out). I would expect these to have some other kind of obvious error message or flag. 
error2

### Versions
Windows, Firefox
 - OpenRefine: the new 3.4 release, as well as 3.4 beta 2 (last week)

WaltonG:
Fixes #3172

- Added reconciliation identifiers notification
- 
-


WaltonG:
@wetneb Am checking it out

main/webapp/modules/core/scripts/reconciliation/recon-dialog.html
```diff
@@ -12,6 +12,8 @@
 				</div>
 				<div class="recon-dialog-service-panel-container"
 					bind="servicePanelContainer">
+					<div class="recon-dialog-service-panel-message-note"
+						bind="servicePanelMessageNote"></div>
```
wetneb, line 16: When using this dialog, identifiers are indeed validated, so we should not add this note in this dialog. The operation which does not validate its identifiers is the "Use values as identifiers" operation.

wetneb (commented):
Thanks! It looks like your PR is still deleting two unrelated files (in `server/IDEs/eclipse`).

main/webapp/modules/core/scripts/reconciliation/recon-dialog.html
WaltonG: How did I miss that, and its clearly stated on how to reproduce the issue . Working on it.

> Thanks! It looks like your PR is still deleting two unrelated files (in `server/IDEs/eclipse`).

Will try and figure out how those files are generated. It could be my eclipse ide

WaltonG:
Hi @wetneb, I have submitted a pull request with the recommended changes

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -140,7 +140,7 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
     var frame = DialogSystem.createDialog();
     frame.width("400px");
 
-    var header = $('<div></div>').addClass("dialog-header").text($.i18n('core-views/use-values-as-identifiers')).appendTo(frame);
+    var header = $('<div></div>').addClass("dialog-header").text($.i18n('core-views/use-values-as-identifiers-note')).appendTo(frame);
```
wetneb, line 143: As mentioned above, I do not think we should be making any changes to this dialog in this pull request.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -409,7 +409,7 @@
     "core-recon/error-contact": "Error contacting recon service",
     "core-recon/wd-recon-lang": "en",
     "core-recon/recon-col": "Reconcile column",
-    "core-recon/pick-service": "Pick a service or extension on left",
+    "core-recon/pick-service": "Pick a service or extension on left.",
```
wetneb, line 412: I would leave this message untouched since it is used in a dialog that is unrelated to the problem we are trying to solve.

wetneb (commented):
Thanks! However the files in `server/IDEs/eclipse` are still deleted by your pull request.

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
WaltonG: The same has introduced a conflict with master. Am currently working on a different approach of displaying the message

main/webapp/modules/core/langs/translation-en.json
WaltonG: The message is untouched. The deletion is the message I had included earlier before the first review.

wetneb:
Hi @WaltonG, to see the files that are modified by this PR, you can have a look at the "Files changed" tab of this PR: https://github.com/OpenRefine/OpenRefine/pull/4655/files

If you are struggling with restoring those files to their original content, then you could simply start over from a fresh branch and open a new PR.

WaltonG:
Am for the option of submitting a new pull request. Working on it now

wetneb:
Then let us close this one.

Issue #4585, opened 2022-03-02
Menu items that open popups and new windows should have ellipsis


elebitzero:
Some of the menu items don't have an ellipsis but should because they open a popup, file explorer, or new window.

The following options need ellipsis:

All->Edit all columns->Transform 
All->Edit all columns->Trim leading and trailing whitespace
All->Edit all columns->Collapse consecutive whitespace
All->Edit all columns->Unescape HTML entities
All->Edit all columns->Replace smart quotes with ASCII
All->Edit all columns->To titlecase
All->Edit all columns->To uppercase
All->Edit all columns->To lowercase
All->Edit all columns->To number
All->Edit all columns->To date
All->Edit all columns->To text
All->Edit all columns->To null
All->Edit all columns->To empty string
Column->Facet->Custom text facet on column 
Column->Facet->Custom numeric facet on column 
Column->Facet->Scatterplot facet
Column->Edit cells->Replace 
Column->Edit Column->Add column by fetching URLs based on column X 
Column->Edit Column->Add column based on column X 
Column->Edit Column->Split column X into several columns 
Column->Edit Column->Join columns  
Column->Edit Column->Rename this column
Column->Transpose->Transpose cells across columns into rows... 
Column->Transpose->Transpose cells in rows into columns... 
Column->Transpose->Columnize by key/value columns... 
Column->Reconcile->Start reconciling
Column->Reconcile->Use values as identifiers
Column->Reconcile->Add entity identifiers column
Export->Google Sheets
Export->Wikibase schema
Wikibase->Select Wikibase instance
Wikibase->Manage Wikibase account
Wikibase->Import schema
Wikibase->Upload edits to Wikibase

### Proposed solution
Add ellipsis to the necessary menu items.

thadguidry:
@elebitzero Yeap, agree, I was just thinking about this today as I was testing our 3.6 Snapshot build with your earlier merged changes for the alignment of dialog buttons.  I'm wondering on this particular issue if we might use the Unicode ellipsis char and bake this into the menu's that need them via a JS variable?  

<html><body>


Name: | Horizontal Ellipsis
-- | --
PostScript Name: | `ellipsis`
HTML Entity: | `&#8230; &#x2026; &mldr;`
UTF-8 Encoding: | `0xE2 0x80 0xA6`
UTF-16 Encoding: | `0x2026`


</body>
</html>

[...]

PR #4661, opened 2022-03-29
Fixed dialog headers with ellipsis

elebitzero:
Fixes regression introduced from #4585 where the titles of some dialogs had ... since the headers used the same translation as the menu option.

- Added separate translations for the headers of those dialogs.


wetneb (approved):
Thanks a lot!

PR #4663, opened 2022-03-30
fix OpenRefine#4656 - clarify location of workspace

### LINKED ISSUE #4656: Unclear documentation in "Set where data is stored" of installing section
**Problem 1**: 
It is confusing where to find the workspace when first reading the user manual.

The reason is as below (screenshots below):

- The first appearance of the term "workspace directory" in section _Back up your data_ doesn't point to the correct section #**set-where-data-is-stored**, which now points to #where-is-data-stored. 

-  But **Set where data is stored** section's instruction is not clear enough.

- The clearest guidance to find the workspace is under section **Find the right place to install**

**Problem 2**: the default location of program files cannot be reached. As there is a space in the path, copying and pasting this path in the terminal with cd command gives 

> string not in pwd: Application error.

`~/Library/Application Support/OpenRefine/`

### To Reproduce
1. First, go to [set where data is stored](https://docs.openrefine.org/manual/installing#set-where-data-is-stored)
2. Then, copy path and paste in the terminal.

### Current Results

1. Cannot find workspace directory
2. Cannot locate the program directory using the path.

### Expected Behavior

1. Find program files in the web app.
2. Find program files in the local directory

### Screenshots

image

image

image

### Versions
 - Operating System:  iOS
 - Browser Version:  Chrome 99
 - JRE or JDK Version:  openjdk version "17.0.2"
 - OpenRefine:  3.5.2 [e3efd4e]

SIXIN-LI:
Fixes #4656

- Correct the inline link of "locate your workspace directory" in _**_Installing or upgrading --Back up your data_**_ section.
- Reconfigure the guidance to find workspace in section **Set where data is stored**
- Simplify steps to install extensions in section **Find the right place to install**
- Modify the path to be reachable

### DIFF
--- a/docs/docs/manual/installing.md
+++ b/docs/docs/manual/installing.md
@@ -62,7 +62,7 @@ OpenRefine may also work in other environments, such as [Chromebooks](https://gi
 
 If you are upgrading from an older version of OpenRefine and have projects already on your computer, you should create backups of those projects before you install a new version.
 
-First, [locate your workspace directory](#where-is-data-stored). Then copy everything you find there and paste it into a folder elsewhere on your computer.
+First, [locate your workspace directory](#set-where-data-is-stored). Then copy everything you find there and paste it into a folder elsewhere on your computer.
 
 For extra security you can [export your existing OpenRefine projects](exporting#export-a-project).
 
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thanks a lot!

PR #4664, opened 2022-03-30
fixes #3172 Reconcile > Use values as identifiers" does not reconcile #4655

### LINKED ISSUE #3172: "Reconcile > Use values as identifiers" does not reconcile

WaltonG:
Fixes #{3172}
Notice added when selecting service


WaltonG:
@wetneb I submitted anew pull request https://github.com/OpenRefine/OpenRefine/pull/4664 but more files are being added.
This is what I did
1.Forked the repository
2. Built the repository-This is when the files were introduced via npm install. 
3. Made a commit with message First build.
4. Created a branch to work on the issue , made a commit and pushed to origin.

Thats when the pull request was opened but it included the files from the first commit

wetneb:
When you make your commit, you should have an option to select which files to add to your commit. You should only add those which are relevant to your pull request. If you are using git from the command line, then the relevant step is when you do `git add`. You should only run this command on the files that you actually want to update in your pull request.

wetneb:
Another small detail: in your pull request description, you can write "fixes #3172" rather than "fixes #{3172}" so that GitHub recognizes it as an issue number.

WaltonG:
Thank you for the clarification. I have always been using git add . which happens to add all the files. I guess I'll have to submit another pull request

PR #4670, opened 2022-03-31
added margin prop to move cell editor 1cell size down

### LINKED ISSUE #1116: Move "Edit" cells window to lower position
Modal window, that opened by cells "Edit", should be lower, to be able to see full string you edit (just to size of one row lower of current position).

elroykanye:
Fixes #1116

- an additional margin rule for the `data-table-cell-editor` to move cell editors one cell down when editing the content

### DIFF
--- a/main/webapp/modules/core/styles/views/data-table-view.less
+++ b/main/webapp/modules/core/styles/views/data-table-view.less
@@ -344,6 +344,10 @@ a.data-table-flag-off {
   background-position: -17px -17px;
   }
 
+.data-table-cell-editor {
+  margin: 20px 0;
+}
+
 .data-table-cell-editor, .data-table-topic-popup {
   border: 1px solid @chrome_primary;
   background: @chrome_secondary;


elroykanye:
### Before
 - cell-edit before
 Case 1
 - cell-edit before 1
 Case 2

### After
- ce--edit after
Case 1
 - ce--edit after 1
 Case 2

wetneb:
It seems like a relatively safe move to me! In general cells can contain multiple lines, so it does not guarantee that we still see the entire original cell content while editing, but perhaps that is not a big deal. I would like to check what happens when editing a cell at the very bottom of the screen, I suspect the edit dialog could then be hidden.

elroykanye:
> It seems like a relatively safe move to me! In general cells can contain multiple lines, so it does not guarantee that we still see the entire original cell content while editing, but perhaps that is not a big deal. I would like to check what happens when editing a cell at the very bottom of the screen, I suspect the edit dialog could then be hidden.

Oh! That is true. I will look into this.

wetneb:
Hi @elroykanye, are you still working on this?

wetneb:
Ooops, I closed the PR accidentally while commenting: feel free to reopen if you are still working on it!

PR #4672, opened 2022-03-31
Enhance random number function

### LINKED ISSUE #3143: Enhance randomNumber() function

elroykanye:
Fixes #3143

- a test class to fit the expectations of this function
- a change to the ControlFunctionRegistry introducing the randint and change from randomNumber -> random
- addition of a case that handles 0-args to this function


main/src/com/google/refine/grel/ControlFunctionRegistry.java
```diff
@@ -306,7 +306,8 @@ static public void registerControl(String name, Control c) {
         registerFunction("combin", new Combin());
         registerFunction("degrees", new Degrees());
         registerFunction("radians", new Radians());
-        registerFunction("randomNumber", new RandomNumber());
```
wetneb, line 309: I would not remove this registration, as it would break expressions which currently use it. We should also try to keep the function backwards compatible, meaning that arguments which were previously accepted should still be, and should be interpreted in the same way.

main/src/com/google/refine/grel/ControlFunctionRegistry.java
```diff
@@ -306,7 +306,8 @@ static public void registerControl(String name, Control c) {
...
+        registerFunction("random", new RandomNumber());
+        registerFunction("randint", new RandomNumber());
```
wetneb, line 310: I do not think it is very useful to register the function as `randint`, as the `int` part specifies that it should return random *integers*, and that is not the case for this function.

wetneb (commented):
I would say it is quite important to keep existing GREL expressions working, so I would like to make sure we are backwards compatible here. Perhaps the simplest way would be to implement a completely new function and keep `randomNumber` untouched.

main/src/com/google/refine/grel/ControlFunctionRegistry.java
elroykanye: Oh! Okay, makes a lot of sense. Will do 😁

main/src/com/google/refine/expr/functions/math/RandomNumber.java
```diff
@@ -42,13 +42,27 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
...
+                if (args[0] instanceof Long && args[1] instanceof Long) {
+                    // return a long
+                    return ThreadLocalRandom.current().nextLong(((Number) args[0]).longValue(), ((Number) args[1]).longValue()+1);
+                } else {
+                    // return a double
+                    return ThreadLocalRandom.current().nextDouble(((Number) args[0]).doubleValue(), ((Number) args[1]).doubleValue());
+                }
+            }
         }
         return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects two numbers, the first must be less than the second");
```
wetneb, line 67: Maybe this error message could also be adapted now that calling the function without arguments is allowed.

wetneb (commented):
Sorry for the delay in reviewing this. I think it is nearly good to go: I would just adapt the user manual accordingly. You could change the name of the function there to `random` (forgetting about `randomNumber` in the docs but keeping it in the code for backwards-compatibility), and mention the different behaviours there. @tfmorris any other wishes?

main/src/com/google/refine/expr/functions/math/RandomNumber.java
elroykanye: Alright, noted!

main/src/com/google/refine/expr/functions/math/RandomNumber.java
```diff
@@ -64,7 +63,7 @@ public Object call(Properties bindings, Object[] args) {
                 }
             }
         }
-        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects two numbers, the first must be less than the second");
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + "expects no arguments or two numbers, the first must be less than the second");
```
wetneb, line 66: I think the space before "expects" is a good thing.

elroykanye: This is why I should go everywhere with my lenses 😅

docs/docs/manual/grelfunctions.md
```diff
@@ -489,7 +489,7 @@ Some of these math functions don't recognize integers when supplied as the first
 |`pow(n1, n2)`|Returns n1 raised to the power of n2. Note: value.pow(3)` will work, whereas `2.pow(3)` will not work.|`pow(2, 3)` returns 8 (2 cubed) and `pow(3, 2)` returns 9 (3 squared). The square root of any numeric value can be called with `value.pow(0.5)`.|
 |`quotient(n1, n2)`|Returns the integer portion of a division (truncated, not rounded), when supplied with a numerator and denominator.|`quotient(9,2)` returns 4.|
 |`radians(n)`|Converts an angle in degrees to radians.|`radians(10)` returns 0.17453292519943295.|
-|`randomNumber(n lowerBound, n upperBound)`|Returns a random integer in the interval between the lower and upper bounds (inclusively). Will output a different random number in each cell in a column.|
+|`randomNumber(n lowerBound, n upperBound)`|Returns a random integer in the interval between the lower and upper bounds (inclusively). Will output a different random number in each cell in a column. If no arguments are provided, returns a number in the range 0.0 <= x < 1.0|
```
wetneb, line 492: I would change `randomNumber` to `random` since that is probably the prefered name for this function, as suggested by the issue.

elroykanye: Oh! Yes, right

wetneb (approved):

PR #4673, opened 2022-03-31
fixes #3919 OpenRefine 3.4.1: Reconciliation Service does not use proxy defined in openrefine.l4j.ini

### LINKED ISSUE #3919: OpenRefine 3.4.1: Reconciliation Service does not use proxy defined in openrefine.l4j.ini
OpenRefine 3.4.1 on Windows (with or without integrated JRE) does not use the proxy set in openrefine.l4j.ini to connect to reconciliation service (https://lobid.org/gnd/reconcile)

### To Reproduce
1. First, set proxy in openrefine.l4j.ini using the following parameters:
  `-Dhttps.proxyHost=proxy-name.domain.ch`
  `-Dhttps.proxyPort=8080`
  `-Dhttp.proxyHost=proxy-name.domain.ch`
  `-Dhttp.proxyPort=8080`
  `-Dhttp.nonProxyHosts=localhost`
2. Then, start OpenRefine using openrefine.exe, open project, select column "CDU - GND" for reconciliation
3. Then start reconciling using https://lobid.org/gnd/reconcile
4. Finally, OpenRefine throws the following error 
`16:50:55.909 [                  command] Failed to guess cell types for load
{"q0":{"query":"Landquart (Fluss)","limit":3}} (0ms)
org.apache.http.conn.HttpHostConnectException: Connect to lobid.org:443 [lobid.org/193.30.112.187] failed: Connection timed out:...`

### Current Results
The following messages are shown in OpenRefine (with verbosity "all"): 
image
Additionally we did monitor the network traffic using wireshark and did find no connection from our localhost to our proxy but instead a connection request from our localhost to "emphytos.hbz-nrw.de"  which bears the alias "lobid.org" with the  IP 193.30.112.187.
This leads us to the conclusion that OpenRefine does not use the proxy settings specified in openrefine.l4j.ini.

### Expected Behavior
OpenRefine connects to the Reconciliation service using the proxy specified in openrefine.l4j.ini using the parameters  
  `-Dhttps.proxyHost=proxy-name.domain.ch`
  `-Dhttps.proxyPort=8080`
  `-Dhttp.proxyHost=proxy-name.domain.ch`
  `-Dhttp.proxyPort=8080`
  `-Dhttp.nonProxyHosts=localhost`

### Versions
 - Browser Version: Edge or Firefox
 - JRE or JDK Version: java version "1.8.0_281" Java(TM) SE Runtime Environment (build 1.8.0_281-b25) Java HotSpot(TM) 64-Bit Server VM (build 25.281-b25, mixed mode)
 - OpenRefine: Version 3.4.1 [437dc4d]

### Datasets
[Dummy.openrefine.tar.gz](https://github.com/OpenRefine/OpenRefine/files/6509899/Dummy.openrefine.tar.gz)

### Additional context
OpenRefine runs behind a proxy that does not allow direct connections to the internet. 
We tried both packages that are available for windows (windows with java and widows kit without java)
We also tested with the setting "-Djava.net.useSystemProxies=true", set the proxy-server in the java control panel and got the same results.
We also tried to run openrefine directly from the commandline and added the proxy parameters (we could not test refine.bat because the use of batch files is blocked by group policy).

WaltonG:
Fixes #3919

- Include httpproxy during HttpClient creation
- 
-


WaltonG:
@wetneb Am trying to get the values of  Dhttps.proxyHost, Dhttps.proxyPort, Dhttp.proxyHost=proxy, Dhttp.proxyPort, Dhttp.nonProxyHosts which has proven to be a challenge. My thought is if the values are available during HttpClient creation then the proxy defined will be used

WaltonG:
Am checking out the System.getProperty method or any other ways of getting the jvm runtime properties

main/src/com/google/refine/util/HttpClient.java
```diff
@@ -75,7 +80,113 @@ public HttpClient(int delay, int retryInterval) {
...
+                            if (delay > 0) {
+                                try {
+                                    Thread.sleep(delay);
+                                } catch (InterruptedException e) {
+                                }
+                            }
+                            nextRequestTime = System.currentTimeMillis() + _delay;
+
+                        }
+                    });
```
wetneb, line 153: It looks like you have duplicated a lot of code here! I think there should be a way to rewrite this to minimize the code duplication.

WaltonG: Am looking into it

WaltonG: @elroykanye Could there be an issue with my PR. 3 checks have been in progress for hours

wetneb:
Thank you! I think we could avoid duplication even more. Remember that calls to a builder do not need to be chained.

WaltonG:
@wetneb Thank you for the light-bulb moment on methods to a builder. I"ll get to work on it right away

WaltonG:
@wetneb Changes ready for review

main/src/com/google/refine/util/HttpClient.java
```diff
@@ -105,6 +111,24 @@ public void process(
...
+            proxyHost = System.getProperty("http.proxyHost");
+
+        if (System.getProperty("https.proxyHost") != null)
+            proxyHost = System.getProperty("https.proxyHost");
+
+        if (System.getProperty("http.proxyPort") != null)
+            proxyPort = Integer.parseInt(System.getProperty("http.proxyPort"));
+
+        if (proxyHost != null && proxyPort != 0) {
+            proxy = new HttpHost("http", proxyHost, proxyPort);
```
wetneb, line 124: Question (I have not looked through it myself): does this work for both HTTP and HTTPS proxies? Or do we need to add a configuration variable to let the user choose between the two?

main/src/com/google/refine/util/HttpClient.java
```diff
@@ -105,6 +111,24 @@ public void process(
                     }
                 });
 
+        if (System.getProperty("http.proxyHost") != null)
+            proxyHost = System.getProperty("http.proxyHost");
+
```
wetneb, line 116: I think we generally try to avoid if blocks without `{` and `}` because some consider it bad practice. This is something that should be caught by the linter, it looks like we need to configure it better. In the meantime, could you add the braces to those `if` blocks?

wetneb (commented):
Thanks a lot, it looks much better to me. Here are some small things.

wetneb (commented):
I had a closer look and it looks like HTTPS proxies are not actually supported by OkHTTP: https://github.com/square/okhttp/issues/3787
So I wonder how this is supposed to work at all? Have you actually tested this with an HTTPS proxy? Note: this is not about reaching an HTTPS website via an HTTP proxy, but also having the connection from the client to the proxy encrypted.

Also, it looks like the system properties used in this PR are not actually documented in our user manual, it would be good to add them in the same go.

Perhaps I would suggest removing the `https.proxyPort` variable altogether if it is not actually supported.

WaltonG:
Actually I had not considered this:
>Note: this is not about reaching an HTTPS website via an HTTP proxy, but also having the connection from the client to the proxy encrypted.

Working on this right away:
>Also, it looks like the system properties used in this PR are not actually documented in our user manual, it would be good to add them in the same go.
Perhaps I would suggest removing the https.proxyPort variable altogether if it is not actually supported.

docs/docs/manual/running.md
```diff
@@ -290,6 +290,15 @@ JAVA_OPTIONS=-Drefine.data_dir=usr/lib/OpenRefineWorkspace
...
 
 
+##### Proxy Configuration
+
+To use a proxy add the following configurations for either HTTP Proxy or HTTPS Proxy in the relevant [modification file](#modifications-set-within-files) : 
+
+-Dhttps.proxyHost= \
+-Dhttp.proxyHost= \
+-Dhttp.proxyPort=
+
```
wetneb, line 300: We have a section dedicated to Java variables, perhaps it would make sense to integrate these variables there?

antoine2711:
@WaltonG: you should rebase your PR because the running.md file was updated today, in fact, the table for the JVM preferences is changed.

https://github.com/OpenRefine/OpenRefine/blob/dba2b8b1311ce58e98f88d767726955ab098e38b/docs/docs/manual/running.md#L202-L211

Regards, Antoine

WaltonG:
@antoine2711 Oh that slipped me. Thanks

main/src/com/google/refine/util/HttpClient.java
```diff
@@ -131,6 +137,28 @@ public void process(
                     }
                 });
 
+        if (System.getProperty("http.proxyHost") != null) {
+            proxyHost = System.getProperty("http.proxyHost");
+        }
+
+        if (System.getProperty("https.proxyHost") != null) {
+            proxyHost = System.getProperty("https.proxyHost");
```
wetneb, line 145: Let us remove this section since HTTPS proxies are not supported.

docs/docs/manual/running.md
```diff
@@ -286,13 +293,16 @@ Locate the `refine.ini` file, and add `JAVA_OPTIONS=` before the `-Drefine.prefe
 JAVA_OPTIONS=-Drefine.autosave=2
 JAVA_OPTIONS=-Drefine.port=3334
 JAVA_OPTIONS=-Drefine.data_dir=usr/lib/OpenRefineWorkspace
+JAVA_OPTIONS=-Dhttps.proxyHost=yourproxyhost
+JAVA_OPTIONS=--Dhttp.proxyHost=yourproxyhost
```
wetneb, line 297: Double `-` here

main/src/com/google/refine/util/HttpClient.java
WaltonG: Sure

docs/docs/manual/running.md
WaltonG: Doing it right away

wetneb: The double `-` is still here.

WaltonG: Oh I overlooked that. Updating it in a few

PR #4676, opened 2022-04-01
Mention our main class or Java entry point

thadguidry:


wetneb:
I think it rather belongs to the `Running` section, instead of the  `Building` section, so I have made another proposal here: #4680.

thadguidry:
Ah ok. Then closing.

PR #4677, opened 2022-04-01
Also mention our main class in our current docs

thadguidry:


wetneb:
I propose to put it in `Running` instead: #4680

PR #4681, opened 2022-04-01
Add specific UI to edit different entity types in Wikibase manifest

### LINKED ISSUE #4270: Add specific UI to edit different entity types in Wikibase manifest
Following #4268, we will be able to use the current Wikibase schema interface to edit any entity type that has statements and/or terms: for instance, Properties or MediaInfo entities.

One problem with this integration is that it will propose irrelevant fields in the Wikibase schema. For instance, if I am editing MediaInfo entities, the OpenRefine user interface will offer adding labels, descriptions and aliases. This is confusing, because MediaInfo entities only have captions (which are internally the same thing as labels). So, as long as the user does not try to add description or aliases on MediaInfo entities, they will be able to edit, but this is still very brittle (for instance, it would be quite natural to assume that the "captions" that MediaInfo entities have correspond to "descriptions" on Wikidata, and therefore mistakenly select "description" when wanting to add a caption).

Therefore, we want to be able to adapt the user interface depending on the type of entity that we are editing.
For instance, when editing MediaInfo entities, the user should only be provided with the following data import options:
* captions
* statements
* wikitext

When editing items, the user should be provided with the following fields:
* labels, descriptions and aliases
* statements
* sitelinks (not supported yet)

### Proposed solution

Wikibase schemas currently consist of a list of item templates. Users can add more of these templates using the `+ add item` button in the schema window.
We could replace this button by a few others:
* `+ add item`
* `+ add mediainfo` (or other user-facing term, I suspect `MediaInfo` is just an internal term)
* `+ add property` (if we want to add support for properties…)

Those buttons could be shown only for the entity types actually supported by the Wikibase instance (for instance it would not make sense to offer editing MediaInfo entities on Wikidata, and conversely you should not be able to edit items on Commons).

Then, based on the choice made by the user, we would be able to show an entity template whose fields match the entity type (as hinted above).

### Alternatives considered

We could simply have an `+ add entity` button and then let the user select the entity type as a particular field in the entity template (which would then control which other fields are visible).

### Additional context

The UI choices for this task should be discussed with Lozana.

This task is part of #2144.

wetneb:
Fixes #4270.

This introduces different classes to distinguish between edits on Items and on MediaInfo entities. Before this, such edits were treated in the same way, meaning that we would offer the same fields to the user regardless of the entity type. For instance, the user still had the option of adding aliases to MediaInfo entities, which does not make sense. Now the UI can be adapted to the entity type, paving the way for the introduction of other fields specific to some entity types (such as file path, for MediaInfo entities, which will be needed to enable upload).

For now, visually the main difference is in the schema editor:
image


wetneb:
@j-sal @antoine2711 if you feel like reviewing this let me know, otherwise I would merge it soon.

PR #4684, opened 2022-04-01
Invalid URL detection in cell UI

### LINKED ISSUE #4546: Invalid URL detection in cell UI
Sometimes the cell UI gets URL boundaries wrong.

### To Reproduce
Create a project and put the following string in a cell:
```
{{Information
|Description    = {{en|1=<br>
:'''Species''' {{Species-inline|vernacular=Wood Duck|binominal=Aix sponsa}}
:'''Recordist''' Jonathon Jongsma
:'''Remarks''' Bird seen. Several widely-spaced and rather strange hoarse 'screams' from a female wood duck sitting about 15m directly above my head in a large cottonwood at the edge of an urban park pond. Recording quality is rather poor because all I had with me was a mobile phone.
:'''Bird seen''' yes
:'''Elevation''' 250 m
:'''Country''' United States}}
|Source         = [http://www.xeno-canto.org/101287
|Author ](http://www.xeno-canto.org/101287|Author)        = Jonathon Jongsma
|Date           = 2012-05-10 07:53
|Permission     = 
|other_versions = 
}}{{Location dec|44.942|-93.259}}
{{Xeno-canto}}

=={{int:license-header}}==
{{cc-by-sa-3.0}}

[[Category:Sound files uploaded by Fæ]]
[[Category:Xeno-canto]]
[[Category:Audio files of Aix sponsa]]
```

### Current Results

image

### Expected Behavior

The URL should not continue on the following line.

### Versions
 - Operating System: irrelevant
 - Browser Version: Firefox
 - JRE or JDK Version: irrelevant
 - OpenRefine: master branch

Binita-tech:
Fixes #4546 

-  Adapted the regex that detects the URL so that it does not accept newlines directly

### DIFF
--- a/docs/docs/manual/grelfunctions.md
+++ b/docs/docs/manual/grelfunctions.md
@@ -141,6 +141,12 @@ You cannot find or replace nulls with this, as null is not a string. You can ins
 
 Returns the string obtained by replacing a character in s, identified by find, with the corresponding character identified in replace. For example, `"Téxt thát was optícálly recógnízéd".replaceChars("áéíóú", "aeiou")` returns the string “Text that was optically recognized”. You cannot use this to replace a single character with more than one character.
 
+###### replaceEach(s, a find, a replace) {#replaceeach-s-a-find-a-replace}
+
+Returns the string obtained by replacing each element in s of a find array with the corresponding element of a replace array, sequentially. For example, `"The cow jumps over the moon and moos".replaceEachFromArrays(["th", "moo"], ["ex", "mee"])` returns the string “The cow jumps over exe meen and mees”.
+
+The length of the find array must be the same as the length of the replace array. If the length of the find array is greater than the length of the replace array, the last element of the replace array is used for all remaining elements in the find array.
+
 ###### find(s, sub or p) {#finds-sub-or-p}
 
 Outputs an array of all consecutive substrings inside string s that match the substring or [regex](expressions#grel-supported-regex) pattern p. For example, `"abeadsabmoloei".find(/[aeio]+/)` would result in the array [ "a", "ea", "a", "o", "oei" ].
--- a/main/src/com/google/refine/expr/functions/strings/ReplaceEach.java
+++ b/main/src/com/google/refine/expr/functions/strings/ReplaceEach.java
@@ -1,0 +1,64 @@
+package com.google.refine.expr.functions.strings;
+
+import com.google.refine.expr.EvalError;
+import com.google.refine.grel.ControlFunctionRegistry;
+import com.google.refine.grel.Function;
+import org.apache.commons.lang.StringUtils;
+
+import java.util.Arrays;
+import java.util.Properties;
+
+public class ReplaceEach implements Function {
+    /**
+     * Replace each occurrence of a substring in a string with another substring.
+     * @param bindings bindings
+     * @param args arguments
+     * @return result
+     */
+    @Override
+    public Object call(Properties bindings, Object[] args) {
+        if(args.length == 3) {
+            if(args[0] instanceof String && args[1] instanceof String[] && (args[2] instanceof String[] || args[2] instanceof String)) {
+                String str = (String) args[0];
+                String[] search = (String[]) args[1], replace;
+                if(args[2] instanceof String) {
+                    replace = new String[]{(String) args[2]};
+                }else {
+                    replace = (String[]) args[2];
+                }
+
+                // Check that the search array is greater than or equal to the replace array in length
+                if(search.length >= replace.length) {
+                   // check that the replace array and search array are the same length
+                    if(search.length != replace.length) {
+                        // make multiple replacements of the last element in replace
+                        String[] newReplace = new String[search.length];
+                        Arrays.fill(newReplace, replace[replace.length - 1]);
+                        replace = newReplace;
+                    }
+
+                    // replace each occurrence of search with corresponding element in replace
+                    return StringUtils.replaceEachRepeatedly(str, search, replace);
+                } else {
+                    return new EvalError(ControlFunctionRegistry.getFunctionName(this) + ": the number of replacements must be equal to or less than the number of search strings.");
+                }
+            }
+        }
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects 3 arguments: string, array of strings to replace, array of replacement strings");
+    }
+
+    @Override
+    public String getDescription() {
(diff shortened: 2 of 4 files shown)


main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -122,7 +122,7 @@ DataTableCellUI.prototype._render = function() {
       }
       if (spanArr.length != 0) {
         var span = document.createElement('span');
-        span.textContent = spanArr.join(" ");
+        span.innerHTML = spanArr.join(" ").replace(/\n/, '<br />');
```
wetneb, line 125: If cells contain HTML markup, then this would interpret the HTML markup directly by embedding the elements in OpenRefine, which is not what we want, I think :)

elroykanye, line 125: Hey, creating the br tags using the document object may help with this 💯

Binita-tech: Thanks @elroykanye. You are helping me to resolve this issue. Really appreciated :)

Binita-tech: Updated. Please check @wetneb

wetneb:
Thanks for your attempts! I think it is important to remember that cells can contain arbitrary strings, including HTML markup that should be just rendered as strings (not as HTML elements in OpenRefine's DOM), so if you use `innerHTML` then you need to make sure you are escaping things before.

elroykanye:
Hey @Binita-tech , noticed that some other files were changed in your commits which don't relate to the issue. That could cause some issues.
It would be better if the PR that fixes the issue tagged here only contained commits that modified the file(s) required 😁 
Consider isolating a new branch from the master and then modifying only the file you intend to. After that, you can push the changes you have made. If you need any help, please do let me know. :)

elroykanye:
> Hey @Binita-tech , noticed that some other files were changed in your commits which don't relate to the issue. That could cause some other issues. It would be better if the PR that fixes the issue tagged here only contained commits that modified the file(s) required grin Consider isolating a new branch from the master and then modifying only the file you intend to. After that, you can push the changes you have made. If you need any help, please do let me know. :)

Ah! Perfect! That fixes it. I was looking at the wrong change list.

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -99,25 +99,35 @@ DataTableCellUI.prototype._render = function() {
...
           if (spanArr.length != 0) {
             var span = document.createElement('span');
             span.textContent = spanArr.join(" ");
             divContent.appendChild(span).appendChild(document.createTextNode('\u00A0'));
             spanArr = [];
           }
+          var [realURL, extra] = ((str) => {
+            var parts = str.split('\n');
+            return parts.length > 1 ? [parts[0], parts.slice(1).map((s) => ("\n" + s))] : [parts[0], []];
+          })(arr[i]);
```
elroykanye, line 113: The IIFE used here does not really require any parameters as you can just use the arr[i] variable directly in it. 
I would imagine that it should look more like this

```js
var [realURL, extra] = (() => {
     var parts = arr[i]split('\n');
     return parts.length > 1 ? [parts[0], parts.slice(1).map((s) => ("\n" + s))] : [parts[0], []];
})();
```

Apart from that, looks everything looks fine to me. 👍🏾

elroykanye (changes requested):
Thanks for this @Binita-tech

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
Binita-tech: what about str, should I remove str also? @elroykanye

elroykanye: Yes

elroykanye:
Thanks!

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -99,25 +99,35 @@ DataTableCellUI.prototype._render = function() {
...
+      for (var i=0; i<arr.length; i++) {
         if (URL.looksLikeUrl(arr[i])) {
           if (spanArr.length != 0) {
             var span = document.createElement('span');
             span.textContent = spanArr.join(" ");
             divContent.appendChild(span).appendChild(document.createTextNode('\u00A0'));
             spanArr = [];
           }
+          var [realURL, extra] = (() => {
+            var parts = str.split('\n');
```
elroykanye, line 111: You may want to adjust this from your previous update :)
```suggested
var parts = arr[i].split('\n');
```

Binita-tech: Done

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
Binita-tech: Done

elroykanye (approved):

wetneb (commented):
Thank you both!
I tried it out and it works well. However, at the preview stage (before project creation, when you can tweak import settings), the bug is still there. Could you investigate what it would take to also fix the bug there?

elroykanye:
> Thank you both! I tried it out and it works well. However, at the preview stage (before project creation, when you can tweak import settings), the bug is still there. Could you investigate what it would take to also fix the bug there?

Hi @wetneb , I have gone through this for some time and tried to see how we could improve right at the models (Column -> Cell -> Value). However, I have been unable to find a way out because I noticed that data is mostly written with Cell value types serialised through the `ParsingUtilities.mapper`.

The approach @Binita-tech used came from us realising that this is how urls with proceeding words are stored in the `data.txt`. The image below is an example.
Screenshot_20220421_161613

I wish to ask if you have any suggestion for how we can solve this from the project creation? Maybe we both have been going round in circles where the solution isn't, so we might some pointers in the right direction. :)

wetneb:
I think this should be fairly simple: it is just about locating the javascript code that is in charge of rendering values at the preview stage. This rendering is slightly different than in the final project, so that is why your changes do not fix the bug at the preview stage. You should not need to worry about Java things or project serialization at all for this.

elroykanye:
OH! I get it now. On it then 😁

wetneb:
Hi @Binita-tech, are you still working on this?

Binita-tech:
I am not working on this issue anymore @wetneb

PR #4687, opened 2022-04-02
Fixes #2316 "Match other cells with same content" keeps appearing as the default choice #2316

### LINKED ISSUE #2316: "Match other cells with same content" keeps appearing as the default choice

WaltonG:
Fixes #2316

- Use a global variable to store user's choice of radio button
- Checked radio button is set by the user's previous choice
-

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-ui.js
@@ -41,7 +41,7 @@ function DataTableCellUI(dataTableView, cell, rowIndex, cellIndex, td) {
   this._render();
 }
 
-var defaultChoice = true;
+var reconMatchSilimilarCellsByDefault = true;
 
 DataTableCellUI.previewMatchedCells = true;
 
@@ -363,7 +363,7 @@ DataTableCellUI.prototype._searchForMatch = function(suggestOptions) {
   elmts.clearButton.html($.i18n('core-buttons/dont-reconcile'));
   elmts.cancelButton.html($.i18n('core-buttons/cancel'));
   
-	if (!defaultChoice) {
+	if (!reconMatchSilimilarCellsByDefault) {
 		elmts.radioSimilar[0].setAttribute("checked", false);
 		elmts.radioOne[0].setAttribute("checked", true);
 	}
@@ -371,14 +371,9 @@ DataTableCellUI.prototype._searchForMatch = function(suggestOptions) {
   var level = DialogSystem.showDialog(frame);
   var dismiss = function() {
 	  if (elmts.radioSimilar[0].checked) {
-		  defaultChoice = true;
-		  elmts.radioSimilar[0].setAttribute("checked", true);
-		  elmts.radioOne[0].setAttribute("checked", false);
-
+		  reconMatchSilimilarCellsByDefault = true;
 	  } else {
-		  defaultChoice = false;
-		  elmts.radioSimilar[0].setAttribute("checked", false);
-		  elmts.radioOne[0].setAttribute("checked", true);
+		  reconMatchSilimilarCellsByDefault = false;
 	  }
 	
     DialogSystem.dismissUntil(level - 1);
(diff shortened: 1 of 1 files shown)


main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -41,6 +41,8 @@ function DataTableCellUI(dataTableView, cell, rowIndex, cellIndex, td) {
   this._render();
 }
 
+var defaultChoice = true;
```
wetneb, line 44: Since you are introducing this as a global variable, I think it would be better to use a more specific name, making it clear which choice it affects. For instance `reconMatchSilimilarCellsByDefault` or something like that.

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -360,9 +362,25 @@ DataTableCellUI.prototype._searchForMatch = function(suggestOptions) {
...
+		elmts.radioSimilar[0].setAttribute("checked", false);
+		elmts.radioOne[0].setAttribute("checked", true);
+	}
 
   var level = DialogSystem.showDialog(frame);
   var dismiss = function() {
+	  if (elmts.radioSimilar[0].checked) {
+		  defaultChoice = true;
+		  elmts.radioSimilar[0].setAttribute("checked", true);
+		  elmts.radioOne[0].setAttribute("checked", false);
```
wetneb, line 376: It is not clear to me why the two lines above are needed - why do you need to update the DOM at this stage? (similarly below)

wetneb (commented):
Thank you, I think this is on the right track!

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
WaltonG: Well noted

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
WaltonG: Actually I think its not needed since the users choice has been stored in the global variable and the DOM should be updated using the global variable value when the reconciliation match dialog is opened. 
Let me work on it an see the outcome

elroykanye:
I don't know what happened to the workflows  @WaltonG but I guess it is all good now. I had to cancel and rerun.

WaltonG:
> I don't know what happened to the workflows @WaltonG but I guess it is all good now. I had to cancel and rerun.
This PR is fine. The issue was on PR #4673  The check have failed with null poniter exception. Can you check it out

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -360,9 +362,20 @@ DataTableCellUI.prototype._searchForMatch = function(suggestOptions) {
...
   elmts.cancelButton.html($.i18n('core-buttons/cancel'));
+  
+	if (!reconMatchSilimilarCellsByDefault) {
+		elmts.radioSimilar[0].setAttribute("checked", false);
+		elmts.radioOne[0].setAttribute("checked", true);
+	}
 
   var level = DialogSystem.showDialog(frame);
   var dismiss = function() {
+	  if (elmts.radioSimilar[0].checked) {
```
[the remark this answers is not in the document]
WaltonG: I think its logical because if radioSimilar is checked it automatically implies the radioOne is uchecked and vice versa.
Thanks

antoine2711 (approved):
@WaltonG 👍 Good work. This is complete for me. @wetneb will merge soon.

Regards, Antoine

WaltonG:
@antoine2711 👍

wetneb (approved):
Thank you!

PR #4689, opened 2022-04-03
set division operator call to return NaN on 0/0

### LINKED ISSUE #377: Not seeing a NaN when expected on initial import
_Original author: PaulMake...@gmail.com (May 10, 2011 17:48:18)_

1. Import the attached file (containing simply the number 0 and a header)
2. Edit cells -&gt; Transform: value / value

Expect to see NaN - see instead null.

<b>What version of Google Refine are you using?</b>

2.0

<b>What operating system and browser are you using?</b>

Windows Chrome

If I transform the column to text and then back to numbers I get the expected NaN as though the initial import missed something in construction.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=377_

elroykanye:
Fixes #377

- fix the division operator call to return a Double.NaN when `n1 <- 0, n2 <- 0`

### DIFF
--- a/main/src/com/google/refine/grel/ast/OperatorCallExpr.java
+++ b/main/src/com/google/refine/grel/ast/OperatorCallExpr.java
@@ -74,7 +74,9 @@ public class OperatorCallExpr implements Evaluable {
                     } else if ("*".equals(_op)) {
                         return n1 * n2;
                     } else if ("/".equals(_op)) {
-                        if(n2 == 0 && n1 == 0) { return Double.NaN; }
+                        if (n2 == 0 && n1 == 0) {
+                            return Double.NaN;
+                        }
                         return n1 / n2;
                     } else if ("%".equals(_op)) {
                         return n1 % n2;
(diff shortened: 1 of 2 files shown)


main/src/com/google/refine/grel/ast/OperatorCallExpr.java
```diff
@@ -101,6 +102,7 @@ public Object evaluate(Properties bindings) {
                     } else if ("*".equals(_op)) {
                         return n1 * n2;
                     } else if ("/".equals(_op)) {
+                        if(n2 == 0&& n1 == 0) { return Double.NaN; }
```
wetneb, line 105:
```suggestion
                        if (n2 == 0 && n1 == 0) {
                            return Double.NaN;
                        }
```

wetneb (commented):
Thank you! Could you add a unit test as well?

main/src/com/google/refine/grel/ast/OperatorCallExpr.java
wetneb: We normally have linting checks which should pick this up but it looks like they are not quite strict enough. You could do a similar change above.

elroykanye: Alright, I will. Thanks! :)

wetneb (approved):
Thanks @elroykanye!

elroykanye:
> Thanks @elroykanye!

My pleasure 😁

PR #4697, opened 2022-04-05
New grel function parseURI() and access to key aspects (host, port, query…)

### LINKED ISSUE #1857: Implement GREL to parse URIs and extract key aspects
Given a URI or URL it can be very useful to break down the URI into its constituent parts. Also to check the URI for validity.

The gokbutils extension implements an 'extractHost' function, but this is only a single part of the set of things you might want to extract from a valid URI. Other aspects include:

* scheme or protocol
* host
* port
* path
* fragment (after a #)
* query (after a ?)

A GREL command or set of GREL commands to make it easy to extract these from a string that looks like a URI. For example:

if value  = "https://www.openrefine.org:80/documentation#download":

value.parseURI() -> a set of elements which can be accessed via dot notation or 'get()' function. If 'value' is not a valid URI return an error
value.parseURI().host -> www.openrefine.org
and/or
value.parseURI().get("host") -> www.openrefine.org
and/or
value.parseURI().authority -> www.openrefine.org

and similarly (using either dot or 'get' syntax)

value.parseURI().path -> /documentation
value.parseURI().port -> 80
value.parseURI().scheme -> https
value.parseURI().fragment -> dowload

etc.
Some or all of this can be achieved using match or split with the appropriate regular expressions, but the parseURI would allow for the validity of the string as a URI to be checked, and makes it more accessible.

**Additional context**
What I've basically described here is implementing some of the methods of https://docs.oracle.com/javase/7/docs/api/java/net/URI.html 

e.g. URI.create(String) would be used to create a URI object from the string
URI.getAuthority would be used to get the host
etc.

elroykanye:
Fixes #1857 

- a grel function to parse uri and extract properties into a map
- an addition of this function to the function registry
- a test for this function

### DIFF
--- a/main/src/com/google/refine/expr/functions/strings/ParseUri.java
+++ b/main/src/com/google/refine/expr/functions/strings/ParseUri.java
@@ -1,6 +1,6 @@
 package com.google.refine.expr.functions.strings;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.refine.expr.EvalError;
 import com.google.refine.grel.Function;
 import com.google.refine.util.ParsingUtilities;
@@ -10,8 +10,6 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.Properties;
 
 /**
@@ -41,30 +39,28 @@ public class ParseUri implements Function {
                     }
                 }
 
-                // initial
-                Map<String, String> queryParams = new HashMap<>(Map.of());
+                // split the query string into key-value pairs
+                ObjectNode queryParamsNode = ParsingUtilities.mapper.createObjectNode();
                 if(qp.length() != 0) {
                     // get the query parameters as a list of name-value pairs
                     Arrays.stream(qp.split("&"))
-                            .forEach(pair -> queryParams.put(pair.split("=")[0], pair.split("=")[1]));
+                            .forEach(pair -> queryParamsNode.put(pair.split("=")[0], pair.split("=")[1]));
                 }
 
-                return ParsingUtilities.mapper.readTree(ParsingUtilities.mapper.writeValueAsString(Map.of(
-                        "scheme", uri.getScheme(),
-                        "host", uri.getHost(),
-                        "port", String.valueOf(uri.getPort() == -1 ? 80 : uri.getPort()),
-                        "path", uri.getPath(),
-                        "query", uri.getQuery(),
-                        "authority", uri.getAuthority(),
-                        "fragment", fragment,
-                        "query_params", ParsingUtilities.mapper.writeValueAsString(queryParams)
-                )));
+                ObjectNode uriNode = ParsingUtilities.mapper.createObjectNode();
+                uriNode.put("scheme", uri.getScheme());
+                uriNode.put("host", uri.getHost());
+                uriNode.put("port", String.valueOf(uri.getPort() == -1 ? 80 : uri.getPort()));
+                uriNode.put("path", uri.getPath());
+                uriNode.put("query", uri.getQuery());
+                uriNode.put("authority", uri.getAuthority());
+                uriNode.put("fragment", fragment);
+                uriNode.put("query_params", queryParamsNode);
+                return uriNode;
 
 
             } catch (URISyntaxException | MalformedURLException e) {
                 return new EvalError("Invalid URI: " + s);
-            } catch (JsonProcessingException e) {
-                e.printStackTrace();
             }
         }
         return new EvalError("ParseUri takes a single string argument.");
--- a/main/src/com/google/refine/expr/functions/strings/ParseUri.java
+++ b/main/src/com/google/refine/expr/functions/strings/ParseUri.java
@@ -42,17 +42,17 @@ public class ParseUri implements Function {
                 e.printStackTrace();
             }
         }
-        return new EvalError("ParseUri takes a single string argument");
+        return new EvalError("ParseUri takes a single string argument.");
     }
 
     @Override
     public String getDescription() {
-        return "Parses a URI and extracts its components";
+        return "Parses a URI and extracts its components.";
     }
 
     @Override
     public String getParams() {
-        return "String url";
+        return "string url";
     }
 
     @Override
--- a/main/src/com/google/refine/expr/functions/strings/ParseUri.java
+++ b/main/src/com/google/refine/expr/functions/strings/ParseUri.java
@@ -50,7 +50,7 @@ public class ParseUri implements Function {
                 ObjectNode uriNode = ParsingUtilities.mapper.createObjectNode();
                 uriNode.put("scheme", uri.getScheme());
                 uriNode.put("host", uri.getHost());
-                uriNode.put("port", String.valueOf(uri.getPort() == -1 ? 80 : uri.getPort()));
+                uriNode.put("port", uri.getPort());
                 uriNode.put("path", uri.getPath());
                 uriNode.put("query", uri.getQuery());
                 uriNode.put("authority", uri.getAuthority());
--- a/main/src/com/google/refine/expr/functions/strings/ParseUri.java
+++ b/main/src/com/google/refine/expr/functions/strings/ParseUri.java
@@ -1,0 +1,47 @@
+package com.google.refine.expr.functions.strings;
+
+import com.google.refine.expr.EvalError;
+import com.google.refine.grel.Function;
+
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.Map;
+import java.util.Properties;
+
+public class ParseUri implements Function {
+    @Override
+    public Object call(Properties bindings, Object[] args) {
+        if(args.length == 1 && args[0] instanceof String) {
+            String s = (String)args[0];
+            try {
+                URL url = new URL(s);
+                URI uri = url.toURI();
+
+                return Map.of(
+                        "scheme", uri.getScheme() == null ? "" : uri.getScheme(),
+                        "host", uri.getHost() == null ? "" : uri.getHost(),
+                        "port", String.valueOf(uri.getPort()),
+                        "path", uri.getPath() == null ? "" : uri.getPath(),
+                        "query", uri.getQuery() == null ? "" : uri.getQuery(),
+                        "fragment", uri.getFragment() == null ? "" : uri.getFragment(),
+                        "authority", uri.getAuthority() == null ? "" : uri.getAuthority()
+                );
+            } catch (URISyntaxException | MalformedURLException e) {
+                return new EvalError("Invalid URI: " + s);
+            }
+        }
+        return new EvalError("ParseUri takes a single string argument");
+    }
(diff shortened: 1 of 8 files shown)


antoine2711:
@elroykanye: in the help pane, the documentation shows `undefined` as the parameter type. Shouldn't it string?
image

Regards, Antoine

antoine2711:
Also, this works fine:
image

But, shouldn't this work also?
image

Regards, Antoine

ostephens:
Returning a map isn't something we do anywhere else in GREL. I'm slightly concerned about introducing a new data format for a single GREL function and also wondering what other behaviour we can expect (e.g. how does this interact with other functions like `type()` or `length()` etc. Can a map be stored in a cell? Does `toString()` work?

We would also need to update the documentation with this new type.

I'm not sure I have the right answer here, and it maybe that a map is a good idea, but it could be simpler (from a behaviour perspective) to return a JSON representation of the object (as is returned by the existing `parseJson()` function) , and then we have a set of better known behaviour?

elroykanye:
> Also, this works fine: image
> 
> But, shouldn't this work also? image
> 
> Regards, Antoine

Thanks for this @antoine2711 , I missed out the getParams()
Will work on that right now.

elroykanye:
> I'm not sure I have the right answer here, and it maybe that a map is a good idea, but it could be simpler (from a behaviour perspective) to return a JSON representation of the object (as is returned by the existing `parseJson()` function) , and then we have a set of better known behaviour?

To be honest, I was sceptical about returning a map. Actually hoped to get a resolution after opening this PR. Thanks for this @ostephens . I am looking through other methods of returning this following the current paradigm.

antoine2711 (commented):
image

One detail: I think that port 80 should be returned even when not explicit.

main/src/com/google/refine/expr/functions/strings/ParseUri.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Parses a URI and extracts its components";
+    }
+
+    @Override
+    public String getParams() {
+        return "String url";
```
antoine2711, line 55:
```suggestion
        return "string url";
```

main/src/com/google/refine/expr/functions/strings/ParseUri.java
```diff
@@ -0,0 +1,62 @@
...
+            } catch (JsonProcessingException e) {
+                e.printStackTrace();
+            }
+        }
+        return new EvalError("ParseUri takes a single string argument");
+    }
+
+    @Override
+    public String getDescription() {
+        return "Parses a URI and extracts its components";
```
antoine2711, line 50:
```suggestion
        return "Parses a URI and extracts its components.";
```

main/src/com/google/refine/expr/functions/strings/ParseUri.java
```diff
@@ -0,0 +1,62 @@
...
+                        "fragment", uri.getFragment() == null ? "" : uri.getFragment(),
+                        "authority", uri.getAuthority() == null ? "" : uri.getAuthority()
+                )));
+            } catch (URISyntaxException | MalformedURLException e) {
+                return new EvalError("Invalid URI: " + s);
+            } catch (JsonProcessingException e) {
+                e.printStackTrace();
+            }
+        }
+        return new EvalError("ParseUri takes a single string argument");
```
antoine2711, line 45:
```suggestion
        return new EvalError("ParseUri takes a single string argument.");
```

main/tests/server/src/com/google/refine/expr/functions/strings/ParseUriTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class ParseUriTest extends RefineTest {
+
+    private String sampleUri;
+
+    @Override
+    @BeforeTest
+    public void init() {
+        logger = LoggerFactory.getLogger(this.getClass());
+        sampleUri = "https://www.openrefine.org:80/documentation#download";
```
antoine2711, line 23:
```suggestion
        sampleUri = "https://www.openrefine.org:80/documentation#download?format=xml&os=mac";
```
Could our example and test be more complex? More exhaustive?

antoine2711 (commented):
Just some comments.

Regards, Antoine

elroykanye:
Thanks very much @antoine2711 
I will work on this 😁

main/tests/server/src/com/google/refine/expr/functions/strings/ParseUriTest.java
```diff
@@ -20,22 +20,23 @@
...
+
         Assert.assertNotNull(res);
         Assert.assertEquals(resMap.get("path"), "/documentation");
         Assert.assertEquals(resMap.get("host"), "www.openrefine.org");
         Assert.assertEquals(resMap.get("port"), "80");
         Assert.assertEquals(resMap.get("query"), "");
         Assert.assertEquals(resMap.get("fragment"), "download");
         Assert.assertEquals(resMap.get("scheme"), "https");
-
+        Assert.assertNotNull(resMap.get("query_params"));
```
antoine2711, line 39: Could you validate the data value (`format=xml&os=mac`) and not just that there is data?

elroykanye: oh yes I can do that!

antoine2711: @elroykanye : just a comment, don't change the code, but here, you could have just compare the value of the whole string, since we don't manipulate it's composition.

Also, but I'm not asking it, we could return the data in such a way that we could do:
`parseUri("https://www.openrefine.org:80/documentation#download?format=xml&os=mac").search[0].key`
or
`parseUri("https://www.openrefine.org:80/documentation#download?format=xml&os=mac").search['format']`

Then, the more precise test case would be necessary! ;-)

Regards, Antoine

P.S. I'm not asking to do that, merely suggesting it could be done.

main/src/com/google/refine/expr/functions/strings/ParseUri.java
```diff
@@ -0,0 +1,89 @@
...
+public class ParseUri implements Function {
+
+    @Override
+    public Object call(Properties bindings, Object[] args) {
+        if(args.length == 1 && args[0] instanceof String) {
+            String s = (String)args[0];
+            try {
+                URL url = new URL(s);
+                URI uri = url.toURI();
+                ObjectMapper objectMapper = new ObjectMapper();
```
wetneb, line 31: I think Jackson's docs recommend not to instantiate an `ObjectMapper` too often and instead rely on a statically stored one. We have one you can use at `ParsingUtilities.mapper` (you are actually already using that below, so you might just be able to delete this instance :) )

main/src/com/google/refine/expr/functions/strings/ParseUri.java
```diff
@@ -0,0 +1,89 @@
...
+
+                // initial
+                Map<String, String> queryParams = new HashMap<>(Map.of());
+                if(qp.length() != 0) {
+                    // get the query parameters as a list of name-value pairs
+                    Arrays.stream(qp.split("&"))
+                            .forEach(pair -> queryParams.put(pair.split("=")[0], pair.split("=")[1]));
+                }
+
+                return ParsingUtilities.mapper.readTree(objectMapper.writeValueAsString(Map.of(
```
wetneb, line 54: Perhaps we can avoid the slightly inelegant serialization round-trip here, with something of this form:
```java
ObjectNode node = ParsingUtilities.mapper.createObjectNode();
node.put("scheme", uri.getScheme());
// ...
```

main/src/com/google/refine/expr/functions/strings/ParseUri.java
```diff
@@ -0,0 +1,89 @@
...
+                        "scheme", uri.getScheme() == null ? "" : uri.getScheme(),
```
wetneb, line 55: I am not sure why we should convert `null` to `""` here: JSON can store `null` values too, which do exist in GREL, right?

main/src/com/google/refine/expr/functions/strings/ParseUri.java
elroykanye: Oh! Yes, thanks 😁

main/src/com/google/refine/expr/functions/strings/ParseUri.java
elroykanye: Will do! Thanks

main/src/com/google/refine/expr/functions/strings/ParseUri.java
elroykanye: We actually do. 
I faced some issues with an earlier implementation that returned a Map instead of a JSONObject, so I had to replace `null` with `""`. This was just carried on from there. 

I will update accordingly, thanks ☺️

main/src/com/google/refine/expr/functions/strings/ParseUri.java
```diff
@@ -43,30 +39,28 @@ public Object call(Properties bindings, Object[] args) {
...
-                        "path", uri.getPath() == null ? "" : uri.getPath(),
-                        "query", uri.getQuery() == null ? "" : uri.getQuery(),
-                        "authority", uri.getAuthority() == null ? "" : uri.getAuthority(),
-                        "fragment", fragment,
-                        "query_params", objectMapper.writeValueAsString(queryParams)
-                )));
+                ObjectNode uriNode = ParsingUtilities.mapper.createObjectNode();
+                uriNode.put("scheme", uri.getScheme());
+                uriNode.put("host", uri.getHost());
+                uriNode.put("port", String.valueOf(uri.getPort() == -1 ? 80 : uri.getPort()));
```
wetneb, line 53: I guess this fallback could introduce a wrong port: for instance, if parsing `https://foo.com/` then you will return 80 as port, which is incorrect since the implied port for HTTPS connections (443). For simplicity you could just forward `uri.getPort()` without trying to add more logic to it, I would say.

elroykanye: I see 🤔 alright then

antoine2711: > I guess this fallback could introduce a wrong port: for instance, if parsing `https://foo.com/` then you will return 80 as port, which is incorrect since the implied port for HTTPS connections (443). For simplicity you could just forward `uri.getPort()` without trying to add more logic to it, I would say.

@wetneb: good catch! I should've seen it… ;-) 

I wonder here if it wouldn’t be better to return 443 for https and 80 for http. Because if this information can be eluded at the full URL level, it always implies those values. But at the `.port` level, we have no context. If you want to know the port, "" says nothing, since it could be 80 or 443, like you toughtfully mentioned.

Anyway, I understand why "" can also be what the user would want to see… I guess that since we are working the data, and not interpreting it, "" might be closer to the data.

Regards, Antoine

wetneb:
This looks good to me, but we miss an update of the user manual, which describes the new GREL function there.

elroykanye:
Noted! I'll be sending that soon.

wetneb (approved):

antoine2711:
> Noted! I'll be sending that soon.

Good work @elroykanye! One less issue... ;-)

Regards, Antoine

elroykanye:
Thanks for the help @antoine2711 @wetneb

PR #4703, opened 2022-04-06
Added fix for dot notation

### LINKED ISSUE #3328: Math functions behave inconsistently with dot notation
Separate but perhaps related to #3062. 

Some of the math functions ([see here for a handy reference in draft](https://deploy-preview-3264--openrefine-docs.netlify.app/manual/expressions#math-functions)) behave strangely when using dot notation for supplying arguments.

Three I've found:
1. `value.mod(9)` will work, whereas `74.mod(value)` will not work - it just returns 74.
2. `value.pow(3)` will work, whereas `2.pow(value)` will not work - it just returns 2.
3. `5.max(value)` just returns 5.

The syntax to fix this is just to throw the first argument in brackets - `(74).mod(value)` will work. But this is inconsistent across the other math functions (`-3.7.ceil()` works fine, e.g.). 
`cos(5)` works fine, but `5.cos()` doesn't. Same with 5.cosh(), 5.degrees(), 5.exp()....

Interestingly, `-3.7.abs()` works fine (returns 3.7), but `-3.abs()` doesn't (returns -3). 

### Expected Behavior
I'm not sure if this is a bug or just some complicated behaviour based on each function and its purpose. If someone can explicate it for me, I'll write it down! 

If not, I think users should be able to supply a number argument consistently across all math functions either with or without brackets. 

I haven't tested everything to see whether it treats floats and integers differently, but that abs() behaviour worries me too. I am guessing that is a separate bug, but I could be wrong.

elebitzero:
Fixes #3328

- Added fix for dot notation for math functions:

image
image
image
image

### DIFF
--- a/main/src/com/google/refine/grel/Scanner.java
+++ b/main/src/com/google/refine/grel/Scanner.java
@@ -134,7 +134,8 @@ public class Scanner {
                 _index++;
             }
 
-            if (_index < _limit && (c == '.' || c == 'e')) {
+            if (_index + 1 < _limit && (c == '.')
+                    && Character.isDigit(_text.charAt(_index + 1))) {
                 double value2 = value;
                 if (c == '.') {
                     _index++;


wetneb:
Could we add a few tests to demonstrate the behaviour? There are a few cases mentioned on the issue.
Overall I am a bit worried that we might be breaking parsing in other contexts, precisely because our test suite is not great (but I could have a closer look at the code to convince myself otherwise).

elebitzero:
After looking at this more, I understand the potential for breaking existing behavior.  Since I'm not that familiar with the intricacies of GREL, I am going to close the PR for now.

PR #4704, opened 2022-04-06
[I #387] No error message when you have 5 or less rows in the XML

### LINKED ISSUE #387: No error message when you have 5 or less rows of XML
_Original author: jinyoung...@gmail.com (May 22, 2011 20:06:00)_

1. Have an XML file with 5 or fewer rows.
2. Open it up in Google Refine.
3. No rows are shown, no error is mentioned.

I expect either an error saying that there needs to be an minimum of 6 rows or to see the rows even if there are only 5 rows.

<b>What version of Google Refine are you using?</b>

Version 2.0 for Mac.

<b>What operating system and browser are you using?</b>

Mac OS X 10.6.7, Firefox 3.6.13

<b>Is this problem specific to the type of browser you're using or it happens</b>
<b>in all the browsers you tried?</b>

Didn't try other browsers.

We thought for a while that our app was outputting data in a odd way because it seemed to not import into Google Refine.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=387_

WaltonG:
Fixes #387

- Show all child rows in the uploaded xml file


main/src/com/google/refine/importers/tree/XmlImportUtilities.java
```diff
@@ -237,9 +234,9 @@ static protected RecordElementCandidate detectRecordElement(TreeReader parser, S
             sortRecordElementCandidates(descendantCandidates);
 
             RecordElementCandidate candidate = descendantCandidates.get(0);
-            if (candidate.count / descendantCandidates.size() > 5) {
-                return candidate;
-            }
+
```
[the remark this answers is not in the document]
antoine2711: @WaltonG : importing less than 5 rows is still a good way to proceed.

The question that @wetneb raised was: are you sure this safe-guard was not implemented for a particular reason? There might be scenarii where such a check prevents other problems…

Regards, Antoine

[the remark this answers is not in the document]
WaltonG: @antoine2711 I have just seen your response after after submitting another PR. 

>I think the challenge of this issue is also to find that. In any case, if there is 4 rows in a XML, we should import it. I can really think a of scenario where I would want to do that, and now, if I wanted to do that with OR, I would have to play with the data to make OR import it. Not ideal.

I get your point, the code should not limit a user in such a scenario. I will have to dig into the code to gain more insight into its logic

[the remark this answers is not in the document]
antoine2711: > Is it OK to go with the option of introducing an error message ?

@WaltonG: I personaly strongly feel that we should import all we can, even a 1x2 or 2x1 (more than 1 cell).

That being said, the error handling you are adding is also important and necessary. But is it the only error that can occur?

Regards,
   Antoine

wetneb:
@WaltonG I think it is really not worth attempting to tackle this issue without first understanding why this check is currently here. I do not think we want to expose this error message to the user: they should indeed be able to import a project with fewer than 5 records.

main/src/com/google/refine/importers/tree/XmlImportUtilities.java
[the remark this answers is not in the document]
WaltonG: >That being said, the error handling you are adding is also important and necessary. But is it the only error that can occur?
@antoine2711 Not really. There is a possibilty that other unchecked errors occuring.

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
```diff
@@ -265,10 +265,13 @@ Refine.DefaultImportingController.prototype.getPreviewData = function(callback,
           "limit" : numRows || 100 // More than we parse for preview anyway
         }),
         null,
-        function(data) {
-          result.rowModel = data;
-          callback(result);
-        },
+		  function(data) {
```
antoine2711, line 268: @WaltonG: your spacing seams off a bit. What do you think?

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
```diff
@@ -265,10 +265,13 @@ Refine.DefaultImportingController.prototype.getPreviewData = function(callback,
...
+			  if (data["rows"].length == 0) {
```
antoine2711, line 269: How about checking for a few rows, but empty columns? Is that a possible scenario?

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
```diff
@@ -265,10 +265,13 @@ Refine.DefaultImportingController.prototype.getPreviewData = function(callback,
...
+				alert($.i18n('core-index-import/load-xml-rows-error'));
+			  }
+			  result.rowModel = data;
+			  callback(result);
+		  },
         "jsonp"
```
antoine2711, line 275: @WaltonG: I understand that it was not there before, but how about putting some error handling here? It seams to me that if the call fails, the user is going to be stuck...

antoine2711 (changes requested):
Overall, a good move forward.
With this start data:
```
{
"a":["A1", "A2"],
"b":["B1", "B2"]
}
```
I get this:
image

Imagine just 4 cells, but with thousands of character! This is good. Still, I noted a few points.

Regards, Antoine

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
WaltonG: True its too indented. Will sort it out.

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
WaltonG: I will have to try and see if its possible

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
WaltonG: @antoine2711 Actuallythe callback was there before.  I only added
https://github.com/OpenRefine/OpenRefine/blob/50e84ebb43bee103113540a47191151856e319f3/main/webapp/modules/core/scripts/index/default-importing-controller/controller.js#L269-L271

WaltonG: > @WaltonG: I understand that it was not there before, but how about putting some error handling here? It seams to me that if the call fails, the user is going to be stuck...

@antoine2711: I cant seem to get the errors that will make the user get stuck. I would appreciate some pointers to proceed

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
WaltonG: I think to check for a few rows there has to be some rows in the received data i.e `data["rows"].length > 0` which implies there was no parse error.

antoine2711: @WaltonG : I think we could refuse to import a table that is 1x1. Anything above that is fine. But at least one row and one column. 

Even 1x1, we could accept, but no user would do that. And the solution would be to add an return at the end of the file.

It will be a good improvement from 1x5 minimum ! ;-)

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
```diff
@@ -265,10 +265,13 @@ Refine.DefaultImportingController.prototype.getPreviewData = function(callback,
...
-          result.rowModel = data;
-          callback(result);
-        },
+		  function(data) {
+			  if (data["rows"].length == 0) {
+				alert($.i18n('core-index-import/load-xml-rows-error'));
+			  }
+			  result.rowModel = data;
+			  callback(result);
+		  },
```
antoine2711, line 274: You have a spacing problem. Also, if you get there, it's because there's an error. It could be a simple network error… (They are VERY common…) Not sure why you check that, and why you don't report other errors.

In terms of user interface, if it fails there, should we go back?

Regards, Antoine

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
antoine2711: > @antoine2711: I cant seem to get the errors that will make the user get stuck. I would appreciate some pointers to proceed

Network error. It's an API call, no?

antoine2711 (changes requested):
The error handling is still to manage more.

Regards, Antoine

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
WaltonG: The original code was:
https://github.com/OpenRefine/OpenRefine/blob/8e5e06a4b11843bbfcb5d1db128c08aef46d2b5c/main/webapp/modules/core/scripts/index/default-importing-controller/controller.js#L268-L271

WaltonG: The original code was:
https://github.com/OpenRefine/OpenRefine/blob/8e5e06a4b11843bbfcb5d1db128c08aef46d2b5c/main/webapp/modules/core/scripts/index/default-importing-controller/controller.js#L268-L271

Actually there is no error. Because of the row limit, if rows < 6  then no rows are shown and no error is mentioned, hence I added the check to atleast mention an error.

WaltonG: On further analysis this check is supposed to be in xml-parser-ui.js. Am a pushing commit for further review.

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
WaltonG: > @WaltonG : I think we could refuse to import a table that is 1x1. Anything above that is fine. But at least one row and one column.
> 
> Even 1x1, we could accept, but no user would do that. And the solution would be to add an return at the end of the file.
> 
> It will be a good improvement from 1x5 minimum ! ;-)

This would mean by-passing  the 1x5 minimum check in the backend. Would that be in order ?

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
```diff
@@ -265,10 +265,13 @@ Refine.DefaultImportingController.prototype.getPreviewData = function(callback,
           "limit" : numRows || 100 // More than we parse for preview anyway
         }),
         null,
-        function(data) {
```
antoine2711, line 268: This is the callback for a $.POST with jQuery. But, where is the failure callback?

See this example:
https://github.com/OpenRefine/OpenRefine/blob/577caabec1cce968df56156d7561a1ffde4b01c4/main/webapp/modules/core/scripts/preferences.js#L85-L91

Here, the error from the API is handled. At the very least, a dialog should be displayed. But maybe a move back in the interface also…

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
antoine2711: > @antoine2711: This would mean by-passing the 1x5 minimum check in the backend. Would that be in order ?

Have you investigated it?!

WaltonG: The usage found so far is when parsing XML :
https://github.com/OpenRefine/OpenRefine/blob/8e5e06a4b11843bbfcb5d1db128c08aef46d2b5c/main/src/com/google/refine/importers/tree/XmlImportUtilities.java#L122-L130

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
WaltonG: Sure there needs to be a faliure callback . To gain more insight on error handling on callbacks and how to implement it I have searched the code base but there seems to be no code the hanldles  callback errors. Could I be missing something ?

antoine2711: > @antoine2711: Sure there needs to be a faliure callback . To gain more insight on error handling on callbacks and how to implement it I have searched the code base but there seems to be no code the hanldles callback errors. Could I be missing something ?

@WaltonG: there is at least ONE place that handles the errors of API well, and it's in the setPrefs and getPrefs. I know because I coded it. And yes, you are right to point that it is missing in a lot of places.

Now, I could ask you to fix many (all of them?) in code surrounding yours, but It would take a lot of time. What I want here is to make sure that you know at least how to write it ONCE. That's my mentor work. And you have learned enough for me to ask you this level of quality. ;-) Keep up the good work!

Regards, Antoine

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
antoine2711: > The usage found so far is when parsing XML (...)

Could you explain to me why there is this limitation? And should it also apply to JSON?!

https://github.com/OpenRefine/OpenRefine/blob/8e5e06a4b11843bbfcb5d1db128c08aef46d2b5c/main/src/com/google/refine/importers/tree/XmlImportUtilities.java#L161

And also, is the 5 rows limit coming from this?

Regards, Antoine

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
```diff
@@ -270,7 +270,9 @@ Refine.DefaultImportingController.prototype.getPreviewData = function(callback,
 			  callback(result);
 		  },
         "jsonp"
-      );
+	   ).fail(function(jqXhr, textStatus, errorThrown) {
+		  alert($.i18n('core-index/rows-loading-failed'));
+	  });
```
antoine2711, line 275:
```suggestion
	   ).fail(() => { alert($.i18n('core-index/rows-loading-failed')); });
```
Since you don't use the param, might as well use an arrow function, don't you think? It makes the code simplier.

Regards, Antoine

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
WaltonG: No the limit does not apply to JSON however I didnt get any iplementation of the check in the code base and on further analysis I managed to create a project from an element with 1 record

```
<wb:data xmlns:wb="http://www.worldbank.org" >
        <wb:indicator id="SP.POP.TOTL">Population, total</wb:indicator>
</wb:data>
```
image

Could it have been fixed in one of the releases  ?
>From [dfhu...@gmail.com](mailto:dfhu...@gmail.com) on May 23, 2011 13:55:50:
I think this problem occurs because there isn't enough content in the file for Refine to guess where the record elements are. The next version of Refine will require you to pick the record elements yourself and hopefully won't have this problem.

WaltonG: No the limit does not apply to JSON however I didnt get any iplementation of the check in the code base and on further analysis I managed to create a project from an element with 1 record

```
<wb:data xmlns:wb="http://www.worldbank.org" >
        <wb:indicator id="SP.POP.TOTL">Population, total</wb:indicator>
</wb:data>
```
image

Could it have been fixed in one of the releases  ?
>From [dfhu...@gmail.com](mailto:dfhu...@gmail.com) on May 23, 2011 13:55:50:
I think this problem occurs because there isn't enough content in the file for Refine to guess where the record elements are. The next version of Refine will require you to pick the record elements yourself and hopefully won't have this problem.

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
```diff
@@ -265,12 +265,14 @@ Refine.DefaultImportingController.prototype.getPreviewData = function(callback,
           "limit" : numRows || 100 // More than we parse for preview anyway
         }),
         null,
-        function(data) {
-          result.rowModel = data;
-          callback(result);
-        },
+		function(data) {
+			  result.rowModel = data;
```
antoine2711, line 269: I see you removed the error checking here. I think it's a mistake. I think that if the API connection works, we end up here, but still, we might receive no valid data from the back-end, because input data was invalid.

@WaltonG: below (`.fail(function()`), it an error check for an API CONNECTION. Lost of network like. I don't care that we have the same error message, though an error message saying the connection was the problem, could be good. Also, if you check jQuery doc, you'll see that both textStatus and errorThrown can have error information.

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
antoine2711: > > From [dfhu...@gmail.com](mailto:dfhu...@gmail.com) on May 23, 2011 13:55:50:
> > I think this problem occurs because there isn't enough content in the file for Refine to guess where the record elements are. The next version of Refine will require you to pick the record elements yourself and hopefully won't have this problem.

That is great! I guess you have the honor to do the fixe, since we choose the record since a few version. So, we do not this constraint any longer.

Regards, Antoine

main/webapp/modules/core/scripts/index/default-importing-controller/controller.js
WaltonG: I moved the error check to xml-parser-ui.js because the error message was specific to xml. We can have the error check back here and generalize the error message to cater for both xml and json

antoine2711: > I moved the error check to xml-parser-ui.js because the error message was specific to xml. We can have the error check back here and generalize the error message to cater for both xml and json

That sound logic @WaltonG. Both moving the data error to the function that handles the data, while still checking network & communication errors at the level of the function that initiate the API call. 

So yes, implementing such a check for the JSON (or any code that receives data from this) would be mandatory, since you removed a data check there that was shared.

wetneb (approved):
So, I had a good look at this, and if I am not wrong:
* issue #387 is actually already solved: I could at least not reproduce it. The comment by @dfhuynh in that issue actually indicates that it was already meant to be solved by the interface to pick the particular record in the XML/JSON file, which indeed exists now. So, we could close the issue without any PR.
* this PR removes some dead code that is indeed leftover from the previous record selection mechanism, and adds some more error handling on the frontend. Although it does not "solve" the issue (since it is already fixed), that is always nice to have!

So, happy to merge this.

wetneb:
@antoine2711 since you have a "changes requested" review I am letting you approve and merge

antoine2711 (approved):
Nice work. Thank you @WaltonG.

Regards, Antoine

PR #4706, opened 2022-04-06
Improved browser launch in Linux distribution

### LINKED ISSUE #2640: Improve browser launch in Linux distribution
**Describe the bug**

It would be worth making sure we are using the most appropriate API call to launch the browser on Linux. In some circumstances, this can happen:
```
You have 15876M of free memory.
https://github.com/OpenRefine/OpenRefine/wiki/FAQ:-Allocate-More-Memory
Starting OpenRefine at 'http://127.0.0.1:3333/'

15:02:40.754 [            refine_server] Starting Server bound to '127.0.0.1:3333' (0ms)
15:02:40.755 [            refine_server] refine.memory size: 1400M JVM Max heap: 1407188992 (1ms)
15:02:40.763 [            refine_server] Initializing context: '/' from '/home/projadmin/dj/openrefine-3.3/webapp' (8ms)
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/projadmin/dj/openrefine-3.3/server/target/lib/slf4j-log4j12-1.7.18.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/projadmin/dj/openrefine-3.3/webapp/WEB-INF/lib/slf4j-log4j12-1.7.18.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Log4jLoggerFactory]
15:02:41.204 [                   refine] Starting OpenRefine 3.3 [58b839b]... (441ms)
15:02:41.204 [                   refine] initializing FileProjectManager with dir (0ms)
15:02:41.204 [                   refine] /home/projadmin/.local/share/openrefine (0ms)
15:02:44.359 [                   refine] Sorry, some error prevented us from launching the browser for you.
```

It would be worth checking that we are using the best approach to open `http://localhost:3333/` - the API to do this might have evolved. We could try different APIs in turn, too (making sure that we do not launch it twice).

 - OS: Linux

 - Version 3.3

**Additional information**

Mailing list thread: https://groups.google.com/forum/#!topic/openrefine/oo4W-zlSqc8

Binita-tech:
Fixes #2640 

- Improved browser launch in Linux distribution

### DIFF
--- a/server/src/com/google/refine/Refine.java
+++ b/server/src/com/google/refine/Refine.java
@@ -50,6 +50,7 @@ import java.util.concurrent.TimeUnit;
 
 import javax.swing.JFrame;
 
+import org.apache.commons.lang.SystemUtils;
 import org.apache.log4j.Level;
 import org.eclipse.jetty.server.HttpConfiguration;
 import org.eclipse.jetty.server.HttpConnectionFactory;
--- a/server/src/com/google/refine/Refine.java
+++ b/server/src/com/google/refine/Refine.java
@@ -494,11 +494,11 @@ class RefineClient extends JFrame implements ActionListener {
     } 
     
     private void openBrowser() {
-        if (!Desktop.isDesktopSupported()) {
+        if (!Desktop.getDesktop().isSupported(Desktop.Action.BROWSE)) {
             try {
-                openBrowser2();
+                openBrowserFallback();
             } catch (IOException e) {
-                logger.error("Failed to open browser. Please open %s in your browser",uri.toString());
+                throw new RuntimeException(e);
             }
         } else {
             try {
@@ -509,28 +509,17 @@ class RefineClient extends JFrame implements ActionListener {
         }
     }
 
-    private void openBrowser2() throws IOException {
-        logger.info("Opening browser to {}", uri.toString());
-        String os = System.getProperty("os.name").toLowerCase();
-        Runtime runtime = Runtime.getRuntime();
-        if(os.contains("win")) { // If we are on Windows
-            runtime.exec("rundll32 url.dll,FileProtocolHandler " + uri);
-        } else if(os.contains("mac")) { // If we are on Mac
-            runtime.exec("open " + uri);
-        } else if(os.contains("nix") || os.contains("nux")) { // If we are on unix
-            String[] browsers = {"google-chrome", "epiphany", "firefox", "mozilla", "konqueror",
-                    "netscape", "opera", "links", "lynx" };
-            StringBuilder cmd = new StringBuilder();
-            for (int i = 0; i < browsers.length; i++) {
-                if (i == 0) {
-                    cmd.append(String.format("%s \"%s\"", browsers[i], uri));
-                } else {
-                    cmd.append(String.format(" || %s \"%s\"", browsers[i], uri));
-                }
-            }
-            runtime.exec(new String[] { "sh", "-c", cmd.toString() });
+    private void openBrowserFallback() throws IOException {
+        Runtime rt = Runtime.getRuntime();
+
+        if(SystemUtils.IS_OS_WINDOWS) {
+            rt.exec("rundll32 url.dll,FileProtocolHandler " + uri);
+        } else if (SystemUtils.IS_OS_MAC_OSX) {
+            rt.exec("open " + uri);
+        } else if (SystemUtils.IS_OS_LINUX) {
+            rt.exec("xdg-open " + uri);
         } else {
-            logger.warn("Java Desktop class not supported on this platform");
+            throw new RuntimeException("Unable to open browser");
         }
     }
 }
--- a/server/src/com/google/refine/Refine.java
+++ b/server/src/com/google/refine/Refine.java
@@ -520,7 +520,7 @@ class RefineClient extends JFrame implements ActionListener {
         } else if (SystemUtils.IS_OS_LINUX) {
             rt.exec("xdg-open " + uri);
         } else {
-            throw new RuntimeException("Unable to open browser");
+            logger.warn("Java Desktop class not supported on this platform. Please open %s in your browser",uri.toString());
         }
     }
 }
--- a/server/src/com/google/refine/Refine.java
+++ b/server/src/com/google/refine/Refine.java
@@ -495,12 +495,42 @@ class RefineClient extends JFrame implements ActionListener {
     
     private void openBrowser() {
         if (!Desktop.isDesktopSupported()) {
-            logger.warn("Java Desktop class not supported on this platform.  Please open %s in your browser",uri.toString());
+            try {
+                openBrowser2();
+            } catch (IOException e) {
+                logger.error("Failed to open browser. Please open %s in your browser",uri.toString());
+            }
+        } else {
+            try {
+                Desktop.getDesktop().browse(uri);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
         }
-        try {
-            Desktop.getDesktop().browse(uri);
-        } catch (IOException e) {
-            throw new RuntimeException(e);
+    }
+
+    private void openBrowser2() throws IOException {
+        logger.info("Opening browser to {}", uri.toString());
+        String os = System.getProperty("os.name").toLowerCase();
+        Runtime runtime = Runtime.getRuntime();
+        if(os.contains("win")) { // If we are on Windows
+            runtime.exec("rundll32 url.dll,FileProtocolHandler " + uri);
+        } else if(os.contains("mac")) { // If we are on Mac
+            runtime.exec("open " + uri);
+        } else if(os.contains("nix") || os.contains("nux")) { // If we are on unix
+            String[] browsers = {"google-chrome", "epiphany", "firefox", "mozilla", "konqueror",
+                    "netscape", "opera", "links", "lynx" };
+            StringBuilder cmd = new StringBuilder();
+            for (int i = 0; i < browsers.length; i++) {
+                if (i == 0) {
+                    cmd.append(String.format("%s \"%s\"", browsers[i], uri));
+                } else {
+                    cmd.append(String.format(" || %s \"%s\"", browsers[i], uri));
+                }
+            }
+            runtime.exec(new String[] { "sh", "-c", cmd.toString() });
+        } else {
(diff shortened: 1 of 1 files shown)


elroykanye (commented):
Thanks so much for this @Binita-tech . I was able to try this out on Kali Linux running a KDE desktop env and it worked just fine!

How about windows? Did you face any issues?

server/src/com/google/refine/Refine.java
```diff
@@ -494,12 +494,42 @@ public void actionPerformed(ActionEvent e) {
     
     private void openBrowser() {
         if (!Desktop.isDesktopSupported()) {
-            logger.warn("Java Desktop class not supported on this platform.  Please open %s in your browser",uri.toString());
+            try {
+                openBrowser2();
+            } catch (IOException e) {
+                logger.warn("Java Desktop class not supported on this platform.  Please open %s in your browser",uri.toString());
```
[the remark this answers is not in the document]
Binita-tech: Done

Binita-tech:
Yes, it is working fine in windows @elroykanye

elroykanye:
Awesome

Binita-tech:
Please review @wetneb

server/src/com/google/refine/Refine.java
```diff
@@ -494,12 +494,42 @@ public void actionPerformed(ActionEvent e) {
...
+        logger.info("Opening browser to {}", uri.toString());
+        String os = System.getProperty("os.name").toLowerCase();
+        Runtime runtime = Runtime.getRuntime();
+        if(os.contains("win")) { // If we are on Windows
+            runtime.exec("rundll32 url.dll,FileProtocolHandler " + uri);
+        } else if(os.contains("mac")) { // If we are on Mac
+            runtime.exec("open " + uri);
+        } else if(os.contains("nix") || os.contains("nux")) { // If we are on unix
+            String[] browsers = {"google-chrome", "epiphany", "firefox", "mozilla", "konqueror",
+                    "netscape", "opera", "links", "lynx" };
```
wetneb, line 521: Perhaps that is a bit much… Also, it will try browsers in this exact order, whereas we might want to open the URL with the default browser instead. How about using `xdg-open` (and potentially its alternatives) instead?

server/src/com/google/refine/Refine.java
```diff
@@ -494,12 +494,42 @@ public void actionPerformed(ActionEvent e) {
...
+                throw new RuntimeException(e);
+            }
         }
-        try {
-            Desktop.getDesktop().browse(uri);
-        } catch (IOException e) {
-            throw new RuntimeException(e);
+    }
+
+    private void openBrowser2() throws IOException {
```
wetneb, line 511: I would prefer a more informative name, like `openBrowserWithoutDesktopSupport` or `openBrowserFallback`.

server/src/com/google/refine/Refine.java
```diff
@@ -494,12 +494,42 @@ public void actionPerformed(ActionEvent e) {
...
+        logger.info("Opening browser to {}", uri.toString());
+        String os = System.getProperty("os.name").toLowerCase();
+        Runtime runtime = Runtime.getRuntime();
+        if(os.contains("win")) { // If we are on Windows
```
wetneb, line 515: Perhaps OS detection would be a bit cleaner if we used a library (which we already rely on): https://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/SystemUtils.html#IS_OS_WINDOWS

server/src/com/google/refine/Refine.java
```diff
@@ -494,12 +494,42 @@ public void actionPerformed(ActionEvent e) {
     
     private void openBrowser() {
         if (!Desktop.isDesktopSupported()) {
-            logger.warn("Java Desktop class not supported on this platform.  Please open %s in your browser",uri.toString());
+            try {
+                openBrowser2();
+            } catch (IOException e) {
+                logger.error("Failed to open browser. Please open %s in your browser",uri.toString());
```
wetneb, line 500: Looking at your code, `openBrowser2` can fail without throwing an `IOException` at the moment, so this message might not be displayed in those cases. The control flow of the two methods could probably be adapted to avoid this.

server/src/com/google/refine/Refine.java
Binita-tech:
```
import java.awt.Desktop;
import java.io.IOException;
import java.net.URI;
import java.net.URISyntaxException;

public class MultiBrowPop {

    public static void main(String[] args) {
        OUT("\nWelcome to Multi Brow Pop.\nThis aims to popup a browsers in multiple operating systems.\nGood luck!\n");

        String url = "http://www.birdfolk.co.uk/cricmob";
        OUT("We're going to this page: "+ url);

        String myOS = System.getProperty("os.name").toLowerCase();
        OUT("(Your operating system is: "+ myOS +")\n");

        try {
            if(Desktop.isDesktopSupported()) { // Probably Windows
                OUT(" -- Going with Desktop.browse ...");
                Desktop desktop = Desktop.getDesktop();
                desktop.browse(new URI(url));
            } else { // Definitely Non-windows
                Runtime runtime = Runtime.getRuntime();
                if(myOS.contains("mac")) { // Apples
                    OUT(" -- Going on Apple with 'open'...");
                    runtime.exec("open " + url);
                } 
                else if(myOS.contains("nix") || myOS.contains("nux")) { // Linux flavours 
                    OUT(" -- Going on Linux with 'xdg-open'...");
                    runtime.exec("xdg-open " + url);
                }
                else 
                    OUT("I was unable/unwilling to launch a browser in your OS :( #SadFace");
            }
            OUT("\nThings have finished.\nI hope you're OK.");
        }
        catch(IOException | URISyntaxException eek) {
            OUT("**Stuff wrongly: "+ eek.getMessage());
        }
    }
```
Will this be ok to use @wetneb as it is using xdg-open?

wetneb: It looks pretty good to me indeed, thank you!

elroykanye:
@wetneb , do you think it is best to replace the current method `Desktop.browse(url)` with `Runtime.exec(cmd)` completely?
In this way, we won't need the openBrowser() method anymore, or we could just replace its content with @Binita-tech 's `openBrowser2()` implementation. 

I suggest this since it is a cross-platform solution...

wetneb:
Hmm, intuitively I would still use `Desktop.browse` when it is supported - it is more cross-platform than this new fallback strategy (since we do not need to distinguish between OSes there). Or did I misunderstand something?

elroykanye:
> Hmm, intuitively I would still use `Desktop.browse` when it is supported - it is more cross-platform than this new fallback strategy (since we do not need to distinguish between OSes there). Or did I misunderstand something?

Oh! You are right. I am looking into @Binita-tech 's implementation again.

server/src/com/google/refine/Refine.java
```diff
@@ -494,12 +494,42 @@ public void actionPerformed(ActionEvent e) {
     
     private void openBrowser() {
         if (!Desktop.isDesktopSupported()) {
```
elroykanye, line 496: This returns true in every operating system apparently, because the class itself is supported. We need to specifically determine if the action we want to perform is.
According to this [link](https://stackoverflow.com/questions/5226212/how-to-open-the-default-webbrowser-using-java/21676290#21676290), GNOME libraries are required for the `Desktop.browse` action. Unfortunately, it is possible to use Linux without them (for example, I use KDE as my desktop env and this solution does not work for me).

A more precise approach will be to replace this check with this:
```java
if(!Desktop.getDesktop().isSupported(Desktop.Action.BROWSE)) {
```
The `openBrowser()` method is indeed very necessary.

server/src/com/google/refine/Refine.java
```diff
@@ -494,12 +494,42 @@ public void actionPerformed(ActionEvent e) {
...
+            for (int i = 0; i < browsers.length; i++) {
+                if (i == 0) {
+                    cmd.append(String.format("%s \"%s\"", browsers[i], uri));
+                } else {
+                    cmd.append(String.format(" || %s \"%s\"", browsers[i], uri));
+                }
+            }
+            runtime.exec(new String[] { "sh", "-c", cmd.toString() });
+        } else {
+            logger.warn("Java Desktop class not supported on this platform");
```
elroykanye, line 532: @Binita-tech 
You may surely have modified to using the `xdg-open` and `SystemUtils` recommendation.
I think your else here should throw a RuntimeException so the `openBrowser()` can know what to do from there.

elroykanye (changes requested):
Here is what I noticed may need to be updated about the code because I tried this currently (even with the recommendations from @wetneb ) but the only thing blocking was the:

```java
if(!Desktop.isDesktopSupported()) {
```
Check it out :)

elroykanye (approved):

elroykanye:
@wetneb , I checked out @Binita-tech 's update and they correspond with what I suggested. Awaiting your response. :)

server/src/com/google/refine/Refine.java
```diff
@@ -494,13 +495,32 @@ public void actionPerformed(ActionEvent e) {
...
+        Runtime rt = Runtime.getRuntime();
+
+        if(SystemUtils.IS_OS_WINDOWS) {
+            rt.exec("rundll32 url.dll,FileProtocolHandler " + uri);
+        } else if (SystemUtils.IS_OS_MAC_OSX) {
+            rt.exec("open " + uri);
+        } else if (SystemUtils.IS_OS_LINUX) {
+            rt.exec("xdg-open " + uri);
+        } else {
+            throw new RuntimeException("Unable to open browser");
```
wetneb, line 523: Perhaps in this last case we should resort to the previous behaviour (`logger.warn("Java Desktop class not supported on this platform.  Please open %s in your browser",uri.toString());`) because otherwise by throwing the exception we will likely stop OpenRefine completely, and that is probably not the desired behaviour.

Binita-tech: Done

wetneb (approved):

PR #4707, opened 2022-04-06
initial research on Weblate

### LINKED ISSUE #4197: Provide more context to translators about the message they are translating
As a translator, it would sometimes be useful to know a bit more context about the message being translated: beyond the message itself, a description of where the message is used in the UI, or even a screenshot of the UI in English where that message appears.

### To Reproduce
Go to Weblate and start translating: sometimes it is not so clear where the messages you are translating will appear.
### Tasks

- [ ] Figure out what Weblate actually supports to contextualize messages
- [ ] Investigate how we could use those features for the OpenRefine project

### Additional context

In any case we should not over-use this feature: a lot of messages are self-explanatory and can be translated without much context (things like "OK", "Cancel"…).

Babi-B:
Fixes #4197

- Outlines understanding of Weblate internals

### DIFF
--- a/docs/docs/technical-reference/translation-context.md
+++ b/docs/docs/technical-reference/translation-context.md
@@ -32,9 +32,25 @@ Alongside these, there is also a glossary of words to help the translator. The w
 
 In summary, Weblate uses a **text-based** support for providing message contexts
 
-## Outline Of How OpenRefine Could Implement the Weblate Support for Message Contextualization
+## Outline Of How OpenRefine Could Utilize the Weblate Support for Message Contextualization
 
 ### Setting Up The Glossary
 As previously mentioned, every project in Weblate has a glossary. The glossary, as Weblate [defines](https://lstn.wolba.ch/weblate-terms/#:~:text=Every%20project%20on%20Weblate%20is,consistently%20translated%20throughout%20the%20project.), contains strings that may repeat throughout different components and ensures that these strings will be consistently translated throughout the project.
 
 All OpenRefine projects should have glossaries with terminologies specific to OpenRefine. Words added to the glossary should be those that are used multiple times in the application. For example clusters, facets, reconciliation, metadata, schema, tags, etc
+
+### Using Weblate's `nearby texts`, `history`, and `other languages` options
+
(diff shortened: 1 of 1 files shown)


Babi-B:
@antoine2711 @wetneb before continuing I wish to know if I am on the right track with what is requested in this [issue](https://github.com/OpenRefine/OpenRefine/issues/4197)

antoine2711:
@Babi-B : good thing you ask. The objective of the issue, in the way I understand it, to document the translation process (which you did well from what I see), but rather to starting using a context string.

So right now, for translations, we have the key and the text. The question is to know: can we add a context string to help us know where the translation will be used? If so, can an example show how-to?

I think your documentation is good, but it's not what needed. The issue is about addind context to translation strings. Can we do it?

I hope my explanations are clear. Don't hesitate to ask more questions.

Regards, Antoine

Babi-B:
Hello @antoine2711. I have redone the report based on my understanding of your comment. Before I continue I wish to know if I am now on par with what is needed

antoine2711:
I think you are overdocumenting. Here's then task from the issue:
>- [ ] Figure out what Weblate actually supports to contextualize messages
>- [ ] Investigate how we could use those features for the OpenRefine project

The need here is not documentation or to write a file. It's interesting, but our real aim is to incorporate that in OR, if we can… ;-) So this issue is about having an answer: can we — or not — add a context string with Weblate…

I read your documentation, and I can't make for sure if we have the possibility now…

What I see with a first glance at the software is that I can't put a context string in translated strings, but, in the glossary, Weblate seams to have such a feature…

Translated string edit pane:
image

Glossary string edit pane:
image

In the second picture, the glossary « QuickStatements » string, I have put some context data in the « Explanation » field ("Software for mass transfert to Wikidata."). It would be great if we could do the same with translated string… Can you find out? That’s the issue. Not writing documentation.

Understanding an issue is never easy, and it's often a lot of discussions. If I'm not clear, feel free to ask again. I'm here for you. You can also email me: antoine@beaubien.qc.ca.

Regards, Antoine

wetneb:
Thanks @antoine2711 for the thorough review! I agree this should not go in the repository itself. @Babi-B feel free to take notes in the issue itself and open a PR when you are ready to make actual changes.

PR #4717, opened 2022-04-08
(I#2442) Setup glossary for translations with Weblate 

### LINKED ISSUE #2442: Set up glossary for translations with Weblate
We currently use [Weblate](https://hosted.weblate.org/engage/openrefine/?utm_source=widget) to translate the user interface. Translators get a list of untranslated strings (or strings needing review) in their language, write translations, and this gets committed to the repository via regular pull requests.

Weblate makes it possible to set up a [glossary](https://docs.weblate.org/en/latest/user/translating.html#glossary), which determines how the terminology that is specific to OpenRefine should be translated in each language. For instance, terms like "facet", "clustering" or "reconciliation" have a pretty specific meaning in our context - so we want to make sure these terms are translated uniformly in all strings.

We should look into setting up such a glossary and identifying the terms that should go there.

WaltonG:
Fixes #2442

- Added glossary in json format
- 
-


wetneb (commented):
Thanks!
I wonder if this is the right place to store the glossary: you have put it in the `docs/docs` folder, which has nothing to do with localization, right?
Looking at weblate, it looks like we already have something there, so perhaps we do not need to store anything in the repository and the issue can be closed, no?
https://hosted.weblate.org/projects/openrefine/glossary/

antoine2711:
> Looking at weblate, it looks like we already have something there, so perhaps we do not need to store anything in the repository and the issue can be closed, no?

@wetneb: what we have in Weblate is a very good start in English. But I lack to see where we can link different language for the Glossary. From the description of this issue, you clearly stated that you wanted to link Facet with Facette. But nothing so far as address that part of the issue.

I think the issue still need work, but correct me if I'm wrong.

Also, having the values of the Glossary saved along it's corresponding Repo evolution seams a good enough reason to have the JSON export in our repo. Am I wrong? But I agree the selected place might not be the best.

Regards, Antoine

wetneb:
OK! I haven't actually looked deep into how Weblate's glossary currently works.

wetneb:
@antoine2711 @trnstlntk can I leave this PR to you then, as you have been thinking about this topic recently?

trnstlntk (commented):
While I will decline this PR, I think the definitions you have created here are certainly useful and I will look into integrating them into Weblate itself (through its front end user interface).

trnstlntk:
> @antoine2711 @trnstlntk can I leave this PR to you then, as you have been thinking about this topic recently?

Yes, I'm happy to look at this further.

I translate OpenRefine's interface to my native language, Dutch. I think I agree with @wetneb's assessment that a JSON file in our own repo is not what we need here; the glossary should be maintained and updated through Weblate's end user interface by the community of translators, and translations should be added in each language there. So perhaps this issue has been a bit out of place here.

That said @WaltonG: although I will be declining your PR, I think your work has not in vain at all - I will take the files you created and will look into integrating the content additions in Weblate itself. Thanks for looking into this!

antoine2711:
@trnstlntk: I think we can keep this issue openend as there are still questions not answered.

Yes, we currently have a glossary on Weblate, but it's only in English. We should also have the translation/link to our main languages IN the glossary. The picture of the documentation suggest it is possible.

> that a JSON file in our own repo is not what we need here

The data for the glossary is already on Weblate, and will stay there in any case. But, on Weblate, we have no history, only the latest version. By keeping a JSON on our repo, it stores the changes along the way with no efforts. Also, have a JSON file enables manipulation, like building a table for translation, etc. This is just a small +, there probably exists an export option on Weblate, but it's not obvious.

Also, I still have questions about the glossary.
1. Should the terms that starts with an uppercase, or a lowercase?
2. Should we keep the plural form, or the singular, or both?!

I think we could still investigate a bit this issue.

Regards, Antoine

trnstlntk:
> Yes, we currently have a glossary on Weblate, but it's only in English. We should also have the translation/link to our main languages IN the glossary. The picture of the documentation suggest it is possible.

I actually added quite a few terms to the general glossary ('Terminology') through / from Dutch. And those terms that I added, are now available for other languages to translate as well (example of [untranslated terminology in French](https://hosted.weblate.org/browse/openrefine/glossary/fr/?q=state:%3Ctranslated))

That said, Weblate's own interface for finding, managing and working on this multilingual glossary is not optimal, and their [documentation](https://docs.weblate.org/no/latest/user/glossary.html) is only super basic and quite technically oriented.

> > that a JSON file in our own repo is not what we need here
> 
> The data for the glossary is already on Weblate, and will stay there in any case. But, on Weblate, we have no history, only the latest version. By keeping a JSON on our repo, it stores the changes along the way with no efforts. Also, have a JSON file enables manipulation, like building a table for translation, etc. This is just a small +, there probably exists an export option on Weblate, but it's not obvious.

Yeah, I also saw indications in Weblate's documentation that such automated export / syncing should be possible. But how/where? I've clicked through the entire admin interface but could not find it.

My main perspective is that of our translators (who will usually not have a technical background). For their ease of use I think it's best to keep all functionalities in Weblate itself (it's already not obvious to point people to Weblate; how will we make it clear to them that they need to translate the glossary elsewhere?). I see the arguments re: usefulness of a JSON backup, but for me, more and better translations (and an easy, single interface for that) are the main priority.

> Also, I still have questions about the glossary.
> 
> 1. Should the terms that starts with an uppercase, or a lowercase?
> 2. Should we keep the plural form, or the singular, or both?!

I have these questions too. This is where Weblate's own documentation is really lacking. They explain (extremely minimally) where to find certain functionalities, but there are no indications or examples of the 'social' side, of best practices for communities.
I guess, for now, we'll have to figure this out for ourselves.

My first impression is that:
1. We can use lowercase everywhere - Weblate will recognize if a term is used with capitalization
2. But we need to create entries for each singular/plural/abbreviation variant (`column`/`columns`) because Weblate does _not_ recognize the difference between these.

antoine2711:
> That said, Weblate's own interface for finding, managing and working on this multilingual glossary is not optimal, and their [documentation](https://docs.weblate.org/no/latest/user/glossary.html) is only super basic and quite technically oriented.

@trnstlntk: maybe we should compensate with our own documentation, simple and centered around the task our contributors do.

> Yeah, I also saw indications in Weblate's documentation that such automated export / syncing should be possible. But how/where? I've clicked through the entire admin interface but could not find it.

Having a copy file in our repo would make that data available and seeable. Behind Weblate's servers, is harder to see/manipulate. While still being a small cost, I also a small advantage. So I have not a strong opinion. But I do see some value.

> how will we make it clear to them that they need to translate the glossary elsewhere?).

I think there might be a misunderstanding here: users don't need to manipulate this file. They don't even need to know it exists.It makes nothing more complicated. Users will always go (or be able to go) to Weblate. It's not one or the other.

> I guess, for now, we'll have to figure this out for ourselves.
> 
> My first impression is that:
> 
> 1. We can use lowercase everywhere - Weblate will recognize if a term is used with capitalization
> 2. But we need to create entries for each singular/plural/abbreviation variant (`column`/`columns`) because Weblate does _not_ recognize the difference between these.

1. Agree
2. Wow. This is really a problem. Are they working on a solution for that... It seams hard to manually create these copies.

Overall, I would be in favor of having on our side more documentation, with our working process described.

Regards, Antoine

PR #4718, opened 2022-04-08
Improve clustering method documentation; make it skimmable

### LINKED ISSUE #3033: Document clustering algorithms
The documentation on the clustering algorithms is scant, and in some cases, non-existent. The user interface should provide a brief description of each algorithm as well as guidance on how to pick the most appropriate one for their application as well as any pertinent tradeoffs.

Some are obvious from their names (or at least easily Google-able), but other are much more opaque.

PPM is a particularly egregious example, but since I just looked at the implementation, here's what it does:

PPM stands for [Prediction by Partial Matching](https://en.wikipedia.org/wiki/Prediction_by_partial_matching) which is an arithmetic coding technique. The implementation that OpenRefine uses is based on the MIT SIMILE project's Vicino package, which, in turn, uses Bob Carpenter's arithcode package, which has a tutorial available here: https://raw.githubusercontent.com/bob-carpenter/java-arithcode/master/arithcode-1.2/doc/tutorial.html. The two strings to be compared are concatenated together and then "compressed" using the coder. The PPMModel uses 8 bytes of context and returns as the score the length of the encoded output. This is used four times to compute distance scores for (x,y), (y,x), (x,x), (y,y) (because the scores are directional) which are then combined together in this fashion to generate a final score: 

`10.0d * ((d(x,y) + d(y,x)) / (d(x,x) + d(y,y)) - 1.0d)`

This score is relatively computationally expensive to compute and has to be done in a pairwise fashion for every pair of strings. To prevent an N^2 explosion, both the "nearest neighbor" clustering algorithms first divide the strings into manageable sized blocks and only does the pairwise comparison for members of the block. Depending on the blocking algorithm used, it's possible for two similar strings to end up in different blocks and never get compared to each other.

Babi-B:
Fixes #3033

- makes clustering method documentation more skimmable for superficial readers


Babi-B:
> I'm not 100% sure what you are proposing when you say you could
>
> >make the [clustering-methods](https://docs.openrefine.org/manual/cellediting#clustering-methods) document for the end-users more skimmable
>
>I don't know if you can explain here before doing the work or if it's easier to make a PR for the docs and get feedback on that - I'd be happy either way

@ostephens @antoine2711. 

I thought perhaps it would be good to edit the content of the document so superficial readers could easily sieve through for what they want.

Babi-B:
@WaltonG thanks.  Missed that

wetneb:
Looking at the preview at https://deploy-preview-4718--openrefine-docs.netlify.app/next/manual/cellediting#cluster-and-edit, it looks fine to me, although I am not really sure what is the problem with [the original](https://docs.openrefine.org/next/manual/cellediting#cluster-and-edit) to be honest.

However the new version has an inconsistency in the hierarchy of headings, with the following list of subparts:
1. Fingerprinting
2. N-gram fingerprinting
3. Metaphone3 fingeprinting
4. …

The problem is that 1. is not a fingerprinting method on its own (rather an intro to the general technique of fingerprinting), so it should not be at the same hierarchical level as specific fingerprinting methods.

Also the indentation of the new version looks a bit off, perhaps because these numbered sub-headings are not actually meant to be used like that, but are more for enumerated lists?

Babi-B:
Ok @wetneb. So would you advice that I update the file to correct the inconsistencies or should I discard the changes all together?

antoine2711:
> Looking at the preview at https://deploy-preview-4718--openrefine-docs.netlify.app/next/manual/cellediting#cluster-and-edit, it looks fine to me, although I am not really sure what is the problem with [the original](https://docs.openrefine.org/next/manual/cellediting#cluster-and-edit) to be honest.

@wetneb: I would tend to disagree here.
image

In this image, Phonetic clustering is a section title, and yet, it's smaller than the next title, a sub-section of it. So the inconsistancy you spotted in @Babi-B works was already there.

Also, why does the Baider-Morse section doesn't get it paragraph? There is a paragraph break missing there.

As a bonus, I would put links to the relevant section in the summary above.
Cleaning the hierarchy and putting links would be, in my opinion, « making the doc more skimmable ».

Regards, Antoine

Babi-B:
Ok @antoine2711 . Thanks

docs/docs/manual/cellediting.md
```diff
@@ -103,55 +103,79 @@ You can also export the currently identified clusters as a JSON file, or close t
...
-    *   fingerprint
-    *   ngram-fingerprint
-    *   metaphone3
-    *   cologne-phonetic
-    *   Daitch-Mokotoff
-    *   Beider-Morse
-*   nearest neighbor
-    *   levenshtein
-    *   ppm
+The clustering pop-up window offers you a variety of clustering methods; 6 different key collision clustering methods and 2 nearest neighbour clustering methods.
```
antoine2711, line 110:
```suggestion
The clustering pop-up window offers you two categories of clustering methods: 6 different key collision clustering methods and 2 nearest neighbour clustering methods.
```

wetneb:
This PR adds two files, [docs/docs/manual/img.png](https://github.com/OpenRefine/OpenRefine/pull/4718/files#diff-9389edea530131d27b5867fa4b213432159d85e3f54089a4388c9b847cdcee08) und docs/docs/manual/img_1.png, but I cannot figure out why they are there. Should they be removed?

@antoine2711 can I leave it to you review and merge this?

antoine2711 (commented):
I think we are getting there. Is there a reason you deleted the summary? I think it helps to browse them.

Regards, Antoine

docs/docs/manual/cellediting.md
```diff
@@ -103,52 +103,95 @@ You can also export the currently identified clusters as a JSON file, or close t
...
-    *   metaphone3
-    *   cologne-phonetic
-    *   Daitch-Mokotoff
-    *   Beider-Morse
-*   nearest neighbor
-    *   levenshtein
-    *   ppm
+The clustering pop-up window offers you two categories of clustering methods: 6 different key collision clustering methods and 2 nearest neighbour clustering methods. 
+
+- [Key collision](#key-collision)
```
antoine2711, line 112:
```suggestion
- [Key Collision](#key-collision)
```

Could you capitalise every summary items? Like the title there are below. There is more than 10 that could be capitalized from what I saw.

docs/docs/manual/cellediting.md
```diff
@@ -103,52 +103,95 @@ You can also export the currently identified clusters as a JSON file, or close t
...
+N-gram fingerprinting allows you to set the _n_ value to whatever number you’d like and will create n-grams of _n_ size (after doing some cleaning), alphabetize them, then join them back together into a  fingerprint.
+
+**For example**, a 1-gram fingerprint will simply organize all the letters in the cell into alphabetical order - by creating segments one character in length. A 2-gram fingerprint will find all the two-character segments, remove duplicates, alphabetize them, and join them back together (for example, “banana” generates “ba an na an na,” which becomes “anbana”).
+
+This can help match cells that have typos, or incorrect spaces (such as matching “lookout” and “look out,” which fingerprinting itself won’t identify because it separates words). The higher the _n_ value, the fewer clusters will be identified. With 1-grams, keep an eye out for mismatched values that are near-anagrams of each other (such as “Wellington” and “Elgin Town”).
+
+For an in-depth understanding of N-gram fingerprinting, check this [document](https://github.com/OpenRefine/OpenRefine/wiki/Clustering-In-Depth#n-gram-fingerprint)
 
 ##### Phonetic clustering {#phonetic-clustering}
```
antoine2711, line 152:
```suggestion
_**Phonetic clustering {#phonetic-clustering}**_
```

Here, we have a complicated problem. The Header level 4 is much smaller than the paragraph title. So, could we put it the same size, but in italic? It's really an informative header…

antoine2711 (changes requested):
I like this. Only 2 small details.

Regards, Antoine

antoine2711 (approved):

antoine2711:
Good work! One more done, @Babi-B. 

Best Regards, Antoine

PR #4719, opened 2022-04-08
Technical documentation for clustering

### LINKED ISSUE #3033: Document clustering algorithms

Babi-B:
Fixes #3033

- describes metaphone3 algorithm for OpenRefine


Babi-B:
Hello @antoine2711. Excuse me but I edited the [wiki](https://github.com/OpenRefine/OpenRefine/wiki/Clustering-In-Depth#metaphone3) before realizing I had to do a pull request for your permission first. If you do not approve this changes I will take out what I had done.

antoine2711:
> Hello @antoine2711. Excuse me but I edited the [wiki](https://github.com/OpenRefine/OpenRefine/wiki/Clustering-In-Depth#metaphone3) before realizing I had to do a pull request (…).

Actually, you did right. We do not have a « request first » modus operandi, people just make changes there as they see fit.

We are slowly phasing out of the Wiki, so I would say: make sure that the information you add in the Wiki is also in the Docs, and that will pass thru our supervision mechanism, and reviewers will see you changes there, so if a correction to your work is requested in the PR for the Docs, you could also apply that correction to the Wiki.

Are my explanation clear? If not, please ask me to clarify.

Regards, Antoine

Babi-B:
>Are my explanation clear? If not, please ask me to clarify.

@antoine2711 it is clear. Thank you

wetneb:
@Babi-B it looks like your PR fails to build on Netlify: that might mean Docusaurus has found some errors in your changes. Can you have a look? You can click on the `Details` links in the checks for this PR to find out more.

Babi-B:
Ok @wetneb. I'm on it

Babi-B:
Hello @antoine2711, I have completed the documentation. Could you please review it?

wetneb (changes requested):
As far as I can tell, the page is still not reachable from the rest of the documentation. Also, this PR adds an image which should probably be deleted.

Babi-B:
@wetneb I don't know where I am making the error. I followed the instructions outlined in this [readme](https://github.com/OpenRefine/OpenRefine/tree/master/docs#readme) file and was able to run the docs locally (http://localhost:3000/next). 

I made some test changes to other files and they displayed however the file I added did not

wetneb:
You could add a link to your file from other files, or add it to the sidebar (see `sidebar.json`).

Babi-B:
@wetneb thank you. I added it successfully

wetneb (approved):
Thank you!

PR #4729, opened 2022-04-10
Adding checkbox for unprintable characters

shriyasankhyan:
Fixes{#1286}

I have added a checkbox so that if we mark the checkbox , we can toggle unprintable characters.

### DIFF
--- a/main/webapp/modules/core/project.vt
+++ b/main/webapp/modules/core/project.vt
@@ -76,6 +76,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
             <!-- row/record counts -->
           </div>
           <div bind="extensionBarDiv" id="extension-bar">
+            <input type = "checkbox" id = "toggle-display-characters" name = "nonPrintableCharacters">
+            <label class = "non-printable" type = "checkbox" for = "nonPrintableCharacters" > Non-Printable Characters</label>
             <span id="or-proj-ext"></span> <span bind="menuContainer" id="extension-bar-menu-container"></span>
           </div>
         </div>


wetneb:
Hi @shriyasankhyan, it looks like you have two pull requests for the same issue: #4729 (this one) and #4730. Could we have just one, which would solve the issue entirely, in one go?

shriyasankhyan:
> Hi @shriyasankhyan, it looks like you have two pull requests for the same issue: #4729 (this one) and #4730. Could we have just one, which would solve the issue entirely, in one go?

Yeah surely. But it is not completely solved yet. So should I wait and whenever it is completely solved, then only I should have a combined pull request or should send pull requests as long as the code starts working in bits?

shriyasankhyan:
> Yes it is better to open the pull request when it is ready for review, especially because we have a lot of new pull requests at the moment, due to Outreachy :)

Ok surely, then I shall wait and close both of these pull requests. Thanku

antoine2711:
> But it is not completely solved yet. So should I wait and whenever it is completely solved, then only I should have a combined pull request or should send pull requests as long as the code starts working in bits?

@shriyasankhyan: One other thing you can do if you know you are still working on a Pull Request, is to change it's status to a « Draft Pull Request ». That way, it's clear you are still working on it and we don't bother you about it.

Regards, Antoine

shriyasankhyan:
> > But it is not completely solved yet. So should I wait and whenever it is completely solved, then only I should have a combined pull request or should send pull requests as long as the code starts working in bits?
> 
> @shriyasankhyan: One other thing you can do if you know you are still working on a Pull Request, is to change it's status to a « Draft Pull Request ». That way, it's clear you are still working on it and we don't bother you about it.
> 
> Regards, Antoine

Ok Thankyou

PR #4730, opened 2022-04-10
Checkbox in the top right corner

### LINKED ISSUE #1286: Add a DISPLAY UNPRINTABLE CHARS toggle mode

shriyasankhyan:
Fixes #1286

Shows the non-printable characters checkbox in the top right corner


wetneb:
Hi @shriyasankhyan, it looks like you have two pull requests for the same issue: #4729 and #4730 (this one). Could we have just one, which would solve the issue entirely, in one go?

PR #4733, opened 2022-04-10
(I#2770) Improve error handling for JSON Import during Create Project

### LINKED ISSUE #2770: Improve error handling for JSON Import during Create Project
Attempting to import invalid JSON leaves the UI in an inconsistent state.

### To Reproduce
1. Import this data from the clipboard:
```
[
{"foo":NaN},
{"bar":Infinity}
]
```
2. Select "JSON Files" for "Parse data as..." 
1. Note that the preview doesn't show any data except for the JSON delimiters
1. Click on `{}` and note the error "Non-standard token 'NaN': enable JsonParser.Feature.ALLOW_NON_NUMERIC_NUMBERS to allow"
1. Dismiss the error alert dialog 
1. Note that the screen continues to show spinner with "Updating preview..."

### Expected behavior

* The "Updating Preview..." spinner should be dismissed on error (or when the error dialog is dismissed)
* The parse error should be shown immediately without the user having to select a record

### Versions
- Browser Version: Chrome 83
- OpenRefine  3.4-SNAPSHOT (HEAD of current master)

WaltonG:
Fixes #2770

- Use window alert to present error
- Reload window when alert is dismissed

### DIFF
--- a/main/src/com/google/refine/importers/tree/XmlImportUtilities.java
+++ b/main/src/com/google/refine/importers/tree/XmlImportUtilities.java
@@ -289,7 +289,7 @@ public class XmlImportUtilities extends TreeImportUtilities {
                 }
             }
         } catch (TreeReaderException e) {
-            logger.error("Exception from XML parse :" + e.getMessage(), e);
+            logger.error("Exception from XML parse :" + e.getMessage());
             return;
 
         }
--- a/main/src/com/google/refine/importers/tree/XmlImportUtilities.java
+++ b/main/src/com/google/refine/importers/tree/XmlImportUtilities.java
@@ -282,9 +282,8 @@ public class XmlImportUtilities extends TreeImportUtilities {
                 }
             }
         } catch (TreeReaderException e) {
-            logger.error("Exception from XML parse :" + e.getMessage());
-            return;
-
+            logger.error("Exception from XML parse",e);
+            throw e;
         }
     }
 
--- a/main/src/com/google/refine/importers/tree/XmlImportUtilities.java
+++ b/main/src/com/google/refine/importers/tree/XmlImportUtilities.java
@@ -289,8 +289,8 @@ public class XmlImportUtilities extends TreeImportUtilities {
                 }
             }
         } catch (TreeReaderException e) {
-            logger.error("Exception from XML parse :" + e.getMessage());
-            return;
+            logger.error("Exception from XML parse", e);
+            throw e;
 
         }
     }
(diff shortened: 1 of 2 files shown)


main/src/com/google/refine/importers/tree/XmlImportUtilities.java
```diff
@@ -356,23 +357,29 @@ static protected void findRecord(
...
+                                String desiredFieldName = recordPath[pathIndex + 1];
+                                String currentFieldName = parser.getFieldName();
+                                if (desiredFieldName.equals(currentFieldName)) {
+                                    processFieldAsRecord(project, parser, rootColumnGroup, trimStrings, storeEmptyStrings, guessDataTypes);
+                                }
                             }
                         }
                     }
+                } catch (TreeReaderException e) {
+                    logger.error("Exception from XML parse", e);
```
wetneb, line 380: That is not the best way to report errors to the user: in general, users should not have to look at the server logs to understand what is wrong. I would prefer to surface this message in the UI.

WaltonG: Very true, am working on it

antoine2711:
@WaltonG: I'm a bit confused here. The issue is about fixing a JSON importation error handling, and you are in the XmlImporter files. Do you plan to do both, JSON and XML error handling?

If so, it ok, just tell me. Otherwise, please help me understand.

Regards, Antoine

WaltonG:
> @WaltonG: I'm a bit confused here. The issue is about fixing a JSON importation error handling, and you are in the XmlImporter files. Do you plan to do both, JSON and XML error handling?
> 
> If so, it ok, just tell me. Otherwise, please help me understand.
> 
> Regards, Antoine

@antoine2711 I was also confused by the classname XmlImportUtilities. The class itself is a utility class that extends TreeImportUtilities and provides import utility methods for json and xml files. A different name for the class would have been better to avoid confusion

WaltonG:
> > @antoine2711 I was also confused by the classname XmlImportUtilities. The class itself is a utility class that extends TreeImportUtilities and provides import utility methods for json and xml files. A different name for the class would have been better to avoid confusion
> 
> @wetneb: would it be appropriate to rename the file from `XmlImportUtilities.java` to `TreeImportUtilities.java`? Does it have implications?
> 
> Regards, Antoine

@antoine2711 `TreeImportUtilities.java` already exists as a superclass of  `XmlImportUtilities.java`

main/src/com/google/refine/importers/tree/XmlImportUtilities.java
```diff
@@ -356,23 +357,29 @@ static protected void findRecord(
...
+                                String currentFieldName = parser.getFieldName();
+                                if (desiredFieldName.equals(currentFieldName)) {
+                                    processFieldAsRecord(project, parser, rootColumnGroup, trimStrings, storeEmptyStrings, guessDataTypes);
+                                }
                             }
                         }
                     }
+                } catch (TreeReaderException e) {
+                    logger.error("Exception from parse", e);
+                    throw new TreeReaderException("Exception from parsing: " + e.getLocalizedMessage());
```
wetneb, line 381: It is not completely clear to me what value there is in catching and rethrowing the exception. Why do we need to log it since it is shown to the user anyway?

WaltonG: @wetneb Sure there is no need to log the error. The value of rethrowing the exception is to present the error message on the UI since it will not be logged to the console

wetneb: then how about not catching the exception at all?

WaltonG: Not catching the exception is what caused this issue that the PR is trying to fix

WaltonG: I think there is a work around in not catching the error and hadling it from the UI via the recived response payload. Let me try it out

WaltonG: Not catching the exception and presenting the error messge from the front end works fine

main/src/com/google/refine/importers/tree/XmlImportUtilities.java
```diff
@@ -580,40 +575,38 @@ static protected void processSubRecord(
...
                 Serializable value = parser.getValue();
                 String colName = parser.getFieldName();
                 if (value instanceof String) {
                     String text = (String) value;
-                    if(trimStrings) {
+                    if (trimStrings) {
                         text = text.trim();
                     }
-                    addCell(project, thisColumnGroup, record, colName, text, 
+                    addCell(project, thisColumnGroup, record, colName, text,
```
wetneb, line 609: All the changes in this file seem to be formatting changes, is that correct? If so I would prefer to leave this file untouched in this PR. If not, all the formatting changes should be removed so that we can see the functional changes more easily.

WaltonG: Am sorting it out

WaltonG: I have reverted the file to its original status

WaltonG: @wetneb  I have not been able to dicard the formatting changes. Closing this PR and submitting a new one

WaltonG:
Github workflow has been messed so am closing this PR

PR #4741, opened 2022-04-12
(I#4625) Manage illegal character when importing from URL

### LINKED ISSUE #4625: Unable to create a project from a URL on Windows if the URL path contains ":" character
On Windows, creating a project from a URL that contains the colon ":" character in the URL path throws an error related to the naming of a temporary directory/file.

### To Reproduce
1. Create a new project from a URL as `protocol://domain/foo/:bar/foo`
2. Click on the Next button
3. An error message is displayed that states "Error uploading data illegal char <:> at index [...]" (see screenshot below)

### Current Results
An error message is displayed that states "Error uploading data illegal char <:> at index [...]". The project creation fails.

### Expected Behavior
The project should be created from a URL that contains a colon character or any other character that is valid in a URL.

### Screenshots
openrefine-win-error

### Versions
 - Browser Version: Chrome 99
 - JRE or JDK Version: JRE 1.8.0_301
 - OpenRefine: 3.5.2 [e3efd4e]

### Additional context
Creating a project from a URL containing a colon character seems to work fine on Linux and macOS.

zkl-ai:
Fixes #4625

- add a method to normalize the file/path name in ImportingUtilities.java


Babi-B:
Hello @zkl-ai. You should remove the curly braces when stating "Fixes #4625"

zkl-ai:
> Hello @zkl-ai. You should remove the curly braces when stating "Fixes #4625" to link your PR to its issue successfully

Thank you!

Babi-B:
😄 have had a similar issue

antoine2711:
> hi, now, i'm confused with the failed test, i have tried kinds of branch, but i still fail the test. The code I add is just about some string normalize, I don't think it can influence someting about the linux, could you give me some advice?

@zkl-ai: 
image

Why does it expects: `a/b-c/dummy-test`
but gets: `a/b:c/dummy:test`?

(Notice the disparity TWICE.)

Regards, Antoine

zkl-ai:
> > hi, now, i'm confused with the failed test, i have tried kinds of branch, but i still fail the test. The code I add is just about some string normalize, I don't think it can influence someting about the linux, could you give me some advice?
> 
> @zkl-ai: image
> 
> Why does it expects: `a/b-c/dummy-test` but gets: `a/b:c/dummy:test`?
> 
> (Notice the disparity TWICE.)
> 
> Regards, Antoine


The `:` is illegal charater in windows, I want to check whethe the file need to be created by the url contains the illegal charater in its filename (`dummy:test`) and path (`b:c`). If it contains some illegal charater, I replace them with the charater `-`.

zkl-ai:
> > > hi, now, i'm confused with the failed test, i have tried kinds of branch, but i still fail the test. The code I add is just about some string normalize, I don't think it can influence someting about the linux, could you give me some advice?
> > 
> > 
> > @zkl-ai: image
> > Why does it expects: `a/b-c/dummy-test` but gets: `a/b:c/dummy:test`?
> > (Notice the disparity TWICE.)
> > Regards, Antoine
> 
> The `:` is illegal charater in windows, I want to check whethe the file need to be created by the url contains the illegal charater in its filename (`dummy:test`) and path (`b:c`). If it contains some illegal charater, I replace them with the charater `-`.

I will check it in linux. Thank you!

zkl-ai (commented):
fix the bug in the test

antoine2711:
> fix the bug in the test

```
Error:  Failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.17.1:validate (default-cli) on project main: 
File '/home/runner/work/OpenRefine/OpenRefine/main/tests/server/src/com/google/refine/importing/ImportingUtilitiesTests.java' 
has not been previously formatted.  Please format file and commit before running validation! -> [Help 1]
```

Still… :-(

Regards, Antoine

main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -438,13 +439,29 @@ static public String getRelativePath(File file, File dir) {
...
         return (location.startsWith(File.separator)) ? location.substring(1) : location;
     }
-    
+
+    private static String normalizeFileName(String fileName){
+        FileSystem currentFileSystem = FileSystem.getCurrent();
+        if (currentFileSystem != FileSystem.WINDOWS) return fileName;
+        String normalizedLocalName = new String();
+        // normalize the file name if the current system is windows
+        String[] path = fileName.split("/");
```
wetneb, line 448: I do not understand why you are splitting by `/` and normalizing each part independently. Since `/` is not a allowed as part of a filename on Windows, why would you want to preserve those characters? Why not replace them by `-` like the other illegal characters? It feels like this method should basically just call `FileSystem::toLegalFileName`. I also wonder if it is worth delegating to `FileSystem::toLegalFilename` regardless of the operating system: surely, other operating systems also have illegal characters for filenames, and this method probably also handles those cases, no?

zkl-ai: The filename pass in is part of url and the url is splited by `/`. To minimize the influence to the create project process, I just normalize the file name and its path in the url  since only windows has such illegal while linux and mac doesn't. The method `FileSystem::toLegalFilename` is just need a file name or diretory name. However, the parameter passed in contains path and file name together. That why I split it.

wetneb: ok! Could it make sense to use existing classes like `Path` or `File` to split by path separators? Intuitively there could be cases where splitting by `/` is not quite right (for instance if we had an absolute path - not sure if that can happen at this stage though).

zkl-ai: OK, I'll try fix it quickly.

zkl-ai (commented):
Change and test the separator

zkl-ai (commented):
Fix test bugs in Linux

zkl-ai (commented):
fix the bug test in Linux

main/tests/server/src/com/google/refine/importing/ImportingUtilitiesTests.java
```diff
@@ -110,6 +112,72 @@ public void testAllocateFileDeduplication() throws IOException {
         Assert.assertEquals(allocated, new File(dirA, "dummy-2"));
     }
 
+    @Test
+    public void testNormalizePath() {
+        // Test for issue Unable to create a project from a URL on Windows if the URL path contains ":" character #4625
+        // https://github.com/OpenRefine/OpenRefine/issues/4625
+        String urlPath = "/a/b:c/dummy:test";
+        String urlPathFixed = "\\a\\b-c\\dummy-test";
```
wetneb, line 120: I am surprised that a forward slash is converted to a backwards slash here - does that actually work? If we originally have a local path, then the normalization function should probably also return a local path, no?

zkl-ai: I have check this in the `File` class. And I found that when I create a new object by a path (in OpenRefine the path is from the method `url.getpath()`). The bug I fixed is actually here, using this path to create a file. In the construct, it will check the current Operation System and create a corresponding file with the path passed in if the path doesn't contain the illegal charater in current system. If not, there will be an crash. That's the issue about. And this work well in linux and mac but not well in windows. To make minimum impact on OpenRefine, my modification only work when current system is Windows, otherwise it will return to the original path. 
- For the first question, I have tested this and it work well. And the convertion only occurs in Windows. To implement the convertion, I use the library `org.apache.commons.io.FilenameUtils`  and the eaxct method ` String pathWithWSeparator = FilenameUtils.separatorsToWindows(path)`.
 - For the second question, I think it doesn't matter because the method `static public File allocateFile(File dir, String name)`  invokes this normalize. And the `allocateFile()` method only return a file object, it doesn't affect the method. Since if normalize was invoked means the current System is Windows, and the path in the file object is  separated by backward slash.  It works. If the current system is not Windows, it won't invoke the normalize method. It also works.

wetneb:
Sorry for the delay in reviewing this! Your contribution was rebased and merged as #5787. Thank you so much!

PR #4752, opened 2022-04-13
(I#2606) Add 2 new grel functions for replacing strings matching a search list

### LINKED ISSUE #2606: Add 2 new GREL functions for replacing strings matching a search list
Currently we do not have a nice single replace() function that takes a List of Strings as an argument to search for in a String and replace any found match with a replacement character.
We have to resort to use of cross() or forEach() with inArray() to perform the lookup/inspection of a List or Array and then perform our String replacement.

Mailing list thread comment: 
https://groups.google.com/d/msg/openrefine/XIfKTL6sv0w/BjUedqn0AgAJ

GIVEN:  2 columns where column A has the string we want to replace when a list of substrings are found.
AND: column B has a list (or can be split into an Array) of those substrings to search for
AND: and when matching,
THEN: replace ANY matching substring in our column A with a single replacement value like "" empty string or whatever the user desires.

The following GREL performs this:

```
forEach(
  forEach(
    value.split(","),
    i,
    i.trim()
  ),
  i,
  if(
    inArray(cells.B.value.split("|"),i),
    "",
    i
    )
)
.join(",")
```

however, it is very verbose.

Instead we can simply this large expression to something like this that takes as an argument a list of strings to search for and those that match get replaced with replacement char.

```
value.trim().replaceEach(cells.B.value.split("|"), "")
```

headaches and more coffee and continue with GREL snakes.

**Additional context**
Apache StringUtils seems to have already 2 nice functions to help `replaceEach()` and `replaceEachRepeatedly()` :
http://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#replaceEach-java.lang.String-java.lang.String:A-java.lang.String:A-

... as well as many "index" methods.

Example Problem OpenRefine Project with provided "GREL snake" current solution:
[vidal_santos.openrefine.tar.gz](https://github.com/OpenRefine/OpenRefine/files/4560531/vidal_santos.openrefine.tar.gz)

elroykanye:
Fixes #2606

- a new function `replaceEach` to address the concerns stated in the issue
- tests for this function
- update to the function registry

### DIFF
--- a/docs/docs/manual/grelfunctions.md
+++ b/docs/docs/manual/grelfunctions.md
@@ -141,6 +141,12 @@ You cannot find or replace nulls with this, as null is not a string. You can ins
 
 Returns the string obtained by replacing a character in s, identified by find, with the corresponding character identified in replace. For example, `"Téxt thát was optícálly recógnízéd".replaceChars("áéíóú", "aeiou")` returns the string “Text that was optically recognized”. You cannot use this to replace a single character with more than one character.
 
+###### replaceEach(s, a find, a replace) {#replaceeach-s-a-find-a-replace}
+
+Returns the string obtained by replacing each element in s of a find array with the corresponding element of a replace array. For example, `"The cow jumps over the moon and moos".replaceEachFromArrays(["oo", "moo"], ["ee", "mee"])` returns the string “The cow jumps over the meen and mees”.
+
+The length of the find array must be the same as the length of the replace array. If the length of the find array is greater than the length of the replace array, the last element of the replace array is used for all remaining elements in the find array.
+
 ###### find(s, sub or p) {#finds-sub-or-p}
 
 Outputs an array of all consecutive substrings inside string s that match the substring or [regex](expressions#grel-supported-regex) pattern p. For example, `"abeadsabmoloei".find(/[aeio]+/)` would result in the array [ "a", "ea", "a", "o", "oei" ].
--- a/docs/docs/manual/grelfunctions.md
+++ b/docs/docs/manual/grelfunctions.md
@@ -143,7 +143,7 @@ Returns the string obtained by replacing a character in s, identified by find, w
 
 ###### replaceEach(s, a find, a replace) {#replaceeach-s-a-find-a-replace}
 
-Returns the string obtained by replacing each element in s of a find array with the corresponding element of a replace array. For example, `"The cow jumps over the moon and moos".replaceEachFromArrays(["oo", "moo"], ["ee", "mee"])` returns the string “The cow jumps over the meen and mees”.
+Returns the string obtained by replacing each element in s of a find array with the corresponding element of a replace array, sequentially. For example, `"The cow jumps over the moon and moos".replaceEachFromArrays(["th", "moo"], ["ex", "mee"])` returns the string “The cow jumps over exe meen and mees”.
 
 The length of the find array must be the same as the length of the replace array. If the length of the find array is greater than the length of the replace array, the last element of the replace array is used for all remaining elements in the find array.
 
--- a/main/src/com/google/refine/expr/functions/strings/ReplaceEach.java
+++ b/main/src/com/google/refine/expr/functions/strings/ReplaceEach.java
@@ -1,0 +1,71 @@
+package com.google.refine.expr.functions.strings;
+
+import com.google.refine.expr.EvalError;
+import com.google.refine.grel.ControlFunctionRegistry;
+import com.google.refine.grel.Function;
+
+import java.util.Arrays;
+import java.util.Properties;
+
+public class ReplaceEach implements Function {
+    /**
+     * Replace each occurrence of a substring in a string with another substring.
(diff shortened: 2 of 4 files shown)


wetneb (commented):
Thank you! At a first glance it looks good to me, but since the specification of the function does not seem fully set in stone in the issue, I would leave some time for others to comment on the desired behaviour.

elroykanye:
> Thank you! At a first glance it looks good to me, but since the specification of the function does not seem fully set in stone in the issue, I would leave some time for others to comment on the desired behaviour.

Actually. Thanks.

antoine2711:
> Thank you! At a first glance it looks good to me, but since the specification of the function does not seem fully set in stone in the issue, I would leave some time for others to comment on the desired behaviour.

@wetneb would it be appropriate for @elroykanye to start to write the documentation in `/docs/manual/grelfunctions`? Because otherwise, it's not explicit what is his implementation. At least, with docs, it would be much easier to understand. And even if the implementation changes, not a lot of work would be waisted.

Regards, Antoine

wetneb:
Makes sense!

elroykanye:
Alright, will do that. I guess should I make this a draft then 😁

antoine2711:
> Alright, will do that. I guess should I make this a draft then 😁

@elroykanye: oh no! Keep is as a PR, it IS ready, not a work in progress (WIP). 

But, to get more exposure and better understand your logic, could you write the documentation for the users? That way, it will be easier to understand your work. For me, I would ship it. ;-)

Regards, Antoine

elroykanye:
Okay, noted. Thanks.

docs/docs/manual/grelfunctions.md
```diff
@@ -141,6 +141,12 @@ You cannot find or replace nulls with this, as null is not a string. You can ins
 
 Returns the string obtained by replacing a character in s, identified by find, with the corresponding character identified in replace. For example, `"Téxt thát was optícálly recógnízéd".replaceChars("áéíóú", "aeiou")` returns the string “Text that was optically recognized”. You cannot use this to replace a single character with more than one character.
 
+###### replaceEach(s, a find, a replace) {#replaceeach-s-a-find-a-replace}
+
+Returns the string obtained by replacing each element in s of a find array with the corresponding element of a replace array. For example, `"The cow jumps over the moon and moos".replaceEachFromArrays(["oo", "moo"], ["ee", "mee"])` returns the string “The cow jumps over the meen and mees”.
```
antoine2711, line 146:
```suggestion
Returns the string obtained by replacing each element in s of a find array with the corresponding element of a replace array. For example, `"The cow jumps over the moon and moos".replaceEachFromArrays(["th", "moo"], ["ex", "mee"])` returns the string “exe cow jumps over exe meen and mees”.
```

Your other example is ambigious, as once "oo" are converted to "ee", there is no more "woo", because they are already "wee". It defeats the purpuse of multiple replace... What do you think? Make sense @elroykanye?

EDIT: By the way, is it case-sensitive? It could be in the documentation.

antoine2711 (commented):
Some details.

Regards, Antoine

docs/docs/manual/grelfunctions.md
elroykanye: Oh! That's true 😅
In my mind at the time, this was happening async. Will update as suggested 😁
Thanks

antoine2711: > Oh! That's true 😅 In my mind at the time, this was happening async. Will update as suggested 😁 Thanks

He, he, he... But wait! Maybe you should document if it's async or sequencial (and if so, in what axis...) ;-)
This really never ends...

elroykanye: It doesn't 😁

antoine2711: @elroykanye: sure. But how do you proceed? You search the whole string with the first item, you replace, then you search the replaced string with the second item, and so on? If so, you should say it, because your implemantation could be a single pass algorythm.

elroykanye: Yes, it does so sequentially.
```java
// replace each occurrence of search with corresponding element in newReplace
for(int i = 0; i < search.length; i++) {
    str = str.replace(search[i], newReplace[i]);
}
```

elroykanye: > EDIT: By the way, is it case-sensitive? It could be in the documentation.

Oh it is not. However, it does not have any support for regex. I think I should update that.

antoine2711: > Yes, it does so sequentially.
> (...)

Well, let's keep our user informed for the next years by putting this detail in the Documentation. Fair enough?

wetneb:
I wonder why we are not directly using `StringUtils.replaceEach` here? @thadguidry mentioned in the issue that it exists in Apache Commons (`commons-lang`), which we already depend on, so I would be in favour of reusing that.

wetneb (approved):

PR #4754, opened 2022-04-14
(I#2071) Remove double escaping quotes in TSV export

### LINKED ISSUE #2071: TSV export escapes double quotes (") unnecessarily

TheTrailblazer09:
Fixes #2071

- In the file CSVExporter, I changed the original constructor to be called only if the separator is a comma, otherwise (in the case of a TSV file) the constructor is called with the last argument changed to be CSVWriter.NO_ESCAPE_CHARACTER
-  I tested this change with the example mentioned in the issue


wetneb:
Thank you for this PR! It looks like it fails to build at the moment, and it also includes unrelated changes (see https://github.com/OpenRefine/OpenRefine/pull/4754/files).

TheTrailblazer09:
Hi, I think my second commit passes all checks and solves the issue. Could one of the mentors take a look at it, please?

Thanks,
Vani

main/src/com/google/refine/exporters/CsvExporter.java
```diff
@@ -97,9 +98,19 @@ public void export(Project project, Properties params, Engine engine, final Writ
             (params != null && params.getProperty("printColumnHeader") != null) ?
                 Boolean.parseBoolean(params.getProperty("printColumnHeader")) :
                 true;
-        
-        final CSVWriter csvWriter = 
-            new CSVWriter(writer, separator.charAt(0), CSVWriter.DEFAULT_QUOTE_CHARACTER, lineSeparator);
+
+        final CSVWriter csvWriter;
+        if(separator.charAt(0) == '\t'){
+            csvWriter = new CSVWriter(writer, separator.charAt(0), CSVWriter.DEFAULT_QUOTE_CHARACTER, CSVWriter.NO_ESCAPE_CHARACTER);
```
wetneb, line 104: What happens if one of the cells of the project contains a "\t" character? It would be good to have a test case to demonstrate the behaviour there.

server/IDEs/eclipse/Refine with extensions.launch
```diff
@@ -1,18 +1,18 @@
...
-<stringAttribute key="org.eclipse.jdt.launching.CLASSPATH_PROVIDER" value="org.eclipse.m2e.launchconfig.classpathProvider"/>
-<stringAttribute key="org.eclipse.jdt.launching.MAIN_TYPE" value="com.google.refine.Refine"/>
-<stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="server"/>
-<stringAttribute key="org.eclipse.jdt.launching.SOURCE_PATH_PROVIDER" value="org.eclipse.m2e.launchconfig.sourcepathProvider"/>
-<stringAttribute key="org.eclipse.jdt.launching.VM_ARGUMENTS" value="-Drefine.autoreloading=true -Dbutterfly.modules.path=modules,&quot;${workspace_loc:extensions}&quot;,&quot;${workspace_loc:jython}&quot;,&quot;${workspace_loc:extensions}&quot;,&quot;${workspace_loc:database}&quot;"/>
-<stringAttribute key="org.eclipse.jdt.launching.WORKING_DIRECTORY" value="${workspace_loc:OpenRefine}"/>
+    <listAttribute key="org.eclipse.debug.core.MAPPED_RESOURCE_PATHS">
+        <listEntry value="/server/src/com/google/refine/Refine.java"/>
+    </listAttribute>
+    <listAttribute key="org.eclipse.debug.core.MAPPED_RESOURCE_TYPES">
```
wetneb, line 6: Changes to this file (and to `main/webapp/package-lock.json` are unrelated to this PR, I would keep them out.

wetneb (commented):
Thank you! I think we probably cannot disable escaping entirely in the case of a TSV export, as cells could contain tab characters: see below.

main/src/com/google/refine/exporters/CsvExporter.java
TheTrailblazer09: Hi,
I tried recreating the issue with a "\t"  character in the cell, however, it does not split the cell. I believe \t is taken care of by the Quote character.

With regard to the test case, I think exportTsvWithComma in TsvExporterTests.java has a test for a tab instead of a comma, Copied below is the snippet for ready reference

    @Test
    public void exportTsvWithComma() {
        CreateGrid(3, 3);

        project.rows.get(1).cells.set(1, new Cell("with\t tab", null));
        try {
            SUT.export(project, options, engine, writer);
        } catch (IOException e) {
            Assert.fail();
        }

        Assert.assertEquals(writer.toString(), "column0\tcolumn1\tcolumn2\n" +
                "row0cell0\trow0cell1\trow0cell2\n" +
                "row1cell0\t\"with\t tab\"\trow1cell2\n" +
                "row2cell0\trow2cell1\trow2cell2\n");
    }
   
   
  Please correct me if I have not understood anything, and please guide me on a further course of action for this 
  issue. I will be highly obliged for the same.

server/IDEs/eclipse/Refine with extensions.launch
TheTrailblazer09: Okay I will do that! 

Thank you for the review!

sherrif10: cc @TheTrailblazer09 there is a review to address.

main/src/com/google/refine/exporters/CsvExporter.java
antoine2711: @wetneb.

wetneb: ok thanks, this case seems fine to me - but I am not sure about another one, let me comment there.

main/tests/server/src/com/google/refine/exporters/TsvExporterTests.java
```diff
@@ -171,7 +171,7 @@ public void exportTsvWithQuote() {
 
         Assert.assertEquals(writer.toString(), "column0\tcolumn1\tcolumn2\n" +
                 "row0cell0\trow0cell1\trow0cell2\n" +
-                "row1cell0\t\"line has \"\"quote\"\"\"\trow1cell2\n" +
+                "row1cell0\t\"line has \"quote\"\"\trow1cell2\n" +
```
wetneb, line 174: This new test case seems incorrect to me. If we do not have two double quotes inside the cell value, then the cell value should not be quoted at all. So it should be:
```suggestion
                "row1cell0\tline has \"quote\"\trow1cell2\n" +
```

But I suspect it will be hard to get this behaviour without modifying CSVWriter itself.

TheTrailblazer09: Two double quotes are removed inside the cell value to make sure that double quotes are not escaped. However, the  quote outside cell value are added in the assert statement as the "cells.set" method is preserving the double quotes as depicted in screenshot attached. 


I would also like to add that I have manually also tested these cases in openrefine locally after making the changes and it is working in all these scenarios.
TSV test

wetneb: Sorry I still think the output specified by this test case is wrong (see the original issue).

Here is another test case worth thinking about: exporting a cell which contains a quote character followed by a TAB character. In this case, we do need the quotes around the cell value, and therefore we also need escaping the quote inside the cell value, right?

TheTrailblazer09: I have tried this scenario also. The input text copied on clipboard with various combination in cell such as `,",\t is attached. The local tsv export of that project is also enclosed (.txt is added at the end just to attach it here).
input

[tsv_exportfile.tsv.txt](https://github.com/OpenRefine/OpenRefine/files/8541809/tsv_exportfile.tsv.txt)

If you still feel this is not right, I will try implementing CSV Writer but I strongly believe it meet the requirements as mentioned in the issue
Thanks & Warm Regards

wetneb: I think what is really important to keep in mind is how a program can parse the exported file back into a table. For instance, OpenRefine itself should normally be able to parse a file it exported itself back as a table (and ideally other programs, such as LibreOffice or Excel).

If I open your exported TSV file above in OpenRefine, I cannot find a set of options which would give me the original table. If I disable escaping, I get this:
image

Did I miss anything here?

TheTrailblazer09: Thanks for looking at it. Here, the spitted column has a tab character which i have manually inserted just to check various combination which is spitted into two columns by the **CSVImporter.** Tab character here is the special case case while importing however  there is no other issue in any other character.

Warm Regards

wetneb: The exported file should be parsable regardless of the characters contained in the cells.

antoine2711: @TheTrailblazer09: I think this is a situation where you gain a lot from writing your test case before coding. Try to find a few bizarre cases. I would be interested to see if a cell that is already in double quotes, would have them escaped before having normals double quotes being added.

Regards, Antoine

TheTrailblazer09: Hi,
I have added a test case where the cell is enclosed by the quote character and the CSVWriter seems to pass the test. Below is the output produced in both scenarios for ready reference:-

Input:  "cell enclosed by (") quote character"

Output produced without escaping:    """cell enclosed by ("") quote character"""

Output produced with escaping:  ""cell enclosed by (") quote character""

Regards,
Vani

main/tests/server/src/com/google/refine/exporters/TsvExporterTests.java
```diff
@@ -171,10 +171,30 @@ public void exportTsvWithQuote() {
...
+        } catch (IOException e) {
+            Assert.fail();
+        }
+
+        Assert.assertEquals(writer.toString(), "column0\tcolumn1\tcolumn2\n" +
+                "row0cell0\trow0cell1\trow0cell2\n" +
+                "row1cell0\t\"\"cell enclosed by (\") quote character\"\"\trow1cell2\n" +
+                "row2cell0\trow2cell1\trow2cell2\n");
+    }
+
```
sherrif10, line 195: Remove atleast one space to make them code more cleaner.

main/tests/server/src/com/google/refine/exporters/TsvExporterTests.java
antoine2711: > Input: "cell enclosed by (") quote character"
> 
> Regards, Vani

Yes, that's one scenario @TheTrailblazer09. I'm talking about a lots of scenarii…

1. "cell enclosed by (") double quote character"
2. cell containing (") double quote character
3. "cell starting with a double quote character
4. cell ending with a double quote character"
5. cell containing (") double quote and (`\t`) tab character
6. `\t`cell containing (") double quote and starting with (`\t`) tab character
7.  `\t`cell enclosed by (`\t`) tab character`\t`
8. …
9. cell containing (') single quote character
10. …
11. cell containing (`) single quote character
12. 

It's really to save you time. There are so many time this is used, so it used with so much different data, that more test seams to me a saving net. @wetneb, how would you feel about that?

Regards, Antoine

TheTrailblazer09: @antoine2711  Added 25 additional test cases. Kindly review them.

Thanks & Warm Regards

> > Input: "cell enclosed by (") quote character"
> > Regards, Vani
> 
> Yes, that's one scenario @TheTrailblazer09. I'm talking about a lots of scenarii…
> 
> 1. "cell enclosed by (") double quote character"
> 2. cell containing (") double quote character
> 3. "cell starting with a double quote character
> 4. cell ending with a double quote character"
> 5. cell containing (") double quote and (`\t`) tab character
> 6. `\t`cell containing (") double quote and starting with (`\t`) tab character
> 7. `\t`cell enclosed by (`\t`) tab character`\t`
> 8. …
> 9. cell containing (') single quote character
> 10. …
> 11. cell containing (`) single quote character
> 
> It's really to save you time. There are so many time this function is used by the users and in so mamny contexes, and it used with so much different data, that more test seams to me like a saving net. @wetneb, how would you feel about that?
> 
> Regards, Antoine

@antoine2711  Added 25 additional test cases. Kindly review them.

Thanks & Warm Regards
Vani

antoine2711: @wetneb: just to make sure... Here, did I suggest well in asking more tests, and is this too much, or correct? Maybe the caret tests are too much...

Also, is it possible/better to regroup them instead of having a single test for each 20+?

Regards, Antoine

sherrif10: @antoine2711  @wetneb  tests are vital .

> did I suggest well in asking more tests, and is this too much, or correct?

main/tests/server/src/com/google/refine/exporters/TsvExporterTests.java
```diff
@@ -193,6 +193,468 @@ public void exportTsvWithEmptyCells() {
...
+        try {
+            SUT.export(project, options, engine, writer);
+        } catch (IOException e) {
+            Assert.fail();
+        }
+
+        Assert.assertEquals(writer.toString(), "column0\tcolumn1\tcolumn2\n" +
+                "row0cell0\trow0cell1\trow0cell2\n" +
+                "row1cell0\t\"cell containing (\") double quote and (\t) tab character\"\trow1cell2\n"
```
wetneb, line 282: This test case seems to depart from the desired behaviour stated here:
https://github.com/OpenRefine/OpenRefine/issues/2071#issuecomment-515681043

Double quotes which appear in cells enclosed by double quotes should be escaped.

TheTrailblazer09: > This test case seems to depart from the desired behaviour stated here: [#2071 (comment)](https://github.com/OpenRefine/OpenRefine/issues/2071#issuecomment-515681043)
> 
> Double quotes which appear in cells enclosed by double quotes should be escaped.

This test does not seems to depart from the original issue to me as it is mentioning that without escaping there is a double/duplicate occurrence of double quote character. However, in this case there is no duplicate occurrence of double quote character.

 Below is the output produced from openrefine in both scenarios for ready reference:-

Input: "cell enclosed by (") quote character"

Output produced without escaping: """cell enclosed by ("") quote character"""

Output produced with escaping: ""cell enclosed by (") quote character""

Regards,
Vani

main/tests/server/src/com/google/refine/exporters/TsvExporterTests.java
TheTrailblazer09: > @wetneb: just to make sure... Here, did I suggest well in asking more tests, and is this too much, or correct? Maybe the caret tests are too much...
> 
> Also, is it possible/better to regroup them instead of having a single test for each 20+?
> 
> Regards, Antoine

Hi
@Antoine. In CI/CD some tests have failed which shows some I/O error. However, all these tests have passed locally (./refine test). could you please guide me on this if anything is needed from my side.

Thanks & Warm Regards
Vani

main/tests/server/src/com/google/refine/exporters/TsvExporterTests.java
TheTrailblazer09: Hi @antoine2711 @wetneb 

Do I need to work any further on this issue.

Thanks & Warm Regards

wetneb:
It looks like there is disagreement about what the behaviour of this quoting should be. Without a clear agreement on this I think it is a rather risky thing to change. I would welcome more opinions here, but if that does not happen I would just close this per inactivity.

PR #4757, opened 2022-04-14
Option to remove Blank Columns upon project creation

### LINKED ISSUE #1750: Remove Blank Columns upon project creation/import

wsmmxmm:
Fixes #1750 
-  Determine whether a column is blank: (based on https://github.com/OpenRefine/OpenRefine/pull/3497): Create a new list<Boolean> in the TabularImportingParserBase class, that represents whether each column has data.
-  If the store blank column is false, delete this column (only happens when store blank column == false): removeColumn method is created in the ColumnModel class. This will use columns.remove(all empty columns), which will remove both data lines and header lines. 
- The checkbox of the store blank column added to the front-end page. After testing with real files (csv, xls, tsv), this function works. I think it should work on other Tabular files as well, since I'm optimizing directly in the TabularImportingParserBase class. But I'm not really sure how to test in TabularImportingParserBase directly. In addition, I've tested tree files (xml, json), and now it is possible to do blank columns. So I didn't make changes.

### DIFF
--- a/main/tests/cypress/cypress/e2e/create-project/preview_project.cy.js
+++ b/main/tests/cypress/cypress/e2e/create-project/preview_project.cy.js
@@ -197,4 +197,27 @@ describe(__filename, function () {
     cy.get('table.data-table tr').eq(1).should('to.contain', '01002');
     cy.get('input[bind="disableAutoPreviewCheckbox"]').uncheck();
   });
+
+  it('Tests save blank columns of parsing options', function () {
+    cy.visitOpenRefine();
+    cy.createProjectThroughUserInterface('food-blank-column.mini.csv');
+    cy.get('.create-project-ui-panel').contains('Configure parsing options');
+    // FIXME: data-table in preview has no thead, so first tr in tbody is a header
+    cy.get('table.data-table > tbody > tr:nth-child(2) > td:nth-child(1)').should('to.contain', '1.');
+    cy.get('table.data-table > tbody > tr:nth-child(2) > td:nth-child(2)').should('to.contain', '01001');
+    cy.get('table.data-table > tbody > tr:nth-child(2) > td:nth-child(3)').should('to.contain', 'BUTTER,WITH SALT');
+    cy.get('table.data-table > tbody > tr:nth-child(2) > td:nth-child(4)').should('to.contain', '15.87');
+    // empty cells are filled with NBSP when rendered
+    cy.get('table.data-table > tbody > tr:nth-child(2) > td:nth-child(5)').should('to.contain', '\u00a0');
+    cy.get('table.data-table > tbody > tr:nth-child(2) > td:nth-child(6)').should('to.contain', '717');
+
+    cy.get('input[bind="storeBlankColumnsCheckbox"]').uncheck();
+
+    cy.get('table.data-table > tbody > tr:nth-child(2) > td:nth-child(1)').should('to.contain', '1.');
+    cy.get('table.data-table > tbody > tr:nth-child(2) > td:nth-child(2)').should('to.contain', '01001');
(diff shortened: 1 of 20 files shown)


main/src/com/google/refine/importers/ImporterUtilities.java
```diff
@@ -249,4 +249,50 @@ protected long track(long bytesRead) {
...
+//                }
+            }
+        }
+//        if (deletedSoFar != 0){//set col[row.size()-1] to col[row.size()-1-deletedSoFa]  =null;
+//            for (int i = 0; i < project.rows.size(); i++) {
+//                Row row = project.rows.get(i);
+//                for (int j = 0; j < deletedSoFar ; j++) {
+//                    row.setCell(row.cells.size()-1-j, null);
+//                }
+//            }
```
wetneb, line 295: Is this commented out code meant to be submitted in this PR? If it is not needed, I would remove it entirely.

wetneb (commented):
Thank you, it looks pretty good to me! It would be good to have unit tests to demonstrate the feature, for instance you could test the feature on a few tabular importers (CSV, Excel…)

main/src/com/google/refine/importers/ImporterUtilities.java
wsmmxmm: > Is this commented out code meant to be submitted in this PR? If it is not needed, I would remove it entirely.

I have removed the extra code in the new commits(?). I think it is more reasonable to remove the automatically added column names when deleting blank columns.

wsmmxmm:
> Thank you, it looks pretty good to me! It would be good to have unit tests to demonstrate the feature, for instance you could test the feature on a few tabular importers (CSV, Excel…)

And I add some unit tests in the test part.
It's weird that the feature is worked in the website on excel importer. 
image
image


But it has different results in unit test.
I found the reason for this was that the results of the unit test processing on the blank cell were not consistent with expectations. But I decided not to look further for a solution after comparing it with csv importor test, perhaps because of historical reasons.

image

wsmmxmm:
I found the failed email just now. This is my first pull request. Is there anything else I should do?

WaltonG:
@wsmmxmm The reason for failling intergation is explained when you click on detaills. In your case its formatting issues.

From the developers manual
>If you made Java changes, run linting to make sure they conform to our code style, with mvn formatter:format.

WaltonG:
@wsmmxmm The integration error still persists. 
>/OpenRefine/main/tests/server/src/com/google/refine/importers/ImporterUtilitiesTests.java' has not been previously formatted.  Please format file and commit before running validation! ->

tfmorris:
I've rebased this on the current HEAD and done a cursory functional test, but haven't reviewed the code or done any detailed testing.

tfmorris:
@wetneb Since you did the original review, could you review again please when you get a chance?

tfmorris (commented):
@wsmmxmm I apologize on behalf of the team for the delay in bringing this to closure. Since it's been such a long time, I imagine you've moved on, but I've provided feedback in case it's useful to you. No action is needed on your part though, because I've pushed the suggested changes.

Thank you VERY much for your contribution and, again, apologies for the delay.

tfmorris:
@wetneb This needs an end-to-end test, so you might want to defer your review until I have a change to add that and make this PR non-draft.

tfmorris:
@wetneb I've added a basic e2e test, so this should be ready for review.

tfmorris:
@wetneb ping

wetneb (approved):
Thanks to everyone involved! Nice to get old PRs merged!

PR #4758, opened 2022-04-14
(I#1286) Implement option to add a "display unprintable chars" toggle mode

### LINKED ISSUE #1286: Add a DISPLAY UNPRINTABLE CHARS toggle mode

shriyasankhyan:
Fixes #1286

-  It shows unprintable characters in the parsing panel as well as in the final data table.


thadguidry:
@shriyasankhyan One thing I am noticing immediately while reviewing...
It seems that you chose to retain the unprintable char, rather than doing a replacement of it during the toggle?
Notice on the left side (original) and the right side (your code changes) and where any particular "" character is not replaced but instead suffixed by the various `SOH`, `STX`, etc.

image

shriyasankhyan:
> @shriyasankhyan One thing I am noticing immediately while reviewing... It seems that you chose to retain the unprintable char, rather than doing a replacement of it during the toggle? Notice on the left side (original) and the right side (your code changes) and where any particular "�" character is not replaced but instead suffixed by the various `SOH`, `STX`, etc.
> 
> image

Yes I shall change it as soon as possible and submit the changes. Thanks for reviewing it :).

thadguidry:
Another issue I see is that of unnecessary repeated elements if I change the number of rows to show (going from default 10 to 50 or 1000 and back down to 10 or 5) all while keeping the toggle on "Show Control Chars".
image

shriyasankhyan:
> Another issue I see is that of unnecessary repeated elements if I change the number of rows to show (going from default 10 to 50 or 1000 and back down to 10 or 5) all while keeping the toggle on "Show Control Characters". Hmm, also that label is uppercase above the datagrid and in the preview it is lowercase. Let's just use titlecase "Show control characters". image

Thankyou again I shall make the changes and revert as soon as possible.

wetneb:
I have not found the time to go thouroughly into this but I would not place the toggle in this position: this is a very valuable space that we need for extensions and other things that people will want to access very often. Instead, I would imagine a "Display options" dialog where we could eventually configure other options (such as a maximum cell length beyond which contents would be hidden by an ellipsis, maybe word wrapping options, another checkbox to control the display of leading/trailing whitespace, and so on).

thadguidry:
@wetneb Do you think a small gear icon after HELP button would be ok as the "display options" dialog entry point?

wetneb:
I think it would deserve its own menu entry with a proper label, actually! Not sure in which menu, perhaps the "All" one…

thadguidry:
@wetneb All -> View seems to be what @ostephens had suggested before, yeah, that works.

shriyasankhyan:
> @wetneb All -> View seems to be what @ostephens had suggested before, yeah, that works.

Can you please tell me whether parsing panel checkbox is going to be at the same place or is it to be changed to some other location as well??

thadguidry:
@shriyasankhyan so the bottom is where parsing options are (so lets avoid that area, as we'd only be able to maybe squeeze it under the phrase "Parse data as").  So that leaves the top portion 'wizardHeader` as a common area for Create Project workflow, as well as the left side area, which is not common for the Create Project workflow other than it's `action-area-tab`.  I think leaving it inside that `wizardHeader` where you have it is likely the best place, which really flows well into the next datagrid area header where you have it beforre, but alas @wetneb and @ostephens disagree with me here.

Let @wetneb make the final decision on this.

shriyasankhyan:
I did reproduce the changes somehow but there is a problem I guess.
OpenRefine – data-table-view js  main  17-04-2022 19_08_36
OpenRefine – data-table-view js  main  17-04-2022 19_08_49

I can not get how to show control characters by labelling label: $.i18n('core-views/display-controlChars'),

shriyasankhyan:
> @shriyasankhyan so the bottom is where parsing options are (so lets avoid that area, as we'd only be able to maybe squeeze it under the phrase "Parse data as"). So that leaves the top portion `wizardHeader` as a common area for Create Project workflow, as well as the left side area, which is not common for the Create Project workflow other than it's `action-area-tab`. I think leaving it inside that `wizardHeader` where you have it is likely the best place, which really flows well into the next datagrid area header where you have it before, but alas @wetneb and @ostephens disagree with me here.
> 
> Let @wetneb make the final decision on this.

If parsing panel checkbox is to be kept at the same place and only final data table checkbox had to be removed and placed under All -> View-> ,then I have done the changes. Please review them once again. 
Thank you

wetneb:
The parsing panel is shown before project creation, and cannot be accessed after project creation, so I think it makes sense to distinguish between both stages.

* Before project creation, when previewing the project to be created, I would say that it is worth displaying unprintable characters unconditionally. Users will not be working with the data yet, so I do not see how that would annoy them. As a user I would find it useful to know ahead of time that those characters are there: I would likely clean them up immediately after project creation. So I would say that we should not need a toggle at this stage to control whether unprintable characters are displayed.

* After project creation, I think it makes sense to offer this toggle. As I wrote above it could go in the "All columns" menu, in a new dialog which could control other display options. I think this option could be enabled by default.

Let me know if anything is unclear or if you disagree with those design opinions :)

shriyasankhyan:
@wetneb I have placed it under the All -> View -> Show/ Hide 'null' values in cells in the table, so whether I need to place it there or Under All -> directly.

wetneb:
I would just replace the current "Show/Hide 'null' values in cells" by a menu item "Settings" or so, and that would open a dialog which would let you:
* set whether you want to show or hide 'null' values
* set whether you want to display special characters

shriyasankhyan:
> I would just replace the current "Show/Hide 'null' values in cells" by a menu item "Settings" or so, and that would open a dialog which would let you:
> 
> * set whether you want to show or hide 'null' values
> * set whether you want to display special characters

Ok I shall work on the same as soon as possible :)

shriyasankhyan:
When I am reverting to the original file it is showing errors. Can I just close this pull request and redo them in some other branch and then open a pull request again?

shriyasankhyan:
I am closing this one and whatever new changes are there to be done, I shall do it in new PR as the code has messed up a lot and I want to do it from the start. Even after reverting, it is showing errors.

https://github.com/OpenRefine/OpenRefine/pull/4798

PR #4764, opened 2022-04-15
(I#4339) Upgrade database and gdata extensions to httpclient5 

### LINKED ISSUE #4339: Upgrade database and gdata extensions to httpclient5
Our database extension (`extensions/database/pom.xml`) depends on `org.apache.httpcomponents:httpclient` but our `main` module (`main/pom.xml`) depends on `org.apache.httpcomponents:httpclient5`. We should investigate if the former can be migrated to `httpclient5` so that we avoid duplicating those dependencies.

This is also the case for the `gdata` extension. We should investigate if this is the case for other modules.

WaltonG:
Fixes #4339

- Upgrade database and gdata extensions dependency to httpclient5

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,7 +86,6 @@
     <odfdom-java.version>0.9.0</odfdom-java.version> <!-- do not update to 0.10.0, see issue #4397 -->
     <vicino.version>1.2</vicino.version>
     <clojure.version>1.11.1</clojure.version>
-    <httpclient.version>4.5.13</httpclient.version>
     <httpclient5.version>5.1.3</httpclient5.version>
     <httpcore.version>4.4.15</httpcore.version>
     <opencsv-multichar.version>2.4</opencsv-multichar.version>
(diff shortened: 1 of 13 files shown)


wetneb:
I am surprised it is that easy! Have you tested those two extensions after this change? They are basically not covered by our integration tests, so it is worth giving it a good go.

pom.xml
```diff
@@ -86,6 +86,7 @@
     <odfdom-java.version>0.9.0</odfdom-java.version> <!-- do not update to 0.10.0, see issue #4397 -->
     <vicino.version>1.2</vicino.version>
     <clojure.version>1.11.1</clojure.version>
+    <httpclient.version>4.5.13</httpclient.version>
```
wetneb, line 89: Ah, this likely indicates that we also pull the version 4 in other modules, which is perhaps why the extensions did not require any adaptation: they probably can still use the version 4 through indirect dependencies.
It is worth using `mvn dependency:analyze` to make sure we only depend on one version of this library throughout the application.

WaltonG: Sure, am looking into it

PR #4765, opened 2022-04-15
(I#4304) Provide a link to an overview of existing/active reconciliation services in the reconciliation dialog window 

### LINKED ISSUE #4304: Provide a link to an overview of existing/active reconciliation services in the reconciliation dialog window
Especially for users who are new to the variety of reconciliation services that are out there, it's not obvious to discover / find a list of them. Knowing which ones exist and can be used can be quite inspiring and helpful.

### Proposed solution

In a recent [user interview](https://phabricator.wikimedia.org/T294939), one OpenRefine user mentioned that they would find it very helpful if OpenRefine's user interface itself - possibly the reconciliation dialog window - would contain (a pointer to) a list of existing / active reconciliation services.

Perhaps, at first, this could be a simple link to https://reconciliation-api.github.io/testbench/

### Alternatives considered

OpenRefine's [manual also collects various pointers](https://docs.openrefine.org/manual/reconciling) to places where one can find lists of reconciliation services. But in order to get there, the user first needs to be aware of the manual itself, find this specific page, and then click through to the various options.

### Additional context

This suggestion came up during a [user interview for Wikimedia Commons batch upload functionalities](https://phabricator.wikimedia.org/T294939).

WaltonG:
Fixes #4304

- Added a button which when clicked opens a link to existing reconciliation services


antoine2711:
There is something weird about the size of the string, and it placement. The problem was a bit there with the previous heading, but now, with the second string, bigger, kinda placed to the left... 
image

I personnaly think that the link should go below the existing title, and be smaller.

Regards, Antoine

WaltonG:
@antoine2711 Thanks for the suggestion, my look and feel skill need ssome polishing.

>I personnaly think that the link should go below the existing title, and be smaller

Should the link go below the reconcile dialog title or below the Service title

antoine2711:
> @antoine2711 Thanks for the suggestion, my look and feel skill needs some polishing.
Thruth be said, it not something developer are usually good at. It's another skillset!

> Should the link go below the reconcile dialog title or below the Service title

image

How about that? Maybe someone in the community has other ideas.

Regards, Antoine

WaltonG:
>How about that? Maybe someone in the community has other ideas.

I'd go with that since it clearly brings out the context of the link.

wetneb:
Overall I would like to redesign this dialog from scratch, so I am not so comfortable making small changes like this, but I do not want to block this PR with this concern. Also, the testbench is not designed for end users but rather for reconciliation service developers, so I wish we had a better link to point the users to (again, it should not be a blocker though).

Other UI suggestion: another button, next to the one to add another reconciliation service.

WaltonG:
Well noted. I'll proceed to work on the suggestions

antoine2711 (changes requested):
Hi @WaltonG, we've got a problem here:
image

Also, could you put more space between the buttons? The new one you created, but also the other OK and Cancel. Definitely too close.

> Also, the testbench is not designed for end users but rather for reconciliation service developers, so I wish we had a better link to point the users to (again, it should not be a blocker though).

Is there a Wiki page on Recon Services, or a page with a list of services somewhere? Do we have one in our doc or in our Wiki?

Regards, Antoine

wetneb:
Ideally the spacing between the buttons should be standard across all dialogs and not adjusted in an ad-hoc manner: @elebitzero has been working on normalizing button placement recently.

> Is there a Wiki page on Recon Services, or a page with a list of services somewhere? Do we have one in our doc or in our Wiki?

Yes, but it is not as up to date: https://github.com/OpenRefine/OpenRefine/wiki/Reconcilable-Data-Sources

antoine2711:
> Yes, but it is not as up to date: https://github.com/OpenRefine/OpenRefine/wiki/Reconcilable-Data-Sources

@WaltonG: I see a nice opportunity to make this PR and issue consequent.
Can you update a bit our Wiki with the data from the techical link? Like make sure each is mentionned, and maybe a short description?

What would be great would be a little table at the top, with 2 or 3 columns, or maybe not. Just a list of the services, with a link to the section below would be a great improvement.

Regards, Antoine

wetneb:
I have mixed feelings about pointing the user to our GitHub wiki to be honest, it is not very welcoming and I would not want to encourage users to register new reconciliation services there instead of the testbench :-/

antoine2711:
> I have mixed feelings about pointing the user to our GitHub wiki to be honest, it is not very welcoming and I would not want to encourage users to register new reconciliation services there instead of the testbench :-/

@wetneb: a page in the documentation would be a logical choice, no?

Regards, Antoine

wetneb:
I would say we need something a bit more dynamic and easier to edit than the docs, perhaps.
In an ideal world, we would just have a website similar to the testbench, but more user-oriented (like a marketplace), which could potentially use the same data source (Wikidata) as the testbench, so that you only add a service once to both.
Perhaps we can indeed use the testbench for now? Sorry for the trouble.

antoine2711:
> In an ideal world, we would just have a website similar to the testbench, but more user-oriented (like a marketplace), which could potentially use the same data source (Wikidata) as the testbench, so that you only add a service once to both.

@wetneb: how about a Wikipedia page that contains a list build with some ListeriaBot magic? And that page would be in many languages, so the link shown could be adjusted with the language of the user... Could that be the solution?

Regards, Antoine

wetneb:
I am not sure Wikipedia is the right place to host such a catalogue.

antoine2711:
@WaltonG: so, we really have no better option then the link to the Recon Status page. Better than nothing! You can move forward.

Regards, Antoine

wetneb (approved):

PR #4766, opened 2022-04-16
Improved the UI for the select Wikibase instance dialog.

### LINKED ISSUE #3824: UI tweaks in Wikibase selection dialog
I would like to propose some minor tweaks to the dialog to select the Wikibase instance:
* The currently selected Wikibase instance is not displayed prominently enough. I would expect the item in the list to be somehow highlighted (different background color, border…)
* Since the logos take a bit of vertical space I think we could have space to also display the URL of the Wikibase instance underneath the name. I expect it could be useful to distinguish between various instances with similar names/logos.
* When I click on another Wikibase instance to select it, I find it confusing that the dialog does not go away by itself. What else could I want to do in this dialog after that? Its only purpose is to select the Wikibase instance, so once that is done I should be able to just go back to the project and carry on.

Pinging @afkbrb for his thoughts on this :)

elebitzero:
Fixes #3824

* Currently selected Wikibase instance is shown at the top of the list and shown selected.
* Remaining Wikibase instances are shown in alphabetical order.
* Selecting/Adding/Deleting a Wikibase instance updates the list.
* Wikibase URL (mediawiki.root) is shown under the name for each Wikibase.
* Delete icon changed to jquery ui-trash-icon

image

### DIFF
--- a/extensions/wikidata/module/scripts/dialogs/wikibase-dialog.html
+++ b/extensions/wikidata/module/scripts/dialogs/wikibase-dialog.html
@@ -4,9 +4,9 @@
     <p bind="explainSelectWikibase"></p>
     <p bind="currentSelectedWikibase"></p>
     <div class="wikibase-list-wrapper">
-      <table class="wikibase-dialog-wikibase-list" bind="wikibaseList">
+      <ol class="wikibase-list" bind="wikibaseList">
         <!-- dynamically generated -->
-      </table>
+      </ol>
     </div>
   </div>
   <div class="dialog-footer" bind="dialogFooter">
--- a/extensions/wikidata/module/scripts/dialogs/wikibase-dialog.js
+++ b/extensions/wikidata/module/scripts/dialogs/wikibase-dialog.js
@@ -24,23 +24,58 @@ WikibaseDialog.launch = function () {
 };
 
 WikibaseDialog.populateDialog = function () {
-  let wikibases = WikibaseManager.getAllWikibases();
+  let wikibases = WikibaseManager.getAllWikibaseManifests();
+  let selectedWikibase =  WikibaseManager.getSelectedWikibaseName().toLowerCase();
+
+  wikibases.sort((a, b) => {
+    let ret;
+    let aName = a.mediawiki.name;
+    let bName = b.mediawiki.name;
+    let aSelected = aName.toLowerCase() === selectedWikibase;
+    let bSelected = bName.toLowerCase() === selectedWikibase;
+    // ascending order
+    // compare selected, then by name
+    if (aSelected) {
+      ret = -1;
+    } else if (bSelected) {
+      ret = 1;
+    } else {
+      if (aName < bName) {
+        ret = -1;
+      } else if (aName > bName) {
+        ret = 1;
+      } else {
+        ret = 0;
+      }
+    }
+    return ret;
+  });
 
   WikibaseDialog.elmts.wikibaseList.empty();
-  for (let wikibaseName in wikibases) {
+  for (let manifest of wikibases) {
+    let wikibaseName = manifest.mediawiki.name;
+
+    let rootURL = manifest.mediawiki.root;
+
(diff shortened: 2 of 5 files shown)


elebitzero:
We also might not need the "Current selected Wikibase:" line since we show the selected Wikibase. If we still want the links to be clickable, then the URLs can be clickable in the list or have an icon to go to the link.

wetneb:
I simply love it! Not merging immediately because of the discussion in #3824

lozanaross:
Hi all -- check out the images I posted in #3824 -- On the whole this is quite close to what we're planning with the SDC extension but perhaps making the dialogs in both WB / SDC extensions identical would make more sense.

My observations on the above: the current "Select" dialog has a problem in that it conflates 2 separate functions into 1 dialog - adding & selecting an instance. The current design I'm proposing for the SDC extension tries to decouple the two functions more clearly through several design decisions:
1) Change in top copy
2) Intro of a radio button
3) highlight background color (this happens in the design proposal above, too, so that's great)
4) Moving the "Add new" button to the bottom left corner; Introducing a "Confirm selection button"

Final note: while the trash icons are great and very clear visual signs, I wonder if we should start introducing icons in such a fragmented way - I think for the most part the current OpenRefine framework does not make much use of icons? Introducing icons should be done in a holistic way across the entire UI framework, so it doesn't start looking patchy. As this holistic framework "re-thinking" is on the list of things to be resolved within the next months, I suggest we pause introducing new visual language into the UI until we have set the long-term strategy.

elebitzero:
Hey @lozanaross,

I like the idea of separating selection from management of the Wikibase instances.

This PR is focused on fixing the immediate issues @wetneb has raised in Issue#3824 with the current design.  I think the bigger redesign should be filed as a separate enhancement issue.  

If someone is working on the SDC extension, when that is implemented, they could make the same changes to the WB extension so they are consistent. I'm not familiar with the SDC extension, so correct me if that doesn't make sense.

elebitzero:
New changes:

* Removed 'Current selected Wikibase:' since it is no longer needed.
* Moved Add Wikibase button to the left.


Updated screenshot above to reflect latest changes.

elebitzero:
As for the trash icon, I agree it is better to minimize the use of icons. Since they were already using an icon, I replaced it with the jquery trash icon which I believe is more appropriate.  Since it is a jquery-ui icon, I think that is consistent. But if people have an issue with the change, I can revert that part.  Or I can change it to use text e.g. 'Delete'. Example of what that would look like:

wetneb:
Lozana is making a good point: let us go for your textual variant ("Delete") instead of the bin button (it does look intuitively different from the rest of our style, perhaps too bold and rounded?)

I tested this on Firefox (78.15.0esr, Linux) and the logos do not show up anymore:

elebitzero:
@wetneb, I have not been able to reproduce the missing logo problem you mentioned.

It is working fine for me on Chrome/Brave/Firefox/Edge on Windows 10 and Firefox 99.0.1 (64-bit) on Ubuntu 21.10 Linux.

Can you tell me if you are getting any errors on the network tab or the console of the developer tools panel?

elebitzero:
@wetneb, I found that there was a very tight timeout set for getting the logo image, so I bumped it up to five seconds and also made it so the URL is cached on success. Now you can see if there are any errors in your console.

wetneb:
Thanks! Sadly it does not solve the issue on my side:
image
I also tried clearing the cache: the requests to fetch the images are then HTTP 200 but the images still do not show up.
The console does not seem to indicate any problem either:
image

elebitzero:
@wetneb, I'll try downoading the 78.15.0esr version of Firefox on Linux and test with that.

elebitzero:
@wetneb, I was able to reproduce the problem on 78.15esr for Windows, and I have checked in a fix to workaround what looks like a Firefox bug in that version. Please check if the issue is resolved for you. Thanks.

PR #4768, opened 2022-04-17
(I#4402) update project metadata file reader

### LINKED ISSUE #4402: Problems with special characters in project name
This issue describes problems with special characters in the project name and other fields from `metadata.json`.

### To Reproduce
1. Create a project with special characters in the project name ("Testing äüöß").
2. Close OpenRefine and check `metadata.json` (everything ok!)
3. Reopen OpenRefine and check special characters in the project name ("Testing Ã¤Ã¼Ã¶ÃŸ").
4. Close OpenRefine and check `metadata.json` (not ok!)
5. Reopen OpenRefine and check special characters in the project name ("Testing ÃƒÂ¤ÃƒÂ¼ÃƒÂ¶ÃƒÅ¸")

### Current Results

The special characters are somehow completely wrong and their number doubles with each restart of OpenRefine.

### Expected Behavior

Reading and writing metadata to `metadata.json` preserves special characters.

#### After creating the project (step 1):

openrefine_special_chars_1

#### First restart (step 3):

openrefine_special_chars_2

##### Second restart (step 5):

openrefine_special_chars_3

### Versions
 - Browser Version: Firefox 95
 - JRE: 11.0.13 (using the bundled version "openrefine-win-with-java-3.5.1.zip")
 - OpenRefine: observed in 3.5.0 and 3.5.1

### Additional context

1. The list of garbled up characters grows exponentially which results in realy large `metadata.json` files and OutOfMemory errors.
2. This might be related or even be the same problem as described in #3431.
3. I also observed this problem in `metadata.json` in the fields `.importOptionMetadata` and `.preferences.entries."exporters.templating.template"`.

elroykanye:
Fixes #4402 

- Modified the reader for loading project metadata to FileInputStream
- Set the charset to UTF-8 in the reader

### DIFF
--- a/main/src/com/google/refine/io/ProjectMetadataUtilities.java
+++ b/main/src/com/google/refine/io/ProjectMetadataUtilities.java
@@ -34,15 +34,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package com.google.refine.io;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.FileReader;
 import java.io.IOException;
-import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
-import java.io.Reader;
 import java.io.Writer;
-import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.time.LocalDateTime;
@@ -62,7 +58,7 @@ import com.google.refine.util.ParsingUtilities;
 public class ProjectMetadataUtilities {
     final static Logger logger = LoggerFactory.getLogger("project_metadata_utilities");
 
-    public static void save(ProjectMetadata projectMeta, File projectDir) throws IOException {
+    public static void save(ProjectMetadata projectMeta, File projectDir) throws IOException  {
         File tempFile = new File(projectDir, "metadata.temp.json");
         saveToFile(projectMeta, tempFile);
         if (tempFile.length() == 0) {
@@ -118,7 +114,7 @@ public class ProjectMetadataUtilities {
         File file = new File(projectDir, fileName);
         if (file.exists()) {
             try {
-               pm = loadFromFile(file, StandardCharsets.UTF_8);
+               pm = loadFromFile(file);
             } catch (Exception e) {
                 logger.warn("load metadata failed: " + file.getAbsolutePath());
                 logger.error(ExceptionUtils.getStackTrace(e));
@@ -170,12 +166,7 @@ public class ProjectMetadataUtilities {
     }
 
     static protected ProjectMetadata loadFromFile(File metadataFile) throws Exception {
-        Reader reader = new FileReader(metadataFile);
-        return ParsingUtilities.mapper.readValue(reader, ProjectMetadata.class);
-    }
-
-    static protected ProjectMetadata loadFromFile(File metadataFile, Charset charset) throws Exception {
-        Reader reader = new InputStreamReader(new FileInputStream(metadataFile), charset);
+        FileReader reader = new FileReader(metadataFile);
         return ParsingUtilities.mapper.readValue(reader, ProjectMetadata.class);
     }
 }
--- a/main/src/com/google/refine/io/ProjectMetadataUtilities.java
+++ b/main/src/com/google/refine/io/ProjectMetadataUtilities.java
@@ -34,11 +34,15 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package com.google.refine.io;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.FileReader;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
+import java.io.Reader;
 import java.io.Writer;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.time.LocalDateTime;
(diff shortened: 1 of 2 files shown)


main/src/com/google/refine/io/ProjectMetadataUtilities.java
```diff
@@ -33,12 +33,7 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
 
 package com.google.refine.io;
 
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.FileReader;
-import java.io.IOException;
-import java.io.OutputStreamWriter;
-import java.io.Writer;
+import java.io.*;
```
sherrif10, line 36: i would rather leave it as full package.   cc @antoine2711

elroykanye: Oh yes, thank you. Didn't realise that slipped through 😄

elroykanye:
> (It would be great to check that the test fails without your fix!)

Very well. Will do just that.

elroykanye:
> (It would be great to check that the test fails without your fix!)

Hello @wetneb , I have thought about this and I think that we will have to make the charset optional and call another method (probably that which will have the fix) to be able to have a demonstration from the loadFromFile where it will fail if the UTF-8 charset is not specified. 
I will push some commits with this so you see what I mean.

main/tests/server/src/com/google/refine/io/ProjectMetadataUtilitiesTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    ProjectMetadata actualMetadata = new ProjectMetadata();
+    ProjectMetadata expectedMetadata = new ProjectMetadata();
+    File file = null;
+
+    @Before
+    public void setUp() {
+        actualMetadata.setName("Velmi dobře, použijeme vícejazyčný neurální stroj, abychom to vyřešili. ÜÜÖÖÄÄ");
+
+        file = new File("tests/server/src/com/google/refine/io/example_project_metadata_charset_test.json");
```
wetneb, line 23: It would be better to use a temporary file instead (files created in this location will not be cleaned by `mvn clean`).
You should be able to use `File.createTempFile` for that.

elroykanye: 🤔 true! Alright, thanks.

wetneb:
Thanks! So, should we do an experiment with the CI? We could temporarily revert your changes to the main code and keep your test, and see if the Windows CI fails.

wetneb:
Could you do it?

elroykanye:
I am not sure I know how to do it. If I could get guidelines or a sample, I will surely find my way around.

wetneb:
So: the CI is all green, including the Windows tests, so there is a chance that the test does not fully capture the bug to be fixed.
Perhaps this only happens if a default encoding is set: we could potentially change the CI configuration (in `.github/workflows` in this repository) so that the Windows CI runs with a non-UTF8 default encoding. Maybe that would help reproducing this. What do you think?

elroykanye:
> So: the CI is all green, including the Windows tests, so there is a chance that the test does not fully capture the bug to be fixed. Perhaps this only happens if a default encoding is set: we could potentially change the CI configuration (in `.github/workflows` in this repository) so that the Windows CI runs with a non-UTF8 default encoding. Maybe that would help reproducing this. What do you think?

I had to modify the `Assert.assertNotEqual` to `Assert.assertEqual` to see if the tests will fail on windows. I did so because I think the general expectation is for the initially set name before saving the file, to be equal to the name gotten after loading the file. Is that okay?

wetneb:
Well, if that passes again then that is an indication that your test is not getting run at all! That is probably because you have imported the Junit `@Test` annotation, not the one from TestNG. You should make sure you only import things from `org.testng` (and I should make sure Junit is not even on the classpath…).

wetneb:
Amazing, we have some wonderful failures here. Do you fancy reverting my commit and seeing what happens?

elroykanye:
I will have to revert up to this point: https://github.com/OpenRefine/OpenRefine/pull/4768/commits/11d43a4e32411712edd8b274fe0e6e65ed788c15, right?

elroykanye:
Okay, I reverted just at the point where you made the change. If this is adopted, there won't be a need for overloading the `loadFromFile`, right? 
And consequently, no need for the test which checks the original `loadFromFile` method?

wetneb:
Ok great! (by the way, we already run the tests with a non-UTF8 encoding even for Linux, hence the failures on Linux too: https://github.com/OpenRefine/OpenRefine/blob/89fdc837ee5b8bafada26fed5aa7209ddc7d0c38/pom.xml#L167)

>  And consequently, no need for the test which checks the original loadFromFile method?

I would do the opposite: I would keep the original loadFromFile method, no need to have one where the encoding is specified, I would say, unless I missed something.

elroykanye:
> I would do the opposite: I would keep the original loadFromFile method, no need to have one where the encoding is specified, I would say, unless I missed something.

I think we need the encoding specified because it is what solves the issue. The 2 tests are for the 2 variants of `loadFromFile`, so I thought we could have just one `loadFromFile` with the UTF-8 encoding specified and its corresponding test since the `loadFromFile` without specified encoding is not used anywhere else in the code except in its own test.

wetneb:
> I thought we could have just one loadFromFile with the UTF-8 encoding specified and its corresponding test since the loadFromFile without specified encoding is not used anywhere else in the code except in its own test.

There we agree! Yes let us only keep the method which has less arguments (that was what I tried to write earlier).

elroykanye:
Awesome! Coming up with that then 😁

main/tests/server/src/com/google/refine/io/ProjectMetadataUtilitiesTest.java
```diff
@@ -0,0 +1,68 @@
...
+    File file = null;
+
+    @BeforeTest
+    public void setUp() {
+        actualMetadata.setName("Velmi dobře, použijeme vícejazyčný neurální stroj, abychom to vyřešili. ÜÜÖÖÄÄ");
+
+        try {
+            file = File.createTempFile("example_project_metadata_charset_test", ".json");
+        } catch (IOException e) {
+            e.printStackTrace();
```
wetneb, line 51: One last thing: it is better not to catch the exception here, so that the `setUp` method properly fails (and similar for the test below). You can just add the corresponding `throws` statement to the methods.
This is especially important for the test: if an Exception occurs, we want the test to fail, and currently it would succeed even though it printed a stack trace in the console.

elroykanye: Thanks for this @wetneb
I understand now 😊

wetneb (approved):
Thank you!

PR #4685, opened 2022-04-02
[SQL Exporter] Substitute invalid characters in column names with underscore (#1940)

shelomi99:
Fixes #1940

Changes proposed in this pull request:
- Substituting invalid characters in column names with an underscore for SQL Exporter
- The invalid characters are detected through the following regex: `[^a-zA-Z0-9_]`

PR #4776, opened 2022-04-19
(I#4716) Escape character for fields name when exporting in SQL

### LINKED ISSUE #4716: Escape character when exporting in SQL
It would be good to be able to have data escaped with the SQL importer, and to be able to choose by which character.
This was discussed in Issue #1940

### Proposed solution
Add a text field for an escape character in the SQL Exporter.

### Alternatives considered
data manipulation at the text level after exportation.

### Additional context
image

the-star-sea:
Fixes #4716

- Change  main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java to support escape character
- Change main/webapp/modules/core/langs to offer text for escape character
- Change main/webapp/modules/core/scripts/dialogs/sql-exporter-dialog.html,main/webapp/modules/core/scripts/dialogs/sql-exporter-dialog.js to offer the escape layout

### DIFF
--- a/main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
+++ b/main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
@@ -62,13 +62,14 @@ public class SqlCreateBuilder {
 
         List<JsonNode> columnOptionArray = options == null ? Collections.emptyList() : JSONUtilities.getArray(options, "columns");
         boolean trimColNames = options == null ? false : JSONUtilities.getBoolean(options, "trimColumnNames", false);
-
+        String EscapeCharacter=JSONUtilities.getString(options, "EscapeCharacter", null);
         int count = columnOptionArray.size();
 
         for (int i = 0; i < count; i++) {
             JsonNode columnOptions = columnOptionArray.get(i);
             if (columnOptions != null) {
                 String name = JSONUtilities.getString(columnOptions, "name", null);
+
                 String type = JSONUtilities.getString(columnOptions, "type", SqlData.SQL_TYPE_VARCHAR);
                 String size = JSONUtilities.getString(columnOptions, "size", "");
                 boolean allowNull = JSONUtilities.getBoolean(columnOptions, "allowNull", true);
@@ -83,11 +84,12 @@ public class SqlCreateBuilder {
                 
                 if (name != null) {
                     if(trimColNames) {
-                        String trimmedCol = name.replaceAll("[^a-zA-Z0-9_]", "_");
-                        createSB.append( trimmedCol + " ");
-                    }else{
-                        createSB.append(name + " ");
+                        name = name.replaceAll("[^a-zA-Z0-9_]", "_");
+
+                    }if (EscapeCharacter!=null){
+                        name=EscapeCharacter+name+EscapeCharacter;
                     }
+                    createSB.append( name + " ");
                    
                     if (type.equals(SqlData.SQL_TYPE_VARCHAR)) {
                         if (size.isEmpty()) {
@@ -160,7 +162,7 @@ public class SqlCreateBuilder {
         sql.append(")").append(";" + "\n");
         
         String createSQL = sql.toString();
-        if(logger.isDebugEnabled()){
+        if(logger.isDebugEnabled()) {
             logger.debug("Create SQL Generated Successfully...{}", createSQL);
         }
         return createSQL;
--- a/main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
+++ b/main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
@@ -85,8 +85,8 @@ public class SqlCreateBuilder {
                 if (name != null) {
                     if(trimColNames) {
                         name = name.replaceAll("[^a-zA-Z0-9_]", "_");
-
-                    }if (EscapeCharacter!=null){
+                    }
+                    if (EscapeCharacter!=null){
                         name=EscapeCharacter+name+EscapeCharacter;
                     }
                     createSB.append( name + " ");
--- a/main/webapp/modules/core/langs/translation-zh.json
+++ b/main/webapp/modules/core/langs/translation-zh.json
@@ -128,6 +128,7 @@
     "core-dialogs/no-cluster-found": "选中的操作并没有生成簇",
     "core-dialogs/try-another-method": "请尝试上面的其他操作或者修改操作的参数",
     "core-dialogs/clustering": "簇集中…",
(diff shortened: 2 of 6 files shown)


wetneb:
Thanks! Personally I am not sure I understand the motivation of the original issue. As far as I can tell, https://github.com/OpenRefine/OpenRefine/pull/4685 removes non-standard characters so there is no need for escaping anymore. Allowing the user to specify an escape character is only useful if we also let them opt out of the newly-introduced column name normalisation, right?

antoine2711:
> Thanks! Personally I am not sure I understand the motivation of the original issue. As far as I can tell, #4685 removes non-standard characters so there is no need for escaping anymore. Allowing the user to specify an escape character is only useful if we also let them opt out of the newly-introduced column name normalisation, right?

@wetneb: maybe I could review this issue personaly? I also need to learn a bit, and it would remove that burden. Is it acceptable?

Regards, Antoine

the-star-sea:
I have another version without the replacement merged into.I have doubt that I need to delete the replacement function when I did the job.

the-star-sea (commented):
I have review the fix part.

the-star-sea (commented):
I have reviewed all my code

antoine2711:
> I have reviewed all my code

When you are ready for code review, go to the top right of the PR:
<kbd>image</kbd>

And there, you can choose @wetneb, it will warn Antonin.
<kbd>image</kbd>

Regards, Antoine

main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
```diff
@@ -83,11 +84,12 @@ public String getCreateSQL() {
...
                     if(trimColNames) {
-                        String trimmedCol = name.replaceAll("[^a-zA-Z0-9_]", "_");
-                        createSB.append( trimmedCol + " ");
-                    }else{
-                        createSB.append(name + " ");
+                        name = name.replaceAll("[^a-zA-Z0-9_]", "_");
+
+                    }if (EscapeCharacter!=null){
+                        name=EscapeCharacter+name+EscapeCharacter;
                     }
```
antoine2711, line 91: @the-star-sea: here, your name for the variable starts with an uppercase. But, other options are in lowercase. Also, the name could be more precise: since we have `trimColNames`, `escapeColNames` is close, yet significative and different enough. `EscapeCharacter` doesn't really say to what it applies to, only what it does. `escapeColNamesCharacter` say both.

main/webapp/modules/core/langs/translation-zh.json
```diff
@@ -128,6 +128,7 @@
     "core-dialogs/no-cluster-found": "选中的操作并没有生成簇",
     "core-dialogs/try-another-method": "请尝试上面的其他操作或者修改操作的参数",
     "core-dialogs/clustering": "簇集中…",
+    "core-dialogs/sqlExportEscapeCharacter": "转义字符",
```
antoine2711, line 131: Here also, there is naming convention to follow. No uppercase, and dash to separate words.
```suggestion
    "core-dialogs/sql-export-escape-char": "转义字符",
```

What do you think of that?

Regards, Antoine

main/webapp/modules/core/scripts/dialogs/sql-exporter-dialog.html
```diff
@@ -57,7 +57,9 @@
 			     
 			     <td width="1%"><input type="checkbox" bind="sqlExportTrimAllColumnsCheckbox" id="sqlExportTrimAllColumnsCheckboxId" /></td>
 			     <td width="25%"><label for="sqlExportTrimAllColumnsCheckbox" bind="sqlExportTrimAllColumnsLabel"></label></td>
-                
+
+                  <td width="1%"><input type="text" style="width:20px" bind="sqlExportEscapeCharacterTextBox" id="EscapeCharacterTextBoxId" /></td>
```
antoine2711, line 61:
```suggestion
                  <td width="1%"><input type="text" style="width:20px" bind="sqlExportEscapeCharacterTextBox" id="sqlExportEscapeCharacterTextBoxId" /></td>
```

main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
```diff
@@ -83,11 +84,12 @@ public String getCreateSQL() {
                 
                 if (name != null) {
                     if(trimColNames) {
-                        String trimmedCol = name.replaceAll("[^a-zA-Z0-9_]", "_");
-                        createSB.append( trimmedCol + " ");
-                    }else{
-                        createSB.append(name + " ");
+                        name = name.replaceAll("[^a-zA-Z0-9_]", "_");
+
+                    }if (EscapeCharacter!=null){
```
antoine2711, line 89:
```suggestion
                    }
                    if (EscapeCharacter!=null){
```
This is a visual change, but was misleading the other way.

antoine2711 (changes requested):

antoine2711 (changes requested):
```
- [ ] First, it's been requested that both option can not be applied at the same time, `Trim Column Names` and `Escape Character`. So, if `Trim Column Names` is checked, then `Escape Character` should be disabled.
- [ ] The escape character must be valid, and not any character. Logical choices are: `'`, `"`,
```. Is there others?
- [ ] Only ONE escape character at the same time is possible. Now, you can enter multiple.

It's a good start, but UI interaction is missing.

Regards, Antoine

main/webapp/modules/core/langs/translation-zh.json
the-star-sea: I think my way is better because I follow the other three labels.

the-star-sea (commented):
Thanks for your review.However,I think sqlExportEscapeCharacter is better because its neighbour checkboxes which are also under core-dialogs use the name way

main/webapp/modules/core/langs/translation-zh.json
antoine2711: > I think my way is better because I follow the other three labels. 

Wow @the-star-sea! You are making an intersting observation. But the sad thruth is probably that you've unearthed what are mistakes. There are a few other cases, but it's rare.
"core-index/rowCount", "core-index/customMetadata", "core-index/id": "项目编号：", "core-index/importOptionMetadata"

@wetneb: dou we have a rule for the naming of the translation strings?

Regards, Antoine

wetneb:
@antoine2711 I am tentatively "assigning" you here, to signal that your review is requested. As I mentioned in the issue, I think it is not worth giving the user a choice of the character to use to quote values, since `"` should always work as it is mandated by the standard.

wetneb:
Since this PR is inactive and getting stale I would close it soon.

PR #4779, opened 2022-04-20
Fix date formatting in custom tabular exporter

### LINKED ISSUE #3368: Export > Custom tabular exporter: date formatting doesn't work
Using the custom tabular exporter to export dates into locale formats (short, medium, long, full) doesn't seem to have any effect in any of the available output formats. 

custom

### To Reproduce
Export some columns of identical dates into the five available locale formats. 

### Current Results

HTML tables, XLS, XLSX, CSV, and TSV all output the original data into each cell, no matter which locale format you've chosen. Omitting time does nothing. The "use local time zone" checkbox does nothing. 

> Short,Medium,Long,Full,ISO
2020-09-16T13:15:31Z,2020-09-16T13:15:31Z,2020-09-16T13:15:31Z,2020-09-16T13:15:31Z,2020-09-16T13:15:31Z
2020-09-15T13:15:00Z,2020-09-15T13:15:00Z,2020-09-15T13:15:00Z,2020-09-15T13:15:00Z,2020-09-15T13:15:00Z

### Expected Behavior
Presumably (like some of the other settings for the custom tabular exporter) this is intended to work in at least one of the available output formats, but maybe not all of them. Can anyone tell me which formats? 

There seem to be other discussions (#3322 for example) around this right now so I just want to register this as something to test when those are worked on.

xbdeng:
Fixes #3368

- main/src/com/google/refine/exporters/CustomizableTabularExporterUtilities.java
The CellFormatter constructor is always defaulting and the data formatting is always missing. I made some changes to invoke the constructor with parameters, so the data formatting can be set by these parameters.
I will pull some test cases soon if it works.


xbdeng:
@wetneb I clone the newest code of this repository, but it can not pass the test cases as well. I told some friends to try but they also failed to pass the test. So I think the source code may have some problems. 
If it is my blame, I need your help because I am a newcomer.
Thank you!

antoine2711:
> @wetneb I clone the newest code of this repository, but it can not pass the test cases as well. I told some friends to try but they also failed to pass the test. So I think the source code may have some problems. If it is my blame, I need your help because I am a newcomer. Thank you!

@xbdeng: on what OS are you? I have a lots of problems on Window… none on Mac or Linux.

Regards, Antoine

xbdeng:
> > @wetneb I clone the newest code of this repository, but it can not pass the test cases as well. I told some friends to try but they also failed to pass the test. So I think the source code may have some problems. If it is my blame, I need your help because I am a newcomer. Thank you!
> 
> @xbdeng: on what OS are you? I have a lots of problems on Window… none on Mac or Linux.
> 
> Regards, Antoine

@antoine2711 I use Ubuntu, my friend uses Mac, but we failed to pass the test. I wander If you could pass the test with the newest code. Thank you!

elroykanye:
Hi @xbdeng 🙋🏾, 

I wish to help you with this in checking why your checks are failing, but I noticed you opened a PR from the master branch on your fork. Following the [contributing guidelines](https://docs.openrefine.org/technical-reference/contributing#your-first-code-pull-request), we have to make branches of the master which have the issue number and fix description. I think this helps us have a central space for keeping safe code and opens up multiple contributions to be made from your fork.

I would suggest starting clean from a different appropriate branch and running the corresponding tests for the SQL exporter per change you make. Find the tests [here](https://github.com/OpenRefine/OpenRefine/blob/master/main/tests/server/src/com/google/refine/exporters/sql/SqlExporterTests.java).

xbdeng:
I think I fix the bug, I can pass the test now, should I pull a new request? @elroykanye

antoine2711:
> I think I fix the bug, I can pass the test now, should I pull a new request? @elroykanye

@xbdeng: I'd say, since he's overlooking you, that you can put him as the reviewer, and I'll know then that you are ready for a review.
image

Regards, Antoine

xbdeng:
I have pull a new PR, may be this can be closed?
I don't know how to pull a new request under this PR. @antoine2711

xbdeng:
@wetneb I pass the test!

main/src/com/google/refine/exporters/CustomizableTabularExporterUtilities.java
```diff
@@ -114,27 +114,33 @@ static public void exportRows(
...
                     String name = JSONUtilities.getString(columnOptions, "name", null);
                     if (name != null) {
                         columnNames.add(name);
-                        try {
-							columnNameToFormatter.put(name, ParsingUtilities.mapper.treeToValue(columnOptions, ColumnOptions.class));
-						} catch (JsonProcessingException e) {
-							e.printStackTrace();
-						}
+                        JsonNode reconSettingsNode = columnOptions.get("reconSettings");
+                        JsonNode dateSettingsNode = columnOptions.get("dateSettings");
```
wetneb, line 126: Instead of writing deserialization code by hand, I think it would be better to rely on Jackson deserialization with `ParsingUtilities.mapper.treeToValue(columnOptions, ColumnOptions.class)`. If this does not currently work well, there should be a way to fix the constructor of this class to make sure it deserializes things correctly.

main/src/com/google/refine/exporters/CustomizableTabularExporterUtilities.java
```diff
@@ -234,16 +240,31 @@ private enum DateFormatMode {
...
+
     static private class ReconSettings {
         @JsonProperty("output")
         ReconOutputMode outputMode = ReconOutputMode.ENTITY_NAME;
         @JsonProperty("blankUnmatchedCells")
         boolean blankUnmatchedCells = false;
         @JsonProperty("linkToEntityPages")
         boolean linkToEntityPages = true;
+
+        ReconSettings(){}
```
wetneb, line 252: This should not be needed, I think: the default constructor should exist by default, no?

main/src/com/google/refine/exporters/CustomizableTabularExporterUtilities.java
```diff
@@ -253,29 +274,55 @@ static private class DateSettings {
         boolean useLocalTimeZone = false;
         @JsonProperty("omitTime")
         boolean omitTime = false;
+
+        DateSettings(){}
```
wetneb, line 278: Same here.

main/src/com/google/refine/exporters/CustomizableTabularExporterUtilities.java
xbdeng: I found that `ParsingUtilities.mapper.treeToValue(columnOptions, ColumnOptions.class)` does not work well. It will lose the date format. The date format will always be as the default. But I do not know how to fix it, could you give some advice?
Writing deserialization code by hand can solve it, so I think the util may cause some bug.

main/src/com/google/refine/exporters/CustomizableTabularExporterUtilities.java
xbdeng: I add other constructor, I think....

xbdeng:
I am so sorry I misunderstood the issue before. But now I think I have fixed it!
@wetneb

xbdeng:
Could you help me approval the workflow? @antoine2711

xbdeng:
I fix the format bug, could  you help me approval the workflow again? @antoine2711

xbdeng:
I don't know why I can not pass my own added test case, I can pass it on my Ubuntu. I need your help to solve it, sorry....
image

antoine2711:
> I don't know why I can not pass my own added test case, I can pass it on my Ubuntu. I need your help to solve it, sorry.... 

Don't be sorry. I also have those problems! :-( 
https://groups.google.com/g/openrefine-dev/c/EB_ofpAu3jY

I'll try to relaunch them. It a pleasure to help you, @xbdeng.

antoine2711:
> I fix the format bug, could you help me approval the workflow again? @antoine2711

@wetneb: I have questions.
1. I run the test on Mac & Linux Ubuntu, no problem.
2. Here on the CD/CI, it break both for Linux (x2) and Windows
3. On windows, refine.bat test or refine.bat test do not seams to work.

How can I help this contributor?

Regards, Antoine

wetneb (commented):
Thanks, it looks like your fix is the right one!
I am not sure how you are running the test locally, but according to the CI output, your expected CSV output is indeed missing some quotes since your cell values contain `,`, which is also used as separator.
I would add the required quotes to the expected export output and your test should pass as expected.

wetneb:
Thinking about it more, this seems to indicate that your machines are using different locale settings than the CI, and for those locale settings the generated dates do not have a comma in them.
To make sure the test passes reliably, you could perhaps use TSV instead of CSV, to make sure the formatted dates do not contain any separator regardless of the locale.

antoine2711:
> Thinking about it more, this seems to indicate that your machines are using different locale settings than the CI, and for those locale settings the generated dates do not have a comma in them. To make sure the test passes reliably, you could perhaps use TSV instead of CSV, to make sure the formatted dates do not contain any separator regardless of the locale.

@wetneb: should't our tests be agnostics of the OS or user setting, and shouldn't our test do a few different local setting? Do we have a few languages that covers 80 or 90% of our user base? 

I guess that both date setting, and the list separator setting could have an effet...

Regards, Antoine

wetneb:
Hi @xbdeng, are you still working on this?

xbdeng:
Yes, maybe I need more time......

xbdeng:
I noticed #2071 may solve this?

wetneb:
Closing per inactivity, we can reopen if you get round to working on this again.

PR #4780, opened 2022-04-20
(I#4048) Fix Documentation for Clustering

### LINKED ISSUE #4048: Fix documentation for clustering
[The documentation about clustering](https://docs.openrefine.org/manual/cellediting#cluster-and-edit) shows a screenshot of the clustering dialog where all the values in a given cluster look identical, which should not normally happen. This screenshot was taken before #4004 was fixed, which probably explains that. This screenshot should be replaced by a screenshot which shows different values being clustered together.

Current image:
image

Additionally it could be useful to rework the text to make the clustering feature clearer because it looks like the current documentation does not convey the intended usage very well.

This was noticed in #3422 by @fgiroud.

Babi-B:
Fixes #4048

- Create a new screenshot of the clustering dialog
- Replace the old screenshot in the documentation with the new one


docs/docs/manual/cellediting.md
```diff
@@ -93,7 +93,7 @@ OpenRefine will do a number of cleanup operations behind the scenes in order to
 
 You can start the process in two ways: using the dropdown menu on your column, select <span class="menuItems">Edit cells</span> → <span class="menuItems">Cluster and edit…</span>; or create a text facet and then press the “Cluster” button that appears in the facet box. 
 
-![A screenshot of the Clustering window.](/img/cluster.png)
+![A screenshot of the Clustering window.](/img/new_cluster.jpg)
```
wetneb, line 96: I think you could just replace the existing image, so that we do not store unneeded copies.

wetneb (commented):
Thanks!
It would be great if the screenshot could be fit tightly around the frame of the dialog, I would find that a bit cleaner (I know we do not enforce that throughout the rest of the documentation).

docs/docs/manual/cellediting.md
antoine2711: @Babi-B: have you seen this?

Babi-B: > have you seen this?

@antoine2711 @wetneb: I must have missed this notification

docs/docs/manual/cellediting.md
```diff
@@ -93,7 +93,7 @@ OpenRefine will do a number of cleanup operations behind the scenes in order to
 
 You can start the process in two ways: using the dropdown menu on your column, select <span class="menuItems">Edit cells</span> → <span class="menuItems">Cluster and edit…</span>; or create a text facet and then press the “Cluster” button that appears in the facet box. 
 
-![A screenshot of the Clustering window.](/img/cluster.png)
+![A screenshot of the Clustering window.](/img/cluster.jpg)
```
wetneb, line 96: I think it is generally preferable to use the `png` format rather than `jpg` for images on the web.
See also my previous comment about cropping the screenshot tightly around the border of the dialog.

wetneb (approved):
Thank you!

PR #4784, opened 2022-04-21
(I#608) epoch to date grel function

### LINKED ISSUE #608: Support of conversion of millisecond Unix time to Refine datetime

keerthybreddy:
Fixes #608 

- Added epochToDate function and epochToDateTest class, similar code to @SustechcsKing 's contribution in issue #3928
- Changed hard-coded "America/Chicago" timezone to "UTC" universal time zone
- fixed conditional statements to check arguments correctly by replacing "==" with ".equals()"

### DIFF
--- a/main/src/com/google/refine/expr/functions/EpochToDate.java
+++ b/main/src/com/google/refine/expr/functions/EpochToDate.java
@@ -13,6 +13,10 @@ public class EpochToDate implements Function {
 
     @Override
     public Object call(Properties bindings, Object[] args) {
+        long timeStampMilliSecond = 1650547184707L;
+        long timeStampSecond = timeStampMilliSecond / 1000;
+        long timeStampMicroSecond = timeStampMilliSecond * 1000;
+        System.out.println("Seconds timeStamp: " + timeStampSecond + " Microseconds timeStamp: " + timeStampMicroSecond);
         if (args.length >= 1 && args[0] instanceof Number) {
             long epoch = (long) args[0];
             ZoneId zoneId = ZoneId.of("UTC");
--- a/main/src/com/google/refine/expr/functions/EpochToDate.java
+++ b/main/src/com/google/refine/expr/functions/EpochToDate.java
@@ -26,33 +26,32 @@ import com.google.refine.grel.ControlFunctionRegistry;
 import com.google.refine.grel.Function;
 import com.google.refine.util.ParsingUtilities;
 
-public class EpochToDate implements Function{
+public class EpochToDate implements Function {
 
     @Override
     public Object call(Properties bindings, Object[] args) {
-        if(args.length >= 1 && args[0] instanceof Number) {
+        if (args.length >= 1 && args[0] instanceof Number) {
             long epoch = (long) args[0];
             ZoneId zoneId = ZoneId.of("UTC");
-            //ZoneId zoneId = ZoneId.systemDefault();
             Instant instant = null;
             OffsetDateTime date = null;
-            if(args.length == 1){
-                instant = Instant.ofEpochSecond ( epoch );
+            if (args.length == 1) {
+                instant = Instant.ofEpochSecond(epoch);
                 date = OffsetDateTime.ofInstant(instant, zoneId);
                 return date;
-            } else if(args.length == 2 && args[1] instanceof String){
+            } else if (args.length == 2 && args[1] instanceof String) {
                 Object o2 = args[1];
-                String type = ((String ) o2).toLowerCase();
-                if(type.equals("second")){
-                    instant = Instant.ofEpochSecond ( epoch );
+                String type = ((String) o2).toLowerCase();
+                if (type.equals("second")) {
+                    instant = Instant.ofEpochSecond(epoch);
                     date = OffsetDateTime.ofInstant(instant, zoneId);
                     return date;
-                } else if (type.equals("millisecond")){
-                    instant = Instant.ofEpochSecond ( epoch/1000 );
+                } else if (type.equals("millisecond")) {
+                    instant = Instant.ofEpochSecond(epoch / 1000);
                     date = OffsetDateTime.ofInstant(instant, zoneId);
                     return date;
-                } else if (type.equals("microsecond")){
-                    instant = Instant.ofEpochSecond ( epoch/1000000 );
+                } else if (type.equals("microsecond")) {
+                    instant = Instant.ofEpochSecond(epoch / 1000000);
                     date = OffsetDateTime.ofInstant(instant, zoneId);
                     return date;
                 }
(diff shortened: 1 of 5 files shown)


wetneb:
Thanks! I am not reviewing yet because the tests fail. Feel free to ping me when they do or you need help to get there.

elroykanye:
@keerthybreddy , you may have to 
- run the formatter tool to make sure your code is correct: `mvn formatter:format`
- run the server tests locally to ensure your changes integrate well with the application
- verify what is failing on the checks too to be sure you are not missing anything

keerthybreddy:
> Thanks! I am not reviewing yet because the tests fail. Feel free to ping me when they do or you need help to get there.

Thanks for getting back to me! 
I have just made another commit but it still seems to fail one test. In order to make my code pass all the tests, do I need to write a Cypress test or a JUnit test, or both? @SustechcsKing had written a test class that I used to test my code changes, but it doesn't seem to be sufficient to pass all of the GitHub tests for me.

keerthybreddy:
> @keerthybreddy , you may have to
> 
> * run the formatter tool to make sure your code is correct: `mvn formatter:format`
> * run the server tests locally to ensure your changes integrate well with the application
> * verify what is failing on the checks too to be sure you are not missing anything

Got it, Thank you for the feedback! I am currently working on running the server tests locally.

keerthybreddy:
Hi @wetneb , I think I was able to pass the all of the tests, please let me know if you are able to review my my pull request. Thank you!

WaltonG:
Hi @keerthybreddy The integration is still failing. There is a decrease in coverage because the unit test did not cover/test the return statement on lines 43, 51, 58, 63 in EpochToDate. java

keerthybreddy:
> Hi @keerthybreddy The integration is still failing. There is a decrease in coverage because the unit test did not cover/test the return statement on lines 43, 51, 58, 63 in EpochToDate. java

Thank you for clarifying that! I was confused as to why I was failing that test case.

keerthybreddy:
@wetneb , I believe that I was able to write and run tests that cover the lines that @WaltonG  mentioned in the previous comment successfully. Please let me know if you are able to review my code. Thank you!

main/src/com/google/refine/expr/functions/EpochToDate.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expect one argument, " +
+                "which is a number of epoch time\n" +
+                "expect two argument, the first is a epoch time(second, millisecond, microsecond), the second " +
+                "is the type");
+    }
+
+    @Override
+    public String getDescription() {
```
ostephens, line 50: Again suggest that this could be slightly more concise. I suggest reviewing https://github.com/OpenRefine/OpenRefine/blob/235b5957e22c73328e9f84d6fac44a9cb92318f5/main/src/com/google/refine/expr/functions/strings/Phonetic.java#L104 for an example of a description.

e.g.
"Returns a number converted to a date based on Unix Epoch Time. The number can be Unix Epoch Time in one of the following supported units: second, millisecond, microsecond. Defaults to 'decode'."

main/src/com/google/refine/expr/functions/EpochToDate.java
```diff
@@ -0,0 +1,66 @@
...
+            Instant instant = null;
+            OffsetDateTime date = null;
+            if (args.length == 1) {
+                instant = Instant.ofEpochSecond(epoch);
+                date = OffsetDateTime.ofInstant(instant, zoneId);
+                return date;
+            } else if (args.length == 2 && args[1] instanceof String) {
+                Object o2 = args[1];
+                String type = ((String) o2).toLowerCase();
+                if (type.equals("second")) {
```
ostephens, line 28: Would 'unit' be a more intuitive name for this property rather than 'type'? Whichever is used in the code, IMO for the user 'unit' makes more sense

main/src/com/google/refine/expr/functions/EpochToDate.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public String getDescription() {
+        return "Returns a number converted to a date. Can parse one parameter or two parameters. When parsing one parameter, the number is the epoch second."
+                +
+                "When parsing two parameters, the first is the number, the second is the numbers type, such as second, millisecond, microsecond.";
+    }
+
+    @Override
+    public String getParams() {
```
ostephens, line 57: The getParams response should follow the same pattern of other functions. Refer to existing functions for the appropriate form for the getParams response. e.g. see https://github.com/OpenRefine/OpenRefine/blob/235b5957e22c73328e9f84d6fac44a9cb92318f5/main/src/com/google/refine/expr/functions/date/Inc.java#L93

More like:
"number n, string unit (optional, defaults to 'seconds')";

wetneb:
Is the name of this function already agreed on? I think "epoch" rarely refers to a date format, but rather the date it is counted from.  People usually refer to those dates as "POSIX" or "UNIX" in my experience, so I would rather expect the function to be named "posixToDate" or "unixToDate", but it is not a hill I will die on of course!

ostephens:
[Wikipedia lists the names](https://en.wikipedia.org/wiki/Unix_time) as:
* Unix time
* Epoch time (although later in the article this is described as incorrect)
* Posix time
* seconds since the Epoch
* UNIX Epoch time

Although strictly Unix time is always in seconds since the Unix epoch

Unix and Posix also aren't date formats so I'm not so keen on posixToDate or unixToDate for the same reason as @wetneb  doesn't like epochToDate.

This might be better discussed on the original issue #608 rather than on the PR?

keerthybreddy:
Hello! I was able to push the changes that @ostephens recommended to me and I changed the name of the function to TimeSinceUnixEpochToDate() , please let me know if anyone can review my code. @wetneb, tagging you here for visibility :) Thank you!

wetneb (commented):
Great! One last thing: the function must be registered so that it can be used in GREL expressions (in `ControlFunctionRegistry`).

keerthybreddy:
> Great! One last thing: the function must be registered so that it can be used in GREL expressions (in `ControlFunctionRegistry`).

Got it! Just registered the function properly and pushed the change again. Please let me know if you have any further feedback @wetneb

main/src/com/google/refine/expr/functions/TimeSinceUnixEpochToDate.java
```diff
@@ -0,0 +1,62 @@
...
+                }
+            }
+        }
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this)
+                + " accepts a number and an optional second argument containing a string specifying the units");
+    }
+
+    @Override
+    public String getDescription() {
+        return "Returns a number converted to a date based on Unix Epoch Time. The number can be Unix Epoch Time in one of the following supported units: second, millisecond, microsecond. Defaults to 'decode'.";
```
ostephens, line 49: I think I made an error in my comment - I guess this should be "Defaults to 'second'" rather than "Defaults to 'decode'"?

ostephens (changes requested):

main/src/com/google/refine/expr/functions/TimeSinceUnixEpochToDate.java
keerthybreddy: Got it! Just pushed the change that you requested :)

keerthybreddy: > I think I made an error in my comment - I guess this should be "Defaults to 'second'" rather than "Defaults to 'decode'"?
Got it! Just pushed the change that you requested :)

ostephens (approved):

PR #4786, opened 2022-04-21
(I#2326) Support custom rank in Wikidata schema

### LINKED ISSUE #2326: Support for custom ranks in Wikidata schema

Wikidata statements can have three ["ranks"](https://www.wikidata.org/wiki/Help:Ranking): Preferred, Normal or Deprecated.

When creating Wikidata statements from OpenRefine, in the Wikidata schema, it is not currently possible to set a custom rank for the statements created: the rank is always "normal".

There is already an image for ranks in the UI but is not clickable.

It would be useful if the rank image in each statement could be clickable and one could define a custom rank, similarly to the way it is done on Wikidata itself.

**Additional context**

The rank icon looks like this:  image. This represents the three ranks (the grey filling moves depending on which rank is active).

Here is what it looks like in Wikidata:
image
When editing the statement:
image

Current UI in OpenRefine:
image

SIXIN-LI:
Partly Fixes #2326 except uploading statements with ranks to wikibase since the design detail is still in process.

- Made the rank icon in wikibase schema clickable
- Added rank to wikibase statement expression


wetneb:
Thanks for the PR!
It looks like the tests are failing, so we would need to fix that first.
Also, there are quite some spurious whitespace changes in the JS file, ideally it would be better to avoid them (although I know it is painful to restore those trailing whitespaces…)

wetneb:
Hi @SIXIN-LI, are you still working on this PR?

SIXIN-LI:
Sorry, I didn't see the first review comment.

Also, in the discussion of the #2326 issue, the need for this enhancement becomes undetermined. Should I continue?

wetneb:
It is up to you! I think the design work is doable, but it does require some good familiarity with Wikidata/Wikibase. In any case the work you did is definitely useful so it should be a good basis towards this feature, so it would be a good basis for any PR that tackles it!

SIXIN-LI:
It is inspirational! I will fix this PR for future use.

wetneb:
Closing per inactivity then.

Issue #3719, opened 2021-03-14
Can't see projects (but can open them)


jkjkjkjkjk:
No project is showing in open project dialog box, only overlapping All a and Browse workspace directory buttons (same like https://github.com/OpenRefine/OpenRefine/issues/3277).  I am able to open workspace directory and I am able to open the projects using direct link (http://127.0.0.1:3333/project?project=xxx).

### To Reproduce
I have deleted one project directory from workspace directory. 

### Screenshots
<a href="https://ibb.co/Df8SRJf"><img src="https://i.ibb.co/Df8SRJf/OpRef.jpg" alt="OpRef" border="0"></a>

### Versions
 - Operating System: Windows 7
 - Browser Version: Chrome 89
 - OpenRefine: OpenRefine 3.4.1

gitonthescene:
Did you see [this comment](https://github.com/OpenRefine/OpenRefine/issues/3277#issuecomment-721305890) in the issue you referenced?

Can you try starting with the `-d` pointing to a directory with only ASCII characters to see if that helps?

gitonthescene:
@thadguidry - Do you know where the requirement for ASCII directory names comes from?

thadguidry:
@gitonthescene not sure in the code.  It's a core Java lib.  @tfmorris would know.

ToVie:
@thadguidry You can reproduce the behavior described by @jkjkjkjkjk like this:

### Steps to reproduce the problem
1. Fresh install of openrefine
2. Create first project in standard workspace directory and change some data
3. Create second project in standard workspace directory and change some data
4. Quit OpenRefine
5. Check the  "projectIDs"-array of workspace.json => it contains two valid project ids
6. Restart OpenRefine 
7. Delete both projects  on the UI ("Open project")
8. Quit OpenRefine
9. Check the  "projectIDs"-array of workspace.json => it still contains the id of one of the deleted projects
10. Open OpenRefine => The screen shows only "create project" and "open project"
11. Create a new project and change some data
12. Quit OpenRefine
13. Reopen OpenRefine and the UI still only shows "create project" and "open project"
14. Quit OpenRefine and reopen it => the problem persists as the id of the deleted project remains in workspace.json

### Solution
To resolve the issue I had to remove the invalid project id from the "projectIDs"-array of workspace.json

### Tested with
[...]

Issue #3277, opened 2020-10-19
Can't see projects or open Workspace Directory with non-ASCII characters in workspace path


fhc1958:
Hi,
I have installed OpenRefine 3.3. in Windows10 and open with Chrome
Everything works ok, but I can't see Created Projects nor open the Browse directory
Also, I can't import projects.
![image](https://user-images.githubusercontent.com/73102849/96425213-14016a80-11fc-11eb-9dcd-d5cb55f734ad.png)
![image](https://user-images.githubusercontent.com/73102849/96425451-53c85200-11fc-11eb-8681-7a400882741c.png)

Many thanks

tfmorris:
Sorry to hear that you're having trouble. Were you able to figure out what this issue was?

I'd recommend using the latest release 3.4.1 since it has all the latest features and bug fixes, but we have many users successfully using OpenRefine 3.3 and 3.4.1 on Windows 10, so I suspect this is something specific to your local setup.

thadguidry:
@fhc1958 This might be because your workspace directory contains non-ASCII characters, but I'm not certain.  One thing you could do is make a new folder using ASCII characters only... such as just `openrefine_workspace` and then point OpenRefine to use that folder.

You can change this 2 ways:
* by adding this line to the file `openrefine.l4j.ini` and specifying your desired drive and folder path:

  `-Drefine.data_dir=D:\MyDesiredFolder`

If your folder path has spaces, use neutral quotation marks around it:

  `-Drefine.data_dir="D:\My Desired Folder"`

* when you run OpenRefine from the terminal, by pointing to the workspace directory through the -d parameter:

`./refine -p 3333 -i 0.0.0.0 -m 6000M -d /My/Desired/Folder`

If the folder does not exist, OpenRefine will create it.

fhc1958:
Hi Thad
Many thanks.
You're right the problem was in the path of my workspace. I have followed
your instructions and now is working

Muchísimas gracias

thadguidry:
@fhc1958 You are welcome!  We'll close this issue now.

devyzr:
Had a similar issue on Linux, using Open Refine 3.5.2 on Ubuntu 20.04.03. There were no non-ascii in my workspace directory, it is  /home/coconut/.local/share/openrefine. I didn't have any non-ascii characters in my title either, only alpha-numeric and lower case characters. I had to delete my entire project folder and then restart open refine and my browser to get this to stop happening.

PR #4796, opened 2022-04-23
(I#1418 I#3719 I#3277) Fix projectID remains in workspace.json after deleting a project

### LINKED ISSUE #1418: id of deleted project will be kept in workspace.json
I have a project that makes use of another project (as mapping table). After the processing (i.e. application of the mapping rules (JSON history) + exporting the result), I'm deleting both projects. Only the id of the main project will be kept in workspace.json. The auxiliary project (with the mapping table) disappears as expected. When I restart OpenRefine, it tries to recover the project whose id still has been left in the workspace.json. When I try to reimport both projects and do the processing afterwards, then the mapping table won't be applied. So I guess it has something to do with the incomplete removal (and the unsuccessful recovery after OpenRefine restart) of the main project.
My current workaround is that I'll give the main project each time a new name at import.
  
tested with OpenRefine 2.7 and 2.8

WaltonG:
Fixes #1418 #3719 #3277 

- When there is no project in the workspace after deleting the only existing project, set the projectIDs property in the workspace.json file to empty.


antoine2711:
@wetneb: I'll leave it to you to review once @WaltonG is ready.

WaltonG:
The PR is ready for review

wetneb:
Very nice! It looks like you have found the cause of the bug indeed :)
One thing I wonder about, though, is that with this change we will be saving the workspace over and over as long as it stays empty, which is perhaps not a big deal (saving an empty workspace does not cost much) but is perhaps not so elegant. More generally, when reading this `saveNeeded()` method, it might not be immediately obvious to another developer why an empty workspace always needs saving (so the check is likely to be removed by someone else in the future).
To avoid that, I would propose two things:
* Perhaps introduce a `projectRemoved` boolean flag, which would be set to true when we delete a project, and reset to false when the workspace is saved. This would have two benefits: avoid the repeated saving of empty workspaces, and also cater for the case where a project was removed but the workspace is not empty either
* Add a unit test (with a reference to the GitHub issues mentioned in your PR) to demonstrate the behaviour

WaltonG:
Sure will work on the proposals

WaltonG:
Unit test is needed for the integration test to pass. Working on it.

WaltonG:
> Very nice! It looks like you have found the cause of the bug indeed :) One thing I wonder about, though, is that with this change we will be saving the workspace over and over as long as it stays empty, which is perhaps not a big deal (saving an empty workspace does not cost much) but is perhaps not so elegant. More generally, when reading this `saveNeeded()` method, it might not be immediately obvious to another developer why an empty workspace always needs saving (so the check is likely to be removed by someone else in the future). To avoid that, I would propose two things:
> 
> * Perhaps introduce a `projectRemoved` boolean flag, which would be set to true when we delete a project, and reset to false when the workspace is saved. This would have two benefits: avoid the repeated saving of empty workspaces, and also cater for the case where a project was removed but the workspace is not empty either
> * Add a unit test (with a reference to the GitHub issues mentioned in your PR) to demonstrate the behaviour

The unit test seems quite complicated as it involves quite a number of methods, creating a project, saving the project, deleting the project, reading the workspace file. Let me see how it will workout.

main/src/com/google/refine/io/FileProjectManager.java
```diff
@@ -288,15 +290,20 @@ protected void saveWorkspace() {
                 file.renameTo(oldFile);
             }
 
+            if (oldFile.exists() && _projects.isEmpty()) {
+                oldFile.delete();
+            }
```
wetneb, line 295: It is not clear to me why we need to delete the old file in this case, can you explain a bit more?

WaltonG: When a workspace is saved after removing a projectID its previous state is stored in wospace.old.json. When the only existing project is removed and the workspace is saved, workspace.old.json while retain the deleted projectID. 
On restarting the server, an empty project directory for the projectID  in workspace.old.json is created and this brings back the bug we are fixing.

wetneb: But why is the empty project directory leftover? Should we not rather delete that directory instead of deleting `workspace.old.json`?

WaltonG: The project directory is actually deleted as expected but when the server is restarted projects are loaded from :https://github.com/OpenRefine/OpenRefine/blob/1cb874819284ae21a43117decc5f10297a6092a7/main/src/com/google/refine/io/FileProjectManager.java#L361-L372

Since `workspace.old.json` has a projectID and there is no existing project directory for it an empty project directory is created

main/tests/server/src/com/google/refine/io/FileProjectManagerTests.java
```diff
@@ -92,4 +102,28 @@ public void saveReloadMultinationalCharacter() throws IOException {
...
+        manager.saveWorkspace();
+        System.out.println(workspaceFile.getPath());
+        System.out.println(ParsingUtilities.inputStreamToString(f));
+
+        manager = new FileProjectManagerStub(workspaceFile);
+        manager.deleteProject(5555);
+        System.out.println(workspaceFile.getPath());
+        System.out.println(ParsingUtilities.inputStreamToString(f));
+    }
 }
```
WaltonG, line 129: @elroykanye Can I get some assistance here. Seems am stuck.

main/src/com/google/refine/io/FileProjectManager.java
wetneb: Ah, then that sounds like the `loadFromFile` method incorrectly considers that loading `workspace.json` failed if it did not load any project: it should instead recognize that it successfully loaded an empty workspace, no?

WaltonG: Correct. That should be the case

main/tests/server/src/com/google/refine/io/FileProjectManagerTests.java
elroykanye: Hi @WaltonG , let me check this out and give you feedback.

elroykanye: Similar to how you created a temporary workspace file, I think you should do same for the project file and call this `manager.saveToFile()` for it.

Am I on track with you?

WaltonG: Sure. Let me look into that

main/src/com/google/refine/io/FileProjectManager.java
wetneb: Hi @WaltonG, is this still needed after your changes then?

WaltonG: Hi @wetneb I don't think it's needed. Let me confirm

WaltonG: Hi @wetneb, I have confirmed that loading of projects from workspace.temp.json and workspace.old.json is not required

wetneb: Hmm, that is curious! Intuitively, this fallback mechanism should still be useful if `workspace.json` disappeared (after a failed save, perhaps).
I would expect that this code to load from `workspace.temp.json` and `workspace.old.json` stays, but the return value of `loadWorkspace` is not whether a project was found or not, but rather whether the loading of the workspace succeeded or not (which is different, since it is possible to successfully load an empty workspace). This is what I have tried to explain in my comment above - is this new comment any clearer?

WaltonG: Sure the comment is clearer and a fallback mechanism is actually needed. With that in mind am rechecking if deleting `workspace.old.json`  is still needed.

WaltonG: > Hmm, that is curious! Intuitively, this fallback mechanism should still be useful if `workspace.json` disappeared (after a failed save, perhaps).

Considering this scenario, when a project directory is deleted but saving the workspace  failed then the fallback mechanism will reference a non existing project directory. In the following code is it ok to move `saveWorkspace()` to the synchronized block

https://github.com/OpenRefine/OpenRefine/blob/f151be913b5d7de25b312a48300080190e68a605/main/src/com/google/refine/io/FileProjectManager.java#L327-L338

main/tests/server/src/com/google/refine/io/FileProjectManagerTests.java
```diff
@@ -92,4 +101,35 @@ public void saveReloadMultinationalCharacter() throws IOException {
...
+     */
+    @Test
+    public void deleteProjectWithoutSavingWorkspace() throws IOException {
+        FileProjectManager manager = new FileProjectManagerStub(workspaceDir);
+        manager.saveToFile(workspaceFile);
+        manager.deleteProject(5555);
+
+        InputStream inputStream = new FileInputStream(workspaceFile);
+        JsonObject json = JSON.parse(inputStream);
+        assertFalse(json.get("projectIDs").getAsArray().isEmpty());
```
wetneb, line 117: I am not sure about this test: your comment above says "deleting the only existing project without saving the workspace will *not* remove the projectID from workspace.json" but in the test you do assert that the array is empty… That seems inconsistent, no? I think the test below should actually be sufficient, no?

WaltonG: > but in the test you do assert that the array is empty… That seems inconsistent, no?

That is not the case. The test uses `assertFalse`  on `isEmpty()` which confirms that the array is not empty

> I think the test below should actually be sufficient, no?

Thats true since it confirms that deleting and saving the workspace removes projectID from wokspace.json thus fixing the issues

wetneb: Ah thanks, looks like I was not really awake :-D
Sounds good, then let's only keep the second test.

PR #4798, opened 2022-04-23
(I #1286) Add a "display unprintable chars" toggle mode

### LINKED ISSUE #1286: Add a DISPLAY UNPRINTABLE CHARS toggle mode

shriyasankhyan:
Fixes #1286

- Control characters will be shown in the parsing panel by default.
- A 'settings' option will be added in place of show nulls which after clicking, shows us a dialog box from which we can toggle nulls as well as control characters.


shriyasankhyan:
I have reproduced a dialog box when clicked on All->View->Settings from where we can show/hide control characters and nulls. But when I am trying to show control characters in the parsing panel, it is not failing after I am adding the function made in parsingpanel.js but failing when I am calling the function in previewtable.js. Can you please point to any resource or any advice that you can give so that I can solve this problem? 
Thank you! I shall be grateful to you if you help :)

shriyasankhyan:
Whenever I am doing  $('<span>').html(showNonPrintableChars(cell.v)).appendTo(divContent); 
, my code is failing due to cypress error.
I checked the tests on my local system and it pointed me towards 
https://www.slf4j.org/codes.html#StaticLoggerBinder but still I am not able to understand in which direction I should work for fixing the parsing panel. I would be grateful if you help me with the same. @wetneb @thadguidry

wetneb:
Hello @shriyasankhyan, sorry for the delay in replying to this.
The link "https://www.slf4j.org/codes.html#StaticLoggerBinder" is about Java logging so it is unlikely that it is related to your problem, which seems to be on the frontend side. I would recommend to check which Cypress test faiils when you use this particular code, and then run this Cypress test locally. This will give you the opportunity to step through the stages of the test and understand which undesirable behaviour this code is introducing.

shriyasankhyan:
> Hello @shriyasankhyan, sorry for the delay in replying to this. The link "https://www.slf4j.org/codes.html#StaticLoggerBinder" is about Java logging so it is unlikely that it is related to your problem, which seems to be on the frontend side. I would recommend to check which Cypress test faiils when you use this particular code, and then run this Cypress test locally. This will give you the opportunity to step through the stages of the test and understand which undesirable behaviour this code is introducing.

Ok thanks for the reply :) I shall check it in this way.

Issue #4808, opened 2022-04-26
Standardize ordering of Java modifiers


sherrif10:
This is issue is very low and the main idea is to emphasis clean code. i know we might overlook but its much vital in organisation code of conduct  cc @antoine2711  @wetneb

wetneb:
I would not do such a normalization by hand without enforcing it in our linter in the CI.

sherrif10:
Thanks for quick response. Do you mind helping me enforce it in or linter in the ci. We would definately enforce it.

wetneb:
Sorry I do not have time to work on this and it is not clear to me why we would need this normalization.

sherrif10:
The idea to me is to have a standadized code of conduct. Am thinking in the next 10 years from now if this code is not having a clear code convention, it would not be appropriate. That's my kind of thinking.

antoine2711:
> The idea to me is to have a standadized code of conduct. Am thinking in the next 10 years from now if this code is not having a clear code convention, it would not be appropriate. That's my kind of thinking.

@sherrif10: I think that way you are raising here, are kind of issues that are not simple, nor short term planification. It involves so many variables, and also, which is very costly, a LOT of community discussions. And just that, while we are in the middle of the Outreachy project here at OR, it's an overwhelming tasks. So, if you really care about this kind of issue, I suggest that you come back in a month with it. And prepare yourself for a few weeks of discussions. This is not, unfortunatelty, a trivial task! ;-)

To understand the timing, you have to take in consideration that the quantity of task on Antonin's list are pretty numerous. For a month, he will not have time to discussion this kind of issue.

Regards, Antoine

tfmorris:
The Checkstyle static checker could be used to check/enforce this. https://checkstyle.sourceforge.io/config_modifier.html#ModifierOrder

I'm not sure it's worth the churn in the code base, but if it ever were to be done, it'd be good to combine with as many other normalizations (e.g. import ordering, etc) as can be reasonably bundled together.

wetneb:
I wonder if our existing linter would be able to enforce that too (it is based on Eclipse style files). If that is the case, then all it would require is a change in the configuration file and re-linting: the integration in the CI is already there.

wetneb:
Given that there is not a huge appetite in tackling this specific linting change, I am closing this.

PR #4810, opened 2022-04-26
Change static final private to Private static final

sherrif10:
Fixes [#{issue number here}](4808)

Issue  https://github.com/OpenRefine/OpenRefine/issues/4808

- We can configure use of any specific linters for tracking these improvements. Could we use this as a starting point for using linters in this code base cc @antoine2711  @wetneb 

YOu can only see these issue by configuring any linter in IDE Forexample  linter in vscode
- 
-


antoine2711:
If you want to move forward now with this issue, the first phase will really be some documentation. Here some ideas of what could be put in it:
- What rules would be enforced
- What changes to the code would that imply
- Our would it affect code comparaison between code revisions?
- What OS setup would have problem installing a compatible linter tool?
- Do a test yourself on the 3 platforms, or at least 2
- ...

The point of this is to mesure the impact such a change will have on the community and the developers. This should always be your main concern (with the users, of course), and it this case, it is a critical change with many implications.

Regards, Antoine

sherrif10:
Thanks fo the detailed explanation. This is how we should handle this. Probably will add bit of documents thats regards to this work.

AtesComp:
See my comment in issue #4808.

wetneb:
Closing per https://github.com/OpenRefine/OpenRefine/pull/4844#issuecomment-1137442589

PR #4816, opened 2022-04-28
#3515 notify user about disk space availability 

### LINKED ISSUE #3515: Notify user about disk space availability
When a certain volume of data is loaded in OpenRefine is reached, it stops loading the files. This is solved by deleting some files from the history.

### Proposed solution
Generate an available space notification or indicator

elroykanye:
Fixes #3515 

- addition to pom deps to include oshi-core from [here](https://github.com/oshi/oshi).
- a suggestive action area for system information
- command for retrieving system information 
- an ajax request that communicates with the servlet for the command stated above.


elroykanye:
Hi @antoine2711 @wetneb , I wish to ask for your guidance on this.

The approach I followed to come up with a solution for this, was to setup a `GetSystemInfoCommand` which has a `doGet` method for responding to requests which require the system information.

Using the oshi tool suggested in [this](https://groups.google.com/g/openrefine-dev/c/eJbPPYnqJiw/m/_iQqCOdbAwAJ), I was able to return a serialised object containing info like the RAM, hostname, etc.

On the frontend, I used ajax to request for this information from the command stated above. This information has to be updated so I set a 1000ms interval using the `setInterval` function for repeatedly requesting the data.

All this was done on an isolated html + js file so I can see the results without affecting much of the codebase.

My problem is that the logs are almost completely filled with the calling of the command I created, and integrating it into an open project may create some problems. Is there some better you think I could use to maybe subscribe to the command so I get the data realtime or avoid the logging?

elroykanye:
@wetneb @antoine2711 
From the discussion yesterday, I went through the `RefineServlet` again and realised the commands are already checked for the value returned by `logRequests()`. I just updated the new command then by overriding the default `logRequests()` to return false.

This solves the issue I had previously.

main/webapp/modules/core/scripts/project.js
```diff
@@ -591,6 +591,45 @@ Refine.getPermanentLink = function() {
...
+        systemInfo.availableMemory = data["available_memory"];
+        systemInfo.totalMemory = data["total_memory"];
+      }
+    }).fail(function(jqXhr, textStatus, errorThrown ) {
+      alert( textStatus + ':' + errorThrown );
+      clearInterval(intervalId);
+    });
+
+    if(systemInfo.memoryWarn()) {
+      alert(`Your memory is running out\n`)
```
antoine2711, line 628:
```suggestion
      alert(`Your memory is running out.`+ this.availableMemory +` is left.\n`)
```
What do you think? Maybe even the **out of**?

main/webapp/modules/core/scripts/project.js
```diff
@@ -591,6 +591,45 @@ Refine.getPermanentLink = function() {
...
+        systemInfo.availableMemory = data["available_memory"];
+        systemInfo.totalMemory = data["total_memory"];
+      }
+    }).fail(function(jqXhr, textStatus, errorThrown ) {
+      alert( textStatus + ':' + errorThrown );
+      clearInterval(intervalId);
+    });
+
+    if(systemInfo.memoryWarn()) {
+      alert(`Your memory is running out. ${systemInfo.byteToMB(systemInfo.availableMemory)}MB is left.\n`)
```
wetneb, line 628: A few thoughts here:
* the message should be translatable with i18n
* A javascript alert is perhaps a bit too violent for this. Maybe a non-modal notification (like those displayed when an operation is executed) could work too?
* The message is not very actionable: what sort of action should we suggest to the user? Increase the amount of memory allocated to OpenRefine? Close some projects and let OpenRefine unload them? I am not sure we have a workable workflow for that, that's probably out of scope for this issue.

wetneb:
@elroykanye the ball is in your camp on this PR too :)

main/webapp/modules/core/scripts/project.js
elroykanye: I think "Increase the amount of memory allocated to OpenRefine" is a good fit for now

wetneb: Makes sense. Ideally we should add a link to the documentation to explain how, but that is perhaps not so easy in an "alert" dialog. Perhaps that is another reason to use something less blocking, in the DOM (see above)?

elroykanye: Is this the right link? https://docs.openrefine.org/manual/running#starting-with-modifications

main/webapp/modules/core/scripts/project.js
```diff
@@ -591,6 +591,45 @@ Refine.getPermanentLink = function() {
...
+    }).fail(function(jqXhr, textStatus, errorThrown ) {
+      alert( textStatus + ':' + errorThrown );
+      clearInterval(intervalId);
+    });
+
+    if(systemInfo.memoryWarn()) {
+      const msg = $.i18n('core-project/memory-warn');
+      var notification = new ProcessNotification(msg, 10000);
+      notification.display();
+    }
```
wetneb, line 629: This `if` block should rather be inside the `success` callback above, to avoid race conditions. With this code structure you would not even need to store the memory values in a global object (`systemInfo`).

main/webapp/modules/core/scripts/project/process-notification.js
```diff
@@ -0,0 +1,57 @@
...
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+ */
+
+class ProcessNotification {
```
wetneb, line 34: Is there any chance we can reuse this code in the existing places where notifications are displayed (not just the new places you are introducing)?

main/src/com/google/refine/commands/workspace/GetMemoryInfoCommand.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private static final GlobalMemory systemMemory = new SystemInfo().getHardware().getMemory();
+
+    public static class MemoryInfo {
+
+        @JsonProperty("available_memory")
+        protected Long availableMemory = systemMemory.getAvailable();
+
+        @JsonProperty("total_memory")
+        protected Long totalMemory = systemMemory.getTotal();
```
wetneb, line 24: I am a bit suspicious about those calls. Given their description it seems that they refer to the total memory available and used by the system. This is not what we should be watching out for: rather, we should check how much memory our JVM is using, and how much was allocated to it.

I am hopeful that this information can be accessed in pure Java, without the need for a dependency like `oshi`.

elroykanye: Yes, it can

thadguidry:
@elroykanye @elroykanye Where are we at on this PR? Almost ready to merge? Ready?  Needs heavy testing by all on all platforms?

wetneb:
Closing per inactivity.

PR #4834, opened 2022-05-07
[#3243] Ignore JSON comments

### LINKED ISSUE #3243: Allow option for comments in Json files to be ignored during import
Json comments are not supported with the Json Importer and it would be nice if an option was added.

### Proposed solution
Add checkbox option to Json Importer to **Allow  comments in json files**
When selected, `/*...*/` blocks and the line comments `//`, and `#` are interpreted as comments and get ignored (instead of causing silent errors within Jackson importer, I think)

The following would be parsed with the Json Importer and when the option is checked, it will ignore all the comment lines and parse the remaining Json object without errors.  If unchecked, then it would parse and detect an error and advise user onscreen somehow (popup, or directly within the preview panel as an error text string, whatever works best to notify user at this current architecture level of OpenRefine).

```json
/* These are a block of
line comments for this JSON file and
should not cause errors ...
*/
/* This is also a block */
// This line is also ignored instead of causing errors
# This line is also interpreted as comments and gets ignored instead of causing errors
[{
  "foo":"bar"
}]
```

### Alternatives considered
keep silent as we currently do without any clue to the user why they are staring at no data on the preview and no errors or sign of something happening.

### Additional context
Jackson supports this option https://github.com/FasterXML/jackson-core/wiki/JsonParser-Features#support-for-non-standard-data-format-constructs

> ALLOW_COMMENTS (default: false) (for textual formats with concept of comments)
>
>    For textual formats that do not have official comments, but for which "de facto" conventions exist (like JSON), determines whether use of such unofficial comments is allowed or not
>    Supported for: JSON
>    For JSON: enabling the feature allows recognition and handling of "C comments" (/* ... */) and "C++ comments" (// ....)
>    YAML supports its own "hash comments" regardless of this setting

SeliasB:
Fixes #3243

- added the possibility to ingore comments if in JSON file imported
- the idea is to check if the checkbox is checked, if so, enable ALLOWS_COMMENTS and ALLOW_COMMENTS_YAML from JasonFactory method
- skippable comments are { // ... , # ... , /* ... */  } format


wetneb (commented):
I wonder about two things:
* as a user, is there any case in which I would want to turn off this option? When is it important that `ignoreComments == false`? Intuitively, I feel like this option could be always left to true - and therefore not be exposed in the UI at all.
* are there other features of `JsonFactory` that make the JSON parsing more forgiving, and if so, would it also make sense to turn them on systematically?

SeliasB:
@wetneb 

About setting the ignore comment to true, I feel like it could be always setted to skip comments. on another hand, I can double check, but I have noticed that if I set the ignore comments to true by default, the importer creates new empty rows on the viewer, so it probably creates unmanaged artifacts. 
two possible reasons: I enabled comments on the parser as it has the same behaviour as the Jsonfactory or :
The way it preset at the importation and during the project creation is slightly different so it creates artifacts

I will try to set it true by default and double check if I can manage to have consistent behaviour on the viewer. if so, yes it is relevant to set it to true by default

wetneb:
Sounds like a great plan, thank you!

wetneb:
Hi @SeliasB, are you still working on this?

SeliasB:
@wetneb Sorry, I've been in a rush,
should finish the work by the end of june !

wetneb:
Hi @SeliasB, are you still working on this?

wetneb:
Closing per inactivity, but happy to reopen if activity resumes.

thadguidry:
> Thank you for this pull request! I wonder about two things:
> 
> * as a user, is there any case in which I would want to turn off this option? When is it important that `ignoreComments == false`? Intuitively, I feel like this option could be always left to true - and therefore not be exposed in the UI at all.
> * are there other features of `JsonFactory` that make the JSON parsing more forgiving, and if so, would it also make sense to turn them on systematically?

The use case of having an option was so that users could edit and use the data that is sometimes inside the comment lines.  Using facets they can filter out the first X rows, and also copy/paste metadata from within comment lines into actual real row cell values as necessary.  The comment lines sometimes hold partial schema annotations and metadata.  That's why the option to keep the comment lines is useful, thereby letting the users data wrangle with the lines themselves and create a full project for export that includes the incorporated metadata into JSON record row headers and then export.

@wetneb **Please make this a checkbox option with default OFF.  We want to continue to give choices to users, not make them for them.  "Smart defaults with options - always."  Thanks.**

wetneb:
I don't understand how turning this off would let people extract metadata from JSON comments. If we make this option, then:
* when the option is on, JSON files with comments would be successfully imported (ignoring the comments)
* when the option is off, JSON files with comments would be rejected

I suspect you have another option in mind (which could be indeed interesting to have): one that would turn on importing the contents of those comments in the OpenRefine project.
The challenge to support that are that:
* I don't know if our JSON parser makes it possible to parse JSON files with comments and inspect the contents of such comments
* We would need to decide how those comments should be inserted in the table generated by our tree to table machinery. Since those comments can probably show up anywhere, I can imagine it being a bit messy

thadguidry:
@wetneb Sorry!  I guess a picture is worth a thousand words.  Here is what an example of what I was thinking for that JsonFactory might write out where it would be useful to know from a facet perspective which ones are lines and which ones are blocks (instead of just all the comment lines using the same key `comment:`, so I thought it was evident to just output as key:value objects like so:

```json
{
    "commentline1": "a single line // of some metadata comment",
    "commentline2": "another single line // of metadata comments",
    "commentblock3": "a block /* ... */ of schema annotations that is really just a very very very very very long paragraph inside that block",
    "records": [
        {
            "id": 1,
            "type": "String",
            "description": "a textual concept made of characters"
        },
        {
            "id": 2,
            "type": "Integer",
            "description": "a base10 numeric type"
        }
    ]
}
```

Happy to help explain more, but perhaps you see the idea now...and the rub when it comes to Jackson parser and writer handling, but it's doable I've been told.

thadguidry:
@wetneb the comments don't show up just anywhere, they are always at the beginning of the JSON files.  I've NEVER seen them at the bottom or in the middle of records, all of which would completely break JSON formatting anyways and that doesn't seem to be supported by Jackson parser (looks like it only knows about parsing out comment lines at the beginning of JSON files?

thadguidry:
ugh this is the PR !
@wetneb Perhaps I should delete those above comments from me and put all this explanation into the Issue #3243 ?

PR #4844, opened 2022-05-12
Change static private final To Private static final #4808

### LINKED ISSUE #4808: Standardize ordering of Java modifiers

shihlei:
Fixes #4808

- Change the remaining "Static private final" to "Private static final" in all the java files.

I didn't change the "Static private" to "private static" because the title of the issue only mentioned "Change static private final To Private static final". I'm not sure if I should also change all the other "Static private", please update the description of the issue so that everyone can know the main idea of the issue is to have a good coding convention.

### DIFF
--- a/extensions/database/src/com/google/refine/extension/database/mysql/MySQLDatabaseService.java
+++ b/extensions/database/src/com/google/refine/extension/database/mysql/MySQLDatabaseService.java
@@ -115,6 +115,8 @@ public class MySQLDatabaseService extends DatabaseService {
                 DatabaseInfo dbInfo = new DatabaseInfo();
                 dbInfo.setColumns(columns);
                 dbInfo.setRows(rows);
+                queryResult.close();
+                statement.close();
                 return dbInfo;
         } catch (SQLException e) {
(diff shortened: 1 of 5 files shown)


wetneb (changes requested):
As mentioned in the issue, I would be reluctant to do this sort of normalization without enforcing it in the CI. If this is supported by our linter, this could perhaps be configured there? Also, the PR contains unrelated changes.

AtesComp:
Since the "static" keyword is a class bound indicator (vs instance) and "private" applies to both class and instance (as well as "final'), my preference is to ALWAYS set "static" first and group these together to separate the class variables and function from the instance variables and function.

IMHO: It's a handy reminder when it's first.  When not, it's somewhat hidden / over looked and can cause a bit of confusion.

wetneb:
So this is rather a matter of taste. Given that the CI configuration has not been adapted, I would therefore close this.

PR #4845, opened 2022-05-12
Unclosed Statement and ResultSet Resources #4839

### LINKED ISSUE #4839: Unclosed Statement and ResultSet Resources
After working on issue #4080, we reliazed that Statement and ResultSet resources are not properly closed inside any of the DatabaseService classes. 

A Pull Request has already been made by us for the MariaDBDatabaseService class, but the same issue seems to appear on MySQLDatabaseService, SQLiteDatabaseService, etc.

### Current Results

Statement and ResultSet instances are not handled properly after they are created

### Expected Behavior

Statement and ResultSet resources are closed in a proper way (try-with-resources / finally-block)

### Screenshots

image
image

shihlei:
Fixes #4839

- Closed Statement and ResultSet objects properly.


wetneb (changes requested):
How about using a try-with-resources statement instead, as in https://github.com/OpenRefine/OpenRefine/commit/bbd7c955975557c2c5b00db44e0fdb2da82f98bf?

wetneb:
The issue was closed by #4839

PR #4848, opened 2022-05-12
configuration.md Update Wikimedia Commons manifest to version 2.1

VDK:
"reconciliation" parameter was required by OR, copied to doc

### DIFF
--- a/docs/docs/manual/wikibase/configuration.md
+++ b/docs/docs/manual/wikibase/configuration.md
@@ -27,8 +27,9 @@ To work with OpenRefine, your Wikibase instance needs an associated reconciliati
 Here is the manifest of Wikimedia Commons:
 
 ```json
+
 {
-  "version": "2.0",
+  "version": "2.1",
   "mediawiki": {
     "name": "Wikimedia Commons",
     "root": "https://commons.wikimedia.org/wiki/",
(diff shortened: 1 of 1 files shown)


docs/docs/manual/wikibase/configuration.md
```diff
@@ -73,6 +72,9 @@ Here is the manifest of Wikimedia Commons:
   },
   "editgroups": {
     "url_schema": "([[:toollabs:editgroups-commons/b/OR/${batch_id}|details]])"
+  },
+    "reconciliation": {
+    "endpoint": "https://commonsreconcile.toolforge.org/${lang}/api"
```
wetneb, line 77: This should not be needed: the reconciliation endpoints are now configured on a per-entity type basis, in the section above.

wetneb (commented):
Thank you! Let me try to understand what prompted you to suggest those changes. Could it be that you tried to use the Commons manifest with OpenRefine 3.5?

docs/docs/manual/wikibase/configuration.md
VDK: OpenRefine gives you an error if its missing.

wetneb: ah yes, that is because this manifest is not compatible with version 3.5.2 :)

VDK: Well 3.5.2. refers to this repository for manifests and it would fix the error

wetneb: Maybe it would let you add the manifest but then you would likely run into problems when you try to make edits, because OpenRefine 3.5 does not support editing on Wikibase instances which have federation enabled… it could potentially work for very restricted sorts of edits though.

I do agree that we should display a clearer error message, such as "This manifest version is not supported by OpenRefine 3.5.2", or something like that!

wetneb:
Closing this per the above.

PR #4850, opened 2022-05-12
Update .gitignore

antoine2711:
- Put main/webapp/package-lock.json in the ignored files because it is local to a user.

@wetneb: is this a good move?

Regards, Antoine

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ logs
 build/
 dist/
 server/classes/
+main/webapp/package-lock.json
 main/webapp/WEB-INF/classes/
 main/tests/server/classes/
 main/test-output/


wetneb:
I think it should indeed either be in `.gitignore` or be checked in. I think the recommended practice is to add it to the git repository. @elebitzero do you agree?

antoine2711:
> I think it should indeed either be in `.gitignore` or be checked in. I think the recommended practice is to add it to the git repository. @elebitzero do you agree?

@wetneb: isn't it a file that records what's is installed on a machine, a as such, recreated on each machine? If if some modules are installed globally, the file will not be the same, if I understand correctly. Why not let each user created to tailor his needs?

Regards, Antoine

wetneb:
According to this, the recommended practice seems to be to commit the file to the repository: https://nodejs.dev/learn/the-package-lock-json-file

wetneb:
@elebitzero can you confirm that the current situation is fine (so we should close this PR without merging)?

wetneb:
(sorry, I closed it by mistake - the intention was to wait for @elebitzero's feedback)

wetneb:
Given that it is officially recommended to check it in per the link above, let us close this.

PR #4869, opened 2022-05-19
Re-fetch Service Manifest when selecting a service

### LINKED ISSUE #3645: Double check the reconciliation manifest each time reconciliation is fired.

### Proposed solution

### Additional context

v1scal:
Fixes #3645

- Re-fetch Service Manifest when selecting a service
- Update the Manifest in the Preferences

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
+++ b/main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
@@ -174,7 +174,7 @@ ReconDialog.prototype._selectService = function(record) {
             oldRecord.handler.deactivate();
           }
         }
-
+        console.log("buenos dias")
         this._elmts.servicePanelMessage.hide();
 
         record.selector.addClass("selected");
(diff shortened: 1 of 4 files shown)


wetneb (commented):
Thanks! I haven't tried it yet, but I noticed two things:
* the identation seems a bit off
* this is likely to fail for services which only support fetching the manifest via JSONP. It would be good to preserve support for that. For this reason, it might make sense to encapsulate the existing logic of fetching the manifest (from the dialog to add a new service) into a function that you could reuse here.

v1scal:
Ok, I'll work on that. Thanks!

v1scal:
@wetneb I've already added the JSONP but didn't fully understand what you meant by reusing the function. Should I put the refetch in a separate method and call it or try to use the one in recon-manager.js?

wetneb:
I mean that the logic to fetch a manifest first via CORS and then via JSONP should intuitively be the same in two places:
* when adding a new service
* when checking its manifest again before using it

so it would probably make sense not to duplicate code and use a common javascript function / method to implement this fetching.

Also, there are still a lot of indentation changes so it is hard to figure out where you made actual changes. Could you revert those indentation changes?

wetneb:

wetneb:
Closing per inactivity, but happy to reopen if activity resumes.

PR #4887, opened 2022-05-25
#4249 general date parsing similar to strtotime

### LINKED ISSUE #4249: General date parsing similar to strtotime.
It would be easier if OpenRefine was able to parse any date without giving the structure. PHP has a function called [strtotime](https://www.php.net/manual/en/function.strtotime.php) which works quite generally for parsing date found in many ways of styling.

### Proposed solution
implement a date parser, not sure exactly how. 

### Additional context
To convert a date now I have to put something like `value.replace("th","").toDate('MMM d, y').toString('yyyy-MM-dd')`
October 4th, 2021	=>	2021-10-04
Easier would be value.parseDate().toString('yyyy-MM-dd')

PS. thanks for the amazing TOOL!

elroykanye:
Fixes #4249 

- a grel function for parsing dates 
- tests for this function
- updates to the function registry


ostephens:
I'm not so keen on the introduction of a new GREL function to meet this requirement. The 'toDate' GREL is already in place as a way of converting a string form of a date to a date object - I think it would be better to look at how we improve toDate than introduce a completely new function

wetneb:
Note to self: this PR adds a new dependency, so we need to review it as usual (size, license, being actively maintained, and so on)

ostephens:
> I think it will be better to use this case when no pattern for the date has been provided in toDate() since this does not require a pattern to work.

Agreed this seems like the best option. Unfortunately toDate() is a bit of a mess at this point :( (I feel I can say this as I'm responsible for some of the mess...)

The biggest issue I can see for this work is that currently using `toDate()` with no arguments is the same as `toDate(true)` where `true` brings in some assumptions about the format of the day - so you may need to be a bit careful there.

I definitely feel that 'toDate()' could be substantially improved, but I've always been slightly reluctant to pull it apart completely because of the risk of breaking backwards compatibility - but maybe we've reached the point where we need to decide that this is something we can live with. Would be interested to get comments from @thadguidry @antoine2711 @wetneb

wetneb:
I hear you! If it is difficult to work with `toDate` in a backwards compatible way, perhaps that is the case for making a new function. But I guess we could also have some backwards incompatibility in some corner cases, it's not something I would oppose if we can see a clear improvement or simplification.

ostephens:
> But providing users with a transition function could be the best of both worlds I would say since we know from recent talks, mailing list and survey's that users do have repeated workflows they use and those would need to be updated (find/replace the toDate() in the JSON). Not convenient, but I think easy enough for all our users when given the choice.

Are you suggesting that you are in favour of having two separate date parsing functions then @thadguidry ?

ostephens:
> I am in favor of improving OpenRefine's date parsing functionality overall.

Absolutely - I don't think there's any disagreement here!

> taking the time to make toDate() have slightly better Smart defaults and the functionality our users want

This is absolutely my preferred option. My concern was about backwards compatibility - as you mention:

>  users do have repeated workflows they use and those would need to be updated (find/replace the toDate() in the JSON). Not convenient, but I think easy enough for all our users when given the choice.

But it sounds like you don't have too many concerns about this if we are improving the date parsing overall.
In which case @elroykanye I'd be in favour of making this improved date parsing part of the existing `toDate()` function. 

My suggestion would be that if only a single argument is present (the string to be converted), use the new date parsing, but in the case of >1 arguments fall back on the existing options. That way any existing scripts that use `value.toDate()` can be left as they are (take advantage of the new date parsing) or converted by the user to `value.toDate(true)` if they want to maintain the existing behaviour

I think this meets @thadguidry's goals as he states above and walks a line between intuitive behaviour, backwards compatibility and (perhaps most importantly) improving OpenRefine's handling of dates

elroykanye:
Hi. Thanks for the feedback. I think I will have to take some time to carefully go through the the implementation toDate() to understand its logic before integrating this case as @ostephens described. Thanks again :)

main/src/com/google/refine/expr/functions/ToDate.java
```diff
@@ -81,52 +82,65 @@ public Object call(Properties bindings, Object[] args) {
...
             try {
-               return CalendarParser.parseAsOffsetDateTime( o1, (month_first) ? CalendarParser.MM_DD_YY : CalendarParser.DD_MM_YY);
-            } catch (CalendarParserException e) {
+               return CalendarParser.parseAsOffsetDateTime( o1, (monthFirst) ? CalendarParser.MM_DD_YY : CalendarParser.DD_MM_YY);
+            } catch (CalendarParserException ignored) {
            }
         }
-        return parse(o1,formats);
+        OffsetDateTime date = parse(o1, formats);
+        if (Boolean.FALSE.equals(monthFirst)) {
```
thadguidry, line 140: @elroykanye What test covers this fully?

elroykanye: https://github.com/OpenRefine/OpenRefine/blob/master/main/tests/server/src/com/google/refine/expr/functions/ToDateTests.java#L76-L78

It's actually what has kept me on hold for some time 😃  I noticed that formats were not in the tests when `monthFirst` was explicitly stated.

thadguidry: OK, well, I think we need to be descriptive in our tests (to cover our assumptions of formats) so that in the future, in case we make Date library changes, or the JDK itself goes crazy, then at least our test cases are described well with comments, the formats described so eliminate ambiguity for other developers in the future, and basically make OpenRefine Date parsing a bit more bulletproof with robust testing here.

Is your intention to improve the tests in a few more commits on this?

elroykanye: I did plan to improve the tests for cases where just 1 or 2 args are provided which is the case handled by the introduced fix. I thought it wise to leave the other tests as they are just so I am sure the improvements do not break any uses of toDate()

thadguidry (commented):
@elroykanye Do you see a problem here, or do I have the wrong assumption when trying Finnish dates?  `toDate("15.01.13",false)` because I didn't notice a change in the output with true or false.

Separate issue?

ostephens:
@thadguidry

ostephens:
>Do you see a problem here, or do I have the wrong assumption when trying Finnish dates? toDate("15.01.13",false) because I didn't notice a change in the output with true or false.

I don't know what the Finnish date format is, but as I understand it the boolean is meant as a "hint" to the parser not an absolute - so in this case I'd expect the parser to interpret this as 15th January 2013 whatever the setting of the boolean (you can see this behaviour currently with `toDate("15-01-13",true)` and `toDate("15-01-13",false)` which give the same output (15th Jan) vs `toDate("05-01-13",true)` and `toDate("05-01-13",false)` which give different outputs (1st May and 5th Jan respectively)

Whether this is good behaviour (treating the boolean as a hint rather than an absolute statement) I'm not sure, but the behaviour you describe seems consistent with the current behaviour.

thadguidry:
> Whether this is good behaviour (treating the boolean as a hint rather than an absolute statement) I'm not sure, but the behaviour you describe seems consistent with the current behaviour.

@ostephens Thanks for explaining the current behavior, very helpful!  Hmm, in that case should we document that the boolean is a hint in our docs? I'd vote for that.

ostephens:
@thadguidry agreed that the documentation should reflect behaviour. I don't know where I saw the boolean described as a "hint" originally but the behaviour, and a quick look at the code (which is quite complex tbh so I'm not 100%) seems to back up the idea that if the number can't be treated as a month (i.e. >12) it can end up being treated as a day instead

docs/docs/manual/grelfunctions.md
```diff
@@ -413,6 +413,10 @@ As of OpenRefine 3.4.1, uniques() reorders the array items it returns; in 3.4 be
 
 Returns the current time according to your system clock, in the [ISO 8601 extended format](exploring#data-types) (converted to UTC). For example, 10:53am (and 00 seconds) on November 26th 2020 in EST returns [date 2020-11-26T15:53:00Z].
 
+###### parseDate(s dateString) {#parsedates-dateString}
+
+Returns the inputted date as a string in any valid format to a date object in the [ISO 8601 extended format](exploring#data-types) (converted to UTC). For example, "October 4th, 2021" returns [date 2021-10-04T00:00:00Z]. Generally, date formats supported by this function are stated [here](https://github.com/sisyphsu/dateparser#showcase).
+
```
wetneb, line 419: This documentation change can be removed because you are no longer introducing this `parseDate` function.

main/pom.xml
```diff
@@ -348,12 +348,19 @@
...
       <artifactId>language-detector</artifactId>
       <version>0.6</version>
     </dependency>
+
+    <!-- date parser -->
+    <dependency>
+      <groupId>com.github.sisyphsu</groupId>
+      <artifactId>dateparser</artifactId>
+      <version>1.0.10</version>
+    </dependency>
```
wetneb, line 367: I am actually a bit worried about including this dependency. Although it is actively maintained, it does not seem to be used a lot, so it is likely that we get surprises with it: https://mvnrepository.com/artifact/com.github.sisyphsu/dateparser
Also, the license it declares on GitHub (MIT) is not the one included in the Maven Metadata (Apache license).

wetneb (commented):
I think this needs quite some tests to demonstrate the new formats that are supported with this change. Also I am not sure about the new dependency (see below).

This is a PR that (I would say) has high changes of breaking things. But if @ostephens is confident about it I am not going to stand in the way.
Otherwise, by default I would not merge it, to be on the safe side.

main/pom.xml
tfmorris: Additionally, it was implemented as a showcase for the underlying RE implementation, rather than as something driven by user requirements and it also appears to be English only for month & day names.

Another possibility might be a JNA wrapper to https://github.com/araddon/dateparse which seems to be a little bit more general and much more popular.

However, one of the things that I think to current date guesser has shown us is that something which works *most* of the time is kind of a recipe for unforeseen problems and complaints.

PR #4817, opened 2022-04-29
Updated tests for jQuery migration

elebitzero:
Changes proposed in this pull request:
- Updated tests for change preferences and project metadata in preparation for the jQuery migration.

PR #4891, opened 2022-05-25
Upgrading jQuery to 3.6.0

elebitzero:
- Upgrade jQuery and jQuery Migrate.

The jQuery migration from 1.12.4 to the latest (3.6.0) version required many fixes due to deprecated and removed methods, and also some behavior changes.

Luckily, all the work was able to be done incrementally since all of the fixes were compatible with the old version (1.12.4).

All of the Cypress tests are passing, and I have done some additional manual testing.

Summary of Changes:
* Replaced usages of jQuery functions removed in 3.6.0 or earlier. (Ref: https://api.jquery.com/category/removed/).
* Replaced usages of jQuery functions deprecated in 3.6.0 or earlier. (Ref:  https://api.jquery.com/category/deprecated/).
* Fix broken code due to jQuery no longer auto-inserting tbody tags.
* Correct <textarea/> to <textarea></textarea> becuase latest jQuery did not like self-closing <textarea/>.
* Fix deprecations in external dependencies jquery.imgareaselect.js and suggest-4_3a.js.
* Due to a change in behavior in jQuery $.post(), calls to /command/core/toggle-starred-expression would fail if a "json" is expected, but an empty response is returned. By omitting the expected content type (""), then the content-type is set by the server response. (Ref: https://github.com/jquery/jquery/issues/3973).
* Updated select2 package which fixed jQuery deprecations.
* Resolves security issues with jquery 1.12.4. (Ref: https://snyk.io/vuln/npm:jquery?lh=1.12.4):
image

Pending:
* The tablesorter dependency has an open PR to remove calls to deprecated functions. (Ref: https://github.com/Mottie/tablesorter/pull/1786). The current tablesorter does work with jQuery 3.6.0, so it's still okay to go ahead.
* May want to remove jquery-migrate.  They don't recommend you have it enabled in production.  We should just add it in dev builds when we are doing migrations from one jQuery version to the next.

Associated PRs:
* https://github.com/OpenRefine/OpenRefine/pull/4886
* https://github.com/OpenRefine/OpenRefine/pull/4882
* https://github.com/OpenRefine/OpenRefine/pull/4874
* https://github.com/OpenRefine/OpenRefine/pull/4872
* https://github.com/OpenRefine/OpenRefine/pull/4862
* https://github.com/OpenRefine/OpenRefine/pull/4861
* https://github.com/OpenRefine/OpenRefine/pull/4853
* https://github.com/OpenRefine/OpenRefine/pull/4852
* https://github.com/OpenRefine/OpenRefine/pull/4851
* https://github.com/OpenRefine/OpenRefine/pull/4840
* https://github.com/OpenRefine/OpenRefine/pull/4837
* https://github.com/OpenRefine/OpenRefine/pull/4827
* https://github.com/OpenRefine/OpenRefine/pull/4822
* https://github.com/OpenRefine/OpenRefine/pull/4817
* https://github.com/OpenRefine/OpenRefine/pull/4811

### DIFF
--- a/main/webapp/package.json
+++ b/main/webapp/package.json
@@ -7,8 +7,8 @@
   "dependencies": {
     "@wikimedia/jquery.i18n": "1.0.7",
     "datejs-coolite": "1.0.0",
-    "jquery": "1.12.4",
-    "jquery-migrate": "1.4.1",
+    "jquery": "3.6.0",
+    "jquery-migrate": "3.4.0",
     "js-cookie": "2.2.1",
     "moment": "2.29.3",
     "select2": "4.1.0-rc.0",


wetneb:
Very nice, thank you for the amazing work!
In terms of schedule, I wonder if we want to have this in 3.6. This is going to be a breaking change for some extensions, I suspect. I guess we could decide to put this in 4.0 only and keep the 3.x series as compatible as possible. But then jQuery 1.12.4 does have some vulnerabilities so I guess there is also a case for migrating it in 3.x directly.

elebitzero:
Is OpenRefine 3.6 going to be released soon? 

In the OpenRefine UI code, I only found 10 actual issues which broke either the code or the cypress tests, which resulted in 53 tests failing.  In some cases, one bug broke multiple tests.  

A lot of the changes were migrating deprecated functions, but of course that isn't required.  JQuery has deprecated many functions, but they have been very slow to remove them. If you look at the list, they have only removed functions in early jQuery 1.3-1.10.   (Ref: https://api.jquery.com/category/removed/).  There were only a handful of places OpenRefine was using removed jQuery functions.  

For extensions, I would expect there would be few issues. If there are popular extensions outside of this repo, link them here and I can try them out with OpenRefine 3.6 and jQuery 3.6.00 to see if they work.

And yes, the upgrade does resolve three security issues. Not sure how critical that is if OpenRefine is mainly used locally or within a private network. I've added info to the PR for that.

wetneb:
Yes I would like to release 3.6 soon.
We have a list of popular extensions here: https://openrefine.org/download.html

elebitzero:
I looked at the list of extensions but most were not active and haven't been updated to work with OpenRefine 3.5.x.

If we postpone upgrading jQuery, that means any new extensions on 3.6.x will be using jQuery 1.12.4, and may break when jQuery gets upgraded in the next OpenRefine version.

wetneb (approved):
Merging and adding to 3.6 since no concerns have been raised.

PR #4892, opened 2022-05-26
Add support of Mardown Exporter

Unnamed-1408:
Fixes #{2532}

- Add Markdown options to Export menu
- Add Markdown exporter
- Add Markdown exporter test file
- Pass cypress (failed in the last pr because `actions/setup-node@v3.2.0` change)


wetneb (commented):
Thanks a lot for fixing those tests!
Unrelatedly, I just thought: what about reconciled cells? Perhaps it would be nice if cells that are matched get rendered as links, with the name of the matched item as link text?

Unnamed-1408:
> Thanks a lot for fixing those tests! Unrelatedly, I just thought: what about reconciled cells? Perhaps it would be nice if cells that are matched get rendered as links, with the name of the matched item as link text?

OK, I will try to solve it, set it to draft temporarily

Unnamed-1408:
> Thanks a lot for fixing those tests! Unrelatedly, I just thought: what about reconciled cells? Perhaps it would be nice if cells that are matched get rendered as links, with the name of the matched item as link text?

The reconciled cells can rendered as link now, and I add the special characters I found that can't be displayed directly in the markdown table (if anyone finds other special characters, they can be added directly to the regrex expression)
image

main/src/com/google/refine/exporters/MdExporter.java
```diff
@@ -0,0 +1,64 @@
...
+
+            @Override
+            public void addRow(List<CellData> cells, boolean isHeader) {
+                String[] cellData = new String[cells.size()];
+                for(int i = 0; i < cells.size(); i++){
+                    CellData cellDatatmp = cells.get(i);
+                    cellData[i] = "";
+                    if(cellDatatmp != null && cellDatatmp.text != null){
+                        String textInCell = cellDatatmp.text;
+                        textInCell = textInCell.replaceAll("([`|])", "\\\\$1");
```
wetneb, line 44: I am surprised that this is not handled directly by the `net.steppschuh.markdowngenerator` library: this ought to be done there natively, no? Perhaps it would be worth reporting this to them.

Unnamed-1408: This library doesn't handle the situation that `|` in the markdown table, I will report it to them (but I don't think he will handle it quickly, cause the last update was a year ago)

wetneb: I suspect we might need to escape more characters here (for instance, the ones that are used to format text as bold or italics).

wetneb:
Given this observation I am worried about the quality of this dependency. Escaping text properly is one of the main features such a library should provide. The ability to control alignment and padding of cells so that the text representation is nice makes sense, but I would say it does not justify pulling in such a weak dependency for this.
Is there any alternative available?

wetneb:
Hi @Unnamed-1408, are you still working on this?

Unnamed-1408:
> Hi @Unnamed-1408, are you still working on this?

Sorry I forgot to reply, I have searched A considerable part of libraries but didn't see one that can generate markdown in java

wetneb:
Thank you for looking into this! Then let us close this for now, we can revisit this later if we find another solution, perhaps?

PR #4893, opened 2022-05-26
Added error messages for scatterplot facet

### LINKED ISSUE #4890: Scatterplot facet should show error on non-numeric column
Column->Facet->Scatterplot facet requires the selected column to be numeric and at least one other numeric column.

### To Reproduce
1. Create Project. Choose sample file foods.csv
2. Next then Create Project
3. Calcium Column -> Facet -> Scatterplot Facet 

### Current Results
Scatterplot shows a blank dialog when invoked on a non-numeric column.

### Expected Behavior
An error message should be displayed.

### Screenshots
image

elebitzero:
Fixes #4890

- Added error messages to prevent users from attempting to do a scatter plot facet without the requisite columns.

Let me know if the wording of the error message is alright:

When column is not numeric:

"Scatterplot requires a numeric column (green values). The column $1 has less than 50% numeric data. If a column has numerical data but is stored as text (black color), then use Column->Edit Cells->Common Transforms->to Number to convert."

image

When column is numeric, but there are no other numeric columns to compare against:

"No other numeric columns, (columns with more than 50% numeric data (green values)), to compare with Column $1."

image


wetneb (approved):
Works for me, leaving open for longer if people want to chime in.

ostephens:
Looking at the code the test of whether a column is Numeric or not seems to be that more than half the values in the column are numeric. So two questions

1. Are we happy with this as our threshold for allowing scatterplots? (I don't have any strong feelings as I never use this functionality!)
2. And perhaps more importantly - should we be more explicit in the user feedback to state that this is the definition?

wetneb:
It is definitely not ideal to rely on such a threshold… I guess it would make sense to mention it in this error message indeed!

elebitzero:
I updated the error messages to mention the threshold, and I noticed the scatterplot documentation doesn't mention what a numeric column is: (Ref: https://docs.openrefine.org/manual/facets#scatterplot-facet). Scatterplot seems to be the only feature which uses the concept of a numeric column.  Other features work on individual cells, like numeric facet.

wetneb (approved):

PR #4894, opened 2022-05-27
csv to json conversion challenge - smhumraan

smhumraan:
csv to json conversion challenge - smhumraan

Fixes #{issue number here}

- 
- 
-


wetneb:
As you titled this PR "csv to json conversion challenge", I wonder who gave you this challenge? Is this an exercise for a course perhaps?
The templating exporter is currently playing the role of our JSON exporter. This is definitely not perfect (for instance because it is hard or impossible to export back a JSON file to its original format), so there could be some improvement on this. However, I would expect those improvements to be first discussed in an issue (or elsewhere) to have a clear design that people agree on.
In this form, it looks like your JSON exporter is generating a fairly flat JSON which is probably equivalent to that of our templating exporter, so I would rather decline that PR on this basis.

smhumraan:
#@AhmadRumi1 @AhmadRumi1

PR #4897, opened 2022-05-27
(I#3805) Fix problem with frequent writes to metadata files

### LINKED ISSUE #3805: Workspace files being written unnecessarily
### Current Results

It looks like all of the individual projects' `metadata.json` files are getting written every 5 minutes, even if the project hasn't been modified in any way - or even opened. This is both wasteful and error prone because it provides many more windows for project corruption when no file write is needed.

### Expected Behavior
File updates should be minimized and restricted to those that are actually needed. Unmodified files should not be re-written.

### Screenshots
Screen Shot 2021-04-10 at 8 20 00 PM

### Versions
 - Operating System: macOS 11.2.3 Big Sure
 - Browser Version: Chrome Version 89.0.4389.90 (Official Build) (x86_64)
 - JRE or JDK Version: bundled
 - OpenRefine: 3.4.1

ComgLq24:
Fixes #3805

Frequent writes to metadata files are due to deprecated class property `written` in `ProjectMetadata.java` which has been assigned null and remain unchanged. When `FileProjectManager` try to judge whether there is changed in metadata, it utilize this property for every project, resulting in metadata files written for every project.

Solution is to store temporary property recording the last save time for each metadata and update it every time `FileProjectManager` writes the metadata.

- Modified property `written` by `_lastSave` in `ProjectMetadata.java`
- Add setter for `_lastSave` which store the current localdatetime in `ProjectMetadata.java`
- Modified `isDirty` method in `ProjectMetadata.java` which compares `_lastSave` and `_modified`
- For method `saveToFile` first get ids of the modified project judging by `isDirty`, then save those project with the ids
- State update metadata files as need  if length of id list is not null
- Make `ProjectMetadataUtilities.java` to update '_lastSave` everytime metadata file is updated.

A test is shown in the following screeshots
**Projects**
project_struct
**Original Metadata Files**
origin
**Modify one of the project**
modified
**Metadata Files After Modification(Only metadata files of one project has been written)**
aftermodified
terminal

At last, I feel sorry as it seems a bit messy in the `file changes` window. It seems that my editor adds some blank space before some part of the codes for no reason.

### DIFF
--- a/main/src/com/google/refine/ProjectMetadata.java
+++ b/main/src/com/google/refine/ProjectMetadata.java
@@ -387,6 +387,6 @@ public class ProjectMetadata {
 
     @JsonIgnore
     public void setLastSave() {
-        lastSave = LocalDateTime.now();
+        lastSave = Instant.now();
     }
 }
--- a/main/src/com/google/refine/ProjectMetadata.java
+++ b/main/src/com/google/refine/ProjectMetadata.java
@@ -68,7 +68,7 @@ public class ProjectMetadata {
     @JsonProperty("modified")
     private Instant _modified;
     @JsonIgnore
-    private Instant written = null;
+    private Instant lastSave = null;
     @JsonProperty("name")
     private String _name = "";
     @JsonProperty("password")
@@ -125,23 +125,22 @@ public class ProjectMetadata {
 
     protected ProjectMetadata(Instant date) {
         _created = date;
+        _modified = _created;
         preparePreferenceStore(_preferenceStore);
     }
 
     public ProjectMetadata() {
         this(Instant.now());
-        _modified = _created;
     }
 
     public ProjectMetadata(Instant created, Instant modified, String name) {
         this(created);
-        _modified = modified;
         _name = name;
     }
 
     @JsonIgnore
     public boolean isDirty() {
-        return written == null || _modified.isAfter(written);
+        return lastSave == null || _modified.isAfter(lastSave);
     }
 
     static protected void preparePreferenceStore(PreferenceStore ps) {
@@ -384,5 +383,10 @@ public class ProjectMetadata {
             logger.error(ExceptionUtils.getFullStackTrace(e));
             throw new RuntimeException(e);
         }
+    }
+
+    @JsonIgnore
+    public void setLastSave() {
+        lastSave = LocalDateTime.now();
     }
 }
--- a/main/src/com/google/refine/io/FileProjectManager.java
+++ b/main/src/com/google/refine/io/FileProjectManager.java
@@ -44,6 +44,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.stream.Collectors;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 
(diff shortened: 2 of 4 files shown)


main/src/com/google/refine/ProjectMetadata.java
```diff
@@ -373,4 +375,9 @@ public void setAnyField(String metaName, String valueString) {
             logger.error(ExceptionUtils.getFullStackTrace(e));
         }
     }
+
+    @JsonIgnore
+    public void set_lastSave(){
```
wetneb, line 380: To follow the existing naming conventions, I would rather name this `setLastSave`.

main/src/com/google/refine/io/FileProjectManager.java
```diff
@@ -292,29 +297,36 @@ protected void saveWorkspace() {
             logger.info("Saved workspace");
         }
     }
-    
-    protected boolean saveNeeded() {
-        boolean projectSaveNeeded = _projectsMetadata.entrySet().stream()
-                .anyMatch(e -> e.getValue() != null && e.getValue().isDirty());
-        return projectSaveNeeded || _preferenceStore.isDirty();
+
+    protected List<Long> getModifiedProject(){
```
wetneb, line 301:
```suggestion
    protected List<Long> getModifiedProjectIds() {
```

wetneb (commented):
Thank you! I have not tested it but it looks sensible to me. Have you considered writing unit tests for this change? I think it should be doable and worthwhile.

ComgLq24:
@wetneb  Sure, I'm glad to give a hand on the junit test. However, I'm having several final exams coming up. Once I finish, I would continue the work. Thanks for understanding!

ComgLq24:
@wetneb I've added one unit test in `FileProjectManagerTests` and names of methods are also modified as suggested. Thanks! :smiley:

main/src/com/google/refine/ProjectMetadata.java
```diff
@@ -123,23 +123,25 @@ public class ProjectMetadata {
...
         this(created);
         _modified = modified;
         _name = name;
     }
-    
+
     @JsonIgnore
     public boolean isDirty() {
-        return written == null || _modified.isAfter(written);
+        return _modified.isAfter(lastSave);
```
wetneb, line 144: The case where `_modified` (previously, `written`) was `null` seemed to be an indication that the project medatada was never saved, so I am not sure it is safe to initialize it with `LocalDateTime.now()` in the constructor. If I create a brand new `ProjectMetadata` object, then I expect that it `isDirty()` because it has never been saved yet, right? Also, it is not clear to me that `_modified` is always initialized (in one of the constructors it is not) so I suspect there could be a Null Pointer Exception in some cases.

main/tests/server/src/com/google/refine/io/FileProjectManagerTests.java
```diff
@@ -116,4 +120,34 @@ public void deleteProjectAndSaveWorkspace() throws IOException {
...
+        manager.saveWorkspace();
+        long idA = manager.getProjectID("A");
+        long idB = manager.getProjectID("B");
+        Path pathA = Paths.get(workspaceDir.getAbsolutePath(), String.valueOf(idA) + ".project", "metadata.json");
+        Path pathB = Paths.get(workspaceDir.getAbsolutePath(), String.valueOf(idB) + ".project", "metadata.json");
+        File metaAFile = pathA.toFile();
+        File metaBFile = pathB.toFile();
+        long timeBeforeA = metaAFile.lastModified();
+        long timeBeforeB = metaBFile.lastModified();
+        Thread.sleep(3000);
```
wetneb, line 145: This seems like a rather long sleep, I would expect we could at least halve it and the test would still pass reliably (even on systems where the lastModified() method has a second granularity - I do not actually know if this varies from system to system but I would expect all systems are at least precise to the second if not more).

wetneb (approved):

ComgLq24:
@wetneb I totally agree with your idea. I modified  `ProjectMetadata.java` based on the following logic:

1. First, following the original logic, `_modified` attribute is to indicate the time when the metadata in ROM is updated and `lastSave` attribute is to indicate the time when the local metafiles is updated. 
2. When a metadata object is created, the `_modified` attribute should be the create time and `lastSave` should be null since the metadat has never been saved to the local.
3.  We implement `2` by adding `_modified = _created` in the first constructor(as the other two constructors will always invoke the first) and by initializing `lastSave` with null
4. `isDirty()` will return true if `lastSave` is null denoting that metadata is never saved into local files.

As for the unit test, the time for the thread to sleep has shrunk to `1 second` .
Thanks for the advice and apologize for my mis consideration

wetneb (approved):
Sorry for the delay in merging this!

ComgLq24:
> Sorry for the delay in merging this!

Actually it feels great to contribute to the society, thanks a lot for giving the opportunity!

tfmorris (commented):
I need to look at this some more to fully understand what's going on, but I don't think it actually fixes #3805 because each project will get written every time OpenRefine starts, flushing the users' backup copies of the project metadata.

A contributing, pre-existing issue seems to be that all of the projects' metadata files are loaded at every startup which seems to be a change that could date back as far as 2018.

I'll try and dig into this more in the next few days and come up with a patch.

PR #4898, opened 2022-05-27
Added whitespaces to the "collapse consecutive whitespaces" command

### LINKED ISSUE #4883: "Collapse consecutive whitespace" operation does not collapse all possible unicode whitespace
The "Collapse consecutive whitespace" operation does not work when applied to certain whitespace unicode characters.

### To Reproduce
1. First, import this openrefine project: [Clipboard.openrefine.tar.gz](https://github.com/OpenRefine/OpenRefine/files/8762921/Clipboard.openrefine.tar.gz)
2. Then, run the "Collapse consecutive whitespace" operation on the only column

### Current Results
The cell is not edited

### Expected Behavior
The cell should be edited to "hello world"

### Versions
 - JRE or JDK Version: 11
 - OpenRefine: 3.6-SNAPSHOT

### Datasets

Real world dataset where this appears: https://opendata.paris.fr/explore/dataset/lieux-de-tournage-a-paris/information/?disjunctive.type_tournage&disjunctive.nom_tournage&disjunctive.nom_realisateur&disjunctive.nom_producteur&disjunctive.ardt_lieu

### Additional context
Discovered while doing a demo at Dataharvest 2022

carlos-montano-hub:
fixes #4883

- replaced `value.replace(/\\s+/,' ')` for `value.replace(/[\\p{Zs}\\s]+/,' ')` 
- 
-

### DIFF
--- a/main/tests/cypress/cypress/integration/project/grid/column/edit-cells/common-transforms/collapse-consecutive-whitespace.spec.js
+++ b/main/tests/cypress/cypress/integration/project/grid/column/edit-cells/common-transforms/collapse-consecutive-whitespace.spec.js
@@ -3,8 +3,8 @@ describe(__filename, function () {
     const fixture = [
       ['NDB_No', 'Shrt_Desc'],
       ['01001', 'THIS    IS A     TEST'],
-      ['01002', 'THIS    IS ANOTHER     TEST'],
-      ['01003', 'THIS IS a THIRD TEST'],
+      ['01002', 'THIS    IS ANOTHER     TEST'], // this line contains U+00A0 (non-breaking space)
+      ['01003', 'THIS IS a THIRD TEST'], // this one too
     ];
     cy.loadAndVisitProject(fixture);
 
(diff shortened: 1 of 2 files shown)


wetneb (commented):
Thanks a lot! That looks very good to me. Here are a few ideas for improvement:
* Do we actually need to keep the `\s` in the regex? I would suspect those characters are already included in `\p{Zs}`, no?
* We have some functional tests (written in Cypress) and one of them covers this functionality. Perhaps it would be not too hard to adapt this functional test so that the example data includes some interesting Unicode whitespace? You can find the test at `main/tests/cypress/cypress/integration/project/grid/column/edit-cells/common-transforms/collapse-consecutive-whitespace.spec.js`

ostephens:
>Do we actually need to keep the \s in the regex? I would suspect those characters are already included in \p{Zs}, no?

I think we need to keep both as \p{Zs} does not include the tab character (I'm not sure if there are other characters included in \s but not \p{Zs} but the tab character is definitely an issue)

wetneb (approved):

carlos-montano-hub:
Hello!
Sorry for the delay, I was a little absent the past month. About the testing, should I make another issue for that, or was it already fixed?

wetneb:
Thanks for checking! I added the test to this pull request already (see https://github.com/OpenRefine/OpenRefine/pull/4898/files)

PR #4900, opened 2022-05-28
Deleted the first default Sheet in OdsExporter

### LINKED ISSUE #4864: ODS exporter adds an additonal empty sheet to the project
The ODF exporter creates files with two sheets: the first is empty, and the second contains the project data. We should remove the first one.

### To Reproduce
1. First, export any project to ODF
2. Then, open it (for instance with LibreOffice)

### Current Results
The exported file contains two sheets, the first one being empty.

### Expected Behavior
The exported file should contain a single sheet with the project data.

### Versions
 - Operating System: Debian
 - JRE or JDK Version: 11
 - OpenRefine: 3.6-SNAPSHOT

### Datasets

Example exported file: [Clipboard.ods](https://github.com/OpenRefine/OpenRefine/files/8716506/Clipboard.ods)

### Additional context
Similar to #4738

xiao2k:
Fixes #4864 

- I think we can use `getTableByName("Sheet1")` and `defTable.remove();` to solve this issue.
- I made a little change for your reference
-


main/tests/server/src/com/google/refine/exporters/OdsExporterTests.java
```diff
@@ -126,8 +126,8 @@ public void exportSimpleOds() throws IOException {
         try {
             OdfDocument odfDoc = OdfDocument.loadDocument(new ByteArrayInputStream(stream.toByteArray()));
             List<OdfTable> tables = odfDoc.getTableList();
-            Assert.assertEquals(tables.size(), 2); // don't know how the first sheet generate yet
-            OdfTable odfTab = tables.get(1);
+            Assert.assertEquals(tables.size(), 1); // deleted the first sheet generated by defult
```
wetneb, line 129:
```suggestion
            Assert.assertEquals(tables.size(), 1); // we deleted the first sheet generated by default
```

main/src/com/google/refine/exporters/OdsExporter.java
```diff
@@ -76,7 +76,13 @@ public void export(final Project project, Properties params, Engine engine,
...
             public void startFile(JsonNode options) {
                 table = OdfTable.newTable(odfDoc);
-                table.setTableName(ProjectManager.singleton.getProjectMetadata(project.id).getName());
+                String tableName = ProjectManager.singleton.getProjectMetadata(project.id).getName();
+                table.setTableName(tableName);
+                if(!tableName.equals("Sheet1") && odfDoc.getTableList().get(0).getTableName().equals("Sheet1")){
+                    OdfTable defTable;
+                    defTable = odfDoc.getTableByName("Sheet1");
+                    defTable.remove();
+                }
```
wetneb, line 85: Thanks, that seems to work! However, this is likely to break if the default sheet is not named `Sheet1`, which could potentially happen in a later version (for instance because the name gets localized? `Feuille1` in French?), or if the default table somehow gets moved to index 1 (because the new table gets created at index 0, maybe?).

I can think of other approaches:
* before calling `OdfTable.newTable`, delete all existing tables in the document (I suspect that fails because we must always have at least one table in the document)
* after calling `OdfTable.newTable` and setting its name, loop through all the tables in the document and delete those whose name is not the name we just set - that I think should work and should be more reliable.

wetneb (commented):
Very nice, thank you!

wetneb (approved):

PR #4903, opened 2022-05-28
Replace 'parse-next' and it's translations with 'parse-first'

### LINKED ISSUE #4902: Parse next X line(s) as column headers should be renamed "Parse first"
It is a bit confusing when someone imports a file or pastes from clipboard and wants to enable column headers.
The wording that we chose "Parse next" isn't quite what we meant as "Parse first".
We have several translation files that likely need to be updated.
For example, the EN translation: 
 https://github.com/OpenRefine/OpenRefine/blob/3cbb5b143df3caa6e50d5493a34cee3f12cbcf77/extensions/database/module/langs/translation-en.json#L53

Perhaps also in our code, the variable `parse-next` might also need to be updated to `parse-first` or `parse-header` to avoid confusion in the future.

 Screenshot:
 
image

Sync271:
Fixes #4902 

- Replace `parse-next` with `parse-first` everywhere
- Change translations of `Parse Next` to translations of  `Parse First` everywhere

I've used Google for all the translations so they might not be accurate so I need help with that!
A few already that were translated already meant "Parse First", so I didn't touch those translations.


wetneb:
The issue is being discussed, so I would wait for this discussion to conclude before making any progress here, but independently of this - please only edit translations in languages you master well yourself, and do not use automated translation for this - instead, leave the task to our community of translators.

Sync271:
Yup, alright.
My bad, I'll keep that in mind.
Depending on where the discussion heads, I can revert the changes for translation and keep everything else.

wetneb:
So, given the discussion in the issue, I think we need to go back to the drawing board and come up with a new UI proposal for this. I would therefore propose to close this PR in the meantime, if that is cool with you?

Sync271:
Yup, no worries.

PR #4911, opened 2022-06-01
Fixed jitter in column -> sort dialog

elebitzero:
- Fixed jitter found in column -> sort dialog, as shown here:

Before:
sort_jitter
After:
sort_jitter_fixed


wetneb (approved):
Nice one, thank you!

PR #4929, opened 2022-06-06
Implement a SPARQL Importer

### LINKED ISSUE #1212: Create project from SPARQL query
As suggested by @Daniel-Mietchen at https://github.com/wetneb/openrefine-wikidata/issues/7 it would be great to be able to create a project using a SPARQL query (for instance against Wikidata).

Ideally, cells corresponding to entities would be automatically reconciled against the right identifier and schema space (this seems hard to do in a generic way for any SPARQL endpoint but it is quite easy for Wikidata).

OpenRefine would embed a query editor (ideally, with most of the features of the Wikidata Query Service, although again that seems hard to do in a generic way). Perhaps the way to go is just to provide two import methods, one for generic SPARQL and one for Wikidata SPARQL. This would make sense as Wikidata is already the default reconciliation service.

--------------------------
_This is a [proposed Outreachy project in 2022](https://github.com/OpenRefine/OpenRefine/wiki/GSoC-Outreachy-2022%C2%A0Ideas). If you are **not** planning to apply for an internship via Outreachy/GSoC, we kindly ask that you do not work on this task yet, in order to leave the floor to potential interns._

WaltonG:
Hi @antoine2711, this PR is a work in progress, I have only added SPARQL to the list of default importing sources. I will be working on it to close #1212 (Create project from SPARQL query).


wetneb:
Let's mark it as draft since it is not meant to be merged yet.

antoine2711:
@wetneb: should this code go into the full OpenRefine repo, or a new repo, like the Commons extension?

Regards, Antoine

wetneb:
Good question! I can think of three possibilities:
1. adding this to the core features (`main` module of this repository)
2. implementing this in an extension which is stored in this repository
3. implementing this in an extension which is stored in another repository

Perhaps it would be cleaner to work on this as an extension? No strong feelings about whether it should be in this repository or another one.

WaltonG:
@antoine2711 I concur

PR #4936, opened 2022-06-07
Issue 4877 reconciliation waiting screen

### LINKED ISSUE #4877: Waiting screen while guessing reconciliation types should be internal to the reconciliation dialog
Currently, when opening the reconciliation dialog, we have the following user experience:
1. user selects the reconciliation service
2. a "Working…" screen appears while OpenRefine guesses reconciliation types. The user cannot do anything while this is shown.
3. the reconciliation configuration dialog appears and the user can start configuring the reconciliation

Sometimes the guessing of reconciliation types actually takes quite some time (depending on the recon endpoint, the internet connection speed, the specifics of the dataset…) and it can be quite frustrating to wait for this process to complete - especially if you already know which type you want to reconcile the data to, or if the service does not use types at all.

### Proposed solution
Instead of having this "Working…" screen, we should directly show the reconciliation configuration dialog. We would have a waiting spinner (similar to the one used in data extension preview) in the sub-area of the dialog which shows guessed types. It could look like this (with an animated spinner):
image

### Alternatives considered

None

### Additional context
Problem observed in a session at Dataharvest 2022 (in a conference room with pretty bad wifi)

v1scal:
Fixes #4877

- Remove the "Working ..." pop-up
- Add the spinning wheel inside the types column

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
+++ b/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
@@ -42,7 +42,7 @@ function ReconStandardServicePanel(column, service, container) {
 
 ReconStandardServicePanel.prototype._guessTypes = function(f) {
   var self = this;
-  var dismissBusy = DialogSystem.showBusy();
+  var dismissBusy = DialogSystem.showBusyReconciling();
 
   Refine.postCSRF(
     "command/core/guess-types-of-column?" + $.param({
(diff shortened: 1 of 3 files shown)


wetneb:
Also, I tested it and it works very well! I would still have a suggestion: the panel on the right-hand side where we configure properties could appear as soon as the dialog is set up, it should not need to wait for the suggested types to show up.

v1scal:
Ok I think I also managed to show the properties panel before and treat the error of pressing Start Reconciling before the type guessing finished

wetneb:

wetneb:
Closing per inactivity

PR #4942, opened 2022-06-08
Remove references to specific versions of OpenRefine in the docs.

wetneb:
Those mentions are getting out of date as we publish new versions.
I also removed the entire "What's new" section because the docs is not
an appropriate place for our changelogs, and the contents of the section
were redundant with the other download links listed before.


trnstlntk (approved):
👍

PR #4945, opened 2022-06-09
Rearranged "Search for Match" dialog so the dropdown doesn't cover the buttons

### LINKED ISSUE #4821: 'Search for match' dialog: search dropdown mostly covers 'Don't reconcile cell' button, and fully covers 'Cancel' button
During the (Wikidata) reconciliation process... In the 'Search for match' dialog window that is displayed when searching for a match, the search dropdown mostly covers 'Don't reconcile cell' button, and fully covers 'Cancel' button

### To Reproduce
1. Reconcile a column in your dataset (I have only seen the behavior when reconciling against Wikidata)
2. Find an unreconciled value (or any value really) and click 'Search for match'
3. In the 'Search for match' pop up dialog, when I search for the thing, my 'Don't reconcile cell' button is almost fully covered and the 'Cancel' button is fully covered. In many cases (usually when no reconciled entity is found) I want to choose 'Don't reconcile' and then I need to do some precision clicking, which shouldn't be necessary. I can imagine that, in some cases, the user may just want to cancel the operation too.

Maybe the UI for this operation can be improved so that the buttons are not covered by the search dropdown anymore.

### Expected Behavior

Buttons should remain visible.

### Screenshots

image
Clicking 'Choose new match':
image
Sad face

### Versions
 - Operating System: MacOS 10.13.6
 - Browser Version: Chrome 100.0.4896.127 (Official Build) (x86_64)
 - JRE or JDK Version: n/a?
 - OpenRefine: 3.6-SNAPSHOT [b0f3282]

### Datasets
Any with Wikidata reconciliation going on :-)

elebitzero:
Fixes #4821

- Rearranged "Search for Match" dialog so the dropdown doesn't cover the buttons.
- Fixed escape key handling so hitting escape will close the dropdown without closing the dialog. To close the dialog hit escape again.

New layout:
search-for-match-new-layout

### DIFF
--- a/main/webapp/modules/core/scripts/util/dialog.js
+++ b/main/webapp/modules/core/scripts/util/dialog.js
@@ -74,25 +74,36 @@ DialogSystem.showDialog = function(elmt, onCancel) {
 
   var level = DialogSystem._layers.length;
 
-  $(window).on('keydown',escapeKey);
-  
+  DialogSystem.setupEscapeKeyHandling();
+
   return level;
 };
+
+
+DialogSystem.pauseEscapeKeyHandling = function() {
+  $(window).off('keydown',escapeKey);
+}
+
+DialogSystem.setupEscapeKeyHandling = function() {
+  $(window).on('keydown',escapeKey);
+}
 
 DialogSystem.dismissLevel = function(level) {
     var layer = DialogSystem._layers[level];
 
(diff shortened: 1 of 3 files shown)


wetneb (approved):
Yes please!
Leaving this open a bit longer to leave the opportunity for others to chime in.

PR #4949, opened 2022-06-10
Updated refine.bat to use %JAVA_HOME% and be consistent with the refine script

### LINKED ISSUE #4930: Java version messaging to users is inconsistent between Windows and Linux 'refine' scripts
If a user has set their `%JAVA_HOME%` env variable on Windows but the `%PATH%` variable has a different Java version for the runtime and is lower than the compiled class file version 55.0 (Java 11) then users will notice an error and failure during startup even though the Java version check was successful.

### To Reproduce
1. First, set `%JAVA_HOME%` to a path having Java 8 Update 333
2. Keep `%PATH%` env variable set to Java 11 runtime
3. Then, run `refine.bat`

### Current Results
```
E:\GitHub Repos\OpenRefine>refine
Getting Java Version...
openjdk version "11.0.14.1" 2022-02-08
OpenJDK Runtime Environment Temurin-11.0.14.1+1 (build 11.0.14.1+1)
OpenJDK 64-Bit Server VM Temurin-11.0.14.1+1 (build 11.0.14.1+1, mixed mode)
=====================================================
Getting Free Ram...
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.UnsupportedClassVersionError: com/google/refine/Refine has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

#### Windows
Since we now really depend on `%JAVA_HOME%` when setting the CLASSPATH in the `refine.bat` script for Windows, we should fully use %JAVA_HOME% for the Java version inspection and logging to support.log.

We currently have good support for checking the Java versions we support and reporting errors in our `refine` script for Linux and should ideally incorporate that `checkJavaMajorVersion()` functionality and adapt it into our `refine.bat` script for Windows.  That way we are consistent with our Java error messaging to users on both platforms.

#### Linux
One of the additional things in the `refine` Linux script seems to be that we are not using the `$JAVA_HOME` env variable during the `checkJavaMajorVersion()` either but instead always a `$JAVA` variable.  We should ideally always refer to and use `$JAVA_HOME` as a best practice in our version checking and classpath setup for refine startup, no matter the OS, since that is standard practice and gives users the ability to control their own Java environment or symbolically link to a different one.

### Versions
 - Operating System: Windows 11 and Linux
 - Browser Version: Edge latest and Chrome
 - JRE or JDK Version: openjdk version "11.0.14.1" 2022-02-08
 - OpenRefine: master branch

elebitzero:
Fixes #4930 

Changes for refine.bat:
-  Only use java found in %JAVA_HOME% to check the version and run.  
-  Exit if Java Release < 11 
-  Warn if Java Release > 17

Changes for refine:
- Exit if Java Release < 11

Running with Java 8:
```
>refine.bat run
Java 8 (1.8.0_332)
OpenRefine requires Java version 11 or later. If you have multiple versions of Java installed, please set the environment variable JAVA_HOME to the correct version.
```
Running with Java 17:
```
>refine.bat run
Java 17 (17.0.2)
Getting Free Ram...
23:47:36.994 [            refine_server] Starting Server bound to '[127.0.0.1:3333](http://127.0.0.1:3333/)' (0ms)

```

### DIFF
--- a/refine
+++ b/refine
@@ -776,8 +776,8 @@ checkJavaMajorVersion() {
     # Java 9+ starts with x using semver versioning
     major=`echo ${java_ver} | sed -E 's/([0-9]+)(-ea|(\.[0-9]+)*)/\1/g'`
   fi
-  if (( ${major} < 8 )); then
-    error "OpenRefine requires Java version 8 or later. If you have multiple versions of Java installed, please set the environment variable JAVA_HOME to the correct version."
+  if (( ${major} < 11 )); then
+    error "OpenRefine requires Java version 11 or later. If you have multiple versions of Java installed, please set the environment variable JAVA_HOME to the correct version."
   fi
   if (( ${major} > 17 )); then
     echo "WARNING: OpenRefine is not tested and not recommended for use with Java versions greater than 17."
--- a/refine.bat
+++ b/refine.bat
@@ -235,25 +235,46 @@ if ""%ACTION%"" == """" goto doRun
 
 :doRun
 rem --- Log for troubleshooting ------------------------------------------
-echo Getting Java Version...
-java -version 2^>^&1
-echo.=====================================================
-for /f "tokens=*" %%a in ('java -version 2^>^&1 ^| findstr "version"') do (set JVERSION=%%a)
-echo Getting Free Ram...
+set JAVA="%JAVA_HOME%/bin/java"
+set JAVA_VERSION=""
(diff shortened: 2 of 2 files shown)


wetneb (approved):
Thank you! Adding this to 3.6 as well.

PR #4953, opened 2022-06-12
[4952] Save project metadata after import

### LINKED ISSUE #4952: Project not saved after creation
Hi,
I noticed a behaviour where a freshly created project is not saved.

### To Reproduce
- Start OpenRefine
- Create a new project usign e.g. a [sample csv file addresses.csv](https://people.sc.fsu.edu/~jburkardt/data/csv/addresses.csv)
- Optionally modify the first entry of the data
- Stop OpenRefine by closing the terminal window
- Start OpenRefine again
- Select Open Project ..

### Current Results
The  previously created OR project is not present in the list effectively loosing all changes made to the project.
The AppData folder contains a `1864769305521.project.corrupted` 

### Expected Behavior
The project is present 

### Screenshots

openrefine

### Versions
 - Browser Version:  Chrome Version 102.0.5005.63 (Official Build) (64-bit) 
 - JRE or JDK Version: 
 - OpenRefine: 3.5.2

### Datasets
 [addresses.csv](https://people.sc.fsu.edu/~jburkardt/data/csv/addresses.csv)

### Additional context
I tried on macOS where it seems to be working fine.  I believe that closing the OpenRefine  window kills the app and there is no time to save the project. Should not OpenRefine save the project directly after it's created?

JanC:
Fixes #4952

The `createProjectSynchronously` calls `registerProject` which registers it only in memory. The project metadata are not persisted so if the application stops, the project is lost.

This PR adds the call to `ensureProjectSaved` which perists the newly imported project metadata

Please let me know if there are any tests that should be modified or added

### DIFF
--- a/main/src/com/google/refine/commands/project/CreateProjectCommand.java
+++ b/main/src/com/google/refine/commands/project/CreateProjectCommand.java
@@ -127,7 +127,12 @@ public class CreateProjectCommand extends Command {
             
             List<Exception> exceptions = new LinkedList<Exception>();
             
-            long projectId = ImportingUtilities.createProject(job, format, optionObj, exceptions, true);
+            long projectId = ImportingUtilities.createProject(job,
+                    format,
+                    optionObj,
+                    exceptions,
+                    true,
+                    ProjectManager.singleton);
 
             HttpUtilities.redirect(response, "/project?project=" + projectId);
         } catch (Exception e) {
--- a/main/src/com/google/refine/importing/DefaultImportingController.java
+++ b/main/src/com/google/refine/importing/DefaultImportingController.java
@@ -49,6 +49,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.refine.ProjectManager;
 import com.google.refine.RefineServlet;
 import com.google.refine.commands.Command;
 import com.google.refine.commands.HttpUtilities;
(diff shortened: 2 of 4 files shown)


wetneb (commented):
Thanks!
I guess the main risk with doing this synchronously, directly after project registration, is that it could add quite some processing time - because that does not just save the metadata, but also the project data, right?
The alternative would be to schedule a save immediately after (done in another thread). It is not clear to me if that is worth the complexity.

In any case, it would be nice to have a test for this. `ImportingUtilities` is a big mess full of static stuff, so it might not be completely trivial, but I hope it should be doable.

JanC:
hey @wetneb 

>  because that does not just save the metadata, but also the project data, right?

so I had a second look at what is done inside the `ProjectManager.singleton.ensureProjectSaved` method and as you said both metadata and project data are saved and both are needed for a proper project loading.

> I guess the main risk with doing this synchronously, directly after project registration, is that it could add quite some processing time

It could indeed in large projects. But right now, if one closes OpenRefine, it ends in a corrupt state which is worse that waiting till the project is correctly saved.

> The alternative would be to schedule a save immediately after (done in another thread). It is not clear to me if that is worth the complexity.

I think introducing a async way could be done in several ways and I'd probably need guidance as far as the architecture goes.

Should the `ProjectManager.ensureProjectSaved` become asynchronous (or have a async equivalent)?
or should the async be part of the `ImportingUtilities` or `ImportingManager`?

wetneb:
Thanks for checking back!

> But right now, if one closes OpenRefine, it ends in a corrupt state which is worse that waiting till the project is correctly saved.

That waiting should normally happen if OpenRefine is *correctly* closed - I know this is not the case on Windows if one simply closes the console, but that sounds like something we should solve differently (change the Windows packaging).

Other thoughts:
* in the new architecture (which will become version 4.0) the project is indeed saved at creation. So I am definitely not opposed to the idea, but I feel like this is touching a rather high-risk area, with quite some potential to make some use cases worse - hence my reluctance! Perhaps it is better to wait for 4.0, which will be extensively tested for such sorts of issues before a release (and with less backwards-compatibility expectations from the users, for instance because we are not sharing workspaces with 3.x instances).
* after all, maybe your initial version (synchronous saving) is safer, because it guarantees that the project will only be opened by the user after this saving step…

JanC:
I made the `ProjectManager` injectable instead of using directly the singleton and wrote some basic unit tests

JanC:
I'm closing the PR as discussed in the ticket
https://github.com/OpenRefine/OpenRefine/issues/4952#issuecomment-1270819551

PR #4955, opened 2022-06-12
[I #3373] edit recon service url

### LINKED ISSUE #3373: Editing scope for added reconciliation services
It would be really nice if there is an option/feature to edit (name /URL / label) of an already added reconciliation services either added by using RDF extension (add reconciliation service > based on RDF file) or through "Add standard  Service" in place of  the present option  to delete and add again.

### Proposed solution

A feature may be added in the panel of reconciliation.

### Additional context

It  could help when a user needs to see the URL of a reconciliation service added earlier but missed to store/record the URL.

v1scal:
Fixes #3373

- Add icon to edit reconciliation service
- Select the changed reconciliation service after a successful change

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/recon-manager.js
+++ b/main/webapp/modules/core/scripts/reconciliation/recon-manager.js
@@ -65,7 +65,7 @@ ReconciliationManager.registerService = function(service) {
 ReconciliationManager.registerStandardService = function(url, f, silent) {
   var dismissBusy = function() {};
   if (!silent) {
-    dismissBusy =  DialogSystem.showBusy($.i18n('core-recon/contact-service')+"...");
+    dismissBusy =  DialogSystem.showBusy($.i18n('core-util-enc/working')+"...");
   }
 
   var registerService = function(data, mode) {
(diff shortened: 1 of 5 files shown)


main/webapp/modules/core/scripts/reconciliation/recon-manager.js
```diff
@@ -120,6 +120,67 @@ ReconciliationManager.registerStandardService = function(url, f, silent) {
...
+                ReconciliationManager.standardServices[i] = data
+                break;
+            }
+        }
+        ReconciliationManager._rebuildMap();
+        ReconciliationManager.save(f);
+        dismissBusy();
+    };
+
+    // First, try with CORS (default "json" dataType)
```
wetneb, line 154: As mentioned earlier, I would like that we do not duplicate this logic with the addition of a new service.

main/webapp/modules/core/scripts/reconciliation/recon-manager.js
```diff
@@ -120,6 +120,67 @@ ReconciliationManager.registerStandardService = function(url, f, silent) {
...
+            "handler" : "ReconStandardServicePanel",
+            "access" : mode
+        };
+
+        for (var i = 0; i < ReconciliationManager.customServices.length; i++) {
+            if (ReconciliationManager.customServices[i].url === oldUrl) {
+                ReconciliationManager.customServices[i] = data
+                break;
+            }
+        }
```
wetneb, line 142: Why not just call the function to delete the service, and then add it again from scratch?

wetneb (commented):
Thanks! I have not tried it out yet, but here are a few suggestions just based on the code changes.

wetneb:

wetneb:
Closed per inactivity.

PR #4957, opened 2022-06-14
Updated UI button size and dialog footer style

elebitzero:
- Updated the CSS  to match the latest jQuery UI dialog style (which is also similar to other modern CSS dialog styles.) Ref: https://jqueryui.com/dialog/#modal-confirmation
- Changed footer styling so the buttons stand out more against the white background.

Before:
image

After:
image

Before:
image

After:
image

Before:
image

After:
image


wetneb:
Thanks!
We do not currently have written-up guidelines about the visual style of the UI, so it is hard for me to know how to review pull requests like this. It is really a matter of taste!

I am not sure we really follow jquery-ui style in the rest of the application, so it is not clear to me if that is really unifying the style. My main worry is that we would be departing from the old Google style which we follow relatively consistently throughout the rest of the application.

In terms of taste (which is really just my own): I think the bigger buttons make sense, but I am not sure around the padding in the dialog border. It does not seem so useful to me.

elebitzero:
Sure, we don't have to follow latest jquery-ui exactly.  I have reverted the dialog border padding. But I do think there is consensus amongst modern CSS frameworks for the larger button size and also the higher contrast in the footer.

After reviewing this further, I realize to be consistent within OpenRefine, we have to update the button size everywhere (except the navigation bar).  I have updated the screenshot examples.

elebitzero:
Thank you for the feedback.

I have clicked through most of the app, and the buttons look fine. If there is any side effect found, it should be easy to fix. Or worst case, we can revert this before 3.7 is released.

I have thought about this more, and after comparing the CSS of OR to other modern CSS I am familiar with like Bootstrap, I'm thinking we don't even need to divider line above the footer. If I remove that, it would look even cleaner.

So, the main change is to just make the buttons bigger and higher contrast by removing the footer background.  

Yes, it would be great if the app could be migrated to new CSS/web framework, but that doesn't seem likely to happen, unless project gets more backing because that is a large undertaking.

Modernizing the button style seemed like low hanging fruit, but if there isn't consensus on that, we can hold off.

wetneb:
Sure! My concerns are really a matter of taste, so they are not blocking. Maybe it is worth gathering more feedback from other people. Perhaps some of @ostephens @trnstlntk @lozanaross @thadguidry? In short: should we change the style of buttons in OpenRefine as suggested by the screenshots above?

trnstlntk:
I like the screenshots with the new style buttons: more airy, and they give a slightly more modern feel. OpenRefine's buttons are indeed on the small side and it will be more pleasant if they are slightly larger.

That said, one of my computers is an old macBook Air with a small screen resolution (1366x768) and I already struggle to keep some dialog windows visible there; extra pixels in our dialogs will not help. However, users with such smallish resolutions are probably getting more and more rare.

tl;dr In general I'm mostly in favor.

thadguidry:
So the larger button sizes provide a slightly bigger and more accessible clickable area.  It does not address text size (WCAG recommends 16px and no smaller than 12px) but that was to be addressed later with a new modular framework.
As long as the button text is compatible with users increasing font sizes through changing their browsers settings under __Appearance__ and __Customize fonts__ then 👍 from me.

lozanaross:
Sorry for sluggish response here. 

In general I am in favor of addressing current accessibility issues in the design system, but my main issue is font size (and sometimes contrast) - making the buttons bigger will not necessarily solve those issues and I think it will introduce 2 new issues: 1) current margin spaces between (and sometimes around) buttons are extremely tiny by modern standards, they just about work with small buttons, but making the buttons bigger only exacerbates this issue rather than alleviate it; 2) modern frameworks are very heavily influenced by e.g. Google's Material design framework, which looks a certain way because it's heavily driven by need to work on mobile / touch devices. This is not really applicable to OpenRefine because it's not a mobile / touch device oriented app. Maybe one day it will be :-) 

TL;DR I am not in favor of changing the design system look piece meal - i.e. change one small thing, but not the dependent elements around it. I am however in favor of upgrading the system to modern standards based on complete interaction patterns - i.e. we rethink a particular interaction from start to finish and redesign that. My idea how to best to organize that is to start a penpot file (https://penpot.app/) that documents the complete current design system and then start iterating on specific components within the design file first - develop a complete vision for a new interaction pattern there before committing code. I can definitely use help with this endeavour, so if this is something you'd be interested in getting involved in @elebitzero, let's start a dedicated issue for it & discuss further.

elebitzero:
@lozanaross Thanks for the comments.

Regarding the font size, the button paddings are in em units, and would be the same even if the default font is increased.

As for the new issues:
- (Issue 1) I can adjust the spacing between buttons as part of this or in a follow-up PR.  
- (Issue 2) From my experience, any modern webapp CSS framework has bigger buttons.  It has nothing to do with mobile design.  This PR updates the button style to match the latest jquery-ui button CSS. Ref: https://jqueryui.com/dialog/#modal-confirmation

> I am not in favor of changing the design system look piece meal - i.e. change one small thing, but not the dependent elements around it. I am however in favor of upgrading the system to modern standards based on complete interaction patterns  

As mentioned above, if the dependent item you are referring to is the spacing between the buttons, I can adjust that.  I'll just copy what the latest jquery-ui is using.

This is a proposed style change. It affects appearance and accessibility and usability of buttons. How the user interacts with the UI, i.e. the user workflow, does not change.  So, it is unclear to me how this requires a deep redesign of user interaction patterns.

From my perspective, this PR clearly improves the button UX. And if there is some issue, we can make further refinements during 3.7. But if there is consensus that these kind of changes should wait for some later redesign of the UI, then I'll withdraw this PR.

lozanaross:
@elebitzero, thanks for the comments. 

My general thinking in terms of what makes the app currently less or more usable has more to do with font size, color contrast and overall layout of where elements are positioned on the page, rather than the size of the buttons alone. While your design proposal looks nicer, of course, I think it's good to keep in mind there are more accessibility challenges. My original idea was to think through the app as a whole and suggest more global changes, but that will require a lot of work of course, so maybe we start small - e.g. the buttons - and go from there. 

So with regards to your concrete proposal, here are my thoughts:
Issue 1: OK, if you can show us some screenshots with new spacing around buttons, that will be great.

> As mentioned above, if the dependent item you are referring to is the spacing between the buttons, I can adjust that. I'll just copy what the latest jquery-ui is using.

This is one dependency, but I am wondering what will happen to other areas where we have buttons in smaller / tight spaces, e.g. see these screenshots:
Screenshot 2022-07-06 at 23 19 45

Screenshot 2022-07-06 at 23 19 35
(In this latter screenshot - there is also some strange hover effect happening with the descriptions, but I'll make a separate issue about that)

Also - will the new button size affect also the size of value / text-entry fields - because the previous fields I think were sized in line with the buttons? That is another aspect that I think we should just double-check how it looks before deploying. 

> This is a proposed style change. It affects appearance and accessibility and usability of buttons. How the user interacts with the UI, i.e. the user workflow, does not change. So, it is unclear to me how this requires a deep redesign of user interaction patterns.

Actually, I think the style changes highlight how awkward the dialog layout really is - for example in the SQL exporter dialog in your original screenshot above - the white footer now makes the whole "cancel" button seem obsolete - why not just have an "x" icon in the top right corner of the dialog & deploy that across the entire design system for dialog window interaction? I think the pale grey line separator instead of the block grey footer looks more elegant, but it makes the design of some dialogs less clear - the choice of why some buttons are in the footer and others not becomes less clear (e.g. I also wonder how it will look in the 'Search for match' screenshot I've posted in this comment). Anyways, my point is - ideally we should always think about overall user workflows even when considering fairly simple style changes. Happy for the others to make final decision though, these are just some general thoughts on the future development of a new design system. I don't want to be a deal breaker here :-)

elebitzero:
@lozanaross

I have updated the spacing between buttons in the dialog-footer to match the latest jQuery 0.4em.

As for the menu buttons at the top, I left them at their current size. Those can be adjusted separately if the main console is redesigned.

elebitzero:
This style change didn't make it in and now the branch is out-of-date.  Closing for now.  This can be revisited if OR's style is modernized in the future.

PR #4971, opened 2022-06-18
#816 improve toString to work with numbers

### LINKED ISSUE #816: Extend toString(value,format) to work with numbers
Currently the format string is only used when formatting dates, but it would be desirable to have it work with numbers as well.  This would allow for things like radix conversion, integer formatting of large numbers, etc.  Dates would continue to get formatted using SimpleDateFormat, but numbers would get passed to java.util.Formatter.

elroykanye:
Fixes #816

- used the java.util.Formatter to format numbers


main/src/com/google/refine/expr/functions/ToString.java
```diff
@@ -54,7 +55,7 @@ public Object call(Properties bindings, Object[] args) {
                     OffsetDateTime odt = (OffsetDateTime)o1;
                     return odt.format(DateTimeFormatter.ofPattern((String)o2));
                 } else if (o1 instanceof Number) {
-                    return String.format((String) o2, (Number) o1);
+                    return new Formatter().format((String) o2, o1).toString();
```
wetneb, line 58: From the issue discussion it is not clear to me what difference it makes to use `Formatter::format()` rather than `String.format`. Could you demonstrate the use cases that motivate the change with unit tests?

elroykanye: This is actually true. Initially, I thought the Formatter was a better option to use that String.format(), but I did more digging and landed on this [stackoverflow page](https://stackoverflow.com/questions/2809633/what-is-the-difference-between-messageformat-format-and-string-format-in-jdk-1-5#:~:text=String.format%20is%20just%20a%20shortcut%20to%20Formatter%2C%20this,formatting%20convention%2C%20as%20described%20in%20the%20linked%20documentation.) - referring to the last answer on it.

This is how the method is implemented in java too:

```java
public static String format(String format, Object... args) {
        return new Formatter().format(format, args).toString();
}
```

That said, I think I may have misunderstood the issue and will return to the issue's page to clarify all doubts.

main/src/com/google/refine/expr/functions/ToString.java
```diff
@@ -54,11 +57,14 @@ public Object call(Properties bindings, Object[] args) {
...
                 } else if (o1 instanceof Number) {
-                    return String.format((String) o2, (Number) o1);
+                    DecimalFormat df = new DecimalFormat((String) o2);
+                    return df.format(o1);
                 }
             } else if (args.length == 1) {
                 if (o1 instanceof String) {
                     return (String) o1;
+                } else if (o1 instanceof Number) {
+                    return NumberFormat.getInstance().format(o1);
```
ostephens, line 67: The "default" transformations to strings for other objects are in the com.google.refine.util.StringUtils class - wouldn't it make sense for this default number->string behaviour to also be in that class also?

ostephens: Note that there is already some default behaviour for number.toString() in GREL (I haven't tracked down where this happens - but it definitely already works - do we need any additional code here?)

elroykanye: This is true @ostephens 
I talked this out with Antonin and resolved that this may have already been fixed. I pushed some commits showing tests which demonstrate that and a try-catch for invalid formats.

wetneb (approved):
Looks good to me, thank you!

PR #4973, opened 2022-06-19
3082 improve handling of dates in cell editing UI

### LINKED ISSUE #3082: Improve handling of dates in cell editing UI

Say you want to manually edit a cell to input a date there. This is what it currently looks like:
image

The problems are:
* As a user, how do I know which formats are supported? Especially if I start with a text value and want to transform it to a date, I should not have to refer to the user manual to find out how to format it.
* When I try to input a wrong date, a JS alert pops up: "Not a valid date." That is not very helpful!

Find a way to add a short description of the supported format(s) to the dialog itself, and/or improve the error message to mention the formats.

We could consider using an off-the-shelf date picker but it is not clear to me whether that would be really more user-friendly in a data cleaning context. Perhaps it could be used optionally, with still the ability to modify the date textually?

**Additional context**
Brought up in a documentation meeting with @allanaaa and @thadguidry.

elroykanye:
Fixes #3082 

- updated the cell-editor file to include a section for binding help text
- added conditional help text setting on date cell types when cell is being edited


main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -604,6 +604,13 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
   cellDataType = (this._cell !== null && "t" in this._cell && this._cell.t !=  null) ? this._cell.t : cellDataType;
   elmts.typeSelect.val(cellDataType);
 
+  if (cellDataType === "date") {
+    var simpleDateFormatRef = $('<a></a>').attr('href', 'https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html').append('here');
```
wetneb, line 608: There are a few problems here:
* the `here` word will not be translatable
* the position of the `here` word with the link is also not configurable by the translators (there could be languages where it does not make sense to have the link at the end of the text
* if we are to use this link, I would rather formulate like this "Supply dates based on the [SimpleDateFormat syntax](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html)" (shorter, more effective, avoids [the problems of "click here" links](https://heyoka.medium.com/dont-use-click-here-f32f445d1021))

wetneb (commented):
Thanks!
I do not think we should refer our users to this SimpleDateFormat page, because first it is very technical and second it does not actually tell users which formats are supported (since this page only describes how to specify a date format with SimpleDateFormat).

I think it would be much simpler to refer to the ISO format (not sure which ISO number) in which the date is formatted by default, and give an example accepted value (so that people can imitate it directly without having to read lengthy standards)

elroykanye:
> I think it would be much simpler to refer to the ISO format (not sure which ISO number) in which the date is formatted by default, and give an example accepted value (so that people can imitate it directly without having to read lengthy standards)

That should be ISO 8601 right? I got this from the [docs](https://docs.openrefine.org/manual/exploring#dates).

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
elroykanye: Oh, this is true. Thanks! 😃

wetneb:
This is how it currently looks for me:
image

Here are a few suggestions:
* For the message, how about something like this? `Dates are formatted in the ISO 8601 format, such as "1987-03-04T03:43:30Z"`
* The message is currently only shown when the original datatype of the cell was a date, but not when the user has manually changed it from its original datatype (edit a string-valued cell, pick date datatype: you do not see the message). I think we should show the message in that case as well.

main/webapp/modules/core/scripts/views/data-table/cell-ui.js
```diff
@@ -604,6 +604,21 @@ DataTableCellUI.prototype._startEdit = function(elmt) {
...
+      elmts.cell_help_text.html($.i18n('core-views/cell-edit-date-help'));
+      $(elmts.cell_help_text).css({'font-style': 'italic', 'display': ''});
+    } else {
+      $(elmts.cell_help_text).css({'display': 'none'});
+    }
+  });
+
+  if (cellDataType === "date") {
+    elmts.cell_help_text.html($.i18n('core-views/cell-edit-date-help'));
+    $(elmts.cell_help_text).css({'font-style': 'italic', 'display': ''});
```
wetneb, line 619: I think we can improve this code a little:
* use `.show()` or `.hide()` rather than explicit calls to `.css(...)` to show and hide an element
* setting up the element (by adding the text and setting its font-style) should be done only once, when setting up the dialog (even if the element is not shown initially). That makes the rest of the code cleaner: when an event happens, you should only need to `.show()` or `.hide()` the element.

wetneb (commented):
Current screenshot:
image


I am not too sure about the italics actually, I do not think we are using this elsewhere to give indications to the user. Perhaps we could simply use a normal font-style? No strong feelings about it, but perhaps it is best to gather opinions from others (random pinging: @ostephens @thadguidry @lozanaross)

thadguidry:
We're binding context sensitive help text for the cell type, so that's a good thing now.

As for as styling the help text, italics is discouraged on long paragraphs but short phrases or words is kinda OK by WCAG standards (dyslexia reduction in 1 study).

I am more worried about the context portrayed and phrasing.  This is an input from user and we are providing that input's help text, but are we saying how the input **needs** to be formatted? Or how the input will be represented in our storage?  I think we can shorten the phrase to whichever meaning we are trying to tell the user and then the stylized help text would work better in whichever style.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -764,6 +764,7 @@
     "core-views/column-join-copy-to-new-column": "Write result in new column named",
     "core-views/column-join-delete-joined-columns": "Delete joined columns.",
     "core-views/column-join-dont-escape": "In separator and nulls substitutes, use \\n for new lines, \\t for tabulation, \\\\n for \\n, \\\\t for \\t.",
+    "core-views/cell-edit-date-help": "Dates are formatted in the ISO 8601 format, such as \"1987-03-04T03:43:30Z\"",
```
thadguidry, line 767:
```suggestion
    "core-views/cell-edit-date-help": "Input in ISO 8601 date format such as 1987-03-04T03:43:30Z",
```

thadguidry: I'd avoid the quotes and make the phrase direct and more concise.

thadguidry (changes requested):

wetneb (approved):

Issue #4972, opened 2022-06-18
Import Project and Language buttons are not visible


fabiolinus:
Hi. I have this type of errors when I open chrome devloper tools. I use Win 10 but I can't see Import Project and Language buttons on the left. Is sounds very strange. Thanks for any help.

Uncaught TypeError: Cannot set properties of null (setting 'id')
    at Refine.OpenProjectUI._renderProjects (index-bundle.js:64956:18)
    at Object.success (index-bundle.js:64942:26)
    at fire (index-bundle.js:3234:31)
    at Object.fireWith [as resolveWith] (index-bundle.js:3364:7)
    at done (index-bundle.js:9842:14)
    at callback (index-bundle.js:10313:8)
    at Object.send (index-bundle.js:10323:7)
    at Function.ajax (index-bundle.js:9740:15)
    at Refine.OpenProjectUI._fetchProjects (index-bundle.js:64937:7)
    at Refine.OpenProjectUI._buildTagsAndFetchProjects (index-bundle.js:64913:10)
index-bundle.js:64486 Uncaught TypeError: Cannot read properties of undefined (reading 'resize')
    at resize (index-bundle.js:64486:32)
index-bundle.js:64486 Uncaught TypeError: Cannot read properties of undefined (reading 'resize')
    at resize (index-bundle.js:64486:32)
    at dispatch (index-bundle.js:5228:27)
    at elemData.handle (index-bundle.js:4880:28)

wetneb:
Thanks for reporting this! It could be useful to have:
* a screenshot
* details of which extensions you have installed
* which operating system and browser you use

Thank you!

fabiolinus:
Thanks a lot, Antonin.
1) screenshot in this mail2) No extensions installed3) Windows 10 and Chrome Versione 102.0.5005.115 (Build ufficiale) (a 64 bit). Both of them are updated.
    Il sabato 18 giugno 2022, 22:03:12 CEST, Antonin Delpeuch ***@***.***> ha scritto:  
 
 


Thanks for reporting this! It could be useful to have:
   
   - a screenshot
   - details of which extensions you have installed
   - which operating system and browser you use

Thank you!

wetneb:
We cannot see your screenshot, I think you need to post it directly on GitHub. Also, it would be useful to know which version of OpenRefine you are using.

fabiolinus:
Hi and thanks a lot. Here is waht you wanted.
[...]

PR #4976, opened 2022-06-21
Added checks for the project data and warnings to the console to skip any malformed projects

elebitzero:
Partial fix for #4972

- Added checks for the project data and warnings to the console to skip any malformed projects so the UI will continue to work with valid projects.


wetneb (approved):
Looks safe to me, thank you!

Issue #2443, opened 2020-03-20
Back-end localization


wetneb:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**

Some strings that appear in the UI are generated in the backend: exceptions, descriptions of operations, documentation of GREL functions, and probably others that I am not aware of.
We currently do not have any way to translate these.

**Describe the solution you'd like**
We should investigate an architecture to make these strings translatable. I can think of two approaches:
* Replace these strings by identifiers (possibly with template parameters in the case of operation description) and delay their formatting to the front-end, such that we can use the existing frontend localization system (Wikimedia jQuery i18n plugin). This would mean that the backend can continue to ignore in which language the application is being used;
* Bring in server-side localization using an appropriate Java library. We should make sure the messages which appear in the backend are picked up by Weblate somehow.

**Describe alternatives you've considered**
We might use one solution for a particular class of messages (say operation descriptions) and another for other messages (say exceptions).

--------------------------
_This is a [proposed Outreachy project in 2022](https://github.com/OpenRefine/OpenRefine/wiki/GSoC-Outreachy-2022%C2%A0Ideas). If you are **not** planning to apply for an internship via Outreachy/GSoC, we kindly ask that you do not work on this task yet, in order to leave the floor to potential interns._

thadguidry:
In the enterprise, I've seen lots of usage with Thymeleaf templates [which are dead simple to use](https://www.baeldung.com/spring-boot-internationalization).

wetneb:
I have proposed this as a potential Outreachy project (and maybe GSoC if we participate to that as well). This means I think it should be a good topic for a 3 months internship. This is therefore not suitable as a first issue: have a look at [our list of good first issues](https://github.com/OpenRefine/OpenRefine/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) instead.

thadguidry:
Some possible libraries:
1. Plain Java with `PropertyResourceBundle` (Since Java 9+ it can handle reading UTF-8 encoding) https://www.baeldung.com/java-resourcebundle
2. Rincl might be able to help here (and uses ICU4J).  It's Apache 2 licensed and I've reached out to the author to see what's up. https://rincl.io/ 
3. Localizer library like Jenkins uses? https://www.jenkins.io/doc/developer/internationalization/i18n-source-code/ and https://github.com/kohsuke/localizer

garretwilson:
[...]

PR #4978, opened 2022-06-22
#2443 back end localization

elroykanye:
Week 3 of #2443 

- added localizer plugin and dependency to application's pom.xml
- updated the pom.xml of the main module to enable maven recognise `resources` directory at the module's root


pom.xml
```diff
@@ -434,8 +451,19 @@
...
+          <url>https://repo.jenkins-ci.org/releases/</url>
+      </repository>
   </repositories>
 
+    <pluginRepositories>
+        <pluginRepository>
+            <id>jenkins-repo</id>
+            <url>https://repo.jenkins-ci.org/releases/</url>
+        </pluginRepository>
+    </pluginRepositories>
```
wetneb, line 465: Is this really needed? It looks like this dependency is available in Maven Central:
https://mvnrepository.com/artifact/org.jvnet.localizer/localizer/1.31

main/pom.xml
```diff
@@ -43,7 +46,7 @@
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>build-helper-maven-plugin</artifactId>
-	<version>${build-helper-maven-plugin.version}</version>
+	    <version>${build-helper-maven-plugin.version}</version>
```
wetneb, line 49: It looks like you have some spurious indentation changes in this file. It would be good to avoid them. You can see the full list here: https://github.com/OpenRefine/OpenRefine/pull/4978/files

wetneb (commented):
Thanks!
Perhaps it would be good to already introduce a few localized messages in the backend as part of this pull request, to validate the process. Which sorts of messages do you intend to localize with this library, among the different areas you have identified already? I guess we should be able to identify some messages that should clearly be localized in the backend already, instead of being forwarded to the frontend and be localized there.

Also, for transparency it could be worth publishing your survey of localizing libraries (for instance on your Outreachy blog? just a random suggestion) as well as your survey of the places that could be localized in the backend.

main/pom.xml
elroykanye: Oh! I did not notice that. I will get to working on this

pom.xml
elroykanye: I thought as much. But when I remove this pluginRepository, I get this error:

```
[ERROR] Failed to execute goal org.jvnet.localizer:localizer-maven-plugin:1.31:generate (default) on project openrefine: Execution default of goal org.jvnet.localizer:localizer-maven-plugin:1.31:generate failed: 
Plugin org.jvnet.localizer:localizer-maven-plugin:1.31 or one of its dependencies could not be resolved: 
Failed to collect dependencies at org.jvnet.localizer:localizer-maven-plugin:jar:1.31 -> org.kohsuke:access-modifier-annotation:jar:1.4 -> org.jenkins-ci:annotation-indexer:jar:1.4: Failed to read artifact descriptor for org.jenkins-ci:annotation-indexer:jar:1.4: org.jenkins-ci:jenkins:pom:1.26 was not found in https://repo.maven.apache.org/maven2 during a previous attempt. 
This failure was cached in the local repository and resolution is not reattempted until the update interval of central has elapsed or updates are forced -> [Help 1]

```

Is there something I am missing out?

wetneb: Ok, of course that makes sense, I had not thought about transitive dependencies.

wetneb (approved):
Thanks a lot, and sorry about my hasty "fix" for the benchmark module…

elroykanye:
> Thanks a lot, and sorry about my hasty "fix" for the benchmark module…

Thanks to you too 😁 I did learn much through it.

PR #4979, opened 2022-06-22
Cleaned up the spacing between the Select and Deselect buttons in multiple file select panel

elebitzero:
- Cleaned up the spacing between the Select and Deselect buttons in multiple file select panel.

Before:
image

After:
image

### DIFF
--- a/main/tests/cypress/cypress/integration/project/grid/viewpanel-header/sort.spec.js
+++ b/main/tests/cypress/cypress/integration/project/grid/viewpanel-header/sort.spec.js
@@ -16,7 +16,6 @@ describe(__filename, function () {
     cy.get('.viewpanel-sorting a').contains('Sort').click();
     cy.get('.menu-container').contains('Reorder rows permanently').click();
     cy.assertNotificationContainingText('Reorder rows');
-    cy.reload();
     cy.getCell(0, 'Shrt_Desc').should('to.contain', 'BUTTER,WITH SALT');
     cy.getCell(1, 'Shrt_Desc').should('to.contain', 'BUTTER,WHIPPED,WITH SALT');
   });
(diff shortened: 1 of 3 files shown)


main/webapp/modules/core/scripts/index/default-importing-controller/file-selection-panel.js
```diff
@@ -223,7 +224,8 @@ Refine.DefaultImportingController.prototype._renderFileSelectionPanelControlPane
     $('<button>')
     .text($.i18n('core-buttons/deselect'))
     .addClass("button")
-    .appendTo($('<td>').appendTo(tr))
+    .css('margin-left','3px')
+    .appendTo(tr)
```
wetneb, line 228: Is it really okay to be appending an element directly inside a `<tr>`? If the problem comes from having the two buttons in separate `<td>` elements, perhaps it would be a bit better to append both elements in the same `<td>`?

wetneb (approved):

elebitzero:
I had to fix flaky tests in sort.spec by removing cy.reload() to get the CI check to pass.

main/tests/cypress/cypress/integration/project/grid/viewpanel-header/sort.spec.js
```diff
@@ -108,7 +105,6 @@ describe(__filename, function () {
     cy.get('.viewpanel-sorting a').contains('Sort').click();
     cy.get('.menu-container').contains('Reorder rows permanently').click();
     cy.assertNotificationContainingText('Reorder rows');
-    cy.reload();
```
wetneb, line 111: Interesting that this introduces flakiness! Intuitively those instructions are really part of the test: their purpose is to check that the reordering did indeed happen permanently (unlike the temporary sorting that is initially created). Perhaps there is a better way to assert this?

elebitzero: I added back `cy.reload()` and fixed the flakiness with `cy.waitForProjectTable()` since the reload is similar to `cy.loadAndVisitProject()`.

wetneb (approved):

PR #4987, opened 2022-06-25
547 handle gzip compressed files without gz extension

### LINKED ISSUE #547: Handle gzip compressed files without .gz extension
_Original author: tfmorris (March 08, 2012 00:28:29)_

Decompress Google Earth kmz files.

More generally, perhaps we should be sniffing the file to see if it's compressed (similar to the Unix file command using /etc/magic) or attempting to decompress all files so that we don't have to depend on a list of file extensions.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=547_

elroykanye:
Fixes #547 

- a method returning a boolean to check if a file is gzip compressed. This method is then used in the if branch for gzip compressions.
- renaming the exception object to ignored as nothing is done in that case.


main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -637,10 +639,19 @@ static public InputStream tryOpenAsArchive(File file, String mimeType, String co
...
         return null;
     }
+
+    private static boolean tarIsGZipped(File file) {
+        int magic = 0;
+        try (RandomAccessFile raf = new RandomAccessFile(file, "r");) {
+            magic = raf.read() & 0xff | ((raf.read() << 8) & 0xff00);
+        } catch (IOException ignored) {
+        }
+        return magic == GZIPInputStream.GZIP_MAGIC;
```
wetneb, line 653: This seems to be checking whether the file is gzipped, but that does not guarantee that the gzipped file is a tar archive, right? What happens when we run this with a simply gzipped file (without enclosing tar archive)?

elroykanye: Let me try that out :)

wetneb: Very nice, thanks a lot! Perhaps we can just add one or two tests and it should be good to go?

elroykanye: Yes! I am not sure how I will but I will :)
Probably get an idea from looking at the current tests.

wetneb (approved):
Works as expected, thank you!

PR #4992, opened 2022-06-27
Added fix to fit longer descriptions of suggest properties

### LINKED ISSUE #4988: Longer property description not fully appearing in schema builder
When searching for properties to build statements in the Wikibase extension's schema builder, longer property descriptions don't fully appear. Sometimes such descriptions are quite informative, and users can thus miss helpful info.

Example: 
image
I am quite curious what is said at the end of the description of P8791.

### To Reproduce
1. Start building a schema in the schema builder
2. Search for properties with a longer description, e.g. (partial) search terms `digitiz` and `depic` will provide good examples where the description is cut off in OpenRefine's interface
 
### Expected Behavior
A bit more space is built in, either horizontally or vertically or both, so that property descriptions are fully displayed.

elebitzero:
Fixes #4988

-  Search item description will now wrap to a 2nd line before truncation.
- Switched to non-minified suggest CSS.  (And deleted minified version)
- Adjusted search dropdown max-height to 400px from 320px to give more space for items with 2 line descriptions.
- Removed orange highlighting for the item description for suggest search selection.

This affects all places that are using the suggest dropdown menu, such as the Schema Editor, and Reconciliation Dialog.

After (Edit Schema View):
image

After (Recon Dialog):
image


wetneb (approved):
Thanks!
Now that the descriptions can span multiple lines, I find it a little bit harder to visually identify the names of the entities. Intuitively those should stand out a little more, so I would suggest a lighter shade of grey for the descriptions, or a darker one for the title, so as to increase the contrast between the two. Would it make sense to you?

trnstlntk:
OMG thank you!! :D

elebitzero:
We can make the label darker and/or description lighter. One complication is that the text matching on the keyword is highlighted in darker bold font.  I am concerned if we make the description light enough to stand out from the label, it might make it harder to read. 

Another option would be to make the label a bigger font.

Let me play with it and I can post some screenshots of different options tomorrow.

elebitzero:
Also, do you know why I get all these network requests with 404 Errors when I open the recon dropdown menu, like: https://wikidata.reconci.link/search?filter=(all%20mid:Q468700)&output=(notable:/client/summary%20(description%20citation%20provenance)%20type)&key=null

I am logged into Wikidata. Is there anything else I need to do?

wetneb:
This is a leftover from Freebase integration which is making spurious HTTP calls, as far as I know. There should be a way to clean this up so that those calls are not made.

elebitzero:
Below are the samples to distinguish the label from the description text in the following ways:
- Make Label Font 33% bigger than Description Font.  This does not require any more vertical space since the 1.2em font fits with the 1.4em line space.
- Make description font color gray #646464.
- Make description font color gray #747474.

My preference is the bigger font for the label.  I don't like making the description lighter because it reduces contrast.

Also, for the item in focus, it looks weird that the color of the description is changed during selection while the label isn't. I would just not change the text color at all since we already have a selection border and change the background of the item to white. Both of those should be enough to show selection.

image
image
image

trnstlntk:
Thanks so much for these 3 proposals @elebitzero. I'm not a designer myself, but intuitively I like option 2 best (only a very slight preference, not a very strong opinion) because I feel it fits most with current patterns in OpenRefine. @lozanaross Do you have any thoughts?

lozanaross:
Thank you all for comments and proposals. 
My thoughts are as follows: 
1) Definitely great to have 2 lines description (than 1)
2) I personally don't think it's a problem that the sizes of the entities and the descriptions are the same - white space plays a good role in differentiation + the Q-numbers, too.
3) I would neither make the description text lighter (for accessibility reasons - lower contrast will be problematic), nor make the entity text size bigger - the bigger size looks very clunky within the current margin / padding provisions (overwhelmingly tight in that small space and ultimately still not visually distinct enough). If we want to make that size bigger, we'll need to rethink the overall distribution of space in these suggestion menus, and that's a bigger job that requires thinking though use of space in the design system of the app as a whole.
4) What @elebitzero said here: "Also, for the item in focus, it looks weird that the color of the description is changed during selection while the label isn't. I would just not change the text color at all since we already have a selection border and change the background of the item to white. Both of those should be enough to show selection." > I completely agree - we don't need that color change, if anything better to change the color of the entity and not the description.

So TL;DR - I recommend to implement the original proposal of @elebitzero, don't implement any of the secondary options, but do change the color of the focus item to either not make the description orange, OR to make only the entity label orange.

elebitzero:
CI linux_server_tests failing with some unrelated error:
```
ERROR 2003 (HY000): Can't connect to MySQL server on '[12](https://github.com/OpenRefine/OpenRefine/runs/7210269069?check_suite_focus=true#step:9:13)7.0.0.1:49154' (111)
```

wetneb (approved):
Agreed, let's go with this as it stands!

PR #4997, opened 2022-06-29
Cleaning up and removing old freebase methods and references

### LINKED ISSUE #2435: Remove or replace the calls made to freebase.com
Context: Freebase reconciliation services have been shut down now. So, there's a need to clean up the code which is redirecting the call to freebase.com which will eventually cause an error 404. Code including freebase.com in it should be removed or replaced accordingly without disturbing any other functionality.

Such code can be found: https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/importers/ExcelImporter.java
https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/importers/OdsImporter.java

and many other places may be.

elebitzero:
Fixes #2435

- Replace Freebase credits with GitHub credits where applicable in about.html.
- Remove unused tests that mention Freebase.
- Remove invalid methods related to Freebase.


main/tests/server/src/com/google/refine/importers/RdfTripleImporterTests.java
```diff
@@ -64,115 +64,6 @@ public void setUp() {
...
-        Assert.assertEquals(project.rows.get(0).cells.get(1).value, "http://rdf.freebase.com/ns/en.blood_on_the_tracks");
-    }
-
-    @Test
-    public void canParseMultiLineTriple() throws UnsupportedEncodingException {
-        String sampleRdf = "<http://rdf.freebase.com/ns/en.bob_dylan> <http://rdf.freebase.com/ns/music.artist.album> <http://rdf.freebase.com/ns/en.blood_on_the_tracks>.\n"
-                +
-                "<http://rdf.freebase.com/ns/en.bob_dylan> <http://rdf.freebase.com/ns/music.artist.album> <http://rdf.freebase.com/ns/en.under_the_red_sky>.\n"
-                +
-                "<http://rdf.freebase.com/ns/en.bob_dylan> <http://rdf.freebase.com/ns/music.artist.album> <http://rdf.freebase.com/ns/en.bringing_it_all_back_home>.";
```
wetneb, line 94: Although those triples refer to Freebase, this tests the RDF importer, which is agnostic about which domains are used in the URIs it parses, so I would expect that this test should be kept.

For the sake of making clear that this is unrelated to Freebase integration, we could replace `rdf.freebase.com` by any other domain.

elebitzero: Alright, I added back the tests and changed the URLs.

wetneb (approved):

thadguidry:
This makes me a little sad today... like a child that lost her mother.  But the child has grown up to become healthy, independent, and full of life now.

PR #5025, opened 2022-07-06
API: make the get-importing-configuration command available over GET

Abbe98:
This makes the `get-importing-configuration` command accessible through GET requests rather than through `POST` requests as it does not change state.

Notes:

The `POST` option could maybe be safely removed considering that migration should be straightforward for potential API clients.


wetneb (approved):
Looks good to me. It is covered by Cypress tests already.

PR #5028, opened 2022-07-06
Add support for editing wikitext associated to MediaInfo entities

### LINKED ISSUE #4989: Wikimedia Commons: make Wikitext editable/overwritable
Relevant for OpenRefine 3.6+ in which Wikimedia Commons editing is possible.

In a previous meeting for the [Wikimedia Commons integration project](https://github.com/orgs/OpenRefine/projects/2/views/1), we discussed that it would be possible to make Wikitext of a file editable / overwritable. I then recommended to not enable this, mainly with the argument that I find it a bit scary (Wikitext is fiddly and it would not be good if users would, for instance, massively replace, remove or create faulty Wikitext, which could make the Commons community angry).

But based on user interviews / tests organized by @lozanaross (and based on my own experience and needs) I think I've changed my mind: users are actively asking to allow replacing existing Wikitext (to replace complex infobox templates with more simple Wikidata-driven ones), and we think OpenRefine / Wikimedia Commons users should generally be savvy enough to know what they are doing.

### Proposed solution
Let's add an 'override if present' check box for Wikitext in the Wikimedia Commons schema builder, which will trigger overwriting all existing Wikitext upon editing.

### Further considerations
I'm not sure yet if we will indeed want to include a separate field in the schema builder for Wikimedia Commons categories, or if we will skip this and ask the users to include categories in the Wikitext instead. If we add a separate field / functionality to add one or more Commons categories, for consistency an overwrite checkbox there would make sense too.

wetneb:
Closes #4989.

- let users edit wikitext of existing files
- add an override checkbox in the schema editor to specify whether the supplied wikitext should override the existing one (unchecked by default)

@elebitzero if you feel like reviewing the code changes, feel free (it's quite boring though, mostly passing a boolean around!)

Screenshot:
image


extensions/wikidata/src/org/openrefine/wikidata/editing/EditBatchProcessor.java
```diff
@@ -182,6 +183,14 @@ public void performEdit()
                 // Existing entities
                 EntityUpdate entityUpdate = update.toEntityUpdate(currentDocs.get(update.getEntityId().getId()));
                 editor.editEntityDocument(entityUpdate, false, summary, tags);
+                if (entityUpdate instanceof FullMediaInfoUpdate) {
```
wetneb, line 186: In this method we are increasingly relying on `instanceof` checks, that is a sign we need a different architecture I think. The existing one worked as long as we were doing all of our editing through Wikidata-Toolkit, but this is no longer the case with this. I think this can stay for now but I would welcome proposals for other architectures.

wetneb, line 186: Our increased reliance on `instanceof` checks here is a sign that we need a better architecture for this… suggestions welcome :)

wetneb:

PR #5031, opened 2022-07-07
Purge wikitext after making edits to MediaInfo entities

### LINKED ISSUE #4924: SDC: do null edit to trigger page update while adding structured data
[Category:INaturalist images missing SDC depicts](https://commons.wikimedia.org/wiki/Category:INaturalist_images_missing_SDC_depicts) is an example of a so-called 'maintenance category' for Structured Data on Wikimedia Commons; it indicates that certain files still lack a certain statement. I have recently added [depicts (P180)](https://www.wikidata.org/wiki/Special:EntityPage/P180) statements to many of these files, which should cause them to disappear from this category, but they keep showing up there. (I edited more than 20,000 files so I'd expect the category to still contain 20K to 30K files - not 50K.)

On the file side, though, the category itself doesn't appear with the individual files anymore ([example](https://commons.wikimedia.org/wiki/File:2014-08-19_Zebra_Finch,_Sumba,_Nusa_Tenggara_Timur,_Indonesia_3.jpg))

@multichill [told me](https://commons.wikimedia.org/wiki/User_talk:Multichill#Category:INaturalist_images_missing_SDC_depicts_may_need_a_poke) that this is due to [a bug on the Wikimedia Commons side](https://phabricator.wikimedia.org/T237991). SDC edits with bots and batch tools usually don't trigger a page refresh. [He advises us](https://commons.wikimedia.org/wiki/User_talk:Multichill#Category:INaturalist_images_missing_SDC_depicts_may_need_a_poke) to implement such a page refresh (by doing a null edit) in OpenRefine.

### Proposed solution
OpenRefine does a 'null edit' to trigger an update.

### Alternatives considered
Maybe this is not the most pressing issue (although maintenance categories on Wikimedia Commons are certainly really useful and it's nice if our edits become visible there). In the end, it's an upstream issue which should ideally be fixed on the Commons side though.

### Additional context
Phabricator ticket about the upstream issue on Wikimedia Commons: https://phabricator.wikimedia.org/T237991

wetneb:
Closes #4924.

- add API calls to purge the cache of the rendered wikitext, after edits to the structured data part of Commons files

This is really not pretty since we are adding yet another custom processing part on top of Wikidata-Toolkit. The corresponding bug in Wikibase is already quite old so it seems unlikely that it gets fixed soon, so I think there is still value in providing a temporary workaround.

Alternatively we could consider pushing this logic into Wikidata-Toolkit itself.


wetneb:

PR #5036, opened 2022-07-07
Assume that unknown datatypes are string-based.

### LINKED ISSUE #3263: Support for custom data types in OpenRefine-Wikibase integration
There is a problem when trying to work with custom data types in the OR-Wikibase integration extension.

### To Reproduce
1. First, reconcile data to Rhizome's custom Wikibase.
2. Then, build schema which includes property with custom data type (in this case – media file that is not Commons media file).
3. Finally, try to upload to Rhizome's Wikibase instance.

### Current Results
Right now I get a message that indicates the desired data type is unsupported when adding the new property to the schema. The corresponding value field goes red initially, but then it still allows me to drag a column with values into it and even renders a correct preview. I am then able to select Upload option, but get stuck at 0%.

### Expected Behavior
If this feature doesn't work, it should't even allow me to get to Preview and Upload stage. It would be great if custom data types were supported in the future, though. At least custom media files, as this is pretty fundamental feature to every independent Wikibase instance outside Wikidata.

### Screenshots
Screenshot 2020-10-14 at 14 46 18
Screenshot 2020-10-14 at 14 46 29
Screenshot 2020-10-14 at 14 46 42

### Versions
- Operating System: MacOS Catalina 10.15.6
- Browser Version: Google Chrome Version 84.0.4147.135 (Official Build) (64-bit)
- JRE or JDK Version: JRE 1.8.0_261
- OpenRefine: OpenRefine master branch as of 14.10.2020->

### Datasets
Property in question is here: https://artbase.rhizome.org/wiki/Property:P126

### Additional context
Thanks to a new extension built by Professional.wiki, Wikibase instances can now use a new data type for custom images that are not Commons media files. In the future even further data types could be added. This is not handled well within the current OR interface, but it is an important feature moving forward.

wetneb:
Fixes #3263.

Changes proposed in this pull request: Added support for custom Wikibase datatypes by assuming that their datavalue type is "string". Looking at the datatypes developed recently, this seems to be a reasonable assumption and should work with all extensions I am aware of.

To make this more robust, one should ideally lookup the datavalue type from the Wikibase instance (or store this information in the manifest). I could not find a way to lookup the datavalue type of a given datatype with the MediaWiki API.

### DIFF
--- a/extensions/wikidata/module/langs/translation-cs.json
+++ b/extensions/wikidata/module/langs/translation-cs.json
@@ -16,7 +16,6 @@
     "wikibase-schema/save-button": "Uložit schéma",
     "wikibase-schema/invalid-schema-warning-preview": "Vaše schéma není kompletní. Dokončete jej a uvidíte náhled.",
     "wikibase-schema/invalid-schema-warning-issues": "Vaše schéma není kompletní. Dokončete jej a uvidíte problémy.",
-    "wikibase-schema/datatype-not-supported-yet": "Tento datový typ bohužel ještě není podporován.",
     "wikibase-schema/geoshape-with-prefix": "název souboru začínající na \"Data:\"",
     "wikibase-schema/math-expression": "matematický vzorec",
     "wikibase-schema/commons-media": "Název souboru",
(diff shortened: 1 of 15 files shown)


wetneb:

PR #5042, opened 2022-07-08
Enable locating rows generating certain Wikibase issues with facets.

### LINKED ISSUE #5033: Provide facets to locate rows which generate certain quality assurance issues in the Wikibase extension
In the "Issues" tab, we report issues about Wikibase edits generated by the project grid and the schema. We give some examples of the values which generate the issues, but it would be much more convenient if users could simply get a facet which selects which rows generate each issue.

Not every issue can be tracked to a single row: for instance, the addition of two different identifier values on a given item is likely due to two different rows contributing those identifier values independently.

### Proposed solution

For issues which can be tracked back to a single row (such as string formatting issues), add a button on the issue to create a facet which selects the rows generating that issue. I propose the following internal architecture:
* the wikibase extension registers a new GREL function (for instance called `wikibaseIssues()`), whose task is to evaluate the schema on the current row, run scrutinizers on it, and return the list of issue identifiers that it generates
* the quality issues which are "facetable" are declared as such, and when they are, the UI displays a button to create the corresponding facet (for instance defined by `grel:wikibaseIssues().inArray('duplicate-whitespace')`)

### Alternatives considered

We could also try to extend this feature to issues which can be generated by multiple rows, but that is likely to require a more complicated architecture, so I would propose to leave that out first.

wetneb:
Closes #5033.

Not sure about the UX and language to use: suggestions welcome!

## Screenshots

image
Clicking on the button gives:
image


wetneb:

PR #5044, opened 2022-07-08
Semantics: change two buttons to links

Abbe98:
So that the security and accessibility features of the browser work as intended.

### DIFF
--- a/extensions/wikidata/module/scripts/dialogs/wikibase-dialog.html
+++ b/extensions/wikidata/module/scripts/dialogs/wikibase-dialog.html
@@ -11,7 +11,7 @@
   <div class="dialog-footer" style="justify-content: space-between" bind="dialogFooter">
       <div>
           <button class="button" bind="addButton"></button>
-          <button class="button" bind="discoverManifestsButton"></button>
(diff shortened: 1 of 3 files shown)


elebitzero:
I am wondering if we should just show the 'Discover Wikibase instances...' button as a link. I found when it is a button, my expectation is that it was going to fetch the list of manifests and display them in the webapp, whereas with a link I would expect it to open a new window (current behavior).
image

elebitzero:
Also, on further reflection, do we even need the 'Discover Wikibase instances...' button on the 'Select Wikibase' dialog,  because the link is already on the 'Add Wikibase' dialog ("manifest for some public Wikibases can be found here" )

image

Maybe we could just add the 'Discover Wikibases' link to the bottom of the 'Add Wikibase' dialog.

image

wetneb:
A link feels fine to me too.

> Maybe we could just add the 'Discover Wikibases' link to the bottom of the 'Add Wikibase' dialog.

I am not opposed to having it there as well, but I think there is also value in having it in the dialog where the existing Wikibase instances are listed, as it is perhaps more visible (no strong feeling though).

wetneb (approved):
Fine with me to do it step by step.

wetneb:
Merging now, with the understanding that converting those buttons to actual links (as proposed above) would be a sensible follow-up.

PR #5045, opened 2022-07-08
2443 set the locale based on userLang from preferences

elroykanye:
Related to #2443 

- set the locale based on the userLang value of the preferences from the workspace


main/src/com/google/refine/io/FileProjectManager.java
```diff
@@ -377,6 +378,13 @@ protected boolean loadFromFile(File file) {
         if (file.exists() || file.canRead()) {
 	        try {
 	        	ParsingUtilities.mapper.readerForUpdating(this).readValue(file);
+
+                PreferenceStore ps = this.getPreferenceStore();
+                if (ps.get("userLang") != null) {
+                    Locale locale = new Locale.Builder().setLanguage((String) ps.get("userLang")).build();
```
wetneb, line 384: Do you know if the language codes we use actually match those that are supported by `java.util.Locale`? I know there are subtle differences between various language identifier schemes, it would be good to double-check that Weblate and java.util.Locale agree.

wetneb (commented):
Very nice, thank you!
Would it make sense to also update the locale when the user changes language, using the `set-preference` command?

main/src/com/google/refine/io/FileProjectManager.java
elroykanye: I did not check for this, but I will. Thanks very much.

elroykanye: I did notice some differences in the codes for Japanese, Traditional/Simplified Chinese. Made some updates addressing this.

main/src/com/google/refine/util/LocaleUtils.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.Locale;
+import java.util.MissingResourceException;
+import java.util.ResourceBundle;
+
+public class LocaleUtils {
+    public static void setLocale(String code) {
+        String localeCode;
+        try {
+            ResourceBundle langResource = ResourceBundle.getBundle("language_codes");
+            localeCode = (String) langResource.getObject(code);
```
wetneb, line 12: Since most of the language codes are the same in both schemes, I wonder if it would make sense to only store the mappings where it differs (suggested by @nerdyProjects who had a glimpse at my screen, thank you!), and adapt the code accordingly (if the mapping does not exist, keep the same value).

elroykanye: Thanks to you and @NerdyProjects 🙏🏾 I'll do just that

elroykanye:
> Interesting, so the test failures are indeed consistent! Curious. Let me know if you need help to debug that.

Oh yes please, I do need help

main/src/com/google/refine/util/LocaleUtils.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.Locale;
+import java.util.Map;
+
+public class LocaleUtils {
+    private final static Map<String, String> languageCodeMap = Map.of(
+            "bn", "ben",
+            "jp", "ja",
+            "zh", "zh_CN",
+            "zh_Hant", "zh_TW"
+    );
```
wetneb, line 12: The initialization of `languageCodeMap` is not run before `setLocale` is run, probably because of the specifics of in which order Java executes static initialization.

Instead, I would recommend you initialize `languageCodeMap` to `null`, and properly initialize it (to `Map.of(...)`) the first time it is needed. Alternatively, because this map is rather small and we are not expecting `setLocale` to be run many times, it is probably also fine not to have it as a static variable, but a local one inside of `setLocale()`.

wetneb:
@elroykanye do you want to debug the next failure? I think you should be able to reproduce this locally by running the tests on your side. It should be a similar initialization problem (perhaps the PreferenceStore is not initialized when you are trying to use it?)

main/src/com/google/refine/util/LocaleUtils.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.Locale;
+import java.util.Map;
+
+public class LocaleUtils {
+    public static void setLocale(String code) {
+        if (code == null) {
+            return;
+        }
+        Map<String, String> languageCodeMap = Map.of(
```
wetneb, line 11: It could be worth explaining why this translation is needed (what are the language code schemes used on both sides, how to find out if more mappings should be added). By the way, it could be worth mapping for all the cases where the language codes differ in those schemes, even if those languages are not supported by OpenRefine yet, just to be on the safe side (we will probably not remember to edit this when new languages are added).

wetneb (commented):
Well done! Some suggestions for improvements below.

main/src/com/google/refine/util/LocaleUtils.java
elroykanye: The language code scheme for java's locale system follows [ISO 639-2](https://www.loc.gov/standards/iso639-2/php/code_list.php).

I tried accessing the docs to retrieve that for weblate but noticed that there was some issue with the [docs website](https://docs.openrefine.org/)

wetneb: Sorry about that, I just fixed the docs, it should be up in a few minutes.

wetneb (approved):
LGTM!

PR #5049, opened 2022-07-09
Improved Wikibase selection UI

### LINKED ISSUE #5009: More UI improvements for Wikibase selection dialog
In the (resolved) issue #3824, the Wikibase selection dialog window has been improved.

image

@lozanaross and I have tested the "new style" window in v3.7 while observing people during Wikimedia Commons upload user tests, and we notice that its behavior is still confusing to folks, especially the way in which Wikibase instances change place in the window as a new one is selected (the active one will jump to the top). A UX where instances are e.g. in alphabetical order, and the active one is highlighted otherwise, would feel more natural/intuitively understandable.

### Proposed solution
This design by @lozanaross is probably still a better way to go (or do we already have a newer version?). Radio button to select the desired active instance, cross to delete an instance, `Add another instance` button to add a new one. Instances stay in the same place/order and don't switch around after selection/addition.
image

### Additional context
Also see #5007 - may be good to integrate a link to the Wikibase manifest repository as well?

I may miss elements that are relevant for better UX of this dialog, so feel free to chime in with more factors to take into account.

elebitzero:
Fixes #5009

Changes to 'Select Wikibase' dialog proposed in this pull request:
- Added radio button to show selection.
- Added display label: 'active' for the currently targeted Wikibase.
- The user cannot delete the 'active' and currently 'selected' Wikibase. 
- 'Cancel' will ignore any selection change, and close the dialog. The targeted Wikibase does not change.
- 'OK' confirms the selection and closes the dialog.
- Increased the max height of the Wikibase list to 500px so more items are visible.

After:
image


wetneb (approved):
Looks really good to me, thank you so much!
There is a small merge conflict with #5044, but that will be solved when we merge the PRs.

lozanaross (commented):
Thanks for adding all these improvements! It looks really good to me and I'm happy to approve, but just 2 quick questions: 1) Is there some convention in how dialog buttons in OpenRefine are ordered. I wonder if we need both "Apply" and "OK" or if just one of these is enough (what's the difference between the two operations?). 2) I wonder if we need to spell out "active" once the Wikibase is selected with a radio button - if it's selected and then the user hits the button "Apply" or "OK" then we know it's going to be "active". Keeping "Delete" there instead of "active" will just retain the consistency of the Delete operation which should be possible even if something is currently selected rather than asking users to select something else first before being able to delete a service. 

But these are minor points, and I don't mind if others overrule my suggestions here.

elebitzero:
For the order of the buttons, most dialogs have 'Ok', and 'Cancel', but a few have it the other way.

The reason for the 'Apply' button is so the user can switch the active Wikibase and then delete the old one without closing the dialog.  If you do not have 'Apply', then you have to select the new Wikibase, click 'OK', and reopen the dialog to delete the old Wikibase.

'Active' denotes the currently active Wikibase. The active Wikibase doesn't change until you select a new Wikibase and click 'OK' or 'Apply'.

lozanaross:
> For the order of the buttons, most dialogs have 'Ok', and 'Cancel', but a few have it the other way.

OK, good to know. I think for future we can try to standardize that, so the order is always consistent.

> The reason for the 'Apply' button is so the user can switch the active Wikibase and then delete the old one without closing the dialog. If you do not have 'Apply', then you have to select the new Wikibase, click 'OK', and reopen the dialog to delete the old Wikibase.

I think I understand that, but I'm afraid introducing 2 steps to do something (first clock on the radio button area & then apply it) could be confusing to users (based on observing how users do things during user testing sessions). Maybe @trnstlntk could give feedback here as well.

> 'Active' denotes the currently active Wikibase. The active Wikibase doesn't change until you select a new Wikibase and click 'OK' or 'Apply'.

I think the reason to introduce the radio button is exactly to make it obvious that you're selecting something in order to activate it immediately. I thought in the past versions of the dialog window deletion was possibly even when something was active, so i thought if we keep the "delete" button always there even on an active wikibase users can perform all actions they want to in one go - i.e. delete an instance, add new instance, select an instance, etc. However, my technical understanding of how the code works is limited, so if it's not possible to keep the delete option always there, let's go with this current proposal. My only concern was to keep it simplest as possible for users, therefore avoiding the need to "apply" the selection and just having one button to "OK" all operations in the dialog.

elebitzero:
> I think the reason to introduce the radio button is exactly to make it obvious that you're selecting something in order to activate it immediately.

The user is selecting immediately, not activating immediately.  Referring to your mockup, you had a button 'Confirm Selection' which implies the user has to confirm the selection to activate it.

I will simplify it for now by removing the Apply button.  Let's go with this version now, and once we get additional feedback, we can adjust it further.

wetneb:
@elebitzero feel free to merge when this is ready

lozanaross:
> The user is selecting immediately, not activating immediately. Referring to your mockup, you had a button 'Confirm Selection' which implies the user has to confirm the selection to activate it.

Yes, I used 'confirm selection' to imply "OK the selection" - the dialog window is called "Select Wikibase instance", so I thought we should give users some way to confirm their expectations as to what happens in this dialog (before there was no button of this kind at all). What I meant was indeed just an "OK" button.

> I will simplify it for now by removing the Apply button. Let's go with this version now, and once we get additional feedback, we can adjust it further.

Great, let's try that - we'll do some user testing and we'll see if it works or we need some further adjustments.

extensions/wikidata/module/styles/dialogs/wikibase-dialog.less
```diff
@@ -65,11 +65,15 @@
   color: gray;
 }
 
-.wikibase-list li a {
+.wikibase-list li a,span {
```
wetneb, line 68: @elebitzero it looks like there is a small error here: this CSS selector seems to be understood as "`.wikibase-list li a` OR `span`" (which is probably not what you intended). This means that the style instructions in it are applied to all spans in OpenRefine (even outside of this dialog). This results in unusually wide margins in quite a few places, for instance here:
image
Both Chromium and Firefox seem to agree on the interpretation of this CSS selector.

PR #5052, opened 2022-07-10
drop legacy CSS

Abbe98:
Drop various legacy CSS features/prefixes.

These changes should only affect browsers older than:

 - IE 10
 - Safari 13
 - Firefox 50
 - Chrome 56

### DIFF
--- a/main/webapp/modules/core/styles/index/open-project-ui.less
+++ b/main/webapp/modules/core/styles/index/open-project-ui.less
@@ -147,7 +147,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 /* Making metadata table header sticky and darker*/
 #metadata-body table > tbody > tr >th{
-	position: -webkit-sticky;
 	position:sticky;
 	top:0;
 	color:black;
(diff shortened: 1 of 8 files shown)


main/webapp/modules/core/externals/jquery-ui/css/ui-lightness/theme.css
```diff
@@ -172,20 +172,14 @@ a.ui-button:active,
 .ui-widget-content .ui-priority-secondary,
 .ui-widget-header .ui-priority-secondary {
 	opacity: .7;
-	-ms-filter: "alpha(opacity=70)"; /* support: IE8 */
```
wetneb, line 175: This seems to be an external library that we have copied into the repository, I would rather not make changes to it here. If there is a newer version of it we could upgrade to it.

elebitzero: Hi,

JQuery UI is generated from the website https://jqueryui.com/download/ , so we cannot add it to the package.json. We are already using the latest version (1.13.1.), so I don't think we should modify those files unless it is causing an issue.

Abbe98: A mistake on my side, I did not intend to modify external dependencies.

PR #5053, opened 2022-07-11
Wait a little before reloading in Cypress sort test

### LINKED ISSUE #5051: Unstable Cypress test for the sorting functionality
We regularly have CI failures due some Cypress tests which test our sorting functionalities:
* https://github.com/OpenRefine/OpenRefine/runs/7269100211?check_suite_focus=true
* https://github.com/OpenRefine/OpenRefine/runs/7246541652?check_suite_focus=true

We should fix those tests or OpenRefine itself so that this does not happen.

wetneb:
Closes #5051.

The test sometimes fails apparently because some of the UI update code gets interrupted by the `cy.reload()` action. Intuitively such errors should be ignored by Cypress, but it is not currently the case.

I am not happy with this solution, but it is the only thing I could think of, because I do not see what DOM element I could wait on instead. It does seem to make the test more reliable though. Any other ideas very much welcome!

### DIFF
--- a/main/tests/cypress/cypress/integration/project/grid/viewpanel-header/sort.spec.js
+++ b/main/tests/cypress/cypress/integration/project/grid/viewpanel-header/sort.spec.js
@@ -16,6 +16,7 @@ describe(__filename, function () {
     cy.get('.viewpanel-sorting a').contains('Sort').click();
     cy.get('.menu-container').contains('Reorder rows permanently').click();
     cy.assertNotificationContainingText('Reorder rows');
+    cy.wait(250); // wait for the rest of the update code to run before reloading, for #5051 (TODO: find a better way?)
     cy.reload();
     cy.waitForProjectTable();
     cy.getCell(0, 'Shrt_Desc').should('to.contain', 'BUTTER,WITH SALT');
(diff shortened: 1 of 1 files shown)


elebitzero:
Yes, these tests are flaky, but they haven't failed on my local setup, only in the CI tests.  Is it reproducible for you locally?

If the exception is due to cy.reload() interrupting the UI, you can add the same check on the cell data to ensure the table data is reloaded before calling cy.reload().

wetneb:
Yes I can reproduce it locally, by running this single test spec with `cypress open`.

> If the exception is due to cy.reload() interrupting the UI, you can add the same check on the cell data to ensure the table data is reloaded before calling cy.reload().

The problem is that we are not expecting any change in grid data here (since the sorting configuration is already active before it is made permanent). The error is happening when updating the undo/redo tab, to add the history entry there, so I guess one option could be to switch to that tag, wait for the entry to appear there, and then reload.

antoine2711:
> Yes, these tests are flaky, but they haven't failed on my local setup, only in the CI tests. Is it reproducible for you locally?
> 
> If the exception is due to `cy.reload()` interrupting the UI, you can add the same check on the cell data to ensure the table data is reloaded before calling `cy.reload()`.

@elebitzero: for me, I have different results when I run the tests on my machine or virtual machines.
And if I run my tests from the visual, I'm more likely to see errors. If I run them from the command line, they usually succeed all. But, when I have errors, it's not at the same place.

I really think our test are VERY vulnerable to timing issues.

Regards, Antoine

wetneb:
Merging to see if it makes a difference to stability on the long run.

PR #5056, opened 2022-07-13
Dialog footer cleanup

elebitzero:
- Updated Upload edits to Wikibase Dialog to use a dialog footer.
- Cleaned up other dialog footers that were using tables to be more consistent.

Upload Edits to Wikibase:

Before:
image

After:
image


wetneb (approved):
Looks great to me!

PR #5058, opened 2022-07-14
Updated dependabot to check dependencies in the main webapp.

elebitzero:
- Updated dependabot to check dependencies in the main webapp.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,6 +28,11 @@ updates:
     - "< 2"
   target-branch: 4.0
     
+# For main webapp
+- package-ecosystem: "npm"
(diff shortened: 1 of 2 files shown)


wetneb:
I suspect we also need to add a `package-lock.json` file to the repository, no?

elebitzero:
I don't think the package-lock.json is required.  The other package.json that Dependabot is checking in docs/ and main/tests/cypress do not have package-lock.json checked in either.

wetneb:
At least that is what I am seeing in the GitHub UI:

image

elebitzero:
I added the package-lock.json.

wetneb (approved):

PR #5065, opened 2022-07-15
Added ignore to Dependabot for Cypress dependency

elebitzero:
- Added ignore to Dependabot for the Cypress dependency since we do not want to upgrade until Cypress adds back 'Run all tests,' and based on the comments of the maintainer, 'Run all tests' will not be back for another 6 months to a year.


wetneb (approved):
Thank you!

PR #5095, opened 2022-07-20
2443 backend localization operation descriptions/history entries

elroykanye:
Related to #2443
- add properties files for operation descriptions and operation history entries 
- localise messages and change to reference methods from localizer generated classes
-

### DIFF
--- a/main/src/com/google/refine/operations/cell/FillDownOperation.java
+++ b/main/src/com/google/refine/operations/cell/FillDownOperation.java
@@ -72,8 +72,9 @@ public class FillDownOperation extends EngineDependentMassCellOperation {
     protected String createDescription(Column column,
             List<CellChange> cellChanges) {
         
-        return "Fill down " + cellChanges.size() + 
-            " cells in column " + column.getName();
+        // return "Fill down " + cellChanges.size() +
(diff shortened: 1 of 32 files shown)


main/src/com/google/refine/operations/cell/FillDownOperation.java
```diff
@@ -63,15 +64,17 @@ public FillDownOperation(
...
     }
 
     @Override
     protected String createDescription(Column column,
             List<CellChange> cellChanges) {
         
-        return "Fill down " + cellChanges.size() + 
-            " cells in column " + column.getName();
+        // return "Fill down " + cellChanges.size() +
+        //    " cells in column " + column.getName();
```
wetneb, line 76: I would say it is worth removing those comments before merging.

wetneb (commented):
Very nice, the tests all pass!

main/src/com/google/refine/operations/cell/FillDownOperation.java
elroykanye: On it ✌🏾

wetneb (approved):

Issue #5043, opened 2022-07-08
Ability to save "schema templates": schemas where some values are left empty


wetneb:
Wikibase communities generally establish some data modelling conventions, which users are asked to follow when importing data.
For instance, a book on Wikidata will generally have a certain set of statements on it (title, author, date of publication…).

When importing data in a Wikibase instance, users need to have the knowledge of those modelling conventions. If they do not have this knowledge, a blank schema page can be quite daunting.

### Proposed solution
We should make it possible for users to define "schema templates" which represent a prototypical Wikibase schema for an import in a certain domain. Those schema templates would be incomplete Wikibase schemas, in the sense that some statement values could be left blank, for instance. After loading such a template, the user could then drag and drop the columns from their project to the appropriate locations in the schema.

For the implementation, we would just use the same classes as the ones that define schemas, but relax the validation in the constructors to allow for empty value. We would then add methods to check that a given schema is complete. This would be the opportunity to return more precise error messages when the schema is incomplete (#4724).

### Alternatives considered
Use [ShEx](https://en.wikipedia.org/wiki/ShEx) or [SHACL](https://en.wikipedia.org/wiki/SHACL). Those formats are not designed for the Wikibase data model, but rather for RDF. This means that it would not be possible to take a schema represented in one of those formats, and render it as a Wikibase schema with holes. Therefore this does not seem fitting here.

### Additional context
Requested by @trnstlntk @lozanaross in the context of the Wikimedia Commons integration project.

johnsamuelwrites:
The proposal of “schema templates”, especially for the [Wikimedia Commons Integration project](https://meta.wikimedia.org/wiki/Grants:Project/CS%26S/Structured_Data_on_Wikimedia_Commons_functionalities_in_OpenRefine) is interesting. I understand the requirement where the users would propose “incomplete” schemas or “schemas with holes” suiting their particular or temporary requirements. We have a number of such use cases. For example, different language communities may describe a hospital in different ways, suiting their local needs. Hence, during [WikiTechstorm 2019](https://diff.wikimedia.org/2020/01/07/wikimedian-john-samuel-and-his-experiences-at-wiki-techstorm-2019/), we proposed to simplify the shape expressions (ShEx) in a way that helps users easily create schemas for their needs and validate their data with their own proposed schema (for example, the local definition of a hospital with the required properties). This is also very important for lexicographical data, where we required different schemas for lexemes of multiple languages.

[...]

Issue #4724, opened 2022-04-09
Saving the Wikidata Schema sometimes gives an error message that is imprecise and unhelpful. The problem should be describe instead.


antoine2711:
We have non-negligeable problems in the Wikidata Schema’s validation process. A lot a times, even if the schema is good (what's visually in our face), the validation fails with a genereic and very unhelpful error message that can make a normal user grow some impatiance trying to find why it failed.

This is also for the error when saving a schema. Sometime, when there's problems, it just says it can't save the schema.

### Proposed solution
A clue to what might actually be not working not just a simple and obvious « I won't do it ».
I would like to know EXACTLY where the problem is.

- if it's the data, which column (and maybe the index of fautive rows?)
- if it's a schema field, which one? Item # + Term ID or Statement #, Reference # and/or Value # if also applicable.
- I could live with only the next error.

### Alternatives considered
Either search everything till it works, or cry…

### Additional context
[Please note that I did experience that issue a lot and it might have an impact on how I describe this issue.]

### Screenshots
![image](https://user-images.githubusercontent.com/12217525/162548603-027cb82e-e726-429b-9b70-1649a9d1b1de.png)

Aizko:
I have been going through some of these issues lately myself. The screenshot above (default at the start of the schema) is the smallest of them, but gives a clue, with no clear idea of what is going wrong really. More clues as to the error would be appreciated in order to put a remedy.

antoine2711:
> More clues as to the error would be appreciated in order to put a remedy.

@Aizko: I would go as far as to say that to DEBUG this problem, more precise error messages are going to be mandatory. It’s too obscure in the front-end code.

Regards, Antoine

Aizko:
Thanks for the hints, Antoine. This is not my worst issue right now with OpenRefine, the worst being I cannot upload the project despite progress starting, but actually blocked (see the googlegroups conversations). Will remember this conversation if I see vague messages appearing again. See the [googlegroups conversations](https://groups.google.com/g/openrefine/c/oOSg-0MelBQ). Regards

WolfgangFahl:
Personally i think this part of OpenRefine is what makes it a showstopper to be used for mere mortals.

PR #5096, opened 2022-07-20
Backend support for incomplete schemas

wetneb:
This is first step towards those two issues:
* #5043
* #4724 

It decouples the parsing of the schema (done by the Jackson library, which calls constructors of our classes which make up the schema) and its validation. So far, the constructors of each of those classes were checking that they had valid (non-empty) arguments.

By delaying this to an explicit validation step, we are enabling two things:
* to return better error messages which can point to the location in the schema where a value is missing. I plan to expose this in the frontend in a follow-up PR (which will also tackle related issues in the schema UI)
* to make is possible to represent "schemas with holes", also known as "schema templates". The alternative would have been to declare another tree of classes to represent templates, but that feels overkill to me. Some follow-up PRs will expose this in the frontend (saving, loading, deleting those schema templates).

Since this PR is already rather big I am submitting it as is (although it does not address any of the issues yet), to make it easier for anyone who would want to review this to do so.

### DIFF
--- a/extensions/wikidata/src/org/openrefine/wikidata/commands/SaveWikibaseSchemaCommand.java
+++ b/extensions/wikidata/src/org/openrefine/wikidata/commands/SaveWikibaseSchemaCommand.java
@@ -26,6 +26,8 @@ package org.openrefine.wikidata.commands;
 import static org.openrefine.wikidata.commands.CommandUtilities.respondError;
 
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Properties;
 
 import javax.servlet.ServletException;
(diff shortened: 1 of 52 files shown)


extensions/wikidata/src/org/openrefine/wikidata/commands/SaveWikibaseSchemaCommand.java
```diff
@@ -55,23 +58,28 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
             AbstractOperation op = new SaveWikibaseSchemaOperation(schema);
             Process process = op.createProcess(project, new Properties());
 
             performProcessAndRespond(request, response, project, process);
-
-        } catch (IOException e) {
-            // We do not use respondException here because this is an expected
-            // exception which happens every time a user tries to save an incomplete
-            // schema - the exception should not be logged.
-            respondError(response, String.format("Wikibase schema could not be parsed: ", e.getMessage()));
```
wetneb, line 74: We used to catch all IOExceptions and assume they were due to the schema being invalid, but that was a very sketchy approximation: now schema invalidity errors are handled separately, above.

PR #5099, opened 2022-07-21
Lint all source files, not just the test ones.

### LINKED ISSUE #5097: Java linting not enforced on source files
The linting of our Java source files seem to work only for our test files, not our main source files.
For instance, a lot of our source files have mixed tabs and spaces.

This seems to be due to the non-standard source directories we are using.

We should add those source directories in the POM files to make sure they are linted. This will require fixing the corresponding linting issues, and rebase / merge a lot of pull requests… Sorry about that, I should have checked that when I introduced the linting in the first place!

wetneb:
Closes #5097.

I apologize in advance for the mess this will cause to pull requests, I am keen to wait for big PRs to be merged before this.

Most (but not all) changes are re-indentation so `git merge -Xignore-space-change` should make this a lot more manageable than it looks.


wetneb:
This is also discussed on the mailing list: https://groups.google.com/g/openrefine-dev/c/fw_iFbSNz9o

wetneb:
I have just rebased this PR on master and the option to ignore whitespace changes made the process fairly straightforward: I did have some merge conflicts to solve, but they were non-trivial. Therefore I think it should be manageable for other PRs to also adapt after this one is merged.

PR #5112, opened 2022-07-25
4928 xml import fails if the file contains invalid xml characters

### LINKED ISSUE #4928: XML import fails if the file contains invalid XML characters
When trying to parse a file which contains invalid characters in XML, we get an error. Instead it would be nice if those characters could be ignored, perhaps via an importer option.

### To Reproduce

Import the following file: 
[expert_groups.zip](https://github.com/OpenRefine/OpenRefine/files/8839692/expert_groups.zip)

### Current Results

Error message: `javax.xml.stream.XMLStreamException: ParseError at [row,col]:[296,30402]
Message: Un caractère XML non valide (Unicode : 0xb) a été détecté dans la section CDATA`

### Expected Behavior

The dataset could be imported anyway

### Versions
 - Operating System: Debian
 - Browser Version: irrelevant
 - JRE or JDK Version: 11
 - OpenRefine: 3.6-SNAPSHOT

### Datasets

[expert_groups.zip](https://github.com/OpenRefine/OpenRefine/files/8839692/expert_groups.zip)

### Additional context
Similar issue to make the XML importer more accepting: #3638

elroykanye:
Fixes #4928 

- A private method in the nested class `XmlParser` in `com.google.refine.importers.XmlImporter` to remove invalid characters from an xml input stream

These changes are to fix occurrence of invalid characters in an xml file, which include those not specified in [XML recommendation 1.0, §2.2 Characters](http://www.w3.org/TR/xml/#charsets).


main/src/com/google/refine/importers/XmlImporter.java
```diff
@@ -213,8 +216,16 @@ static public class XmlParser implements TreeReader {
...
         public XmlParser(InputStream inputStream) throws XMLStreamException, IOException {
+            inputStream = removeInvalidCharacters(inputStream);
             parser = createXMLStreamReader(inputStream);
         }
+
+        private InputStream removeInvalidCharacters(InputStream inputStream) throws IOException {
+            String in = IOUtils.toString(inputStream, StandardCharsets.UTF_8);
+            Pattern p = Pattern.compile("[^\\u0009\\u000A\\u000D\\u0020-\\uD7FF\\uE000-\\uFFFD\\u10000-\\u10FFF]+");
+            String result = p.matcher(in).replaceAll("");
+            return IOUtils.toInputStream(result, StandardCharsets.UTF_8);
```
wetneb, line 227: I wonder if we could do this in a streaming fashion instead, to avoid loading the whole file in memory?

wetneb: Note: there could also be the possibility of configuring the parser so that it ignores invalid characters itself (ideally). If it's not supported then there might be utilities to replace characters in an InputStream in a streaming fashion.

wetneb:
I thought a bit more about this and I suspect you could do the replacement on a `Reader` object, not on the `InputStream` itself. If the XML parser requires an `InputStream` then you could convert the replaced Reader back to an InputStream ([with commons-io](https://www.baeldung.com/java-convert-reader-to-inputstream))

elroykanye:
> I thought a bit more about this and I suspect you could do the replacement on a `Reader` object, not on the `InputStream` itself. If the XML parser requires an `InputStream` then you could convert the replaced Reader back to an InputStream ([with commons-io](https://www.baeldung.com/java-convert-reader-to-inputstream))

I have been checking on this actually but I have not had any luck because the XMLStreamReader used is not compatible with IOUtils or the ReaderInputStream.

elroykanye:
> I thought a bit more about this and I suspect you could do the replacement on a `Reader` object, not on the `InputStream` itself. If the XML parser requires an `InputStream` then you could convert the replaced Reader back to an InputStream ([with commons-io](https://www.baeldung.com/java-convert-reader-to-inputstream))

Hello @wetneb , please check the recent updates on this. I have an alternative approach which involves a static InvalidXmlFilterReader class extending the FilterReader. I haven't really tried it out, but I just wish to get your feedback on this first.

main/src/com/google/refine/importers/XmlImporter.java
elroykanye: > Note: there could also be the possibility of configuring the parser so that it ignores invalid characters itself (ideally). If it's not supported then there might be utilities to replace characters in an InputStream in a streaming fashion.

I'll swear to you that getting to understand how to do this perfectly and implement took me over 2 weeks of continuous trials 😆 

I managed to get the processing done by replacing invalids with a '?' (I do not know if this is okay or if there is an alternative way to handle this).

wetneb: Very nice, that looks really good to me!
Replacing invalid characters with `?` makes sense, but I am thinking that removing them entirely could be preferable in some situations. But no strong feeling on this.

main/src/com/google/refine/importers/XmlImporter.java
```diff
@@ -216,7 +219,45 @@ static public class XmlParser implements TreeReader {
...
+
+            @Override
+            public int read() throws IOException {
+                int c = super.read();
+                if (c == -1) return c;
+                while (pattern.matcher(String.valueOf((char) c)).matches()) {
+                    c = super.read();
+                }
+                return c;
+            }
```
wetneb, line 249: This `read()` method is inconsistent with the other one: the one above removes invalid characters, the one below replaces them by `?`. So this might lead to unstable behaviour, depending on which of the methods is used. I think both methods should have a consistent behaviour (perhaps with a slight preference for removing, since it is perhaps more likely that this produces valid XML markup).

main/src/com/google/refine/importers/XmlImporter.java
```diff
@@ -216,7 +219,45 @@ static public class XmlParser implements TreeReader {
         static final int WHITESPACE_CHARACTERS_TOKEN = 15;
 
         public XmlParser(InputStream inputStream) throws XMLStreamException, IOException {
-            parser = createXMLStreamReader(inputStream);
+            parser = createXMLStreamReader(new InvalidXmlFilterInputStream(inputStream));
+        }
+
+        static class InvalidXmlFilterInputStream extends BufferedInputStream {
+
+            private final Pattern pattern = Pattern.compile("[^\\u0009\\u000A\\u000D\\u0020-\\uD7FF\\uE000-\\uFFFD\\u10000-\\u10FFF]+");
```
tfmorris, line 227: Many of these characters are legal in the CDATA elements of XML 1.1 documents https://www.w3.org/TR/xml11/#NT-Char

tfmorris (changes requested):
These changes appear to affect ALL XML parsing, both XML 1.0 and XML 1.1, whether or not the parse would produce an error. Please consider either not changing this code or restricting the changes to only XML 1.0 documents which produces errors on the first attempt.

In addition to the functional risk of making global changes to such a critical code path, there's also the performance impact of introducing a regex into an input path that regularly handles many MiB of data.

wetneb:
Hi @elroykanye, it looks like this PR is stalled by the issues above. I am closing it as such, but if you are still working on this, feel free to reopen.

PR #5121, opened 2022-07-26
Update openrefine-api.md

VladimirAlexiev:
- fix link name of refine-ruby
- fix link of refine-java (the old one is a redirect)
- add ontorefine-cli


docs/docs/technical-reference/openrefine-api.md
```diff
@@ -279,7 +279,8 @@ Libraries using the [OpenRefine API](openrefine-api):
 * [openrefine-php-client](https://github.com/keboola/openrefine-php-client)
 
 ### Java {#java}
-* [refine-java](https://github.com/dtap-gmbh/refine-java)
+* [refine-java](https://github.com/ancore/refine-java)
+* [ontorefine-cli](https://github.com/Ontotext-AD/ontorefine-cli) that is a fork of the above for Ontotext's OntoRefine
```
wetneb, line 283: Since this seems to be specific to OntoRefine, I am not sure it has its place here. If ontorefine-cli actually works for OpenRefine too, perhaps it would be worth advertising it as such on the repository first?

VladimirAlexiev: AFAIK all additions of ontorefine-cli on top of refine-java are specific to OntoRefine.

I thought this is a place to list all Refine clients, like the Recon testbench lists all Recon endpoints?
What if we change the description to `a fork of the above adapted for OntoRefine` ?

wetneb: I would keep ontorefine-cli out of this list if it is specific to OntoRefine, but maybe others think differently

PR #5124, opened 2022-07-27
Switched the icon for removing a URL from trash-icon to the x icon

elebitzero:
-  Switched the icon for removing a URL from the trash icon to the x icon in create a project with URL:

image


wetneb (approved):
Seems intuitively fine to me (but no strong views).

main/webapp/modules/core/scripts/index.js
```diff
@@ -68,24 +68,7 @@ Refine.postCSRF = function(url, data, success, dataType, failCallback) {
...
-//		lang : lang
-	},
-	success : function(data) {
-		dictionary = data['dictionary'];
-        lang = data['lang'];
-	}
-});
-$.i18n().load(dictionary, lang);
-$.i18n({ locale: lang });
-// End internationalization
```
wetneb, line 88: That change seems to be unrelated to the change of icon, so perhaps it would be cleaner to have it as a separate PR, no?

elebitzero: I meant to put that change in a separate PR, I have reverted it and added some slight improvements (see updated screenshot).

Issue #5118, opened 2022-07-25
Japanese Translation File Naming


AtesComp:
Japanese translation files erroneously use "jp" in the file name.  The "jp" designation is a country code.  They should use "ja" instead as this designation is the language code.

### To Reproduce
1. See the translation files for the main program and extensions.  For Japanese, they are "translation-**jp**.json".
2. Consult the Internet on language versus country code for Japanese.

See: https://www.andiamo.co.uk/resources/iso-language-codes/
See: https://news.ycombinator.com/item?id=5047300

wetneb:
Agreed. This language code was introduced in bedb8c03bf7d1a80a6e978ff8dc10c5ff8a98384, before we used Weblate, and seems to be wrong indeed. We could rename this. Weblate would probably handle this just fine, but on OpenRefine's side I think it would be good to introduce a fallback mechanism to make sure that people using OpenRefine in japanese currently do not get an error when they upgrade. Maybe in the load-language command, for instance?

isaomatsunami:
Wow. The very first translator got off on the wrong foot! (I just wanted to use this phrase. Thanks AtesComp / Antonin!)

AtesComp:
Right! I noticed this when I had a translation submission to the RDF Transform extension using a "-ja.json" file.  I couldn't understand why it wasn't loading until I saw the "jp" files.

Could the load-language code report a warning that the current language wasn't found and show the expected 2 char language code and defaulting? Server side? It's a small corner case but could help in the debug process.

Also, supporting a "debug" preference in OR for log messages could be helpful...maybe for 4.0.

Also, is the browser's navigator.language (and languages array) a default based on the country code given the IP's geolocation?  This would imply the default language code--we would need to translate a country code to a language code.  Also, are there country codes that match other language codes? If so, there is another problem if we don't strongly differentiate country code input from language code.  I'll look at the related code.

wetneb:
> Also, is the browser's navigator.language (and languages array) a default based on the country code given the IP's geolocation? This would imply the default language code--we would need to translate a country code to a language code. Also, are there country codes that match other language codes? If so, there is another problem if we don't strongly differentiate country code input from language code. I'll look at the related code.

I think it is much simpler than that: you do not need any IP geolocation or mapping countries to languages. Most browsers simply expose the locale they prefer using the `Accept-Language` header, such as `Accept-Language: en-GB,en;q=0.5`.
[...]

PR #5125, opened 2022-07-27
Cleaned up duplicate i18n code

elebitzero:
- Move duplicate code to initialize the i18n system to i18n.js so it can be used by the index.vt, preferences.vt, and project.vt templates.  
- Removed duplicate code in extensions.
- Updated controller.js to include i18n.js in the script bundles for each template.
- Added check to compare the browser's language to the one returned by the load-language command, and print a warning if they are different.  Requested in Issue #5118.

### DIFF
--- a/main/webapp/modules/core/scripts/util/i18n.js
+++ b/main/webapp/modules/core/scripts/util/i18n.js
@@ -6,7 +6,7 @@ var dictionary = "";
 /*
    Initialize i18n and load message translation file from the server.
 
-   Note that the language is set by the 'userLang' user preference setting.  You can chnage that by
+   Note that the language is set by the 'userLang' user preference setting.  You can change that by
    clicking on 'Language Settings' on the landing page.
 */
 I18NUtil.init = function (module) {
@@ -15,8 +15,7 @@ I18NUtil.init = function (module) {
         type: "POST",
         async: false,
         data: {
-            module: module,
-            lang: lang
+            module: module
         },
         success: function (data) {
             dictionary = data['dictionary'];
--- a/main/webapp/modules/core/scripts/util/i18n.js
+++ b/main/webapp/modules/core/scripts/util/i18n.js
@@ -1,0 +1,38 @@
+I18NUtil = {};
+
+var lang = (navigator.language || navigator.userLanguage).split("-")[0];
+var dictionary = "";
+
+/*
+   Initialize i18n and load message translation file from the server.
+
+   Note that the language is set by the 'userLang' user preference setting.  You can chnage that by
+   clicking on 'Language Settings' on the landing page.
+*/
+I18NUtil.init = function () {
+    $.ajax({
+        url: "command/core/load-language?",
+        type: "POST",
+        async: false,
+        data: {
+            module: "core"
+        },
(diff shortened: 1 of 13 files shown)


AtesComp:
For `main/webapp/modules/core/scripts/util/i18n.js`
The older `navigator.userLanguage` only applies (AFAICT) to the deprecated Internet Explorer (IE), so we should probably remove.
The newer `navigator.languages` (plural) exists so we should likely add it and use the first (primary?) language.

Example:
```
// Get document language code...
//    navigator.userLanguage  removed, IE Only
var lang = (navigator.language || navigator.languages[0]).split('-')[0];
...
```

wetneb (commented):
Very sensible refactoring, thank you! I wonder if it would make sense to do it in the extensions included in this repository, too?

elebitzero:
> The older navigator.userLanguage only applies (AFAICT) to the deprecated Internet Explorer (IE), so we should probably remove.

@AtesComp,  I wasn't sure if OpenRefine still supported IE, but I just checked the documentation, and it says IE is not supported. (Ref: https://docs.openrefine.org/manual/installing#compatible-browser )

So I will remove the navigator.userLanguage check.

For navigator.languages ,  if the browser has navigator.languages, wouldn't they have set navigator.language? That seems like an unnecessary fallback.

elebitzero:
@wetneb, I removed the duplicate code in extensions as well.

main/webapp/modules/core/scripts/util/i18n.js
```diff
@@ -0,0 +1,39 @@
...
+   clicking on 'Language Settings' on the landing page.
+*/
+I18NUtil.init = function (module) {
+    $.ajax({
+        url: "command/core/load-language?",
+        type: "POST",
+        async: false,
+        data: {
+            module: module,
+            lang: lang
```
wetneb, line 19: This `lang: lang` binding seems to have been commented out in most of the blocks that this new function is replacing. Are you clear on what we are changing by including this parameter?

elebitzero: I looked at the implementation of core/command/load-language in LoadLanguageCommand.java, and it looks like if 'lang' is not passed it defaults to the user preference 'userLang'. If we always pass in 'lang', then the user preference 'userLang' is never used.  If people were using the 'userLang' preference to test different locales then this PR would break that.

I will remove passing the 'lang' parameter since most of the i18n init code was not passing 'lang.' I'm not sure how the 'userLang' preference was being set, and how it is working for locales other than English.

It makes sense to use the browser's locale, but if people are installing OpenRefine on their own personal desktops it might not be an issue.

AtesComp: Yep, I saw that also and was changing it so that the lang passed in was a suggestion and the stored preference overrides.  Otherwise, why set a language preference in OR if the browser overrides.

I was waiting for your finished code before sending mine in.

wetneb: Makes sense, let's merge this and then @AtesComp can have a go at tweaking the logic :)

wetneb (approved):

PR #5131, opened 2022-07-30
Better error reporting in Wikibase schema

### LINKED ISSUE #4724: Saving the Wikidata Schema sometimes gives an error message that is imprecise and unhelpful. The problem should be describe instead.

wetneb:
Fixes #4724.

This adds explicit error messages indicating which parts of the schema are incomplete.
For now they can only be partially localized, but once backend localization is introduced in the Wikidata extension, the error messages that are generated there could also be localized too.

## Screenshot

image


elebitzero (approved):
1. If I add statement, and leave the property empty and click 'Save Schema', I see an exception. (Screenshot#1):
image

2.  If I add a term, and leave it empty and then click 'Save Schema', it appears to do it (Screenshot#2), but the previous behavior would not save and displays an error message "Your schema is incomplete so it cannot be saved yet" (Screenshot#3):
image

image

wetneb:
Thanks a lot, this was very sloppy on my side!

In terms of user experience, I am not completely satisfied with the current situation. The validation errors are displayed in the Issues and Preview tab, but will users look for them there? Should we change the message in the alert to indicate that users can go to the issues tab to see the validation errors? Alternatively, we could remove this alert and simply switch users to the Issues tab whenever they try to save an incomplete schema, perhaps?

wetneb:
Since this is getting stale, let's just merge it and deal with any improvements in follow-up PRs.

lozanaross:
> In terms of user experience, I am not completely satisfied with the current situation. The validation errors are displayed in the Issues and Preview tab, but will users look for them there? Should we change the message in the alert to indicate that users can go to the issues tab to see the validation errors? Alternatively, we could remove this alert and simply switch users to the Issues tab whenever they try to save an incomplete schema, perhaps?

I think a better alert message would be preferable to just unexpetedly switching users to another tab. Something like: "You schema is incomplete so cannot be saved yet. Go to the Issues tab to review validation errors."

PR #5146, opened 2022-08-03
Fix facetability of Wikibase issues

### LINKED ISSUE #5144: "Locate offending rows" does not work
Clicking on the "Locate offending rows" button in any Wikibase issue does not have any effect.

Also, the warning below should not have a "Locate offending rows" button because the issue is not always generated by a single row in the project:
image

Not writing more here because I intend to fix this directly.

wetneb:
Closes #5144

- fix frontend code for setup of the facet (not sure how the current bug made it through my testing…)
- fix facetability of Wikibase issues (not being preserved by merge, and absence of issues should not be facetable either)

### DIFF
--- a/extensions/wikidata/module/langs/translation-en.json
+++ b/extensions/wikidata/module/langs/translation-en.json
@@ -285,5 +285,5 @@
     "wikibase-statement-settings/strategy-options/property/description": "Compare statements by their main property only.",
     "wikibase-issues/no-warning": "No warning",
     "wikibase-issues/warning-raised": "Warning raised",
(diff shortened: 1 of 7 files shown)


wetneb:
Still to do on this pull request:
- [ ] Remove the "offending" adjective: for issues with "INFO" warning level, it feels weird to consider those rows "offending", since there might not be any problem with them. So the label of the button should just be "Locate rows"
- [ ] The buttons also show up in the dialog to upload edits to the Wikibase instance. Either they should be removed, or clicking on them should close the dialog and take the user to the grid.

wetneb:
Merging this soon, let me know if you can think of things to improve.

PR #5148, opened 2022-08-04
Update Japanese translation filename and related language corrector code

AtesComp:
Fixes https://github.com/OpenRefine/OpenRefine/issues/5118

- Renaming Japanese Translation files from *-jp.json to *-ja.json
- Modify LoadLanguageCommand.java adding temporary CORRECTOR code to handle an existing userLang parameter in the PreferenceStore converting "jp" to use the proper "ja" language code
- Modify LoadLanguageCommand.java to prefer the PreferenceStore's userLang over the request language code
  - Treat the request language (extracted from the navigator) as a last resort when no OpenRefine user preference is provided.
- Clean LoadLanguageCommand.java for error handling
- Clean incidental whitespace in language related .js files


main/src/com/google/refine/commands/lang/LoadLanguageCommand.java
```diff
@@ -113,32 +130,46 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
         try {
-            Reader reader = new BufferedReader(new InputStreamReader(new FileInputStream(langFile), "UTF-8"));
-            return ParsingUtilities.mapper.readValue(reader, ObjectNode.class);
-        } catch (FileNotFoundException e1) {
+            fisLang = new FileInputStream(langFile);
+        } catch (FileNotFoundException e) {
             // Could be normal if we've got a list of languages as fallbacks
-        } catch (IOException e) {
-            logger.error("JSON error reading/writing language file: " + langFile, e);
+            logger.warn("Language file " + strMessage + " not found", e);
```
wetneb, line 156: Perhaps an `info` level would be more appropriate here as it is an expected scenario: for instance, if I use OpenRefine in Japanese and some extension has not been translated to Japanese, it is normal to fall back on English for that extension.

wetneb (commented):
Very nice, thank you! Just some thoughts about logging, see below.

wetneb:
Ah, and of course we would need to make sure the tests pass :)

main/src/com/google/refine/commands/lang/LoadLanguageCommand.java
AtesComp: Yeah, I'm onboard with that.

AtesComp:
Yep, I need to manage / update the tests for the new process.  On it.

AtesComp:
I keep forgetting to do the formatting.  I've updated the code, added two tests, and modified one test.

elebitzero (approved):

main/src/com/google/refine/commands/lang/LoadLanguageCommand.java
```diff
@@ -79,32 +79,52 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+
+        // Always replace suggested with user preference language, if available...
+        PreferenceStore ps = ProjectManager.singleton.getPreferenceStore();
+        if (ps != null) {
+            String strLang = (String) ps.get("userLang");
+            // If user preference language exists...
+            if (!(strLang == null || strLang.isEmpty())) {
+
+                // CORRECTOR...
+                // TODO: This code may be removed sometime after the 3.7 release has been circulated.
```
elebitzero, line 96: The logic looks good to me. Although the code under the TODO should probably be kept in case people wait until a version later than 3.7 to update.

PR #5149, opened 2022-08-04
Added the ability to submit the form data by pressing enter on the 'add column based on this column' dialog

### LINKED ISSUE #5143: In the "Add column based on this column" dialog, validating a form by pressing Enter should run the operation
When adding a column based on a column, I often first type the expression I want, and then set a column name in the field at the top, and then press Enter, expecting it to have the same effect as clicking the "OK" button. It currently does not do anything.

### Proposed solution
This field should be part of a form, and when that form is submitted, that should be equivalent to pressing the OK button.
This should not require setting up any event listener on keyboard events, but instead integrate the existing input fields in a form, and binding the submission event on that form to the execution of the operation (and closing the dialog).

### Alternatives considered
None.

### Screenshot
image

elebitzero:
Fixes #5143

- Added the ability to submit the form data by pressing enter on the 'add column based on this column' dialog.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/add-column-dialog.html
+++ b/main/webapp/modules/core/scripts/views/data-table/add-column-dialog.html
@@ -22,7 +22,7 @@
       </table></div>
     </div>
     <div class="dialog-footer" bind="dialogFooter">
-      <button class="button" bind="okButton"></button>
+      <button class="button" type="submit" bind="okButton"></button>
       <button class="button" bind="cancelButton"></button>
     </div>
   </form>
--- a/main/webapp/modules/core/scripts/views/data-table/add-column-dialog.html
+++ b/main/webapp/modules/core/scripts/views/data-table/add-column-dialog.html
@@ -1,6 +1,7 @@
 <div class="dialog-frame" style="min-width: 700px;">
   <div class="dialog-border">
     <div class="dialog-header" bind="dialogHeader"></div>
+    <form bind="form">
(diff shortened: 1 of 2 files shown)


main/webapp/modules/core/scripts/views/data-table/add-column-dialog.html
```diff
@@ -24,5 +25,6 @@
       <button class="button" bind="okButton"></button>
       <button class="button" bind="cancelButton"></button>
     </div>
+  </form>
```
wetneb, line 28: Thanks a lot! I wonder if wie need a 'type="submit"' in the "OK" button to keep it working?

elebitzero: It is working, but I added the type="submit" to the ok button since that is standard.

wetneb: Interesting, do you understand why it was working without? Does the browser interpret the "OK" label by itself, to understand that the form should be submitted? I find that surprising!

elebitzero: I checked and it looks like the type defaults to submit if inside a form.

Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type

wetneb (approved):
Thanks a lot as always!

PR #5158, opened 2022-08-07
Support for file upload to Wikibase instances without structured data support

### LINKED ISSUE #5039: Add support for uploading local media files on third-party Wikibases without structured data
OpenRefine should be able to upload files to Wikibase instances which are their own media repository. By this, I mean that the media files which can be referenced in Wikibase statements are stored in the same MediaWiki instance as the Wikibase itself.

In such instances, the media files typically do not have structured data associated with them, as that data can already be represented in the item namespace if needed.

### Proposed solution

Repurpose the existing Commons integration, which already includes support for file upload.
* [ ] adapt the Wikibase manifest format to make it possible to keep file upload while disabling support for MediaInfo entities
* [ ] potentially it would be useful to remove the need for a reconciliation service entirely, using #5038. This would make sense because using the reconciliation service to fetch extra data is not super useful if structured data is not enabled on the wiki (although the wikitext could potentially be of interest)
* [ ] make sure the resulting filenames can be used as values in media-valued statements on the Wikibase instance. This requires adapting the edit scheduler to detect the relationship between the upload of the file and its use in a statement.

### Alternatives considered
Not sure!

### Additional context

@lozanaross your input would be very welcome to frame this issue better (I hope I have understood it correctly!)

wetneb:
Fixes #5039.

- Rework the backend to avoid trying to fetch MediaInfo entities when we are not uploading any structured data. This enables working with the existing Wikibase integration with "virtual" MediaInfo entities, even when the target Wikibase instance does not actually have Wikibase entities of that type, but only plain old MediaWiki files. This is shoehorning a new sort of file upload feature in the Wikibase extension (#5039), even if it does not have anything to do with Wikibase in the end.
- Add a field to the manifest to disable the structured data fields on media files (captions and statements). When this flag is set in the Wikibase manifest, we simply offer the file name, file path and wikitext fields in the schema editor, and nothing else.
- Add documentation of this feature.

This supports both uploading new files and editing the wikitext of existing files.

To test this integration, you need:
- a Wikibase instance which has file upload enabled
- an instance of the [Commons recon service](https://gerrit.wikimedia.org/r/q/project:labs%252Ftools%252Fcommons-recon-service) running for that Wikibase instance, with [this patch](https://gerrit.wikimedia.org/r/c/labs/tools/commons-recon-service/+/820900) applied

I have deployed [a test instance](http://bubbletea.reconci.link/wiki/Main_Page) (which will disappear in 10 days) that you can also use to test this functionality.

### DIFF
--- a/docs/docs/manual/wikibase/configuration.md
+++ b/docs/docs/manual/wikibase/configuration.md
@@ -71,6 +71,7 @@ Here is the manifest of Wikimedia Commons:
        "reconciliation_endpoint": "https://commonsreconcile.toolforge.org/${lang}/api"
     }
   },
+  "hide_structured_fields_in_mediainfo": false,
   "editgroups": {
     "url_schema": "([[:toollabs:editgroups-commons/b/OR/${batch_id}|details]])"
   }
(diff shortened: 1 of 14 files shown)


wetneb:
I have been testing this but have hit an issue in Wikidata-Toolkit, so we might need to cut out a new release of Wikidata-Toolkit with that fix (https://github.com/Wikidata/Wikidata-Toolkit/pull/717) before this can really be tested thouroughly.

Issue #3017, opened 2020-07-30
Dark mode


MavropaliasG:
Hi, it would be fantastic to have a dark mode, it will make OpenRefine look more modern as well.

thadguidry:
@MavropaliasG Yes I agree! How fast can you hack on that and submit a pull request? :-)

Taruun:
> @MavropaliasG Yes I agree! How fast can you hack on that and submit a pull request? :-)

Can i work on that?

MavropaliasG:
> Can i work on that?

That would be amazing. I found that https://darkreader.org/ creates a nice dark mode, can we access the CSS that it produces and just create a dark mode from that? Is it going to be easier?

wetneb:
@Taruun thanks for volunteering, please go ahead!

eddo888:
I too would love dark mode, is there a simple tick box on a menu we can try ? not wanting to get too into the whole CSS world at the moment.

PR #5165, opened 2022-08-09
Dark mode and theming support #3017

Abbe98:
**This is very much a work in progress and has a long way to go.**

The intent of this pull request is to bring theming and in particular dark-mode support(#3017) to the OpenRefine GUI.

 - [x] migrate LESS variables to CSS
 - [ ] consolidate and reduce shades and colors used in OpenRefine (*wip* / *input wanted*)
 - [ ] rename and make variable names color independent (*wip*)
 - [ ] ensure all colors are exposed as CSS variables (*wip*)
 - [ ] migrate raster images to vectors (*help wanted*) 
 - [ ] decide on a color scheme for dark mode (*help wanted*)
 - [ ] migrate extensions to use the built-in theme options (*not started*)

Expected side effects:

 - changes to GUI colors as a result of consolidating the 10+ shades of greys, 10+ shades of blue/white, etc
 - a period in between versions in which non-core extensions might not adhere to theming
 - improved contrast and overall accessibility
 - ...

**Note that the dark-mode theme in place as of the opening of this pull request is only a proof of concept.**

### DIFF
--- a/extensions/database/module/styles/database-import.less
+++ b/extensions/database/module/styles/database-import.less
@@ -53,7 +53,7 @@
 
 .database-importing-wizard-header {
   font-size: 1.3em;
-  background: @chrome_primary;
+  background: var(--chrome-primary);
   padding: @padding_tight;
   }
 
@@ -76,7 +76,7 @@
   font-size: 1.3em;
   position: absolute;
   overflow: auto;
-  border-top: 5px solid @chrome_primary;
+  border-top: 5px solid var(--chrome-primary);
   background: white;
   padding: @padding_looser;
   }
--- a/extensions/gdata/module/styles/importing-controller.less
+++ b/extensions/gdata/module/styles/importing-controller.less
@@ -48,7 +48,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
   }
 
 .gdata-document-container > table {
-  color: var(--metadata-grey);
+  color: var(--text-tertiary);
 }
 
 a.gdata-doc-title {
--- a/extensions/gdata/module/styles/importing-controller.less
+++ b/extensions/gdata/module/styles/importing-controller.less
@@ -48,11 +48,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
   }
 
 .gdata-document-container > table {
-  color: @metadata_grey;
+  color: var(--metadata-grey);
 }
 
 a.gdata-doc-title {
-  color: @link_primary;
+  color: var(--link-primary);
   text-decoration: none;
   }
 a.gdata-doc-title:hover {
@@ -66,7 +66,7 @@ a.gdata-doc-preview {
   font-size: 80%;
   text-decoration: none;
   vertical-align: super;
-  color: @link_secondary;
+  color: var(--link-secondary);
   margin-left: 0.5em;
   }
 a.gdata-doc-preview:hover {
@@ -81,7 +81,7 @@ a.gdata-doc-preview:hover {
 
 .gdata-importing-wizard-header {
   font-size: 1.3em;
-  background: @chrome_primary;
+  background: var(--chrome-primary);
   padding: @padding_tight;
   }
 
@@ -104,7 +104,7 @@ a.gdata-doc-preview:hover {
   font-size: 1.3em;
   position: absolute;
   overflow: auto;
-  border-top: 5px solid @chrome_primary;
+  border-top: 5px solid var(--chrome-primary);
   background: white;
   padding: @padding_looser;
   }
--- a/extensions/wikidata/module/styles/dialogs/wikibase-dialog.less
+++ b/extensions/wikidata/module/styles/dialogs/wikibase-dialog.less
@@ -34,7 +34,7 @@
 }
 
 .wikibase-list li:not(.selected):hover {
-  background: @dialog_footer;
+  background: var(--dialog-footer);
 }
 
 .wikibase-list li img {
--- a/main/webapp/modules/core/styles/common.less
+++ b/main/webapp/modules/core/styles/common.less
@@ -34,6 +34,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 @import-less url("theme.less");
 
 :root {
+  --text-primary: black;
+  --text-secondary: #222;
   --link-primary: #11c;
   --link-secondary: #4272db;
   --chrome-primary: #bcf;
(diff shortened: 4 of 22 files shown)


Abbe98:
For those of you who would like to help work on the color scheme you can play with the colors by inspecting the page in your browser and changing the color of variables to see changes to all elements using that particular variable. If you do please share screenshots of the UI! :1st_place_medal: 
Screenshot from 2022-08-09 16-24-53

Abbe98:
One area in which I would very much appreciate help would be to migrate raster images to vector ones that could be themed, this is particularly relevant for various icons. 

Screenshot 2022-08-09 at 16-31-56 pattypan slm - OpenRefine

tfmorris:
> One area in which I would very much appreciate help would be to migrate raster images to vector ones that could be themed, this is particularly relevant for various icons.

Is that more complicated than just grabbing the star, flag, etc icons from something like FontAwesome?
https://fontawesome.com/icons/caret-down?s=solid
https://fontawesome.com/icons/star?s=solid
https://fontawesome.com/icons/flag?s=solid

wetneb:
I guess ideally it would be good to keep a consistent visual style across the application, so using symbols from the same library everywhere… I know we have departed from that already quite a bit though :-/

Abbe98:
@tfmorris yeah, using one single library like FontAwesome is the way to go for icons, however, there are some other assets that would need work(the OpenRefine logo with text comes to mind(should maybe even be reproduced with plain HTML)).

elebitzero:
Hey @Abbe98,

Instead of one big PR, you could organize it into smaller PRs which build on each other. This way, the work can be checked in incrementally, which has its advantages.  It is easier to review smaller PRs.  And most of the work is preparation which has no visible change. It is easy to spot if something is wrong if everything is supposed to look the same.

PR #1 parameterize the colors for existing light theme, replace LESS with CSS vars

What is the reason to use CSS vars? Does that make it easier to switch themes at runtime?
This brings up a separate topic whether we should just get rid of LESS and use CSS vars, since all modern browsers support the feature.  (Ref: https://caniuse.com/css-variables)

PR #2 consolidate and reduce shades and colors used in OpenRefine for light theme

PR #3 migrate raster images to vectors for light theme

The jQuery-ui bundle was generated using light-theme which has an icon set for the light-theme.
It also has a dark theme, and an alternative icon set for the dark theme.  We could just use that, rather than trying to replace all the icons with different icons.

For the few icons (star,flag,etc.) that are custom pixel icons, we could switch them to something else. JQuery-ui has a star and flag icon.

Ref: https://jqueryui.com/themeroller/

PR #4 dark mode theme, user preference to select theme, mechanism in code to switch themes.

Abbe98:
> This way, the work can be checked in incrementally, which has its advantages. It is easier to review smaller PRs.

That's what I imagine, this branch being cherry-picked and merged multiple times. For development five out of the seven tasks are very dependent on each other. For example, migrating all colors to variables before reducing the colors used would create a lot of extra work, initial dark-mode supports help with "debugging", etc.

I would imagine your suggested merge 1 and merge 2 being one as 2 reduces the work needed for 1 and vice versa. CSS variables should also preferably be available one release earlier than theming to at least give extensions a chance to be compatible from the start.

> What is the reason to use CSS vars? Does that make it easier to switch themes at runtime?

Yes, that's the one and only reason.

> This brings up a separate topic whether we should just get rid of LESS and use CSS vars

I think it's a great follow-up PR as we do use some other LESS features(to a minimal extent) unrelated to this effort.

> The jQuery-ui bundle was generated using light-theme which has an icon set for the light-theme.
It also has a dark theme, and an alternative icon set for the dark theme. We could just use that, rather than trying to replace all of the icons with different icons.

I think this would make custom theming very difficult as jQuery UI is entirely driven by raster image sprites, so we would end up hardcoding a binary set of background images into variables rather than just working with colors.

Sadly I think we are also stuck with jQuery UI overrides for general theming as it has no native support for CSS variables or prefers-color-scheme.

trnstlntk:
> I guess ideally it would be good to keep a consistent visual style across the application, so using symbols from the same library everywhere… I know we have departed from that already quite a bit though :-/
> 
> @lozanaross any suggestions?

One idea, not sure if this makes sense: Wikimedia's [OOUI](https://www.mediawiki.org/wiki/OOUI) icons have undergone a lot of scrutiny from WMF design teams and it's a vast collection - https://commons.wikimedia.org/wiki/OOUI_icons / https://doc.wikimedia.org/oojs-ui/master/demos/?page=icons&theme=wikimediaui&direction=ltr&platform=desktop#CheckboxMultiselectWidget

wetneb:
> That's what I imagine, this branch being cherry-picked and merged multiple times. For development five out of the seven tasks are very dependent on each other.

Since this branch is not meant to be merged in master, I think we do not need keep a PR open for it.
Feel free to open PRs for the sub-parts as suggested by @elebitzero.
If the intention is to provide a discussion space around your work, I suggest to use our brand new Discourse forum.

Therefore closing this (but I am very much supportive of all the tasks mentioned here and the overall goal).

Abbe98:
I'm reopening this as it otherwise get really tricky for others to follow and various Github features break on closed merge requests(checks, conflicts, etc).

wetneb:
Fine with me, but let's mark it as draft then.

smierz:
would it make sense to convert the list of tasks into an issue with sub-issues?
As a draft PR all the important information here are hard to find

tfmorris:
That's similar to what @elebitzero asked here https://github.com/OpenRefine/OpenRefine/pull/5165#issuecomment-1210258482 to which @Abbe98 replied https://github.com/OpenRefine/OpenRefine/pull/5165#issuecomment-1210463367.  (although they were talking about PRs, not issues)

I'll also note that these tasks likely overlap with some of the comments/ToDos in [OpenRefine Design System](https://openrefine.org/docs/technical-reference/openRefine-design-system) that @Lydiaofficial and @lozanaross put together on [Figma](https://www.figma.com/file/2fGTsMdExFEf4ReUNFzB0V/OpenRefine-Outreachy?type=design&node-id=131%3A315&mode=design&t=sqE5Yl0QJtw5TAHy-1).

A work breakdown would be great. I'm not sure whether starting from a 3 yr old PR would be productive. Perhaps the To-Do list instead? 

@Abbe98 is there an associated issue (or set of issues)? #6538 is one, but I don't see anything else relevant [here](https://github.com/OpenRefine/OpenRefine/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22Theme%3A%20UI%2FVisual%22)

Abbe98:
We implemented this downstream ~2 years ago so I will look my commits and share what we did. I know the larger effort was around some thirdparty dependencies that had their own GUIs that OpenRefine simply embedded.

tfmorris:
I've created #7585 per @smierz's suggestion.

PR #5169, opened 2022-08-10
add properties files in french

### LINKED ISSUE #2443: Back-end localization

elroykanye:
add properties files in french

Fixes #2443 

- Noticed some alerts when adding new components for the properties files and there was some issue with conflicting default and *_en.properties, so I am bringing this as an attempt to clear it up

### DIFF
--- a/main/resources/com/google/refine/operations/OperationHistoryEntry.properties
+++ b/main/resources/com/google/refine/operations/OperationHistoryEntry.properties
@@ -1,5 +1,0 @@
-row_flag=Flag {0} rows
-row_unflag=Unflag {0} rows
-row_removal=Remove {0} rows
-row_star=Star {0} rows
-row_unstar=Unstar {0} rows


elroykanye:
Closing this as I just realised there was a misunderstanding.

PR #5172, opened 2022-08-12
Migrated substr calls to substring

elebitzero:
- Migrated substr calls to substring.
Ref:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr


main/webapp/modules/core/scripts/util/date-time.js
```diff
@@ -203,7 +203,7 @@ DateTimeUtil.setIso8601 = function (dateObject, string){
                 offset = (Number(d[3]) * 60) + Number(d[5]);
                 offset *= ((d[2] == '-') ? 1 : -1);
             }
-            comps[1] = comps[1].substr(0, comps[1].length - d[0].length);
+            comps[1] = comps[1].substring(0, comps[1].length - d[0].length);
```
tfmorris, line 206: This page https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#the_difference_between_substring_and_substr says that the second argument should be end position, not length

wetneb: In this case the start position is 0, so the end position should be equal to the length of the desired substring, no? Or do you think this is a problem in other places, potentially?

elebitzero: @tfmorris, I had the same doubts as you and had to double-check.  But it is correct because the substrings starts at 0.

from: 0
`length: comps[1].length - d[0].length`

start: 0
`end: 0 + comps[1].length - d[0].length`

image

As an aside, I was looking into possibly removing the date-time.js file since it is only called in one place in cell-ui.js as a backup if Date.parse() fails to parse a date string.

wetneb: > As an aside, I was looking into possibly removing the date-time.js file since it is only called in one place in cell-ui.js as a backup if Date.parse() fails to parse a date string.

I haven't looked into the problem but that sounds like a sensible move - I was actually not aware we had any clever date parsing code in the frontend because I thought we were requiring the user to input dates in ISO format… If something more flexible is available then we need to make that clear in the UI.

elebitzero:
I submitted another PR to remove date-time (#5182).

wetneb (approved):

main/webapp/modules/core/scripts/util/date-time.js
tfmorris: > In this case the start position is 0, so the end position should be equal to the length of the desired substring, no? Or do you think this is a problem in other places, potentially?

Oops! Sorry, you're right.

Issue #488, opened 2012-10-15
ISO 8601 dates not supported in cell editing - cell-ui.js


tfmorris:
_Original author: tfmorris (November 18, 2011 22:21:30)_

Dates formatted in ISO 8601 format can't be converted to a date by the individual cell editor cell-ui.js  This is ironic since that's the format that converts dates to when creating strings.

An example which doesn't work 1987-08-01T03:00:00Z

Apparently Javascript 5 will include better parsing, but until then, the Date.parse() needs to be replaced with something else here:

```
} else if (type == &quot;date&quot;) {
  value = Date.parse(text);
  if (!value) {
    alert(&quot;Not a valid date.&quot;);
    return;
  }
  value = value.toString(&quot;yyyy-MM-ddTHH:mm:ssZ&quot;);
```

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=488_

PR #5182, opened 2022-08-15
Removed unneeded date-time.js file

elebitzero:
The date-time.js file was added to provide IS08601 date parsing when browsers did not fully support ES5's Date.parse specification. Today's browsers support ECMAScript 1-6, so the workaround is no longer necessary.

I looked up the original issue, and found the date string that was failing ("1987-08-01T03:00:00Z").  I tried that with the date-time.js workaround removed, and it still worked.

So, I think it's safe to remove this code.

Ref: https://github.com/OpenRefine/OpenRefine/issues/488

- Removed unneeded date-time.js file


wetneb (approved):
Thanks a lot for the very thorough analysis!

PR #5188, opened 2022-08-16
Fetch URL throttle delay update

### LINKED ISSUE #5173: Is the 5000ms default delay for the fetch URL operation too high?
I had feedback from a user who initially did not notice the <s>1000 ms</s> 5000 ms delay in the "Add column by fetching URLs" operation, and had to wait a long time for the operation to complete.

### Proposed solution

There is a case for saying that <s>1000 ms</s>5000 ms is perhaps a bit too high for a default value. It makes sense to have a relatively generous wait by default, so that users who are not aware of the problems caused by sending too many requests at a time do not accidentally hammer services.

### Alternatives considered

Add some contextual help to highlight this field (which is perhaps not so visible). This could be an occasion to educate users about responsible crawling, perhaps?

### Additional context

Perhaps this is less of an issue when we are able to show partial resrults for long-running operations (so that users do not have to wait for the completion of the operation to keep working on the project).

elebitzero:
Fixes #5173

- Changed default throttle input delay for create column by fetching URLs to 500ms.
- Added an error message for submitting a delay less than the minimum:
image


thadguidry (approved):
tested and works as expected in Edge.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -567,6 +567,7 @@
     "core-views/store-err": "store error",
     "core-views/cache-responses": "Cache responses",
     "core-views/copy-val": "copy value from original column",
+    "core-views/warning-throttle-delay-input": "Throttle delay must be 200ms or greater.",
```
tfmorris, line 570: I don't believe this is correct.

main/webapp/modules/core/scripts/views/data-table/menu-edit-column.js
```diff
@@ -158,15 +158,19 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
         alert($.i18n('core-views/warning-col-name'));
         return;
       }
-      
+      let delay = Number.parseInt(elmts.throttleDelayInput[0].value);
+      if (Number.isNaN(delay) || delay < 200) {
+        alert($.i18n('core-views/warning-throttle-delay-input'));
+        return;
+      }
```
tfmorris, line 165: I think this needs to be reverted. See my comments on the associated issue.

main/webapp/modules/core/scripts/views/data-table/menu-edit-column.js
```diff
@@ -158,15 +158,19 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
       Refine.postCoreProcess(
         "add-column-by-fetching-urls", 
         {
           baseColumnName: column.name, 
           urlExpression: previewWidget.getExpression(true), 
           newColumnName: columnName, 
           columnInsertIndex: columnIndex + 1,
-          delay: elmts.throttleDelayInput[0].value,
+          delay: delay,
```
tfmorris, line 173: Also to be reverted.

main/webapp/modules/core/scripts/views/data-table/add-column-by-fetching-urls-dialog.html
```diff
@@ -9,7 +9,7 @@
           <td><input bind="columnNameInput" size="40" /></td>
 
           <td width="1%" style="white-space: pre;" bind="or_views_throttle"></td>
-          <td><input bind="throttleDelayInput" size="10" value="5000" /> <span bind="or_views_milli"></span></td>
+          <td><input bind="throttleDelayInput" size="10" value="500" /> <span bind="or_views_milli"></span></td>
```
tfmorris, line 12: I like the new default. That's always my starting point.

tfmorris (commented):
As I mentioned in my comment on #5173, I don't think the 200ms minimum applies to the delay, but instead to the retry interval.

main/webapp/modules/core/langs/translation-en.json
tfmorris: This was addressed in a follow-up PR (#5197) after this was merged.

Issue #4819, opened 2022-04-29
Selection of Wikibase instance directly in the schema editor


wetneb:
Our Wikibase schema editor currently looks like this:
![image](https://user-images.githubusercontent.com/309908/165939943-90415650-4561-471c-8b87-1ac4e1dc0544.png)

This has a few problems:
* the currently selected Wikibase instance is not very well identifiable, because it is just mentioned in passing in a big blob of text
* the big blob of text may be read the first time, but then it is really just taking up space on the screen
* the way to change the Wikibase instance is not obvious

### Proposed solution

@lozanaross proposed the following design (from [this series of wireframes developed for the Wikimedia Commons integration project](https://upload.wikimedia.org/wikipedia/commons/2/2e/Structured_Data_on_Commons_file_upload_in_OpenRefine_-_wireframes_draft_v2.pdf)):
![image](https://user-images.githubusercontent.com/309908/165941106-f637252b-ae7a-419f-9dbb-b994f002063d.png)

(disregard the schema template part: that is for a separate issue).
Note also the change of layout for the buttons on the right-hand side, which also makes sense to me.

I am not sure if the explanations removed in the process should be displayed elsewhere or if we can remove them entirely.

### Alternatives considered

None

### Additional context

Design proposed for the Wikimedia Commons integration project.

lozanaross:
Slight update here: https://xd.adobe.com/view/d8b086f2-199a-4fcf-a1e8-8d939842b790-8de3/
![Image](https://user-images.githubusercontent.com/25489750/166142767-02fe9f03-a5f4-4c1b-bbc6-4e49a53da5e6.png)

@wetneb @trnstlntk Should I make a separate version just for the Wikibase extension in general, because this is specific to SDC extension, but it can be cleaned up more just for the WB extension as well.

wetneb:
Yes I would hope we can have one version that can be integrated in the Wikibase extension and which can be used for Commons as well as any other Wikibase instance.

lozanaross:
> Yes I would hope we can have one version that can be integrated in the Wikibase extension and which can be used for Commons as well as any other Wikibase instance.

@wetneb Thanks. Given your answer to https://github.com/OpenRefine/CommonsExtension/issues/24 I assume there is actually no need to simplify the design for Wikibase extension then, because it will also include a template dropdown? And the schema template option will also be there?

[...]

Issue #1713, opened 2018-08-19
Add "no value" and "unknown value" as possible values for a Wikidata statement


SannitaSSJ:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
It would be great if there could be a possibility to put "no value" or "unknown value" as a value for statements in a Wikidata scheme.

**Describe the solution you'd like**
When editing the Wikidata scheme for exporting values, it would be nice to have an additional command to add a "no value" or "unknown value" option, much like the Wikidata one (see image below).

![immagine](https://user-images.githubusercontent.com/31443821/44308657-12d51880-a3ba-11e8-8ce7-3730676e9548.png)

**Describe alternatives you've considered**
none

**Additional context**
none

wetneb:
Indeed! Adding this to the todo list.

https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing/FAQ#Can_I_add_unknown_value_/_no_value_claims?

SannitaSSJ:
Fantastic, thanks for updating the FAQs. :)

wetneb:
Olivier from packed.be would also be interested in this.

lozanaross:
@wetneb  just checking in to see if there's been any development on this or if there are any plans to implement this soon. It would actually be really valuable for custom Wikibase integrations, because apparently unknown values are not supported by wikidata integrator either, so not many options to work with that without resorting to manual editing.

wetneb:
We can try to attract a bit more attention to this issue, since it is not that hard.

sjathin:
I would like to work on this. As I am new to OpenRefine and Wikidata. I went through the following tutorial to get an idea - https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing/Tutorials/Basic_editing 

I followed the above tutorial, during the edit wikibase scheme, I added an item, and a statement(see the following image) should "no value" or "unknown value" be present here?

![Screenshot 2021-03-01 at 15 22 57](https://user-images.githubusercontent.com/29340368/109510623-87689200-7aa2-11eb-97e0-93d5b15daab0.png)

And also it would be helpful if I can get to know how to test this? Or if there is any documentation available that can help me to test this.

wetneb:
Yes this is the area where we would add support for this. Here are a few ideas.

[...]

PR #5191, opened 2022-08-17
Initial support for schema templates.

### LINKED ISSUE #5043: Ability to save "schema templates": schemas where some values are left empty

wetneb:
Closes #5043.

This adds the ability to start a schema from a previously saved schema template.
A schema template is a schema with some missing values.
It is also possible to pre-define some templates in the Wikibase manifest, so that those are added to OpenRefine when the manifest is first loaded.

I think it would be useful to add an interface to list existing schema templates, which would enable:
* deleting schema templates
* exporting a schema template to JSON (useful to add it to the manifest afterwards)

### DIFF
--- a/extensions/wikidata/module/MOD-INF/controller.js
+++ b/extensions/wikidata/module/MOD-INF/controller.js
@@ -66,6 +66,7 @@ function init() {
         "scripts/wikibase-manifest-schema-v1.js",
         "scripts/wikibase-manifest-schema-v2.js",
         "scripts/wikibase-manager.js",
+        "scripts/template-manager.js",
         "scripts/menu-bar-extension.js",
         "scripts/warnings-renderer.js",
         "scripts/lang-suggest.js",
@@ -79,6 +80,7 @@ function init() {
         "scripts/dialogs/import-schema-dialog.js",
         "scripts/dialogs/wikibase-dialog.js",
         "scripts/dialogs/statement-configuration-dialog.js",
+        "scripts/dialogs/save-new-template-dialog.js",
         "scripts/jquery.uls.data.js",
       ]);
 
--- a/extensions/wikidata/module/langs/translation-en.json
+++ b/extensions/wikidata/module/langs/translation-en.json
@@ -310,5 +310,6 @@
     "wikibase-save-new-template/existing-template-label": "Or select an existing template to override",
     "wikibase-save-new-template/select-template": "Select template...",
     "wikibase-save-new-template/save": "Save",
+    "wikibase-save-new-template/empty-name": "Please provide a name for the schema template.",
     "wikibase-schema/template-load-erases-schema": "This will discard the current schema, which has unsaved changes."
 }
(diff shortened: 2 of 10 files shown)


trnstlntk:
I have just tested this, and I indeed am able to create and save new schema templates and then switch between them.
I can save them with or without prefilled values. I think that's good. See the difference between information/artwork and the Art photo example in the short screen recording I made:

https://user-images.githubusercontent.com/9881645/185386296-0e24659b-9951-45f7-9643-be706531eebb.mov

I noticed that I can't save schemas (not schema templates, but schemas) with empty values, is that indeed the behavior we want to maintain? I have no opinion (or memory of any decision) either way.

Also, if I recall correctly, @lozanaross suggested some changes in placement of certain buttons; that may be something to look at in a separate issue I can imagine.

wetneb:
> I noticed that I can't save schemas (not schema templates, but schemas) with empty values, is that indeed the behavior we want to maintain? I have no opinion (or memory of any decision) either way.

Yes, a schema is meant to be something that is able to translate rows to edits: if there is a missing value in a schema, then we cannot use it to generate edits. This is consistent with the previous behaviour (not being able to use a schema if it is incomplete).

Of course this is different from the "novalue" / "somevalue" support (#1713), where the user would explicitly specify that they want to use those "special values" (rather than the schema being incomplete.

> Also, if I recall correctly, @lozanaross suggested some changes in placement of certain buttons; that may be something to look at in a separate issue I can imagine.

Yes, I am waiting for those issues and I think @lozanaross is aware.

lozanaross:
What issues are missing? I thought I made everything that was agreed to be feasible within this round of funding as issues a while back in July / August.

wetneb:
Correct, there is https://github.com/OpenRefine/OpenRefine/issues/4819 already actually!

PR #5199, opened 2022-08-19
Re-enabled tests that create project from URL(s)

### LINKED ISSUE #4015: Test for creating project from web URLs are not working anymore
Those two tests doesn't work anymore and have been disabled to keep the test suite stable

- Test the create project from Web URL based on CSV
- Test the create project from Multiple Web URLs based on CSV

it looks likes it's from the latest version of Cypress 7.5

elebitzero:
Fixes #4015

- Re-enabled following tests:
  - Test the create project from Web URL based on CSV
  - Test the create project from Multiple Web URLs based on CSV

This PR fixes Cypress tests which were broken due to Cypress having trouble following iframes. (Ref: https://github.com/cypress-io/cypress/issues?q=is%3Aissue+is%3Aopen+iframe+in%3Atitle ) 

Create project was making ajax calls using the outdated technique of calling `form.submit()` to a hidden iframe (#create-project-iframe).  The rest of the code uses `$.post()` or `$.ajax()` to make async backend API calls.

Cypress does not work well with projects using iframes.  When I debugged the above tests to see why they were failing, Cypress was hanging at the point of `form.submit()` to the hidden iframe.

To fix this, I changed the code in default-importing-controller/controller.js to use the standard `$.ajax()` call instead of `form.submit()`. After this, the Cypress tests worked again.


wetneb (approved):
Ouch, submitting a form in a hidden iframe? That is definitely something I am happy to see the back of. I assume they had a good reason to introduce that back then, probably because file upload did not play well with XHR? And that we can now get rid of this hack because the problem has been fixed in all non-ancient browsers?

elebitzero:
I'm cleaning this up, I'll commit and push the changes later today.

PR #5201, opened 2022-08-20
Fix file permissions in DMG

### LINKED ISSUE #5160: OpenRefine 3.6 will not launch on Mac when installed for all the users of the machine (root:wheel)
In chatting with the devs on Gitter, it seems like this needs to be baked into the build process:

sudo chmod +r /Applications/OpenRefine.app/Contents/Resources/webapp/WEB-INF/*

otherwise, if the application is owned by "root:wheel" -- which is how Munki would deploy it for those orgs that use that -- the application will not launch.

wetneb:
Closes #5160.

- adapt packaging process to fix file permissions

### DIFF
--- a/packaging/pom.xml
+++ b/packaging/pom.xml
@@ -303,6 +303,12 @@
                     <arg line="-r"/>
                     <arg line="&quot;${mac.build.directory}/OpenRefine.app/Contents/PlugIns/JRE/Contents/Home/jre/legal&quot;"/>
                 </exec>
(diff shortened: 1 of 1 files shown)


wetneb:
I did not test this, not having a Mac machine at hand. If someone wants to check if it indeed solves the problem, that would be very useful (@antoine2711 perhaps?). Otherwise, I would be happy to just release 3.6.1 with this commit and report back to the user and see if it works for them.

wetneb:
Let's go ahead and release this :)

PR #5231, opened 2022-09-04
Make browser list more useful

katrinleinweber:

- Hyperlink browser list to their homepages
- Add Safari as support WebKit browser
- Correct WebKit capitalisation ;-)

---

I hope it's correct to update all `*/versioned_docs/version-*/` files as well.

- [ ] Discuss, whether the list has to be duplicated in `running.md` & `installing.md` files

### DIFF
--- a/docs/docs/manual/installing.md
+++ b/docs/docs/manual/installing.md
@@ -30,6 +30,7 @@ OpenRefine works best on browsers based on Webkit, such as:
 *   [Chromium](https://ungoogled-software.github.io/)
 *   [Opera](https://www.opera.com/)
 *   [Microsoft Edge](https://www.microsoft.com/edge)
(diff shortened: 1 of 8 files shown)


wetneb (approved):
Welcome back @katrinleinweber and thanks for this!

ostephens (approved):

katrinleinweber:
Thank you, @ostephens! I guess the list has to be duplicated in `running.md` & `installing.md` files ;-D

ostephens:
@katrinleinweber the way it's currently structured I think it's safer because we may find someone reads one page but not the other? Perhaps a better alternative approach would be to have a page on browser compatibility which is linked from both locations?

katrinleinweber:
Good point! I've started https://github.com/OpenRefine/OpenRefine/issues/5234 as a follow-up to this.

PR #5235, opened 2022-09-06
Improve documentation about absence of support for editing sitelinks

wetneb:
In Wikibase schemata we do not let people edit sitelinks yet, this fixes the docs about that.


trnstlntk (approved):
Great 👍

PR #5249, opened 2022-09-12
Added labels and cleaned up "Add column from reconciled values..." dialog

elebitzero:
While adding labels to the "Add column from reconciled values..." dialog, I found there were two copies of the HTML, and that one is a leftover from a previous refactoring.  

- Added labels.
- Removed unused HTML template.
- Renamed the related files to more appropriate names.
- PR which introduced the duplicate HTML files:
https://github.com/OpenRefine/OpenRefine/pull/1210


wetneb (approved):
Many thanks and sorry about this duplicate I introduced years ago!

PR #5253, opened 2022-09-14
Adjusted the colors in range and timerange facet to have more contrast

### LINKED ISSUE #3319: adjust colors of filtered numeric facet
I taught a workshop on OpenRefine today, and some learners could not see the dark-blue on light-blue histograms when filtering data in a numeric facet.

### Proposed solution
Add a color choosing menu that adjusts the "filtered histogram color" (and perhaps the initial histogram color).

### Alternatives considered
I don't know of another way in OpenRefine to access a histogram of filtered data, or to compare the distribution 

### Additional context
The color contrast was not strong enough for them:

image

elebitzero:
Fixes #3319

Adjusted the numeric and timeline facet histogram bin colors to improve the contrast so that it is 3 to 1 between the three colors: background (white), bin color 1, and bin color 2.

Previously, timeline facet was using a separate color palette. This change makes numeric and timeline use the same colors. See the pictures below to see the difference.

Before and After:
image

Contrast:
image

### DIFF
--- a/main/webapp/modules/core/scripts/facets/range-facet.js
+++ b/main/webapp/modules/core/scripts/facets/range-facet.js
@@ -172,7 +172,7 @@ class RangeFacet extends Facet {
     this._elmts.removeButton.on('click',function() { self._remove(); });
     this._elmts.minimizeButton.on('click',function() { self._minimize(); });
 
-    this._histogram = new HistogramWidget(this._elmts.histogramDiv, { binColors: [ "#bbccff", "#88aaee" ] });
+    this._histogram = new HistogramWidget(this._elmts.histogramDiv, { binColors: [ "#668CFF", "#174092" ] });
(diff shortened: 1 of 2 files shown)


wetneb (commented):
I like it! Since this touches visual design I would leave this open a bit longer to let more people chime in.

tfmorris:
LGTM 

Do the "change" and "reset" labels now meet the accessibility guidelines? My [comment](https://github.com/OpenRefine/OpenRefine/issues/3319#issuecomment-722704249) from a couple of years ago indicated that they didn't at the time.

wetneb (approved):
Good to merge - the "change" and "reset" buttons are generic parts of the facet component, so we can keep the scope of this PR to the histograms.
We plan to do an accessibility review of the tool to work on those topics more.

PR #5263, opened 2022-09-16
Make respondException print the exception stacktrace

Abbe98:
When errors happen in the API it can be rather opaque to the end user as the error is only printed in the resulting JSON response. This change ensures that the stacktrace is also natively printed.


wetneb (approved):
Fine with me!

PR #5264, opened 2022-09-16
drop unused imports from Wikibase extension

Abbe98:
Mainly to remove noise in IDE problem windows.


Abbe98:
It seems like CI is never reaching the tests.

wetneb:
You probably just need to run the linter with `mvn formatter:format`

wetneb (approved):

PR #3624, opened 2021-02-12
(I#3430) Tests for the Wikidata extension

fgiroud:
First PR of a serie, in order to cover the wikidata extension with Cypress tests
issue #3430

Issue #3430, opened 2020-12-16
Add UX tests about the Wikidata extension


fgiroud:
This is a placeholder ticket so far.
Community feedback is welcome to understand if, what and how to test this extension properly.

![Screen Shot 2020-12-16 at 14 46 39](https://user-images.githubusercontent.com/6267288/102356498-92387e00-3fad-11eb-96f9-d872e7bf4271.png)

### Proposed solution

To be determined

fgiroud:
@riteshsoni123 that's a tough one, in case you're interested you can dig in

riteshsoni123:
Yes surely I would like to give it a try , but a little bit explanation about this issue would be helpful.

PR #5273, opened 2022-09-21
Wikidata extension tests

elebitzero:
Partial fix for #3430
This PR is adapted from @fgiroud PR #3624

- Added basic tests for the Wikibase extension that are possible without credentials:
  - Check extension is loaded
  - Check login dialog
  - Select Wikibase instance
  - Add/Delete Wikibase manifests by URL and JSON


wetneb (approved):
Thank you so much!

PR #5286, opened 2022-09-25
Fix reconciliation dialog overflow issue.

### LINKED ISSUE #5285: Dysfunctional reconciliation dialog window in OpenRefine 3.7 snapshots (since around Sep 20, 2022)
I tried reconciling against Wikidata in a large-ish dataset using our 3.7 snapshot release, and the reconciliation dialog appears wonky at the bottom. The `Start reconciling` and `Cancel` buttons are dysfunctional (nothing happens when clicking them).

image

### To Reproduce

The bad things happened with this dataset: [homosaurus-v3-jsonld.openrefine.tar.gz](https://github.com/OpenRefine/OpenRefine/files/9639142/homosaurus-v3-jsonld.openrefine.tar.gz)
I tried reconciling the `http://www.w3.org/2004/02/skos/core#prefLabel` column
In Version 3.7-SNAPSHOT [549f83b] and the September 21 snapshot release.

I loaded this same dataset in OpenRefine 3.6.1 on Wikimedia PAWS, where the window does behave as expected, so it seems like the issue was recently introduced.

### Versions
 - Operating System: MacOS 12.6
 - Browser Version: Chrome Version 105.0.5195.125 (Official Build) (arm64)
 - JRE or JDK Version: n/a?
 - OpenRefine: Version 3.7-SNAPSHOT [549f83b] and the September 21 snapshot release

wetneb:
We plan to redesign this dialog completely - there are a lot of issues with it. This is just a fairly minimal fix to avoid its contents overflowing and hiding the "Reconcile" and "Cancel" buttons.

Closes #5285.

- make sure the lower part of the form can expand to the full width of the dialog by uncoupling it with the separation between the two upper parts
- remove the (fairly arbitrary) `height: 500px` set on the dialog so that it can expand further down to fit its contents

New version:

image

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
+++ b/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
@@ -15,24 +15,18 @@
       <td width="50%"><div class="detail-container" bind="detailContainer"></div></td>
     </tr>
     <tr>
-      <td><input type="radio" name="type-choice" id="againstType" value="">
+      <td colspan="2"><input type="radio" name="type-choice" id="againstType" value="">
         <label for="againstType"><span bind="or_proc_againstType"></span></label> <input size="20" bind="typeInput" /></td>
-      <td>
-      </td>
     </tr>
     <tr>
-      <td><input type="radio" name="type-choice" id="noType" value="-">
+      <td colspan="2"><input type="radio" name="type-choice" id="noType" value="-">
         <label for="noType"><span bind="or_proc_noType"></span></label></td>
-      <td>
-      </td>
     </tr>
     <tr>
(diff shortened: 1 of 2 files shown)


trnstlntk:
Thank you :D

elebitzero (approved):
@wetneb

Looks good, except without `width: 500px`, when the reconcile dialog first opens the main panel is empty until you select a recon service, and it looks odd with the service dropdown menu extended outside the dialog, as seen below. I would change it to `min-width: 500px`.

(Only need to set min-width on recon-dialog-service-panel-container, not necessary for recon-dialog-service)

I noted on the issue that this is not a regression, but has been existing bug. If we will have a 3.6.2 patch, this is a good candidate.

image

wetneb:
Thanks a lot, I should have caught that…

PR #5288, opened 2022-09-25
Wikibase selection in schema editor

### LINKED ISSUE #4819: Selection of Wikibase instance directly in the schema editor

wetneb:
Fixes #4819

- add a drop-down menu to select the Wikibase directly in the schema
- simplify the explanation text at the top of the schema
- limit the width of the schema to keep the buttons on the right-hand side of the statements close to the body of the statements
- change the styling of the "add entity" buttons to be actual buttons instead of links


wetneb:
Oh that's exciting, I need to update the Cypress tests of the Wikibase extension! How cool :)

PR #5303, opened 2022-09-30
Add dialog to manage schema templates.

### LINKED ISSUE #5280: Add ability to rename, delete and export schema templates
We recently introduced the notion of "schema template", which helps the user start from a skeleton of Wikibase schema instead of creating one from scratch.
Users can currently create their own schema templates, but they are not able to:
* rename them
* delete them
* export them to JSON

### Proposed solution
We should have an additional dialog which lists the templates stored for the current Wikibase instance, letting the user perform the actions above.

### Alternatives considered
It was also suggested by @lozanaross to store those schema templates in a directory of the file system, and let the user perform those operations using their file manager. While this is attractive (no need to re-implement our own file manager), templates are currently stored in the preference store (so in `workspace.json`), so it would be involved to change that.

### Additional context
Part of the Wikimedia Commons integration project.

wetneb:
Closes #5280.

To validate the schema templates that are imported by the user, I considered using a JSON schema (like we do for Wikibase manifests). I tried generating the JSON schema automatically from the Java classes using [jsonschema-generator](https://github.com/victools/jsonschema-generator) but it does not quite work out of the box and has no guarantees to be completely faithful to Jackson deserialization. So I went for a simpler approach of introducing a new command which parses a schema template in the backend, and reports the status to the frontend.

The UI is not beautiful and style changes would be welcome. I did not add support for exporting multiple templates at a time because it would require bundling up in some archive, and that is not something I would want to do generate in the frontend directly.


elebitzero (changes requested):
Some things I found:

 ### Save new schema template
- Save new... could just be Save... since you can save new or save existing.  (Also update header to Save schema template)

- Manage schema templates should have ellipsis ...
- I would pull Manage schema templates out of the Save dialog.

- Padding needs to be adjusted.

###  Manage schema templates

- Change header from 'Schema templates' to 'Manage Wikibase schema templates'

- A better presentation would be to have the actions (Rename | Export | Delete) to the right of each template name.
- The list of templates needs to handle overflow (scrollbar)

- If I delete a template, then close the manage schema dialog, the save new dialog still has the template in the 'Select template...' dropdown. (This is another reason to move Manage schema templates out of here.)

extensions/wikidata/module/styles/dialogs/schema-template-dialog.less
```diff
@@ -0,0 +1,15 @@
+.schema-template-list {
+     list-style: none;
```
elebitzero, line 2: For .schema-template-list, add a `max-height:450px` and `overflow-y:auto` in case list gets too long.

extensions/wikidata/module/scripts/dialogs/schema-template-dialog.js
```diff
@@ -0,0 +1,91 @@
...
+};
+
+SchemaTemplateDialog.prototype.deleteTemplate = function (wikibaseName, templateName) {
+  WikibaseTemplateManager.deleteTemplate(wikibaseName, templateName);
+  WikibaseTemplateManager.saveTemplates();
+  this.populateDialog();
+};
+
+SchemaTemplateDialog.prototype.renameTemplate = function (wikibaseName, templateName) {
+  let newName = prompt($.i18n('wikibase-schema-template/enter-new-template-name'), templateName);
```
elebitzero, line 85: Need to check if `newName` is valid. It's easy to hit escape or click away which closes the browser prompt, and then `newName` is null.

wetneb:
Thanks a lot for the thorough review!

> 'Save new...' could just be 'Save...' since you can save new or save existing. (Also update header to 'Save schema template')

That would also work, but I think @lozanaross had some precise thoughts about the naming of this button to distinguish it from the "Save schema" button.
In my opinion, whether "new" is there or not, this might still be a bit unclear for users (even if the button is next to the selection of templates). I think I would need something like "Save current schema as template" to emphasize the relationship to the contents of the schema editor below, but that is quite verbose…

> I would pull 'Manage schema templates' out of the 'Save' dialog.

I agree it feels weird to have to go through the Save dialog for that. Would you add it as a button in the schema editor (next to the button that currently says "Save new…")? Or in the extension menu perhaps?

lozanaross:
Thank you both for all the comments (note -  still have not seen what this PR does UI-wise - screenshots would be handy).

After running some more user testing sessions last week, I've noticed the following: 
- People have pretty different mental models of what schema is vs schema template. It gets hairy when we also start thinking about structured data templates vs wikitext templates...
- "Save new..." indeed does not work. I agree it should probably be just "Save..." although it's potentially a bit unclear.
- I realized that schema templates are not just templates (that is if we agree that something called template is more typically assumed to be "unfilled" - with properties, but no values), they are like what we currently call schemas, but more powerful, because they can contain a mix of filled / unfilled statements, wikitext, and more... So we lose the "usefulness" of schemas alone - i.e. why would a user want to save, import or export a schema, when they can do all that and more with a template which is the same, just more powerful. 
- The above leads me to think that: 
1. We should bring back one of my earlier designs where I called the button that now says "Save schema" something like "Save progress" or "Save schema progress" because this save can be automatic as well (like auto-save) and the use-case is more about not losing your progress while building a schema for a specific project than saving something that will be transferrable. 
2. For all other needs, users can just use schema templates, and if we call the previous button "progress", then we may even be able to scrap the term "template", why not just call them schemas.
3. Lastly, what is currently "Import / Export" schema options from the Extensions dropdown is superfluous if we introduce management dialog for schema templates (or just schemas) - I don't see the need to import or export a schema if you can do that for a template, which as stated above is more powerful. We can indeed add the Manage schema templates (or just Manage schemas) dialog as an item accessible directly from the Extensions dropdown, but I have to say users never find that menu intuitive, so it should be there and probably also still in the "Save..." dialog, too. So accessible from two points, may be not in line with other sections of OpenRefine, but would be nice for users. 

I will lay this all out in more detail in a report after all user testing is complete this week. But let's also discuss here if the above is unclear.

elebitzero (approved):

wetneb:
> Thank you both for all the comments (note - still have not seen what this PR does UI-wise - screenshots would be handy).

Here is one (but I would really like to ensure you can try out a PR on your machine, so that you can have the full user experience - let me know if you need help to set this up):
image


> 1. We should bring back one of my earlier designs where I called the button that now says "Save schema" something like "Save progress" or "Save schema progress" because this save can be automatic as well (like auto-save) and the use-case is more about not losing your progress while building a schema for a specific project than saving something that will be transferrable.

Let's leave this for a separate issue since this can be changed independently from this PR.

> 2. For all other needs, users can just use schema templates, and if we call the previous button "progress", then we may even be able to scrap the term "template", why not just call them schemas.

I think calling those schemas would work. We can then distinguish from complete and incomplete schemas when needed (which is already something we do since #5131).

> 
> 3. Lastly, what is currently "Import / Export" schema options from the Extensions dropdown is superfluous if we introduce management dialog for schema templates (or just schemas) - I don't see the need to import or export a schema if you can do that for a template, which as stated above is more powerful.

Yes I think it makes sense. Some users might find it a bit annoying (because applying a schema to a project will require extra steps, and same for the export), but it is probably worth it for the sake of simplifying the UI.

That does require some changes to this PR because at the moment importing a schema exported from a previous version of OpenRefine as a schema template in 3.7 will not work (because of the missing name). But I can introduce a workaround for that.

> We can indeed add the Manage schema templates (or just Manage schemas) dialog as an item accessible directly from the Extensions dropdown, but I have to say users never find that menu intuitive, so it should be there and probably also still in the "Save..." dialog, too. So accessible from two points, may be not in line with other sections of OpenRefine, but would be nice for users.

I think it is fine to have a dialog accessible from multiple places. We already have that for the Wikibase upload dialog, accessible both from the global Export menu and the Wikibase extension dropdown, for instance.

wetneb:
So I have renamed schema templates to just schemas and simplified the extension menu (removed import/export and replaced by the schema templates):
image

I feel like it's a nice usability gain, thanks @lozanaross!

lozanaross:
> So I have renamed schema templates to just schemas and simplified the extension menu (removed import/export and replaced by the schema templates): image
> 
> I feel like it's a nice usability gain, thanks @lozanaross!

Very nice! I would say maybe final small change - move "Manage WB account..." below the line & above "Upload..." because Manage schemas should be closer to Edit schema & the account login is anyway more handy for upload I think...

lozanaross:
> Here is one (but I would really like to ensure you can try out a PR on your machine, so that you can have the full user experience - let me know if you need help to set this up): image

OK, agreed, I'll try it w/ GH desktop app and see if I manage. Shall I then make a separate issue if I have small tweaks for this particular dialog window?

> Let's leave this for a separate issue since this can be changed independently from this PR.

OK, I will make the issue (maybe after the user testing report is complete).

> I think calling those schemas would work. We can then distinguish from complete and incomplete schemas when needed (which is already something we do since #5131).

Sounds good. 

> Yes I think it makes sense. Some users might find it a bit annoying (because applying a schema to a project will require extra steps, and same for the export), but it is probably worth it for the sake of simplifying the UI.

To be honest all Wikimedia users I've spoken with didn't really understand the schema concept very well (including myself), so I doubt there will be many users who used this functionality extensively to be annoyed once we change it.

> That does require some changes to this PR because at the moment importing a schema exported from a previous version of OpenRefine as a schema template in 3.7 will not work (because of the missing name). But I can introduce a workaround for that.

OK, great.

> I think it is fine to have a dialog accessible from multiple places. We already have that for the Wikibase upload dialog, accessible both from the global Export menu and the Wikibase extension dropdown, for instance.

Great, good to know. I may have accidentally navigated users away from Export, because I thought you can only upload via the extension dropdown 🫣

wetneb:
This is getting stale, so merging it. Further tweaks can be suggested in follow-up issues.

thadguidry:
@wetneb @lozanaross @trnstlntk Curious, do you know where folks are saving schema templates for collaboration? Is it under individual WikiProjects?

PR #5307, opened 2022-10-01
Preserve aspect ratio of Wikibase logos

### LINKED ISSUE #5306: Proportions of Wikibase logos are not respected in the Wikibase management dialog
In this dialog, the logos that are not natively square are stretched, not preserving their aspect ratio:
image

Not providing more details because I am intending to tackle this issue directly.

wetneb:
Fixes #5306.
# Before
image
# After
image

### DIFF
--- a/extensions/wikidata/module/styles/dialogs/wikibase-dialog.less
+++ b/extensions/wikidata/module/styles/dialogs/wikibase-dialog.less
@@ -39,9 +39,7 @@
 
 .wikibase-list li img {
   flex: 0 0;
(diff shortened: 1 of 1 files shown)


wetneb:
I guess the intention with the original CSS rules is that the image is as big as possible. If a Wikibase logo has natively both dimensions smaller than 60px, then it will probably not expand to its maximum possible size. I wonder if there is a standard solution around for this?

elebitzero (approved):

PR #5317, opened 2022-10-06
Introduce expensive Wikibase QA checks

### LINKED ISSUE #5037: Implement expensive Wikibase QA checks only triggered before upload
The Wikibase extension comes with some quality assurance checks, which validate the candidate edits against some constraints, to warn the user of potential issues in the edits.

The checks implemented so far are designed to run quickly, so that we can update the issues tab in real time, as the user transforms their dataset and filters it to display a specific subset of rows.

Sometimes it would be useful to implement more expensive checks, which we would not want to run every time a change in project data happens, because it would delay the delivery of issues to the user and waste resources. For instance, I am thinking about the following checks:
* when uploading new items with external unique identifiers, check if there are any existing items with those identifier values. For now we only report an issue if we are creating two different items with the same identifier value
* when uploading new media files, check that their filenames do not exist on the MediaWiki instance yet
* checking for the presence of suitable inverse statements (for now we only check for them within the edits to be made)
* and probably more constraint checks that we cannot implement fully without fetching data from the Wikibase instance

### Proposed solution

Constraint checks could be run into two modes:
* the fast mode (the current one), in which the check must be performed quickly (without fetching external information except for small metadata like constraint definitions)
* the slow mode (the new one), in which the checks are allowed to take longer, for instance by making some network operations

The slow mode would only be triggered when opening the "Upload edits to Wikibase" dialog. This dialog would show those more thoroughly computed checks (currently it already displays the existing fast checks).

### Alternatives considered

One other solution could be to change the architecture used to compute QA issues, so that each scrutinizer could return its issues asynchronously. This would mean that the issues that are the fastest to compute would be displayed first, and then the slower ones would appear. This would be very nice to have, but I think it would not get rid of the need to have two modes (slow / fast), because I would not want to run the expensive checks over and over, every time a change is made (it would generate a lot of requests to the target wikibase, a lot of which would likely be redundant).

### LINKED ISSUE #5298: Media file upload: provide clear error report when user has mistakes in local file paths
I've encountered a case where an OpenRefine user tried to upload media files to Wikimedia Commons. The upload failed because the (MacOS) file paths in the user's dataset didn't start with a `/` (very subtle!)

The user didn't get any error messages and it took a bit of searching to discover this issue. Maybe this will occur more frequently if users start from datasets with pre-entered local file paths.

### Proposed solution
Perform a general/rough check if the file path is formatted correctly and provide the user with a clear error message if there are obvious mistakes in there (e.g. missing drive letter in Windows paths, missing starting `/` for MacOS paths, wrong directions of (back)slashes, missing file extensions...?)

Related / similar: https://github.com/OpenRefine/OpenRefine/issues/5192

wetneb:
Closes #5037. Closes #5298.

This also introduces a sample expensive check, which verifies that files to be uploaded have a valid path.

image

### DIFF
--- a/extensions/wikidata/module/langs/translation-en.json
+++ b/extensions/wikidata/module/langs/translation-en.json
@@ -267,6 +267,8 @@
     "warnings-messages/new-media-without-file-name/body": "Uploading new media files requires supplying a file name to be used on the wiki. No file name was provided for {example_entity}, for instance.",
     "warnings-messages/new-media-without-wikitext/title": "New media without wikitext",
     "warnings-messages/new-media-without-wikitext/body": "Uploading new media files requires supplying accompanying wikitext. No wikitext was provided for {example_entity}, for instance.",
+    "warnings-messages/invalid-file-path/title": "Invalid file paths",
+    "warnings-messages/invalid-file-path/body": "Some paths of files to be uploaded are invalid, such as <span class=\"wb-issue-preformat\">{example_path}</span>. Make sure those are valid absolute paths on the system which runs OpenRefine.",
     "wikibase-statement-settings/dialog-header": "Statement settings",
     "wikibase-statement-settings/mode-label": "Editing mode",
     "wikibase-statement-settings/strategy-label": "Matching strategy",
--- a/extensions/wikidata/module/scripts/dialogs/perform-edits-dialog.js
+++ b/extensions/wikidata/module/scripts/dialogs/perform-edits-dialog.js
@@ -126,7 +126,11 @@ PerformEditsDialog.checkAndLaunch = function () {
       var discardWaiter = DialogSystem.showBusy($.i18n('perform-wikibase-edits/analyzing-edits'));
       Refine.postCSRF(
           "command/wikidata/preview-wikibase-schema?" + $.param({project: theProject.id}),
-          {manifest: JSON.stringify(WikibaseManager.getSelectedWikibase()), engine: JSON.stringify(ui.browsingEngine.getJSON())},
+          {
+             manifest: JSON.stringify(WikibaseManager.getSelectedWikibase()),
+             engine: JSON.stringify(ui.browsingEngine.getJSON()),
+             slow_mode: true
+          },
           function (data) {
             discardWaiter();
             if (data['code'] !== 'error') {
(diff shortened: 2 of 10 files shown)


trnstlntk:
Just doublechecking: what does 'valid path' mean?
* Is it just a check whether the formatting of the path is correct according to some basic rules (e.g. on my machine: `/Users/sandra/Desktop/Open_Refine_is_Magic.png` is technically 'valid' in terms of MacOS file path syntax, but ehh, I don't have a user directory called `sandra` and that file is really in a subdirectory and not in the given place)
* Or is it a complete check whether the path is indeed correct, all directories are there, and if the file is indeed in that place?

From previous discussions I have the impression it's the latter, right? That'd be super nice 🤩

That said, I find the term 'valid path' a bit vague. Can we be more specific and use slightly different language? For instance along the lines of
* "Your directory path(s) seem(s) incorrect; OpenRefine could not find a directory `/Users/sandra/Desktop/` _or_ could not find this/these directory/ies on your machine. Please double-check and correct the path(s)."
* "OpenRefine can't find the file(s) at the locations you indicated. Please double-check the location of the file(s) and correct the file path(s) accordingly."

I think I'd personally understand such messages just a bit better. Now I want to learn more about good interface copywriting, haha 😅

wetneb:
> Or is it a complete check whether the path is indeed correct, all directories are there, and if the file is indeed in that place?

Yes it's this version.

> "Your directory path(s) seem(s) incorrect"

In my understanding  "directory path" refers to a path to a directory, and here we want a path to a file, so I think this version could mislead the users into thinking they need to provide paths to directories in their cells.

I personally think it is useful to give an example path to the user. Shortening it to the first directory that could not be found could make sense, but I will probably not have time to do that (this is a PR which does not fit in any work packages but feels important to me).

wetneb:

trnstlntk (approved):

lozanaross:
Small (late) suggested edit to the copy, although I see this is already merged:
"Incorrect file paths
Some paths of the files to be uploaded appear incorrect, such as /this/does/not/exist. Make sure those are correct absolute paths on the system which runs OpenRefine. "

I agree with Sandra that invalid can be vague, but want to keep "file path" vs "directory path".

PR #5322, opened 2022-10-07
Add dialog to let user pick recon service when marking cells as new

### LINKED ISSUE #4985: Feature request: "shortcut" menu item for new item creation against a chosen reconciliation service
Idea from @lozanaross, endorsed by me, which we discussed in the context of Wikimedia Commons user testing sessions.

Creating new items via reconciliation services is a confusing process, as we regularly see in demos or user tests. People first have to reconcile non-existent entities (eg not yet existing Wikidata items or Wikimedia Commons filenames) against a reconciliation service, and then mark them as new. We propose a shortcut to make this process easier and probably more logical.

### Proposed solution
Add a menu item which may be something like `Create new items via reconciliation service...`
When selecting this menu item, the user then gets a small dialog, very similar to the one produced by the `Use values as identifiers` menu command, saying: `Create a new item for each of the selected cells, against reconciliation service:` with a drop-down menu that allows the user to choose between all the reconciliation services they have installed.
image

(The idea is that prior reconciliation before this process then would not be necessary.)

### Alternatives considered
Keep the current situation with very confused users.

wetneb:
Closes #4985.

Screenshot
image

Maybe I could add a Cypress test as well.


wetneb:

trnstlntk:
Tried it, works fine!

PR #5323, opened 2022-10-07
Add Wikimedia Commons manifest

### LINKED ISSUE #5220: Ship Commons manifest with OpenRefine by default
I notice that it takes quite a bit of explanation/documentation to help people install the [Wikimedia Commons manifest](https://github.com/OpenRefine/wikibase-manifests/blob/master/wikimedia-commons-manifest.json) in OpenRefine if they want to get started editing or uploading there. Could we perhaps have this manifest pre-installed for every OpenRefine user, just like Wikidata's? Seeing that Wikimedia Commons is a Wikimedia wiki, i.e. a public one that anyone can edit (and re-use files from). 

### Alternatives considered
Also totally OK to not do this. Or, alternatively, have the Wikimedia Commons manifest shipped in the Wikimedia Commons extension (but not in OpenRefine itself).

wetneb:
Closes #5220.

This manifest is quite big because it comes with 4 predefined templates.

I realized that because we import the predefined templates when users add the Wikibase manifest, this does not work for the manifests which we ship by default, so I had to add this import when we select the Wikibase instance as well.
It's really worth thinking about a centrally hosted (on wiki) hosting solution for manifests, since this whole system of shipping templates in manifests is definitely not ideal. But that's well beyond this PR.


Abbe98:
> It's really worth thinking about a centrally hosted (on wiki) hosting solution for manifests, since this whole system of shipping templates in manifests is definitely not ideal. But that's well beyond this PR.

[Most Mediawiki instances can host JSON out of the box.](https://commons.wikimedia.org/wiki/User:Abbe98/test.json)

wetneb:
Oh nice, I had no idea!

Issue #5266, opened 2022-09-16
Move linting from CI to a Git hook


Abbe98:
Rather than to prompt the developer about linting issues post CI and then having them fix the issues we could always run the formatter prior to a commit  for a better developer experience.

This would be especially benefit for work that goes on in branches for a while before reaching master, for newcomers, and for people outside of the OpenRefine org as they do not trigger CI automatically.

As a bonus it would speed up CI.

thadguidry:
We have another issue (somewhere) that suggests to use IDE plugins with style guides which could be preferred and easier setup for devs?

wetneb:
I have come up with the following git hook (to be added as `.git/hooks/pre-commit`, with executable flag):

```
#!/bin/bash
set -eu
if git status -s -uno | grep -P "\.java$" > /dev/null; then
   echo "Linting java files, please wait..."
   mkdir -p build
   mvn formatter:format -q -Dcachedir=build/formatter_cache.properties;
   git status -uno | grep -P "\.java$" | xargs git add
fi
```

When no changes to Java files have been made, it is instant, otherwise it takes about 20 seconds (if the cache of the linter has not been previously cleared by a `mvn clean`).

wetneb:
Another possibility would be to have some automation built around GitHub (as a GitHub Action or App, for instance), to offer people to lint directly from the pull request discussion. For instance, by automatically pushing a linting commit on top of a PR. Not sure if this is really helpful because adding commits on top of a PR branch can be seen as rude (the PR author will need to pull from their branch before they make further changes, or potentially have to solve conflicts).

If we go down the Git hook route, then we could also install the hook automatically in `.git/hooks` via the `./refine` script the first time it is run (which is apparently a common practice). This avoids the need for people to type a `git config` command manually.

PhilipVigus:
I think the git hooks options is the best from a developer experience perspective, particularly if the hook can automatically be installed. From my experience, this kind of thing can and should happen with no input from the developer, which the hook approach achieves.

On a technical note, I don't think the -P grep flag is available on macOS. It complains when I run the script on my machine, and I've checked the Linux and macOS man pages. It's clearly listed on the former but there's no sign of it on the latter.

thadguidry:
The author of this https://github.com/marketplace/actions/action-lint might know better than us and wouldn't hurt to ask them?  Let's use the GitHub community of experts floating around.
[...]

PR #5326, opened 2022-10-08
Fix failing test assert when locale is set to French.

### LINKED ISSUE #5277: ToString test fails when run in a French locale
Running the server tests (with `mvn test`) with a `fr_FR.UTF-8` locale currently fails with a
```
ToStringTests.testToString:50 expected [100.0] but found [100,0]
```

This test should pass regardless of the system locale.

PhilipVigus:
Closes #5277

This changes an assert that was failing when the environment locale was set to French.

I considered forcing the locale to be en_US when tests are run but rejected this because in my opinion, tests that require a specific locale to pass are brittle and should ideally be avoided. The actual fix dynamically sets the expected value based on the locale the test is being run in, which is a more resilient solution.

While working on this issue, an inconsistency in the toString method was identified. Ideally, when the function is called on a number, it should always return a string that takes the locale into consideration. However, there is an edge case where this does not happen. The following assert is included in the `testToString` test and succeeds with a French locale when it should fail:
```
assertEquals(invoke("toString", 100.0), "100.0");
```
Fixing this would require changes to the `toString()` method itself. I spent some time looking into this but any changes I made broke the function in other ways. Given that this is a GREL function that users have been using it for some time, I chose not to dig into this further, as they have not reported this as an issue.

### DIFF
--- a/main/tests/server/src/com/google/refine/expr/functions/ToStringTests.java
+++ b/main/tests/server/src/com/google/refine/expr/functions/ToStringTests.java
@@ -75,5 +75,5 @@ public class ToStringTests extends RefineTest {
         assertEquals(invoke("toString", CalendarParser.parseAsOffsetDateTime(inputDateTime), "yyyy-MM-dd hh:mm:ss"), "2013-06-01 01:12:11");
         assertEquals(invoke("toString", CalendarParser.parseAsOffsetDateTime(inputDateTime), "yyyy-MM-dd HH:mm:ss"), "2013-06-01 13:12:11");
     }
-    
+
 }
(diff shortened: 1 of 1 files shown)


wetneb:
Thanks a lot! If linting is still a problem, you can run `mvn formatter:format` which will fix it.

wetneb:
It looks like it is (also, if you have thoughts about how we should handle this linting, feel free to chime in in https://github.com/OpenRefine/OpenRefine/issues/5266)

wetneb (approved):
Thanks a lot!

PR #5327, opened 2022-10-08
Add QA check for identical label and description in new Wikibase items.

### LINKED ISSUE #4980: Check that created Wikibase items have distinct (name,description) pairs
Hi all,

I am creating new Wikidata entries for a museum via OR (version 3.5) using the Wikidata extension. However, I have had several instances where the edits do not perform from a certain point onwards. Strange but perhaps important detail: OR shows the edits as 78% complete, while in Wikidata I can see in my contributions that only about 30 of the 293 items of _Part 4 of the public domain upload for the DIVA museum_ have been uploaded. Many titles are the same (Lepel, Vork, Modelprent met verschillende voorwerpen, ...), but it seems to me that this cannot be the problem, right? They are in fact different objects, as you can see in the object number column, and thus I created new items for each cell.

### To Reproduce
1. Reconciliation of column Title (= item) 
2. If the item doesn't exist > Actions > Create new item for each cell
3. Edit Wikibase schema
4. Upload edits to Wikibase

### Screenshots

Screenshot 2022-06-22 at 12-13-21 78% complete - Publiek domein upload DIVA - OpenRefine
afbeelding

### Versions
 - Operating System: Windows 10,
 - Browser Version: Chrome 19, Firefox 61
 - OpenRefine: OpenRefine 3.5

wetneb:
Closes #4980.

Screenshot
image


wetneb:

PR #5328, opened 2022-10-10
Fix caching of Wikibase suggest results

### LINKED ISSUE #5190: Suggestion of properties and items in the Wikibase schema builder sometimes totally refuses to load any suggestions (or only wrong ones)
Sometimes, item and property suggestion in the Wikibase extension's schema builder simply refuses to work (refusing to provide any suggestions at all, or only wrong suggestions), making schema building quite painful.

This issue is difficult to reproduce - I have seen the behavior quite often but haven't discovered a pattern in it. I was recently creating a screen recording while it happened badly and [you can see it happening in this video](https://drive.google.com/file/d/1g0g2SsbVKAMJk70vawf-swkkBtFLiz2N/view?usp=sharing). 

Reproducing my process (but I have the idea it doesn't only happen under these circumstances):
* I added what I thought was a wrong statement in the schema builder, deleted the statement, and then tried to recreate it. I wanted the same property but another value. Property (Source of file) simply refuses to load.
* Then you see me fiddling a bit, working on other statements, trying if moving out of the schema builder and back fixes it (it does not)
* Finally I decide to exit my project and go back to OpenRefine's start screen and re-open the project. Then I can add the statement just fine and the suggester does its job.

Unfortunately I didn't have the Terminal running, so I don't have a log of what happened.

This issue has been reported on Wikimedia's OpenRefine Telegram group as well.

### Expected Behavior
Properties and items display correctly and promptly.

### Versions
 - Operating System: MacOS 12.5
 - Browser Version: Something probably right before Chrome 104.0.5112.79 (Official Build) (x86_64)
 - JRE or JDK Version: n/a?
 - OpenRefine: 3.7 snapshot release of 13 August 2022

wetneb:
Closes #5190.

This bug was happening because we were storing the results of Wikibase requests incorrectly in the cache. They needed to be translated to a common format supported by the library.

### DIFF
--- a/extensions/wikidata/module/scripts/wikibase-suggest.js
+++ b/extensions/wikidata/module/scripts/wikibase-suggest.js
@@ -91,7 +91,6 @@
               self.input.data("request.count.suggest", calls);
             },
             success: function(data) {
-              $.suggest.cache[url] = data;
               // translate the results of the MediaWiki API to that of the reconciliation API
               var translated = {
                   prefix: val, // keep track of prefix to match up response with input value
@@ -100,6 +99,7 @@
                       name: result.label,
                       description: result.description};})
               };
+              $.suggest.cache[url] = translated;
               self.response(translated, cursor ? cursor : -1);
             },
             error: function(xhr) {


Imug3tsu:
Hi, i don't know how start one conversation in github, can you help me?

PR #5330, opened 2022-10-11
Allowed the cell edit popup and dialogs with textbox inputs to be resizable

elebitzero:
- Allowed the cell edit popup and the templating, export and apply history dialogs to be resizable when changing the size of the textbox inputs.

Edit cell before:
Edit cell resizable before

Edit cell after:
Edit cell resizable


wetneb (approved):
oh yeah! Thanks for the gifs!

PR #5332, opened 2022-10-11
Replace use of 'which' with 'command -v'

### LINKED ISSUE #5240: "Error: We need either 'curl' or 'wget' present" error is misleading
refine checks for `curl` or `wget` using `which`. Such test fails if `which` is not installed.

https://github.com/OpenRefine/OpenRefine/blob/abab670399924daf2222b366e078b758f6192d80/refine#L121-L122

PhilipVigus:
Fixes #5240

This PR fixes a problem where the absence of the `which` command on users' computers was causing `refine.sh` to give a misleading error message.

After reading around the issue, various approaches were considered:
1. Just update the error message to make it clear that `which` might be one of the commands that's missing
2. Do an individual check against the `which` command with its own error message before the command is used. This would use `command -v`
3. Rework the script to not use `which` at all, replacing it with `command -v`. There seems to be quite a strong sentiment that it's not ideal to use, eg https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then, and replacing it should reduce instances where users are missing a command to run the script.
4. Do something that tries to use `which`, and if that fails then tries `command -v`. If both fail then it errors

I ultimately chose 3. A comment from frafra against the issue sums up the reason:

> `command` is not always available (it is not always available in busybox), but it is available as built-in in bash and other POSIX-compliant shells; it seems a good solution, since the script starts with `#!/usr/bin/env bash`, so it clearly targets bash

The downside of this solution is that it touches a lot more areas of the script. I've tried to test it as thoroughly as I can, but give the script's importance I would strongly suggest that others give it a run through and make sure I haven't missed anything.

Also, apologies for the number of white space changes. They were introduced by the autoformatting on my machine when I saved the file.

### DIFF
--- a/refine
+++ b/refine
@@ -986,7 +986,7 @@ fi
 # ----- Respond to the action given --------------------------------------------
 case "$ACTION" in
   build) build_prepare; do_mvn process-resources; do_mvn compile\ test-compile;;
-  clean) mvn clean;;
+  clean) do_mvn clean;;
   whitespace) whitespace $1;;
   test) test $1;;
   tests) test $1;;
--- a/refine
+++ b/refine
@@ -118,16 +118,16 @@ check_macosx() {
 }
 
 check_downloaders() {
-    CURL="`which curl 2> /dev/null`"
-    WGET="`which wget 2> /dev/null`"
-    
+    CURL="`command -v curl 2> /dev/null`"
+    WGET="`command -v wget 2> /dev/null`"
+
     if [ -z "$CURL" ] && [ -z "$WGET" ] ; then
         error "We need either 'curl' or 'wget' present in PATH to download external dependencies."
     fi
 }
 
 check_unzip() {
-    UNZIP="`which unzip 2> /dev/null`"
+    UNZIP="`command -v unzip 2> /dev/null`"
     
     if [ -z "$UNZIP" ] ; then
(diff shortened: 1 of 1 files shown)


refine
```diff
@@ -378,9 +378,9 @@ virtualenv_prepare() {
     
 # ----------------------------------------------------------------------------------------------
 
-mvn() {
-    mvn_prepare   
-    
+do_mvn() {
```
PhilipVigus, line 381: I had to change the name of this function, as using `command -v` caused recursive infinite loop in the `test()` function because of a clash between the name of function and the `mvn` CLI command. (no idea why. I tried a bunch of other ways of fixing it, but renaming the function was the simplest and only one I could get working).

refine
```diff
@@ -985,11 +985,11 @@ fi
 
 # ----- Respond to the action given --------------------------------------------
 case "$ACTION" in
-  build) build_prepare; mvn process-resources; mvn compile\ test-compile;;
+  build) build_prepare; do_mvn process-resources; do_mvn compile\ test-compile;;
```
PhilipVigus, line 988: See my earlier comment as to why this function name had to be changed

wetneb (approved):
Thank you very much!
It works on my side after converting a remaining `mvn` to `do_mvn`.

PR #5334, opened 2022-10-12
Made it so the Wikibase schema panel gets resized when the window gets resized

### LINKED ISSUE #5333: Schema editor (Wikimedia Commons) doesn't scroll down
Testing various brand new Commons features just created by @wetneb in #5322 and others. Most things seem to work super swimmingly. However, while constructing a schema in the schema editor, the schema refuses to scroll down.

Video showing the behavior:
https://user-images.githubusercontent.com/9881645/195176893-ed26fb2c-8c78-48da-9582-face5042f037.mov

Happens with this project: [Jan-Geerling-photographs-Rotterdam-1503.openrefine (1).tar.gz](https://github.com/OpenRefine/OpenRefine/files/9759034/Jan-Geerling-photographs-Rotterdam-1503.openrefine.1.tar.gz)

I'm testing with Version 3.7-SNAPSHOT [5f3fa43] on MacOS 12.6 using Chrome Version 105.0.5195.125 (Official Build) (arm64)

elebitzero:
Fixes #5333

- Made it so the Wikibase schema panel gets resized when the window gets resized, which fixes the scrollbar problem.

### DIFF
--- a/main/webapp/modules/core/scripts/project.js
+++ b/main/webapp/modules/core/scripts/project.js
@@ -105,7 +105,9 @@ function resizeAll() {
   ui.processPanel.resize();
   ui.historyPanel.resize();
   ui.dataTableView.resize();
-  SchemaAlignment.resize();
+  if (SchemaAlignment) {
+    SchemaAlignment.resize();
+  }
 }
 
 function initializeUI(uiState) {
--- a/main/webapp/modules/core/scripts/project.js
+++ b/main/webapp/modules/core/scripts/project.js
@@ -105,6 +105,7 @@ function resizeAll() {
   ui.processPanel.resize();
   ui.historyPanel.resize();
   ui.dataTableView.resize();
+  SchemaAlignment.resize();
 }
 
 function initializeUI(uiState) {
--- a/extensions/wikidata/module/scripts/schema-alignment.js
+++ b/extensions/wikidata/module/scripts/schema-alignment.js
@@ -309,16 +309,21 @@ SchemaAlignment.switchTab = function(targetTab) {
   $('.main-view-panel-tab-header[href="'+targetTab+'"]').addClass('active');
   $(targetTab).show();
   resizeAll();
-  var panelHeight = this._viewPanel.height();
-  this._schemaPanel.height(panelHeight);
-  this._issuesPanel.height(panelHeight);
-  this._previewPanel.height(panelHeight);
-  // Resize the inside of the schema panel
-  var headerHeight = this._schemaElmts.schemaHeader.outerHeight();
-  this._schemaElmts.canvas.height(panelHeight - headerHeight - 10);
 
   if (targetTab === "#view-panel") {
-     ui.dataTableView.render();
+    ui.dataTableView.render();
+  }
+};
+
+SchemaAlignment.resize = function() {
+  if (this._viewPanel) {
+    var panelHeight = this._viewPanel.height();
+    this._schemaPanel.height(panelHeight);
+    this._issuesPanel.height(panelHeight);
+    this._previewPanel.height(panelHeight);
+    // Resize the inside of the schema panel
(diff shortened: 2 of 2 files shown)


main/webapp/modules/core/scripts/project.js
```diff
@@ -105,6 +105,7 @@ function resizeAll() {
   ui.processPanel.resize();
   ui.historyPanel.resize();
   ui.dataTableView.resize();
+  SchemaAlignment.resize();
```
wetneb, line 108: This `SchemaAlignment` is coming from the Wikidata/Wikibase extension, so it might not be there if OpenRefine is running without this extension. I wonder if we have any extension point for this, or if we should build one?

I assume that this `resize()` business should really not be there in the first place, and we should just let the browser do its job of figuring out the layout, given sensible CSS directives. But changing to something else is not going to happen overnight.

wetneb (commented):
Thanks a lot for having a go at this!

main/webapp/modules/core/scripts/project.js
elebitzero: There already is an API: the `window.resize` event handler.  I tried registering the `SchemaAlignment.resize()` independently, but it didn't seem to work.  I think there is some dependency on the order of the other resize code in `resizeAll()` and then `SchemaAlignment.resize()`.

I can put a guard to check if `SchemaAlignment` is defined in case the extension is missing.

There is quite a bit of work to untangle the Wikidata UI from the rest of the UI, that is beyond the scope of this PR.

wetneb: > There is quite a bit of work to untangle the Wikidata UI from the rest of the UI, that is beyond the scope of this PR.

Where do you see other coupling points? I am not aware of anywhere else in the main code that refers to the Wikidata extension. The Wikidata extension inserts its Schema, Issues and Preview tabs in a hacky way, without having a corresponding extension point in the core, but that is not a dependency from the core to the extension.

I would rather try to introduce an extension point which lets extensions define new tabs in the main panel. The logic of switching between tabs would be handled by the core, and so would be the resizing.

elebitzero: Regarding the coupling points, I was just talking about how the sizing of the Wikidata tabs/panels is coupled to the resizing of the right-hand-side view panel somehow.  I tried decoupling by having Wikidata extension separately subscribe to `window.resize` event but that didn't work.

This PR is just trying to fix the issue, the coupling was already present. If you look at `switchTab()`, it was already calling `resizeAll()` in projects.js.  I just corrected the logic so it is happening on resize also.

If/when the Wikidata extension will be extracted into a separate repo, the decoupling work can be taken up as part of that.

Also my IDE says there are 27 dependencies from just wikidata/schema-alignment.js to the main webapp, pictured in the screenshot.  I haven't looked at the extension APIs, so not sure how many of these dependencies are using official extension API, versus private/internal APIs/variables.   That is just one file of Wikidata extensions. There are probably 100s of dependencies in the whole extension.

RHS: schema-alignment.js   LHS: main webapp scope
image

wetneb: > Also my IDE says there are 27 dependencies from just wikidata/schema-alignment.js to the main webapp, pictured in the screenshot.

Sure, in this direction it is totally expected.

>  I haven't looked at the extension APIs, so not sure how many of these dependencies are using official extension API, versus private/internal APIs/variables.

I think we have just not documented at all what the frontend extension points are meant to be, so it's hard to tell!

wetneb (approved):
I am fine with merging this as it stands, it is going to make it even more urgent to have a clean extension point for introducing those tabs, and that is probably a useful nudge to get it done.

frafra:
This PR breaks faceting on 3.7: #6070.
Master is not affected thanks to #5953.

wetneb:
We should backport this commit and release it in 3.7.6

PR #5357, opened 2022-10-19
Made the textarea for the edit cell/text facet editors larger

### LINKED ISSUE #5349: Make multi-line "edit cell dialog" larger if it contains more text and if user's screen resolution can handle it
When editing individual cells with quite a bit of text, I end up dragging the little 'enlargement corner' in the "edit cell" dialog all the time, which is not fun. I think, today, many user's screen resolution can totally deal with a slightly larger dialog window. Perhaps even one that flexibly adapts a bit to the length of text in it.

image

versus

image

elebitzero:
Fixes #5349

Increased the initial height for the cell and text facet editor when the content has more rows. Minimum initial height is 3 rows; maximum initial height is 15 rows (5x more than before). Scrollbars are displayed if the content has more than 15 rows.

The user can still manually make the textarea even taller if they want.  See recording for examples.

edit-cell-larger-size

### DIFF
--- a/main/webapp/modules/core/scripts/facets/list-facet.js
+++ b/main/webapp/modules/core/scripts/facets/list-facet.js
@@ -610,6 +610,8 @@ class ListFacet extends Facet {
     .trigger('select')
     .trigger('focus');
 
+    setInitialHeightTextArea(elmts.textarea[0]);
+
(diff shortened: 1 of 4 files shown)


wetneb (approved):
Yes I also thought adding an external library for this would be a bit much. Thank you!

PR #5360, opened 2022-10-21
wikibase: initial somevalue novalue editing support

Abbe98:
Part of #1713.

This makes it possible to set novalue / somevalue statements when editing by adding two "magic" keywords `#NOVALUE#` and `#SOMEVALUE#`.

Usage:
1. Have a column of magic keywords
2. Reconciliate column 
3. Clear reconciliation data
4. Use the magic keyword column in the schema editor

Known issues / remaining tasks:
 - ~~Not all data types are supported (only properly tested with dates so far)~~
 - ~~Statement merger doesn't work~~
 - ~~Preview is blank~~

Implementation details: 
 - `SpecialValueNoValueException` and `SpecialValueSomeValueException` extends `SkipSchemaExpressionException`, these should probably all instead extend a generic `SchemaException`
 - The "Some value" terminology is used rather than the "Unknown value" as the later is known to be confusing.
 - The magic keywords both ends and starts with a hash sign to avoid looking like possible fragment URIs.

### DIFF
--- a/extensions/wikibase/module/scripts/preview-renderer.js
+++ b/extensions/wikibase/module/scripts/preview-renderer.js
@@ -313,9 +313,13 @@ EditRenderer._renderValue = function(json, container) {
   if (mode === "wikibase-item" || mode === "wikibase-property") {
     EditRenderer._renderEntity(json.datavalue, container);
   } else if (json.snaktype === "somevalue") {
-    $('<span>somevalue</span>').appendTo(container);
+    $('<span></span>')
+      .text($.i18n('wikibase-schema/some-value'))
+      .appendTo(container);
   } else if (json.snaktype === "novalue") {
-    $('<span>novalue</span>').appendTo(container);
+    $('<span></span>')
+      .text($.i18n('wikibase-schema/no-value'))
+      .appendTo(container);
   } else {
     var jsonValue = JSON.stringify(json.datavalue);
     var fullJsonValue = JSON.stringify(json);
--- a/extensions/wikibase/module/scripts/preview-renderer.js
+++ b/extensions/wikibase/module/scripts/preview-renderer.js
@@ -312,6 +312,10 @@ EditRenderer._renderValue = function(json, container) {
   
   if (mode === "wikibase-item" || mode === "wikibase-property") {
     EditRenderer._renderEntity(json.datavalue, container);
+  } else if (json.snaktype === "somevalue") {
+    $('<span>somevalue</span>').appendTo(container);
+  } else if (json.snaktype === "novalue") {
+    $('<span>novalue</span>').appendTo(container);
   } else {
     var jsonValue = JSON.stringify(json.datavalue);
     var fullJsonValue = JSON.stringify(json);
--- a/extensions/wikibase/src/org/openrefine/wikibase/schema/WbVariableExpr.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/schema/WbVariableExpr.java
@@ -46,6 +46,9 @@ import com.google.refine.model.Cell;
 @JsonIgnoreProperties(ignoreUnknown = true)
 public abstract class WbVariableExpr<T> implements WbExpression<T> {
 
+    public static final String NO_VALUE_KEYWORD = "#NOVALUE#";
+    public static final String SOME_VALUE_KEYWORD = "#SOMEVALUE#";
+
     private String columnName;
 
     /**
@@ -100,9 +103,9 @@ public abstract class WbVariableExpr<T> implements WbExpression<T> {
             throws SkipSchemaExpressionException, QAWarningException, SpecialValueNoValueException, SpecialValueSomeValueException {
         Cell cell = ctxt.getCellByName(columnName);
         if (cell != null) {
-            if ("#NOVALUE#".equals(cell.toString())) {
+            if (NO_VALUE_KEYWORD.equals(cell.toString())) {
                 throw new SpecialValueNoValueException();
-            } else if ("#SOMEVALUE#".equals(cell.toString())) {
+            } else if (SOME_VALUE_KEYWORD.equals(cell.toString())) {
                 throw new SpecialValueSomeValueException();
             }
 
--- a/extensions/wikibase/src/org/openrefine/wikibase/schema/WbVariableExpr.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/schema/WbVariableExpr.java
@@ -26,6 +26,8 @@ package org.openrefine.wikibase.schema;
 
 import org.openrefine.wikibase.schema.exceptions.QAWarningException;
 import org.openrefine.wikibase.schema.exceptions.SkipSchemaExpressionException;
+import org.openrefine.wikibase.schema.exceptions.SpecialValueNoValueException;
+import org.openrefine.wikibase.schema.exceptions.SpecialValueSomeValueException;
 import org.openrefine.wikibase.schema.validation.ValidationState;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
@@ -90,12 +92,20 @@ public abstract class WbVariableExpr<T> implements WbExpression<T> {
      * Evaluates the expression in a given context, returning
      * 
      * @throws QAWarningException
+     * @throws SpecialValueNoValueException
+     * @throws SpecialValueSomeValueException
      */
     @Override
     public T evaluate(ExpressionContext ctxt)
-            throws SkipSchemaExpressionException, QAWarningException {
+            throws SkipSchemaExpressionException, QAWarningException, SpecialValueNoValueException, SpecialValueSomeValueException {
         Cell cell = ctxt.getCellByName(columnName);
         if (cell != null) {
+            if ("#NOVALUE#".equals(cell.toString())) {
+                throw new SpecialValueNoValueException();
+            } else if ("#SOMEVALUE#".equals(cell.toString())) {
+                throw new SpecialValueSomeValueException();
+            }
+
             return fromCell(cell, ctxt);
         }
         throw new SkipSchemaExpressionException();
--- a/extensions/wikibase/src/org/openrefine/wikibase/schema/strategies/SnakOnlyStatementMerger.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/schema/strategies/SnakOnlyStatementMerger.java
@@ -89,14 +89,18 @@ public class SnakOnlyStatementMerger implements StatementMerger {
         // Deliberately only comparing the pids and not the siteIRIs to avoid spurious mismatches due to federation
         if (!existingSnak.getPropertyId().getId().equals(addedSnak.getPropertyId().getId())) {
             return false;
-        } else if (existingSnak instanceof NoValueSnakImpl && addedSnak instanceof NoValueSnakImpl) {
-            return true;
-        } else if (existingSnak instanceof SomeValueSnakImpl && addedSnak instanceof SomeValueSnakImpl) {
-            return true;
-        } else {
+        } else if (existingSnak instanceof NoValueSnak) {
+            return addedSnak instanceof NoValueSnak;
+        } else if (existingSnak instanceof SomeValueSnak) {
+            return addedSnak instanceof SomeValueSnak;
+        } else if (addedSnak instanceof ValueSnak) {
+            // existingSnak is also a ValueSnak per the cases above
             Value existingValue = ((ValueSnak) existingSnak).getValue();
             Value addedValue = ((ValueSnak) addedSnak).getValue();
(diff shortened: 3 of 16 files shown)


extensions/wikibase/src/org/openrefine/wikibase/schema/WbDateVariable.java
```diff
@@ -57,13 +59,18 @@ public TimeValue fromCell(Cell cell, ExpressionContext ctxt)
...
             // parsed dates are accepted by converting them to strings
-            return WbDateConstant.parse(cell.value.toString());
+            return WbDateConstant.parse(value);
         } catch (ParseException e) {
-            if (!cell.value.toString().isEmpty()) {
+            if (value.equals("#NOVALUE#")) {
+                throw new SpecialValueNoValueException();
+            } else if (value.equals("#SOMEVALUE#")) {
+                throw new SpecialValueSomeValueException();
+            } else if (!value.isEmpty()) {
```
wetneb, line 71: Do we actually need special handling here given that you have added support in `WbVariableExpr` already?

extensions/wikibase/src/org/openrefine/wikibase/schema/WbVariableExpr.java
```diff
@@ -90,12 +92,20 @@ public void setColumnName(String columnName) {
...
     public T evaluate(ExpressionContext ctxt)
-            throws SkipSchemaExpressionException, QAWarningException {
+            throws SkipSchemaExpressionException, QAWarningException, SpecialValueNoValueException, SpecialValueSomeValueException {
         Cell cell = ctxt.getCellByName(columnName);
         if (cell != null) {
+            if (cell.toString() == "#NOVALUE#") {
+                throw new SpecialValueNoValueException();
+            } else if (cell.toString() == "#SOMEVALUE#") {
+                throw new SpecialValueSomeValueException();
+            }
```
wetneb, line 107: It's probably worth storing those special strings in some `public static final String` constant declared somewhere, reused thorough the codebase, especially because they will be used by a lot of different classes.

Also, we should not use ` == ` to compare Strings, but the `.equals()` method. It is customary to write the constant first (so that you are sure that you are calling the `.equals()` on a non-null object) -> so `"#NOVALUE#".equals(cell.toString())`.

wetneb (commented):
Very nice, congrats!
Do not forget to write unit tests as you go along - this part of the codebase has ample server-side tests so you should be able to imitate existing ones for most classes you touch.

extensions/wikibase/src/org/openrefine/wikibase/schema/WbDateVariable.java
Abbe98: @wetneb I'm afraid so as parsing appears to be used for things like QA. I also think such things needs to be added to many more data types :-/

extensions/wikibase/src/org/openrefine/wikibase/schema/WbVariableExpr.java
Abbe98: > It's probably worth storing those special strings in some public static final String constant declared somewhere, reused thorough the codebase, especially because they will be used by a lot of different classes.

Maybe in the "entityvalues" definitions?

>Also, we should not use == to compare Strings, but the .equals() method. It is customary to write the constant first (so that you are sure that you are calling the .equals() on a non-null object) -> so "#NOVALUE#".equals(cell.toString()).
Yeah I have a couple of things like that which I need to cleanup.

extensions/wikibase/src/org/openrefine/wikibase/schema/WbDateVariable.java
wetneb: I do not really understand why - could it be that your implementation in `WbVariableExpr` did not work so far because of the `==` instead of the `.equals()`? And what does this have to do with QA?

Abbe98: > I do not really understand why - could it be that your implementation in WbVariableExpr did not work so far because of the == instead of the .equals()? 

That might indeed be it.

> And what does this have to do with QA?

Given a magic keyword the `!value.isEmpty()` will raise a QA Warning/Error.

Abbe98: > >I do not really understand why - could it be that your implementation in WbVariableExpr did not work so far because of the == instead of the .equals()?

>That might indeed be it.

That does indeed take care of the issue!

extensions/wikibase/src/org/openrefine/wikibase/schema/strategies/SnakOnlyStatementMerger.java
```diff
@@ -87,9 +89,9 @@ public boolean match(Snak existingSnak, Snak addedSnak) {
         // Deliberately only comparing the pids and not the siteIRIs to avoid spurious mismatches due to federation
         if (!existingSnak.getPropertyId().getId().equals(addedSnak.getPropertyId().getId())) {
             return false;
-        } else if (existingSnak instanceof NoValueSnak && addedSnak instanceof NoValueSnak) {
+        } else if (existingSnak instanceof NoValueSnakImpl && addedSnak instanceof NoValueSnakImpl) {
```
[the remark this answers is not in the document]
wetneb: I am surprised it is necessary actually! In which situations does it make a difference to check for `NoValueSnakImpl` instead of `NoValueSnak`? In general for this library the user (us) is expected to refer to the interfaces only, not their implementations, so this is a sign that something fishy is going on.

But yes, this code was indeed not used before, since we could not have an `addedSnak` that would be a `NoValueSnak`.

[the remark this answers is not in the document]
Abbe98: In this case I never got the `NoValueSnak` variant to work for either `addedSnak` or `existingSnak`. Debugging indicated that `*Impl` instances were passed around.

However, I think the normal value snaks were passed around correctly.

Abbe98:
@trnstlntk there are some implementation details left but it should already work as expected.

extensions/wikibase/src/org/openrefine/wikibase/schema/strategies/SnakOnlyStatementMerger.java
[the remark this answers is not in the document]
wetneb: Ok! I guess the tests should help us get this to work. For the naming, the *Impl instances are also instances of the corresponding interfaces, so I really think this should not make any difference. Because the interfaces are meant to be the things we use here, I would prefer if we did not refer to the implementations.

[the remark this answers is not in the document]
Abbe98: Do you have suggestions for good tests here, I'm drawn to the JSON-blob based ones but they aren't the easiest to follow on my end.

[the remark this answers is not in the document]
wetneb: I would write at least something in `WbStatementExprTest`. Just copy one of the existing test cases there (for instance `testEvaluate`) and replace one of the cell contents (defined in `setRow`) by one of your magic constants. Then, construct the expected value with the appropriate No/SomeValueSnak. That should be fairly straightforward.

Then, for the SnakOnlyStatementMerger, you could follow the same strategy: copy some existing test case, change the some/both of the involved statements being matched to have some no value / somevalue case, and specify the expected matching result.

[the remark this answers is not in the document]
Abbe98: `WbStatementExprTest` doesn't seem to test anything other than `WbExpression` as the `evaluateTo` method is inherited from `WbExpressionTest`? At best I could catch the new exceptions.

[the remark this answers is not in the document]
wetneb: No, it does test `WbStatementExpr`. The evaluateExpr method can indeed work with any WbExpression, it is a utility that is used in many test classes. In this case it is used to test `WbStatementExpr` via a wrapper defined in the test class.

trnstlntk:
I have tested it (on the user side) and it works great. One example edit: https://commons.wikimedia.org/w/index.php?title=File%3AKoninklijk_Theater_Tuscinski_%E2%80%93_cinema_organ_02.jpg&type=revision&diff=703109719&oldid=703088605

The workflow is ... "specific" but I think it makes sense.

wetneb:
> The workflow is ... "specific" but I think it makes sense.

Just to emphasize: stated as such the workflow is very convoluted, but this is not the intended use case: this implementation is not meant to serve the case where the added value is constantly novalue/somevalue, but rather the case when this depends on which row.

See the issue for a proposed workflow for constant novalue/somevalue (which could be added as a follow-up PR).

extensions/wikibase/module/scripts/preview-renderer.js
```diff
@@ -312,6 +312,10 @@ EditRenderer._renderValue = function(json, container) {
   
   if (mode === "wikibase-item" || mode === "wikibase-property") {
     EditRenderer._renderEntity(json.datavalue, container);
+  } else if (json.snaktype === "somevalue") {
+    $('<span>somevalue</span>').appendTo(container);
+  } else if (json.snaktype === "novalue") {
+    $('<span>novalue</span>').appendTo(container);
```
wetneb, line 318: Since those strings are shown to the user, it would be good to:
- use user-friendly terms instead (Wikibase displays "unknown value" / "no value")
- make it possible to localize them with `$.i18n`

extensions/wikibase/src/org/openrefine/wikibase/schema/WbVariableExpr.java
```diff
@@ -90,12 +92,20 @@ public void setColumnName(String columnName) {
...
     @Override
     public T evaluate(ExpressionContext ctxt)
-            throws SkipSchemaExpressionException, QAWarningException {
+            throws SkipSchemaExpressionException, QAWarningException, SpecialValueNoValueException, SpecialValueSomeValueException {
         Cell cell = ctxt.getCellByName(columnName);
         if (cell != null) {
+            if ("#NOVALUE#".equals(cell.toString())) {
+                throw new SpecialValueNoValueException();
+            } else if ("#SOMEVALUE#".equals(cell.toString())) {
+                throw new SpecialValueSomeValueException();
```
wetneb, line 106: Since the `"#NOVALUE#"` and `"#SOMEVALUE#"` Strings have a special meaning defined by this class, I still think it would be good to have them as `public static final String` in this class, to make them part of the public interface of this class. This could be reused in tests, initially.

extensions/wikibase/src/org/openrefine/wikibase/schema/strategies/SnakOnlyStatementMerger.java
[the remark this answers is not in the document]
wetneb: We still need to change this to use the interfaces and not the implementations, just as a matter of good practice. Note that we are defining [FullyPropertySerializingNoValueSnak.java](https://github.com/OpenRefine/OpenRefine/pull/5360/files#diff-a9faad4d5d80e5c915fd5fe989e788ddf41d0205c9d7eb7d0734ea12d7d468d8) ourselves so we know those interfaces can have more implementations.

wetneb (changes requested):
Are you still working on this @Abbe98? I have had a look at the outstanding things needing to be resolved before we merge the PR, see below. As mentioned before, we also need Java tests for this.

extensions/wikibase/src/org/openrefine/wikibase/schema/strategies/SnakOnlyStatementMerger.java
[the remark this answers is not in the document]
Abbe98: Yeah sure, problem still is that changing to the interfaces break it.

extensions/wikibase/module/scripts/preview-renderer.js
Abbe98: +1 I would suggest "some value" and "no value" as Commons does it.

wetneb:
Hi @Abbe98, are you still working on this? I just started a discussion on releasing 3.7-beta1: if you plan to finish this soon and want us to wait for you, let us know: https://forum.openrefine.org/t/releasing-3-7-beta1/214

wetneb:
I would merge this and release it in 3.7. I'll open a documentation PR for it too.

Abbe98:
Tanks @wetneb! :rocket:

wetneb (approved):

Issue #5221, opened 2022-08-30
Allow regular Wikimedians to edit Wikidata or Wikimedia accounts with their regular accounts, without requiring bot passwords


trnstlntk:
I notice during tests and demos that 'regular' Wikimedians (the kind of people who, like me, usually only do small batches and who don't write/operate bots) get confused when OpenRefine asks them to provide / get bot passwords or log in with an owner-only consumer 
<img width="558" alt="image" src="https://user-images.githubusercontent.com/9881645/187475792-b2dda5e4-583b-4c38-aeeb-683ba58fde58.png">
Can we simplify this process, or make this even unnecessary somehow?

### Additional context
I'm creating this issue because at some point I got the following feedback in my log:
```
19:20:19.823 [..kibaseapi.ApiConnection] API warning [main]: Subscribe to the mediawiki-api-announce mailing list at <https://lists.wikimedia.org/postorius/lists/mediawiki-api-announce.lists.wikimedia.org/> for notice of API deprecations and breaking changes. Use [[Special:ApiFeatureUsage]] to see usage of deprecated features by your application. (2283ms)
19:20:19.824 [..kibaseapi.ApiConnection] API warning [login]: Main-account login via "action=login" is deprecated and may stop working without warning. To continue login with "action=login", see [[Special:BotPasswords]]. To safely continue using main-account login, see "action=clientlogin". (1ms)
```
and while I'm not well-versed in this stuff, it looks as if `"action=clientlogin"` is something worth looking into?

Abbe98:
If the issue is mainly concerned with WMF wikis then OAuth using a server side component might be a good solution. 

OR would store the access token and secret. The server would store the application token and secret. Shared instances like on PAWS could set a custom OAuth callback URL.

Abbe98:
> OAuth requires bundling a secret in the app, and as an open source desktop application this is a difficult thing for us to do. 

OAuth directly targeting the app would be tricky to begin with as the callback URI could vary. I'm, however, suggesting that one uses a server side component to deal with this. The secrets could be stored in environment variables similarly to how most OAuth tools on Tool Forge do it.

wetneb:
[...]

PR #5374, opened 2022-10-24
wikibase: remove bot password alert upon login

Abbe98:
Part of #5221

This removes the bot password alert that is raised upon signing into Wikibase, as there is a less alarming notice on the main dialog.


wetneb (approved):
The CI failure is spurious. Merging soon.

PR #5378, opened 2022-10-25
Add See Also tips for find() match()

thadguidry:
This might help a bit for issues such as https://groups.google.com/g/openrefine/c/tsA0FYnOLv4/m/CyUxSzzkAAAJ


wetneb (commented):
I would rather put the "See also" tips at the end, not the beginning, of each section. Also, I would rather capitalize it as "See also", not "See Also"

wetneb (approved):

PR #5381, opened 2022-10-25
Also add recon `cell` tip to current docs

thadguidry:


elebitzero (approved):
Looks good after checking the docs and directions.

PR #5383, opened 2022-10-25
Fixed flaky export project tests to properly verify downloads

elebitzero:
- Fixed flaky export project tests failing with error:
`Timed out retrying after 4000ms: cy.readFile("cypress/downloads/1659083769462.xlsx") failed because the file does not exist at the following path:
`

The old code would return as soon as file was created, but possibly before file contents were flushed to disk, so sometimes the file would be empty (filesize: 0).

Solution:
Use Cypress's file download recipe to check that the downloaded files have contents. (Ref: https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/testing-dom__download/cypress/e2e/utils.js)

### DIFF
--- a/main/tests/cypress/cypress/integration/project/project-header/export_project.spec.js
+++ b/main/tests/cypress/cypress/integration/project/project-header/export_project.spec.js
@@ -18,6 +18,25 @@ describe(__filename, function () {
       })
     })
   }
+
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thanks for looking into this!

PR #5395, opened 2022-10-31
fix(sec): upgrade com.fasterxml.jackson.core:jackson-databind to 2.14.0-rc1

chncaption:
### What happened？
There are 1 security vulnerabilities found in com.fasterxml.jackson.core:jackson-databind 2.13.4
- [CVE-2022-42003](https://www.oscs1024.com/hd/CVE-2022-42003)

### What did I do？
Upgrade com.fasterxml.jackson.core:jackson-databind from 2.13.4 to 2.14.0-rc1 for vulnerability fix

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
     <!-- dependency versions -->
-    <jackson.version>2.13.4</jackson.version>
+    <jackson.version>2.14.0-rc1</jackson.version>
     <servlet-api.version>4.0.1</servlet-api.version>
     <butterfly.version>1.2.3</butterfly.version>
     <slf4j.version>2.0.3</slf4j.version>
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
     <!-- dependency versions -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -72,7 +72,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
     <!-- dependency versions -->
-    <jackson.version>2.13.4</jackson.version>
+    <jackson.version>2.14.0-rc1</jackson.version>
```
wetneb, line 75: How about 2.13.4.2 instead? I'd rather avoid relying on a RC.

wetneb (commented):
Thank you!

pom.xml
tfmorris: Yes, 2.13.4.2 is definitely better and includes the fix according to https://github.com/FasterXML/jackson-databind/issues/3590

wetneb:
So only some Jackson artifacts got this micro-patch release. `jackson-annotations` and `jackson-core` did not get a micro-patch release (since they were left unchanged), which means that we cannot steer its version from the same Maven property. I'm not so sure whether it's worth introducing more Maven variables or just accept the RC (which is very likely fine too).

wetneb:
Since we are not using the `UNWRAP_SINGLE_VALUE_ARRAYS` deserialization feature, we are not affected by the vulnerability, so I would say it is also fine to wait for the next stable release.

wetneb:
Closing then.

PR #5397, opened 2022-11-01
Add option to import Excel as text only. Closes #4838

### LINKED ISSUE #4838: Option to render non-text Excel cells as text when importing
It would be easier if OpenRefine offered an option to parse all cells as text when importing Excel files, as it does with csv files. When refining e.g. publication dates, the default behaviour of parsing digit characters in Excel files as numbers can lead to unwanted results and is cumbersome to fix.

### Proposed solution
Offer the option to leave a checkbox "Parse cell text into numbers, dates, ..." unchecked at import, as with csv files.

### Alternatives considered
All Excel files could be converted to csv files before importing them into OpenRefine. 

### Additional context
image

tfmorris:
Fixes #4838

Adds an option (default=False) to import all cell values using a textual rendering, rather than the value that is stored in Excel. 

This can be useful for datatypes which aren't supported by OpenRefine such as times, dates, intervals or where the datatype information is only available in the format string (e.g. integers).

### DIFF
--- a/main/src/com/google/refine/importers/ExcelImporter.java
+++ b/main/src/com/google/refine/importers/ExcelImporter.java
@@ -42,12 +42,13 @@ import java.util.ArrayList;
 import java.util.List;
 
 import org.apache.poi.ooxml.POIXMLException;
-import org.apache.poi.ss.usermodel.DateUtil;
 import org.apache.poi.hssf.usermodel.HSSFWorkbook;
 import org.apache.poi.poifs.filesystem.FileMagic;
 import org.apache.poi.openxml4j.exceptions.InvalidFormatException;
 import org.apache.poi.poifs.filesystem.POIFSFileSystem;
 import org.apache.poi.ss.usermodel.CellType;
+import org.apache.poi.ss.usermodel.DateUtil;
+import org.apache.poi.ss.usermodel.DataFormatter;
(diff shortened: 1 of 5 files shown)


main/webapp/modules/core/langs/translation-ar.json
```diff
@@ -195,6 +195,7 @@
     "core-index-parser/json-parser": "انقر على أول عقدة JSON { } المقابلة للسجل الأول المراد تحميله.",
     "core-index-parser/preserve-empty": "الاحتفاظ بالقيم الفارغة",
     "core-index-parser/store-archive": "حفظ ملف الأرشيف",
+    "core-index-parser/force-text": "Import all cells as text",
```
wetneb, line 198: It is not necessary to add English strings to other language files, because there is already an implicit fallback mechanism in place. In fact, by adding those English strings there, we are making it impossible for Weblate to flag those strings as needing translation - so we should definitely remove those.

tfmorris: Thanks for the heads up. I'll remove them before removing the "Draft" label. I've updated the instructions in the wiki to reflect this guidance - https://github.com/OpenRefine/OpenRefine/wiki/Translate-OpenRefine

wetneb (approved):
Perfect, tested and it works great. Even for monetary values! Thank you!

PR #5409, opened 2022-11-06
Implemented splitting a string repeatedly by a sequence

### LINKED ISSUE #3948: Support splitting a string to an array based on a set of repeated string lengths
Originated in #3929.
**Minor Update 2021-05-28:** Make sure to see comments below about preference to create new operation rather than amending splitByLengths() as in proposal

Currently there is no easy way to split a string into an array based on a set of repeated string lengths. For example if a cell contains a string that concatenates several fixed length strings together with no separators there is no straightforward way of splitting that string back to the original fixed length strings.

Example 1:
Cell contains a string created by concatenating four ID numbers each four digits long. The original IDs are:

- 1234
- 5678
- 9012
- 3456

The cell content is `1234567890123456`

The desire is to get an array with the four IDs separator: `["1234", "5678", "9012", "3456"]`

Example 2:
Cell contains a string created by concatenating pairs of fixed length strings. Each pair consists of two book identifiers - the 10 character OCLC number and the 13 character ISBN. The original string pairs are:

- 1128883299,9780399208539
- 0890160519,9780007586547

The cell content is `1128883299978039920853908901605199780007586547`

The desire it to get an array: `["1128883299", "9780399208539", "0890160519", "9780007586547"]`

### Proposed solution
Modified from original proposal by @thadguidry in # #3929

Enhance the existing GREL splitByLengths() function to support final boolean argument indicating whether the list of split lengths should be used as a repeated pattern for the whole length of the string.

`"12345678".splitByLengths(4,true)`
would output
`["1234", "5678"]`

`"12345678".splitByLengths(3,1,true)`
would output
`["123", "4", "567", "8"]`

The boolean value should default to `false` - i.e. if it is omitted then `false` is used so both of

- `"12345678".splitByLengths(4)`
- `"12345678".splitByLengths(4,false)`

Output `["1234"]`

If the repeated pattern does not exactly match the length of the string, the remaining characters should be included as the last item in the array. i.e.

`"12345678".splitByLengths(3,true)`
would output
`["123", "456", "78"]`

This is consistent with the current behaviour of splitByLengths() (`"12345678".splitByLengths(3,3,3) -> `["123", "456", "78"]`)

If this proposal were implemented the two example cases above could be resolved using:

- Example 1: `value.splitByLengths(4, true)`
- Example 2: `value.splitByLengths(10, 13, true)`

### Alternatives considered
Create a new operation that works as specified, that can exist alongside the current "splitByLengths()" operation. e.g. "splitByRepeatedLengths()" or "splitRepeatLengths()" (latter suggested in a comment below for increased brevity). I do not believe this is necessary if we implement as described here because there is no breaking change to the existing splitByLengths operation. If the implementation leads to a situation where the behaviour of "splitByLengths" introduces a breaking change, then we should consider a new operation instead. An advantage of having a new separate operation (as noted by @wetneb below) is that the boolean argument is no longer needed, keeping the inputs to just a list of integers for both the existing and the new operation

Similar outcomes to the proposed function might be currently achieved by:

- Using `match` with an appropriate regular expression 
- Fully specifying the set of repeated splits within splitByLengths()

But these are more complex or time consuming or require certain assumptions to be made about the total repetitions in the starting string

### Additional context
Originally proposed by @thadguidry in #3940. That issue was closed only to allow for a clean start with the proposal

achinthaa:
Fixes #3948  

- Implemented additional function on strings, which allows splitting a string by a sequence as described in Issue #3948 

Examples of functionality (from #3948):

"12345678".splitRepeatLengths(4) would output:
["1234", "5678"]

"12345678".splitRepeatLengths(3,1) would output
["123", "4", "567", "8"]

Splitting proceeds until string termination, and last element may not correspond to one of the sequence lengths in the case that the sequence does not perfectly divide the string.
e.g., "internationalization".splitByLengths(5, 2) would output
[ "inter", "na", "tiona", "li", "zatio", "n" ]

This is consistent with the splitByLengths() function, which this PR aims to extend.


wetneb:
Thank you for this contribution!
To be completely honest I am a bit unsure about the use cases for this function - but @ostephens took the time to write a nice issue about it so that's already enough to make it worth implementing.

I haven't had a close look at the code yet, but it would be nice if you could write some unit tests for your function.
Also, you can run  `mvn formatter:format` to format your code according to the style used in this repository.

achinthaa:
@wetneb Thank you so much for your feedback! Actually, we are university students and this was our first ever contribution to an OSS project. Could you please point us in the right direction for where we can write the tests and which framework to use? We have also never used mvn before, how would we run this command. Your help is greatly appreciated, thank you so much!

wetneb:
To write tests, I recommend that you just imitate existing ones. For instance, you could have a look at `main/tests/server/src/com/google/refine/expr/functions/ToDateTests.java`. There are plenty of other examples in the folders nearby.

To set up Maven (the `mvn` command) you could have a look at our documentation:
https://docs.openrefine.org/technical-reference/build-test-run#maven-optional

wetneb:
Hi @achinthaa, are you still working on this?

achinthaa:
> Hi @achinthaa, are you still working on this?

Hi @wetneb, yes we still are! We have just finished up exams and we will be finishing this up too! Sorry about the delay

PR #5411, opened 2022-11-08
Preserve pagination after actions which trigger an update of the grid

### LINKED ISSUE #33: Manually selecting a reconciliation match returns the grid to page 1.
_Original author: staringmonkey (May 14, 2010 14:09:53)_

1. From the header of a data column select Reconcile -&gt; Start Reconciling.
2. Select and appropriate topic and click Start Reconciling.
3. Wait for the reconcile to complete.
4. Click the &quot;next page&quot; link.
5. Find a cell that which has several possible reconciliation matches and
click the check-mark or double-check-mark image.

Reconciliation is completed correctly, however, grid is returned to page
one.  This makes it nearly impossible to complete manual reconciliation for
large datasets.

<b>What version of the product are you using? On what operating system?</b>

SVN HEAD on OSX.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=33_

### LINKED ISSUE #570: Editing records and advancing forward page by page- how the list refreshes and displays the records

### LINKED ISSUE #572: Undo/Redo page reset problem

wetneb:
Fixes #33, fixes #570, fixes #572.

In short: this preserves the current paging settings when applying an operation which triggers a refresh of the grid.
This change is based on the new architecture (4.0 branch), because this architecture makes this change more natural (although it would not be impossible to implement this in the 3.x series).

## Known issues

* This is quite radical: it does not just affect reconciliation actions, but basically all other operations. The frontend will attempt to preserve the position in the grid for all of them. I am thinking for some of them it does not actually make sense to do that: for instance, when doing data extension, we create additional rows needed by the record structure, and so it is pointless to preserve paging based on row ids after that. Same for the transpose operations. So I will need to refine this to preserve the old behaviour for those operations (basically all operations which are not row/record wise).
* This removes the input field introduced by @antoine2711 in #2639 to let the user pick a particular page number. With this change there is no easily computable notion of page number anymore, and so I have reverted that change to show the range of rows being displayed, as it was before. We could consider making those editable, but I am not sure about the desired UX for that. If we make both of them editable then we let the user input something that is inconsistent with the page size selected on the left-hand side. We could make only the start editable, perhaps? But then I am not sure the UI will be clear enough to show that the user understands the meaning of those two numbers? Also, when a temporary sort is applied, those numbers bear no relationship to the row numbers displayed in the grid.

## Overview of the changes

This required a bit more changes in the backend than I expected:
* introducing the ability to request a page ending at a given row/record id (not just starting at a given row/record id)
* changing the `GridState` interface to support this, which means adapting the three existing implementations of that interface (testing, local and spark)
* changing the sorting API so that both the original row ids and the new row ids (after sorting) are available. This is because we cannot rely on the original row ids for pagination, since they are out of order. This was a significant change but it ended up simplifying the `GridState` interface quite a bit and it opens up the possibility of new optimizations (such as avoiding to re-sort the grid every time a request is made). I have not made those optimizations yet because I wanted to first check if the UX makes sense.

Trying it out interactively, I find it to work as expected.
Perhaps one thing that is a bit counter-intuitive is that when you click on "Previous" to get a previous page, and then apply an action, the grid will be refreshed in a way that preserves the last row being displayed in the page, not the first one. It is something that we can change easily though.

Marking as draft per the additional changes mentioned above.

### DIFF
--- a/extensions/database/module/scripts/index/database-import-controller.js
+++ b/extensions/database/module/scripts/index/database-import-controller.js
@@ -309,7 +309,7 @@ Refine.DatabaseImportController.prototype._getPreviewData = function(callback, n
             result.rowModel = data;
             callback(result);
           },
-          "jsonp"
+          "json"
         );
       },
       "json"
--- a/main/src/main/java/org/openrefine/commands/expr/PreviewExpressionCommand.java
+++ b/main/src/main/java/org/openrefine/commands/expr/PreviewExpressionCommand.java
@@ -201,7 +201,12 @@ public class PreviewExpressionCommand extends Command {
                 Properties bindings = ExpressionUtils.createBindings();
 
                 if (Mode.RowBased.equals(engine.getMode())) {
-                    List<IndexedRow> rows = state.getRows(engine.combinedRowFilters(), sortingConfig, 0, limit);
+                    GridState sorted = state;
+                    if (!SortingConfig.NO_SORTING.equals(sortingConfig)) {
+                        // TODO cache appropriately
+                        sorted = state.reorderRows(sortingConfig, false);
+                    }
+                    List<IndexedRow> rows = sorted.getRowsAfter(engine.combinedRowFilters(), 0, limit);
                     for (IndexedRow indexedRow : rows) {
                         Cell cell = indexedRow.getRow().getCell(cellIndex);
                         Record record = null;
@@ -210,7 +215,12 @@ public class PreviewExpressionCommand extends Command {
                                         repeat, repeatCount));
                     }
                 } else {
-                    List<Record> records = state.getRecords(engine.combinedRecordFilters(), sortingConfig, 0, limit);
+                    GridState sorted = state;
+                    if (!SortingConfig.NO_SORTING.equals(sortingConfig)) {
+                        // TODO cache appropriately
+                        sorted = state.reorderRecords(sortingConfig, false);
+                    }
+                    List<Record> records = sorted.getRecordsAfter(engine.combinedRecordFilters(), 0, limit);
                     for (Record record : records) {
                         for (IndexedRow indexedRow : record.getIndexedRows()) {
                             Cell cell = indexedRow.getRow().getCell(cellIndex);
--- a/extensions/wikibase/module/scripts/dialogs/perform-edits-dialog.js
+++ b/extensions/wikibase/module/scripts/dialogs/perform-edits-dialog.js
@@ -60,7 +60,7 @@ PerformEditsDialog.launch = function(logged_in_username, max_severity) {
         tag: WikibaseManager.getSelectedWikibaseTagTemplate(),
         maxEditsPerMinute: WikibaseManager.getSelectedWikibaseMaxEditsPerMinute()
       },
-      { includeEngine: true, cellsChanged: true, columnStatsChanged: true },
+      { includeEngine: true, cellsChanged: true, columnStatsChanged: true, rowIdsPreserved: true, recordIdsPreserved: true },
       { onDone: function() { dismiss(); } }
     );
   };
--- a/extensions/wikibase/src/org/openrefine/wikibase/schema/WikibaseSchema.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/schema/WikibaseSchema.java
@@ -201,7 +201,7 @@ public class WikibaseSchema implements OverlayModel {
             throw new IllegalStateException("The schema has not been validated before being evaluated");
         }
         List<EntityEdit> result = new ArrayList<>();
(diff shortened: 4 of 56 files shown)


main/src/main/java/org/openrefine/commands/expr/PreviewExpressionCommand.java
```diff
@@ -201,7 +201,12 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
                 Properties bindings = ExpressionUtils.createBindings();
 
                 if (Mode.RowBased.equals(engine.getMode())) {
-                    List<IndexedRow> rows = state.getRows(engine.combinedRowFilters(), sortingConfig, 0, limit);
+                    GridState sorted = state;
+                    if (!SortingConfig.NO_SORTING.equals(sortingConfig)) {
+                        // TODO cache appropriately
+                        sorted = state.reorderRows(sortingConfig, false);
```
wetneb, line 207: This is the sort of place that would benefit from the caching I mentioned above.

extensions/database/module/scripts/index/database-import-controller.js
```diff
@@ -309,7 +309,7 @@ Refine.DatabaseImportController.prototype._getPreviewData = function(callback, n
             result.rowModel = data;
             callback(result);
           },
-          "jsonp"
+          "json"
```
wetneb, line 312: Some parts of the code were still making calls to the `get-rows` command with JSONP, so I have used this opportunity to clean that up.

wetneb:
The failures seem spurious so I consider that this is now good to go!
I will leave this open for a bit longer but will otherwise merge in the coming days, since this is going to 4.0, which is experimental anyway.

thadguidry:
@wetneb Sounds good and makes sense about the revert for @antoine2711 pagination commits.  Can you describe what a "page" is now in the 4.0 architecture (maybe document it also)?  Is a "page" = range of rows or records (depending on OpenRefine mode) or is it something conceptually different and abstract now and not that particular definition?

thadguidry:
Sounds like it is indeed then, "a range of rows" would you agree?  Perhaps we can drop the use of the term "page" and indeed just move to the actual concept term, that of "range" instead of using a term of "size" which seems confusing.

So maybe "range" of a grid is the replacement term for "page" in the 4.0 architecture and we document it as such?

PR #5426, opened 2022-11-11
fix flaky test in main #5425

### LINKED ISSUE #5425: junit flaky test in main
We detected the existence of flaky junit test in main using [NonDex](https://github.com/TestingResearchIllinois/NonDex)

test:
https://github.com/OpenRefine/OpenRefine/blob/ed5e90e59cfd5edb4cac620b5966eae06d5787c7/main/tests/server/src/com/google/refine/clustering/binning/BinningClustererTests.java#L77

### To Reproduce
`mvn -pl main edu.illinois:nondex-maven-plugin:2.1:nondex`

### Current Results
The junit test can fail since the order of elements in clusterer can randomize

### Expected Behavior
The junit should pass always
PR : https://github.com/OpenRefine/OpenRefine/pull/5426

### Screenshots
Debug Screenshot
image

anantdahiya8:
Fixes #5425

The order of the elements in json can changes and cause the junit test to fail. Using a `LinkedHashMap `maintains the order of the elements

### DIFF
--- a/main/src/com/google/refine/clustering/binning/BinningClusterer.java
+++ b/main/src/com/google/refine/clustering/binning/BinningClusterer.java
@@ -38,6 +38,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


wetneb:
Thank you!

I would say that we do not actually care about the order of clusters (or cluster components) from a user perspective, so I am perhaps a bit reluctant to change the main code to preserve order because a test happens to currently require that incidentally. I am not sure how much more expensive a `LinkedHashMap` is compared to a `HashMap`, but even if the cost happens to be affordable it feels like the wrong method here.

The flaky test is a test that is meant to control the serialization of clustering results - so really just the translation from results to the JSON format.
So I would rather fix this by making this test accept various serialization orders (for instance by keeping only one cluster, say the one for `a` and `à`, and checking that the JSON output is one of the two possibilities of ordering `a` and `à` in their common cluster - if the `TreeMap`'s order is not deterministic).
You could also try to make this test narrower - arguably a test for serialization should not include the computation of the clusters at all, it could already set those itself and just test the serialization.

wetneb:
Closing per inactivity.

Issue #5430, opened 2022-11-12
Add QA warning for file upload with filename starting with a namespace prefix


wetneb:
When uploading a file to a Wikibase, if the filename supplied by the user starts with `File:` (or another namespace in use on the wiki) then the upload will fail.

### To Reproduce
1. First, create a wikimedia commons schema for file upload
2. Then, make sure the filenames start with `File:`
3. Finally, attempt the upload

### Current Results
No file is uploaded

### Expected Behavior
The upload dialog should warn that the wiki will reject the uploads based on the filenames.

### Versions
 - OpenRefine: 3.7-snapshot
 

### Additional context
The list of namespace prefixes and their aliases can be extracted from the API, such as https://fr.wikipedia.org/w/api.php?action=query&meta=siteinfo&siprop=namespaces&formatversion=2.

Identified by @trnstlntk and @ookgezellig at the Wikimedia NL hackathon today

wetneb:
This should be superseded by #5433 as the `:` character gets normalized at upload time.

PR #5435, opened 2022-11-14
Add explicit Wikibase QA warning when file names are normalized.

### LINKED ISSUE #5433: Improve handling of file name normalization for file upload in Wikibase extension
Currently, uploading files with the Wikibase extension will fail if the proposed filenames contain characters that should be normalized, such as `:` or `/`.

### Proposed solution

Instead, normalize those characters ahead of time (at schema evaluation time) and generate a QA warning indicating that this normalization has taken place.

### Alternatives considered
We could also not do the normalization and just block the upload with a critical QA warning, forcing the user to do the normalization themselves (so that they control how it is done), but that is needlessly rude I would say.

### Additional context

Bug found by @trnstlntk and @ookgezellig.

wetneb:
This should let the user identify cases such as additional `File:` prefixes too, which was the problem encountered in #5430. Closes #5433.


elebitzero (approved):

extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizer.java
```diff
@@ -49,13 +48,12 @@ public void batchIsBeginning() {
...
      * Attempt of a local implementation of the file name normalization that is done in MediaWiki. Assumes a non-empty
-     * file name as input.
+     * file name as input. This assumes that some special characters have been replaced already: `:`, `/` and `\`
      * 
      * @return
      */
     protected String normalizeFileName(String filename) {
-        String replaced = filename.replaceAll(normalizedFileNameChars.pattern(), "-")
-                .replaceAll("_", " ");
+        String replaced = filename.replaceAll("_", " ");
```
elebitzero, line 56: Since this method is only normalizing spaces and other normalization is done outside, it could be renamed to something like normalizeFileNameSpaces.

PR #5436, opened 2022-11-14
Add desktop file and appstream metadata

### LINKED ISSUE #4229: openrefine: add desktop file and appstream metadata for Linux distributions
Hello,

as part of the effort to package OpenRefine for Debian I have created a desktop file and appstream metadata for better desktop integration in Linux distributions. 

"AppStream is a collaborative effort for enhancing the way we interact with the software repositories provided by the distribution by standardizing sets of additional metadata. "

https://freedesktop.org/software/appstream/docs/chap-Quickstart.html

In order to make OpenRefine more visible on the user's desktop, a desktop file provides a short cut in an application menu (or e.g. Gnome shell) and you just have to click on it to start the application. You can also add keywords like rdf, csv, marc, odf, etc. to list openrefine in search results. Even if you don't know the exact program name but the file formats the program should be presented to users.

The appstream metainfo file adds OpenRefine to software centers, provides a short description of the main functionalities and screenshots, so users can more easily decide whether they would like to install a certain application or not.

I'm attaching preliminary versions of these files and an icon with 128x128 pixel resolution. The screenshots are currently missing. Ideally someone would upload them to openrefine.org and then point the links in the appstream metainfo file to them. When OpenRefine has been accepted into Debian I will update this bug report and add some screenshots as well. I suggest to add these files under the packaging directory. Then other distributions could use them as well.

[openrefine_desktop_appstream.zip](https://github.com/OpenRefine/OpenRefine/files/7380883/openrefine_desktop_appstream.zip)

wetneb:
Closes #4229.


elebitzero (approved):
Looks good if the screenshots will be added later as the example ones are still there.

wetneb:
Oh I didn't realize that, thanks for pointing this out! Let's get this fixed before then.

PR #5447, opened 2022-11-16
Fix CodeQL flagged vulnerability

tfmorris:
Switch from string concatenation to string array for command line construction. I don't think this is really an issue, but hopefully it's an easy fix to quiet CodeQL

### DIFF
--- a/server/src/com/google/refine/Refine.java
+++ b/server/src/com/google/refine/Refine.java
@@ -520,7 +520,7 @@ class RefineClient extends JFrame implements ActionListener {
         } else if (SystemUtils.IS_OS_MAC_OSX) {
             rt.exec(new String[] { "open ", String.valueOf(uri) });
         } else if (SystemUtils.IS_OS_LINUX) {
-            rt.exec(new String[] { "xdg-open ", String.valueOf(uri) });
+            rt.exec(new String[] { "xdg-open", String.valueOf(uri) });
         } else {
             logger.warn("Java Desktop class not supported on this platform. Please open %s in your browser", uri.toString());
         }
--- a/server/src/com/google/refine/Refine.java
+++ b/server/src/com/google/refine/Refine.java
@@ -516,7 +516,7 @@ class RefineClient extends JFrame implements ActionListener {
         Runtime rt = Runtime.getRuntime();
 
         if (SystemUtils.IS_OS_WINDOWS) {
-            rt.exec(new String[] { "rundll32 url.dll,FileProtocolHandler ", String.valueOf(uri) });
+            rt.exec(new String[] { "rundll32 ", "url.dll,FileProtocolHandler ", String.valueOf(uri) });
         } else if (SystemUtils.IS_OS_MAC_OSX) {
             rt.exec(new String[] { "open ", String.valueOf(uri) });
         } else if (SystemUtils.IS_OS_LINUX) {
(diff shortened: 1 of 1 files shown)


elebitzero (approved):

server/src/com/google/refine/Refine.java
```diff
@@ -516,11 +516,11 @@ private void openBrowserFallback() throws IOException {
         Runtime rt = Runtime.getRuntime();
 
         if (SystemUtils.IS_OS_WINDOWS) {
-            rt.exec("rundll32 url.dll,FileProtocolHandler " + uri);
+            rt.exec(new String[] { "rundll32 url.dll,FileProtocolHandler ", String.valueOf(uri) });
```
wetneb, line 519: Looking at the documentation of this method, it is not clear to me if the array of strings is expected to match the canonical tokenization of the command, or if the strings are just internally concatenated together anyway.
So for this case I wonder if it should not rather be 
```suggestion
            rt.exec(new String[] { "rundll32 ", "url.dll,FileProtocolHandler ", String.valueOf(uri) });
```
And I wonder if we can remove the trailing spaces around the strings (also below).

Did someone test if this still works on Windows? @thadguidry perhaps?

wetneb (commented):
It works on Linux, but I'd like to double-check it works on Windows and MacOS too.

server/src/com/google/refine/Refine.java
```diff
@@ -516,11 +516,11 @@ private void openBrowserFallback() throws IOException {
...
         } else if (SystemUtils.IS_OS_MAC_OSX) {
-            rt.exec("open " + uri);
+            rt.exec(new String[] { "open ", String.valueOf(uri) });
         } else if (SystemUtils.IS_OS_LINUX) {
-            rt.exec("xdg-open " + uri);
+            rt.exec(new String[] { "xdg-open ", String.valueOf(uri) });
```
wetneb, line 523: On Linux this seems to work as well, and looks a bit cleaner I guess:
```suggestion
            rt.exec(new String[] { "xdg-open", String.valueOf(uri) });
```

server/src/com/google/refine/Refine.java
thadguidry: @wetneb I tried running tests after checking out PR and seems I get failures.  Here's the surefire test output:
[tests.zip](https://github.com/OpenRefine/OpenRefine/files/10053209/tests.zip)

thadguidry: Looks like it's something to do with the space in my repo path (which is perfectly normal to have).  
`E:\GitHub Repos\OpenRefine`
but after running the tests, I notice that this new folder with empty subfolders structure was made??? weird.
`E:\GitHub%20Repos\OpenRefine\main\target\test-classes`

wetneb: It is curious that this PR would change anything to that, I do not think that the lines affected by the PR are coverered by tests. Do the tests pass on the master branch? If not I would open a specific issue about this.
Did you try running OpenRefine interactively on this branch? I would be interested to know if the browser opens automatically.

thadguidry: OS detected and browser (my default Edge on Windows 11) opened just fine.  Anything else to test on this branch?
Yeah, I'll open an issue for failing tests on Windows.

wetneb: Thanks, that's all I wanted to know :)

PR #5451, opened 2022-11-18
Restore worksheet name in filename source column. Fixes #5034

### LINKED ISSUE #5034: Excel Worksheet name not included with filename when "Store file source" option checked
When importing Excel worksheets previously (pre 3.5.0) if the "Store file source" option was checked both the filename and worksheet name were used to populate the "File" column. From 3.5.0 onwards only the filename is used.

The previous behaviour was the correct behaviour and should be restored.

### To Reproduce

1. Choose "Create Project"
2. Select "This computer"
3. Choose the Excel file linked to this issue
4. Once preview opened, check the "Store file source" option
5. Note that the "File" column added to the preview includes the Excel filename ("demo-excel-file.xlsx") but not the worksheet name ("Sheet 1")

### Current Results
The "File" column added to the preview includes the Excel filename but not the worksheet name

### Expected Behavior
The "File" column added to the preview should include the Excel filename and the worksheet name, joined by a '#' symbol - e.g.:
`demo-excel-file.xlsx#Sheet 1`

### Screenshots
**3.4.1 behaviour**
Excel import 3-4-1 behaviour

**3.6rc1 behaviour (also occurs in 3.5.x)**
Excel import 3-6-rc1 behaviour

### Datasets
Example Excel file
[demo-excel-file.xlsx](https://github.com/OpenRefine/OpenRefine/files/9062651/demo-excel-file.xlsx)

### Additional context
Reported on Google Group https://groups.google.com/g/openrefine/c/ooyERpE57Kg/m/E5tnz99uAgAJ

tfmorris:
Fixes #5034

This fixes the regression that was introduced in the previous refactoring. As part of this, I discovered that all of the importer tests ran at a level below that were the filename source (and archive name) was implemented, so none of it was ever tested.

This PR refactors both the importer framework and the tests, allowing me to verify that the fixe actually works (and providing test coverage for the future). I did not extended this testing to any other importers or test the archive name functionality.

I also fixed a couple of latent bugs discovered as part of this and did a general cleanup of things that my IDE was warning about.

This is a draft for now so that I can take a look at the refactoring with clear eyes and make sure it makes things better, not worse. I'm not really thrilled with the complexity of the importer architecture and don't want to be making it any worse.

### DIFF
--- a/main/src/com/google/refine/importers/FixedWidthImporter.java
+++ b/main/src/com/google/refine/importers/FixedWidthImporter.java
@@ -34,8 +34,8 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.LineNumberReader;
 import java.io.Reader;
-import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import com.fasterxml.jackson.databind.node.ArrayNode;
@@ -96,7 +96,7 @@ public class FixedWidthImporter extends TabularImportingParserBase {
         if (options.has("columnNames")) {
             String[] strings = JSONUtilities.getStringArray(options, "columnNames");
             if (strings.length > 0) {
-                retrievedColumnNames = new ArrayList<Object>();
+                retrievedColumnNames = new ArrayList<>();
                 for (String s : strings) {
                     s = CharMatcher.whitespace().trimFrom(s);
                     if (!s.isEmpty()) {
@@ -145,13 +145,13 @@ public class FixedWidthImporter extends TabularImportingParserBase {
      *            Line to be split
      * @param widths
      *            array of integers with field sizes
-     * @return
+     * @return an ArrayList of cells
      */
     static private ArrayList<Object> getCells(String line, int[] widths) {
-        ArrayList<Object> cells = new ArrayList<Object>();
+        ArrayList<Object> cells = new ArrayList<>();
 
         int columnStartCursor = 0;
-        int columnEndCursor = 0;
+        int columnEndCursor;
         for (int width : widths) {
             if (columnStartCursor >= line.length()) {
                 cells.add(null); // FIXME is adding a null cell (to represent no data) OK?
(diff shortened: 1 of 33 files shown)


main/src/com/google/refine/importers/FixedWidthImporter.java
```diff
@@ -217,7 +214,7 @@ static public int[] guessColumnWidths(File file, String encoding) {
                 }
 
                 if (counts != null && lineCount > 2) {
-                    List<Integer> widths = new ArrayList<Integer>();
+                    List<Integer> widths = new ArrayList<>();
```
wetneb, line 217: I'd be super grateful if we can avoid mixing large-scale style changes and functional changes.
I have no problem with normalizing `new ArrayList<Integer>()` to `new ArrayList<>()` but I would do this in a separate PR (potentially alongside other cosmetic changes). Ideally we should also change the linter configuration for it to enforce this convention across the codebase (if possible).

tfmorris: As a reminder, again, "Draft" means "not ready for review."

If you're looking for something to review, I marked #5397 as ready for review 3 days ago.

wetneb: I feel like it's useful to flag this sort of problem early on, before you bundle up more unrelated changes in the same PR, because they can be tricky to untangle later on. That is something that was a problem in past PRs, so it feels important to me to point this out.

If getting feedback at this stage is problematic for you, I wonder if you could just not open a PR yet - that would help us keep the backlog shorter. If the intention with opening a PR is to show that you are working on this to avoid someone else doing the same, you can just assign you to the issue, no? You could open the PR to run the CI, but in this case you are working on the backend so I trust you should be able to run the test suite locally without too much problem.

Another problem with draft PRs is that I do not actually get notifications when people mark them as ready for review - but I do get notifications when people open them. I wonder if this is something I can fix in my settings.

Also, I would be grateful if you could adopt a friendlier tone. Your comment above reads quite dry.

tfmorris: > I feel like it's useful to flag this sort of problem early on, before you bundle up more unrelated changes in the same PR, because they can be tricky to untangle later on. 

That's why I (usually) segregate them in separate commits. There are also unrelated bug fixes for problems found by my IDE, also separated in their own commits.

> That is something that was a problem in past PRs, 

If you're talking about my PRs, I don't remember it being mentioned. Sorry if I missed it.

> You could open the PR to run the CI, but in this case you are working on the backend so I trust you should be able to run the test suite locally without too much problem.

As you can see from 093c4c212638ede3869cbf13470c5fc38938f18f, CI produces [different results](https://github.com/OpenRefine/OpenRefine/actions/runs/3493620663) from the test suite. If there's something I'm missing in my local setup, I'd welcome a link to the documentation.

If it's project policy that Draft PRs aren't allowed, I will comply, of course, but it'll probably mean fewer contributions.

PR #5456, opened 2022-11-19
Fix signing of Jython dependency on MacOS

### LINKED ISSUE #4568: Dependency to Jython-standalone prevents us from signing our MacOS package
The fact that we depend on the `jython-standalone` distribution prevents us from signing our MacOS package and notarizing it with Apple.

### To Reproduce
Check the CI logs here:
https://github.com/OpenRefine/OpenRefine/runs/5335255475?check_suite_focus=true

### Current Results

When notarizing our DMG build, we get the following errors:
```
 [INFO]      [exec]   "issues": [
[INFO]      [exec]     {
[INFO]      [exec]       "severity": "error",
[INFO]      [exec]       "code": null,
[INFO]      [exec]       "path": "openrefine-mac-3.6-SNAPSHOT.dmg/OpenRefine.app/Contents/Resources/webapp/extensions/jython/module/MOD-INF/lib/jython-standalone-2.7.2.jar/jni/Darwin/libjffi-1.2.jnilib",
[INFO]      [exec]       "message": "The binary uses an SDK older than the 10.9 SDK.",
[INFO]      [exec]       "docUrl": null,
[INFO]      [exec]       "architecture": "i386"
[INFO]      [exec]     },
[INFO]      [exec]     {
[INFO]      [exec]       "severity": "error",
[INFO]      [exec]       "code": null,
[INFO]      [exec]       "path": "openrefine-mac-3.6-SNAPSHOT.dmg/OpenRefine.app/Contents/Resources/webapp/extensions/jython/module/MOD-INF/lib/jython-standalone-2.7.2.jar/jni/Darwin/libjffi-1.2.jnilib",
[INFO]      [exec]       "message": "The binary uses an SDK older than the 10.9 SDK.",
[INFO]      [exec]       "docUrl": null,
[INFO]      [exec]       "architecture": "x86_64"
[INFO]      [exec]     }
[INFO]      [exec]   ]
```

### Expected Behavior
Successful notarization.

### Additional context

We might be able to switch to `jython-slim` which should also let us save some megabytes (but some additional dependencies are probably pulled in then)

### Funding available

The OpenRefine project has funding available to work on this. Do you know someone who would be able to get this fixed? Send them our way!
https://openrefine.org/blog/2022/09/30/windows-macos-packaging.html

wetneb:
Fixes #4568.

This dynamically patches the Jython package at upload time, to replace a library by our own build of it.
This is a stop-gap measure to unblock the signing of the MacOS DMG package for the upcoming release.
I am hoping we can get rid of it soon once https://github.com/jnr/jffi/issues/123 is closed.

This will only be properly tested once the PR is merged in master since the packaging pipeline requires secrets which are not made available to pull requests for security reasons.

It is very well possible that Apple notarization still fails with this, because their tooling reports errors as they are encountered (meaning that solving this particular error might uncover others). In this case, I will disable notarization again and investigate the following problem, to preserve the greenness of the CI.


wetneb:
Merging to see if it works - happy to revert if people have concerns with this approach!

PR #5467, opened 2022-11-23
Fix unary minus sign evaluation. Fixes #5465

### LINKED ISSUE #5465: Negative integer literals (w/unary minus sign) incorrectly interpreted as floats
If you have an integer-only expression you expect the result to remain integer, but here are a couple of instances where using a negative integer can result in a floating point result (note that the only way I know how to test for the resultant type is to use toString):

1. If the first (or only) value of the expression is a negative integer, as in
`toString(-1+2,"%d")`

2. If a variable in a with() or forEach() function is a negative integer, as in
`with(-1,a,toString(2+a,"%d"))`

In both of these cases the result is
_Error: java.util.IllegalFormatConversionException: d != java.lang.Double_
but if you use "%f" in the string conversion you get 1.000000 .

I am using OpenRefine Version 3.6.2 in Safari Version 16.1 on MacOS 12.6.1 .

tfmorris:
Fixes #5465

- Fix unary minus sign evaluation and add tests for that case
- 
-


wetneb (approved):
Looks like a very safe move indeed! Thanks for looking into that.

PR #5468, opened 2022-11-23
Import unsupported datatypes as strings

### LINKED ISSUE #5389: All numbers are imported as floats by XLS(X) importer
### To Reproduce
1. Export a table containing an integer column

### Current Results
A cell showing the integer `123` is exported as `123.0`. This makes SQL exports fail with the following error: `MESSAGE:	123.0 is not compatible with column type :INT`.

### Expected Behavior
What you see is what you get: the software should not export cells in a different format.

### Versions
 - Operating System: Linux
 - Browser Version: Firefox
 - OpenRefine: 3.6.2

### LINKED ISSUE #5390: Unsupported datatypes should be imported as strings
There are a large number of Excel date, time, duration, post code, etc formats which are not supported by OpenRefine, but are currently imported as either numbers or datetimes. Anything unsupported should be imported as a string instead of a number or datetime.

### To Reproduce
1. Import an Excel worksheet containing dates such as `09:46:00 AM` in a column
2. Open the project

### Current Results
The cell is converted to `1899-12-31T09:46:00Z`.

### Expected Behavior
The column should be kept as a string.

### Additional context
I have not found any way to disable cast on import.

tfmorris:
Fixes #5390. Fixes #5389

- Import numbers as integers where possible/required. This includes "General" format numbers with a zero fractional part as well numbers which are explicitly formatted as integers, perhaps masking a fractional part.
- Import unsupported date formats as strings, including dates (w/o times), times, intervals
- Import problematic non-date number formats as strings, including leading zeros, embedded punctuation or anything other than grouping separator, decimal separator, percentage sign, padding (leading/trailing space & trailing zero)

The regex used to identify the number formats that OpenRefine can import cleanly will probably need some tuning. There's also the issue that Apache POI's text rendering isn't feature complete or 100% true to the Excel implementation, so there will be edge cases where there's information loss/corruption, but the hope is that it'll be better fidelity than the current implementation.

We should solicit Excel power users to do significant testing of this before release.

### DIFF
--- a/main/src/com/google/refine/expr/functions/ToDate.java
+++ b/main/src/com/google/refine/expr/functions/ToDate.java
@@ -146,7 +146,7 @@ public class ToDate implements Function {
                     }
                 }
             } catch (IllformedLocaleException e) {
-                // We ignore this. It PROBABLY means we got a date format string, not a language code
+                // We ignore this. It PROBABLY means we got a date format string, not a language code,
                 // although it could be a malformed language tag like zh_TW instead of zh-TW
             }
         }
--- a/main/src/com/google/refine/importers/ExcelImporter.java
+++ b/main/src/com/google/refine/importers/ExcelImporter.java
@@ -1,6 +1,6 @@
 /*
 
-Copyright 2010, Google Inc.
+Copyright 2010, 2022 Google Inc., OpenRefine developers
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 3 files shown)


wetneb (approved):
I tested it manually and it works.

One thing I noticed though: if numbers are formatted as currencies, they are now imported as strings, whereas previously they were imported as numbers. I am not sure what is best, really: by importing them as numbers, we were throwing the currency away (which is obviously a big problem for datasets with mixed currencies), but I can also imagine that some users relied on that and will struggle to convert the string to a number on their own (since they need to strip the unit using some string transformation operation first). It feels like something that could be controlled by a new option in the importer, but I guess we can open a separate issue about it and merge this as is.

Can I leave the merge conflict to you?

tfmorris:
> One thing I noticed though: if numbers are formatted as currencies, they are now imported as strings, whereas previously they were imported as numbers.

That's not the intention. Currencies, as well as numbers with grouping separators (e.g. $1,000.00), should be imported as numbers. Let me add a test case for that.

wetneb:
Interesting! On my side, importing the following file keeps the last cell as a string:
[excel_datatypes_test.zip](https://github.com/OpenRefine/OpenRefine/files/10135566/excel_datatypes_test.zip)

tfmorris:
The format for that cell is `[$K-647]#,##0.00;[RED]\-[$K-647]#,##0.00`. No multi-part formats (ie positive;negative;zero;text) are currently recognized and none of the fancy conditional formatting using square brackets is recognized. I'm guessing you selected a currency symbol other than your locale's default.

Previously all numbers were imported as floating point numbers. Now things are biased the other way so that anything other than simple numeric formats are imported as strings (and they'll be strings as rendered by Apache POI which might not be 100% faithful to Excel's rendering).

If anyone wants to go to town tweaking the regex that's used for numeric recognition, feel free, but the Excel format strings can be exceedingly complex.

tfmorris:
I've added support for scientific notation (which was an oversight), as well as all the built-in numeric formats. I fully expect additional tweaking will need to be done during user testing which is why I recommended that it received significant testing from Excel power users.

wetneb:
Nice, thanks! I cannot really tell to what extent the coverage of Excel's datatypes is complete enough, not knowing much about Excel's data model, but this looks good to merge on my side. We can tweak the logic later on if needed, as users test the feature. The upcoming 3.7-beta1 release would feel like a good opportunity for that, but if you feel it's too risky we can also merge it afterwards.

tfmorris:
I think it's low risk. If people run into problems, it should be obvious at import time and they can work around it by changing the cell format to get the old behavior.

PR #5470, opened 2022-11-23
Make Excel import tests locale independent. Fixes #5464

### LINKED ISSUE #5464: XLS tests are locale-dependent
The tests for XLS importer can fail depending on the locale they are run with.

### To Reproduce
Steps to reproduce the behavior: run `mvn test` in the `fr.FR` locale (unfortunately there does not seem to be a CLI argument to set that).

### Current Results

```
[ERROR] readXlsxAsText(com.google.refine.importers.ExcelImporterTests)  Time elapsed: 0.1 s  <<< FAILURE!
java.lang.AssertionError: expected [1.1] but found [1,1]
	at com.google.refine.importers.ExcelImporterTests.readXlsxAsText(ExcelImporterTests.java:224)
```

### Expected Behavior
The tests should pass.

### Versions
 - OpenRefine: master branch

### Additional context
Intuitively the locale dependency of OpenRefine's behaviour is expected here, since we are just preserving the string representation of values in the XLS file, which are locale dependent, so I would say we should just ensure the tests accept that.

It could be even better to give the user the option to select the locale in which the values should be represented, but I am not even sure POI lets us configure that.

tfmorris:
Fixes #5464

- Use our default locale DecimalFormatter instead of a hardcoded string for the expected value
- ~~Also left in the commented out "big hammer" version which just pins the default locale to US. I don't have a strong opinion as to which is better.~~
-

### DIFF
--- a/main/tests/server/src/com/google/refine/importers/ExcelImporterTests.java
+++ b/main/tests/server/src/com/google/refine/importers/ExcelImporterTests.java
@@ -44,6 +44,8 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
 import java.time.LocalDate;
 import java.time.Month;
 import java.time.OffsetDateTime;
@@ -85,6 +87,7 @@ public class ExcelImporterTests extends ImporterTest {
 
     private static final File xlsFileWithMultiSheets = createSheetsWithDifferentColumns(false);
     private static final File xlsxFileWithMultiSheets = createSheetsWithDifferentColumns(true);
+    private static final NumberFormat NUMBER_FORMAT = DecimalFormat.getInstance();
 
     @Override
     @BeforeTest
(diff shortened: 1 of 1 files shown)


main/tests/server/src/com/google/refine/importers/ExcelImporterTests.java
```diff
@@ -85,11 +88,13 @@ public class ExcelImporterTests extends ImporterTest {
 
     private static final File xlsFileWithMultiSheets = createSheetsWithDifferentColumns(false);
     private static final File xlsxFileWithMultiSheets = createSheetsWithDifferentColumns(true);
+    private static final NumberFormat NUMBER_FORMAT = DecimalFormat.getInstance();
 
     @Override
     @BeforeTest
     public void init() {
         logger = LoggerFactory.getLogger(this.getClass());
+//        Locale.setDefault(Locale.US); // Some Excel formats are locale dependent, so pin our locale to en_US
```
wetneb, line 97: I think your use of `DecimalFormat` is better, so I would drop this comment. If we were setting the locale manually like this, I think we should also restore it at the end of the test.
Alternatively if we leave the comment in, I would add a comment to explain why it is commented out - otherwise I would find it confusing (especially because it's not necessarily obvious that the NumberFormat above is the solution to the locale dependency).

tfmorris: Done

wetneb (approved):

PR #5474, opened 2022-11-25
Bump Cypress 9.7.0 to 11.2.0

elebitzero:
Cypress has restored the 'Run all specs' and 'Run filtered specs' features as of version 11.2.0 (Ref: https://github.com/cypress-io/cypress/discussions/21628), so we can go ahead and upgrade now.

- Bumped Cypress from 9.7.0 to 11.2.0 in main\tests\cypress\package.json
- The upgrade involves some migration of files.  The migration was automated by the Cypress runner. See screenshots below.
- Re-enabled Dependabot for the Cypress package.
- Added `experimentalRunAllSpecs: true` in cypress.config.js to enable 'Run all specs' and 'Run filtered specs'
- I ran the full test suite five times, and all of the tests appear to work as before.

image

image


wetneb (approved):
Thank you so much for taking care of this! Looks like the CI isn't green yet but I feel like we are close :)

Issue #3057, opened 2020-08-11
Configuration for OpenRefine on Windows should use only 1 config file (.ini)


thadguidry:
**Is your feature request related to a problem or area of OpenRefine? Please describe.**
On Windows, OpenRefine has 2 .ini files that can be used for applying JVM settings.  We should only have 1 .ini file.

**Describe the solution you'd like**
Allow the `openrefine.exe` launching to read settings from the refine.ini file only.
Remove the `openrefine.l4j.ini` file from packaging.

**Describe alternatives you've considered**

**Additional context**

wetneb:
I had a quick look to estimate the difficulty of this task and put a reasonable price tag on it.

The `openrefine.l4j.ini` and `refine.ini` do not use the same syntaxes, so one cannot directly point either `refine.bat` to `openrefine.l4j.ini` or `openrefine.exe` to `refine.ini`.

The name `openrefine.l4j.ini` is hardcoded by Launch4j into `openrefine.exe` so this filename does not seem configurable, see:
https://launch4j.sourceforge.net/docs.html#Additional_jvm_options

Therefore the simplest solution for this would probably be to adapt `refine.bat` so that it uses `openrefine.l4j.ini`, changing it so that it accepts the syntax currently used there. This would be slightly odd in the sense that `refine.bat` would rely on a file with `l4j` in its name while Lanch4j is not used at all by `refine.bat`, but I would say this would still be an improvement over the current situation and still solve this issue.

It should not be too hard to submit a patch to Launch4j to make this filename configurable, but it will probably take a while before this is merged and released, so that could be a follow-up issue on this.

wetneb:
Just for fun I tried to make Launch4j configurable: https://sourceforge.net/p/launch4j/git/merge-requests/20/.
But I still think the easiest solution would be to adapt `refine.bat`.

wetneb:
Good news, my PR in Launch4j was merged so this filename should be configurable in the next Launch4j release. After that it should be possible for Windows and Linux to use the same configuration file. I am not sure about Mac OS though - for now the options seem to be loaded from some `info.plist` according to the docs, and it looks not so easy to change that.

wetneb:
Update: this is still blocked by necessary changes in Launch4j. The following PR needs merging and releasing: https://sourceforge.net/p/launch4j/git/merge-requests/21/

PR #5477, opened 2022-11-27
Handle proxy configuration, closes #5476

### LINKED ISSUE #5476: Proxy configuration still not working
In OpenRefine 3.6.2, the configuration of proxy is half working after the implementation of #3919.

Indeed, the nonProxyHosts parameter is no taking into account which prevents the use case where some services are outside (access through the proxy) and others are inside the institution (access without the proxy).

The following configuration in openrefine.l4j.ini:
```
-Dhttp.proxyHost=myproxy.mydomain.org
-Dhttp.proxyPort=8080
-Dhttp.nonProxyHosts="*.mydomain.org|localhost|127.0.0.1"
```
should use the proxy ``myproxy.mydomain.org`` for all the hosts except the ones in ``*.mydomain.org``

currently the proxy isused for all the hosts

tledoux:
Fixes #5476

This pull request aims to properly handle the configuration of proxies for external services called by OpenRefine.
In particular, it allows the scenario where some services are outside (accessible through the proxy) and other are inside (not using the proxy).

3 configurations are possible:
1/ no proxy is needed: nothing has to be defined in openrefine.l4j.ini

2/ we want to use the usual properties for configuring nework in java as described in the [documentation](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html)
In this case, the property ``-Djava.net.useSystemProxies=true`` should be defined as well as ALL the others needed (the latter ones can be defined globally using the environment variable ``JAVA_TOOL_OPTIONS``). Note that, in particular, http and https properties need too be defined to handle both protocols.

3/ we just want a specific configuration for OpenRefine. In which case, only the ``http.proxyHost``, ``http.proxyPort`` and ``http.nonProxyHosts`` properties need to be defined, in the openrefine.l4j.ini configuration file, like this:
```
-Dhttp.proxyHost=myproxy.mydomain.org
-Dhttp.proxyPort=8080
-Dhttp.nonProxyHosts="*.mydomain.org|localhost|127.*"
```
The configuration will wok for both, http and https, protocols.

All the above should probably be described in the documentation but it seems to have move from docs directory...

Lastly, I added log information in order to inform about the configuration used during download of files or reconciliation.

### DIFF
--- a/conf/openrefine.l4j.ini
+++ b/conf/openrefine.l4j.ini
@@ -8,6 +8,16 @@
 
 # Use system defined HTTP proxies
 -Djava.net.useSystemProxies=true
+# You should probably defined the properties below, if they are not already 
+# defined in the environment variable JAVA_TOOL_OPTIONS.
+# Refer to the Java documentation (for example, at
+# https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html
+# for the precise meaning of each option
+#-Dhttp.proxyHost=myproxy.mydomain.org
+#-Dhttp.proxyPort=8080
+#-Dhttps.proxyHost=myproxy.mydomain.org
+#-Dhttps.proxyPort=8080
+#-Dhttp.nonProxyHosts="*.mydomain.org|localhost|127.*"
 
 #-XX:+UseLargePages
 #-Dsomevar="%SOMEVAR%"
--- a/conf/openrefine.l4j.ini
+++ b/conf/openrefine.l4j.ini
@@ -8,11 +8,11 @@
 
 # Use system defined HTTP proxies
 -Djava.net.useSystemProxies=true
-# You should probably defined the properties below, if they are not already 
+# You should probably define the properties below, if they are not already 
 # defined in the environment variable JAVA_TOOL_OPTIONS.
 # Refer to the Java documentation (for example, at
 # https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html
-# for the precise meaning of each option
+# for the precise meaning of each option)
 #-Dhttp.proxyHost=myproxy.mydomain.org
 #-Dhttp.proxyPort=8080
 #-Dhttps.proxyHost=myproxy.mydomain.org
--- a/main/src/com/google/refine/util/HttpClient.java
+++ b/main/src/com/google/refine/util/HttpClient.java
@@ -35,6 +35,8 @@ import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
+import java.util.regex.Matcher;
 
 import org.apache.hc.client5.http.ClientProtocolException;
 import org.apache.hc.client5.http.classic.methods.HttpGet;
@@ -83,6 +85,7 @@ public class HttpClient {
     private HttpHost proxy;
     private int proxyPort;
     private String proxyHost;
+    private Pattern nonProxyHosts;
     private DefaultProxyRoutePlanner routePlanner;
 
     public HttpClient() {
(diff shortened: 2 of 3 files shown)


conf/openrefine.l4j.ini
```diff
@@ -8,6 +8,16 @@
 
 # Use system defined HTTP proxies
 -Djava.net.useSystemProxies=true
+# You should probably defined the properties below, if they are not already
```
wetneb, line 11:
```suggestion
# You should probably define the properties below, if they are not already 
```

conf/openrefine.l4j.ini
```diff
@@ -8,6 +8,16 @@
 
 # Use system defined HTTP proxies
 -Djava.net.useSystemProxies=true
+# You should probably defined the properties below, if they are not already 
+# defined in the environment variable JAVA_TOOL_OPTIONS.
+# Refer to the Java documentation (for example, at
+# https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html
+# for the precise meaning of each option
```
wetneb, line 15:
```suggestion
# for the precise meaning of each option)
```

wetneb (commented):
Sorry for the delay in reviewing this. It looks very good to me, thank you! I will try to test this manually too, but apart from small typos this seems very good to me.

conf/openrefine.l4j.ini
tledoux: Thanks for the review,  the last commit should include the typos.

conf/openrefine.l4j.ini
```diff
@@ -8,6 +8,16 @@
...
+# You should probably define the properties below, if they are not already 
+# defined in the environment variable JAVA_TOOL_OPTIONS.
+# Refer to the Java documentation (for example, at
+# https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html
+# for the precise meaning of each option)
+#-Dhttp.proxyHost=myproxy.mydomain.org
+#-Dhttp.proxyPort=8080
+#-Dhttps.proxyHost=myproxy.mydomain.org
+#-Dhttps.proxyPort=8080
+#-Dhttp.nonProxyHosts="*.mydomain.org|localhost|127.*"
```
wetneb, line 20: It would also be worth adding this to `refine.ini` (for Linux) using the `JAVA_OPTIONS=-Dhttp.proxy....` syntax documented at https://openrefine.org/docs/manual/running#starting-with-modifications
It's really a mess though, we ought to have something uniform across platforms… 

If that feels too fiddly to add to all platforms now, I would be okay with waiting for https://github.com/OpenRefine/OpenRefine/issues/3057 to sort it out.

wetneb (approved):
Thanks a lot again! I wonder if it would actually make sense to turn on the `java.net.useSystemProxies` variable by default?

I can imagine this would make the life of a lot users easier, by not having to figure out how to configure OpenRefine to respect that. This would also have the benefit of treating HTTP calls made from the frontend and the backend in a uniform way, since web browsers generally default to the system's proxy configuration. For reconciliation services (where some calls are made by the frontend, others by the backend) this could be really helpful.

ostephens:
> Thanks a lot again! I wonder if it would actually make sense to turn on the `java.net.useSystemProxies` variable by default?

I thought this was something agreed long long ago! https://code.google.com/archive/p/google-refine/issues/325

tledoux:
Hi, I have no problem with turn on the `java.net.useSystemProxies` variable by default. My only concern is that the system (for this variable) is the Java system and not the platform system, which is quite misleading.
For example, on Windows, the proxy internet settings in the "Internet config panel" is NOT taken into account. However, if you add the `http.proxyHost` variable across all the java instances (for example, by adding it to the environment variable JAVA_TOOL_OPTIONS) then it will work.
In normal configuration, such settings are not made, so you do have to change the `openrefine.l4j.ini` file to set your own proxy.
Since, as soon as you have a proxy you need to edit the ini file, I prefer to have nothing set (to avoid suprises) and have all the possibities documented and commented to make changes easy.
But again if you feel other defaults are better, feel free to change them.

wetneb:
Thanks for the precisions. It's indeed sad that it does not pick up the operating system's proxy settings.
I can imagine that in some corporate environments, the Java proxy settings are also preset to the same proxy settings, so that turning this on would still bring some benefit to those users.

PR #5480, opened 2022-11-30
Cleaned up flaky test that used ajax_in_progress="true" in join_columns.cy

elebitzero:
-  Fixes a flaky Cypress test in join_columns.spec that used `ajax_in_progress="true"`


wetneb:
Looks like this change might have been already merged before?

elebitzero:
This PR has a new change in join_columns.cy, which is similar to the change in #5461 that had an `ajax_in_progress="true"` in rows_records.cy.js.

wetneb (approved):

Issue #528, opened 2012-10-15
Entire Refine workspace metadata trashed unrecoverably


tfmorris:
_Original author: tfmorris (February 09, 2012 01:43:40)_

All of the projects in my Refine workspace directory have been renamed from .project to .project.corrupted and each contains a 0-byte metadata.json file.

I'm not sure exactly how this happened, but I suspect that Refine tried to shutdown when it was in a low-memory situation and ended up writing the 0-length files.  Although it uses a write, rename, delete strategy for updates, this only checks for the existence of the output file, not whether or not it contains valid data.

I think a few things need to be done differently:
- don't update project metadata which hasn't been modified
- don't delete the old metadata json file until the next time Refine has been started and successfully read in the newly written file (so that there's always a backup copy)
- probably other stuff after a little more code review

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=528_

PR #5490, opened 2022-12-02
Untangled project timestamp management. Fixes #3047

### LINKED ISSUE #3047: Fix project metadata dates
In #1383 the project creation and modification dates were changed to the `LocalDateTime` datatype, but this can't represent a point in time (because it has no time zone info), so is completely inappropriate for this use.

This was actually one of the few areas of OpenRefine where `java.util.Date` was perfectly appropriate and suited to the use, so it didn't need to be change. ~~All projects which were written since this change are going to have dates which are only knowable within a 24 hour window, not precisely.~~

We can go back to `java.util.Date` or use fancy new `OffsetDateTime` or `Instant` but it should get serialized as ISO 8601 at UTC and formatted using the user's locale & timezone.

tfmorris:
Fixes #3047

- Switch from LocalDateTime with ad hoc UTC conversion hand coded on top to java.time.Instant which is natively UTC
- Fix test that was broken in 2017 when the previous set of changes were made

NOTE: This changes the datatypes of parameters/returns in an API which is declared as public, but really shouldn't be used externally (and no one apparently noticed/complained when they were changed from java.util.Date before). We could restore compatibility with the pre-2017 API, but I don't really see the point.

Untangling the user data type will be in a separate PR since it's not at all coupled with this.

### DIFF
--- a/main/src/com/google/refine/ProjectManager.java
+++ b/main/src/com/google/refine/ProjectManager.java
@@ -1,6 +1,6 @@
 /*
 
-Copyright 2010, Google Inc.
+Copyright 2010, 2022 Google Inc. & OpenRefine contributors
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -36,11 +36,9 @@ package com.google.refine;
 import com.google.refine.util.GetProjectIDException;
 import java.io.IOException;
 import java.io.InputStream;
-import java.time.LocalDateTime;
-import java.time.ZoneId;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


wetneb:
Thanks Tom, yes this migration has caused quite some pain and sorrow on my side at least. Glad to see a principled approach to the problem. I'll need a bit more time to wrap my head around the changes. Does this preserve all serialization formats?

wetneb (approved):
Answering my own question: yes, the JSON serialization of `ProjectMetadata` is the only one that was touched, and it is covered by untouched tests, so we should be all good. All the changes make sense to me just reading the diff.

tfmorris:
@wetneb Correct, the serialization format is unchanged. The test changes are described in the commit message for this commit: https://github.com/OpenRefine/OpenRefine/pull/5490/commits/661ca8401965f900c7caa992bf6af2db81e6bc32

@thadguidry Because the on-disk formats are unchanged, recover should work the same way it always has. I haven't really looked at that code since I wrote in back in 2013 (https://github.com/OpenRefine/OpenRefine/commit/1d8784e059acc3f249a551df1648b643dad61cbd #528), but I don't see any reason that it would be affected by this.

PR #5496, opened 2022-12-05
Changed "parse-next" to "parse-first"

josephkois:
64 changes staged for testing, changed "parse-next" to "parse-first". CTRL+SHIFT+F to search for all keywords containing "parse-next".

All Language files have been modified, however the translations have not been modified due to the lack of trust in google translate.

Fixes #{issue number here}

- changed every translation file from "parse-next" to "parse-first"
- changed translation .json files in main/webapp to also reflect this change
-

### DIFF
--- a/extensions/database/module/langs/translation-ca.json
+++ b/extensions/database/module/langs/translation-ca.json
@@ -46,7 +46,7 @@
     "database-parsing/disable-auto-preview": "Desactivar previsualització automàtica",
     "database-parsing/ignore-first": "Ignorar primera",
     "database-parsing/ignore": "línia/es al començament de l'arxiu",
-    "database-parsing/parse-next": "Analitza la següent",
+    "database-parsing/parse-first": "Analitza la següent",
     "database-parsing/parse": "línia/es com a capçaleres de columna",
     "database-parsing/limit-next": "Carrega com a màxim",
     "database-parsing/limit": "fila/es de dades",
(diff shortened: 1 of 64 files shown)


wetneb:
Hi @josephkois, thanks for this PR! Did you try to address a problem which is described by one of our issues? If so, you could insert the issue number in the description of your pull request (at `Fixes #{issue number here}`).
If not, could you describe why this change is needed?

Generally, pull requests made here should only change the English translation files: the translators can then take care of updating the translations accordingly. Your change seems to reuse existing translations for a different English source, which is generally not something we should do: if you change the English source, then it is likely the translations need to be changed as well, and as such I suspect that Weblate will not mark those strings as needing new translation.

josephkois:
> Hi @josephkois, thanks for this PR! Did you try to address a problem which is described by one of our issues? If so, you could insert the issue number in the description of your pull request (at `Fixes #{issue number here}`). If not, could you describe why this change is needed?
> 
> Generally, pull requests made here should only change the English translation files: the translators can then take care of updating the translations accordingly. Your change seems to reuse existing translations for a different English source, which is generally not something we should do: if you change the English source, then it is likely the translations need to be changed as well, and as such I suspect that Weblate will not mark those strings as needing new translation.

Hey sorry about that, I thought I did include the issue #.

 I ended up changing all the parse-next, however did not change the translation. I will try another pull request and will only update the English translation, I will also fix the title and description of the pull request to align with the standard.

Sorry, first time contributing to open source.

PR #5509, opened 2022-12-08
Add `/Applications` shortcut to macOS dmg file

### LINKED ISSUE #5205: Provide Applications shortcut in MacOS dmg + potentially add a nice background [$250 bounty]
Our MacOS download provides a .dmg that, upon opening, only contains the OpenRefine application, without any other frills. It works, but it's barebones :-)

I have just downloaded quite a bit of software, and I notice that many applications make their .dmg a bit fancier these days. They usually include a shortcut to the user's Applications folder, sometimes with a background image with some context/help that makes it easier and more obvious to the user that they can just drag the application there in order to install the software.

Some applications also have backgrounds in that .dmg window, which makes everything look just a bit more polished, professional and friendly.

LibreOffice: In my opinion: good to have that arrow + the project logo and a visual. The arrow looks amateurish though. Not sure if the licenses and readmes folders make any sense - I don't think anyone will ever click/open these?
image

Zotero: Simple but good, maybe a bit amateurish because of its simplicity.
image (1)

VLC: Something in between. Balanced and subtle. Not bad IMO!
image

### Proposed solution
Can we do this too?
* Highest priority IMO: the shortcut to the Applications folder
* Nice to have but not necessary: a background with more context. I'd like that background to be nice and tasteful ;-)
* We can talk about: what else makes sense to include in the window without making it too bloated? Perhaps we can even include something like a donation link? ;-) (not sure if that would be possible)

### Alternatives considered
We can also totally leave it like it is - it works. But I think it's a nice touch in general.

### Funding available

The OpenRefine project has funding available to work on this. Do you know someone who would be able to get this fixed? Send them our way!
https://openrefine.org/blog/2022/09/30/windows-macos-packaging.html

messense:
Fixes #5205

- Use [dmgbuild](https://github.com/dmgbuild/dmgbuild) to create macOS dmg file with a shortcut to `/Applications` and a simple arrow background (can be changed to something else if you design a better background image).

image

### DIFF
--- a/.github/workflows/snapshot_release.yml
+++ b/.github/workflows/snapshot_release.yml
@@ -206,6 +206,14 @@ jobs:
         gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
         gpg-passphrase: GPG_PASSPHRASE
 
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3' 
+
+    - name: Install dmgbuild
+      run: pip install dmgbuild
+
     - name: Decrypt Apple certificates
       run: |
         gpg --pinentry-mode loopback --passphrase "${{ secrets.GPG_PASSPHRASE }}" --output .github/workflows/release/apple_cert.cer --decrypt packaging/apple_certs/apple_cert.cer.enc
--- a/packaging/dmg_settings.py
+++ b/packaging/dmg_settings.py
@@ -1,0 +1,181 @@
+import os.path
+import plistlib
+
+#
+# Example settings file for dmgbuild
+#
+
+# Use like this: dmgbuild -s settings.py "Test Volume" test.dmg
+
+# You can actually use this file for your own application (not just TextEdit)
+# by doing e.g.
+#
+#   dmgbuild -s settings.py -D app=/path/to/My.app "My Application" MyApp.dmg
+
+# .. Useful stuff ..............................................................
+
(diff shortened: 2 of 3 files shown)


wetneb:
Thanks a lot for tackling this @messense!

I tried building the DMG with your new pipeline on Linux and was not sure if it would be supported. It did generate a DMG file, but the internal structure seems to be wrong (when opening the DMG in MacOS VM):
image

Since I am not getting the same result as you, I suspect this is just a consequence of the fact that I built on Linux?
In the CI, we are building with MacOS, so I would be tempted to just merge this and see if the DMG built by the CI works as expected.

messense:
> I suspect this is just a consequence of the fact that I built on Linux?

I've only tested on arm64 macOS, looking at the CI configuration of dmgbuild they also only test on macOS: https://github.com/dmgbuild/dmgbuild/blob/main/.github/workflows/ci.yml

messense:
https://github.com/dmgbuild/mac_alias

> While it is written in pure Python, some OS X specific code is required to generate a proper Alias or Bookmark record for a given file, so this module currently is not portable to other platforms.

I think because `dmgbuild` depends on `mac_alias`, it's also not portable to non-macOS platforms..

wetneb (approved):
Alright then let's just merge this and I'll check the output from the CI. Exciting!

wetneb:
It works as expected. Thanks again! I have awarded the corresponding bounty to you on IssueHunt.

PR #5511, opened 2022-12-08
issue 5359-- updated refine.bat

### LINKED ISSUE #5359: Windows refine.bat needs to be updated for new arguments matching Linux refine script
Windows refine script arguments `/d` and `/x` no longer match the Linux refine script and causes confusion for those that maintain across both OS's.

We introduced a few new arguments in the Linux refine script over the last couple of years

- `-d <path> path to the data directory`
- `-x <name=value> additional configuration parameters to pass to OpenRefine`
- `-v <level> verbosity level [from low to high: error,warn,info,debug,trace]`

and replaced two with different arguments (which should no longer are `/d` and `/x` as above but instead mimic the Linux refine script as

- `--debug enable JVM debugging (on port 8000)`
- `--jmx enable JMX monitoring (for jconsole and jvisualvm)`

### To Reproduce
1. Start OpenRefine from Cmd Prompt with `refine /d "D:\myOpenRefineWorkspace\"`
2. Notice that log output does not show the data directory being set but instead JVM debugging is enabled

### Current Results
`refine.bat /?` is not consistent with the output from Linux with `./refine -h`

### Expected Behavior
both startup scripts should provide identical arguments, except for differences in shell syntax.
Additionally, `/debug` and `/jmx` should be the corresponding argument syntax in `refine.bat`.
We should also update our user documentation pages where necessary for Windows to reflect the updated arguments.

### Versions
 - JRE or JDK Version: JDK 11
 - OpenRefine:  Version 3.7-SNAPSHOT [ea70840]

josephkois:
Updated the refine.bat file to replicate the Linux refine counterpart.

changed /d, /x, and /v commands.

Fixes #5359

- Updated /d command in refine.bat
- Updated /x command in refine.bat
- Updated /v command in refine.bat


elebitzero (approved):

elebitzero:
The current `/d` command has to be moved to the suggested `/debug` and `/x` has to be moved to `/jmx`, and `/v` has to be implemented as well.

josephkois:
> Items to complete for refine.bat:
> 
>     * Move current `/d` functionality to the suggested `/debug`.
> 
>     * Implement new `/d` command that sets the data directory.
> 
>     * Move `/x` functionality to the suggested `/jmx`.
> 
>     * Implement new `/x` command that provides additional parameters to pass to OR.
> 
>     * Implement new `/v` command that sets the verbosity level.

I think i fixed it now? I changed --debug to /debug and --jmx to /jmx

elebitzero (commented):
Hi @josephkois, I still don't see the `/d`, `/x`, and `/v` commands implemented in refine.bat so they have the same functionality as the Linux refine scripts.

josephkois:
> Hi @josephkois, I still don't see the `/d`, `/x`, and `/v` commands implemented in refine.bat so they have the same functionality as the Linux refine scripts.

Will go over it tomorrow morning and try to fix it, sorry! I see my mistake now.

PR #5514, opened 2022-12-12
Adds a fix to fit the text facet windows to their content

### LINKED ISSUE #5513: Fit height of text facet to its contents when few options are present
When creating a text facet in a column where few different values are present, I get a facet with a lot of whitespace:

image

### Proposed solution
When few options are present, the facet should shrink to a smaller size, without introducing a scroll bar:
image

When more options are available, it could potentially expand more than its default size, but that should not happen indefinitely: we should introduce a scroll bar past a certain length.

The facets should remain resizable by the user in any case.

### Alternatives considered

None.

### Additional context

Prompted by @ostephens's video on [doing the Advent of Code 2022 with OpenRefine](http://www.meanboyfriend.com/overdue_ideas/2022/12/advent-of-code-with-openrefine/)

elebitzero:
Fixes #5513

  * Sets the initial height of a text facet to the lesser of the item list height or 17em.  
  
 image


wetneb:

Is the `overflow: auto` really needed? It seems to add a spurious scroll bar in my case (both Chromium and Firefox on Linux):

With the initial height of the facet (without any resizing), the scroll bar is present: intuitively it should not, since the width is fitted to the contents:
image

If I compress the facet under its initial height, I get two nested scroll bars:
image

Without the `overflow: auto`, those scroll bars disappear and the text facet still fits to its contents:
image

The inner scroll bar still appears when compressing the facet:
image

elebitzero:
Thanks for reviewing, I was able to get it to work without the `overflow:auto`.

wetneb:
With this new version, it is impossible on my side to expand the facet beyond the set `max-width`. I think it would be a useful thing to preserve. Why do you need to introduce the `max-width` statement on the `facet-body-inner`? It seems to work great without, no?

elebitzero:
@wetneb, I had to rework this to allow the facet to be resizable since we can't use max-height. Instead, I set the initial height when the facet element is added to the document.  

Please try these changes out. I think this fixes the issue #5513 while still preserving the previous functionality.

wetneb (approved):

wetneb:
Works as expected on my side, thanks a lot!

PR #5543, opened 2023-01-03
Clear selected type when "No Type" is selected. Fixes #5523

### LINKED ISSUE #5523: Suggest property query includes a type when none is selected
When reconciling a column, using a service that supports a `/suggest/property` endpoint, if I select `Reconcile against no particular type`, OpenRefine includes the most recently selected type in the query.

For example, in this screenshot, I selected `Reconcile against no particular type` and then started typing `test` in the `As property` box.

Screen Shot 2022-12-14 at 4 40 06 PM

### Current Behavior
In that example, OpenRefine sent the following request to the reconciliation service:

`https://wikidata.reconci.link/en/suggest/property?prefix=test&spell=always&exact=false&scoring=schema&prefixed=true&type=Q223393`

### Expected Behavior
I would expect the type to be sent in the suggest/property query if one is selected so that a reconciliation service has the option to filter the suggested properties by the included type.

If `Reconcile against no particular type` is selected, then I expect no type to be included in the query so that the reconciliation service does no filtering on the suggested properties.

### Versions
 - Operating System: MacOS Monterey
 - Browser Version: Chrome
 - JRE or JDK Version: JDK 13.0.2
 - OpenRefine: Version 3.6.2

### Additional context
I'm creating a reconciliation service and this affects how we need to implement the internal logic of the suggest property endpoint.

tfmorris:
Also align naming to be consistent.

Fixes #5523

- Clear selected type when "Reconcile against any type" is selected
- Align naming to be consistent

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
+++ b/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
@@ -19,8 +19,8 @@
         <label for="againstTypeId"><span bind="or_proc_againstType"></span></label> <input aria-label="" size="20" bind="typeInput" /></td>
     </tr>
     <tr>
-      <td colspan="2"><input type="radio" name="type-choice" id="noType" value="-">
-        <label for="noType"><span bind="or_proc_noType"></span></label></td>
+      <td colspan="2"><input type="radio" name="type-choice" id="noTypeId" bind="noType" value="-">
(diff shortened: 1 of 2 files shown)


elebitzero (approved):
I tested the fix, and it clears the type when "No Type" is selected and retains the type when the options are selected so it looks good to me.

wetneb:
Thank you both!

PR #5549, opened 2023-01-05
header: move "OpenRefine" header text from image

Abbe98:
So that the background and foreground text can be styled.

side effects:
 - minor style changes (acceptable in my opinion)
 - accessibility improvements
 - dropped some absolute positioning
 - fixed a bug causing the slogan to shift from the left during page load

Part of #3017

### DIFF
--- a/main/webapp/modules/core/styles/common.css
+++ b/main/webapp/modules/core/styles/common.css
@@ -277,15 +277,30 @@ a img {
 
 #header {
   height: 40px;
-  margin: 10px;
-  position: relative;
-  overflow: hidden;
+  padding: 10px;
+  display: flex;
+  align-items: center;
+  box-sizing: border-box;
+}
+
+#header h1 {
+  margin: 0 .2em;
+  color: var(--text-primary);
+  font-family: serif;
+  font-size: 1.8em; /* to avoid being styled by pure.css from the database extension */
+}
+
+#header img {
+  height: 30px;
 }
 
 #app-home-button {
-  position: absolute;
-  top: 0;
-  left: 0;
+  display: flex;
+  align-items: center;
+}
+
+#app-home-button:hover {
+  text-decoration: none;
 }
 
 #body-info {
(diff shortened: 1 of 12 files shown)


Abbe98:
Header after these edits:

Screenshot 2023-01-05 at 15-48-52 OpenRefine
Screenshot 2023-01-05 at 15-49-08 types - OpenRefine

Abbe98:
Two things that could be improved:
 - The "Navigate to dashboard" text could be made translatable (I don't know how given that some files currently do not use translation)
 - `<h1>` doesn't make sense on pages with an existing one, however, given that these are inside of the "home button" it could make sense just to hide it with `aria-hidden`

main/webapp/modules/core/styles/common.css
```diff
@@ -277,15 +277,30 @@ a img {
...
+  display: flex;
+  align-items: center;
+  box-sizing: border-box;
+}
+
+#header h1 {
+  margin: 0 .2em;
+  color: var(--text-primary);
+  font-family: serif;
+  font-size: 1.8em; /* to avoid being styled by prue.css from the database extension */
```
wetneb, line 290:
```suggestion
  font-size: 1.8em; /* to avoid being styled by pure.css from the database extension */
```

wetneb (commented):
Thanks @Abbe98 for this!

I think the move makes a lot of sense.
Concerning "Navigate to dashboard", I would rather write "Navigate to the home screen" because I do not think we use the term "dashboard" elsewhere for this page, and [the documentation uses "home screen"](https://openrefine.org/docs/manual/running#the-home-screen). The term "dashboard" intuitively implies to me the visualization of some metrics, graphs or periodically updated information, which is not really the case (we just list the projects in the workspace, essentially).

I think it would be worth localizing it on all the pages which already have localization set up. I think it should be most of them, no? The error page can stay in English I think - it makes sense not to rely on too many systems there, since we want it to be able to display reliably in all situations.

wetneb (approved):

PR #5550, opened 2023-01-05
database extension: drop Bootstrap and Pure #5548

### LINKED ISSUE #5548: The database extension overrides CSS styles from core
The database extensions loads Yahoo's prue.css library which styles elements regardless of classes and as a result one must always ensure that core got styles which take priority to avoid side effects. 

### To Reproduce

Inspect any element on either the index or project page and check for styles coming from prue.css.

Abbe98:
Fixes #5548

This drops Pure and Bootstrap from the database extension as those libs leak into the rest of the UI. 

Side effects:
 - It changes the overall look and feel of the connection UI (smaller buttons, no colors, etc)


Abbe98:
Post-changes screenshot:
Screenshot 2023-01-05 at 20-24-43 OpenRefine

I think it's worth it for the over all maintainability of OpenRefine.

wetneb:
I am very supportive of this but it might be worth tweaking the new styling a bit more to make the whole thing more inviting and more similar to the rest of OpenRefine, no?
As a user, there is a clear degradation of the appearance and I think we can avoid it.

Here is how the same screen looks before the change:
image

Abbe98:
I would be happy to spend time on restoring the layout and enlarging headers. Beyond that I think it would be better if someone actually using the extension would make additional improvements(I usually uninstall this extension).

Abbe98:
I have been restoring the basic layout as well as some padding/alignment. In the best of one would probably like to adopt the Dialog system but I feel like that's beyond scope. 

Screenshot 2023-01-16 at 11-07-31 OpenRefine

elebitzero (commented):
These changes are looking better especially the query editor, still the 'New connection' button probably needs slightly more space above the 'Saved connections' and the query editor's textbox is leaking out of the right border and lined up almost over the left as shown below. The old design also had a more balanced look for the input areas at the start by lining everything up with more space but that can be fixed later.
Query editor textbox leak

elebitzero:
Also the connections area in 'Saved connections' doesn't show a scrollbar past the current size:

Before:
Saved connections overflow before 3


After:
saved connections after

Abbe98:
The overflow issues are taken care off but the spacing still looks kinda off, it's hard to get rights without starting to refactor things or going overboard with hard-coded pixel values.

wetneb (approved):
I made a small tweak to try to bring the UI a bit more in line with the rest. It's not perfect for sure. I'll open an issue to improve it, perhaps our Outreachy design applicants will be inspired.

Issue #550, opened 2012-10-15
Split eats empty pieces


tfmorris:
_Original author: tfmorris (March 14, 2012 23:34:17)_

&quot;/&quot;.split(&quot;/&quot;) returns one piece instead of two and &quot;/user/foo&quot; returns just two, making it impossible to reliably extract the namespace (without using a trick like (&quot; &quot;+&quot;/user/foo&quot;).split(&quot;/&quot;) ) which works as expected.

Split should return N+1 pieces when given a string with N separators independent of whether the separator is a leading or trailing character.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=550_

Issue #456, opened 2012-10-15
GREL split() drops empty pieces


tfmorris:
_Original author: tfmorris (September 30, 2011 18:17:51)_

value.split('/') produces the following preview (string + result):

//Google Inc//Google Calendar 70.9054//EN   [ &quot;Google Inc&quot;, &quot;Google Calendar 70.9054&quot;, &quot;EN&quot; ]

Python thinks the answer is ['', '', 'Google Inc', '', 'Google Calendar 70.9054', '', 'EN'] which fits with my world view.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=456_

Issue #301, opened 2012-10-15
Question about a word count feature/transformation -- is this possible?


tfmorris:
_Original author: elizabet...@gmail.com (January 06, 2011 22:17:27)_

Hello -- I have a question about an operation I would very much like to perform, but do not know how. Specifically, I want to get the word count for the number of words in a given cell, so that I could perform some transformation that would produce a parallel column with numerical values denoting the word count for each cell in the &quot;project description&quot; column. In other words, I want to be able to compare the length of descriptions across a huge number of projects. 

Is there a way to do this? What are the commands one uses? Where can I go to read more about them? 

Thank you so much - I am a very enthusiastic novice here... 

Also, I'm using the latest download of google refine for a Mac OS X.

Best, 
elizabeth

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=301_

PR #5551, opened 2023-01-05
Use internationalized word separator. Fixes #557.

### LINKED ISSUE #557: Word facet should split on all whitespace
_Original author: tfmorris (April 07, 2012 15:42:20)_

In particular, it should consider a no-break space (' ',NBSP) a word delimiter, but probably should use all whitespace characters.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=557_

tfmorris:
Fixes #557

- internationalize word facet
- 
-

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/menu-facets.js
+++ b/main/webapp/modules/core/scripts/views/data-table/menu-facets.js
@@ -158,7 +158,7 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
                 {
                   "name": column.name,
                   "columnName": column.name,
-                  "expression": "value.split(' ')"
+                  "expression": "value.split(/(?U)[\\W]/)"
                 }
             );
           }
--- a/main/webapp/modules/core/scripts/views/data-table/menu-facets.js
+++ b/main/webapp/modules/core/scripts/views/data-table/menu-facets.js
@@ -158,7 +158,7 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
                 {
                   "name": column.name,
                   "columnName": column.name,
-                  "expression": "value.split(/(?U)[\\W]/)"
+                  "expression": "value.split(/(?U)[\\W]+/)"
                 }
             );
           }


ostephens:
LGTM. An update to the documentation at https://github.com/OpenRefine/openrefine.org/blob/master/docs/manual/facets.md#word-facet-word-facet will be needed to reflect this change

ostephens:
Just noticed one minor thing (and it may not matter). The behaviour of the old version `split(" ")` and new `split(/(?U)[\W]+/)` behaves differently for leading and trailing non-word characters in the cell. For example:

```
" abc ".split(" ") -> [ "abc", "" ]
" abc ".split(/(?U)[\W]+/) -> [ "", "abc" ]
```
As can be seen the former results in the trailing character being counted as a 'blank' word, whereas the latter results in the leading character being treated as a 'blank' word.

My feeling is that we should just omit blanks altogether in a word facet (this would already be the behaviour for any consecutive 'non-word' characters. Possibly `trim().split(/(?U)[\W]+/)` is the simplest approach? 

This is definitely a minor issue IMO and I'd be happy to see this change merged either way

wetneb:
Thanks @tfmorris! @ostephens' suggestion to add a call to `trim()` makes sense to me, especially now that `trim()` itself properly supports Unicode.

Just to make sure we are on the same page (because this was not obvious to me), this change is not just changing from splitting on the space character to all Unicode whitespace (as the `W` can intuitively suggest), but all non-word characters, which includes a lot more stuff. For instance:
* Before this change: `"Which sub-directory? No idea! :)"` -> `["Which", "sub-directory?", "No", "idea!", ":)"]`
* After this change: `"Which sub-directory? No idea! :)"` -> `["Which", "sub", "directory", "No", "idea"]`

The fact that trailing punctuation is trimmed automatically seems clearly desirable. I can imagine that one could prefer keeping "sub-directory" as one word. Whether smileys are words can surely spark interesting linguistic debates… But this change is overall a step in the right direction, I would say.

Also, for reproducibility considerations, the GREL expression of the facet is written in JSON workflows so someone executing a workflow built with the old version of the word facet should retain stable functionality, which is a good thing.

tfmorris:
Thanks for the reviews. Both the old and new behaviors seem non-intuitive to me and I'm not sure they don't represent a bug in `split()`, but I'd need to investigate more. I added the `+` because my original without it did:
```
" ab  c ".split(/(?U)[\W]/) => [ "", "ab", "", "c" ]
" ab  c ".split(" ") => [ "ab", "c", "" ]
```
It seems irregular that empty tokens are filtered at the beginning and in the middle but not the end for the old behavior and the end, but not the middle in the new behavior. I would expect empty pieces to either be included or excluded on a consistent basis.

I would actually lean towards filtering all empty pieces for this application (since they're not "words").

> this change is not just changing from splitting on the space character to all Unicode whitespace (as the W can intuitively suggest), but all non-word characters

Correct. Unicode whitespace is `\s` which equates to `[\p{IsWhite_Space}]` (ie has the Unicode property White_Space set), while `\W` is "non-word character " `[^\w]` where `\w` (word character) is defined as   `[\p{Alpha}\p{gc=Mn}\p{gc=Me}\p{gc=Mc}\p{Digit}\p{gc=Pc}\p{IsJoin_Control}]`.

Since it's called a "Word Facet", I just punted the whole definition of "word-ness" to the Unicode Consortium on the assumption that they are the experts. Unicode has a more sophisticated word boundary algorithm, but I don't think it can be supported with a simple regex pattern: https://www.unicode.org/reports/tr29/tr29-37.html#Word_Boundaries

ostephens:
> I would actually lean towards filtering all empty pieces for this application (since they're not "words").
Agreed

ostephens:
>Both the old and new behaviors seem non-intuitive to me and I'm not sure they don't represent a bug in split(), but I'd need to investigate more.

I'd agree
`split` takes a second boolean argument for 'perserve tokens'. This should default to `false` and so I think that the behaviour of both
```
" ab  c ".split(/(?U)[\W]/) => [ "", "ab", "", "c" ]
" ab  c ".split(" ") => [ "ab", "c", "" ]
```
is incorrect.
Note that 
```
" ab  c ".split(" ",true) => [ "", "ab", "", "c", "" ]
```
which seems correct, and further suggests to me that the behaviour when the second argument is false/omitted is not currently what should be expected

thadguidry:
@ostephens Correct, that "looks" like a regression to me along OpenRefine's timeline after we asked for the feature in Google Refine 2.0 https://github.com/OpenRefine/OpenRefine/issues/301 and then @tfmorris did question a few things indeed back then in https://github.com/OpenRefine/OpenRefine/issues/456 and https://github.com/OpenRefine/OpenRefine/issues/550 which was after David first implemented my request in 2010 prior.  I had asked David way back in early days of GEL I think here: https://github.com/OpenRefine/OpenRefine/commit/2ff0184c65c8ea59c36cae70302dfb48a04266f9 for preserving whitespace with `split()` and if the argument is false it should not output a blank element, ever.  I don't know when that regression took place however or if it is one.  Maybe in Apache StringUtils updates?  Or maybe I just didn't test the trailing space! Doh!  But doesn't seem likely since this hasn't changed in StringUtils?

>The StringUtils class defines certain words related to String handling.
>
> null - null
> empty - a zero-length string ("")
> space - the space character (' ', char 32)
> whitespace - the characters defined by [Character.isWhitespace(char)] (https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html?is-external=true#isWhitespace-char-)
> trim - the characters <= 32 as in [String.trim()](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html?is-external=true#trim--)

@tfmorris For Word Facet, perhaps we should just make it right and first do `StringUtils.stripAll()` before anything else?

wetneb:
Stripping whitespace is not enough to prevent empty strings from being added to the result array, see this case:

 `"Which sub-directory? No idea! :)"` -> `["Which", "sub", "directory", "No", "idea", ""]`

We'd need to either strip all leading and trailing non-word characters before splitting, or filtering out empty strings after splitting.

thadguidry:
@wetneb well, I'd agree I guess with @tfmorris on punting on word boundries.  And I guess we should filter out empty strings after splitting to preserve my original ask of the feature to David and how it was described and supposed to work? `:-)`

ostephens:
Rather than digging into the behaviour of `split()` on this issue (which is meant to be a "quick fix") I'd suggest we accept this change (with an update to the documentation) and then dig into the behaviour of `split` and the `preserveAllTokens` parameter separately.

tfmorris:
What is needed to make progress on this? It's been three weeks and I don't see any actionable feedback.

I agree with @ostephens that the `split()` bug is a separate issue. I'll create a ticket for it (unless someone else already has).

wetneb (approved):
Per @ostephens' comment above

PR #5553, opened 2023-01-07
CSS Migration Fix

elebitzero:
- Fixes a regression introduced by #5525, where the slogan text and the browsing panel help text (using facets/filters window) both look lighter.
- Added the CSS variable `text-quarternary` that replaces the LESS variable `@near_black` and its usages.

Slogan before:
Slogan before

Slogan after:
Slogan After

Browsing panel before:
Browsing panel help before

Browsing panel after:
Browsing panel help After


wetneb (approved):

Abbe98 (approved):
+1!

I want to look over/reduce the amount of shades but it should be done separately from LESS/CSS migration and similar changes.

PR #5562, opened 2023-01-16
adopt CSS variables for primary background and text

Abbe98:
Part of #3017.

### DIFF
--- a/main/webapp/modules/core/styles/index/default-importing-parsing-panel.css
+++ b/main/webapp/modules/core/styles/index/default-importing-parsing-panel.css
@@ -87,7 +87,7 @@ td.default-importing-parsing-control-panel-formats {
 }
 
 td.default-importing-parsing-control-panel-options-panel {
-  background: var(--background-primary)white;
+  background: var(--background-primary);
   padding: var(--padding-looser);
   border-top-left-radius: 5px;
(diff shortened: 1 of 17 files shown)


main/webapp/modules/core/styles/index/default-importing-parsing-panel.css
```diff
@@ -81,13 +81,13 @@ td.default-importing-parsing-control-panel-formats {
...
-  background: white;
+  background: var(--background-primary);
   font-weight: bold;
-  color: black;
+  color: var(--text-primary);
 }
 
 td.default-importing-parsing-control-panel-options-panel {
-  background: white;
+  background: var(--background-primary)white;
```
wetneb, line 90:
```suggestion
  background: var(--background-primary);
```

wetneb (approved):
Makes sense!
I think it's hard to make sure all (and only) the relevant places are updated but it should be easier figure it out later when your dark mode is implemented, and it should not break the existing style otherwise.

Abbe98:
> I think it's hard to make sure all (and only) the relevant places are updated but it should be easier figure it out later when your dark mode is implemented, and it should not break the existing style otherwise.

My reasoning as well. Also I think quite a lot of these statements could be removed in favor of inheritance in a later round.

PR #5563, opened 2023-01-16
Use column-removal operation after joining columns

### LINKED ISSUE #4055: 'remove columns after join' should trigger 'column-removal' rather than 'column-reorder' function
When I join columns to create a new one, and I select 'delete joined columns'... this triggers 'column-reorder' which lists and keeps all the columns in this data apart from the ones that I have elected to delete (if 'Write result in selected column' was selected then the other column/s that were joined are removed, if 'Write result in new column' was selected then all the joined columns are deleted).

The problem I have is that I want to extract the processes as JSON and then run on another set of data which may have a different number of (unrelated) columns at the point that the join+reorder happens - the result is that, because 'column-reorder' is triggered, those additional unrelated columns are also unexpectedly deleted... essentially this means that I cannot use 'delete joined columns' option in the 'join columns' interface, and instead run the 'column-removal' processes separately for the joined columns.

eg, I have columns:
A | B | C
-- | -- | --
1 | 2 | 3

I select to 'join columns' B and C (with separator '+'); I select 'Write result in selected column' and 'delete joined columns' option in the popup - the result will be:
A | B 
-- | --
1 | 2+3

So far, so good - but if I extract the process, and then run it on another set of data which contains some additional columns:
A | B | C | D 
-- | -- | -- | --
1 | 2 | 3 | 4 

...becomes:
A | B 
-- | --
1 | 2+3

...whereas I would like it to become:
A | B | D
-- | -- | --
1 | 2+3 | 4

...it is unintuitive that column D is deleted when I have run a process where I selected 'delete joined columns' as it was not part of the 'join columns' process.

The ability to extract the process and run it on a separate set of data is a really powerful application of OpenRefine - eg a process can be designed on a non-final set of data, or repeatedly re-run on updated sets of data - but the way that 'delete joined columns' works prevents that, if the columns in subsequent data are not all present in the initial data that has been used to design the process.

(I am running Version 3.4.1 [437dc4d])

### Proposed solution
It would be much more convenient if 'delete joined columns' ran 'column-removal' process on the appropriate columns, rather than 'column-reorder' - that would mean there was no risk that running the extracted process on other source data would lose columns unexpectedly... and it would also be more consistent with what the user expects when selecting that option (ie that only the specified columns will be removed).

wetneb:
Closes #4055.

This generalizes the `row-removal` operation to support removing multiple columns at once, and uses this operation to remove the joined columns in the `Edit column -> Join columns` operation (when the user selected this option in the dialog).

This way, the JSON representation of the sequence of operation becomes:
```json
[
  {
    "op": "core/column-addition",
    "engineConfig": {
      "facets": [],
      "mode": "row-based"
    },
    "baseColumnName": "NDB_No",
    "expression": "join ([coalesce(cells['NDB_No'].value,''),coalesce(cells['Shrt_Desc'].value,'')],'')",
    "onError": "keep-original",
    "newColumnName": "joined",
    "columnInsertIndex": 1,
    "description": "Create column joined at index 1 based on column NDB_No using expression join ([coalesce(cells['NDB_No'].value,''),coalesce(cells['Shrt_Desc'].value,'')],'')"
  },
  {
    "op": "core/column-removal",
    "columnNames": [
      "NDB_No",
      "Shrt_Desc"
    ],
    "description": "Remove columns NDB_No, Shrt_Desc"
  }
]
```

This does not refer to the rest of the columns in the project, making it easier to generalize the workflow in other situations.

This is a prerequisite for #5559, whose Cypress tests will fail without this.

### DIFF
--- a/main/src/main/java/org/openrefine/commands/column/RemoveColumnCommand.java
+++ b/main/src/main/java/org/openrefine/commands/column/RemoveColumnCommand.java
@@ -34,16 +34,20 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package org.openrefine.commands.column;
 
 import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import com.fasterxml.jackson.core.type.TypeReference;
 import org.openrefine.commands.Command;
 import org.openrefine.model.Project;
 import org.openrefine.operations.Operation;
 import org.openrefine.operations.column.ColumnRemovalOperation;
 import org.openrefine.process.Process;
+import org.openrefine.util.ParsingUtilities;
 
 public class RemoveColumnCommand extends Command {
 
@@ -58,9 +62,11 @@ public class RemoveColumnCommand extends Command {
         try {
             Project project = getProject(request);
 
-            String columnName = request.getParameter("columnName");
+            String columnNames = request.getParameter("columnNames");
 
-            Operation op = new ColumnRemovalOperation(columnName);
+            Operation op = new ColumnRemovalOperation(request.getParameter("columnName"),
+                    ParsingUtilities.mapper.readValue(columnNames, new TypeReference<List<String>>() {
+                    }));
             Process process = op.createProcess(project);
 
             performProcessAndRespond(request, response, project, process);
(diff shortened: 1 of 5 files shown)


wetneb:
Note: although this PR generalizes the column-removal operation to support removing multiple columns at once, the previous serialization format of this operation is still supported, so using JSON representations of operation sequences which include this operation will still work.

As a follow-up to this change, I think the column reordering dialog should use this new column-removal operation (and not the general column-reorder operation) if the user only deletes some columns without reordering any of the remaining columns. This would make the operation more generalizable (as seen in #4055). I have opened an issue for this: #5576.

PR #5568, opened 2023-01-17
Fixes #5567 Migrate PR workflow to JDK Eclipse Temurin

### LINKED ISSUE #5567: Migrate to use Java distribution - Eclipse Temurin since AdoptOpenJDK is archived and not publishing security releases
We are missing out on security updates for the Java distribution since we have not migrated yet.  As noted in https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#adopt
> NOTE: Adopt OpenJDK got moved to Eclipse Temurin and won't be updated anymore. It is highly recommended to migrate workflows from adopt to temurin to keep receiving software and security updates. See more details in the [Good-bye AdoptOpenJDK post](https://blog.adoptopenjdk.net/2021/08/goodbye-adoptopenjdk-hello-adoptium/).

### To Reproduce
1. Run the PR workflow.

### Current Results
Latest JDK semver (currently jdk-17.0.5+8) from matrix is not used for build/test.

### Expected Behavior
We should migrate to using Eclipse `temurin` distribution since Adopt OpenJDK `adopt` has now officially been archived and no longer providing security releases (JCK and AQAvit certified) but instead those are under `temurin`.

thadguidry:
Fixes #5567 

-  replaces `adopt` with `temurin` for Java distribution

@wetneb Should this also be replaced in the other workflows?

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -71,7 +71,7 @@ jobs:
     - name: Set up Java ${{ matrix.java }}
       uses: actions/setup-java@v3
       with:
-        distribution: 'adopt'
+        distribution: 'temurin'
         java-version: ${{ matrix.java }}
 
     - name: Check Java linting
@@ -115,7 +115,7 @@ jobs:
     - name: Set up Java 17
       uses: actions/setup-java@v3
       with:
-        distribution: 'adopt'
+        distribution: 'temurin'
         java-version: 17
 
     - name: Check Java linting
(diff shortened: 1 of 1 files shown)


wetneb:
Thanks, that looks good to me. You could do the same on the snapshot release workflow.
The JREs we bundle with our packaged artifacts are fetched from adoptium.net which serves temurin releases already, so that should not need tweaking.

wetneb:
Superseded by #5571 #5570

PR #5570, opened 2023-01-17
Fixes #5569 Use 'cache' feature in setup-java step.

### LINKED ISSUE #5569: Simplify the CI/CD workflows to use the builtin Maven cache feature in setup-java action that now supports it
We can now remove the actions/cache steps and replace them with a config setting under `setup-java` action.  It automatically uses actions/cache already under the hood and will create the fileHash for the Maven `**/pom.xml` automatically.
Documention: https://github.com/actions/setup-java/blob/main/README.md#caching-packages-dependencies

Usage such as:
```
steps:
- uses: actions/checkout@v3
- uses: actions/setup-java@v3
  with:
    distribution: 'temurin'
    java-version: '17'
    cache: 'maven'
- name: Build with Maven
  run: mvn -B package --file pom.xml
```

### To Reproduce
1. Run a PR workflow.

### Current Results
Extra steps not need for `actions/cache` separately.

### Expected Behavior
cleaner and less verbose build/test output in logs.

thadguidry:
Fixes #5569

- removes separate actions/cache for java-setup steps.
- adds `cache` config to java-setup action.
- but keeps actions/cache setup for Yarn.  I think this is wise?

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -60,19 +60,12 @@ jobs:
         ref: ${{ github.event.pull_request.head.ref }}
         repository: ${{ github.event.pull_request.head.repo.full_name }}
 
-    - name: Restore dependency cache
-      uses: actions/cache@v3
-      with:
-        path: ~/.m2/repository
-        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
-        restore-keys: |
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thanks, you could do it on the snapshot.yml workflow as well.

wetneb (approved):

PR #5577, opened 2023-01-20
Add dialog to warn of history entry deletion

### LINKED ISSUE #3184: Add warnings for Undo/Redo history changes
Currently OpenRefine does not warn the user about losing a portion of their workflow history.  It is non-obvious if a user attempts to rollback and undo some changes, and then apply some new changes, they will silently lose any forward changes without a warning.  We should warn users about destroying forward history and not let them be surprised or reading details in our docs.

### Proposed solution

A confirm dialog could appear before the very next Change operation a user does that will overwrite history after the chosen undo step:
1. User chooses to undo 3 steps
2. User attempts to do a new Transform on a column of text
3. A warning appears saying:
----
**You have Undo/Redo changes after this point in your workflow history that will be lost and replaced with this operation.**
**Are you sure you want to change the workflow history?**

[Yes] / [No]
[ x ] disable this warning in future

----

wetneb:
Closes #3184.

The tick box in the dialog is not doing anything yet, that will need updating before merging.

image


wetneb:
I have introduced a preference, `ui.history.warnAgainstDeletion`, which can be set to `'true'` or `'false'`, to store the users' choice when they tick the checkbox. Let me know if you prefer another key.

wetneb:
> MESSAGE: Are you sure you want to delete the selected items from this user's history?

Intuitively I feel like it is important to make clear that the proposed deletion is due to the addition of a new change. Otherwise new users might not understand why this deletion is proposed (as in "why does it want to delete my super long reconciliation operation? I was just editing a cell!") but perhaps I am overly cautious here.

Also did you mean "this project's history"? We do not really have a notion of user at the moment so it feels awkward for me to mention "this user's history".

> CHECKBOX: Don't show again

That seems better indeed.

> Also I was wondering how it handles showing a higher number of history entries, like 50 entries.

There is a provision for auto scroll when the list gets longer:

```css
max-height: 20em;
overflow: auto;
```

elebitzero:
After commenting on the wording I read the issue #3184, and I'm not sure this is necessary.

Other applications like document editors, IDEs, MS Excel, etc. which have a stack of changes and have redo/undo, do not pop up a warning like this if you undo a few changes, and then start making new changes. The user is expected to know what they are doing in this case.

For someone that knows how it works, it would be potentially confusing to have this warning dialog pop up.

wetneb:
Thanks, I replied on the issue in #3184.

PR #5503, opened 2022-12-07
Bump Cypress from 11.2.0 to 12.3.0

elebitzero:
Changes proposed in this pull request:
- Updates Cypress to 12.3.0.
- Removed usage of `cy.server()` since it was removed in Cypress 12.0.0 and added a workaround from https://stackoverflow.com/questions/71357705/hide-xhr-calls-on-cypress-test-runner to hide XHR requests as a replacement.
- Fixed tests that used `cy.within()` to work on Cypress 12.
- Fixed a test in facets.cy.js to work on Cypress 12.
- Fixed a test in pagination.cy.js to work on Cypress 12.
- Fixed test in rows_records.cy.js to work in Cypress 12.
- Changed `cleanupWikibases()` to use the API to get and restore the Wikibase list after each test.
- Changed URL in `getPreference()` to use the `openRefineUrl` variable instead of a hardcoded value.

PR #5578, opened 2023-01-24
make dialogs focusable

Abbe98:
Makes dialogs focus-able, makes it possible
to use dialog headers as modal labels, and
automatically set focus when a dialog is displayed.

### DIFF
--- a/main/webapp/modules/core/scripts/util/dialog.js
+++ b/main/webapp/modules/core/scripts/util/dialog.js
@@ -76,6 +76,15 @@ DialogSystem.showDialog = function(elmt, onCancel) {
 
   DialogSystem.setupEscapeKeyHandling();
 
+  elmt.attr("aria-role", "dialog");
+  var dialogHeader = elmt.find(".dialog-header");
+  if (dialogHeader.length && dialogHeader[0].id) {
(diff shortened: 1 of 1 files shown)


Abbe98:
Failing tests^^, looks like there is a dialog without a header.

elebitzero:
@Abbe98, I checked and there are two tests failing.
I'm not sure why the tests are failing in the following two files:
  * undo_redo.cy.js
  * importing-data-into-openrefine.cy.js
 

I see your branch doesn't have the latest Cypress update that was just merged https://github.com/OpenRefine/OpenRefine/pull/5503  but that shouldn't matter.

I just created a new PR for Cypress 12.4.0 and all of the tests are passing.

You could just trigger a new build to see if the failures are repeatable, or pull the latest master which has the updated Cypress.

You could also try running the Cypress tests locally and see if you still get the errors.

Abbe98:
>  I'm not sure why the tests are failing in the following two files:

@elebitzero yeah I think there is a dialog somewhere without a header, will have a look next time I'm on it to see if the code changes should be hardened or if another change is motivated.

Abbe98:
I ended up hardening the changes after considering that enforcement of headers might be a breaking change for some extensions.

elebitzero (approved):

PR #5579, opened 2023-01-24
simplify dialog top position

Abbe98:
Have the default top positioning being set by CSS and only use JavaScript as a fallback for very large modals.

### DIFF
--- a/main/webapp/modules/core/scripts/util/dialog.js
+++ b/main/webapp/modules/core/scripts/util/dialog.js
@@ -53,15 +53,14 @@ DialogSystem.showDialog = function(elmt, onCancel) {
   .css("z-index", 102 + DialogSystem._layers.length * 2)
   .appendTo(document.body);
 
-  elmt.css("visibility", "hidden").appendTo(container);
-// Dialogs should be in the upper half of the window
-// Offset by 10% of window height to separate from browser bar
-// Unless dialog is as big or bigger than the window, then set top to 0 + 5px;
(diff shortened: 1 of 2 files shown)


elebitzero (commented):
When the dialog height is greater than 90% of the view height the new logic is different.  Before, it would inch the dialog top higher from 10vh to a minimum 5 pixel offset. The new logic jumps the dialog immediately to a 5 pixel offset as soon as the dialog height exceeds 90% of view height.

I like the idea of having the browser do the work, but this puts part of the logic in CSS, and another part in Javascript.  So, I'm not sure if this is an improvement.

Abbe98:
@elebitzero the goal here is to only have logic for the edge case in JavaScript and leave the default positioning to CSS. I don't think the edge case motivates to have the default positioning in CSS.

The new code does indeed do less than the old code for that edge case but it allows one to remove the `visible`/`hidden` trick(that affects all rendering).

Abbe98:
It's worth noting that the old code comment only address a part of what was going on when the edge case triggered. That said I'm not aware of a dialog that triggers the edge case to begin with, maybe in an extension somewhere?

elebitzero:
For the visible/hidden trick, I just checked and I don't think that is necessary; we can remove that even with the existing code.

> That said I'm not aware of a dialog that triggers the edge case to begin with, maybe in an extension somewhere?

'Cluster and edit' and 'Templating export' are big dialogs. On some laptops, they may be > 100% of the browser window height.

PR #5588, opened 2023-01-27
Filter empty tokens to match doc. Fixes #5587

### LINKED ISSUE #5587: Fix split() function to match its documentation
The documentation for `split()` says:

"Returns the array of strings obtained by splitting s by sep. The separator can be either a string or a regex pattern. If preserveTokens is true, empty segments are preserved."

This implies, but doesn't explicitly say, that (all) empty tokens are elided, but the actual implementation naively follows the default behavior of the external functions used by the implementation, which differ from this (and from each other).

- `org.apache.commons.lang3.StringUtils.splitByWholeSeparator()` - returns empty tokens if there are trailing separator characters, which I suspect is unintentional from looking at the implementation, but it's a long standing behavior that I suspect won't change.
- `java.util.regex.Pattern.split()` - documented to return an empty token for leading pattern patch.

Neither of these behaviors matches the OpenRefine documentation and, worse, they differ from each other.

### Current Results

From https://github.com/OpenRefine/OpenRefine/pull/5551#issuecomment-1373703149 grel examples by @ostephens 

```
" abc ".split(" ") -> [ "abc", "" ]
" abc ".split(/(?U)[\W]+/) -> [ "", "abc" ]
```

### Expected Behavior
Both cases above return a single token "abc"

### Versions
 - OpenRefine: 3.6

### Additional context

While this is technically a breaking change because it changes the behavior, it would bring the behavior in line with the documentation and user expectations.

tfmorris:
Fixes #5587

- Changes `split()` to filter trailing empty token when there's a trailing string separator and leading empty token when there's a leading pattern separator match.
- 
- 

Note in the diff that I actually noted the anomalous behavior when I wrote the original tests, but never investigated (and just wrote the test to match the current, rather than documented, behavior.

### DIFF
--- a/main/src/com/google/refine/expr/functions/strings/Split.java
+++ b/main/src/com/google/refine/expr/functions/strings/Split.java
@@ -33,6 +33,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package com.google.refine.expr.functions.strings;
 
+import java.util.Arrays;
 import java.util.Properties;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


tfmorris:
@wetneb ping

wetneb (approved):

wetneb:
Sorry for the delay! I am a bit torn that the default behaviour of this function is to filter out tokens (I would rather expect that `preserveAllTokens` is true by default), but that's probably a coder's perspective.

tfmorris:
> I am a bit torn that the default behaviour [sic] of this function is to filter out tokens (I would rather expect that preserveAllTokens is true by default), but that's probably a coder's perspective.

That was the original behavior, so when a second parameter was added to allow the behavior to be overridden, the default was set to preserve the historical behavior for backward compatibility (which is always the top priority).

PR #5590, opened 2023-01-28
project tags: expose tag name in a data attribute

Abbe98:
This change adds a `data` attribute to project tags with the name of that tag so that one can give individual tags custom styles:

```css
.project-tag[data-tag-name="client-y"] {
  background: #0ab97e;
  ...
}
```

Previously you would need to do this with JavaScript to check the actual contents of the tag.


wetneb (approved):
This is obviously rather benign but generally speaking I wonder if we want to consider that an official extension point. I wonder what sort of processes we should have around such decisions.

Approving by default as I do not want to hold this up though.

Abbe98:
>This is obviously rather benign but generally speaking I wonder if we want to consider that an official extension point. I wonder what sort of processes we should have around such decisions.

Yeah there is no lack of cases like this :-/ 

@wetneb  thanks!

PR #5595, opened 2023-01-31
Preserve pagination settings when using undo/redo.

### LINKED ISSUE #572: Undo/Redo page reset problem

wetneb:
Fixes #572.

When the sequence of operations undone or redone all preserve rows (or records, if in records mode), the pagination settings are preserved.


elebitzero (approved):
This change looks good as it preserves the pagination settings when using undo/redo on operations that don't sort.

Also shouldn't the undo button in the notification window work similarly?

For example, I was on page 3 of the data, I starred a row, and clicked undo in the notification, and it was reset to page 1.

wetneb:
Thanks for catching that! I had indeed forgotten to update this as well.

PR #5597, opened 2023-01-31
Make it possible to link to a given home screen panel

Abbe98:
This change makes it possible to link to given panel by its identifier.

To for example go straight to the language settings one could navigate to:
`http://127.0.0.1:3333/#lang-settings`


elebitzero (approved):
Looks good.  I like that if OR is refreshed it keeps the tab you are on.  What about the next level of tabs in 'Create project'?

image

Abbe98:
> What about the next level of tabs in 'Create project'?

@elebitzero I'm not sure, it gets tricky especially if one would like to deal with things like focus management (which I would like to follow up with).

PR #5608, opened 2023-02-06
Added a fix to keep the facet height after performing various operations

### LINKED ISSUE #5606: Facets do not keep their size positions any longer in 3.8 snapshot
REGRESSION:  Text Facet (and maybe others?) is not keeping it's dimensions after I click to include a selection from the list.
See Video below (soon).

### To Reproduce
1. Create Text Facet
2. Then, click on include links for a few selections

### Current Results
Notice that upon every click on a *include* link, the facet window shrinks back to default size!

### Expected Behavior
Facet should remain with the size the user has changed it to (especially for long lists for easier browsing/analysis)

### Screenshots

https://user-images.githubusercontent.com/986438/216807735-ce179a85-b5d6-4f27-ae40-e1e36fa92ea5.mp4

### Versions
 - JRE or JDK Version: JDK 17
 - OpenRefine: Version 3.8-SNAPSHOT [4d7571d]

elebitzero:
Fixes #5606

Fixes the Regression from: https://github.com/OpenRefine/OpenRefine/pull/5514

The initial height is set when a facet is added.  The facet height is not adjusted in the following cases:
 - When other facet is added.
 - When other facet is deleted.
 - Include/Exclude Choice
 - Edit Choice -> Apply
 - Sort by: name count
 - Refresh
 - Cluster Choices -> Merge selected & Close
 - Resize/Min/Max Browser Window
 - Show/Hide Left Panel
 - Min/Max a facet

### DIFF
--- a/main/webapp/modules/core/scripts/facets/list-facet.js
+++ b/main/webapp/modules/core/scripts/facets/list-facet.js
@@ -50,6 +50,7 @@ class ListFacet extends Facet {
 
     this._data = null;
 
+    this._initialHeightSet = false;
     this._initializeUI();
     this._update();
   };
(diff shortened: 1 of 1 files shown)


thadguidry:
@elebitzero Tested and now the Facet height is kept in place if user adjusts it. Thanks!
However, while testing I noticed a few things that ultimately need to be improved at some point for `leftPanelDiv` elements.
I will open some new issues for those.

Abbe98 (approved):
+1

PR #5611, opened 2023-02-08
Added a fix to stop the facets panel from shifting on refresh

### LINKED ISSUE #5610: Facet sidebar shifts when it refreshes
The issue is as @thadguidry says:
> Unneccesary jumping up/down of Facet panels while users click on selections. The div browsing-panel-indicator will quickly show a hidden spinner “Refreshing facets…” which is above the browsing-panel-controls and hence causes all child elements underneath to suddenly move down and then back up. The effect to a user is disconcerting with jerky motion.

image

elebitzero:
Fixes #5610

- Added a fix to stop the facets panel from shifting on refresh as shown below:

After:
image


Abbe98 (approved):

thadguidry:
Tested, works beautifully.  Thanks @elebitzero!

PR #5629, opened 2023-02-15
Enhance forRange to accept negative increments (#5520)

### LINKED ISSUE #5520: Enhance `forRange` to accept negative increments
`forRange` allows iteration based on a defined range of numbers and an increment, but currently only works if the increment is positive. Current examples:
```
forRange(0,10,1,i,i) -> [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 ]
forRange(10,0,-1,i,i) -> [  ]
```

This is different to the `range()` function where a negative increment is accepted:
```
range(0,10,1) -> [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 ]
range(10,0,-1) -> [ 10, 9, 8, 7, 6, 5, 4, 3, 2, 1 ]
```

### Proposed solution
forRange could be enhanced to work with a negative increment
It might also be a good idea to reject invalid arguments - for example currently:
```
forRange(10,0,1,i,i)
```
As there is no way of incrementing by 1 to get from 10 to zero, this should be rejected as invalid parameters, whereas instead it currently returns an empty array

### Alternatives considered
It is possible to use `range` combined with `forEach`:
```
forEach(range(10,0,-1),v,v) -> [ 10, 9, 8, 7, 6, 5, 4, 3, 2, 1 ]
```
This isn't hugely more complicated but it seems that `range` and `forRange` should be consistent with each other

Huishin-pie:
Fixes #5520

- forRange can work with a negative increment.
- Fixes the bug to support both integer and floating point values.
- To reject invalid arguments and return an error message instead of an empty array.

### DIFF
--- a/main/src/com/google/refine/grel/controls/ForRange.java
+++ b/main/src/com/google/refine/grel/controls/ForRange.java
@@ -69,7 +69,7 @@ public class ForRange implements Control {
         } else if (ExpressionUtils.isError(stepO)) {
             return stepO;
         } else if (!(fromO instanceof Number) || !(toO instanceof Number) || !(stepO instanceof Number)) {
-            return ControlEvalError.for_range();
+            return new EvalError(ControlEvalError.for_range());
         }
 
         String indexName = ((VariableExpr) args[3]).getName();
@@ -83,32 +83,46 @@ public class ForRange implements Control {
                 long step = ((Number) stepO).longValue();
                 double to = ((Number) toO).doubleValue();
 
-                if (!((from < to && step > 0) || (from > to && step < 0))) {
+                if (step == 0) {
                     return new EvalError(EvalErrorMessage.invalid_arg());
                 }
 
-                int rangeSize = (int) Math.ceil(Math.abs(from - to) / Math.abs(step));
-                for (int i = 0; i < rangeSize; i++) {
-                    bindings.put(indexName, from);
-                    Object r = args[4].evaluate(bindings);
-                    results.add(r);
-                    from += step;
+                if (step <= Math.abs(from - to)) {
+                    while (from < to && step > 0) {
+                        bindings.put(indexName, from);
+                        Object r = args[4].evaluate(bindings);
+                        results.add(r);
+                        from += step;
+                    }
+                    while (from > to && step < 0) {
+                        bindings.put(indexName, from);
+                        Object r = args[4].evaluate(bindings);
+                        results.add(r);
+                        from += step;
+                    }
                 }
             } else {
                 double from = ((Number) fromO).doubleValue();
                 double step = ((Number) stepO).doubleValue();
                 double to = ((Number) toO).doubleValue();
 
-                if (!((from < to && step > 0) || (from > to && step < 0))) {
+                if (step == 0) {
                     return new EvalError(EvalErrorMessage.invalid_arg());
                 }
 
-                int rangeSize = (int) Math.ceil(Math.abs(from - to) / Math.abs(step));
-                for (int i = 0; i < rangeSize; i++) {
-                    bindings.put(indexName, from);
-                    Object r = args[4].evaluate(bindings);
-                    results.add(r);
-                    from += step;
+                if (step <= Math.abs(from - to)) {
+                    while (from < to && step > 0) {
+                        bindings.put(indexName, from);
+                        Object r = args[4].evaluate(bindings);
+                        results.add(r);
+                        from += step;
+                    }
+                    while (from > to && step < 0) {
+                        bindings.put(indexName, from);
+                        Object r = args[4].evaluate(bindings);
+                        results.add(r);
+                        from += step;
+                    }
                 }
             }
             return results.toArray();
--- a/main/src/com/google/refine/grel/controls/ForRange.java
+++ b/main/src/com/google/refine/grel/controls/ForRange.java
@@ -40,10 +40,7 @@ import java.util.Properties;
 import com.google.refine.expr.EvalError;
 import com.google.refine.expr.Evaluable;
 import com.google.refine.expr.ExpressionUtils;
-import com.google.refine.grel.Control;
-import com.google.refine.grel.ControlDescription;
-import com.google.refine.grel.ControlEvalError;
-import com.google.refine.grel.ControlFunctionRegistry;
+import com.google.refine.grel.*;
 import com.google.refine.grel.ast.VariableExpr;
 
 public class ForRange implements Control {
@@ -86,27 +83,31 @@ public class ForRange implements Control {
                 long step = ((Number) stepO).longValue();
                 double to = ((Number) toO).doubleValue();
 
-                while (from < to) {
+                if (!((from < to && step > 0) || (from > to && step < 0))) {
+                    return new EvalError(EvalErrorMessage.invalid_arg());
+                }
+
+                int rangeSize = (int) Math.ceil(Math.abs(from - to) / Math.abs(step));
+                for (int i = 0; i < rangeSize; i++) {
                     bindings.put(indexName, from);
-
                     Object r = args[4].evaluate(bindings);
-
                     results.add(r);
-
                     from += step;
                 }
             } else {
-                double from = ((Number) fromO).longValue();
-                double step = ((Number) stepO).longValue();
+                double from = ((Number) fromO).doubleValue();
+                double step = ((Number) stepO).doubleValue();
                 double to = ((Number) toO).doubleValue();
 
-                while (from < to) {
+                if (!((from < to && step > 0) || (from > to && step < 0))) {
+                    return new EvalError(EvalErrorMessage.invalid_arg());
+                }
+
+                int rangeSize = (int) Math.ceil(Math.abs(from - to) / Math.abs(step));
+                for (int i = 0; i < rangeSize; i++) {
                     bindings.put(indexName, from);
-
                     Object r = args[4].evaluate(bindings);
-
                     results.add(r);
-
                     from += step;
                 }
             }
--- a/main/src/com/google/refine/grel/controls/ForRange.java
+++ b/main/src/com/google/refine/grel/controls/ForRange.java
@@ -87,14 +87,15 @@ public class ForRange implements Control {
                     return new EvalError(EvalErrorMessage.invalid_arg());
                 }
 
-                if (step <= Math.abs(from - to)) {
-                    while (from < to && step > 0) {
+                if (step > 0) {
+                    while (from < to) {
                         bindings.put(indexName, from);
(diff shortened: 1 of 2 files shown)


main/src/com/google/refine/grel/controls/ForRange.java
```diff
@@ -86,27 +83,31 @@ public Object call(Properties bindings, Evaluable[] args) {
                 long step = ((Number) stepO).longValue();
                 double to = ((Number) toO).doubleValue();
 
-                while (from < to) {
-                    bindings.put(indexName, from);
+                if (!((from < to && step > 0) || (from > to && step < 0))) {
```
tfmorris, line 86: We want to make sure to check for `step == 0` here since that will cause an infinite loop, but I don't think we should be more permissive about parameters which will cause zero iterations e.g. from 0 to 10 by -1 because those values might be calculated and we don't want the callers to have to do special handling when zero iterations is an adequate solution.

main/src/com/google/refine/grel/controls/ForRange.java
```diff
@@ -86,27 +83,31 @@ public Object call(Properties bindings, Evaluable[] args) {
...
-
                     results.add(r);
-
                     from += step;
                 }
             } else {
-                double from = ((Number) fromO).longValue();
-                double step = ((Number) stepO).longValue();
+                double from = ((Number) fromO).doubleValue();
+                double step = ((Number) stepO).doubleValue();
```
tfmorris, line 99: Great! Thanks for fixing this bug.

main/src/com/google/refine/grel/controls/ForRange.java
```diff
@@ -86,27 +83,31 @@ public Object call(Properties bindings, Evaluable[] args) {
...
-                while (from < to) {
-                    bindings.put(indexName, from);
+                if (!((from < to && step > 0) || (from > to && step < 0))) {
+                    return new EvalError(EvalErrorMessage.invalid_arg());
+                }
 
+                int rangeSize = (int) Math.ceil(Math.abs(from - to) / Math.abs(step));
+                for (int i = 0; i < rangeSize; i++) {
+                    bindings.put(indexName, from);
                     Object r = args[4].evaluate(bindings);
```
tfmorris, line 93: I think the way that this was originally organized without the intermediate `rangeSize` calculation is more direct and avoids any additional loss of precision.

main/src/com/google/refine/grel/controls/ForRange.java
```diff
@@ -86,27 +83,31 @@ public Object call(Properties bindings, Evaluable[] args) {
...
+                double step = ((Number) stepO).doubleValue();
                 double to = ((Number) toO).doubleValue();
 
-                while (from < to) {
-                    bindings.put(indexName, from);
+                if (!((from < to && step > 0) || (from > to && step < 0))) {
+                    return new EvalError(EvalErrorMessage.invalid_arg());
+                }
 
+                int rangeSize = (int) Math.ceil(Math.abs(from - to) / Math.abs(step));
```
tfmorris, line 106: See comments above for the integer branch.

main/tests/server/src/com/google/refine/grel/controls/ForRangeTests.java
```diff
@@ -27,15 +27,55 @@
...
+
+    @Test
+    public void testEvalError() {
+        bindings = new Properties();
+        bindings.put("v", "");
+        String tests[] = {
+                "forRange(10,0,1,v,v)",
+                "forRange(0,10,-1,v,v)",
+                "forRange(10,10,-1,v,v)",
+                "forRange(10,0,0,v,v)",
```
tfmorris, line 69: The last one should be an error, but all the others should succeed with zero iterations.

main/tests/server/src/com/google/refine/grel/controls/ForRangeTests.java
```diff
@@ -27,15 +27,55 @@
...
 
     @Test
     public void serializeForRange() {
         String json = "{\"description\":\"Iterates over the variable v starting at \\\"from\\\", incrementing by \\\"step\\\" each time while less than \\\"to\\\". At each iteration, evaluates expression e, and pushes the result onto the result array.\",\"params\":\"number from, number to, number step, variable v, expression e\",\"returns\":\"array\"}";
         TestUtils.isSerializedTo(new ForRange(), json);
     }
+
+    @Test
+    public void testForRangeArray() throws ParsingException {
+        String test[] = { "forRange(10.5,8,-0.5,v,v).join(',')", "10.5,10.0,9.5,9.0,8.5" };
```
tfmorris, line 55: Please add a test or two for things like `from == to`, `step > from - to`, and other edge cases. The code is pretty simple, so but it's a good habit to get into to test the corners.

tfmorris (changes requested):
Great start! I made a couple of comments for minor things to improve.
@wetneb may have additional comments.

Huishin-pie:
@tfmorris Thank you for the review. I will improve the code as soon as possible. :)

wetneb (approved):
Thanks for working on this! Nothing to add on my side beyond @tfmorris' review

main/tests/server/src/com/google/refine/grel/controls/ForRangeTests.java
```diff
@@ -51,23 +51,67 @@ public void serializeForRange() {
...
+        parseEval(bindings, test);
+    }
+
+    @Test
+    public void testForRangeWithFloatingPointValues() throws ParsingException {
+        String test[] = { "forRange(0,0.6,0.1,v,v).join(',')", "0.0,0.1,0.2,0.3,0.4,0.5" };
         bindings = new Properties();
         bindings.put("v", "");
+        parseEval(bindings, test);
+    }
```
Huishin-pie, line 83: There was a failure in this test due to floating-point precision. Do we need to resolve this issue?

main/tests/server/src/com/google/refine/grel/controls/ForRangeTests.java
```diff
@@ -27,15 +27,99 @@
...
+        bindings.put("v", "");
+        parseEval(bindings, test);
+    }
+
+    @Test
+    public void testForRangeWithFloatingPointValues() throws ParsingException {
+        String test[] = { "forRange(0,0.6,0.1,v,v).join(',')", "0.0,0.1,0.2,0.3,0.4,0.5" };
+        bindings = new Properties();
+        bindings.put("v", "");
+        parseEval(bindings, test);
```
Huishin-pie, line 82: There was a failure in this test due to floating-point precision. Do we need to resolve this issue?

Huishin-pie:
Please review the changes made. Thank you!

main/tests/server/src/com/google/refine/grel/controls/ForRangeTests.java
tfmorris: Sorry for missing this question! Yes, all tests need to pass before we can merge. The traditional way to do floating point comparisons is to make sure two values are within a small value of each other ie `abs(f1 - f2) < epsilon` rather than `f1 == f2` but since these are string comparisons, a different solution will be needed.

main/src/com/google/refine/grel/controls/ForRange.java
```diff
@@ -83,32 +83,46 @@ public Object call(Properties bindings, Evaluable[] args) {
...
                     return new EvalError(EvalErrorMessage.invalid_arg());
                 }
 
-                int rangeSize = (int) Math.ceil(Math.abs(from - to) / Math.abs(step));
-                for (int i = 0; i < rangeSize; i++) {
-                    bindings.put(indexName, from);
-                    Object r = args[4].evaluate(bindings);
-                    results.add(r);
-                    from += step;
+                if (step <= Math.abs(from - to)) {
```
tfmorris, line 90: This test is unnecessary. It will be handled naturally by the loop logic (ie if step > (from-to) there will be zero iterations.

main/src/com/google/refine/grel/controls/ForRange.java
```diff
@@ -83,32 +83,46 @@ public Object call(Properties bindings, Evaluable[] args) {
...
+                    while (from < to && step > 0) {
```
tfmorris, line 91: `step` is invariant here, so I would either convert this an `if` statement
```
if (step > 0) {
  while (from < to) {
  }
}
else {
  while (from < to) {
  }
}
```
or, to reduce the redundancy, fold both loops together and make the conditional slightly more complex
```
while ((step > 0 && from < to) || to > from) {
```
(we don't need to test `step < 0` because we know it's not zero and not greater than zero)

main/tests/server/src/com/google/refine/grel/controls/ForRangeTests.java
tfmorris: Also, I hadn't noticed that the test workflows required approval to run. Now that they've been approved once, I'm hoping they'll run automatically on future commits.

tfmorris (changes requested):

main/tests/server/src/com/google/refine/grel/controls/ForRangeTests.java
Huishin-pie: Thank you for your response. This is my first time writing a test for floating-point numbers, so I'm not very familiar with it. Could you please give me some more advice? I've researched and only found that the solution as you said is to check if the expected and actual values are equal within a certain delta tolerance.

tfmorris: Since this is operating in the string domain, one option would be to use the precision specifier on `toString()` so change the test to something like 
```
forRange(0,0.6,0.1,v,v.toString('%.3f').join(',')
```
and then adjust to expected values to something like `0.000`. This will give you enough precision to catch real errors, but not so much precision that you're bothered by differences too small to care about.

BTW, I'll be offline for the next 2-3 of days, but will check back in when I return (and @wetneb ) and the other developers can help in the mean time.

main/src/com/google/refine/grel/controls/ForRange.java
Huishin-pie: I misunderstood the meaning of zero iterations. I have adjusted the code.

main/tests/server/src/com/google/refine/grel/controls/ForRangeTests.java
Huishin-pie: I got it! Thanks for your help.

tfmorris: Perfect!

tfmorris (approved):
Looks great. Thanks for sticking with it!

PR #5632, opened 2023-02-16
Fix for failing Cypress CI test

### LINKED ISSUE #5628: Failing Cypress test in the CI
Our CI is red since 37a443d735811be58892d9dcbd5c36c835cddc6f, although the cause is likely external. We should fix this.

elebitzero:
Fixes #5628

-  Adds a check to make sure that there is at least one reconciliation service to select.


tfmorris (approved):
LGTM. Seems a likely fix for the flaky test and no downside risk.

wetneb:
Amazing @elebitzero, so this was actually due to the Wikidata recon service being down… it makes a lot more sense now!

PR #5634, opened 2023-02-16
Fix accessor for "record" field in lists of WrappedRows (output by cross function)

### LINKED ISSUE #5633: Field lookup of .record incorrect for lists of WrappedRows
Evaluation of the `record` field accessor on lists of wrapped rows, such as returned by the `cross()` function incorrectly uses the global context from the bindings when looking up the row index, rather than using the `rowIndex` field in the associated `WrappedRow`s.

### To Reproduce
1. Use the following data to create a new project from the clipboard
```
001,035$a,cross,245$a,020$a
53773,(CaOOAMICUS)000034856822,,Why I love my mummy /,7270208
53806,(CaOOAMICUS)000034856822,,Why I love my daddy /,7877617
```
3. Choose add column based on this column, using the "035$a" column with the expression: `cell.cross("","035$a").record.index`
4. Inspect the results in the preview display

### Current Results

row | value | cell.cross("","035$a").record. 
-- | -- | --
1  | (CaOOAMICUS)000034856822 | [ 0, 0]
2  | (CaOOAMICUS)000034856822 | [ 1, 1 ]

### Expected Behavior

row | value | cell.cross("","035$a").record. 
-- | -- | --
1  | (CaOOAMICUS)000034856822 | [ 0, 1 ]
2  | (CaOOAMICUS)000034856822 | [ 0, 1 ]

### Versions
 - OpenRefine: 3.6.2, 3.7

tfmorris:
Fixes #5633

- Refactor to use static import for asserts so that we won't have a mix of styles in the module (separate commit)
- Fix NullPointerException when null parameters are passed and add a test for the condition (discovered while debugging)
- Fix `record` field access for lists of WrappedRows (only used by `cross()` so test added there)

### DIFF
--- a/main/src/com/google/refine/expr/WrappedRow.java
+++ b/main/src/com/google/refine/expr/WrappedRow.java
@@ -60,8 +60,6 @@ public class WrappedRow implements HasFields {
         } else if ("index".equals(name)) {
             return rowIndex;
         } else if ("record".equals(name)) {
-            int rowIndex = (Integer) bindings.get("rowIndex");
-
             return new WrappedRecord(project.recordModel.getRecordOfRow(rowIndex));
         } else if ("columnNames".equals(name)) {
             Project project = (Project) bindings.get("project");
--- a/main/src/com/google/refine/expr/WrappedRow.java
+++ b/main/src/com/google/refine/expr/WrappedRow.java
@@ -62,8 +62,6 @@ public class WrappedRow implements HasFields {
         } else if ("record".equals(name)) {
             return new WrappedRecord(project.recordModel.getRecordOfRow(rowIndex));
         } else if ("columnNames".equals(name)) {
-            Project project = (Project) bindings.get("project");
-
             return project.columnModel.getColumnNames();
         } else {
             return row.getField(name, bindings);
--- a/main/src/com/google/refine/expr/functions/Cross.java
+++ b/main/src/com/google/refine/expr/functions/Cross.java
@@ -59,13 +59,13 @@ public class Cross implements Function {
             // if 2nd argument is omitted or set to "", use the current project name
             Object targetProjectName = "";
             boolean isCurrentProject = false;
-            if (args.length < 2 || args[1].equals("")) {
+            if (args.length < 2 || "".equals(args[1])) {
                 isCurrentProject = true;
             } else {
                 targetProjectName = args[1];
             }
             // if 3rd argument is omitted or set to "", use the index column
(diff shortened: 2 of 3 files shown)


thadguidry (commented):
@tfmorris My comment to you in your repo (sorry!) was basically about this:
I don't see the coverage happening for some reason in Jacoco report output for some reason or other?  And it maybe that it's getting reported under a different package level than 
`OpenRefine/main/target/site/jacoco/com.google.refine.expr.functions/Cross.html).html`
but I didn't find it under others in the report html structure.
image

Incidentally, something that also worries me just a bit is that we lack coverage for ensuring the fields and in particular the `rowIndex` within a WrappedRecord are accessible under report output /OpenRefine/main/target/site/jacoco/com.google.refine.expr/WrappedRow.java.html

Anyways, that's what I was trying to explain, for some reason the test coverage you added here doesn't seem to be either 1. getting reported or 2. instrumented well enough

tfmorris:
Where are you looking at that report? As far as I can tell, coveralls isn't set up to track pull requests.

Regardless of where you got the report, I don't think you're looking at the right Class/Method. The fix was in WrappedRows#getField and that's where the coverage was added. The fix for the null parameter bug already had coverage for that line, but not in a way that tested the bug, so I wouldn't expect the line-based coverage to change.

As an aside, I don't look at coverage reports unless 1) I'm tracking overall trends or 2) want to look at coverage for a particularly big and/or problematic section of code.  In the context of a bug fix, I test coverage by making sure the test fails without the patch and succeeds with the patch.

thadguidry:
That is from our generated Jacoco report that we have had for a while now.  See the URL path in the address bar. `OpenRefine/main/target/site/jacoco`  We still generate those reports in that path whenever you do `refine test` and I do look at them occasionally.

However, if you and @wetneb think Jacoco reporting is no longer effective, then I'd say "let's cleanup the pom.xml" overall, maybe also removing the coveralls plugin as well if it's not really being used any longer?

wetneb (approved):

tfmorris:
@thadguidry I'm not saying that Jacoco reporting isn't useful, just that I don't look at it for each and every PR or commit.

Having looked at the coverage report & code, though, I added a bunch more tests for all of the WrappedRow and WrappedRecord field accessors. This allowed me to find and fix a similar problem with the `.columnNames` field for `WrappedRow`s.

@wetneb Feel free to re-review, but it's not critical. I'll merge later in the weekend.

thadguidry:
@tfmorris glad it was somewhat useful to improve thing's further.

PR #5638, opened 2023-02-17
Add licenseConfigFile and configuration

### LINKED ISSUE #5637: Fix licensing issues during maven test and packaging
There are a few warnings regarding license information for some dependencies (transitive as well).
We can handle this noise and be a good citizen of open source license information for our dependencies by using and providing more context through `license-maven-plugin:download-licenses`

Below are the current warnings sometimes exhibited depending on connectivity issues:

### To Reproduce
1. First, do `mvn test`
2. Then, notice WARNINGs which happens because of the `generate-sources` lifecycle event in maven which happens during test/packaging.

### Current Results
```
[INFO] --- license-maven-plugin:2.0.0:download-licenses (default) @ packaging ---
[WARNING] No license information available for: oro:oro:2.0.8
[WARNING] Unable to retrieve license from URL 'https://oss.oracle.com/licenses/CDDL+GPL-1.1' for dependency 'javax.servlet:javax.servlet-api:4.0.1': Connect to oss.oracle.com:443 [oss.oracle.com/138.1.117.31] failed: Read timed out
[WARNING] Unable to retrieve license from URL 'https://raw.githubusercontent.com/OpenRefine/OpenRefine/master/LICENSE.txt' for dependency 'org.openrefine:server:3.8-SNAPSHOT': The requested name is valid, but no data of the requested type was found (raw.githubusercontent.com)
[WARNING] There were 3 download errors - check the warnings above
```

### Expected Behavior
The maven plugin: `download-licenses` has several configuration options and we can fallout with ignoring the errors and hiding them through a config flag, but to be a good citizen of open source we should provide manual license information through a `licensesConfigFile` which we can simply store as `manual_licenses.xml` in the `/packaging` folder.
This will eliminate warnings during CI as well as locally for folks that experience connectivity issues occasionally (like me!).
It will also provide the right context and license information for the dependencies that are dual or triple licensed or that no longer are reachable via a URL.

### Versions
 - Browser Version:  n/a
 - JRE or JDK Version: JDK 17
 - OpenRefine: master branch

### Additional context
I'm going to tackle this since I've done this before for another OSS project and I'm almost done with the basic configuration for a PR soon.

thadguidry:
Fixes #5637

- Adds licensesConfigFile for license-maven-plugin
- Adds configuration to packaging/pom.xml

@wetneb for some reason it's not picking up the config file, perhaps I have the wrong path when attempting to use `${project-basedir}/manual_licenses.xml` in the configuration?  Help with that line please?


wetneb:
@thadguidry sorry for the delay! Looking at the rest of `packaging/pom.xml`, it seems that `${basedir}` would be a better property to use, instead of `${project-basedir}` which does not seem to exist. Could you try with that?

thadguidry:
Hmm, still a few issues...

https://github.com/OpenRefine/OpenRefine/actions/runs/4733389973/jobs/8400853802?pr=5638#step:9:4954

I'll look into it more in the morning.

packaging/THIRD-PARTY.txt
```diff
@@ -0,0 +1,111 @@
+
+Lists of 109 third-party dependencies.
+     (Apache License, Version 2.0) Titanium JSON-LD 1.1 (JRE11) (com.apicatalog:titanium-json-ld:1.3.1 - https://github.com/filip26/titanium-json-ld)
+     (Apache License, Version 2.0) jcommander (com.beust:jcommander:1.82 - https://jcommander.org)
```
wetneb, line 4: @thadguidry Thanks for working on this! Is this PR to a stage where we can remove this `THIRD-PARTY.txt` file from git? I think it would be good to keep it dynamically generated if it's not too expensive, given how quickly our dependencies evolve, no?

thadguidry: That was the intent, to generate dynamically for that reason.  I'd love the help to finish things off and merge it with the necessary changes that you see fit.  Does it all just work ok for you if you do `mvn package` as a test locally?  Because I still get issues, but then I get timeouts quite often from open sources mirrors and have to use the ones in China, so maybe it's a local issue.  The timeout changes I did helped but not fully for me.  As I did this before in the past in America, things just worked, but that was for a much smaller dependency tree than what we have.

Be my guest!

wetneb (changes requested):
I am not sure why this `oro` dependency is still marked as having a missing license. I still get this in the build log:
```
[INFO] Cleaning licensesOutputDirectory '.../packaging/target/generated-resources/licenses'
[WARNING] No license information available for: oro:oro:2.0.8
[WARNING] There were 1 download errors - check the warnings above
```
It should in theory be handled by your manual_licenses.xml, but somehow not!

thadguidry:
@wetneb I know right! Saw the same thing.  This Maven license config thing drives me mad.  Perhaps you reaching out to a Maven contributor who can test our project and find some Maven bug...is the logical next step?  Because I've burned my brain on this in the past (and my retina Lolol)!

tfmorris:
The oro dependency is eliminated by the Velocity update in Butterfly. We just need to decide what other things we want to update with it.

wetneb (approved):

PR #5646, opened 2023-02-24
Updated Maven in refine script

### LINKED ISSUE #5484: refine script fails with weird message if maven download fails
Apparently the current Maven version is 3.8.6 and when it fails to download 3.8.5 it gives the error:

```
tar: Error opening archive: Unrecognized archive format
Error: Error while expanding apache-maven-3.8.5-bin.tar.gz
```
when it feeds the 404 HTML page to `tar`

### To Reproduce
1. Uninstall `mvn`
2. Run `refine` script

### Current Results
```
$ ./refine test
You have 1352M of free memory.
Your current configuration is set to use 1024M of memory.
https://docs.openrefine.org/manual/installing#increasing-memory-allocation
tar: Error opening archive: Unrecognized archive format
Error: Error while expanding apache-maven-3.8.5-bin.tar.gz
```

### Expected Behavior

Ideally, it figures out the current version. At a minimum, say that Maven is a requirement and tell the user to install it.

### Versions
 - Operating System: Mac OS
 - OpenRefine: HEAD

elebitzero:
Partially fixes #5484

- Updated `mvn_prepare` in the refine script so it doesn't do automatic download, still need to update documentation
- Updated refine.bat with a similar change.

### DIFF
--- a/refine
+++ b/refine
@@ -306,6 +306,7 @@ tools_prepare() {
 }
 
 mvn_prepare() {
+    tools_prepare
     if [ "$MAVEN_HOME" ] ; then
         MVN="$MAVEN_HOME/bin/mvn"
     elif [ "$M2_HOME" ] ; then
--- a/refine
+++ b/refine
@@ -306,7 +306,7 @@ tools_prepare() {
 }
 
 mvn_prepare() {
- if [ "$MAVEN_HOME" ] ; then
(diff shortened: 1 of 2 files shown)


wetneb (approved):
Thanks for tackling this! Tested on Linux and works as expected.

PR #5650, opened 2023-02-26
Fix a bug of object array convertion and adjust error message

### LINKED ISSUE #5463: replaceEach function not working and mis-documented
At https://docs.openrefine.org/manual/grelfunctions#replaceeach-s-a-find-a-replace in the GREL manual, the documentation describes a function replaceEach with syntax

replaceEach(s, a find, a replace)

But the example in the subsequent description uses a function called replaceEachFromArrays, and if you try this example you do indeed get an Unknown function error. However, using replaceEach with the syntax described also always produces an error

_Error: replaceEach expects 3 arguments: string, array of strings to replace, array of replacement strings_

even when using the correct argument types. Try, for instance,

`"The cow jumps over the moon and moos".replaceEach(["th", "moo"], ["ex", "mee"])`

I am using OpenRefine Version 3.6.2 [579a6f7] in Safari Version 16.1 on MacOS 12.6.1 .

Huishin-pie:
Fixes #5463 

- Fix a bug of object array convertion
- Adjust error message

### DIFF
--- a/main/src/com/google/refine/expr/functions/strings/ReplaceEach.java
+++ b/main/src/com/google/refine/expr/functions/strings/ReplaceEach.java
@@ -6,7 +6,7 @@ import com.google.refine.grel.ControlFunctionRegistry;
 import com.google.refine.grel.EvalErrorMessage;
 import com.google.refine.grel.Function;
 import com.google.refine.grel.FunctionDescription;
-import org.apache.commons.lang.StringUtils;
+import com.google.refine.util.StringUtils;
 
 import java.util.Arrays;
 import java.util.Properties;
@@ -29,7 +29,7 @@ public class ReplaceEach implements Function {
                 String[] searchArgs = new String[((Object[]) args[1]).length];
                 Object[] originalSearchArgs = (Object[]) args[1];
                 for (var i = 0; i < searchArgs.length; i++) {
-                    searchArgs[i] = originalSearchArgs[i].toString();
+                    searchArgs[i] = StringUtils.toString(originalSearchArgs[i]);
                 }
                 String[] search = searchArgs, replace;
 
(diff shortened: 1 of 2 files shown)


main/src/com/google/refine/expr/functions/strings/ReplaceEach.java
```diff
@@ -45,7 +55,7 @@ public Object call(Properties bindings, Object[] args) {
                     }
 
                     // replace each occurrence of search with corresponding element in replace
-                    return StringUtils.replaceEachRepeatedly(str, search, replace);
+                    return StringUtils.replaceEachRepeatedly((String) args[0], search, replace);
```
tfmorris, line 58: This is going to collide with `com.google.refine.util.StringUtils`, so let's switch to the fully qualified name here, since it's the only occurrence.
```suggestion
                    return org.apache.commons.lang.StringUtils.replaceEachRepeatedly((String) args[0], search, replace);
```

main/src/com/google/refine/expr/functions/strings/ReplaceEach.java
```diff
@@ -25,13 +25,23 @@ public class ReplaceEach implements Function {
...
+                String[] search = searchArgs, replace;
+
                 if (args[2] instanceof String) {
                     replace = new String[] { (String) args[2] };
                 } else {
-                    replace = (String[]) args[2];
+                    String[] replaceArgs = new String[((Object[]) args[2]).length];
+                    Object[] originalReplaceArgs = (Object[]) args[2];
+                    for (var i = 0; i < replaceArgs.length; i++) {
+                        replaceArgs[i] = originalReplaceArgs[i].toString();
```
tfmorris, line 42: For consistency in string conversion, please use our utility method from `com.google.refine.util.StringUtils` to handle this (you'll need to adjust the imports):
```suggestion
                        replaceArgs[i] = StringUtils.toString(originalReplaceArgs[i]);
```

tfmorris (changes requested):
Looks really good. We have a utility method for `toString()` conversions, so let's use that for consistency.

Also, please add test coverage for some additional code paths:
- replacement array shorter than search array (to test array extension/fill)
- non-string (e.g. int, float, null) search/replace values (to test the toString() conversion)

Other than that it looks ready to go

tfmorris (approved):
Looks great. Thank you!

PR #5655, opened 2023-03-01
Refactor row count handling for ProjectMetadata. Remove from GetRows.

### LINKED ISSUE #1380: Update "Last modified" date only when project is actually modified and not when it is just opened

### LINKED ISSUE #3475: Row count is returned as 0 occasionally from the server while fetching metadata
While writing tests for project_metadata #3474 , I encountered an occasional bug. Table row metadata is being returned as zero sometimes from the openrefine server and hence Rows is being shown as 0.
### To Reproduce
1. Visit the `Open Project` page
2. See table columns, row is sometimes returned as 0 from the server

### Response returned by the server

```json
{
    "projects": {
        "2610263550897": {
            "created": "2021-01-15T17:30:02Z",
            "modified": "2021-01-15T17:30:02Z",
            "name": "1610731802333",
            "tags": [],
            "creator": "",
            "contributors": "",
            "subject": "",
            "description": "",
            "rowCount": 0,
            "title": "",
            "version": "",
            "license": "",
            "homepage": "",
            "image": "",
            "importOptionMetadata": [
                {
                    "includeFileSources": false,
                    "includeArchiveFileName": false,
                    "encoding": "US-ASCII",
                    "ignoreLines": -1,
                    "headerLines": 1,
                    "skipDataLines": 0,
                    "storeBlankRows": true,
                    "storeBlankCellsAsNulls": true,
                    "separator": ",",
                    "guessCellValueTypes": false,
                    "processQuotes": true,
                    "quoteCharacter": "\"",
                    "trimStrings": true,
                    "projectName": "1610731802333",
                    "fileSource": "food.mini.csv",
                    "archiveFileName": null
                }
            ],
            "customMetadata": {}
        }
    }
}
```
Screenshot 2021-01-16 at 1 00 52 AM

Meanwhile, when I click on the project, it shows two rows.
Screenshot 2021-01-16 at 1 01 01 AM

### LINKED ISSUE #5387: Rows are counted only after the project is opened
### To Reproduce
1. First, upload a file via API
2. Then, open the projects' list

### Current Results
`Rows` column shows 0 rows.

### Expected Behavior
Rows should be displayed.

### Versions
 - Operating System: Linux
 - Browser Version: Firefox
 - OpenRefine: 3.6.2 and 3.6.1

### Additional context
Opening the project link updates the `rows` column.

Workaround: GET request toward `/command/core/get-rows?project=PROJECT&start=0&limit=0`.

tfmorris:
Fixes #1380 Fixes #5387 Probably fixes #3475

- refactor row counter maintenance in ProjectMetadata so it's done when row list is updated rather than on every GetRows command. (#1380)
- Update row count when project is created/registered (#5387)
- Update row count for when opening legacy projects which were created before row count was stored in metadata (but do it in a way that doesn't change the Modified date for the project)

### DIFF
--- a/main/src/com/google/refine/ProjectManager.java
+++ b/main/src/com/google/refine/ProjectManager.java
@@ -130,10 +130,12 @@ public abstract class ProjectManager {
      */
     public void registerProject(Project project, ProjectMetadata projectMetadata) {
         synchronized (this) {
+            // Row count is duplicated in metadata, so make sure it is up-to-date
+            projectMetadata.setRowCount(project.rows.size());
             _projects.put(project.id, project);
             _projectsMetadata.put(project.id, projectMetadata);
             if (_projectsTags == null)
-                _projectsTags = new HashMap<String, Integer>();
+                _projectsTags = new HashMap<>();
             String[] tags = projectMetadata.getTags();
             if (tags != null) {
                 for (String tag : tags) {
--- a/main/src/com/google/refine/ProjectMetadata.java
+++ b/main/src/com/google/refine/ProjectMetadata.java
@@ -328,8 +328,16 @@ public class ProjectMetadata {
 
     @JsonIgnore
     public void setRowCount(int rowCount) {
-        this._rowCount = rowCount;
+        setRowCountInternal(rowCount);
         updateModified();
+    }
+
+    /**
+     * Set row count without updating the last modified time. Internal use only!
+     */
+    @JsonIgnore
+    public void setRowCountInternal(int rowCount) {
+        this._rowCount = rowCount;
     }
 
     @JsonIgnore
@@ -349,6 +357,7 @@ public class ProjectMetadata {
     @JsonIgnore
     public void setUserMetadata(ArrayNode userMetadata) {
         this._userMetadata = userMetadata;
+        updateModified();
     }
 
     private void updateUserMetadata(String metaName, String valueString) {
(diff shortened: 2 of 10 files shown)


ostephens:
>add a ToDo to GetRows command where setRowCount is being called in the first place (which seems totally wrong)

Not sure if this is related, but just noted this issue where the row count is reported incorrectly as zero #3475

tfmorris:
@ostephens Thanks for the pointer. I added a comment which describes ways that failure mode could be triggered. The refactoring for #1380 and  #5387 will likely cure #3475 as well.

wetneb (approved):
Thanks a lot! It looks like there are test failures which are likely not spurious and worth fixing, but the approach looks good to me.

tfmorris:
> It looks like there are test failures which are likely not spurious and worth fixing

Yes, there's a test that knows about internal implementation details and is testing those instead of the external behavior. :(
I didn't change the approach, but adjusted it to track the new internals.

PR #5673, opened 2023-03-10
Refine bat consistency cleanup

### LINKED ISSUE #5359: Windows refine.bat needs to be updated for new arguments matching Linux refine script

elebitzero:
Fixes #5359

- Moved functionality of `/d` to `/debug` and `/x` to `/jmx` for refine.bat.
- Implemented `/d`, `/v` and `/x` for refine.bat.
- Cleaned up some of the argument parsing to be more clear and concise.
- Changed the usage statement to reflect the changes.

### DIFF
--- a/refine.bat
+++ b/refine.bat
@@ -79,56 +79,20 @@ set OPTS=
 rem --- Argument parsing --------------------------------------------
 
 :loop
-if ""%1"" == """" goto readIniFile
-if ""%1"" == ""/?"" goto usage
-if ""%1"" == ""/h"" goto usage
-if ""%1"" == ""/p"" goto arg-p
-if ""%1"" == ""/i"" goto arg-i
-if ""%1"" == ""/H"" goto arg-H
-if ""%1"" == ""/w"" goto arg-w
-if ""%1"" == ""/d"" goto arg-d
-if ""%1"" == ""/m"" goto arg-m
-if ""%1"" == ""/x"" goto arg-x
-if ""%1"" == ""/c"" goto arg-c
-goto readIniFile
-
-:arg-p
-set REFINE_PORT=%2
-goto shift2loop
-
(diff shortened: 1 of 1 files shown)


wetneb (approved):
There are some "breaking" changes in there but I think uniformity is a worthy goal. We will need a corresponding documentation update.

I have not tested this.

ostephens:
Great - sorry this was only linked from the owning issue and I didn't track back to there. All looks good to me

PR #5674, opened 2023-03-10
Issue #5388 fix case in sql column names

### LINKED ISSUE #5388: SQL export does not put quote around column names
### To Reproduce
1. Export project as SQL

### Current Results
Column names are not quoted. If a column is named `Plot`, `Plot VARCHAR(255) NULL,` will be printed. Creating such table in Postgres would result in a lowercase `plot` column, if the column name is not quoted.

### Expected Behavior
Column names should be quoted always or when needed.

### Versions
 - Operating System: Linux
 - Browser Version: Firefox
 - OpenRefine: 3.6.2

davidcardd1:
Fixes #5388 

- Added function to wrap a string with double quotes
- Added said function in SQLExporter and SQLInsertBuilder

### DIFF
--- a/main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
+++ b/main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
@@ -54,7 +54,7 @@ public class SqlCreateBuilder {
     }
 
     public static String addQuotes(String columnName) {
-        return "\"" + columnName + "\" ";
+        return "\"" + columnName.replace("\"", "\"\"") + "\"";
     }
 
     public String getCreateSQL() {
@@ -86,9 +86,9 @@ public class SqlCreateBuilder {
                 if (name != null) {
                     if (trimColNames) {
                         String trimmedCol = name.replaceAll("[^a-zA-Z0-9_]", "_");
-                        createSB.append(addQuotes(trimmedCol));
+                        createSB.append(addQuotes(trimmedCol) + " ");
                     } else {
-                        createSB.append(addQuotes(name));
+                        createSB.append(addQuotes(name) + " ");
                     }
 
                     if (type.equals(SqlData.SQL_TYPE_VARCHAR)) {
--- a/main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
+++ b/main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
@@ -53,6 +53,10 @@ public class SqlCreateBuilder {
 
     }
 
+    public static String addQuotes(String columnName) {
+        return "\"" + columnName + "\" ";
+    }
+
     public String getCreateSQL() {
         if (logger.isDebugEnabled()) {
             logger.debug("Create SQL with columns: {}", columns);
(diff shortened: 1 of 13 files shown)


main/src/com/google/refine/exporters/sql/SqlInsertBuilder.java
```diff
@@ -180,9 +180,9 @@ public String getInsertSQL() {
         }
 
         boolean trimColNames = options == null ? false : JSONUtilities.getBoolean(options, "trimColumnNames", false);
-        String colNamesWithSep = columns.stream().map(col -> col.replaceAll("[^a-zA-Z0-9_]", "_")).collect(Collectors.joining(","));
+        String colNamesWithSep = columns.stream().map(col -> SqlCreateBuilder.addQuotes(col.replaceAll("[^a-zA-Z0-9_]", "_"))).collect(Collectors.joining(","));
         if (!trimColNames) {
-            colNamesWithSep = columns.stream().collect(Collectors.joining(","));
+            colNamesWithSep = columns.stream().map(col -> SqlCreateBuilder.addQuotes(col)).collect(Collectors.joining(","));
```
wetneb, line 185: It looks like in this code path, the column names could contain quotes themselves, so I do not think it would be safe given that `addQuotes` does not do any escaping of its input.

main/src/com/google/refine/exporters/sql/SqlCreateBuilder.java
```diff
@@ -53,6 +53,10 @@ public SqlCreateBuilder(String table, List<String> columns, JsonNode sqlOptions)
 
     }
 
+    public static String addQuotes(String columnName) {
+        return "\"" + columnName + "\" ";
```
wetneb, line 57: It feels a bit arbitrary to have this method return an extra space at the end. If you want to introduce a dedicated method for this, I would argue it should:
* do the quoting properly (meaning that any quote in the column name would be escaped)
* not return a space at the end: the caller is responsible for adding it on their side if they are needed

davidcardd1: In the commit below (https://github.com/OpenRefine/OpenRefine/pull/5674/commits/54a76f0fc9a55fee7a840e16747130873692ffaf) : 

- I added escaping for double quotes. I didn't add escaping for single quotes because they're already working as expected (if I add escaping like this: `...replace("'", "''")` it exports two single quotes). 
- Removed space in the function and added it to each caller when needed

davidcardd1:
I've just noticed these two commits: https://github.com/OpenRefine/OpenRefine/pull/5674/commits/2e5731b5639c35e77f31ba34a3f697027f9ce0d5 and https://github.com/OpenRefine/OpenRefine/pull/5674/commits/a97248a9f64ec305cf76da0d85ad50ff06cddcab .
I think they were autogenerated when I did my other two commits. Is this right? if not, is there a way I can remove them?

wetneb (approved):
The details of your commits do not matter too much since we can squash them when merging, which I will do.

PR #5676, opened 2023-03-10
5449-add-item-button

### LINKED ISSUE #5449: UI fix: improve position of add media / add item button in Schema tab
This issue is transferred from a comment in #5050, so it's easier to track separately.

The UI fix in question is to move the "add item" / "add media" button to the left of the schema building window and thus make it more prominent. See screenshots below (context-specific to Wikibase chosen): 

Screenshot 2022-07-23 at 22 37 41

Screenshot 2022-07-23 at 20 36 44

Screenshot 2022-07-23 at 20 36 47

Complete wireframes (including development parameters are here): https://xd.adobe.com/view/3ae13acb-7afb-4ea5-8252-351a1622abba-6fb2/

_Originally posted by @lozanaross in https://github.com/OpenRefine/OpenRefine/issues/5050#issuecomment-1193187217_

prashasti-7:
Fixes #5449 

- Left-aligned 'Add Item' button
- Right-aligned 'Saved Schema' and 'Delete Changes' buttons
- Border between the Add item button and added items (as in the screenshots)
- Increased width of the div box to cover the whole width (as in the screenshots)
- Side border to the 'type file name or drag reconciled column [Let me know if that needs to be reversed]
- Remove, add term and other buttons aligned


extensions/wikibase/module/styles/schema-alignment.css
```diff
@@ -49,12 +49,12 @@
...
 
 .schema-alignment-dialog-entity-edits-area {
-  max-width: 70em;
+  max-width: 100%;
 }
 
 .schema-alignment-dialog-statements-area {
   padding: 10px;
-  max-width: 900px;
+  max-width: 100%;
```
wetneb, line 57: Why make this change?
The intention with capping the width of this element is to match the behaviour of Wikidata (where statements' width is also capped, see [this example on a wide screen](https://www.wikidata.org/wiki/Q69225951)).
I think it makes sense to cap the width because otherwise the buttons end up being very far from the content they relate to, so it makes it harder to follow which button corresponds to which statement/qualifier/reference.

extensions/wikibase/module/styles/schema-alignment.css
```diff
@@ -66,16 +66,31 @@
   margin: 1em;
 }
 
-.schema-alignment-save {
+.schema-alignment-save-a {
```
wetneb, line 69: This class name is not very clear - I am not sure what the `-a` suffix refers to.

extensions/wikibase/module/styles/schema-alignment.css
```diff
@@ -183,6 +198,7 @@
   margin: 5px;
   white-space: normal;
   word-wrap: break-word;
+  padding: 10px;
```
wetneb, line 201: This sort of padding/margin increase seems unrelated to the original issue, no? What was your motivation to make this change?

extensions/wikibase/module/styles/schema-alignment.css
```diff
@@ -268,7 +284,7 @@ input[type="checkbox"].wbs-wikitext-override {
   display: inline-block;
 }
 
-.wbs-entity-contents {
+.wbs-entity-contents, .wbs-entity-input {
```
wetneb, line 287: I'd keep this out indeed, as it seems unrelated to the issue.

wetneb (commented):
Thank you @prashasti-7 for this first contribution!
It seems that you have included quite some changes in this PR, even some which were not suggested in the original issue. I am unsure about the motivation behind those - could you perhaps elaborate? I am adding a few examples below. It seems that you have generally tried to increase margins and padding around various elements - I would rather keep that for a separate PR, to be proposed in another issue first. It might make sense, but it probably comes at a cost for people working on smaller screens.

extensions/wikibase/module/styles/schema-alignment.css
prashasti-7: I understand your concern, but the image provided by Lozana in the issue #5449 (https://xd.adobe.com/view/3ae13acb-7afb-4ea5-8252-351a1622abba-6fb2/) had the buttons in the very corner of the screen, due to which I got the impression that the buttons had to be pushed to the end of the screen.
Please let me know if it needs to be reverted back to its original width.

wetneb: Intuitively the position of those buttons should be independent from the maximum width of the statements. You might need to change the DOM structure so that they are outside of this statements area.

extensions/wikibase/module/styles/schema-alignment.css
prashasti-7: Originally, the buttons 'Save schema' and 'Discard changes' were placed one below the other. But to make them horizontally aligned, I created another class for each: ' schema-alignment-save-a' and 'schema-alignment-save-b', both floating right.
This way, I found 'Save schema' and 'Discard changes' were horizontally aligned towards the right in the hierarchy mentioned in the issue (XD image).
Is there a better and more efficient method to handle this?

extensions/wikibase/module/styles/schema-alignment.css
prashasti-7: Okay sure I'll change it back to its original position

extensions/wikibase/module/styles/schema-alignment.css
prashasti-7: In the original code, the padding in the Terms and Statements input boxes were not equal,
image

 Which is why I took the liberty of adding padding to the Statement input box.
But yes, I agree it is unrelated to the original issue. I'll remove that line

extensions/wikibase/module/styles/schema-alignment.css
wetneb: How about one div containing both buttons, positioned in the right place, and then the buttons inside it one after the other?

extensions/wikibase/module/styles/schema-alignment.css
wetneb: Ok, it's a sensible move indeed but let's keep it out this time for the sake of having a clearer PR

extensions/wikibase/module/styles/schema-alignment.css
prashasti-7: Okay got it

extensions/wikibase/module/styles/schema-alignment.css
prashasti-7: Is this what you mean?
image

prashasti-7:
I have made some changes in the code. Can you check if this is better?

extensions/wikibase/module/styles/schema-alignment.css
wetneb: Yes! I assume you might want some class name on the containing `div` so that you can position it the right way. Also, `schema-alignment-save` seems like it should not apply to the `discardButton`, but that's a detail.

prashasti-7: The recent commit seems to be working. 
Outer div is called `schema-alignment-save-buttons` which consists of both the buttons 'save schema' and 'discard changes'.

extensions/wikibase/module/styles/schema-alignment.css
wetneb: On my side, the schema still expands to the full width of the screen:
image

wetneb:
Very sorry about the time it took me to get back to this (feel free to ping us in those cases)!

I have added a comment above about the width of the schema, which still seems to be expanding to the full screen width in my case.

Another small thing I have noticed is that the new light blue line you have introduced is thicker than the blue lines which contain the list of columns, above the schema. I would follow the wireframes instead, and keep those with the same thickness.

Otherwise this seems good to go on my side. Thanks a lot for your patience!

extensions/wikibase/module/styles/schema-alignment.css
prashasti-7: Should I change the width back to its original length? I thought the schema should expand to the entire width of the screen.

wetneb: As I wrote above, I think it makes sense to cap the width because otherwise the buttons end up being very far from the content they relate to, so it makes it harder to follow which button corresponds to which statement/qualifier/reference. So yes, I would restore this back to its original length.

wetneb (approved):

PR #5682, opened 2023-03-13
migrate home-view layout-logic to CSS

Abbe98:


elebitzero (commented):
It's good to let the browser handle the resize as much as possible, but I have the following comments.

1. I don't follow the warning "resize() is deprecated, use responsive CSS instead."

The `resize()` function is still defined and called on the 'resize' event for index.js
And the core `CreateProjectUI` and `OpenProjectUI ` define `ActionAreaUI.resize()` handlers:

https://github.com/OpenRefine/OpenRefine/blob/608e2f8fccb161ca03ab224e5c57a6c594cfd46a/main/webapp/modules/core/scripts/index/create-project-ui.js#L72
https://github.com/OpenRefine/OpenRefine/blob/89e6dbe2fe8390a55065e997ff657b89ef065d5f/main/webapp/modules/core/scripts/index/open-project-ui.js#L77

Also for the project.js, a separate resize handler is defined:

https://github.com/OpenRefine/OpenRefine/blob/9703e4e6c7ed150f029c762c92ecd90d05cf5f4d/main/webapp/modules/core/scripts/project.js#L166

which calls resize handlers in:
https://github.com/OpenRefine/OpenRefine/blob/89e6dbe2fe8390a55065e997ff657b89ef065d5f/main/webapp/modules/core/scripts/project/browsing-engine.js#L70
https://github.com/OpenRefine/OpenRefine/blob/eadbbf2af92975d864354606066ae227f05fc297/main/webapp/modules/core/scripts/project/history-panel.js#L40
https://github.com/OpenRefine/OpenRefine/blob/c5bca0e3b9d7c735c0e0a805f12e8f00f4735e7a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js#L76

2. Regardless of the warning's wording, it should be removed because it is logging hundreds of times while resizing, which is too much.

3. I tested the changes and noticed one difference:
On Chrome/Brave/Edge, the 'Create Project' panel looks different on the first page load vs after the 1st resize. On Firefox, I don't see the issue. If I compare to master, I do not see this issue on any of the four browsers.

FIRST PAGE LOAD
image
AFTER 1st RESIZE
image

Abbe98:
> The resize() function is still defined and called on the 'resize' event for index.js
And the core CreateProjectUI and OpenProjectUI define ActionAreaUI.resize() handlers

Good point, I somehow thought I had dealt with those, but it seems like I only tackled empty ones.

> Regardless of the warning's wording, it should be removed because it is logging hundreds of times while resizing, which is too much.

Agreed, especially given the current usage. 

> On Chrome/Brave/Edge, the 'Create Project' panel looks different on the first page load vs after the 1st resize. On Firefox, I don't see the issue. If I compare to master, I do not see this issue on any of the four browsers.

Good catch!

elebitzero (approved):
The changes look good but resize is still in open-project-ui.js and create-project-ui.js:

https://github.com/OpenRefine/OpenRefine/blob/711818b206e7b9c7c0ad64db2f7bfacb20bf152d/main/webapp/modules/core/scripts/index/create-project-ui.js#L72

https://github.com/OpenRefine/OpenRefine/blob/711818b206e7b9c7c0ad64db2f7bfacb20bf152d/main/webapp/modules/core/scripts/index/open-project-ui.js#L77

Removing the contents of `resize()` in create-project-ui.js appears to work, but to do the same in open-project-ui.js the `workspaceControls` CSS would have to be moved to the CSS file since the 'Browse workspace directories' button appears on top of the table after removing the contents of `resize()`.

Abbe98:
> Removing the contents of resize() in create-project-ui.js appears to work, but to do the same in open-project-ui.js the workspaceControls CSS would have to be moved to the CSS file since the 'Browse workspace directories' button appears on top of the table after removing the contents of resize().

Yeah, I dropped the "deprecated" statement for the commit for now.

PR #5685, opened 2023-03-14
fix the tab order for reconciliation create and match buttons

Abbe98:
This fixes the tab order for the two reconciliation create buttons:

[Screencast from 2023-03-14 16-16-09.webm](https://user-images.githubusercontent.com/2631719/225048095-c2e2e8bd-41c5-4dae-ad54-a82f61bbe0e9.webm)

Previously the order would be reverse, as in the right button being focused before the left one.


elebitzero (commented):
You need to fix here also:
https://github.com/OpenRefine/OpenRefine/blob/0fa6dd505bee1139f32e36b8da06b36744d0e8bc/main/webapp/modules/core/scripts/views/data-table/cell-renderers/recon-renderer.js#L57

Right now the tab order is only fixed for the 'Create new item' row:

recon-button-tab-order

elebitzero (approved):

PR #5686, opened 2023-03-14
remove deprecated TestUtils.assertEqualAsJson

Abbe98:
Deprecated since 3.5.


extensions/wikibase/tests/src/org/openrefine/wikibase/commands/LoginCommandTest.java
```diff
@@ -97,38 +97,38 @@ public void setUp() throws Exception {
     public void testNoApiEndpointPost() throws ServletException, IOException {
         when(request.getParameter("csrf_token")).thenReturn(Command.csrfFactory.getFreshToken());
         command.doPost(request, response);
-        assertEqualAsJson("{\"code\":\"error\",\"message\":\"missing parameter 'wb-api-endpoint'\"}", writer.toString());
+        assertEqualsAsJson("{\"code\":\"error\",\"message\":\"missing parameter 'wb-api-endpoint'\"}", writer.toString());
```
tfmorris, line 100:
```suggestion
        assertEqualsAsJson(writer.toString(), "{\"code\":\"error\",\"message\":\"missing parameter 'wb-api-endpoint'\"}");
```

The parameter order is different between the two versions of the functions, so these need to be swapped. Ditto for all the occurrences.

tfmorris (changes requested):

extensions/wikibase/tests/src/org/openrefine/wikibase/commands/LoginCommandTest.java
wetneb: It's a change that should be relatively easy to do with the "Inlining" refactoring method of IDEs like Eclipse or IntelliJ by the way.

Abbe98: Ah, I thought the refactoring took care of that but of course the tests pass anyway! Nice catch.

tfmorris (changes requested):
Looks like one swap got missed. Otherwise looks good.

tfmorris (approved):
LGTM

As a suggestion for the future, if you make a separate commit addressing review comments it makes it easier for the reviewer to see what was changed. You can then squash it away when merging.

In this case, the PR is small enough that it's not a huge burden to re-review everything on the force push, but it's still unnecessary extra work.

Thanks for sticking with this to the end.

Abbe98:
> As a suggestion for the future, if you make a separate commit addressing review comments it makes it easier for the reviewer to see what was changed. You can then squash it away when merging.

@tfmorris  you should still be able to compare the changes individually, the "compare" button next to the force-push message compares only the changes for that forced push:

Screenshot 2023-03-15 at 19-48-12 remove deprecated TestUtils assertEqualAsJson by Abbe98 · Pull Request #5686 · OpenRefine_OpenRefine

wetneb (approved):

PR #5691, opened 2023-03-15
Documentation for toDate()

### LINKED ISSUE #5351: Documentation missing for toDate(locale,format)
When the enhancement to `toDate()` requested in #729 was added in fa072df85c27ce601f867a7f39c5a5d1bf288dec 9 years ago (!), no documentation was included. The online help and the documentation should be updated to reflect the current code.

The test cases show how the list of format strings can be prepended with a string identifying the locale which should be used for the parsing attempts:

https://github.com/OpenRefine/OpenRefine/blob/fa072df85c27ce601f867a7f39c5a5d1bf288dec/main/tests/server/src/com/google/refine/tests/expr/functions/strings/ToFromConversionTests.java#L134-L139

Docs to be updated:

https://github.com/OpenRefine/OpenRefine/blob/d9c235b5b4f763826ab6e4ff1b1d946389286530/main/resources/com/google/refine/grel/FunctionDescription.properties#L123

https://github.com/OpenRefine/OpenRefine/blob/be2d14e72158dfc1c93dce098f97ed59a7eb7dc8/docs/docs/manual/grelfunctions.md#date-functions-date-functions

amyra98:
Fixes #5351

- Updated toDate() properties
- Properties of toDate(locale,format)


main/resources/com/google/refine/grel/FunctionDescription.properties
```diff
@@ -120,7 +120,7 @@ fun_jsonize=Quotes a value as a JSON literal value.
...
-fun_to_date=Returns the inputted object converted to a date object. Without arguments, it returns the ISO 8601 extended format. With arguments, you can control the output format. With monthFirst: set false if the date is formatted with the day before the month. With formatN: attempt to parse the date using an ordered list of possible formats. Supply formats based on the SimpleDateFormat syntax: <a href="http://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html">SimpleDateFormat</a>.
+fun_to_date=Returns the date object from input string.The arguments available to the toDate function control how the inputted object is interpreted as a date.The function can also parse dates in non-default locale like we can give in the first string can be a locale identifier instead of a format string and so when a locale identifier is provided, the parsing function will use the formatting rules for that locale, rather than the default formatting rules and it will give us the output date object.
```
ostephens, line 123: toDate can take a number as an input as well so the first sentence
>Returns the date object from input string

should reflect this (e.g. "input string or number")

There should be a space after each full stop

I think the third sentence is quite long - I'd suggest either trying to re-phrase more concisely or splitting into two sentences


Also suggest we avoid use of a 

I'd suggest it could also break into two sentences so for example:
"also parse dates in non-default locale. For example we can give in the"

amyra98:
Please review @ostephens

ostephens:
Thanks @amyra98. I see you've removed the information about the specific `monthFirst` and `format` parameters. Remembering that this is documentation that the user sees when they look at the GREL function help inside the application (as opposed to looking at the full documentation on https://openrefine.org/docs/) then I think we need to either keep this, or add a link to the full documentation (https://openrefine.org/docs/manual/grelfunctions#todateo-b-monthfirst-s-format1-s-format2-) so a user can actually work out how to use the function

ostephens:
You may also need to update the documentation on the official docs site to make sure all the information is correct (I think some of the corrections you've made here also need making in the docs)

amyra98:
I have added the use of monthFirst and format parameters. For changes in official docs I will open an issue in openrefine.org 's repository and do the required changes. @ostephens

wetneb (commented):
I think this could be a bit more concise and go to the point faster. For instance, the phrase `The arguments availabe to the toDate function` is quite lengthy and that is not needed: in this context, the reader already knows this is about the `toDate` function.
This string should stay relatively short (given where it is displayed in the UI) and let the user understand as easily as possible which combinations of arguments are allowed.

PR #5692, opened 2023-03-15
replace image sprite in cell edit button with HTML and CSS

Abbe98:
To make it:
1. stylable
2. translatable
3. "not blurry" 🧐

Part of #3017

Button style after the changes:
[Screencast from 2023-03-15 19-28-05.webm](https://user-images.githubusercontent.com/2631719/225407617-3c2ee908-b785-4ef5-8668-1e959991a36e.webm)

Minor note: the color when not hovering the button itself is slightly darker then before, the old one appeared to be a unique shade so I took the liberty to consolidate it.


elebitzero (commented):
1) The edit button doesn't look vertically centered:
image

Reducing the vertical padding by 1px on the top and bottom seems to fix that.

padding: 2px 3px 2px 3px;

I like the slightly taller button since it is easier to click, but it should be vertically aligned.  I looked at the HTML and the problem is that while it has top=0, the button is contained in a div which is not flush with the top of the td.

2) There is not much difference between the hover and not hover color of the button.

I prefer sticking with the existing colors for now, or you could adopt the style of the recon buttons and show as blue text in a white box and then when you hover over it show the white text in a blue box.

hover-edit-button

Abbe98:
> The edit button doesn't look vertically centered:

> I looked at the HTML and the problem is that while it has top=0, the button is contained in a div which is not flush with the top of the td.

Hm, I will see if I can fix the later to avoid a pixel-based solution.


>There is not much difference between the hover and not hover color of the button.
>
>I prefer sticking with the existing colors for now, or you could adopt the style of the recon buttons and show as blue text in a white box and then when you hover over it show the white text in a blue box.

I have mixed feelings considering the contrast of the current button, but postponing any color change until I reach the  
recon buttons so it can look consistent is fine with me.

elebitzero (commented):
I don't see any change to the vertical alignment of the edit button. It looks the same as the picture I posted above.  Did you try changing the padding to "2px 3px 2px 3px"?

I see the hover and focus background color of the button is now close to what it was before which seems okay.

Abbe98:
> I don't see any change to the vertical alignment of the edit button. It looks the same as the picture I posted above. Did you try changing the padding to "2px 3px 2px 3px"?

@elebitzero do you mean vertical alignment inside the cell rather than of the button text?

elebitzero:
> @elebitzero do you mean vertical alignment inside the cell rather than of the button text?

@Abbe98 yes.

Abbe98:
I have reduced the padding as suggested so it does not overflow like that in single-line cells, but it wont make it aligned.

elebitzero (approved):

PR #5699, opened 2023-03-16
Increase dynamic range of multinomial() function - fixes #5695

### LINKED ISSUE #5695: Fix code scanning alert in Multinomial - Implicit narrowing conversion in compound assignment
Tracking issue for:
- [x] https://github.com/OpenRefine/OpenRefine/security/code-scanning/43

Use `long`s for calculations instead of `int`s.

https://github.com/OpenRefine/OpenRefine/blob/90efb305f8ede391d9f1e59c20bb80ef4a10b466/main/src/com/google/refine/expr/functions/math/Multinomial.java#L51-L60

tfmorris:
Fixes #5695

Resolves CodeQL warning about narrowing conversion and uses BigInteger for intermediary values to avoid overflow unless absolutely necessary. Allows sum of terms to exceed 20 in some cases. (20! is limit of Java long)

### DIFF
--- a/main/src/com/google/refine/expr/functions/math/FactN.java
+++ b/main/src/com/google/refine/expr/functions/math/FactN.java
@@ -61,7 +61,7 @@ public class FactN implements Function {
 
     }
 
-    /*
+    /**
      * Calculates the factorial of an integer, i, for a decreasing step of n. e.g. A double factorial would have a step
      * of 2. Returns 1 for zero and negative integers.
      */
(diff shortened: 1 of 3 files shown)


wetneb (approved):
I'd be curious to know if anyone has ever used that GREL function and if so for what purpose. But it probably does not hurt to fix this warning.

tfmorris:
From my point of view, the less noise in code quality checks, the more likely people are to develop critical problems when they show up, plus more tests is better (neither `fact()` or `multinomial()` had any tests at all).

13! will overflow a Java int, so the previous implementation was extremely limited.

PR #5705, opened 2023-03-18
Updated the cross function to return an empty array when there is no …

### LINKED ISSUE #5531: cross function produces null when there is no match
I notice that the result of the `cross()` function is not an array when there is no match found - instead, the result is `null`. This seems to contradict the reference documentation which states "Returns an array of **zero** or more rows..." (that is, we should expect an an empty array, `[]`, when there are no matching rows). So if, for instance, you are wanting a count of the number of matches using `cross(...).length()` (which is something I do) then you currently get "Error: length expects an array or a string" where you expect to get 0.

Incidentally, if the function gets fixed to produce an empty array, as per the documentation, then the result of a common operation where you just want to retrieve the first matching row (i.e. `cross(...)[0]`) will, when there is no match, change to "Error: java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0" - currently the result is just null (since `null[0]` is also null). So I'm wondering if the null return was implemented as a specific exception to avoid this ungainly instance. Should we consider making the result of indexing an empty array (e.g. `[][0]`) null instead of producing the error?

I'm looking at OpenRefine 3.6.2 [579a6f7].

jenny-Musah:
…matching rows

Fixes #5531

- made the cross function return and empty row [ ] when there are not matching rows just as the documentation stated
- 
-

### DIFF
--- a/main/src/com/google/refine/LookupCacheManager.java
+++ b/main/src/com/google/refine/LookupCacheManager.java
@@ -143,14 +143,13 @@ public class LookupCacheManager {
             this.targetProjectID = targetProjectID;
             this.targetColumnName = targetColumnName;
         }
-
         public HasFieldsListImpl getRows(Object value) {
-            if (!ExpressionUtils.isNonBlankData(value)) return null;
+            HasFieldsListImpl rows = new HasFieldsListImpl();
+            if (!ExpressionUtils.isNonBlankData(value)) return rows;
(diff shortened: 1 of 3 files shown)


wetneb:
Nice! To make your changes conform to the Java style of the codebase, you can run `mvn formatter:format` and commit and push the changes.

PR #5712, opened 2023-03-21
made the cross function return an empty array when there are no match…

jenny-Musah:
made the cross function return an empty array when there are no matching rows just as the documentation stated 

Fixes #{5531}

- made the cross function return an empty array if there is  no matching rows


wetneb:
Thank you!

Your pull request includes the addition of many unrelated files. You can inspect that here:
https://github.com/OpenRefine/OpenRefine/pull/5712/files
For instance, the file [benchmark/IDEs/eclipse/README.txt](https://github.com/OpenRefine/OpenRefine/pull/5712/files#diff-e605e4004f1f915a433208cb805fff1c23c1416ca02f234dbaf6e6f61bdd8158) is unrelated to the issue you are trying to solve, so it should not be added by this pull request.

jenny-Musah:
okay, so i should make another pull request and remove those files

wetneb:
It's interesting that you had to copy all those files, it really should not be needed. Yes, in any case they should not be part of the pull request you are sending. Just the `Cross` and `CrossTests` classes should be enough.

PR #5719, opened 2023-03-22
Use wmic for getting free ram in refine.bat

elebitzero:
- Make refine.bat use `wmic` to get the available free memory instead of the inefficient `systeminfo ^| findstr /C:"Available Physical Memory"`.

WMIC (Windows Management Instrumentation Command-line) is a command-line interface for the Windows Management Instrumentation (WMI) technology. It has been available since Windows Server 2003 and Windows XP/Vista/7 and later.

On my desktop the time for 'Getting Free Ram' went from 2.5 seconds to instant.  For my laptop, it went from 3.5 seconds to almost instant.

Before:
Command prompt before 2 hidden

After:
Command prompt after 2 hidden

### DIFF
--- a/refine.bat
+++ b/refine.bat
@@ -228,14 +228,14 @@ if %JAVA_RELEASE% GTR 17 (
 )
 
 echo Getting Free Ram...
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Nice!

PR #5724, opened 2023-03-23
Made it so dialogs cannot be dragged past the top of the window

### LINKED ISSUE #5714: Dialog mover
I don't know if this problem is only on the 3.7.1 version or on other versions too but on opening the reconciliation dialogue and moving it up to view the list down below, I couldn't move it back down (mover doesn't display on the footer section) which means I can't  move it back down as screen scroller is also not showing anymore since it is a dialogue

FireShot Capture 023 - Portal rodents 19772002 scinameUUIDs csv - OpenRefine - 127 0 0 1

**Proposal**
Add mover when hovering on footers section in a dialogue and probably to other similar dialogue that maybe affected by this

elebitzero:
Fixes #5714

- Set containment property to prevent user from dragging jQuery UI dialogs above the top of the browser window.

 I originally tried `containment: 'window'` but that didn't work as expected because the dialog-container is the full width of the browser window, so it prevented the user from moving the dialog left or right. So instead of that, I am setting a containment box with top = 0 to prevent the user from dragging the dialog above the top of the screen.  The bottom, right, and left boundaries are set sufficiently large that so user's can drag in those directions without limit on 4k displays.
 
 After:
 
Dialog boundary top

### DIFF
--- a/main/webapp/modules/core/scripts/util/dialog.js
+++ b/main/webapp/modules/core/scripts/util/dialog.js
@@ -63,7 +63,7 @@ DialogSystem.showDialog = function(elmt, onCancel) {
   container.css("top", Math.round((top < 0 ) ? 5 : top) + "px");
   elmt.css("visibility", "visible");
 
-  container.draggable({ handle: '.dialog-header', cursor: 'move' });
+  container.draggable({ handle: '.dialog-header', containment: [ -4096, 0, 4096, 4096 ], cursor: 'move' });
 
   var layer = {
     overlay: overlay,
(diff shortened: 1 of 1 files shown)


Abbe98:
It do break on large screens(like ultra wide ones). I wonder if `4096` should just be replaced with an arbitrary large number? I assume it wouldn't have any bad side effects given that the default is no limits.

elebitzero:
I increased it to a larger number (32768).

Abbe98 (approved):
Works great!

PR #5729, opened 2023-03-25
Contributionayushi

### LINKED ISSUE #4722: « Auto-match candidates with high confidence » is deceitful and should not be pre-check
In the Reconciliation dialog, the checkbox « Auto-match candidates with high confidence » is pre-checked when entering the dialog.

Since the confidence from the Recon service is not always useful (it can say 100% to many items even when some should be defraded, user don't really know what the number means and exactly how it is given). Because of that, I think the checkbox should be un-cheked by default. A preference could be create to control the default value of that control.

Since this has a direct impact on data pushed to Wikidata, and I highly raise the possibility of bad data being pushed, I consider this a bug.

### To Reproduce
1. Open the reconciliation dialog

### Current Results
The checkbox « Auto-match candidates with high confidence » is pre-checked.

### Expected Behavior
The checkbox « Auto-match candidates with high confidence » is unchecked when opening the dialog.

### Screenshots
image

### Versions
 - Operating System: N/A
 - OpenRefine: v3.5.x and below.

ayushrai206:
Fixes #4722

- takes user preference if default settings of auto-matching is not desired by the user


main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -337,7 +337,6 @@ ReconStandardServicePanel.prototype.start = function() {
         identifierSpace: this._service.identifierSpace,
         schemaSpace: this._service.schemaSpace,
         type: (type) ? { id: type.id, name: type.name } : null,
-        autoMatch: JSON.parse(Refine.getPreference("ui.reconciliation.automatch", true)),
```
antoine2711: @ayushrai206 : I see you removed your change, but you alos removed all of the original code here…

ayushrai206: The original code set the default value to true..which i already did in the line i added,so i removed the previous code,also regarding the line i added is it working perfectly?is the variable i pointed the default value to correct?

Abbe98:
It looks like you are working from the same branch as your last pull request(#5722) which causes the diff to look odd. Instead you should make a new branch based of `master`, the preferred way to work with branches and PRs is described [here](https://github.com/OpenRefine/OpenRefine/blob/master/CONTRIBUTING.md#how-to-submit-prs-pull-requests-patches-and-bug-fixes).

The new changes looks good I believe, but I'm not sure about potential side effects of squash-merging a branch that have already been merged once. Give the above link a look and I'm sure we can get this merged rather quickly.

PR #5733, opened 2023-03-26
gives preference to user if they want to automatch or not

### LINKED ISSUE #4722: « Auto-match candidates with high confidence » is deceitful and should not be pre-check

ayushrai206:
Fixes #4722 

- gives preference to user if they want to automatch candidates with high confidence or not, keeping it true by default .

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
+++ b/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
@@ -87,7 +87,7 @@ ReconStandardServicePanel.prototype._constructUI = function() {
   var self = this;
   this._panel = $(DOM.loadHTML("core", "scripts/reconciliation/standard-service-panel.html")).appendTo(this._container);
   this._elmts = DOM.bind(this._panel);
-  this._elmts.automatchCheck[0]=JSON.parse(Refine.getPreference("ui.reconciliation.automatch", true));
+  this._elmts.automatchCheck[0].checked=JSON.parse(Refine.getPreference("ui.reconciliation.automatch", true));
   this._elmts.or_proc_access.html($.i18n('core-recon/access-service'));
   this._elmts.or_proc_cellType.html($.i18n('core-recon/cell-type'));
   this._elmts.or_proc_colDetail.html($.i18n('core-recon/col-detail'));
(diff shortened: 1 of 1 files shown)


main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -337,7 +337,7 @@ ReconStandardServicePanel.prototype.start = function() {
         identifierSpace: this._service.identifierSpace,
         schemaSpace: this._service.schemaSpace,
         type: (type) ? { id: type.id, name: type.name } : null,
-        autoMatch: this._elmts.automatchCheck[0].checked,
```
Abbe98, line 340: This should remain as it just gets the value of "checked" from the DOM?

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -87,7 +87,7 @@ ReconStandardServicePanel.prototype._constructUI = function() {
   var self = this;
   this._panel = $(DOM.loadHTML("core", "scripts/reconciliation/standard-service-panel.html")).appendTo(this._container);
   this._elmts = DOM.bind(this._panel);
-  
+  this._elmts.automatchCheck[0]=JSON.parse(Refine.getPreference("ui.reconciliation.automatch", true));
```
Abbe98, line 90: I expect this to still set the `checked` attribute.

Abbe98 (changes requested):
This only appears to be true by default because that is set in the HTML, meanwhile the new code does not interact with the `checked` attribute so it does not seem to have the intended effect.

Abbe98 (approved):
Nicely done!

wetneb:
done by https://github.com/OpenRefine/openrefine.org/pull/209

ayushrai206:
> Could it be possible to have an accompanying documentation update? Otherwise users are unlikely to discover it.
yes, working on it.

antoine2711:
@ayushrai206 : @Abbe98 took care of it. Check the previous comment by @wetneb.

Thank you very much for your work and congrat for this achievement.

Best Regards, Antoine

Abbe98:
Sorry! I already had a draft...

PR #5739, opened 2023-03-28
API: allow clients to set description and creator with create-project-from-upload

Abbe98:


wetneb (approved):
I think this is minimal enough to be worth merging as is.

@tfmorris I agree it's nice to aim for having each PR close an issue but I think there is a balance to strike between that ideal and keeping the contribution process not overly bureaucratic.

tfmorris:
@wetneb It wasn't an idle make work request. It was because I was trying to figure out what was going on. Perhaps if I saw the paired client side change it would be clearer, but as it is it's server only with no web app change.

wetneb:
Sorry about my hasty approval then! If I am not wrong, the `create-project-from-upload` API action mentioned in the title is one that is typically used by third-party clients such as CLIs, so that's why I was not expecting any frontend change. If I remember correctly we do not even use this command in OpenRefine's frontend (but the Cypress tests do use it).

Edit: actually there is a reference to it in our frontend but it seems to be dead code, let's open an issue about that

wetneb:
Done: https://github.com/OpenRefine/OpenRefine/issues/5747

Abbe98:
TLDR;

Yeah, this is intended for external clients, I have been having the patch downstream for a while but thought that it made sense to upstream it. For the selfish reason of having fewer patches and in case it would be useful for someone else.

tfmorris:
OK, got it. I guess, in that case, it's time for my annual reminder that there's only a single official client for the OpenRefine server's internal REST API.

The reason that the protocol is undocumented and unsupported for external use isn't just to be mean, but because it doesn't have many of the features, like versioning and testing, that would be required to support third-party clients.

tfmorris:
And, speaking of testing, since there are no tests for this, no one will know if they break it in the future. When the front end reference gets removed it'll be at risk for a zealous backend cleanup that notices it is unused.

Abbe98:
Worst case scenario we just end up making downstream a proper fork, in the meanwhile we happily contribute upstream.

PR #5740, opened 2023-03-28
Corrected the error message and doc URL when JAVA_HOME is not set in refine.bat

elebitzero:
- Change doc URL from  http://bit.ly/1c2gkR to https://openrefine.org/docs/manual/installing#java.
- Don't print "Type 'refine /h' for usage." in this case.
- Also add a check to see if $JAVA is executable.

### DIFF
--- a/refine.bat
+++ b/refine.bat
@@ -207,6 +207,7 @@ set JAVA="%JAVA_HOME%\bin\java.exe"
 if not exist !JAVA! (
     echo The specified path !JAVA! does not point to a valid Java Development Kit installation.
     echo Please check that the path is correct and that a Java Development Kit is installed at that location.
+    echo   https://openrefine.org/docs/manual/installing#java
     goto :fail
 )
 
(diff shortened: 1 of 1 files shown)


refine.bat
```diff
@@ -201,8 +201,15 @@ if ""%ACTION%"" == """" goto doRun
...
 :doRun
-rem --- Log for troubleshooting ------------------------------------------
-set JAVA="%JAVA_HOME%/bin/java"
+rem --- Checking Java Version  ------------------------------------------
+set JAVA="%JAVA_HOME%\bin\java.exe"
+
+if not exist !JAVA! (
+    echo The specified path !JAVA! does not point to a valid Java Development Kit installation.
+    echo Please check that the path is correct and that a Java Development Kit is installed at that location.
+    goto :fail
```
wetneb, line 210: Would it be worth also including the link to the docs in this case?

wetneb (approved):

PR #5748, opened 2023-03-28
Apply operation json from a file

### LINKED ISSUE #5022: Apply Operation History from JSON file
Related to #4498, I would like to see a button to Apply Operation History by opening a JSON file instead of having to open the file separately. This would remove any need to copy-paste the contents of these files, and make it easier for users.

IjayAbby:
Fixes #5022

-     "Import" button added to the Apply Operation History dialogue
-     On clicking import, the user prompted to upload the file
-     Once the file is uploaded the content (assuming can be read correctly) is populated into the operation history window
-     User can then click "Perform operations" as usual

So far I have been able to create the Upload button but when I try to upload the file nothing happens because I need the correct URL to download the file in line 355. Let me know your suggestions on how to best approach this. Thanks.

### DIFF
--- a/main/webapp/modules/core/scripts/project/history-apply-dialog.html
+++ b/main/webapp/modules/core/scripts/project/history-apply-dialog.html
@@ -3,7 +3,7 @@
     <div class="dialog-header" bind="dialogHeader"></div>
     <div class="dialog-body" bind="dialogBody">
       <div class="dialog-instruction"><label for="textareaId" bind="or_proj_pasteJson"></label></div>
-      <input type="file" id="file-input" bind="operationJsonButton"/>
+      <input type="file" id="file-input" bind="operationJsonButton" style="margin-bottom: 8px;"/>
       <div class="input-container"><textarea spellcheck="false" wrap="off" bind="textarea" id="textareaId" class="history-operation-json" style="min-width: 764px"></textarea></div>
     </div>
     <div class="dialog-footer" bind="dialogFooter">
--- a/main/webapp/modules/core/scripts/project/history-apply-dialog.html
+++ b/main/webapp/modules/core/scripts/project/history-apply-dialog.html
@@ -10,7 +10,6 @@
     <button class="button" bind="cancelButton"></button>
     <button class="button" bind="operationJsonButton"></button>
     <input type="file" id="file-input" style="display:none;">
-    <pre id="file-contents"></pre>
     </div>
   </div>
 </div>
--- a/main/webapp/modules/core/scripts/project/history-apply-dialog.html
+++ b/main/webapp/modules/core/scripts/project/history-apply-dialog.html
@@ -3,13 +3,12 @@
     <div class="dialog-header" bind="dialogHeader"></div>
     <div class="dialog-body" bind="dialogBody">
       <div class="dialog-instruction"><label for="textareaId" bind="or_proj_pasteJson"></label></div>
-      <input type="file" id="file-input" bind="operationJsonButton"/>
+      <input type="file" id="file-input" bind="operationJsonButton" class="upload-json"/>
       <div class="input-container"><textarea spellcheck="false" wrap="off" bind="textarea" id="textareaId" class="history-operation-json" style="min-width: 764px"></textarea></div>
     </div>
     <div class="dialog-footer" bind="dialogFooter">
     <button class="button" bind="applyButton"></button>
     <button class="button" bind="cancelButton"></button>
-    <input type="file" id="file-input" style="display:none;">
     </div>
   </div>
 </div>
--- a/main/webapp/modules/core/scripts/project/history-apply-dialog.html
+++ b/main/webapp/modules/core/scripts/project/history-apply-dialog.html
@@ -8,7 +8,7 @@
     <div class="dialog-footer" bind="dialogFooter">
     <button class="button" bind="applyButton"></button>
     <button class="button" bind="cancelButton"></button>
+    <button class="button" bind="operationJsonButton"></button>
     </div>
   </div>
 </div>
-
--- a/main/webapp/modules/core/scripts/project/history-apply-dialog.html
+++ b/main/webapp/modules/core/scripts/project/history-apply-dialog.html
@@ -8,7 +8,8 @@
     <div class="dialog-footer" bind="dialogFooter">
     <button class="button" bind="applyButton"></button>
     <button class="button" bind="cancelButton"></button>
-    <button class="button" bind="operationJsonButton"></button>
+    <button class="button" id="upload-btn" bind="operationJsonButton"></button>
+    <input type="file" id="file-input" style="display:none;">
     </div>
   </div>
 </div>
--- a/main/webapp/modules/core/scripts/project/history-apply-dialog.html
+++ b/main/webapp/modules/core/scripts/project/history-apply-dialog.html
@@ -3,12 +3,12 @@
     <div class="dialog-header" bind="dialogHeader"></div>
     <div class="dialog-body" bind="dialogBody">
       <div class="dialog-instruction"><label for="textareaId" bind="or_proj_pasteJson"></label></div>
+      <button class="button" bind="operationJsonButton"></button>
       <div class="input-container"><textarea spellcheck="false" wrap="off" bind="textarea" id="textareaId" class="history-operation-json" style="min-width: 764px"></textarea></div>
     </div>
     <div class="dialog-footer" bind="dialogFooter">
     <button class="button" bind="applyButton"></button>
     <button class="button" bind="cancelButton"></button>
-    <button class="button" bind="operationJsonButton"></button>
     <input type="file" id="file-input" style="display:none;">
     </div>
   </div>
--- a/main/webapp/modules/core/scripts/project/history-apply-dialog.html
+++ b/main/webapp/modules/core/scripts/project/history-apply-dialog.html
@@ -8,8 +8,9 @@
     <div class="dialog-footer" bind="dialogFooter">
     <button class="button" bind="applyButton"></button>
     <button class="button" bind="cancelButton"></button>
-    <button class="button" id="upload-btn" bind="operationJsonButton"></button>
+    <button class="button" bind="operationJsonButton"></button>
     <input type="file" id="file-input" style="display:none;">
+    <pre id="file-contents"></pre>
     </div>
   </div>
 </div>
--- a/main/webapp/modules/core/scripts/project/history-apply-dialog.html
+++ b/main/webapp/modules/core/scripts/project/history-apply-dialog.html
@@ -3,7 +3,7 @@
     <div class="dialog-header" bind="dialogHeader"></div>
     <div class="dialog-body" bind="dialogBody">
       <div class="dialog-instruction"><label for="textareaId" bind="or_proj_pasteJson"></label></div>
-      <input type="file" id="file-input" bind="operationJsonButton" class="upload-json"/>
+      <input type="file" id="file-input" bind="operationJsonButton"/>
       <div class="input-container"><textarea spellcheck="false" wrap="off" bind="textarea" id="textareaId" class="history-operation-json" style="min-width: 764px"></textarea></div>
     </div>
     <div class="dialog-footer" bind="dialogFooter">
--- a/main/webapp/modules/core/scripts/project/history-apply-dialog.html
+++ b/main/webapp/modules/core/scripts/project/history-apply-dialog.html
@@ -3,7 +3,7 @@
     <div class="dialog-header" bind="dialogHeader"></div>
     <div class="dialog-body" bind="dialogBody">
       <div class="dialog-instruction"><label for="textareaId" bind="or_proj_pasteJson"></label></div>
-      <button class="button" bind="operationJsonButton"></button>
+      <input type="file" id="file-input" bind="operationJsonButton"/>
       <div class="input-container"><textarea spellcheck="false" wrap="off" bind="textarea" id="textareaId" class="history-operation-json" style="min-width: 764px"></textarea></div>
     </div>
     <div class="dialog-footer" bind="dialogFooter">
--- a/main/webapp/modules/core/scripts/project/history-panel.js
+++ b/main/webapp/modules/core/scripts/project/history-panel.js
@@ -290,11 +290,10 @@ HistoryPanel.prototype._showApplyOperationsDialog = function() {
   elmts.or_proj_pasteJson.html($.i18n('core-project/paste-json'));
 
   elmts.operationJsonButton.on('click', async function() {
-    const input = document.createElement('input');
-    input.type = 'file';
-    input.accept = '.json'; // optional file types
-    input.onchange = async function() {
-      const file = input.files[0];
+    const fileInput = document.querySelector('#file-input');
+    fileInput.accept = '.json'; // optional file types
+    fileInput.onchange = async function() {
+      const file = fileInput.files[0];
       const reader = new FileReader();
       reader.onload = function(e) {
         const fileContent = JSON.parse(e.target.result);
@@ -305,11 +304,11 @@ HistoryPanel.prototype._showApplyOperationsDialog = function() {
         }
       };
       reader.addEventListener('error', function() {
-        alert('Error : Failed to read file');
+        alert('Error: Failed to read file');
     });
       reader.readAsText(file);
     };
-    input.click();
+    fileInput.click();
   });
   
   
--- a/main/webapp/modules/core/scripts/project/history-panel.js
+++ b/main/webapp/modules/core/scripts/project/history-panel.js
@@ -339,34 +339,21 @@ HistoryPanel.prototype._showApplyOperationsDialog = function() {
     DialogSystem.dismissUntil(level - 1);
   });
 
-  $(document).ready(function() {
-    // Trigger the file input when the button is clicked
-    $('#upload-btn').click(function() {
-      $('#file-input').click();
-    });
-    
-    // Upload the file when the user selects it
-    $('#file-input').change(function() {
-      var file = $('#file-input')[0].files[0];
-      var formData = new FormData();
-      formData.append('file', file);
-      
-      $.ajax({
-        url: 'core-index-import/locate-files', // The URL of the script that handles the file upload
-        type: 'POST',
-        data: formData,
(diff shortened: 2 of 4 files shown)


ostephens:
Thanks for this @ljayAbby suggest you make this a draft PR for now while you get to a point the code works (see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request for information on making a PR a Draft PR)

IjayAbby:
Done that already @ostephens

wetneb:
As discussed in the issue, you should be able to read the file entirely client side, without submitting it to the backend via a command (whose URL you said above you are looking for).
See for instance tutorials like this one: https://usefulangle.com/post/193/javascript-read-local-file

wetneb:
You could also edit your issue description to have not "Fixes #{5022}" but rather "Fixes #5022" so that this pull request gets linked to the corresponding issue.

ostephens:
Thanks @wetneb 

@IjayAbby also I think you shouldn't be wrapping the function in the `$(document).ready(function() {` - you just want to have the function like `elmts.<binding>.on('click',function() {` using the appropriate binding

IjayAbby:
@wetneb I found out that FileReader can only access the contents of files that the user has explicitly selected, either using an HTML <input type="file"> element or by drag and drop. It cannot be used to read a file by pathname from the user's file system. To read files on the client's file system by pathname, use the File System Access API. So I have used File System Access API instead.
However, once the upload button is clicked you can choose a .json file but an error occurs after that. Been debugging but am stuck.

ostephens:
> I found out that FileReader can only access the contents of files that the user has explicitly selected

Isn't that what is happening here?

IjayAbby:
It was throwing an error that FileReader() is not an object.

IjayAbby:
My console error message reads as follows: Uncaught TypeError: FileReader.readAsText: Argument 1 is not an object.

IjayAbby:
@wetneb can I still use this for my outreach application?

wetneb:
You can, as Outreachy lets you record pull requests which habe not been merged yet.

IjayAbby:
Does my approach work or what can you recommend since am stuck

wetneb:
The error you are quoting above looks like a generic programming error, indicating that you are not using the api in the expected way. Perhaps you could try making a prototype for this outside of openrefine, to make sure you get the api right. Once you are able to read the file contents into a textarea, then you can go back to open refine and integrate that.

IjayAbby:
Okay thanks.

wetneb:
There is no need to send the file to the backend at all, as explained above. Therefore you will not need any URL where to post the file to.

IjayAbby:
So I do away with the url but the file reader object is problematic on my end

IjayAbby:
Hello @wetneb how do I create an Outreachy internship project timeline

wetneb:
You could try to break down the project description in some meaningful steps and try to estimate how much time you might need to work on them. You can plan ample time for multiple rounds of reviews and user testing so that it converges towards something solid.

IjayAbby:
Thanks, on my issue what should I put as the end date or is it editable
after today too.

IjayAbby:
@wetneb I used the FileReader object and it worked contents are visible on my console. How do I upload it in the Dialog please?

IjayAbby:
@wetneb this pull awaits your review.

wetneb (changes requested):
Thank you for your work! It does seem to work, but I think the layout of the dialog can be improved. On my side it currently looks like this:
image

Because the button you added is labeled "Upload" and is placed in the bottom right area of the dialog, it intuitively feels to me like it could mean "Upload the operations and perform them", which is actually the role of the "Apply operations" button.

Instead of having it down there, I think it would be better placed in the top part. The help text currently says "Paste an extracted JSON history" but thanks to your changes the user actually has the option not to paste but select a file instead. So intuitively this text should be updated to reflect this, and perhaps this is a good opportunity to integrate the button there.

How about something like this?

> Paste or **Select a file** with an extracted JSON history of operations to perform:

where the **Select a file** is your button.

It could look like this:
image

IjayAbby:
Okay, I will look into it. Thank you.

wetneb:
Actually, I now realize that the proposal above is perhaps not very convenient for localization (inserting a button in the middle of localized text is not great, nor is splitting the localized text into the boundaries defined by the button). So it might be easier to have the button separate from the explanation text.

wetneb:
Maybe @Lydiaofficial is interested in suggesting a design here? Just in case! ^^

IjayAbby:
But either way just have the button up there or I wait for the design?

wetneb:
If you can think of a design yourself, feel free to go ahead and propose it. Otherwise, it can be worth waiting for a few days if others have suggestions.

IjayAbby:
Thank you @Lydiaofficial for the design.

main/webapp/modules/core/scripts/project/history-apply-dialog.html
```diff
@@ -3,12 +3,13 @@
     <div class="dialog-header" bind="dialogHeader"></div>
     <div class="dialog-body" bind="dialogBody">
       <div class="dialog-instruction"><label for="textareaId" bind="or_proj_pasteJson"></label></div>
+      <button class="button" bind="operationJsonButton"></button>
```
wetneb, line 6:
```suggestion
      <input type="file" id="file-input" bind="operationJsonButton" />
```
How about simply using a native file input instead? I think it would be more recognizeable as a file field ("Select" in itself is not very suggestive). I think it should also simplify the code (no need for a hidden input) and improve accessibility.

main/webapp/modules/core/scripts/project/history-panel.js
```diff
@@ -288,9 +288,34 @@ HistoryPanel.prototype._showApplyOperationsDialog = function() {
   
   elmts.dialogHeader.html($.i18n('core-project/apply-operation'));
   elmts.or_proj_pasteJson.html($.i18n('core-project/paste-json'));
+
+  elmts.operationJsonButton.on('click', async function() {
+    const input = document.createElement('input');
```
wetneb, line 293: I suspect we might be able to avoid this creation of a dom element since we already have one in the dialog, no?

main/webapp/modules/core/scripts/project/history-panel.js
```diff
@@ -288,9 +288,34 @@ HistoryPanel.prototype._showApplyOperationsDialog = function() {
...
+    input.type = 'file';
+    input.accept = '.json'; // optional file types
+    input.onchange = async function() {
+      const file = input.files[0];
+      const reader = new FileReader();
+      reader.onload = function(e) {
+        const fileContent = JSON.parse(e.target.result);
+        const textAreaElement = document.querySelector('#textareaId')
```
wetneb, line 301:
```suggestion
        const textAreaElement = elmts.textarea;
```
This would avoid looking up the textarea via a global identifier (which could easily conflict with another element, since the id is pretty generic) and instead select the one in the current dialog. Please check if it works though!

main/webapp/modules/core/scripts/project/history-panel.js
```diff
@@ -288,9 +288,34 @@ HistoryPanel.prototype._showApplyOperationsDialog = function() {
...
+        if (textAreaElement) {
+          textAreaElement.textContent = JSON.stringify(fileContent, null, 2)
+
+        }
+      };
+      reader.addEventListener('error', function() {
+        alert('Error : Failed to read file');
```
wetneb, line 308: This would need to be localized with `$.i18n`, and in English there should not be any space between `Error` and the colon `:`.

server/IDEs/eclipse/Refine with extensions.launch
```diff
@@ -1,18 +1,18 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <launchConfiguration type="org.eclipse.jdt.launching.localJavaApplication">
-<listAttribute key="org.eclipse.debug.core.MAPPED_RESOURCE_PATHS">
```
wetneb, line 3: This change is unrelated and should be reverted so that the PR only touches relevant files. Because automatic updates to this file are a recurrent problem I will update it independently.

wetneb (commented):
Thanks a lot, I think we are nearly there!

main/webapp/modules/core/scripts/project/history-panel.js
IjayAbby: Done.

main/webapp/modules/core/scripts/project/history-panel.js
IjayAbby: Unsuccessful, brings about error so I will maintain how it was initially.

main/webapp/modules/core/scripts/project/history-apply-dialog.html
IjayAbby: Done

server/IDEs/eclipse/Refine with extensions.launch
IjayAbby: Done.

main/webapp/modules/core/scripts/project/history-panel.js
wetneb: Have you tried with `elmts.textarea[0]` instead? Looking at the rest of the code, it seems that it could be better like that.

main/webapp/modules/core/styles/common.css
```diff
@@ -64,6 +64,7 @@ table {
 
 input, select {
   vertical-align: middle;
+  margin-bottom: 8px;
```
wetneb, line 67: I would rather not indiscriminately add a margin to all `input` and `select` elements in the application as this can have consequences in a lot of places! Please add a narrower selector, ensuring you only style things within the dialog you are working on.

main/webapp/modules/core/scripts/project/history-apply-dialog.html
```diff
@@ -3,12 +3,13 @@
     <div class="dialog-header" bind="dialogHeader"></div>
     <div class="dialog-body" bind="dialogBody">
       <div class="dialog-instruction"><label for="textareaId" bind="or_proj_pasteJson"></label></div>
+      <input type="file" id="file-input" bind="operationJsonButton"/>
       <div class="input-container"><textarea spellcheck="false" wrap="off" bind="textarea" id="textareaId" class="history-operation-json" style="min-width: 764px"></textarea></div>
     </div>
     <div class="dialog-footer" bind="dialogFooter">
     <button class="button" bind="applyButton"></button>
     <button class="button" bind="cancelButton"></button>
+    <input type="file" id="file-input" style="display:none;">
```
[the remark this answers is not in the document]
IjayAbby: Deleted it

main/webapp/modules/core/scripts/project/history-panel.js
```diff
@@ -288,9 +288,33 @@ HistoryPanel.prototype._showApplyOperationsDialog = function() {
   
   elmts.dialogHeader.html($.i18n('core-project/apply-operation'));
   elmts.or_proj_pasteJson.html($.i18n('core-project/paste-json'));
+
+  elmts.operationJsonButton.on('click', async function() {
+    const fileInput = document.querySelector('#file-input');
```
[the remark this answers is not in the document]
IjayAbby: I get error when I use elmts so I have commented my code and maintained the previous one.
Need guidance on how to best approach this.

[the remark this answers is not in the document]
wetneb: The identifier that comes after `elmts.` is the string that is in the `bind=` attribute of the DOM element you want to select. You should not need to add any class as you did in your latest commit.

main/webapp/modules/core/scripts/project/history-panel.js
wetneb: No, it does not look like this string is localized. The string is still hard-coded in English. Can you localize it with `$.i18n`?

main/webapp/modules/core/scripts/project/history-panel.js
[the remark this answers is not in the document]
IjayAbby: Fixed, thanks.

main/webapp/modules/core/styles/common.css
IjayAbby: Done.

main/webapp/modules/core/styles/common.css
```diff
@@ -348,6 +347,10 @@ a img {
   margin: 0.4em;
 }
 
+#file-input {
+  margin-bottom: 8px;
+}
```
wetneb, line 352: That's better, but still not a specific CSS selector: your `file-input` element id is very generic, so there could be other file inputs with the same id in other parts of the application (even though ids are supposed to be unique), and this CSS rule would apply to them. Why not just put inline CSS on the element?

IjayAbby: Done.

main/webapp/modules/core/scripts/project/history-panel.js
```diff
@@ -197,6 +197,7 @@ HistoryPanel.prototype._showExtractOperationsDialog = function(json) {
 
   elmts.dialogHeader.html($.i18n('core-project/extract-history'));
   elmts.textarea.attr('aria-label',$.i18n('core-project/operation-history-json'))
+  //elmts.fileInput.attr('aria-label',$.i18n('core-project/upload-json'))
```
wetneb, line 200: We could also remove up that unnecessary line

wetneb (commented):
Thanks! We are almost there: see the pending review comment above about localization.

main/webapp/modules/core/scripts/project/history-panel.js
```diff
@@ -288,9 +288,32 @@ HistoryPanel.prototype._showApplyOperationsDialog = function() {
...
+      const reader = new FileReader();
+      reader.onload = function(e) {
+        const fileContent = JSON.parse(e.target.result);
+        const textAreaElement = elmts.textarea[0];
+        if (textAreaElement) {
+          textAreaElement.textContent = JSON.stringify(fileContent, null, 2)
+        }
+      };
+      reader.addEventListener('error', function() {
+        alert('Error: Failed to read file');
```
wetneb, line 306: Hi @IjayAbby, as mentioned earlier this would still need localizing before we can merge this.

IjayAbby: Hello @wetneb, how do I do that Open Refine has many languages to be translated under? Kindly advise
This is my approach 
First, I need to include the jquery.i18n library in the project and initialize it with the appropriate translation files for the languages I want to support.
```
$.i18n().load({
  'en': 'path/to/en.json',
  'fr': 'path/to/fr.json'
});
```
Then, I can use the $.i18n() function to retrieve the localized string for the error message:

```
reader.addEventListener('error', function() {
  alert($.i18n('error_message'));
});
```

wetneb: Your first step is not necessary because the translation files are already loaded (to translate the rest of the app). So you can directly move on to your second step.

IjayAbby: Done.

main/webapp/modules/core/scripts/project/history-panel.js
```diff
@@ -303,7 +303,7 @@ HistoryPanel.prototype._showApplyOperationsDialog = function() {
         }
       };
       reader.addEventListener('error', function() {
-        alert('Error: Failed to read file');
+        alert($.i18n('Error: Failed to read file'));
```
wetneb, line 306: Not quite there yet. Can you look at other uses of `$.i18n` in the rest of the code base to understand how it is used? We also have some documentation about it: https://openrefine.org/docs/technical-reference/translating-ui

IjayAbby: @wetneb , it works fine now.

main/webapp/modules/core/scripts/project/history-panel.js
```diff
@@ -288,9 +288,35 @@ HistoryPanel.prototype._showApplyOperationsDialog = function() {
...
+        try {
+          const fileContent = JSON.parse(e.target.result);
+          const textAreaElement = elmts.textarea[0];
+          if (textAreaElement) {
+            textAreaElement.textContent = JSON.stringify(fileContent, null, 2)
+          }
+        } catch (error) {
+          {
+            window.alert($.i18n('core-index-import/warning-data-file'));
+          }
```
wetneb, line 308: The `{` and `}` around `window.alert(...)` should not be necessary: no need to duplicate those of the `catch` block.

wetneb:
It is nice that you tried to reuse an existing message, but I am not sure if the content of the message is really fitting in this context.
I tried using an invalid JSON file and I got this message: "You must specify a data file to import."

I see multiple problems with this message in this context:
* I did provide a file already, that's not the problem. The problem is that it is not valid JSON.
* The file I am supposed to provide is not a data file, but the representation of a sequence of OpenRefine operations in JSON.

main/webapp/modules/core/scripts/project/history-panel.js
IjayAbby: Got it I will fix it, used something that existed from the docs, but I will create a new one for this.

wetneb: Looks good to me, if you could just remove the `+"."` at the end it would be great because there might be languages where that does not make sense.

wetneb (approved):
It looks good to me now, thanks a lot for your patience!
I am rebasing your branch to fix the build issue.

IjayAbby:
Thank you so much @wetneb for the support and thanks to Outreachy for such a fantastic opportunity. The past two months have been amazing for me from knowing nothing to diving into this vast codebase to mastering my way around it. 
It would be an honor to work with you and @Abbe98 during the internship if given an opportunity. Again, thank you for the opportunity, I wish you would see the excitement in me.

wetneb:
And thanks to @Lydiaofficial for the help with the design! :)

Also, with this PR, the number of people having made a commit to the repository reached 300. Yay!

@IjayAbby the intern selection decisions have been made about a month ago already, after the end of the contribution period, so at this stage we do not have any influence on it anymore. They should be announced today I think.

Lydiaofficial:
Always a pleasure 😊

PR #5754, opened 2023-03-29
ignore .gitignore file in the extension directory

Abbe98:
Having a .gitignore file ignoring either extension folders or symlinks can be super useful during development. For example when testing extension compatibility with ones changes.


wetneb (approved):
Makes intuitive sense to me.

PR #5759, opened 2023-03-31
[I# 1011] Second implementation of facet renaming

### LINKED ISSUE #1011: Custom facet names

antoine2711:
Fixes #1011 

- Add an edit background when over the title
- Show a tooltip when over the title
- Provide a FR translation
- Has functional tests in Cypress

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -141,7 +141,7 @@
         </plugin>
         <plugin>
           <artifactId>maven-site-plugin</artifactId>
-          <version>3.12.1</version>
+          <version>3.20.0</version>
         </plugin>
         <plugin>
           <artifactId>maven-source-plugin</artifactId>


antoine2711:
@wetneb : I've changed the Source <blank-record> to « Blank records per column » in the ToolTip.

Did I understand your reservation in the meeting? Does this address it?

This should be finished, all tests passed.

Regards,
   Antoine

elebitzero (changes requested):
Looks good but the following adjustments should be made:

Change 'Click here to edit the name of the facet' to 'Click to rename facet' so it is consistent with rename project verbiage:
image

There is no need to put the expression in the rename facet tooltip. The expression is visible in the 'change' tooltip, or by clicking 'change' to popup the expression dialog:
image

main/webapp/modules/core/scripts/facets/facet.js
```diff
@@ -67,3 +110,20 @@ class Facet {
...
+  "(filter(row.columnNames,cn,isNonBlank(cells[cn].value)).length()==0).toString()": 
+    { source: "<blank-rows>", label: 'core-views/blank-rows' },
+  "filter(row.columnNames,cn,isBlank(cells[cn].value))": 
+    { source: "<blank-values>", label: 'core-views/blank-values' },
+  "filter(row.columnNames,cn,isBlank(if(row.record.fromRowIndex==row.index,row.record.cells[cn].value.join(\"\"),true)))": 
+    { source: "<blank-records>", label: 'core-views/blank-records' },
+  "filter(row.columnNames,cn,isNonBlank(cells[cn].value))": 
+    { source: "<non-blank-values>", label: 'core-views/non-blank-values' },
+  "filter(row.columnNames,cn,isNonBlank(if(row.record.fromRowIndex==row.index,row.record.cells[cn].value.join(\"\"),null)))": 
+    { source: "<non-blank-records>", label: 'core-views/non-blank-records' }
```
wetneb, line 128: I would rather store this mapping in the definition of the facets themselves (in the menu). Because it is likely that people forget to update this when they add a new facet in that menu, or for instance when an extension registers a facet there. I think this could be done simply changing the method which adds a new facet, so that the facet name can be set directly there.

wetneb:
@antoine2711 requested help to rebase this branch on the latest master, so I have done so here.

PR #5760, opened 2023-03-31
Change the refine script to print free memory message only for the run action

elebitzero:
The refine script prints out the free memory and run memory config for every action, even though the "run" memory config is only relevant when running OpenRefine. For other actions e.g. "build", "dist", etc. it is not applicable.

```
$ ./refine build
You have 5936M of free memory.
https://openrefine.org/docs/manual/installing#increasing-memory-allocation
```

This PR changes it so this message is only printed for the run action.

### DIFF
--- a/refine
+++ b/refine
@@ -446,6 +446,20 @@ run() {
         fi
     fi
 
+freeRam=UNKNOWN
+if [ "$OS" = "macosx" ] ; then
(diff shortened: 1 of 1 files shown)


wetneb (approved):
I was also thinking about that, for too long already! Thanks a lot for taking care of it.

PR #5761, opened 2023-04-01
Added column by fetching URL redirection destination

sn1616:
Fixes #{5641}

- Added column by fetching URL redirection destination
- Utilizing HTTP HEAD requests 
- returning the final URL


wetneb:
Hi!
Thanks for your contribution! I wonder if you have tested it? Looking at your changes, you have created an independent web app in a separate folder. That does not help us so much: we want to implement this feature in open refine itself.

wetneb:
It is clear that you are not testing your changes. This code does not compile.

I try to do my best being supportive of contributors of all abilities, but my time is limited.
Please only re-open a pull request once you have tested the changes on your side and validated that they work as expected. Thank you!

PR #5762, opened 2023-04-01
[I# 3147]  PR to implement `forEachItem` for JSON object

### LINKED ISSUE #3147: Implement `forEachItem` for JSON object
Users have ongoing complaints about the difficult in parsing  JSON which has arbitrary structure, most often in the form of JSON API responses which were designed to return objects keyed by arbitrary identifiers rather than just returning array as the vast more of APIs do.

Examples provided in #2186 include:
* Wikimedia Commons -> https://commons.wikimedia.org/w/api.php?action=query&titles=File:Google%202015%20logo.svg&prop=imageinfo&iiprop=url&format=json
* MySociety MapIt -> https://mapit.mysociety.org/point/4326/0.1293497,51.5464828

And the issue has been discussed both on the OpenRefine mailing list and StackOverflow:
* https://groups.google.com/d/msg/openrefine/O72bg8GZkj8/vBoEzmGxBwAJ
* https://stackoverflow.com/questions/10782737/google-refine-iterate-over-a-json-dictionary

### Proposed solution

Similar to the `forEachIndex` function that we have for arrays, we would implement a `forEachItem` iterator control which would accept a JSON node object, two variable names (one for key and one for value), and a GREL expression. It would iterate over each item in the JSON object, assigning the string value of the key to the key variable and the value for the item to the value variable and then execute the GREL expression.

The control would **only** accept a JSON object to iterate over and would explicitly **not** support iterating over objects that implement the `HasFields` interface.

### Alternatives considered
* extend `forEach()` to iterate over JSON object items. This wouldn't provide access to the field name / key, but would probably satisfy the vast majority of cases since the key is usually included in the child object as an id.
* implement iteration for every object that implements HasFields. Since OpenRefine `HasFields` objects typically have a small fixed set of fields, this seems to have limit value.

sn1616:
Fixes #3147

- Implemented forEachItem for JSON object

### DIFF
--- a/main/src/com/google/refine/grel/Function.java
+++ b/main/src/com/google/refine/grel/Function.java
@@ -38,7 +38,58 @@ import java.util.Properties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
+package com.google.refine.expr.functions;
 
+import java.util.Properties;
+
+import org.codehaus.jackson.JsonNode;
+
+import com.google.refine.expr.EvalError;
+import com.google.refine.expr.ExpressionUtils;
+import com.google.refine.expr.HasFields;
+import com.google.refine.grel.ControlFunctionRegistry;
+import com.google.refine.grel.Function;
+
+public class GrelFunctions {
+    public static void register(ControlFunctionRegistry registry) {
+        registry.registerFunction(new ForEachItem());
+    }
+
+    private static class ForEachItem implements Function {
+
+        @Override
+        public Object call(Properties bindings, Object[] args) {
+            if (args.length != 3) {
+                return new EvalError("forEachItem: Wrong number of arguments");
+            }
+            Object obj = args[0];
+            String keyName = args[1].toString();
+            String valueName = args[2].toString();
+            if (!(obj instanceof JsonNode)) {
+                return new EvalError("forEachItem: Argument is not a JSON object");
+            }
+            JsonNode json = (JsonNode) obj;
+            if (!json.isObject()) {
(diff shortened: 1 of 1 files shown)


antoine2711:
If you want the Issue to be linked, you must write this on the first line:

`Fixes #3147`

And not:

`Fixes #{3147}`

I corrected your issue and also have put a better name.

Regards, Antoine

sn1616:
Got it . Thanks , wont happen again

antoine2711:
Also, I want to point out that this issue was assigned to @satyamksharma.

By solving it with your PR, you are doing the same thing as this person, which is a lost of time for one or the other.
It can also be seen as disrespectful, because she has asked for this issue to be reserved to her, and you are going over this assignment. You probably didn't know that, but you should be careful before involving yourself in an issue, and make sure it is not already assigned.

See on the right.

image

That being said, your works seams good, so a little effort will help you be more appreciated.

Respectfully, Antoine

sn1616:
Got it . Thanks

sn1616:
I was not aware of it being a beginner , didnt mean to be disrespectful. Is the issue still available , I am interested on working on it

antoine2711:
> I was not aware of it being a beginner , didnt mean to be disrespectful. Is the issue still available , I am interested on working on it

Then, @sn1616, what you have to do is enter in contact with the assignee, @satyamksharma, and ask her if she's still interested in fixing this issue. She might let the issue, but she might say that she wants to keep working it. It's been 2 weeks, and she did nothing visible, so maybe she is not very advanced on it.

Before involving yourself in an issue more than analysing, check if it's assigned. If it's not, and you don't have other issues assigned to you, you can « claim » the issue and assing yourself.

I also recommend you read this page:
https://openrefine.org/docs/technical-reference/contributing

Regards, Antoine

PR #5767, opened 2023-04-03
Fix initial facet height when facets are added while the facet panel is not visible

### LINKED ISSUE #5749: Facet sizing broken if the facet panel is not open when the facet is created
Creating a facet while the Undo / Redo tab is active makes the facet initially collapsed, breaking the automatic sizing introduced in #5514.

### Current results

Peek 29-03-2023 11-15

### Expected Behavior

The facet should be properly sized automatically

### Versions
 - Browser Version: Firefox 102.8.0esr 
 - OpenRefine: 3.8-SNAPSHOT

### Additional context

Ideally, the facet panel should also be switched to automatically, because when I create a facet I want to see it directly (either to look at its statistics or to use it to filter rows / records). This means that the fix should be rather simple and should not worry about resizing an element that is not currently rendered: just switch to the facet panel whenever new facets are created. That being said, I am not completely sure about what to do when facets are automatically created without an action from the user, for instance after a reconciliation operation is run.

elebitzero:
Fixes #5749
- Made it so the initial facet height is properly set when facets are added when the facet panel is not visible.
- Add callback on jQuery event "tabsactivate". If the new active tab is a facet tab, then set the initial height for all facets (if not already set).

facet-initial-height-when-hidden

### DIFF
--- a/main/webapp/modules/core/scripts/facets/list-facet.js
+++ b/main/webapp/modules/core/scripts/facets/list-facet.js
@@ -123,6 +123,23 @@ class ListFacet extends Facet {
     this._update();
   };
 
+  checkInitialHeight() {
+    if (this._elmts) {
+      let innerList = this._elmts.bodyInnerDiv[0];
+      if (!this._initialHeightSet && innerList.offsetHeight !== 0) {
+        let innerHeight = innerList.offsetHeight;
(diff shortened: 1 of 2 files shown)


wetneb (approved):
Thanks a lot for tackling this @elebitzero! It works perfectly on my side.

PR #5768, opened 2023-04-03
Add tests for SeparatorBasedImporter

Deee92:
- test to verify that the separator guessed for a CSV resource file is `,`
- tests to verify that the default guessed separator, and the separator for JSON and TSV resource files is `\t`


tfmorris:
> Confusingly enough there is already a test class, which does not follow the same naming convention: TsvCsvImporterTests

It looks like the test case wasn't renamed in 2011 78edff6f7f410bb53c03e67f7f90a6d3521cee7f when the importer class was renamed. There's a similar but different issue with the exporter:

```
./main/tests/server/src/com/google/refine/exporters/TsvExporterTests.java
./main/tests/server/src/com/google/refine/exporters/CsvExporterTests.java
./main/src/com/google/refine/exporters/CsvExporter.java
```

main/tests/server/src/com/google/refine/importers/SeparatorBasedImporterTests.java
```diff
@@ -632,6 +641,39 @@ public void readCsvWithPropertiesIgnoreQuotes() {
...
+                job, fileRecords, "text/tsv");
+        assertEquals("\\t", options.get("separator").textValue());
+    }
+
+    @Test
+    public void testThatSeparatorIsGuessedCorrectlyForJSON() throws IOException {
+        List<ObjectNode> fileRecords = prepareFileRecords("grid_small.json");
+        ObjectNode options = SUT.createParserUIInitializationData(
+                job, fileRecords, "text/json");
+        assertEquals("\\t", options.get("separator").textValue());
```
wetneb, line 674: I am not sure why the guesser should return `\t` when run on a JSON file. Because a JSON file will not be parseable as a CSV anyway, I would not assert anything about the separator guessed in this case.

Deee92: I see. `\t` is guessed as separator by default in either case. I can remove this test.

wetneb (approved):

PR #5769, opened 2023-04-03
open project: move tag filter state from URL hash to query params

Abbe98:
The ability to filter projects by tags uses URL-hashes for managing state, this, however interferes with the home screen panel links introduced in #5597. This change therefor, moves the filtering state to query params.

Side effects:
 - you can now link to a tag filter
 - fixed some bugs caused by naming a tag "all"
 - ensured that OpenRefine's URL utility does not overwrite the native URL class.


elebitzero (commented):
Looks good, except on refresh the selected tag is not restored:

http://127.0.0.1:3333/?tag=foo#open-project
should select 'foo' and filter the projects with foo tag.

http://127.0.0.1:3333/?tag=#open-project
should select 'All' and show all projects.

elebitzero (changes requested):
^

wetneb:
@Abbe98 quick reminder about this - It looks like a good move to me so it's probably worth addressing @elebitzero's comments to avoid the PR getting stale?

Abbe98:
> @Abbe98 quick reminder about this - It looks like a good move to me so it's probably worth addressing @elebitzero's comments to avoid the PR getting stale?

It's on my radar and I have a solution in mind.

elebitzero (approved):

elebitzero:
The changes look good, OR now keeps the selected tag on refresh.

PR #5770, opened 2023-04-04
Remove unneeded frontend code for open project UI

### LINKED ISSUE #5747: Seemingly dead frontend code refering to create-project-from-upload command
The following reference to the `create-project-from-upload` command seems to be dead code.

https://github.com/OpenRefine/OpenRefine/blob/76152500a5af8c38b88d34e15b8b01fcf6d0da6e/main/webapp/modules/core/scripts/index/open-project-ui.js#L316-L344

This request is bound to an event listener on ` #upload-file-button`, but I cannot find any such element in our code base.

https://github.com/OpenRefine/OpenRefine/blob/76152500a5af8c38b88d34e15b8b01fcf6d0da6e/main/webapp/modules/core/scripts/index/open-project-ui.js#L42-L58

The other elements whose values are encoded in the request also do not seem to exist anymore.

One should double check that this is not needed and remove this code (and any other parts that are also never executed anymore).

elebitzero:
Fixes #5747

I tested the code in question and it doesn't execute after clicking everything in the open project UI.

- Removed unneeded frontend code for open project UI.

### DIFF
--- a/main/webapp/modules/core/scripts/index/open-project-ui.js
+++ b/main/webapp/modules/core/scripts/index/open-project-ui.js
@@ -39,24 +39,6 @@ Refine.OpenProjectUI = function(elmt) {
   this._elmt = elmt;
   this._elmts = DOM.bind(elmt);
 
-  $("#project-file-input").on('change',function() {
-    if ($("#project-name-input")[0].value.length === 0) {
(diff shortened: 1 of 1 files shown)


Abbe98 (approved):

wetneb:
Thanks for taking care of it!

PR #5781, opened 2023-04-11
Made it so numeric facets are properly refreshed when switching tabs

elebitzero:
- Made it so numeric facets are properly refreshed when switching tabs from the undo/redo history tab.

Before:
Numeric slider facet before

After:
Numeric slider facet after

### DIFF
--- a/main/webapp/modules/core/scripts/project/browsing-engine.js
+++ b/main/webapp/modules/core/scripts/project/browsing-engine.js
@@ -198,6 +198,9 @@ BrowsingEngine.prototype.addFacet = function(type, config, options) {
          if (facet.facet.checkInitialHeight) {
            facet.facet.checkInitialHeight();
          }
+         if (facet.facet instanceof RangeFacet) {
+           facet.facet.render();
+         }
        }
      }
   });
--- a/main/webapp/modules/core/scripts/project/browsing-engine.js
+++ b/main/webapp/modules/core/scripts/project/browsing-engine.js
@@ -195,11 +195,10 @@ BrowsingEngine.prototype.addFacet = function(type, config, options) {
      let activeTabId = ui.newTab.children('a').attr("href");
      if (activeTabId === '#refine-tabs-facets') {
        for (let facet of this._facets) {
-         if (facet.facet.checkInitialHeight) {
-           facet.facet.checkInitialHeight();
-         }
-         if (facet.facet instanceof RangeFacet) {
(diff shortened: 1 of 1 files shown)


main/webapp/modules/core/scripts/project/browsing-engine.js
```diff
@@ -198,6 +198,9 @@ BrowsingEngine.prototype.addFacet = function(type, config, options) {
          if (facet.facet.checkInitialHeight) {
            facet.facet.checkInitialHeight();
          }
+         if (facet.facet instanceof RangeFacet) {
+           facet.facet.render();
```
wetneb, line 202: In the future I would like to make facets properly extensible, so that extensions could define their own facets. With that in mind, I would generally try to avoid having code in the core of the code that are specific to a given facet. Would it make sense to render the facet regardless of its type, or do you see any complications with that?
I can imagine that facets defined by extensions could also want to be rendered when the facet tab is activated.

wetneb (commented):
Thanks a lot!

main/webapp/modules/core/scripts/project/browsing-engine.js
elebitzero: I had similar thoughts. One issue is `render()` is not in the base class `Facet()`, and `ListFacet()` does not define the `render()` method.
 
 But we can still make it more general and less brittle by checking if the facet defines `render()`, and if so, call it.
 Otherwise, we still have a special case for `ListFacet`.

wetneb: Or have a default `render()` method in the base `Facet()` class?

main/webapp/modules/core/scripts/project/browsing-engine.js
```diff
@@ -195,7 +195,9 @@ BrowsingEngine.prototype.addFacet = function(type, config, options) {
      let activeTabId = ui.newTab.children('a').attr("href");
      if (activeTabId === '#refine-tabs-facets') {
        for (let facet of this._facets) {
-         if (facet.facet.checkInitialHeight) {
+         if (facet.facet.render) {
+           facet.facet.render();
+         } else if (facet.facet instanceof ListFacet) {
```
wetneb, line 200:
```suggestion
         } else if (facet.facet.checkInitialHeight) {
```
Would that also make sense here?

wetneb (approved):

PR #5784, opened 2023-04-14
Add reconciliation service's documentation link

### LINKED ISSUE #5558: Link to reconciliation service's documentation
Some reconciliation services has a link to documentation specified in their service manifest. Currently an OpenRefine user would need to do the following to go to the documentation from OpenRefine:

1. click the "Access service API" link in the main reconciliation dialog
2. browse the JSON manifest to see if there is a documentation link
3. click it

It should be easier.

### Proposed solution
The main reconciliation dialog should have a "Documentation" link next to the "Access service API" link.

ayushrai206:
Fixes #5558 

- adds the link to service API's documentation beside access service api link
- The first commit made some unnecessary changes in the package-lock.json file,so the second commit fixed that.
-


main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -98,7 +99,10 @@ ReconStandardServicePanel.prototype._constructUI = function() {
   this._elmts.typeInput.attr('aria-label',$.i18n('core-recon/type'))
 
   this._elmts.rawServiceLink.attr("href", this._service.url);
-
+  fetch(this._service.url)
```
Abbe98, line 102: Rather than fetching the manifest a second time we should pick up the link from the existing data, [looking at the registration I guess we already have it in there](https://github.com/OpenRefine/OpenRefine/blob/0ade8a310716d3115eef5ae711170e1d02358350/main/webapp/modules/core/scripts/reconciliation/recon-manager.js#L81).

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
```diff
@@ -1,11 +1,15 @@
 <div class="recon-dialog-service-panel recon-dialog-standard-service-panel">
   <div class="grid-layout layout-normal layout-full grid-layout-for-ui"><table>
     <tr>
+      <td colspan="6" style="text-align: right;">
```
Abbe98, line 4: Lots of services does not have a documentation link and we must handle that as well, as it is currently it would just create a broken link for services missing the property.

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
```diff
@@ -1,11 +1,15 @@
...
+        <a href="" target="_blank" bind="documentationLink"><span bind="or_proc_accessDocumentation"></span>
+        <span class="ui-icon ui-icon-link" style="display: inline-block"></span></a>
```
Abbe98, line 6: We might want to have a different icon for the documentation link.

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -98,7 +99,10 @@ ReconStandardServicePanel.prototype._constructUI = function() {
...
+  .then(response => response.json())
+  .then(apiData => this._elmts.documentationLink.attr("href",apiData.documentation));
```
Abbe98, line 104: This would throw an error and crash if used on a service without a "documentation" property.

Abbe98 (changes requested):

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
ayushrai206: Hey, here is the screenshot of how i handled this 
Screen Shot 2023-04-18 at 3 12 41 AM
it works well!

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
ayushrai206: I removed the icon!

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
ayushrai206: Hey, unable to get this ,isn't the link you are pointing to pushes the reconciliation services into the stack?

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
wetneb: I would rather not display any link if there is no documentation URL available.

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
Abbe98: Yes, but it should push as I read it the entire manifest object, including the documentation property.

ayushrai206: what do you think about this?
Screen Shot 2023-04-19 at 12 53 39 AM

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
ayushrai206: I have hidden the link if there is no documentation url available, please see the screenshot in above conversation.

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
wetneb: Hi @ayushrai206, just a general comment about communication: please do not take screenshots of code, but rather submit that code in the pull request by committing and pushing it. Alternatively, you can use the "suggest changes" function to embed small code snippets in your PR comments. Or simply a code block in a comment. If none of those options work for you, it would still be better to share your code in a textual format (for instance by uploading it to another platform), shared in any other way.

We cannot copy a screenshot and put it in our local copy of the code to try it out. We cannot copy parts of it to quote them in our reply to you. If people use screen readers, they will also have a hard time. So please: no screenshots of textual content here.

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
ayushrai206: i am sorry, will not upload screenshots from now onwards, also i pointed you towards the wrong screenshot, i meant the screenshot in https://github.com/OpenRefine/OpenRefine/pull/5784#discussion_r1170481268 conversation, but leave, it i am gonna do a pr right away .

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
Abbe98: @ayushrai206 you are looking for the CSS "display" property rather than "visibility". You can also use jQuery's "hide()" method.

ayushrai206: @Abbe98 done!

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -98,6 +99,22 @@ ReconStandardServicePanel.prototype._constructUI = function() {
   this._elmts.typeInput.attr('aria-label',$.i18n('core-recon/type'))
 
   this._elmts.rawServiceLink.attr("href", this._service.url);
+// Hide the documentation link initially
+this._elmts.documentationLink.css("display", "none");
```
Abbe98, line 103: I would suggest to refector this so that the documentation link is inserted into the DOM if there is a link.

It's more aligned with the rest of the code base and would get rid of issues like, leftover elements and non-default display values.

Abbe98 (changes requested):
See the outstanding issue as well as my new comment.

wetneb:
I tried out this PR interactively and noticed that the layout of the dialog is not ideal:
image


Because you added the link as an extra `<td>` in this table-based layout, this creates an extra column which is empty for the rows below, creating a lot of unused whitespace. We should avoid that.

ayushrai206:
> I tried out this PR interactively and noticed that the layout of the dialog is not ideal: image
> 
> Because you added the link as an extra `<td>` in this table-based layout, this creates an extra column which is empty for the rows below, creating a lot of unused whitespace. We should avoid that.

Hey, as discussed in our meet earlier i removed the access service api option and added the service documentation option ,does 'service documentation' seem alright or it should be named as only 'documentation'?

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -418,6 +418,7 @@
     "core-project/undo": "Undo",
     "core-recon/add-std-srv": "Add standard service",
     "core-recon/access-service": "Access service API",
+    "core-recon/service-documentation":"Service Documentation",
```
wetneb, line 421:
```suggestion
    "core-recon/service-documentation":"Service documentation",
```
For consistency with the casing we use in the rest of the application.

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
```diff
@@ -1,11 +1,15 @@
 <div class="recon-dialog-service-panel recon-dialog-standard-service-panel">
   <div class="grid-layout layout-normal layout-full grid-layout-for-ui"><table>
     <tr>
+      <td colspan="2" style="text-align: right;">
+        <a href="" target="_blank" bind="documentationLink"><span bind="or_proc_accessDocumentation"></span>
+        <span style="display: inline-block"></span></a>
+      </td>
       <td colspan="2" style="text-align: right;">
         <a href="" target="_blank" bind="rawServiceLink"><span bind="or_proc_access"></span>
-        <span class="ui-icon ui-icon-link" style="display: inline-block"></span></a>
```
wetneb, line 6: It looks like you removed the existing "Access service API" link but you kept its `<td>`, and you added another `<td>` for the documentation link. This means that you are still deforming the table by creating a new column that is empty in the other rows.
I would just reuse the existing `<td>` element instead.

wetneb (commented):
Thanks! It looks much simpler like this. I think the DOM structure can still be improved here.
Also, we need to validate the design proposal with the community, so I will bring up the topic on the issue.

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.html
ayushrai206: I removed the TD element please have a look!
Thankyou

ayushrai206:
> Very nice! I had a try and it looks like we have this "link" icon for the API link but not for the documentation link. I would intuitively expect that both links are styled in the same way, since they are both external links. That means either display the icon for both of them or for none of them. Up to you!
> 
> image

Thank you very much! I have added the icon for both links it seemed nice to me that way!

wetneb (approved):

Issue #5431, opened 2022-11-12
Import project archive from URL


ookgezellig:
From the OR start screen you choose "Import project", which enables the user to select a local project tar.gz file for upload into OR

It would be handy to also have the possiblilty the import a project archive directly from URL, such as (for instance) https://github.com/KBNLwikimedia/OpenRefine-WikimediaCommons-Workshop/blob/main/Dataset-WorkshopOpenRefine-WikiconNL19112022-xlsx.openrefine.tar.gz?raw=true

tfmorris:
Sounds like a reasonable request and we have all the pieces, so it's just a matter of wiring them up. I wouldn't be surprised if writing the tests was harder than the implementation itself.

AbrahamRH:
Hi, I'm interested in contributing to this issue. Can I work on it?

SoniaSun810:
Hey, dear @wetneb ! Can you assign this issue to my group? We've been actively working on it and have developed a promising solution.

PR #5796, opened 2023-04-19
Creating the URL input to import a project. (Draft)

AbrahamRH:
Fixes [#5431](https://github.com/OpenRefine/OpenRefine/issues/5431)

Hi, I create the input in the form to import a project file using a URL instead a tar file.

In this draft I only put the form input of the url:

I would like to know what you think about downloading the file and attaching it to the file field to then send the form so that the project is imported, like this:

Or maybe passing the url to the backend, but I have not been able to find the file where the form data is sent and processed.


wetneb:
Hi @AbrahamRH, thanks a lot for having a go at this!

I would intuitively recommend against downloading the file in the frontend and sending it to the backend. Project archives can be relatively large files so it's worth minimizing the transfers being done, so I think we should rather download it directly in the backend.  This would probably mean adding a way to report the progress of this download like in the workflow where we create a project from a URL, which can be quite some work given that we don't have anything like this yet.

This might suggest that we could remove the distinction between creating a project (from a generic file, say a CSV) and importing an existing project archive. We could aim to have only one mechanism to create a project, which would support OpenRefine project archives as one of the possible formats. This would have the benefit that we do not need to re-implement all this logic and already benefit from the URL fetching support there. This could be the occasion to fix other issues: for instance, importing a project archive currently just unzips the given archive in the workspace, without checking anything about its format. This is not ideal because we are potentially creating an invalid project by doing that.

I realize that this opens a can of worms that feels much more complicated than the issue you wanted to tackle, but those are just my intuitive thoughts looking at your attempt and the issue.

@tfmorris has tagged the issue as "good second issue" and seemed to have a much simpler solution in mind, so maybe I am overthinking this?

AbrahamRH:
Yeah! thank you @wetneb , as you mention it seems a little more complicated implementing a completely new feature, maybe at the moment I can't complete it but I can advance a part of the issue by sending the data to the backend.

Or maybe I can take a closer look of how the creation of a project works using a URL and work on it a bit.

wetneb:
Thanks for the hard work!

For people who want to test it, they can use this URL to import a sample project archive:
https://pintoch.ulminfo.fr/6e551ee9a1/Changing-the-type-of-the-column-for-numeric-1680159450732.openrefine.tar.gz

wetneb (commented):
The import panel looks like this for me (Linux, Firefox):
image

I would prefer the UI that you proposed above (in your first screenshot).
Apart from that it seems to work well! Perhaps the progress bar is not such an important feature, since we don't have one when uploading a project archive as a file. But I would at least disable the "Import project" button once it has been pressed and while we are waiting for the request to complete (to prevent the user from submitting the form multiple times). Also, it would be amazing to have tests for this new command. You should be able to write some using `MockWebServer` (look for usages of that class in the rest of the code base), so that you can spin up a local webserver serving a sample project archive.

wetneb:
Superseded by #5892

PR #5801, opened 2023-04-21
Optimize project deletion after Cypress tests

### LINKED ISSUE #5800: Project deletion after Cypress tests is inefficient
After a Cypress test is run, we attempt to delete all projects created in the test.
When multiple tests are run in the same runner, the list that keeps track of which projects were created grows longer and so we attempt to delete many projects which were already deleted earlier. This is inefficient and can be cleared.

### To Reproduce
1. Run all Cypress tests locally (using the "Run all 80 specs" option)

### Current Results

After running a couple of tests, you can observe that the clean-up phase is making many requests to the backend to delete projects which already no longer exist.

### Expected Behavior

Only the projects created in the said test are deleted.

### Screenshots

image

wetneb:
Closes #5800.

### DIFF
--- a/main/tests/cypress/cypress/support/openrefine_api.js
+++ b/main/tests/cypress/cypress/support/openrefine_api.js
@@ -158,9 +158,11 @@ Cypress.Commands.add('cleanupProjects', () => {
           body: { project: projectId },
           form: true,
         }).then((resp) => {
-          cy.log('Deleted OR project' + projectId);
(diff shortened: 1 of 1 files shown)


wetneb:
It does not seem to make a ton of difference for the CI since the projects do not accumulate as much as we are spreading the load on parallel runners, but for running locally it does speed things up quite a bit.

elebitzero (approved):

PR #5804, opened 2023-04-21
Flush precompute cache on revert of RowRemovalOperation

### LINKED ISSUE #5799: Duplicates facet computing incorrectly after undoing row removal
I often use the 'duplicates' built-in facet and recently I noticed that it was reporting incorrectly after I had undone a row removal operation. I find this is reproducible - see the attached images and the dataset I used for this demo:

1. Load a file with a column that has duplicates
2. Apply the **Facet > Customized facets > Duplicates facet** to that column and note that it shows the duplicates as a 'true' set
3. Remove rows (whichever way you like) so that there are no longer duplicates and note that the facet is now showing only 'false'. (In my example here I starred the second line, faceted by star, and removed matching, but I could have done this a little easier by faceting on Col1, selecting "B", and then removing)
4. Undo the row removal and observe that the facet is still showing only 'false' even though all the duplicates are back. You can even add a new duplicates facet on the column and it will only show 'false'
Original load
Undo row removal
Facet not returning to original state

Note that in my use case I often modify the facet to show the duplication count (i.e. `facetCount(value, 'value', 'Col2')`) but the problem is the same.

### Versions
 - Operating System: MacOS 12.6.5
 - Browser Version: Safari 16.4.1
 - OpenRefine: 3.7.2 [f7ad526]

[Simple.txt](https://github.com/OpenRefine/OpenRefine/files/11293737/Simple.txt)

tfmorris:
Fixes #5799
- consolidate CacheTests into RowRemovalOperationTests
- DRY up verification code
- flush precompute cache on revert of row removal and extend test to cover this case


wetneb (commented):
Thanks for tackling this! Intuitively, any change to the grid has the potential to render those caches stale, no?

So I think the invalidation of those caches should not be left to each change, but rather done centrally, whenever a change is applied or undone.
I'd rather remove this logic from those changes and put it just once, for instance in the `History` or `Project` class. This way, there would not be any risk that the implementer of a Change forgets to invalidate those. What do you think of that?

tfmorris:
> Intuitively, any change to the grid has the potential to render those caches stale, no?

I don't know what a `ProjectLookup` is, so I can't answer the general question, but the historical cache is a per-column cache, so only needs to be flushed when values in the column change. The `RowRemoval` operation is a special case because it affects all columns.

As far as I know, operations don't register the columns that they've modified, but if you know of a central place that has this knowledge, please suggest it and I'll investigate.

wetneb:
Indeed, so far there is no mechanism in place to determine which columns are touched by an operation. Therefore I would recommend to invalidate those caches whatever the change is, just to be on the safe side. What do you think of the `History::setModified` method to do this invalidation? It is already called for any application or rollback of a change, so it feels intuitively appropriate to me, but I have not tried it out.

The `ProjectLookup` is the cache that underpins the `cross` function.

By the way, my next major task for the 4.0 architecture is to introduce the required metadata to determine which columns/rows a change touches, so we should be able to have a more informed cache invalidation there.

tfmorris:
The vast majority of operations operate on a single column only. These all use `Column.clearPrecomputes()`. Only a very few operations affect more than one column. These use `ColumnModel.clearPrecomputes()` and it seems wasteful to switch to using this all the time.

In any case, while refactoring/redesigning the caching is certainly possible, I think it's beyond the scope of this simple bug fix. Similarly, I wouldn't attempt to replace uses of `LookupCacheManager.flushLookupsInvolvingProjectColumn()` with `LookupCacheManager.flushLookupsInvolvingProject()` or attempt to deduplicate the two difference column caching mechanisms.

wetneb (approved):
Sure, happy to keep it granular then.

PR #5807, opened 2023-04-21
Reenable Coveralls plugin. Fixes #5013

### LINKED ISSUE #5013: Re-enable Coveralls in the CI
Our submission of coverage info to Coveralls is broken, so I have disabled it.
It would be great to fix the underlying issue or migrate to another coverage info submission mechanism.
https://github.com/trautonen/coveralls-maven-plugin/issues/147

tfmorris:
This reverts commit bfa8c9af756c5efe570b653067811eadd57bd12d.

Fixes #5013 

- Reenable Coveralls plugin

Seems to work now, so let's see how reliable it is.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -88,6 +88,10 @@ jobs:
     - name: Build and test with Maven
       run: mvn jacoco:prepare-agent test
 
+    - name: Submit test coverage to Coveralls
+      run: |
+        mvn jacoco:report coveralls:report -DrepoToken=${{ env.COVERALLS_TOKEN }} -DpullRequest=${{ github.event.number }}
+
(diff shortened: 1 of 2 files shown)


wetneb:
Thanks! This will only submit coverage info on pull requests, right? I suspect we need to do it for builds on the master branch too if we want to be able to get coverage comparisons between master and PRs. How about reverting https://github.com/OpenRefine/OpenRefine/commit/7ec03039db028d52adeac7e8d6b650a919e26005 too?

wetneb (approved):

wetneb:
I have disabled the PR messages that just started showing up (see #5892) because it feels a bit spammy and I think we can use the PR check Coveralls adds.

PR #5814, opened 2023-04-24
Widen scrutinizer regex pattern - fixes #5656 

### LINKED ISSUE #5656: Wikimedia Commons Wikibase not detecting non-Latin script characters
I've been trying to upload a batch of files that have names containing both Latin and [Odia](https://en.wikipedia.org/wiki/Odia_script) characters. For instance, Wikimedia Commons allows filenames with Odia characters (for instance, this file](https://commons.wikimedia.org/wiki/Category:Audio_files_created_using_Kathabhidhana), but OpenRefine does not. See the screenshot of the error I see below. cc @wetneb

### To Reproduce
1. Create a file with a file name such as "ଫାଇଲ.wav" (that's a WAVE file with a filename with in the Odia script)
2. Try to upload to Wikimedia Commons using OpenRefine. It fails at the Wikibase (Upload edits to Wikibase)

### Current Results

It fails (see below)

OpenRefine-issue-with-Odia

### Expected Behavior

The upload should normally happen, just like any other file.

### Versions
 - Operating System: MacOS
 - Browser Version: Brave 1.46.171
 - JRE or JDK Version: 
 - OpenRefine: 3.7

### Datasets

1. [Sample dataset containing the same files I was trying to upload](https://drive.google.com/drive/u/0/folders/1lyg_63oZCd7fIwWFvqU4v0AwKOIHK8FQ)

santi4o:
Fixes #5656 

Hi there!
I've been working on the issue, thanks for your help in the issue thread

- I updated the regex to only disallow control characters, \\p{Cc} will match control characters (ranges 0x00-0x1F and 0x7F-0x9F)
- I updated the validation for the file name length in bytes, encoded as UTF-8
- I updated the tests, adding some that should raise warnings and disabling testInvalidCharactersInFilenameVerticalBar and testInvalidCharactersInFilenameHTMLEscaped

Please let me know what you think

### DIFF
--- a/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizer.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizer.java
@@ -20,7 +20,8 @@ public class FileNameScrutinizer extends EditScrutinizer {
 
     // see https://commons.wikimedia.org/wiki/Commons:File_naming
     public static final int maxFileNameBytes = 240;
-    public static final Pattern forbiddenFileNameChars = Pattern.compile(".*([\\p{Cc}]).*");
+    public static final Pattern forbiddenFileNameChars = Pattern.compile(
+            ".*([^ %!\"$&'()*,\\-./\\d:;=?@\\p{L}\\p{Mc}\\\\^_`~\\x80-\\xFF+]|%[0-9A-Fa-f]{2}|&[A-Za-z0-9\\x80-\\xff]+;|&#[0-9]+;|&#x[0-9A-Fa-f]+;).*");
 
     public static final String duplicateFileNamesInBatchType = "duplicate-file-names-in-batch";
     public static final String fileNamesAlreadyExistOnWikiType = "file-names-already-exist-on-wiki";
--- a/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizer.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizer.java
@@ -90,9 +90,7 @@ public class FileNameScrutinizer extends EditScrutinizer {
                     issue.setProperty("max_length", Integer.toString(maxFileNameBytes));
                     addIssue(issue);
                 }
-            } catch (UnsupportedEncodingException e) {
-                e.printStackTrace();
-            }
+            } catch (UnsupportedEncodingException e) {}
 
             // Invalid characters
             Matcher matcher = forbiddenFileNameChars.matcher(fileName);
--- a/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizer.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizer.java
@@ -90,7 +90,8 @@ public class FileNameScrutinizer extends EditScrutinizer {
                     issue.setProperty("max_length", Integer.toString(maxFileNameBytes));
                     addIssue(issue);
                 }
-            } catch (UnsupportedEncodingException e) {}
+            } catch (UnsupportedEncodingException e) {
+            }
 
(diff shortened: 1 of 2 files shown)


extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizer.java
```diff
@@ -78,15 +78,20 @@ public void scrutinize(MediaInfoEdit edit) {
...
+                if (utf8Bytes.length > maxFileNameBytes) {
+                    QAWarning issue = new QAWarning(fileNameTooLongType, null,
+                            QAWarning.Severity.CRITICAL, 1);
+                    issue.setProperty("example_filename", fileName);
+                    // This is only accurate for 1 byte characters
+                    issue.setProperty("max_length", Integer.toString(maxFileNameBytes));
+                    addIssue(issue);
+                }
+            } catch (UnsupportedEncodingException e) {
+                e.printStackTrace();
```
wetneb, line 94: I'd remove this `e.printStackTrace()` because it would be printed once for each row every time the issues are refreshed. I think it's safe to just ignore this exception altogether because if UTF-8 is not supported on the system we have bigger problems than not being able to report a quality assurance warning.

wetneb (commented):
Very nice, thank you! Perhaps it would be worth adding a test showing that the filename mentioned in the issue ("ଫାଇଲ.wav") is accepted.

extensions/wikibase/tests/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizerTest.java
```diff
@@ -67,7 +67,67 @@ public void testLongFileName() throws IOException, MediaWikiApiErrorException {
...
+    public void testInvalidCharactersInFilenameBackspace() throws IOException, MediaWikiApiErrorException {
+        MediaInfoEdit edit = new MediaInfoEditBuilder(TestingData.newMidA)
+                .addFileName("Backspaces (\b) are not allowed.png")
+                .build();
+
+        scrutinize(edit);
+        assertWarningsRaised(FileNameScrutinizer.invalidCharactersInFileNameType);
+    }
+
+    @Test(enabled = false) // Vertical bars should be allowed
```
wetneb, line 130: Where did you see that vertical bars should be allowed? I just tried uploading a file to Commons with a vertical bar and it is refused.

extensions/wikibase/tests/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizerTest.java
```diff
@@ -77,7 +137,7 @@ public void testInvalidCharactersInFilenameVerticalBar() throws IOException, Med
         assertWarningsRaised(FileNameScrutinizer.invalidCharactersInFileNameType);
     }
 
-    @Test
+    @Test(enabled = false) // HTML escaped entities should be allowed
```
wetneb, line 140: Same here - I just tried uploading something with this filename and it fails with the error:
`This title is invalid. Make sure to remove characters like square brackets, colons, slashes, comparison operators, pipes and curly brackets.`

wetneb (commented):
It looks like the regex might be a bit too wide now, see below.

santi4o:
You're right, my bad. I read in the issue comments a possible solution involving only disallowing control characters. I was able to reproduce the issue and only tested directly in OpenRefine with Odia-script file names.
This is what I found on [Language, character set and extension
](https://commons.wikimedia.org/wiki/Commons:File_naming): "Avoid "funny" symbols (control characters, unneeded punctuation, etc.) that might be significant in future wiki markup. It is a good idea to stick to [graphemic](https://en.wiktionary.org/wiki/graphemic) characters, numbers, underscore (space), ASCII hyphen/minus/dash, plus, and period (dot)." 
I will leave an update. Thank you

wetneb:
Hi @santi4o, are you still working on this? I think you are nearly there and it would be amazing to fix this!

wetneb (approved):

PR #5820, opened 2023-04-27
removes some unused imports to reduce noise

thadguidry:
- implict loaded from same packages, so no need for explicit
- otherwise also removes some unused testing and debug imports


Abbe98:
Some of these are imports for code that is commented out. Is it worth leaving those in given that it will not be very clear from where such code is importing from?

wetneb (approved):
In my opinion we should not have commented out code in the first place, so I am not worried about that. Both Eclipse and IntelliJ will be happy to help anyone add the missing imports if this code had to be uncommented.

I also wish our linter could take care of removing unused imports, but I don't think it supports that sadly.

Issue #3224, opened 2020-09-26
Allow easier install and uninstall on Windows


thadguidry:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24500%20Rewarded-%237E24E3.svg" />](https://oss.issuehunt.io/r/OpenRefine/OpenRefine/issues/3224)

On Windows, OpenRefine currently has a manual process to extract, install, and create desktop shortcuts and configuration.
 
### Proposed solution
- Create a native installer for OpenRefine on Windows (one that produces an .exe or .msi) that runs without any dependency or other software needed to be preinstalled.
- Install and Uninstall should be available in Add/Remove Programs for OpenRefine.
- Ensure custom folder path input is available during install.
- Have option to create desktop shortcut and add OpenRefine to Start menu.
-

#### INNO
----

[Inno Setup](https://jrsoftware.org/isinfo.php) is the most likely candidate for helping packaging a proper installer for Windows with automatic desktop shortcut handling and uninstall.
Inno provides automation through both Inno Setup Scripts (.iss files) as well as Pascal scripting (modern Delphi-like Pascal)

Licensing - Inno seems to have a compatible license, and a simple acknowledgement in our docs is all that is needed, I think.  Item 3 in Inno Setup License: https://github.com/jrsoftware/issrc/blob/main/license.txt

CI/CD and packaging methods - I think using the GitHub Actions VM image Windows Server 2019 would provide the necessary toolkits that are required for Inno Setup as it already includes InnoSetup 6.0.5 (latest stable version) as well as the .Net CORE components.  See https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md

#### NSIS
----

[Nullsoft Scriptable Install System - SourceForge.net](https://sourceforge.net/projects/nsis/) (NSIS)
https://github.com/NSIS-Dev

There is a Maven plugin: https://github.com/DigitalMediaServer/nsis-maven-plugin (which seems to be cross-platform itself too)
VSCode has the NSIS Plug-ins for IntelliSense https://github.com/idleberg/vscode-nsis-plugins

### Alternatives considered
[WiX Windows Installer](https://wixtoolset.org/) was considered but it doesn't leverage the latest Windows 10 SDK handling.

### Additional context
none for now

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $500.00)

[...]

PR #5833, opened 2023-05-10
Add Windows installer using InnoSetup 6

dori4n:
Possible fix for #3224 😘

This adds an InnoSetup script and modifies the GitHib Actions workflow to run the necessary steps that produce a self-contained installer file from the contents of the ZIP file of the **Windows (with Java)** build artifact.

So far this is untested on the GitHub Actions CI, but a proper installer is produced like this, when the InnoSetup Command Line Compiler is invoked manually with `"%ProgramFiles(x86)%\Inno Setup 6\iscc.exe" -DMyAppVersion="3.8-SNAPSHOT" -DMyProgramFiles="target\openrefine-3.8-SNAPSHOT" "packaging\openrefine.iss"` from the repository root after manually copying the contents of the folder in the ZIP file to `packaging\target\openrefine-3.8-SNAPSHOT` (that's what I got from running `mvn -B deploy -DskipTests -Dgpg.skip`).

It's still running at https://github.com/dorianmuthig/OpenRefine/actions/runs/4931761252 but I can obviously not properly test this, because of reliance on the digital signature workflow.

Also it seems like the Windows action does not wait for the Mac action to finish on first glance.
So that this can run on autopilot, what would you prefer?
* ~~build on Windows, and thus build twice, or~~
* [x]  find a way to make it wait until the macos build process has finished

Any other suggestions and feedback?

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#3224: Allow easier install and uninstall on Windows](https://oss.issuehunt.io/repos/6220644/issues/3224)
---
</details>


dori4n:
Okay, I forgot the `needs: <job_id>` directive.

Seems to looks right now: https://github.com/dorianmuthig/OpenRefine/actions/runs/4932031312

Still can't test the workflow further, but I do expect it to work like this.

thadguidry:
@dori4n Hi and thanks!
I don't think we expect you to fully modify our snapshot release process!  That's something that we can certainly do later on our side.
I think for now, just create a brand new GitHub workflow file on your end... call it `windows_release.yml` and add this as a commit and revert the changes on our `snapshot_release.yml` file for now.

Then we'll be able to check out your branch... run that particular `windows_release.yml` workflow file (and any of your other INNO setup files) to verify in our own private repos that indeed a INNO package is created successfully and I can run it on my Windows 11 machine and do more testing.

Later on we can modify our own CI release process.  We can reuse your `windows_release.yml` as a step in our `snapshot_release.yml`  ... which is done through this process https://docs.github.com/en/actions/using-workflows/reusing-workflows and which we'll gladly take care of on our side.  We can call that `windows_release.yml` and even pass data down into it including secrets, keys, envs, etc.  Which is something you don't have to worry about.

wetneb:
I think we can already work with what is here, I don't think there is a need for a new GitHub workflow file as it would duplicate many steps that we currently have :)

Thanks a lot @dori4n, I will try it out in the coming days and get back to you.

wetneb:
One question: why unzip the artifact in the MacOS job, and upload it (as a duplicate artifact) to the release?
Would it not be simpler to unzip the artifact within the Windows job directly, removing the need for this duplicate artifact?

dori4n:
@wetneb 
> One question: why unzip the artifact in the MacOS job, and upload it (as a duplicate artifact) to the release?

It was easier to do it this way. A GitHub Actions workflow artifact is like a file share, and the upload and download is fast. Going through the release artifacts and downloading them from there and then unzipping takes more time, but also requires using PowerShell on Windows, because a native terminal `unzip` command does not exist, or 7-Zip (which [does come pre-installed](https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#tools "runner-images/Windows2022-Readme.md at main · actions/runner-images · GitHub")).
Also, the ZIP archive generated is generated in such a way that the files in the archive are placed into a directory inside the archive. Dealing with one less naming conflict is one less headache to worry about.
> Would it not be simpler to unzip the artifact within the Windows job directly, removing the need for this duplicate artifact?

It depends. It can be. Right now, and keeping in mind to change your existing workflow as little as possible, it wasn't. And since workflow artifact folders are different from release artifact files, this was to the best of my knowledge, the faster option (though, I don't know if that is still true, it's been a while).

wetneb:
Ok, got it, I had not paid attention to the difference between release and workflow artifacts. That makes sense to me.

With this proposed workflow, the installer will only be published for official releases. I wonder if it should also be published for snapshot releases. We currently publish all four packaged variants (MacOS, Linux, Windows with java and Windows without Java, both as zip files) in Maven Central for snapshot releases, so it would feel consistent to have the installer there too. It would at least make testing the pipeline easier!

I am not too sure how to go about this though. We could call Maven again after your workflow, registering the generated exe as an artifact and deploying it to Maven Central, but that seems rather complicated.

For testing purposes, why don't you upload the generated .exe as a workflow artifact (in all cases, or just when it's not an official release)? I would then merge this so that we can test out the workflow in its intended environment. We think again after how to best publish this afterwards.

dori4n:
@wetneb 
> why don't you upload the generated `.exe` as a workflow artifact (in all cases, or just when it's not an official release)?

Like this? 🥳

dori4n:
I did some more experimenting in a separate branch where I have turned off the GPG signing process and skipped OSSRH deployment, and I've discovered that the generated file name of the build artifact always ends in `-3.8-SNAPHOT`, instead of the actual version number and commit hash. I don't think that's ideal.
But I also don't know where to change that and how to the adapt the pattern in to the name of the build artifact.

In fact, this is even true for a release build, such that I cannot believe the workflow as present in the `master` branch was even working to publish a release as is.
I'm sorry. I'm getting a little confused now.

I can fix your GitHub workflow, but I feel like this is getting a little out of scope for this issue now.
The installer works, and so does the GitHub workflow to create one, even if the file paths may be wrong in a release build.

On the test branch with signing disabled:

Commit build: https://github.com/dorianmuthig/OpenRefine/actions/runs/4961976622#artifacts
Release build: https://github.com/dorianmuthig/OpenRefine/actions/runs/4961996873 (I cheated and created a tag with `#3.8-SNAPSHOT` for this)
Release artifacts: https://github.com/dorianmuthig/OpenRefine/releases/tag/3.8-SNAPSHOT

wetneb:
It is normal that the files generated by the Maven workflow only have `3.8-SNAPSHOT` in their name. The additional metadata is added by Maven Central after upload. You can see that the files available for download have a datestamp in them (just try downloading any of the [snapshot releases](https://github.com/OpenRefine/OpenRefine#snapshot-releases)).

dori4n:
@wetneb 
>It is normal that the files generated by the Maven workflow only have `3.8-SNAPSHOT` in their name

That is not what I meant. I mean the steps at [`.github/workflows/snapshot_release.yml`, lines 239 onwards](https://github.com/OpenRefine/OpenRefine/blob/master/.github/workflows/snapshot_release.yml#L239 "OpenRefine/snapshot_release.yml at master · OpenRefine/OpenRefine · GitHub") (on the `master` branch), where the release artifacts are uploaded *to GitHub* when you publish a tagged release.
Even the workflow script *as is* without my changes would fail, because the `asset_path` uses `${{ steps.get_release_upload_url.outputs.tag_name }}` in the file name, even though it is `3.8-SNAPSHOT` in the output file name, if the tag is set to something else.
The Maven build does include the current commit hash into the build artifact and this is shown in the bottom left corner of the web interface. This means that it should be possible somehow to have it use the right file name.
But I can also just change it to `3.8-SNAPSHOT` and be done with it.

wetneb:
When we do a release, the name of the tag is equal to the version which is stored in the `pom.xml` files, so I am not sure where you see a problem. I do think it works, that's how we have been publishing releases for a while now :) Anyway, if you want to make any change to this, we can consider it, but it's probably irrelevant to this PR so we should be discussing that somewhere else.

dori4n:
@wetneb

Hm... okay.

Then I will leave it as it is now. Just don't forget that you will have to change all occurrences of `3.8-SNAPSHOT` in the workflow file when you change it. I just have a strong preference to having versioning artifacts in as few places as possible, so that you only need to change them once.

It is possible to also upload the other build artifacts as GitHub Actions artifacts in addition to the Maven repository, but I don't think that this is necessary or within scope. I might do that, if I get bored today, maybe.
I don't know how to upload the installer file to the Maven repository, and I cannot test that anyway, so it's better if you do that.

You will probably want to digitally sign the executable file, too, because if you don't Windows SmartScreen will be complaining and preventing users from running the installer after downloading it. I tried it with the workflow artifacts from my tests in a newly installed Windows VM, and I had to click through 5 different prompts to *keep the downloaded file* in Edge and then run it, and I would never recommend anyone who doesn't know exactly what they are doing to click past any of that.

wetneb:
> Then I will leave it as it is now. Just don't forget that you will have to change all occurrences of 3.8-SNAPSHOT in the workflow file when you change it. I just have a strong preference to having versioning artifacts in as few places as possible, so that you only need to change them once.

Ah right, of course the new workflow steps that this PR introduces should not have a hard-coded version number in them. (But that's not a problem with the existing workflow steps, right?)
We could simply store the version number in an environment variable earlier in the workflow. It can be computed with:
```
mvn org.apache.maven.plugins:maven-help-plugin:3.1.0:evaluate -Dexpression=project.version -q -DforceStdout
```

wetneb:
@dori4n what do you think about the approach above? I think it would be good to deduplicate the steps you have introduced, so that we just extract the version number once and then use it in the following steps, regardless of whether it is a snapshot or final release.

dori4n:
@wetneb I've had other things to deal with so I couldn't look at it, yet.

> We could simply store the version number in an environment variable earlier in the workflow. It can be computed with:
> 
> ```
> mvn org.apache.maven.plugins:maven-help-plugin:3.1.0:evaluate -Dexpression=project.version -q -DforceStdout
> ```

I'm going to have to look into how to make that work. All the old ways of passing string data between workflow steps are deprecated and I also don't know if passing values between systems and runners' instances is possible using the [new way](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files "Workflow commands for GitHub Actions - GitHub Docs").

> what do you think about the approach above?

Sounds good in theory.

> I think it would be good to deduplicate the steps you have introduced, so that we just extract the version number once and then use it in the following steps, regardless of whether it is a snapshot or final release.

Well, the way it is now, my changes do add a duplication for the steps that use the version number and also to decide whether to attach a workflow artifact or a release artifact for the installer, depending on whether it is a commit run or a release run, respectively. But like I said, I can't imagine `${{ steps.get_release_upload_url.outputs.tag_name }}` being replaced with the proper value when you publish a release, unless you do this manually and adjust the `pom.xml` accordingly, just before you do.
You can't just create a tag on a commit and then publish the release on GitHub like this, because this definitely fails: https://github.com/dorianmuthig/OpenRefine/actions/runs/4962428433/jobs/8880469348

I'll go get some sleep, and then I'll look into whether I can get rid of the duplications, and ideally make it such that publishing a release on GitHub is as easy as creating a commit tag and hitting publish on the website for it, like I suspect this was supposed to work. 😉

wetneb:
Ok! Perhaps this page is helpful to understand our current workflow:
https://openrefine.org/docs/technical-reference/version-release-process

thadguidry:
@dori4n polite nudge.  What the state of things?  Anything our team can further help you along with on this?  We'd love to see you claim the bounty.  If the blocker is truly about GitHub Actions or limitations you are struggling with, we can even invite GitHub Support here to help you out more. (I know a few folks)

dori4n:
I'm sorry, I've been sick. And then I kinda forgot about it for the past week while dealing with other things.

My thoughts on how to go on from what you can see have been:

1. Differentiating from release and push actions
2. Automatically setting the version to the release tag (possibly escaping invalid characters?), if it's a release, like step 2 from https://openrefine.org/docs/technical-reference/version-release-process
3. Using the command you provided with your comment [here](https://github.com/OpenRefine/OpenRefine/pull/5833#issuecomment-1546813530) to extract the current version number and piping it into a job step output variable (gonna have to experiment with the format of that to see how to use it properly) so that it can be used to to dynamically set the file names and the names of the artifact uploads

What I was struggling with three weeks ago was dealing with transitioning the job step output from the Mac runner to the Windows runner. The documentation says [`jobs.<job_id>.needs`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) should be enough to make [`jobs.<job_id>.outputs`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idoutputs) available, but it wasn't working properly with [bruceadams/get-release](https://github.com/bruceadams/get-release) (or maybe I made a mistake, not sure) so I put it in twice to make it work. I'm a bit worried I may not be able to do 3. Or at least not without duplication.

I've been up all night and am a bit tired now, and I really need to grab some sleep. I'll look at it again in the evening. 🙂

dori4n:
@wetneb @thadguidry
Okay, the last test is running now at https://github.com/dorianmuthig/OpenRefine/actions/runs/5402979084

If it doesn't fail, I would consider this safe to merge.

Open issues (not breaking): 
* Code signing for the Windows binaries
I can't test that, as it requires key configuration in the GitHub account, and a code signing certificate. However, recent changes to rules for certificate authorities now require code signing certificates to be generated on and used with a hardware cryptography token. I'm not sure this is compatible with GitHub actions (and the [nextgens/authenticode-sign-action](https://github.com/nextgens/authenticode-sign-action "nextgens/authenticode-sign-action: This action allows for authenticode code-signing using a key hosted on google key management service") signing tool). One of you would have to configure and test that, if it is wanted.
* How to upload the Windows installer file to OSSRH/Maven for commit snapshots
* You have made some changes to the workflow file since. I hope there are no conflicts with my changes now.

wetneb (approved):
Let's try it out! I'll merge this and test the installer. Thank you for your perseverance.

dori4n:
@wetneb 
Looks like it worked and the installer artifact is available at https://github.com/OpenRefine/OpenRefine/suites/14066426575/artifacts/786206866 🥳

wetneb:
I really like it!

Notes from my own testing
image

dori4n:
@wetneb 

Hi

> * during the install phase we are presented with a license text with "copyright Google 2010" or something like that: we should update that to include "OpenRefine contributors" as well

I have just passed in the `LICENSE.txt` file from the build output root. It is possible to omit that step from the installer entirely, too, if it doesn't contain anything the user has to agree to for merely using the software. This isn't really true considering the disclaimer in all caps in the last line of the file.
However the issue you raise is present in [all copies of the license](https://github.com/OpenRefine/OpenRefine/blob/master/LICENSE.txt "OpenRefine/LICENSE.txt at master · OpenRefine/OpenRefine · GitHub"), and not installer specific

> * At the end of the install process I was given the option to open the README file automatically (and this option was ticked by default). Because the README file is `README.md`, my computer failed to open it. We should likely have a `.txt` file instead or disable this feature entirely:
>   image

That error appears, because no application is associated with the `.md` file extension. JrSoftware recommends using `.txt` or `.rtf` files, which are associated with Notepad and WordPad, respectively, on Windows by default.
You can disable this by commenting out [line 72 in `openrefine.iss`](https://github.com/OpenRefine/OpenRefine/blob/master/packaging/openrefine.iss#L72 "OpenRefine/packaging/openrefine.iss at master · OpenRefine/OpenRefine · GitHub").

PR #5838, opened 2023-05-11
Add recommended VSCode extensions

thadguidry:
- For those checking out the project with VSCode, this helps to recommend 2 useful extensions for developing with OpenRefine and our Docs.
- https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions


wetneb (commented):
I'm not sure to what extent this is needed - isn't VSCode already able to suggest that by itself?
Also, isn't the comment in the JSON file a problem (since it's not valid JSON according to stricter versions of the specs I think)?

Issue #5541, opened 2023-01-03
Show partial results of long-running operations before they complete


wetneb:
When using long-running operations, such as:
* reconciliation
* data extension
* URL fetching
* Wikibase editing
the user initially gets a progress notification (showing the percentage of completion of the operation). It is only once the operation is complete that the results are shown.

This is not ideal, for the following reasons:
* imagine there are issues with the settings of the operation. For instance, when fetching URLs, the URL template I am using is flawed and the server responds with an HTTP error, or when reconciling, my reconciliation settings yield unhelpful scores. Because the results for the first few rows in the project are only shown when the operation is complete, I can only detect those issues at that stage. At that point, I have had to wait for a long time and have wasted resources on the service I have queried (or even my own money if I am using a paid service with an API key).
* after a reconciliation operation, I generally need a significant amount of time to review the results and match certain cells to certain items. It would really help if I could start this work as soon as the first reconciliation results are in.

### Proposed solutions

The results of the operation should appear gradually, as they are fetched. This could take various forms:
* Only show the rows/records where the operation has run (effectively offering a cropped view on the project)
* Also show the rows/records where the operation has not run yet, leaving them in their previous state (for reconciliation) or with a blank fetched value (for URL fetching and data extension)
* Visually highlight that a row has not been processed yet, for instance by showing a waiting spinner inside the affected cell(s)

There are also many different ways we could let the user work with this partially fetched view on the grid:
* Initially, just keep queuing any change made to the grid from that state (such as editing a cell or changing the judgment of a reconciled cell) so that they are only executed once the long-running operation is complete
* If the changes made are also row/record wise, then they could already be executed on the grid before the long-running operation is complete. This would for instance let the user start judging reconciled cells before the reconciliation process is complete. Or let the user extract values out of a JSON response from an API before the URL fetching is complete.
* The user could also run follow-up long-running operations, such as reconciliation followed by data extension, and the data extension operation could already start running on the existing reconciled cells. This would be a way to have #2256.

### Additional context

This requires the ability to handle (and therefore likely be able to serialize) partial results for long-running operations, which has been requested independently: #87, #580, #1231

wetneb:
One last thing to solve before closing this one would be to have a satisfactory solution for refreshing the grid in the front-end. Currently this only happens when the user changes page or does another action, which is not very intuitive.

Issue #1440, opened 2018-01-23
Do not display the entire element returned by "fetch urls"


ettorerizza:
If you click on [this Viaf API URL](http://viaf.org/viaf/9847974/viaf.json), even your browser will have trouble displaying the returned JSON. Imagine then when you have 100 of them and you use "add column by fetching URLS" in Open Refine... Even with only two URLs, the whole interface slows considerably.

Same problem when extracting source code from a large web page for scraping purpose. I wonder if there would be any way to display in Refine only the first lines of the Json/xml or source code. After all, I doubt that users really read this tag soup. Instead, they certainly use the web developper in their browser to identify the path of the elements they later extract with the parseJson() or parseHtml() functions.

wetneb:
I think a cap on text length would be reasonable.

Otherwise, one possible workaround is to do the fetching in Python and extract the interesting data in the same go (so, without storing the full response).

ettorerizza:
@wetneb I had thought of this possibility. Because in the end, we often know in advance what we want to extract from the JSON / XML / HTML. The only case where storing the answer is useful it's when one wants to create several columns from the same xml. The ideal would be to have the possibility to store the answer, or parse it on the fly with something like:

`value.fetchUrl().parseJson()...`

thadguidry:
This sounds like a preference that we should set...and then if users want to override it...they update the preference.

@wetneb   So... any thoughts on what a reasonable cap that we can setup in preferences.vt ???

wetneb:
I'm not sure, maybe something like 1024 characters?

jackyq2015:
Maybe the extraction can be done from the VIAF side. I will ease the burner from the OR side. Though I am not familiar with the syntax.

- VIAF in the OCLC API Explorer:
https://platform.worldcat.org/api-explorer/VIAF

- The SRU syntax used in the API SRUSearch function:
http://www.loc.gov/standards/sru/

wetneb:
Yes if there is a text cap it should be large by default and configurable by the user.

ostephens:
On the VIAF side if you use the SRU interface you can limit the number of records returned, but I don't think this is the problem here. The issue is that requesting a single record that has many alternative name representations etc. you get a very large chunk of JSON which you can't limit through the API.

I'm in favour of a user configurable text cap on the amount of text show in a cell with a couple of caveats:

* There needs to be a 'show everything' setting (rather than just putting large numbers to the limit and hoping)
[...]

PR #5840, opened 2023-05-11
Support for resizing columns

### LINKED ISSUE #4806: Possibility to adjust column width manually
Right now column width is generated arbitrarily based on content of the column headers and/or data values - this can be very problematic particularly with XML files where the column headers and sometimes the data values  can be quite lengthy if the XML is not particularly well structured, then very wide columns fill up a user's screen - preventing efficient work.

### Proposed solution
Ideally, column width should be controlled by the users with a familiar user interaction pattern, such as a dragging motion by the user's cursor, as is the case in other tabular management tools such as Excel / Google sheets, etc.

### Alternatives considered
None yet.

### Additional context
N/A

wetneb:
Fixes #4806. This is an attempt to add support for resizing columns.

resizing

It is not ready yet because the initial sizing of the columns has changed and is likely to be less appropriate in some cases. But I would already welcome feedback about the approach!
The initial sizing has changed because I needed to change the way textual content is wrapped inside cells (and in column headers), to allow for columns to resize correctly. This has an effect on the initial sizing as well.

Some details about the approach:
* to change the size of a column, one can either apply inline CSS to the `<th>` and all the `<td>`s in that column, or use the dedicated `<col>` element. The latter felt cleaner (only applying style in one place) but maybe the other approach has other advantages?
* to handle the resizing itself, one could try using the `resize` CSS property, but translating the effects of that to the entire column seems a little hacky. Instead I went for manual handling in Javascript using event listeners for mouse clicks and moves. Not sure it is the best approach though!
* The column sizing should be preserved when paginating or updating the grid (which might mean columns getting added / removed / renamed) so I implemented a rather simple solution by remembering which column name was associated with which width. It will fail to remember the column width when renaming a column, but maybe that's not so critical. I wonder if this information should be persisted (perhaps in the project-level preference store).

Why do this in 4.0 and not in master:
* while the change is conceptually simple, there is a lot of potential to break things in some use cases - so I'd rather not put this in the stable release series
* I am hoping that this change (combined with capping the row height) will help make table layouts more predictable and therefore make real-time updates of the grid more acceptable (for #5541). Because at the moment I am refreshing periodically the grid when some cell values have not yet been computed by a long-running operation, and doing so can result in big layout changes which are disruptive for the user as they are done without any action from them.

How to improve the initial column widths:
* one lazy approach would be to say that we do not actually care that much about the initial column widths as long as they can be changed easily by the user - maybe invest time into persisting those column widths instead. Arguably the sizing strategy will also be impacted by #1440, which I would also tackle as a follow-up - so maybe it's worth waiting for that to tackle this initial sizing.
* another lazy approach would be to just set equal column widths initially. That's apparently what Google Sheets does. Not amazing but perhaps better than wonky and unpredictable heuristics?
* we could gather some statistics about the maximum and/or average length of the content stored in each column (either backend or frontend side) and come up with some heuristics to come up with sensible initial column lengths.
* one could try to revert my word wrapping changes in CSS, so that the initial sizing of the grid is left unchanged, and only apply those CSS rules once the user starts resizing a column. I think this is likely to have some bad side effects such as content jumping around rather unpredictably at the beginning of the resizing, making it less smooth.
* other approaches?

Another question is how to show column names when they do not fit in the header. For now they are just truncated and one can see the full column name by hovering.
image

Another question is whether the fact that columns are resizable should be made more visible with some sort of visual hint that the boundary between the headers can be dragged.

Also, the boundary is currently only draggable if you click slightly to the left of it, and I find that not so convenient to use: clicking slightly to the right should also work.

cc @lozanaross

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
@@ -94,6 +94,8 @@ DataTableColumnHeaderUI.prototype._startResizing = function(clickEvent) {
   state.columnName = self._column.name;
   state.originalWidth = self._col.width();
   state.originalPosition = clickEvent.pageX;
+  // for conversion from px to em
+  state.emFactor = parseFloat(getComputedStyle($(".data-table-container colgroup")[0]).fontSize);
 
   $('body')
       .on('mousemove', DataTableColumnHeaderUI.mouseMoveListener)
@@ -122,7 +124,9 @@ DataTableColumnHeaderUI.mouseReleaseListener = function(e) {
   }
   var state = DataTableColumnHeaderUI.resizingState;
   if (state.dragging) {
-    DataTableView.columnWidthCache.set(state.columnName, state.col.width());
+    var totalMovement = e.pageX - state.originalPosition;
+    var newWidth = state.originalWidth + totalMovement;
+    state.col.width((newWidth / state.emFactor) + 'em');
     state.dragging = false;
     $('body')
         .off('mousemove', DataTableColumnHeaderUI.mouseMoveListener)
--- a/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
@@ -76,6 +76,7 @@ $(function() {
       var totalMovement = e.pageX - state.originalPosition;
       var newWidth = state.originalWidth + totalMovement;
       state.col.width(newWidth);
+
       e.preventDefault();
     }
   }).on('mouseup', function(e) {
@@ -104,21 +105,18 @@ DataTableColumnHeaderUI.prototype._render = function() {
     self._createMenuForColumnHeader(this);
   });
 
-  elmts.resizer.on('mousedown', function(e) {
-    // only capture left clicks
-    if (e.button !== 0) {
-      return;
-    }
-    e.preventDefault();
-    var state = DataTableColumnHeaderUI.resizingState;
-    state.dragging = true;
-    state.col = self._col;
-    state.columnName = self._column.name;
-    state.originalWidth = self._col.width();
-    state.originalPosition = e.pageX;
-  });
-
   self.updateColumnStats();
+};
+
+DataTableColumnHeaderUI.prototype._startResizing = function(clickEvent) {
+  var self = this;
+  clickEvent.preventDefault();
+  var state = DataTableColumnHeaderUI.resizingState;
+  state.dragging = true;
+  state.col = self._col;
+  state.columnName = self._column.name;
+  state.originalWidth = self._col.width();
+  state.originalPosition = clickEvent.pageX;
 };
 
 DataTableColumnHeaderUI.prototype.updateColumnStats = function() {
--- a/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
@@ -31,11 +31,12 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
  */
 
-function DataTableColumnHeaderUI(dataTableView, column, columnIndex, td) {
+function DataTableColumnHeaderUI(dataTableView, column, columnIndex, td, col) {
   this._dataTableView = dataTableView;
   this._column = column;
   this._columnIndex = columnIndex;
   this._td = td;
+  this._col = col;
 
   this._render();
 }
@@ -58,6 +59,39 @@ DataTableColumnHeaderUI.prototype.getColumn = function() {
   return this._column;
 };
 
+// global state for the resizing of column headers
+DataTableColumnHeaderUI.resizingState = {
+  dragging: false, // whether we are currently resizing any column
+  col: null, // the column being resized
+  columnName: null, // the name of the column being resized
+  originalWidth: 0, // the original width of the header when the dragging started
+  originalPosition: 0 // the original position of the cursor when the dragging started
+};
+
+// global event handlers to react to mouse moves during resizing
+$(function() {
+  $('body').on('mousemove', function(e) {
+    var state = DataTableColumnHeaderUI.resizingState;
+    if (state.dragging) {
+      var totalMovement = e.pageX - state.originalPosition;
+      var newWidth = state.originalWidth + totalMovement;
+      state.col.width(newWidth);
+      e.preventDefault();
+    }
+  }).on('mouseup', function(e) {
+    // only capture left clicks
+    if (e.button !== 0) {
+      return;
+    }
+    var state = DataTableColumnHeaderUI.resizingState;
+    if (state.dragging) {
+      DataTableView.columnWidthCache.set(state.columnName, state.col.width());
+      state.dragging = false;
+    }
+    e.preventDefault();
+  });
+});
+
 DataTableColumnHeaderUI.prototype._render = function() {
   var self = this;
   var td = $(this._td);
@@ -68,6 +102,20 @@ DataTableColumnHeaderUI.prototype._render = function() {
   elmts.nameContainer.text(this._column.name);
   elmts.dropdownMenu.on('click',function() {
     self._createMenuForColumnHeader(this);
+  });
+
+  elmts.resizer.on('mousedown', function(e) {
+    // only capture left clicks
+    if (e.button !== 0) {
+      return;
+    }
+    e.preventDefault();
+    var state = DataTableColumnHeaderUI.resizingState;
+    state.dragging = true;
+    state.col = self._col;
+    state.columnName = self._column.name;
+    state.originalWidth = self._col.width();
+    state.originalPosition = e.pageX;
   });
 
   self.updateColumnStats();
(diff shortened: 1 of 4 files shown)


main/webapp/modules/core/styles/views/data-table-view.css
```diff
@@ -146,14 +163,13 @@ table.data-table td.column-header, table.data-table th.column-header {
...
   padding: 4px 0 0 0;
-  display: block;
+  display: inline-block;
+  position: absolute;
 }
 
 a.column-header-menu {
-  float: left;
-  display: block;
+  display: inline-block;
```
wetneb, line 172: I cleaned up this part to help with the resizing. I am not sure why the column menu was a `float: left` which was intended to lodge itself in the margin of the column header. It feels cleaner with two inline elements next to each other, but there might be ways to improve this even more.

main/webapp/modules/core/styles/views/data-table-view.css
```diff
@@ -191,7 +207,7 @@ div.data-table-cell-content {
   line-height: 1.2;
   color: #222;
   position: relative;
-  white-space: pre-wrap;
+  overflow-wrap: anywhere;
```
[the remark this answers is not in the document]
wetneb: I found a way to cap the initial column width so this is less of a problem. Still, there might be ways to improve the initial layout even more.

main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
```diff
@@ -58,6 +59,39 @@ DataTableColumnHeaderUI.prototype.getColumn = function() {
...
+  dragging: false, // whether we are currently resizing any column
+  col: null, // the column being resized
+  columnName: null, // the name of the column being resized
+  originalWidth: 0, // the original width of the header when the dragging started
+  originalPosition: 0 // the original position of the cursor when the dragging started
+};
+
+// global event handlers to react to mouse moves during resizing
+$(function() {
+  $('body').on('mousemove', function(e) {
```
Abbe98, line 73: Instead of always listening for mouse movements you could listen for mouse down, register a listener for mouse movement, and finally unregister it on mouse up. It should greatly bring down the number of times this runs and because it would only be active when the mouse is down there will be no performance penalty of having it on each element rather than globally.

Abbe98 (commented):
I like this approach! Having the logic in JavaScript like this makes sense here as it opens up some capabilities for doing "smart" defaults in the future and maybe even something that could be exposed in an extension API just like cell rendering.

> The column sizing should be preserved when paginating or updating the grid (which might mean columns getting added / removed / renamed) so I implemented a rather simple solution by remembering which column name was associated with which width. It will fail to remember the column width when renaming a column, but maybe that's not so critical. I wonder if this information should be persisted (perhaps in the project-level preference store).

I think this could be serialized to JSON and stored in Localstorage, I'm not as sure it would be wise to have it in the project preference store given that it could make it harder to share projects with others who might want different preferences.

> we could gather some statistics about the maximum and/or average length of the content stored in each column (either backend or frontend side) and come up with some heuristics to come up with sensible initial column lengths.

This is my favorite option, for large datasets we could use a subset if need be.


These are just my initial thoughts, I'm not sure I will have the time to have a proper look next week but I hope so.

wetneb:
Thanks for the review @Abbe98!

> I think this could be serialized to JSON and stored in Localstorage, I'm not as sure it would be wise to have it in the project preference store given that it could make it harder to share projects with others who might want different preferences.

What sort of scenario about thinking about? If I open a project shared by someone else and it opens with the column widths that they have set, isn't it a good thing? I would be able to resize the columns afterwards too. The alternative is that the project just opens with the default sizing, because if I have never opened this project before, I will not have any preferred widths for those columns, unless you are thinking about something different?

Storing things in LocalStorage could work too (and it's arguably quicker since it does not require an HTTP request to sync up with the backend), but it would also have downsides: opening the same project with a different browser would not preserve the widths, and we would need to ensure that this data gets cleaned up when projects are deleted. Do we actually already use LocalStorage for anything? Overall I would be a little worried that we are spreading our state to too many storage locations and we might lose track of what is stored where (which can lead to confusing UX down the line).

> 
> > we could gather some statistics about the maximum and/or average length of the content stored in each column (either backend or frontend side) and come up with some heuristics to come up with sensible initial column lengths.
> 
> This is my favorite option, for large datasets we could use a subset if need be.

I am also tempted by it but there is a little voice in my head telling me that this is a bad decision, because the browser should best placed to figure out sensible sizing for DOM elements based on their contents. There are currently many places in the UI where we are statically sizing things in Javascript (typically the panels) with ugly event listeners on viewport resizing to adapt accordingly. I am of the impression that there is a consensus that we should migrate out of this to a responsive design which uses CSS as much as possible.

Granted, the backend is able to see a lot more rows that the frontend so it would have more accurate statistics about how much text you have in each column, but the backend is oblivious to the fact that characters do not have the same width, the font size currently used by the user, or predicting the size taken up by a reconciled cell…

Also, it's not clear to me that there is a huge benefit in being able to foresee the presence of a big cell in a column which has only small cells at the beginning… How often does that happen is it actually a good thing to inflate the space used to render those initial cells just so that the user does not have to resize that column later on? It's really unclear to me.

Abbe98:
> What sort of scenario about thinking about? If I open a project shared by someone else and it opens with the column widths that they have set, isn't it a good thing?

I don't think so, the users might work on different parts of the data, they might have different screen sizes, and their general preference might differ. For a one time "share" it's probably fine but if you work collaborative and share the project often or even work in the same instance this would get annoying quickly.

> Overall I would be a little worried that we are spreading our state to too many storage locations and we might lose track of what is stored where (which can lead to confusing UX down the line).

So I'm I and I think this is the main case against, however, given hosted instances I can see additional use cases for introducing LocalStorage as an option for UI state and customization so I think bringing in this as an option might be a good thing for other parts of OpenRefine as well.

>I am also tempted by it but there is a little voice in my head telling me that this is a bad decision, because the browser should best placed to figure out sensible sizing for DOM elements based on their contents. There are currently many places in the UI where we are statically sizing things in Javascript (typically the panels) with ugly event listeners on viewport resizing to adapt accordingly. I am of the impression that there is a consensus that we should migrate out of this to a responsive design which uses CSS as much as possible.
>
>Granted, the backend is able to see a lot more rows that the frontend so it would have more accurate statistics about how much text you have in each column, but the backend is oblivious to the fact that characters do not have the same width, the font size currently used by the user, or predicting the size taken up by a reconciled cell…

Very much an unfinished idea: we could use the browser to compute defaults given a heterogeneous subset of the dataset contents which we the use as a set of fixed widths.

> Is it actually a good thing to inflate the space used to render those initial cells just so that the user does not have to resize that column later on?

I don't think it so much about the user having to resize columns as it would be about things jumping around in the UI, this is an issue when switching pages already today and can be rather annoying when working with heterogeneous data and this might be an opportunity to fix it.

Lastly I don't think either of these issues are blockers for this particular change!

wetneb:
Agreed, let's keep the persistence of column widths outside the scope of this.

> Very much an unfinished idea: we could use the browser to compute defaults given a heterogeneous subset of the dataset contents which we then use as a set of fixed widths.

Yes, I think that would make a lot of sense. I would actually do it on the rows initially shown. Maybe they are not representative of the rest but as you write the point is not to find the optimal widths but rather avoid jumps due to width changes.

> given hosted instances I can see additional use cases for introducing LocalStorage as an option for UI state and customization so I think bringing in this as an option might be a good thing for other parts of OpenRefine as well.

I would be keen to use LocalStorage when (if?) we add support for hosted instances or multi-user workflows but I think it will be a lot simpler to make this move if we have a clear view of the existing state, stored in not so many different places, so intuitively I'd rather not introduce this gradually.

Abbe98:
> I would be keen to use LocalStorage when (if?) we add support for hosted instances or multi-user workflows but I think it will be a lot simpler to make this move if we have a clear view of the existing state, stored in not so many different places, so intuitively I'd rather not introduce this gradually.

I agree but if we do commit the state to the project let's make it optional (both loading and saving) not to break existing hosted instances or multi-user workflows.

wetneb:
> I agree but if we do commit the state to the project let's make it optional (both loading and saving) not to break existing hosted instances or multi-user workflows.

Here it's not a problem, but in principle, as long as we do not consider those use cases as supported I would be reluctant to factor in those use cases for decisions like this - because that's pretty much what it means to "support" those use cases!

Of course I am not opposed to OpenRefine supporting those use cases in the future but at the moment I would find it irresponsible to declare that they are, primarily given the security implications. As a maintainer, I do not want to claim that OpenRefine meets the expected standards to be run as a hosted, multi-user app, because it would be (I think) deceitful to users. Those use cases might have big security flaws which cannot be fixed easily and some future changes might introduce more problems than there currently are.

Abbe98:
> Here it's not a problem, but in principle, as long as we do not consider those use cases as supported I would be reluctant to factor in those use cases for decisions like this - because that's pretty much what it means to "support" those use cases!

Indeed here it still useful to make it optional I would say. On the broader topic I think there still is a middle-ground where we avoid implementations that would need downstream patches for hosted usecases while not taking them into account for feature and design aspects.

main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
wetneb: Done.

PR #5842, opened 2023-05-12
fix star/flag action #5738

### LINKED ISSUE #5738: (Un)star/flag action reverts unrelated cell reconciliation judgments
Starring or flagging an individual row reverts any manual reconciliation action on a cell in that row.

### To Reproduce
1. First, reconcile a column
2. Then, make a reconciliation judgment in one of the cells of that column. The action should only impact that single cell (not similar cells).
3. Finally, change the star or flag status of the same row

### Current Results

The reconciliation action of step 2 is visibly reverted (but only client-side, not server side: if the grid is refreshed, we see that it was not actually reverted).

### Expected Behavior

The action in step 2 should not be visibly reverted.

### Screenshots
Peek 27-03-2023 21-32

### Versions
 - Operating System: Linux
 - Browser Version: Firefox
 - JRE or JDK Version: 17
 - OpenRefine: 3.8-SNAPSHOT

### Additional context

Reported on the forum: https://forum.openrefine.org/t/unstarring-reverts-un-reconciliation/672

SoniaSun810:
Fixes #5738

- We currently execute `renderRow(tr, r, row, even)` every time a row is starred/flagged, which is not necessary for the frontend;
-  Adding or deleting `data-table-flag-on` / `data-table-flag-off` in the classList of `flag`, `data-table-star-on` / `data-table-star-off` in the classList of `star` is sufficient;
- By eliminating this unnecessary rendering, we can reduce the load and improve the overall performance.


wetneb:
Oops, that's an excellent point indeed - thanks for noticing that!

Abbe98 (approved):

PR #5843, opened 2023-05-12
Sonia

SoniaSun810:
Fixes #5842

Thank you a lot for this valuable feedback! @wetneb @Abbe98

-To ensure client-server synchronization, I've placed the toggling function within the `onDone`.
-Considering that toggling the star and flag only affects these specific elements and the rest of the cells in the row remain unaffected, I believe the proposed solution is secure.
-I really appreciate your suggestion about branch usage. I've now set up a new branch named `sonia` and I'll be sure to use it for future pull requests.


wetneb:
I already merged #5842, so this PR is not necessary.
You can make changes to an existing pull request by pushing more commits to the branch it is opened from. There is no need to open a new pull request as long as the original one is still open.
Also, we recommend to use a branch only for one pull request. See the expected workflow here:
https://openrefine.org/docs/technical-reference/contributing#submitting-your-changes

PR #5848, opened 2023-05-15
Change tab

### LINKED ISSUE #5618: Left panel tabs wrap when the font size is large enough
The problem is that the left hand panel is set to 300px width, so if the browser font in browser settings is increased it can cause the tabs to wrap.  

Note: just magnifying your browser window with 'CTRL+' does not cause a problem as everything scales up.

Suggested fixes:
* Removing the x/x count from the tab as it is not that useful, and it can grow from x/x to xx/xx to xxx/xxx. This would help the most in preventing the tabs from wrapping.
* Shrink tab margins from 0.8em to 0.5em
* Change layout of the hide-left-panel-button so it overlays the tabs so it doesn't cause Undo/Redo to wrap.
image

SoniaSun810:
Fixes #5618

- To prevent the left panel tabs wrap when the font size is large, we deleted unnecessary information `x/x`
- As my teammate @skhoylow8 discussed in the [issue](https://github.com/OpenRefine/OpenRefine/issues/5618)


skhoylow8:
Hi. My teammate and I noticed that our PR fails when testing for the component that we deleted. Since we already discussed deleting this extra information in the linked issue, would we be able to delete the test case? Thanks in advance for your help.

PR #5851, opened 2023-05-16
add an outline to tab buttons when focused

Abbe98:
jQueryUI uses a combination of `li`/`a` to build it's tabLists rather than `button` as a result(one of many) it does not get an outline when focused.

This adds such an outline manually using CSS. You can test this by just tabbing through the interface, when the tab navigation buttons now are focused they have an outline just like other elements.


wetneb (approved):
I assume that the color is coming from jQuery UI and so it is not one of the colors from our palette that we define as CSS variables. Otherwise I'd encourage to use those.

Abbe98:
> I assume that the color is coming from jQuery UI and so it is not one of the colors from our palette that we define as CSS variables. Otherwise I'd encourage to use those.

It's the color used by Chrome and Firefox for outlines. It could do with a comment.

PR #5858, opened 2023-05-18
Default "Use quote as separator" to False for TSV import

### LINKED ISSUE #3853: Default "Use quote as separator" to False for TSV imports
I got a TSV file and uploaded it to OpenRefine, but OpenRefine parsed it incorrectly, shifting values over several columns.

### To Reproduce
1. Upload the file to OpenRefine.

2. Facet the categories column and select one of the values with "mm" in it. Look at the information in the size columns of the selected row. 

### Current Results
You should see that the actual size values have been shifted two or three columns to the left so that the height value is now in the categories value.

### Expected Behavior
In the raw XSLX file, the values are in the columns they are supposed to be in, I believe that OpenRefine is parsing the information incorrectly.

### Screenshots
OpenRefineScreenshot
ExcelScreenShot

### Versions
 - Browser Version: Microsoft Edge 90.0.818.46
 - JRE or JDK Version:
 - OpenRefine: 3.4.1

### Datasets
Since Github won't let me upload the original TSV, here is an XSLX I created by exporting the TSV's information from OpenRefine.
[phm-collection.xlsx](https://github.com/OpenRefine/OpenRefine/files/6385815/phm-collection.xlsx)

jnchen1:
Fixes #3853 

- The `columnSeparatorValue` is currently used as a variable to store the separator type. If the value is set to 'tab' to indicate a `TSV` file, the checkbox will be left unchecked. Otherwise, the checkbox will remain checked.
- Two test cases have been added to ensure the correct functionality of this change.

### DIFF
--- a/main/src/com/google/refine/importers/SeparatorBasedImporter.java
+++ b/main/src/com/google/refine/importers/SeparatorBasedImporter.java
@@ -73,10 +73,11 @@ public class SeparatorBasedImporter extends TabularImportingParserBase {
         ObjectNode options = super.createParserUIInitializationData(job, fileRecords, format);
 
         String separator = guessSeparator(job, fileRecords);
-        JSONUtilities.safePut(options, "separator", separator != null ? separator : "\\t");
+        String nonNullSeparator = separator != null ? separator : "\\t";
+        JSONUtilities.safePut(options, "separator", nonNullSeparator);
 
         JSONUtilities.safePut(options, "guessCellValueTypes", false);
-        JSONUtilities.safePut(options, "processQuotes", true);
+        JSONUtilities.safePut(options, "processQuotes", !nonNullSeparator.equals("\\t"));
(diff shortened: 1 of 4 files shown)


wetneb (commented):
Thank you very much for this contribution which should make a significant difference for our users!

I think it would be cleaner to keep the frontend code as it is, just translating the default options provided by the backend into the UI. I think this logic should rather be in the backend itself (in `SeparatorBasedImporter.createParserUIInitializationData`).

It's really nice that you wrote tests for this. But because of the above, I would also recommend that the corresponding tests be Java unit tests, which are much faster to execute. You could add assertions in existing Cypress tests which load TSV files to check the status of the checkbox (that should not add much to the running time of those tests), but new Cypress test cases are quite costly so I would rather avoid them in this context.

jnchen1:
Thank you for taking the time to review it! I will make the necessary changes as soon as possible.

wetneb (approved):
Thanks a lot!

PR #5862, opened 2023-05-22
Fixed the issue with transpose so it treats blank cells as null

### LINKED ISSUE #5229: Transpose cells across columns into Rows        BUGS with blank cells?
Hi. Today I used a menu called 'Transpose cells across columns into Rows'. I checked on 'Ignore blank cells' but it failed. I think that the mistake is on the label. It works with null cells but not with empty cells. So It can't work with blank cells whis is empty OR null. Can you verify it? Thanks a lot, Fabio
Untitled

skhoylow8:
Fixes #5229 

- Fixed the issue so that is will treat blank cells with an empty string the same way null is treated when transposing columns
- Added a test file to make sure it was producing the expected output


wetneb (changes requested):
Thank you very much for this contribution!

I have not tested it yet but it looks like this is the right approach. One important thing though: we normally don't just consider `null` and `""` as blank values, but also any string that consists of whitespace only. You could use `ExpressionUtils.isBlank` to do this check.

Also, for the record, comparisons of strings in Java should be made via the `.equals()` method, because there is generally no guarantee that two equal strings point to the same String object in all circumstances. So instead of `cell.value == ""`, you should use `"".equals(cell.value)` (the empty string comes first because we know for sure it is not null, so we do not need to check if `cell.value` is not null).

To fix the formatting issues detected by the PR checks, you can run `mvn formatter:format` and commit and push the changes.

skhoylow8:
Hi @wetneb thank you for the feedback. Just for clarification, if I were to do the `ExpressionUtils.isBlank` then I wouldn't need the `cell.value == null` or the `"".equals(cell.value)` correct?

wetneb:
Exactly!

fabiolinus:
Hi to all and thanks for working on this ticket. What is the difference between cell.value == '''' and "".equals(cell.value)? Thanks, Fabio

wetneb:
@fabiolinus see https://www.baeldung.com/java-compare-strings

PR #5867, opened 2023-05-23
do not remove element outlines set by the browser

Abbe98:
Most notable change in practice is within the open project table.


wetneb (approved):
Not tested but hard to imagine how that could go wrong.

PR #5869, opened 2023-05-23
fix pointer cursor in column headers

Abbe98:
This ensures that the pointer cursor is only in use for the menu buttons and not the whole column headers.


wetneb (approved):
Much better! Interesting that we did not notice that before…

wetneb:
I just realized that for collapsed columns, it does make sense to have the whole column area with this pointer, since clicking anywhere on the column header uncollapses the column. Worth opening a follow-up issue about this?

PR #5871, opened 2023-05-23
project-permalink: keyboard support

Abbe98:
Instead of watching for 'mouseover' this checks for 'focus' which ensures both clicking and tabbing generates a new permalink.


wetneb (approved):
Works with Firefox and Chromium.

PR #5886, opened 2023-06-01
Make it possible to set custom headers using only the keyboard when fetching urls

Abbe98:
This swaps a custom `<details>` implementation for a HTML based one making it keyboard accessiable.


Abbe98:
[Screencast from 2023-06-01 08-55-00.webm](https://github.com/OpenRefine/OpenRefine/assets/2631719/443699ed-8495-4446-9db3-ec08d5af8f5c)

wetneb (approved):
Nice, I did not know about those HTML tags. Very cool! And also happy to get rid of the animation there, I think it was not very fitting.

PR #5892, opened 2023-06-03
Alternative implementation for importing using url

### LINKED ISSUE #5431: Import project archive from URL

SoniaSun810:
Fixes #5431 

- This enhancement provides users the flexibility to switch their preference between importing from a local file and a URL;
- We altered the logic in both the frontend `import-project-ui.js` and the backend `ImportProjectCommand.java` to differentiate between the two project import options. 
-  In alignment with the existing UI design, we integrated a new feature that permits users to clear the local file input.


SoniaSun810:
Screenshot 2023-06-02 at 20 52 27

Screenshot 2023-06-02 at 20 51 07

Hey @wetneb ! Should the new UI meet your approval, I will proceed to adjust the `preview_project.cy.js` test case accordingly.

wetneb (commented):
Thank you so much @SoniaSun810 and @jnchen1 for this, and sorry about the delay in reviewing this. It looks very good to me.

I wonder if you considered writing tests for this?
You could go for one or more Cypress tests, or test this in the backend only, with a Java test. We currently have a fairly empty `ImportProjectCommandTests` test class which you could work from.

For Java tests, you could use `MockWebServer` to spin up a local web server which would serve the OpenRefine project file (stored somewhere as a test fixture). It's probably easiest if you look for usages of `MockWebServer` in the rest of the code base as examples (although I don't think we use it to serve a test file yet - but that's likely doable).

Cypress tests could be doable, but it's probably harder to avoid relying on an external web service (and Cypress tests are generally a lot slower, so I would use them sparingly).

I wonder if @OpenRefine/designers have suggestions of tweaks for the UI?

SoniaSun810:
Hi, @wetneb ! Thank you for your insightful suggestions. I appreciate your patience as I'm currently tied up with some other tasks. I understand the importance of the test cases and will work on incorporating Java tests (and Cypress tests) as soon as possible. Your guidance has been extremely valuable!

wetneb:
Ok! I think it would be totally okay to have either only java tests or only Cypress tests for this.

main/tests/cypress/cypress/e2e/import-project/import_project.cy.js
```diff
@@ -47,4 +47,31 @@ describe(__filename, function () {
     cy.waitForProjectTable();
     cy.get('#project-name-button').contains(projectName);
   });
+
+  it('Import a project from URL', function () {
+    cy.visitOpenRefine();
+    cy.navigateTo('Import project');
+    const url = "https://pintoch.ulminfo.fr/6e551ee9a1/Changing-the-type-of-the-column-for-numeric-1680159450732.openrefine.tar.gz";
```
wetneb, line 54: This is a URL that points to my personal server, so I'd rather avoid relying on it in the test itself.
In other places in the Cypress tests we simply point to the download URL offered by GitHub. We happen to already have another fixture of OpenRefine project archive that we could likely use directly:
```suggestion
    const url = "https://raw.githubusercontent.com/OpenRefine/OpenRefine/master/main/tests/cypress/cypress/fixtures/food-small-csv.openrefine.tar.zip";
```
This will likely require adapting the assertions afterwards since the project is different.

wetneb (commented):
Amazing, thanks a lot for the great work! Just a tiny detail and I would say we are good to go.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -90,6 +91,10 @@
     "core-index-import/this-computer": "This Computer",
     "core-index-import/warning-data-file": "You must specify a data file to import.",
     "core-index-import/warning-import-file": "You must specify a project file.",
+    "core-index-import/warning-import-url": "You must specify a correct url.",
+    "core-index-import/warning-import-input": "You must specify either a project file / url.",
+    "core-index-import/warning-import-two-input": "You must specify only one input field.",
+    "core-index-import/delete-import-file": "Delete file",
```
thadguidry, line 97: I think less confusing would be to say **Clear file**
```suggestion
    "core-index-import/delete-import-file": "Clear file",
```

thadguidry (changes requested):

wetneb (approved):

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -58,8 +58,9 @@
     "core-index-create/formats": "TSV, CSV, *SV, Excel (.xls and .xlsx), JSON, XML, RDF as XML, and Google Data documents are all supported. Support for other formats can be added with OpenRefine extensions.",
     "core-index-create/from": "Get data from",
     "core-index-import/import-proj": "Import project",
-    "core-index-import/locate": "Locate an existing Refine project file (.tar or .tar.gz):",
+    "core-index-import/locate": "Locate an existing Refine project file or use a URL (.tar or .tar.gz):",
```
tfmorris, line 61: Should this have a new id to reflect the fact that the semantics have changed? How do we signal the translators that the existing translation is no longer valid and needs retranslating?

wetneb: That's handled by Weblate automatically. The existing translations are marked as stale and translators are incentivized to update them. In the meantime, the old translation is still shown in the UI, which is fine I think

tfmorris: OK, thanks. I guess I was putting too strong a semantic value on the identifier. Sorry for the noise!

PR #5895, opened 2023-06-04
Skip Wikibase edits on deleted items

### LINKED ISSUE #5385: OpenRefine upload to Wikidata stalls on deleted items
When you have a list of items and reconcile them "as identifiers", it often happens that some of the items had already been deleted from Wikidata. But they still reconcile. 

However, when the Wikidata upload gets to this item, it gives the following error:
Exception in thread "Thread-12" java.lang.NullPointerException
And then it basically just stops the upload...

Steps to reproduce:
Reconcile Q56640589 as Wikidata identifier
try to import something

Expected behavior:
Skip the deleted item. Or do not reconcile it at all.

wetneb:
Fixes #5385.

In general I try to stay away from working on Wikibase integration in OpenRefine at the moment, but this bug feels pretty important and easy to solve.


wetneb:
Merging to gather feedback from the reporters via snapshot releases.

PR #5898, opened 2023-06-06
Added auto cluster checkbox to cluster dialog

### LINKED ISSUE #241: Provide ability to bypass Levenshtein clustering in order to run PPM clustering
_Original author: ross_gr...@mentor.com (November 19, 2010 19:33:54)_

Current UI seems to issue the cluster command when the Method value is changed.  When I change the method from key collision to nearest neighbor, the application immediately runs a nearest neighbor clustering function with the default values for Distance function, Radius, and Block Chars.  I would like to manually issue a command to generate clusters after changing the values for Distance function, Radius, and Block Chars so I don't have to wait for the cluster command to execute with the default values (Levenshtein,1.0,6).

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=241_

elebitzero:
Fixes #241

- Implemented auto cluster checkbox for the cluster dialog.
- Added cluster button to the cluster dialog.

After:
Auto cluster checkbox after 3

Changes to add in a separate PR:
- Add a preference to uncheck auto cluster by default.

### DIFF
--- a/main/webapp/modules/core/about.js
+++ b/main/webapp/modules/core/about.js
@@ -4,4 +4,4 @@ $("#contributors").text($.i18n('core-about/contributors'));
 $("#definition").text($.i18n('core-about/definition'));
 $("#history-openrefine").html($.i18n('core-about/history' ,'http://www.metaweb.com/' ,'Metaweb Technologies,Inc.', 'http://www.google.com/','Google' ));
 $("#thanks").text($.i18n('core-about/thanks'));
-$('#app-home-button').attr('title', $.i18n('core-index/navigate-home'));//temp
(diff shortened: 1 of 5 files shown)


elebitzero:
@wetneb

The coveralls check is flaky, can we modify the workflow so it only runs when Java code is changed?

wetneb:
I opened an issue about that: #5909

wetneb:
I like it! I am not sure about just "Auto" as label for the checkbox, it might be a bit unclear for folks. How about "Auto-update", still quite short?

wetneb:
maybe @OpenRefine/designers have thoughts about this UI?

elebitzero:
Yeah that sounds better, I changed it to Auto-cluster and updated the video.

PR #5900, opened 2023-06-07
Make the "Get data from" menu keyboard accessible

Abbe98:


elebitzero (approved):
I tried this change out and confirmed it works.

PR #5902, opened 2023-06-07
Swap gif loading spinners to CSS based ones

Abbe98:
Part of #3017.

This is a draft to get some suggestions on the looks and sizing of the loading spinners. The motivation is that we need a loading  spinner/indicator that can have a themable  foreground.

### DIFF
--- a/main/webapp/modules/core/preferences.vt
+++ b/main/webapp/modules/core/preferences.vt
@@ -50,7 +50,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
   </div>
 
   <div id="body-info">
-    <div id="loading-message"><img src="images/large-spinner.gif" /> <span id="or-proj-starting"></span></div>
+    <div id="loading-message"><span class="loader"></span> <span id="or-proj-starting"></span></div>
   </div>
 </body>
 </html>
--- a/main/webapp/modules/core/project.vt
+++ b/main/webapp/modules/core/project.vt
@@ -60,7 +60,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
       </div>
     </div>
     
-    <div id="loading-message"><img src="images/large-spinner.gif" /> <span id="or-proj-starting"></span></div>
+    <div id="loading-message"><span class="loader"></span <span id="or-proj-starting"></span></div>
       
     <div id="body">
       <div bind="leftPanelDiv" id="left-panel">
--- a/main/webapp/modules/core/scripts/dialogs/add-column-by-reconciliation.js
+++ b/main/webapp/modules/core/scripts/dialogs/add-column-by-reconciliation.js
@@ -193,7 +193,7 @@ ExtendReconciledDataPreviewDialog.prototype._show = function(properties) {
 
 ExtendReconciledDataPreviewDialog.prototype._update = function() {
   this._elmts.previewContainer.empty().html(
-        '<div bind="progressPanel" class="add-column-by-reconciliation-progress"><img src="images/large-spinner.gif" /></div>');
+        '<div bind="progressPanel" class="add-column-by-reconciliation-progress"><span class="loader"></span></span></div>');
 
   var self = this;
   var params = {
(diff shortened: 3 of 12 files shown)


Abbe98:
Here is one spot in which it's visible. I would, however, highly recommend people to pull it and check it out locally as the experience from loading indicators is so case dependent. Given the CSS implementation it would also allow one to experiment with the shape and color directly in the browser.

[Screencast from 2023-06-07 15-49-54.webm](https://github.com/OpenRefine/OpenRefine/assets/2631719/0da4b638-1946-4532-b0ed-baede891093c)

tfmorris:
Cool! Thanks for contributing this. I think the concept is great, but perhaps the visuals could be improved:
- the spinners are too small for today's high res displays (a pre-existing problem, not something introduced here)
- the regular (as in lack of variation) design and fine features make the spinners look like they're spinning at a super high rate of speed (or not spinning at all)

Some non-visual suggestions:
- delete the animated GIFs for the old spinners so people aren't tempted to use them (although perhaps we want to keep them with a deprecation period for extension writers?)
- attempting to delete the large spinner showed that the large spinner is still being used by the wikibase extension

I found a CC0/MIT licensed spinner which has some visible variation to the animation (and as an added bonus, is pretty similar to our previous spinner). I've put a branch (#6146) with that code on it for you to look at. 
OpenRefine-spinner

wetneb:
Sorry for having let this linger for so long!
I like the idea too.

I sympathize with @tfmorris' suggestion to have something that looks a bit more like our existing spinner. I was a little bit worried that the one in this PR wouldn't be so easily identifiable as a waiting spinner, but after trying it out I have to say it's pretty clear to me and I also like the style. It does have the benefit of being very simple (so, likely quite easy for browsers to render).

I was also worried about browser support but looking at https://caniuse.com/css-animation it looks like most major browsers implemented this more than 10 years ago so it seems very safe.

I'd say let's migrate the spinners from the wikibase extension and merge this, no?

tfmorris:
Sorry for the delayed reply. I didn't get notified of the comment a couple of weeks ago, so this didn't show up until Antonin replied over the weekend.

> Having tested on both a 4k and 2k screen it looks and it looks good on my end. Do you have a screenshot or example?

Screenshots of the two spinners below, both taken from the same screen.

> 25 elements dependent on timings and absolute positioning to get the opacity changes doesn't seem worth it to me.

It seems to perform fine on my browser(s), but if that's too many elements, we can look for alternatives.

> I sympathize with @tfmorris' suggestion to have something that looks a bit more like our existing spinner.

That's a secondary concern (thus the "added bonus"). My primary concern is that it be easily identifiable as changing visually. The fine pitch, small size, and identical shape of the elements all conspire make the spinner difficult to identify as "spinning" (at least for me). But, I'm fine with going with whatever the designers think is best. There are vast arrays of CSS spinners out there, so we're not limited to these two.

#6146
openrefine-spinner-6146

#5902
openrefine-spinner-5902

thadguidry:
I prefer #6146

wetneb:
I would like to get out of the current stalemate between this PR and #6146, to have one of them merged.
I find it hard to decide on the right tradeoff between implementation complexity and visual appearance.

@cooperzoe what would you go for, from a design perspective?

cooperzoe:
Hi, jumping in here! From a design perspective I would go for #6146.

Abbe98:
I feel like this discussion has derailed from the intent of the pull request. Has someone even played with the properties to change the look?

It should only be the transparency that can't be reproduced with this approach and if that's really important then it should probably be implemented with CSS's `counter()` rather than repeating CSS over and over.

Abbe98:
I'm closing this given that we merged a version of this half a year ago and that we are no longer track upstream. Any contributor is welcome to build upon my changes.

tfmorris:
> I would like to get out of the current stalemate between this PR and https://github.com/OpenRefine/OpenRefine/pull/6146, to have one of them merged.

To be clear, #6146 was never intended as an alternative to be merged. That's why it's marked as draft and warns that it's unfinished. It was intended as a place to hold an implementation with a different visual found in one of the many CSS spinner libraries floating around. It's not my implementation.

It sounds like @Abbe98 objects to some technical aspects of the CSS and also has specific ideas for how something similar could be implemented based on what he has. I don't have the CSS skills/interest to do that work. Since he doesn't want to do it, perhaps one of the @OpenRefine/designers / front end devs could take it on.

PR #5906, opened 2023-06-08
Always including the default types

### LINKED ISSUE #4224: Reconciliation UI only shows one of the available default types
The reconciliation UI does not always show all the default types announced by reconciliation services.

### To Reproduce
1. First, add the reconciliation service `https://reconciliation.brickschema.org/reconcile`
2. Then, try to reconcile any column

### Current Results

The dialog only shows the `BrickClass` type whereas the manifest of the service offers two other types.

### Expected Behavior
All three default types should be offered.

### Screenshots
image

### Versions
 - Operating System: Linux
 - Browser Version: Firefox
 - JRE or JDK Version: 11
 - OpenRefine: master

### Additional context
Brought up in the W3C Entity Reconciliation CG.

ayushrai206:
Fixes #4224 

- Default types available in the manifest are always included in the types list where as previously we only included default types when we did not have any option left.

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
+++ b/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
@@ -55,7 +55,7 @@ ReconStandardServicePanel.prototype._guessTypes = function(f) {
       if (data.code && data.code === 'ok') {
         self._types = data.types;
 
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Very nice, it works! I think we can improve this a little more but it's strictly speaking outside of the issue so I think we can merge this as is.

Issue #5909, opened 2023-06-08
Improve reliability of Coveralls integration


wetneb:
The Coveralls integration could be improved:
* [x] the submission of coverage (and even running Java tests?) could be skipped for PRs which do not make any changes to Java files, resources, and pom.xml files (complete list to be checked!)
* [x] the test coverage of generated Java files (for localization) should be ignored
* [x] any tests showing non-deterministic coverage results should likely be cleaned up to avoid spurious coverage variations (see #5898 for instance)

elebitzero:
@wetneb, tomorrow I can work on splitting the pull_request.yml to only build the Linux and Windows servers based on the path, unless you already doing it. Let me know. Thanks.

wetneb:
That would be amazing, but I would also be happy to do it if you prefer. As you wish!

wetneb:
I looks like one remaining source of coverage statistics going up and down is in the coverage of `CalendarParser.java` which goes up and down depending on the build time (likely because the current time is being parsed somewhere and that triggers different code paths in the parser).

See for instance https://coveralls.io/builds/66381077

If we want more stable coverage statistics we could make sure those code paths are covered by independent tests.

tfmorris:
That particular example looks like a JVM finally enforcing a long standing deprecation of the use of internal class             `sun.util.calendar.ZoneInfo`. Does it correlate with JVM version?

https://coveralls.io/builds/66381077/source?filename=main%2Fsrc%2Fcom%2Fgoogle%2Frefine%2Fexpr%2Futil%2FCalendarParser.java#L764

I have a simple fix buried on a branch somewhere that I could probably dig out.

wetneb:
I can't figure out from the Coveralls report which lines are the 26 lines for which coverage differs from build to build. @pkumar2001 has kindly volunteered to improve the test coverage of this class, perhaps that will surface unreachable paths, but if you have fixes available it's of course great to have them too!

tfmorris:
> I can't figure out from the Coveralls report which lines are the 26 lines for which coverage differs from build to build. 

If you click on the link I posted, they are the lines 785-825 which are highlighted in red (offset by a couple of lines of green in the IllegalAccessException handler).

wetneb:
Closed by #6409 and #6476

tfmorris:
> That particular example looks like a JVM finally enforcing a long standing deprecation of the use of internal class sun.util.calendar.ZoneInfo. Does it correlate with JVM version?

[...]

PR #5910, opened 2023-06-08
Run PR CI on merge commits and not on original PR branch

wetneb:
For #5909.

I am not sure why we were previously running the CI checks on the PR branch itself and not on the merge of the PR branch with master, which is the default behavior. That default makes sense to me since it is more faithful to the end result of the merge.

The reason it is related to #5909 is that it creates spurious coverage decreases as in https://coveralls.io/builds/60565681 (see the file [TransposeColumnsIntoRowsOperation.java](https://coveralls.io/builds/60565681/source?filename=main%2Fsrc%2Fcom%2Fgoogle%2Frefine%2Foperations%2Fcell%2FTransposeColumnsIntoRowsOperation.java#L80), whose coverage increased in master and therefore is reported as a decrease of coverage).

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -56,9 +56,6 @@ jobs:
         echo "COVERALLS_TOKEN=$(echo eUVUVGRHOFJhQm9GMFJBYTNibjVhcWFEblpac1lmMlE3Cg== | base64 -d)" >> $GITHUB_ENV
 
     - uses: actions/checkout@v3
-      with:
-        ref: ${{ github.event.pull_request.head.ref }}
-        repository: ${{ github.event.pull_request.head.repo.full_name }}
 
     - name: Set up Java ${{ matrix.java }}
       uses: actions/setup-java@v3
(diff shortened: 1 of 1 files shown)


elebitzero (commented):
Coveralls said the coverage still decreased: 
`coverage/coveralls — Coverage decreased (-0.09%) to 55.528%`

Even with the change the OR CI might still be run on the original branch and not the merge commits.

wetneb:
Looking at the build logs, we are indeed checking out the merge of master and the PR:
```
Checking out the ref
  /usr/bin/git checkout --progress --force refs/remotes/pull/5910/merge
  Note: switching to 'refs/remotes/pull/5910/merge'.
```

I think the remaining coverage decrease is likely coming from a different problem (some non-determinism in tests, perhaps? see #5909). At least the Coveralls report only mentions a decrease in coverage in [main/src/com/google/refine/expr/util/CalendarParser.java](https://coveralls.io/builds/60607979/source?filename=main%2Fsrc%2Fcom%2Fgoogle%2Frefine%2Fexpr%2Futil%2FCalendarParser.java#L778), whose coverage has not changed recently.

elebitzero (approved):

PR #5912, opened 2023-06-09
Made CI skip Coveralls if Java files aren't changed

elebitzero:
Part of https://github.com/OpenRefine/OpenRefine/issues/5909.

Added job that gets the list of changed files and checks if any are within one of the java directories, then changed=true, otherwise false.

If the Java source is changed, then Java tests are run and the Coveralls report is run.

- Made it so Coveralls are skipped on CI if no Java files are altered.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -26,10 +26,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-        with:
-          repository: ${{ github.event.pull_request.head.repo.full_name }}
-          ref: ${{ github.event.pull_request.head.ref }}
-          fetch-depth: 0
 
       - name: Get changed files
         id: getfiles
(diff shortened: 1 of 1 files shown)


wetneb:
Would it not be cleaner to use the [native path filtering options](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) for that? It would avoid spinning up a job just to do this computation. It looks like this is only available for the entire workflow, and not an individual job, but perhaps we could just split the `pull_request.yml` workflow into two? The Windows CI job could go with the two Linux Java CI jobs.

elebitzero:
I'm not sure, it seems more complicated to split it into multiple workflows, but maybe because I haven't seen it done that way.

I just got it working but I need to try changing a Java source file to see if the Java tests and Coveralls report runs correctly.

I'll do that tomorrow. Let me know by tomorrow if you want me to continue on this.

Thanks,

wetneb:
To me it seems simpler for the following reasons:
* no need for an additional job
* the Java CI jobs would not get executed at all, instead of booting up to check if they should execute or not (producing spurious green checks in the list of checks run on the PR)
* we don't need to pull in another github action
* the `pull_request.yml` file is getting quite long and separating the backend tests from the e2e ones seem like a good opportunity to make the size more manageable

elebitzero:
If we change windows_server_test to always skip the Java tests, then you could move the check_java_src_changes into the linux_server_test and avoid creating another job. I had to have as a separate job because both the Windows and Linux tests were using the result.

But if you are saying to skip all server tests if no Java code is changed, and to rely on the e2e tests, I wasn't thinking that is possible but maybe it is ok.

I thought there may be cases where you make some change that isn't in the Java source directory, the server tests should still be run to make sure nothing got broken, like if the refine or refine.bat script is modified.

PR #5916, opened 2023-06-13
Update import to support lower versions of Nodejs

### LINKED ISSUE #5880: Running into dependency issue on Linux
### To Reproduce
I started a linux instance and installed all dependencies. I am able to run `./refine` after cleaning and building. However I am running into some issues where it cannot locate the `select2`, `jQuery` when running the server on the browser. It looks like some dependency installation issue, but not sure how to fix 

### Current Results
```
GET http://127.0.0.1:3333/3rdparty/select2/select2.css net::ERR_ABORTED 404 (Not Found)
:3333/:44 

 GET http://127.0.0.1:3333/3rdparty/tablesorter/theme.blue.css net::ERR_ABORTED 404 (Not Found)

index-bundle.js:18 Uncaught ReferenceError: jQuery is not defined
    at index-bundle.js:18:12
    at index-bundle.js:20:4

VM12:1 Uncaught TypeError: Cannot set properties of undefined (setting 'encodings')
    at VM12:1:18
```

### Expected Behavior
Fully running open refine

### Screenshots
image

### Versions
 - Operating System: Linux 
 - Browser Version:Chrome
 - 
 - JRE or JDK Version: openjdk 11.0.19 2023-04-18
 - OpenRefine: 4.0 branch

5tigerjelly:
Fixes #5880

-  update import
- 
-


wetneb:
Thanks! Looking at the CI, it seems that your new version fails to install our frontend dependencies (notably, jQuery) so the app is not functional anymore with this change.

Abbe98:
Shouldn't we prefer to stick with ES6 here? Going back to non standard CommonJS feels backwards to me. `require` and `import` also work slightly differently.

wetneb:
That would be fine for me too! Do you happen to know what minimum version of nodejs this entails, so that we could document that?

wetneb:
I can say it works with NodeJS 16 so that's an upper bound on the requirement :)

wetneb:
Let's close this and document that Nodejs >= 16 is required to compile OpenRefine. If anyone is aware of an earlier version which also works, we can change that.

PR #5917, opened 2023-06-13
now the default types are displayed with both name and id

### LINKED ISSUE #5907: Default reconciliation types are rendered without name (only id)
The default types proposed by a reconciliation service are proposed to the user by showing only their id, not their human-readable label.

### To Reproduce
1. First, import this dataset: https://opendata.paris.fr/explore/dataset/lieux-de-tournage-a-paris/ (CSV with `;` delimiter: https://opendata.paris.fr/api/explore/v2.1/catalog/datasets/lieux-de-tournage-a-paris/exports/csv?lang=fr&timezone=Europe%2FBerlin&use_labels=true&delimiter=%3B)
2. Then, open the reconciliation dialog to reconcile the "Titre" column to Wikidata

### Current Results

Observe that the last type in the list appears as an identifier (without a name).

### Expected Behavior

Like other type options in the list, the type should be provided as a pair of (name, id).

### Screenshots

image

### Versions
 - Operating System: Linux
 - Browser Version: Firefox
 - JRE or JDK Version: 17
 - OpenRefine: 3.8-SNAPSHOT

### Additional context
Follow-up to #4224.

ayushrai206:
Fixes #5907 

- Both name and id are displayed for the default types now.

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
+++ b/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
@@ -67,7 +67,7 @@ ReconStandardServicePanel.prototype._guessTypes = function(f) {
             if (defaultTypes.hasOwnProperty(id)) {
               self._types.push({
                 id: id,
-                name: defaultTypes[id].name
+                name: defaultTypes[id]
               });
             }
           }


wetneb (approved):
Let's not add a Cypress test for it since it relies on a third-party service.

PR #5918, opened 2023-06-13
Made facet by judgment resizable

### LINKED ISSUE #5908: Make reconciliation judgment facets resizable like other text facets
The facets for reconciliation judgment seem to be intentionally not resizable and it is not clear to me it is a good thing.
The number of choices display in the facet can grow after the facet was created, which will cause the choices to overflow and make the facet unusable.
Also, the visual inconsistency between this facet and others is confusing to me as a user.

### To Reproduce
1. First, reconcile a column, which creates a judgment facet - this one is resizable.
2. Second, delete that facet and re-create it from the Reconcile -> Facets menu. It is not resizable.
3. Then, mark some cells as being matched to new items
4. Finally, observe the glitch in facet rendering

### Current Results

Initially (note the absence of resizing slider in the judgment facet):
image

After marking items as new (note the overflowing "Facet by choice counts"):
image

### Expected Behavior

Make the facet resizable like all others and do not overflow when choices change.

### Versions
 - Operating System: Debian
 - Browser Version: Firefox
 - JRE or JDK Version: 17
 - OpenRefine: 3.8-SNAPSHOT (master)

elebitzero:
Fixes #5908

- Made facet by judgement resizable and scrollable.

### DIFF
--- a/main/src/com/google/refine/operations/recon/ReconOperation.java
+++ b/main/src/com/google/refine/operations/recon/ReconOperation.java
@@ -148,7 +148,7 @@ public class ReconOperation extends EngineDependentOperation {
                 "    \"name\" : \"" + _columnName + ": " + OpenRefineMessage.recon_operation_judgement_facet_name() + "\"\n" +
                 "    },\n" +
                 "    \"facetOptions\" : {\n" +
-                "      \"scroll\" : false\n" +
+                "      \"scroll\" : true\n" +
                 "    },\n" +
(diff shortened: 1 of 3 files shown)


wetneb (approved):
Thank you!
I actually wonder if any of the facets with `scroll: false` should actually stay like that, because the scenario I described in the issue might well apply to those as well… Happy either way.

elebitzero:
@wetneb I'll also check Best Type Match and Best Candidate Name Match. I will fix them also.

PR #5920, opened 2023-06-13
fix(sec): upgrade org.eclipse.jetty:jetty-server to 12.0.0.beta0

LyonLee-Viktor:
### What happened？
There are 1 security vulnerabilities found in org.eclipse.jetty:jetty-server 9.4.51.v20230217
- [CVE-2023-26049](https://www.oscs1024.com/hd/CVE-2023-26049)

### What did I do？
Upgrade org.eclipse.jetty:jetty-server from 9.4.51.v20230217 to 12.0.0.beta0 for vulnerability fix

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,7 +77,7 @@
     <butterfly.version>1.2.4</butterfly.version>
     <slf4j.version>2.0.7</slf4j.version>
     <log4j.version>2.20.0</log4j.version>
-    <jetty.version>9.4.51.v20230217</jetty.version>
+    <jetty.version>12.0.0.beta0</jetty.version>
     <okhttp.version>4.10.0</okhttp.version>
     <jena.version>4.8.0</jena.version>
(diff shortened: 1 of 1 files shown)


wetneb:
Let's see if a version of Jetty 9 with a fix for this comes out.

Abbe98:
These OSCS pull requests seems like automated spam?

tfmorris (commented):
According to the CVE (the [actual CVE,](https://nvd.nist.gov/vuln/detail/CVE-2023-26049) not the clone linked above), this vulnerability is fixed in 9.4.51 which is the version we are currently running.

Also, it's almost never appropriate to update a dependency from a production version to a beta version AND Jetty 10-12 are incompatible with Jetty 9 (thus the major version bumps), so without additional source code changes, this doesn't even build (as evidenced by the failures above).

wetneb:
Thanks for checking this! Then we can close that.

PR #5921, opened 2023-06-13
removing flyout services

### LINKED ISSUE #3141: XSS vulnerability in suggest flyouts

The flyout services in reconciliation endpoints can return arbitrary HTML which is directly included in the DOM without sanitization:

https://reconciliation-api.github.io/specs/latest/#flyout-services

This means that reconciliation services can execute arbitrary Javascript in the user's browser when they hover a suggested item in an autocompleted field.

Drop flyout services.

Change the specs so that the flyout services only return data following a fixed schema (a description string, for instance, but that would be redundant with the description that can be returned in the suggest responses directly).

Or cook up some way to sanitize the HTML returned by the service, by removing `<script>` tags for instance, but that is bound to be brittle.

**Additional context**

Similar to #3140.

ayushrai206:
Fixes #3141 

- Simply deleting flyout services
- 
-

### DIFF
--- a/main/webapp/modules/core/externals/suggest/css/suggest-4_3.css
+++ b/main/webapp/modules/core/externals/suggest/css/suggest-4_3.css
@@ -16,7 +16,7 @@ Author: Kai Conragan (kai@metaweb.com)
     .fbs-subpane    : Inner container for result list <div>
     .fbs-list       : Contained by .fbs-subpane, this is the actual results list <ul>
     .fbs-item       : Individual item in results list <li>
-    .fbs-flyout-pane : Outermost container for flyout on specific object <div>
+
 
 =======================================================
 */
@@ -67,55 +67,36 @@ below rules into your own stylesheet to customize.
 
 */
 
-.fbs-pane,
-.fbs-flyout-pane,
-li.fbs-nomatch,
-a.fbs-more-link,
-li.fbs-selected,
-.fbs-flyout-images,
-#fbs-topic-image { 
-    background: #fff; 
-}
+
 
 
 .fbs-suggestnew { 
     color: #444;
 }
 
-.fbs-pane,
-.fbs-flyout-pane,
-.fbs-flyout-subtitle,
-.fbs-topic-properties strong { 
-    color:  #666;
-}
+
 
 h3.fbs-topic-properties,
-.fbs-flyout-pane p { 
-    color: #444;
-}
+
 
 .fbs-item-name,
 li.fbs-help-on-focus,
 li.fbs-nomatch,
 .fbs-nomatch-text,
-.fbs-flyout-pane h3,
+
 .fbs-properties-header { 
     color: #333;
 }
 
 .fbs-pane,
-.fbs-flyout-pane { 
-    border: 1px solid #bbb;
-    padding: 2px;
-}
 
-.fbs-flyout-pane {
-    border-color: #ccc;
-}
+
+
+
 
 .fbs-list,
 .fbs-list-icons,
-.fbs-flyout-content,
+
 .fbs-attribution, 
 .fbs-header { 
     background-color: #f5f5f5;
@@ -132,9 +113,7 @@ li.fbs-nomatch {
     border: 1px solid #f5f5f5;
 }
 
-h1#fbs-flyout-title { 
-    color: #f60;
-}
+
 
 li.fbs-selected .fbs-item-name {
     border-color: #f60;
@@ -153,9 +132,7 @@ li.fbs-selected .fbs-item-name {
 }
 
 
-#fbs-flyout-title .fbs-flyout-label { 
-    color: #aaa;
-}
+
 
 .fbs-citation {
   white-space: nowrap;
@@ -165,22 +142,17 @@ li.fbs-selected .fbs-item-name {
 
 /* topic image */
 #fbs-topic-image,
-.fbs-flyout-images {
-    border: 1px solid #a9a9a9;
-}
+
 
 .fbs-suggestnew-button,
-.fbs-flyout-pane { 
-    border: 1px solid #9a9a9a;
-    color: #999999;
-}
+
 
 .fbs-suggestnew-button {
     color: #444444;
 }
 
 ul.fbs-list,
-.fbs-flyout-content,
+
 .fbs-attribution,
 .fbs-header {
     border: 1px solid #dae3e9;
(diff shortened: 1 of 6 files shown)


main/webapp/modules/core/externals/suggest/css/suggest-4_3.css
```diff
@@ -67,55 +67,36 @@ below rules into your own stylesheet to customize.
...
 
-.fbs-pane,
-.fbs-flyout-pane,
-li.fbs-nomatch,
-a.fbs-more-link,
-li.fbs-selected,
-.fbs-flyout-images,
-#fbs-topic-image { 
-    background: #fff; 
-}
```
wetneb, line 78: Here you are removing not just flyout styling, but also the styling on a lot of other elements: `.fbs-pane`, for instance, should likely stay. See the syntax for CSS selectors: there is a difference between `.fbs-pane .fbs-flyout-pane` (which could be removed entirely) and `.fbs-pane, .fbs-flyout-pane` (which should be simplified to just `.fbs-pane`).

main/webapp/modules/core/externals/suggest/css/suggest-4_3.css
```diff
@@ -67,55 +67,36 @@ below rules into your own stylesheet to customize.
...
 .fbs-suggestnew { 
     color: #444;
 }
 
-.fbs-pane,
-.fbs-flyout-pane,
-.fbs-flyout-subtitle,
-.fbs-topic-properties strong { 
-    color:  #666;
-}
```
wetneb, line 90: Similar problem here (and in the rest of your CSS changes).

main/webapp/modules/core/externals/suggest/css/suggest-4_3.css
```diff
@@ -571,13 +487,13 @@ to inline Internet Explor fixes
 */
 
 * html .fbs-list,
-* html .fbs-list-icons { height: expression( this.scrollHeight > 410 ? "410" : "auto" ); /* sets max-height for IE */ }
+* html .fbs-list-icons { height: expression( this.scrolllHeight > 410 ? "410" : "auto" ); /* sets max-height for IE */ }
```
wetneb, line 490: This change is probably a mistake.

main/webapp/modules/core/externals/suggest/suggest-4_3a.js
```diff
@@ -1089,71 +1064,6 @@
...
-      this.flyout_image_url = o.flyout_image_service_url;
-      if (o.flyout_image_service_path) {
-          this.flyout_image_url += o.flyout_image_service_path;
-      }
-      // set api key for image api
-      this.flyout_image_url = this.flyout_image_url.replace(/\$\{key\}/g, o.key);
-
-      if (!$.suggest.cache) {
-        $.suggest.cache = {};
-      }
```
wetneb, line 1113: Those three lines above likely need to be kept, I think the cache is still used by the suggest service independently of the flyout.

main/webapp/modules/core/externals/suggest/suggest-4_3a.js
```diff
@@ -1089,71 +1064,6 @@
...
-          $.suggest.flyout = {};
-        }
-        if (!$.suggest.flyout.cache) {
-          $.suggest.flyout.cache = {};
-        }
-      }
-    },
-
-    _destroy: function() {
-      base._destroy.call(this);
```
wetneb, line 1151: I would likely keep this `_destroy` function since it might be doing things that are unrelated to the flyout functionality, and only remove the flyout-related code in it.

main/webapp/modules/core/externals/suggest/suggest-4_3a.js
```diff
@@ -1163,15 +1073,6 @@
       }
     },
 
-    hide_all: function(e) {
-      this.pane.hide();
```
wetneb, line 1167: Same as above

main/webapp/modules/core/externals/suggest/suggest-4_3a.js
```diff
@@ -1319,34 +1220,10 @@
...
-    },
-
     check_response: function(response_data) {
       return response_data.prefix === this.input.val();
     },
 
-    response_hook: function(response_data, cursor) {
-      if (this.flyoutpane) {
-        this.flyoutpane.hide();
-      }
```
wetneb, line 1341: Same here, I suspect?

main/webapp/modules/core/externals/suggest/suggest-4_3a.js
```diff
@@ -1712,15 +1415,7 @@
         ]
       },
 
-      // CSS default class names
-      css: {
-        item_type: "fbs-item-type",
```
wetneb, line 1717: Is this `item_type` flyout-related?

main/webapp/modules/core/styles/common.css
```diff
@@ -466,10 +466,6 @@ img {
   display: none;
 }
 
-.fbs-pane, .fbs-flyout-pane {
-  z-index: 2000;
```
wetneb, line 470: Same comment about CSS rules.

wetneb (commented):
I haven't tested it yet but there are a couple of places where I suspect you have been a tad zealous ^^ It's just guesswork on my side so I might be wrong though.

In any case it's definitely worth testing the suggest functionality thoroughly after your changes and making sure the visual appearance and functionality is preserved. Did you already do that on your side?

ayushrai206:
> I haven't tested it yet but there are a couple of places where I suspect you have been a tad zealous ^^ It's just guesswork on my side so I might be wrong though.
> 
> In any case it's definitely worth testing the suggest functionality thoroughly after your changes and making sure the visual appearance and functionality is preserved. Did you already do that on your side?

Hey, you are absolutely right! I have now taken care of all your feedback and tested the functionality from my end!
Thank you so much!

elebitzero (commented):
Hi @ayushrai206,  `externals\suggest` is a 3rd party library.  We should not be altering these files if at all possible:
* externals/suggest/suggest-4.3a.js
* externals/suggest/css/suggest-4_3.css

This is a 3rd party UI library. It is not included in package.json because it was already altered awhile ago with some bug fixes. We want to avoid making any other changes to these files.

Is there some other way to disable the flyout without modifying these files directly?

wetneb:
I was also wondering about that… I am not sure to what extent this file is really an external library. I assume it's not maintained by another project (otherwise we would likely be pulling it from npm, no?) and we have already made edits to it, so I wonder if we shouldn't just treat it as an internal file. No strong feelings about this though.

In this particular case I think we can avoid making changes to it, by just making sure no flyout settings are passed to it when initializing it on input fields. We'd have to be sanitizing the suggest settings offered in the manifest.

wetneb:
I'd propose to close this PR and do it again with the approach suggested by @elebitzero. The changes should be quite minimal: before initializing the suggest library, we would remove the flyout settings from it.

tfmorris:
I agree with @elebitzero 's approach, but just to clarify the provenance of this file, it's a private fork of Freebase Suggest 4.3 which is no longer maintained by Google. While we control the fork, it's best to minimize changes when possible, as it appears to be in this case.

PR #5922, opened 2023-06-14
Pull request workflow split

elebitzero:
For #5909.

  - Split pull_request.yml workflow into two separate workflows:
    * pull_request_server.yml: Runs Linux server build and install if any file besides UI and localizations are modified.  
    * pull_request_e2e.yml: Runs e2e tests if any file besides localizations are modified.


wetneb (approved):
Exciting, thank you so much for that! Let's just try it out.

PR #5930, opened 2023-06-16
Bump Java's maximum version to 20

wetneb:
I haven't tried if it actually works - let's see what the CI says.

### DIFF
--- a/.github/workflows/pull_request_server.yml
+++ b/.github/workflows/pull_request_server.yml
@@ -47,7 +47,7 @@ jobs:
   linux_server_tests:
     strategy:
       matrix:
-        java: [ 11, 17 ]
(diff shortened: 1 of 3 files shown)


Abbe98 (approved):
Looks good, and we have no need to continue testing Java 17?

wetneb:
I would intuitively say that testing the lower and upper bounds are enough.

Or to diversify a bit we could take one of the two Java 20 jobs and turn it back into a Java 17 job (say, Linux with Java 17 and Windows with Java 20)?

wetneb:
I've just added a commit to change the E2E tests to use Java 17, so that we have a diversity of java versions (11, 17, 20) in the CI.

PR #5931, opened 2023-06-16
Remove maximum Java version on Windows.

### LINKED ISSUE #5583: OpenRefine 3.7 beta 2 for Windows not executing
OpenRefine 3.7 beta 2 for Windows OS is not installing when attempting to download and install from the .exe file with latest version of java already installed using latest version of Chrome

### To Reproduce
1. Download zip file for Windows OS, OpenRefine 3.7 beta 2 with java
2. Unzip file
3. Update and install latest version of java on local machine and latest version of Chrome
4. Open openrefine.exe (application file)

### Current Results
Command line loaded as expected, but application file didn't execute to open the actual application, with latest Chrome browser installed execution of the file opens a new tab which displays the OpenRefine website in the Java section of the page (https://openrefine.org/docs/manual/installing#java)

### Expected Behavior
Expecting to open the application and execute the file

### Screenshots
NA

### Versions
 - Operating System: Windows 10 Enterprise
 - Browser Version: Chrome Version 109.0.5414.120 (Official build)
 - JRE or JDK Version: Version 8, update 361 (build 1.8.0_361-b09)
 - OpenRefine: OpenRefine 3.7 beta 2 with java

### Datasets
NA

### Additional context
NA

wetneb:
This hopefully fixes #5583, but makes sense independently of it I think.

As mentioned in the issue, the upper bound on Java compatibility is only based on what we test for. Given that Java runtime environments are generally very backwards compatible, it feels wrong to me to refuse launching if Java is more recent than the versions of Java we test OpenRefine with. So I would just propose to drop this maximum version altogether.

I would propose to include this in a 3.7.3 release.

### DIFF
--- a/packaging/pom.xml
+++ b/packaging/pom.xml
@@ -443,7 +443,6 @@
               <jre>
                 <path>%JAVA_HOME%;%PATH%</path>
                 <minVersion>${java.minversion}</minVersion>
-                <maxVersion>${java.maxversion}</maxVersion>
                 <initialHeapSize>512</initialHeapSize>
                 <maxHeapSize>2048</maxHeapSize>
                 <opts>
--- a/pom.xml
+++ b/pom.xml
@@ -139,10 +139,9 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-	<version>${maven-compiler-plugin.version}</version>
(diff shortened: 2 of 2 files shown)


thadguidry (approved):
This change is fine.  Well, as long as our dependencies continue to play nice I would say, but even then, all of them are post Java 9 (last time I checked) and so we shouldn't have binary forward compatibility issues at the runtime level.

PR #5934, opened 2023-06-16
Adding engines field for npm, yarn, and node.

### LINKED ISSUE #5880: Running into dependency issue on Linux

elebitzero:
Fixes #5880

Node 14 and below are end-of-life, and Node 16 reaches end-of-life October 2023.
(Ref: https://nodejs.dev/en/about/releases/)

So we should no longer support Node 10, 12, and 14 for development of OpenRefine.

Added engines field in main/webapp and main/tests/cypress/package.json file to specify the minimum node, npm, and yarn versions.
Note, that currently the build uses npm for main/webapp and yarn for main/tests/cypress. We should probably use one or other.

For the node, npm, and yarn versions I took the minimum of the versions between @wetneb and myself.  By Oct'23, we should bump it up to node 18.x.

If someone runs 'npm install' or 'yarn install' with a lower version, they will get a warning on the command line.


wetneb:
Thanks!
Looks like the CI would otherwise work fine with those versions:
```
  node: v14.21.3
  npm: 6.14.18
  yarn: 1.22.19
```

So maybe Node 14 is still fine? Intuitively, if it works, why not allow it?

elebitzero:
I forgot to update the GitHub workflows to use Node 16 instead of 14.  I think we should be on 16 since 14 has reached end-of-life.

There is already one dependency cypress-wait-until which is requiring Node 18. (https://github.com/OpenRefine/OpenRefine/pull/5924)
```
error cypress-wait-until@2.0.0: The engine "node" is incompatible with this module. Expected version ">=18.16.0". Got "14.21.3"
```

But let's just go to 16 for now.

wetneb:
I'd be in favour of still allowing Node 14 given that people seem to have even much older versions of Node around (Node 10 in #5880).

Node is not something we are using to run OpenRefine itself, it is just used in the build process to download dependencies and to run Cypress tests, so the likelihood that a vulnerability in Node would lead to a vulnerability in OpenRefine is very small.

elebitzero:
I agree that security fixes are not a big concern. It is just generally best practice to keep the build tools updated. But I did set the node version to `>=14.0.0`.

Also, while I was looking at the Cypress dependencies, I found that cypress-wait-until is not used, so we can remove it. I will post as a separate PR for that.

PR #5940, opened 2023-06-20
render geo: URIs as links in cells

Abbe98:
https://en.wikipedia.org/wiki/Geo_URI_scheme

### DIFF
--- a/main/webapp/modules/core/scripts/util/url.js
+++ b/main/webapp/modules/core/scripts/util/url.js
@@ -37,6 +37,7 @@ var URLUtil = {
     "chrome":1,
     "file":1,
     "ftp":1,
+    "geo":0,
     "http":1,
     "https":1,
     "imap":1,


wetneb (approved):
Looks good to me.

wetneb:
It works for me both with Firefox and Chromium on Debian. It offers to open Gnome Maps, which then shows a slippy map with a marker at the desired coordinates.

Abbe98:
With FF and Chromium on Fedora I'm given the following options: Gnome Maps, Google Maps, and OpenStreetMap

ostephens:
I've been trying to find how well supported this is across browsers, and haven't been able to find a lot - but support looks patchy tbh (this page doesn't list browsers but it does offer some insight into adoption https://wiki.openstreetmap.org/wiki/Geo_URI_scheme and it's not great)

However, that said it seems to me reasonably likely that if you are dealing with data with Geo URIs then you'll have some understanding of what they are - so I'm OK with this display if it's helpful to those users

wetneb:
Interesting, I did not suspect this could be platform dependent.

ostephens:
There is a plugin for FF. Not tried it but https://addons.mozilla.org/en-GB/firefox/addon/geo-uri-handler/

Abbe98:
> However, that said it seems to me reasonably likely that if you are dealing with data with Geo URIs then you'll have some understanding of what they are - so I'm OK with this display if it's helpful to those users

I could imagine this being the situation for most schemes other than mailto, tel, http, https, etc.

tfmorris:
By "common" I meant browsers with substantial [market share](https://gs.statcounter.com/browser-market-share/desktop/worldwide) - and, by extension, the [operating systems](https://gs.statcounter.com/os-market-share/desktop/worldwide/) they run on.

Safari on Mac at least pops up an error dialog, but Chrome on Mac only logs an error to the Javascript console.

What happens on **popular** browsers like Chrome on Windows? That's what should shape the UX decisions.

PR #5941, opened 2023-06-20
Make project initialization requests in parallel

Abbe98:
Prior to this the UI would make a call to `get-project-metadata`, then when it finished to `get-models`, then to `get-all-preferences` during project initialization. Now all these requests are all made in parallel, greatly speeding up project initialization.


elebitzero (approved):
I tried this out and it works. I didn't notice a change in speed but the code is definitely more clean.

PR #5944, opened 2023-06-21
Expose Wikibase editing errors in the grid

wetneb:
Closes #1782.

This follows the design proposed [on the forum](https://forum.openrefine.org/t/partial-results-of-long-running-operations/482).
The Wikibase upload operation creates a new column (or reuses the column, if it already exists) to store the editing errors of edits related to a given row:
image
Runnning the operation also creates a facet which can be used to select the rows based on whether they encountered errors or not:
image
This facet is generated by the GREL expression `if(isError(value), 'failed edit', 'successful edit')`.

For successful edits I think it would be nice to store some value (to distinguish it from no edit being made). I thought about adding a URL to the diff view of the edit (such as https://www.wikidata.org/w/index.php?diff=1920784148&oldid=1919995661).

This also closes #5166, since it makes it possible to relate the errors to the offending lines better.

You can try this PR on [this sample project](https://github.com/OpenRefine/OpenRefine/files/11822075/Wikibase-editing-error-example.openrefine.tar.gz): uploading that project to Wikidata will not make any edits (the first row fails to create an item, the second one does not make any edit because the data being pushed matches what is already on Wikidata).

cc @OpenRefine/designers

This is meant to be a basic implementation that can be improved on (for instance, localization can be improved). I would like to release this in 4.0-alpha2 to gather feedback about the UX. One other question is where to insert the column with the editing results. Currently it is added after all other columns, which might be a bit inconvenient for projects with many columns.

Issue #5788, opened 2023-04-17
Eliminate places where the entire grid is loaded in memory in 4.0


wetneb:
In the 4.0 architecture, we add support for working on datasets which do not fit in memory.
This already works broadly speaking, but there are still use cases where the entire grid will be loaded in memory.

Some of those can be avoided: if the effort to do so is reasonable, it is worth doing it.
Some others cannot be avoided for various reasons. In that case we should make sure the corresponding operation is refused to the user ahead of time (so that we have proper error reporting to the user instead of dying with an `OutOfMemoryException`) or at least there is proper warning that the user could run into such an error.

This is a tracker issue to identify all of the places where such work is still required.

## Operations
* [ ] transpose operations. Some of those will unavoidably load the whole dataset in memory (we need a warning for that), but the key-value columnize operation should be able to avoid this.
* [ ] clustering (although that's technically not an operation itself). There should be an option (similar to the one for facets) to only process a sample of the rows by default on large grids.
* [x] all other operations (I think!)

## Browsing
* [x] facet computation. Facets are only computed on a subset of the grid when the grid becomes large, so that the facet statistics are returned quickly. The user can configure how many rows/records are processed.
* [ ] records browsing. This is already working fine on large grids, except if the grid contains an unreasonably large record (for instance spanning the entire grid). The size of records should be bounded by default (with the option to disable this bound) so that people do not accidentally crash OpenRefine by turning on the records mode on such grids.
* [ ] sorting a grid. This loads the entire grid in memory systematically.

## Importers:
* [x] CSV and other line-based importers do not load everything in memory
* [x] ODS can likely not avoid loading the entire file, but that's likely okay because those files have a limited length anyway
* [ ] XLSX should take advantage of Apache POI's streaming support to avoid loading everything
* [ ] JSON/XML importers should not need to load everything in memory
* [ ] RDF importer should not need to load everything in memory
* [ ] SQL importer should not have to load everything in memory
* [ ] Wikibase edits preview and quality assurance should work on a sample if the dataset is too large

## Exporters
* [ ] XLSX: we could take advantage of POI's streaming support again
* [x] ODS: likely impossible
* [ ] Templating: check that we are already doing it in a streaming fashion
* [ ] CSV/TSV: check that we are already doing it in a streaming fashion

I am likely forgetting a lot of places which will show up during testing.

Issue #1782, opened 2018-10-27
Find a way to report failures of item creations for new editors in Wikidata extension


wetneb:
**Describe the bug**
It seems that some new item creations are blocked for new editors. Maybe we should just slow down the edit rate to help with that (check needed with Wikidata admins).
We need a way to report creation failures to the user - for now we need to look up the logs but they are not always easily available (for instance when running OR on a mac).

wetneb:
One possibility would be to add a new column containing the editing errors, when they happen.

madham32:
Hi! Can I try working on this issue? :)

madham32:
Could you tell me how to reproduce these errors?

wetneb:
You could create a fresh Wikidata account, without any edits, and then try to [create a new wikidata item from OpenRefine itself](https://www.wikidata.org/wiki/Wikidata:Tools/OpenRefine/Editing/New_items).

If you already have a Wikidata account, make sure you are logged in with your existing account when you sign up for a new account. This will make sure both accounts are linked and the system recognizes them as belonging to the same person (to avoid breaching the policy on multiple accounts).

In general there are many other reasons why Wikidata edits might be rejected - this ticket could be broadened to improve the reporting of these errors in general.
For instance, if you try to create an item whose English label and description are identical to those of another item, this should be rejected as well (I think).

wetneb:
Cc @darecoder1999 who is interested in this issue

antoine2711:
I get this error on the back-end. I would like to see that in the front-end.
```
01:15:49.130 [                   refine] POST /command/wikidata/perform-wikibase-edits (4ms)
01:15:49.131 [..mWikibaseEditsOperation] Performing edits (1ms)
01:15:49.131 [..ting.EditBatchProcessor] Requesting documents (0ms)
01:15:49.515 [..ting.EditBatchProcessor] MediaWiki error while editing [modification-failed]: Item [[Q111909481|Q111909481]] already has label "puppet" associated with language code en, using the same description text. (384ms)
01:15:49.793 [..ting.EditBatchProcessor] MediaWiki error while editing [modification-failed]: Item [[Q111909481|Q111909481]] already has label "puppet" associated with language code en, using the same description text. (278ms)
01:15:50.077 [..ting.EditBatchProcessor] MediaWiki error while editing [modification-failed]: Item [[Q111909481|Q111909481]] already has label "puppet" associated with language code en, using the same description text. (284ms)
[...]

PR #5951, opened 2023-06-29
New data processing architecture

wetneb:
Following the [discussion on the forum](https://forum.openrefine.org/t/releasing-4-0-alpha2/833/7) I am opening this draft PR to give some visibility to my work on a new data processing architecture.

## Overview of the architectural changes

This changes the way project data is represented internally. Instead of being systematically loaded in Java heap as a two-dimensional array of cells, the project state is represented by a `Grid`, instance of an interface which abstracts away access and processing operations on project data. This has various useful consequences:
* the Grid implementation is free to choose from various strategies to represent the underlying data: it can be backed by a similar in-memory representation, or be read dynamically from disk as needed, or even by dynamically computed from another data source (such as an `Iterable<Row>`). This makes it possible to work on projects which do not fit in RAM.
* Grids are immutable objects. Therefore running a transformation on a project involves deriving a new grid from the existing one. The `History` class contains `Grid` objects for all the steps in the transformation history of the project. This gives more guarantees about data integrity: for instance, it prevents commands from making changes to project data without logging those in the history.
* this simplifies the implementation of operations, which only need to implement the application of their changes and not their rollback. Inconsistencies between those had caused some data corruption issues in the past.

The Grid interface and other auxiliary interfaces (such as `ChangeData`, a data store for the data computed by long running operations, and `Runner`, a factory class for `Grid` and `ChangeData`) come with multiple implementations:
* the `LocalGrid` (associated to the `LocalRunner` and `LocalChangeData`) is the default implementation. It is designed to be efficient in the current usage domains of OpenRefine (data processing on a single machine, hence *local*), while also allowing for larger datasets as well. It uses lazy evaluation to load data on demand, with the option of caching project data in memory when its size allows.
* the `TestingGrid` is a reference implementation, which is designed to be as simple as possible and systematically load project data in memory. It is used for tests.
* the `LoggedGrid` is a wrapper on top of another implementation which logs all operations run on a given Grid and their timings. This is mostly useful for debugging and optimization purposes.
* the `SparkGrid`, available in [a separate extension](https://github.com/OpenRefine/refine-spark). This uses [Apache Spark](https://spark.apache.org/) as execution backend and is therefore more suitable in distributed contexts. Running it locally is also possible but will be generally slower than with the default implementation. As mentioned in the README of this extension, there are a couple of hurdles to this implementation being really usable in a distributed context. Tackling those would be possible but I do not see it as a priority given that my goal is to serve the existing user base first.

To enable those changes, the structure of project serialization was changed. This version uses by default a different workspace directory compared to 3.x. Archives of projects exported from 3.x can be imported in this version, but without their transformation history: the project will be imported in the state it was left in in 3.x.

## Features enabled by those changes

This is a broad overview: other smaller features or bug fixes can be found in the 4.0 milestone.

* Better handling of large datasets. There are still cases where the entire project data will be loaded and some of those could be avoided. Issue #5788 gives an overview of this state;
* Ability to limit the compute the state of facets on a sample of rows or records, to make the interface more responsive;
* Support for displaying partial results of long-running operations. This means showing the results of operations such as reconciliation, fetching URLs as they are computed, such that the user can start working with those without having to wait for the completion of the operation run. This feature is discussed in this thread;
* Ability to recover long-running operations from crashes;
* Ability to pause and resume long-running operations;

It also includes some solutions to long-standing issues whose designs were suggested by the new architecture, such as:
* #33: preservation of the pagination position when applying operations which preserve rows or records
* #1782 (not merged yet, see #5944)

## Timeline

I would like to release a snapshot of this branch as an alpha version in the coming weeks (I propose 4.0-alpha2, since I already published 4.0-alpha1 earlier). I want to get this architecture tested widely after this release. I would only merge this PR once we are ready to publish a stable release out of it. Depending on the scale of the changes to make based on that feedback, I anticipate it could take some number of months.


wetneb:
I am closing this as I opened a series of more focused PRs to review those changes more easily. For now the PRs are in my own fork to avoid generating too much noise in the list of PRs as we evaluate whether the format is appropriate:
https://github.com/wetneb/OpenRefine/pulls
Feel free to comment here, there or on the forum to let me know if you like the approach:
https://forum.openrefine.org/t/releasing-4-0-alpha2/833/20

PR #5952, opened 2023-06-30
Update package prefix to org.openrefine from com.google.refine

tfmorris:
This is an attempt to establish a baseline against which architectural prototype can be diff'd more easily. Since it seems likely that any significant breaking change should include a move to a new package naming, we use that as a starting point. 

Wholesale import statement updates represent a good opportunity to change our import ordering if we desire., but this branch keeps it mostly the same as our existing order, which is based on Apache. Static imports, `java.*`, `javax.*`, all non-project imports, all project imports (`org.openrefine.*`). No wildcard imports used. Two changes from our previous setup seem to be: 1) removed the blank line between `java.*` and `javax.*` and 2) `org.*` used to be in a separate section before `com.*`, where as now they collate lexically with `org.` after `com.` (and `edu.`). It's easy enough to restore those settings, but they seem overly fussy and the changes don't seem to introduce too much noise to the diff.

- Update package prefix to `org.openrefine` from `com.google.refine.*`
- Add fallback class mapping from `com.google.refine.*` to `org.openrefine.*`
- Remove legacy mappings for `com.metaweb` and `com.google.gridworks`
- Organize imports as described above


tfmorris:
This doesn't correctly load existing projects yet (because there's a class name baked into the persisted JSON data).

wetneb:
> This is an attempt to establish a baseline against which architectural prototype can be diff'd more easily.

I have rebased my own branch so that the existing commits can be used as base instead, [as explained on the forum](https://forum.openrefine.org/t/releasing-4-0-alpha2/833/20).

So [this PR](https://github.com/wetneb/OpenRefine/pull/335) (currently in my own fork as it is not meant to be merged as such in master) can probably serve the purpose of this one? If that works for you I'll let you close this one.

tfmorris:
@wetneb Thanks for doing that. I've transferred some of the findings from this experiment to comments on that PR, so that they don't get lost if/when this gets closed.

I think the two PRs provide slightly different views since this is built on the current master and the other PR is built on some other branch/point in time. I'll definitely have a closer look at that PR though to see if I can figure out how you solved some of the problems which are still present in this PR.

Issue #5335, opened 2022-10-13
Add extension point to define new tabs in the main area


wetneb:
The Wikidata extension currently hacks its way into the core UI to define tabs for the Schema editor, Issues and Preview.
This is not a healthy practice: by having this extension do this, we are encouraging other extensions to do the same, and this is a problem for multiple reasons:
* any change in the DOM of the grid view can break those extensions
* two extensions which both do that are likely to be incompatible, since the tabs they define will not be handled by the same code.
* correctly handling resize events has led us to add a reference to `SchemaAlignment` in the main code, which should be avoided at all costs because this is a solution that other extensions (not shipped with OpenRefine) cannot rely on (#5334).

### Proposed solution

Introduce an extension point, which can be used to register new tabs for this panel. Tabs are instances of ES6 classes which have various methods:
* `renderBody`, to render the body of the tab in a `<div>` supplied as argument
* `renderHeader`, to render the title similarly
* `resize`, to react to a resize event
* probably other methods that will turn out to be needed?

### Alternatives considered

Originally the Wikidata extension did not define tabs, but used dialogs instead. That was much less user friendly, but in architectural terms it was much cleaner since it relied on the existing dialog system.

### Additional context

Prompted by #5334.

Abbe98:
Given this, would we like the tab row / header to always look like a tab-row? It's a little odd right now when it changes between the both modes. We could make it look just like the sidebars tabs but with the default of just having one?

![Screenshot 2023-05-20 at 21-08-43 fornvannen articles csv - OpenRefine](https://github.com/OpenRefine/OpenRefine/assets/2631719/fdb8d367-82c9-4f38-be8a-7726065ec88e)


![Screenshot 2023-05-20 at 21-09-26 fornvannen articles csv - OpenRefine](https://github.com/OpenRefine/OpenRefine/assets/2631719/9ae0de80-0181-4c54-bf2f-74a207bc686e)

thadguidry:
@Abbe98 What are sidebars tabs?  Oh do you mean the Facet panel tabs?

Abbe98:
> @Abbe98 What are sidebars tabs? Oh do you mean the Facet panel tabs?

Yes exactly.

Abbe98:
Pinging @OpenRefine/designers and @wetneb

wetneb:
I would find it natural to have the grid wrapped in a tab by default, indeed.

Abbe98:
"Mockups"

[...]

PR #5953, opened 2023-06-30
wikibase: remove the need for SchemaAlignment.resize()

Abbe98:
Part of #5335

This moves some layout logic from JavaScript to CSS eliminating the need for schemaAlignment.resize() which was being called from core.

See also #5334.


elebitzero (approved):
Looks good, I tried it out and it works.

PR #5960, opened 2023-07-04
Extract view/link URL for reconciled cells #1457

### LINKED ISSUE #1457: Extract view/link URL for reconciled cells
In a reconciled cell, the content of the cell is enhanced with a link to the reconciled item in the relevant service/site which has been used for the reconciliation.

From what I can see, the URL for the link is built on the fly by using the cell.recon.service to identify the service being used, to look up the 'view' URL from the service definition, and then building the link by combining the view URL template with the cell.recon.match.id

This makes it impossible (I think?) to extract the URL directly - but this is a feature I've seen asked for more than once.

Is there any reason that the view.url isn't stored directly in the reconciled cell? (either storing the template in cell.recon, or the view/link url in cell.recon.match?

If there is not a strong reason to not store it, I'm inclined to look at adding it in to make it more convenient to extract the information in GREL etc.

@wetneb ?

ayushrai206:
Fixes #1457 

- This basically adds another option of adding a column with url of matched entities in the reconcile->actions menu.


main/webapp/modules/core/langs/translation-en.json
```diff
@@ -541,6 +541,8 @@
     "core-views/judg-hist": "Judgment action timestamp",
     "core-views/hist-entries": "$1: judgment action timestamp",
     "core-views/actions": "Actions",
+    "core-views/Add column with URLs of matched entities":"Add column with URLs of matched entities",
+    "core-views/Add column with URLs of matched entities2":"Add column with URLs of matched entities for all current filtered rows",
```
wetneb, line 545: The keys of the messages could be improved:
* as you can see in the neighboring messages, we have a convention to write the keys in lowercase and with dashes. I would follow this convention for the sake of uniformity (although the convention itself is debatable!)
* rather than adding `2` as suffix to differentiate the message for the tooltip, you could use something more explicit, like `-tooltip`

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -55,6 +55,32 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
+        baseColumnName: column.name,  
+        newColumnName: "Url of Matched entities", 
+        columnInsertIndex: columnIndex + 1,
+        onError: "set-to-blank"
+      },
+      { expression: '"' + service.view.url + '".replace("{{id}}", cell.recon.match.id)' },
+      { modelsChanged: true },
+      {
+        onDone: function(o) {
+          dismiss();
```
wetneb, line 79: I am not sure where this `dismiss` is defined - could it be that it is left over from some other place in the code where you took some inspiration from? I suspect it is not needed here (see the other calls to `Refine.postCoreProcess` above, which do not have any `onDone` callback).

wetneb (commented):
Really great work, congratulations!

We will likely need to improve the GREL expression a little so that it works in all cases.
For Wikidata, the identifiers of entities are of the form `Q3498`, so they can be directly inserted in the template and that gives a valid URL.

But some reconciliation services might use identifiers which include some characters which need to be escaped to make sure we get a valid URL.
The specification has [a note](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#dfn-view-template) about this:

> To guarantee the correctness of the formed URI, clients MUST percent-encode component-delimiting reserved characters in the identifier, i.e. encode it as a URI component [[RFC3986](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#bib-rfc3986)].

So we need to find a GREL function which can be used to do this encoding and then insert it at the right place in the expression to get the desired result.

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
ayushrai206: Oh yes, you are right, I took it from somewhere and forgot to remove it! doing that asap.

main/webapp/modules/core/langs/translation-en.json
ayushrai206: done!

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -76,6 +78,9 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
       { modelsChanged: true },
       
     );
+    }
+    else
+    alert("The reconciliation service does not associate URLs to the entities it contains.");
```
wetneb, line 83: That could be localized as well, via `$.i18n`

wetneb:
Your last commit 55010a6 is likely unintentional, no? The `menu-reconcile.js` file should not be deleted.

ayushrai206:
> Your last commit [55010a6](https://github.com/OpenRefine/OpenRefine/commit/55010a6d7dd593fce802919567c37648cd0c04c1) is likely unintentional, no? The `menu-reconcile.js` file should not be deleted.

yes that was by mistake!

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
ayushrai206, line 1: I have not checked this so not sure about this!

wetneb: It looks really good to me! If installing `reconcile-csv` is proving too complicated let me know and I can try to find other ways to test it.

wetneb:
Congratulations @ayushrai206, it's looking great. However it looks like you have added some files by mistake (see the "Files changed" tab of this PR).

main/tests/cypress/cypress/e2e/project/grid/column/reconcile/actions/Add column_with_URLs_of_matched_entities.cy.js
```diff
@@ -0,0 +1,25 @@
...
+    cy.getCell(0, 'Item').should('to.contain', 'Choose new match');
+    cy.columnActionClick('Item', [
+    'Reconcile',
+    'Actions',
+    'Add column with URLs of matched entities',
+    ]);
+    cy.assertNotificationContainingText(
+        'Create new column Url of Matched entities based on column Item by filling 4 rows with if(cell.recon.match!=null,"https://www.wikidata.org/wiki/{{id}}".replace("{{id}}",escape(cell.recon.match.id,"url")),"")',
+    );
+    cy.getCell(0, 'Url of Matched entities').should('to.contain', 'https://www.wikidata.org/wiki/Q3938');
```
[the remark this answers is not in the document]
wetneb: I think it would make sense to add other assertions in the current test for the empty cells in other rows: that does not cost us much.
To test the scenario of service not having a view URL, it seems more complicated as we are currently not isolating reconciliation services (our tests rely on the live endpoints). So we would need to have a public endpoint which can be trusted not to have a view template in the future, or re-introduce a local service that is run alongside the tests (which @fgiroud moved out of because the tests were not passing reliably if I remember correctly).
Overall, given the high cost of Cypress tests, I'd be inclined not to add too much complexity there: we should not use them as unit tests. In the future it'd be nice to have some sort of a system for frontend unit tests, if we were to migrate to a framework which makes that possible.

main/tests/cypress/cypress/e2e/project/grid/column/reconcile/actions/Add column_with_URLs_of_matched_entities.cy.js
```diff
@@ -20,6 +20,9 @@ describe('Add Column With Url Of Matched Entities', () => {
         'Create new column Url of Matched entities based on column Item by filling 4 rows with if(cell.recon.match!=null,"https://www.wikidata.org/wiki/{{id}}".replace("{{id}}",escape(cell.recon.match.id,"url")),"")',
     );
     cy.getCell(0, 'Url of Matched entities').should('to.contain', 'https://www.wikidata.org/wiki/Q3938');
+    cy.getCell(1, 'Url of Matched entities').should('to.contain', '');
+    cy.getCell(2, 'Url of Matched entities').should('to.contain', '');
+    cy.getCell(3, 'Url of Matched entities').should('to.contain', '');
```
wetneb, line 25: I suspect that asserting that a dom element contains `""` is always true (because `""` is a substring of all strings), so there might be a better way to do these assertions. You could look at the `cy.assertCellEquals` method for instance.

main/tests/cypress/cypress/e2e/project/grid/column/reconcile/actions/Add column_with_URLs_of_matched_entities.cy.js
```diff
@@ -0,0 +1,28 @@
+describe('Add Column With Url Of Matched Entities', () => {
```
wetneb, line 1:
```suggestion
describe('Add column with URLs of matched entities', () => {
```
(and similarly in the resrt of the file)

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -55,6 +55,33 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
     );
   };
 
+  var doAddColumnWithUrlOfMatchedEntities=function(){
```
wetneb, line 58:
```suggestion
  var doAddColumnWithUrlOfMatchedEntities = function() {
```

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -55,6 +55,33 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
+    var serviceUrl = null;
+    var service = null;
+    if (column.reconConfig) {
+        serviceUrl = column.reconConfig.service;
+    }
+    if (serviceUrl) {
+        service = ReconciliationManager.getServiceFromUrl(serviceUrl);
+    } 
+    if(service&&service.view)
```
wetneb, line 67:
```suggestion
    if (service && service.view) {
```

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -55,6 +55,33 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
+    {
+    Refine.postCoreProcess(
```
wetneb, line 69: Here the indentation is off

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -55,6 +55,33 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
+        columnInsertIndex: columnIndex + 1,
+        onError: "set-to-blank"
+      },
+      { expression: 'if(cell.recon.match!=null,"' + service.view.url + '".replace("{{id}}",escape(cell.recon.match.id,"url")),"")' },
+      { modelsChanged: true },
+      
+    );
+    }
+    else
+    alert($.i18n('core-views/service-does-not-associate-URLs-to-the-entities-it-contains'));
```
wetneb, line 83: We generally enclose all `if` or `else` blocks with braces even if it's not needed, to avoid accidents, so I would do it for this `else` block too.

wetneb (commented):
We can now do some style polishing, see the comments below.
Also, the file `Add column_with_URLs_of_matched_entities.cy.js` could be renamed to `add_column_with_URLs_of_matched_entities.cy.js` (to avoid having a space in its filename, and to follow the naming convention of other files in that directory).

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -55,6 +55,35 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
+    var serviceUrl = null;
+    var service = null;
+    if (column.reconConfig) {
+        serviceUrl = column.reconConfig.service;
+    }
+    if (serviceUrl) {
+        service = ReconciliationManager.getServiceFromUrl(serviceUrl);
+    } 
+    if (service && service.view)
+    {
```
wetneb, line 68: Here you can also integrate the brace in the same line as the if statement:
`if (service && service.view) {`

wetneb (commented):
Almost there! Just a couple of nitpicks about code style.

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -64,25 +64,21 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
-    if (service && service.view)
-    {
+    if (service && service.view){
      Refine.postCoreProcess
      (
       "add-column", 
       {
         baseColumnName: column.name,  
-        newColumnName: "Url of Matched entities", 
+        newColumnName: $.i18n('core-views/entity URL'),
```
wetneb, line 73:
```suggestion
        newColumnName: $.i18n('core-views/entity-URL'), 
```

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -778,6 +778,7 @@
     "core-views/use-values-as-identifiers/header": "Use values as identifiers",
     "core-views/use-values-as-identifiers2": "Mark cells as reconciled with their values as identifiers",
     "core-views/choose-reconciliation-service": "Choose a reconciliation service",
+    "core-views/entity URL":"Entity URL",
```
wetneb, line 781: For consistency with other message keys:
```suggestion
    "core-views/entity-URL":"Entity URL",
```

wetneb (commented):
I tried it out as a user and found some more things (sorry!):

The menu item could be moved next to the action which adds a column with entity identifiers, because they seem closely related. The "Actions" submenu seems to be more about changing the reconciliation judgment:
image

That same action prompts the user for a column name, so I now wonder if this new action you are introducing should do the same - I'll ask on the issue.

ayushrai206:
> I tried it out as a user and found some more things (sorry!):
> 
> The menu item could be moved next to the action which adds a column with entity identifiers, because they seem closely related. The "Actions" submenu seems to be more about changing the reconciliation judgment: image
> 
> That same action prompts the user for a column name, so I now wonder if this new action you are introducing should do the same - I'll ask on the issue.

Moving the item next to the action which adds a column with entity identifiers makes total sense!

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -598,6 +592,13 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
       click: doUseValuesAsIdentifiers
     },
     {},
+    {
+      id: "core/Add-column-with-URLs-of-matched-entities",
+      label: $.i18n("core-views/add-column-with-URLs-of-matched-entities"),
+      tooltip: $.i18n("core-views/add-column-with-URLs-of-matched-entities-tooltip"),
+      click:  doAddColumnWithUrlOfMatchedEntities
+    },
+    {},
```
wetneb, line 601: I think this separator is not needed, given that the two actions (for identifiers and URLs) are very similar: they can be in the same section of the menu.

wetneb:
I would wait for #6021 to be merged first. Then we can update this PR so that it prompts the user for a column name in the same way as the operation to add identifiers. Potentially this can be done by extracting a generic utility to do this prompting (to avoid duplicating too much code).

ayushrai206:
Hey everyone,
I have added a prompt that asks the user for a new Column name, though there are some problems in my code which are- before the prompt dialog appears the previous menu and submenu does not disappear and the cypress tests need updation. I am just putting forward the changes till now for reviews while I work on resolving this error.
Thank you
Ayushi

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -55,6 +55,33 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
     );
   };
 
+  var doAddColumnWithUrlOfMatchedEntities = function AddColumnWithUrlOfMatchedEntities (columnName,dismissDialog) {
```
wetneb, line 58: Here you can simplify the way you declare this function:
```suggestion
  var doAddColumnWithUrlOfMatchedEntities = function (columnName,dismissDialog) {
```

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -55,6 +55,33 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
+    promptForColumn(AddColumnWithUrlOfMatchedEntities);
```
wetneb, line 59: This will not work, because the callback you are supplying to the `promptForColumn` function cannot be the same as the outer function that calls `promptForColumn`. If you do that, you will run into the following problem (assuming that `promptForColumn` works properly):
* when the user first clicks on the menu entry to run the operation, `AddColumnWithUrlOfMatchedEntities` is called
* inside the body of `AddColumnWithUrlOfMatchedEntities`, `promptForColumn` is called
* once the user has supplied a column name, the success callback supplied to `promptForColumn` is called, which is `AddColumnWithUrlOfMatchedEntities` again
* inside the body of `AddColumnWithUrlOfMatchedEntities`, `promptForColumn` is called
* once the user has supplied a column name, the success callback supplied to `promptForColumn` is called, which is `AddColumnWithUrlOfMatchedEntities` again
* inside the body of `AddColumnWithUrlOfMatchedEntities`, `promptForColumn` is called
* once the user has supplied a column name, the success callback supplied to `promptForColumn` is called, which is `AddColumnWithUrlOfMatchedEntities` again
* inside the body of `AddColumnWithUrlOfMatchedEntities`, `promptForColumn` is called
* once the user has supplied a column name, the success callback supplied to `promptForColumn` is called, which is `AddColumnWithUrlOfMatchedEntities` again
* ...

so this is an infinite loop!

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -55,6 +55,33 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
+    var serviceUrl = null;
+    var service = null;
+    if (column.reconConfig) {
+        serviceUrl = column.reconConfig.service;
+    }
+    if (serviceUrl) {
+        service = ReconciliationManager.getServiceFromUrl(serviceUrl);
+    } 
+    if (service && service.view){
+     Refine.postCoreProcess
```
wetneb, line 69: This code should only be executed once the user has supplied the column name, which means after the success callback of the `promptForColumn` function is called.

wetneb (commented):
Good start! It looks like there are some issues with the overall structure of your code though.

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
ayushrai206: I have tried to resolve these in my new commits, also I am sorry for the mess due to forced pushes.

ayushrai206:
Hey everyone,
 In my most recent commit I have tried to resolve the problem of infinite loop and I have also commented out the cypress tests for a while since they were giving errors, will update the cypress tests when the code is approved. Also I was unsuccessful in removing force pushes from Github history so please do not compare the code in the force pushes, rather see my most recent commit for the actual changes.
Thank you
Ayushi

wetneb (commented):
Congratulations, it seems to work great!

It is a little difficult to review your code because you did not add any indentation. Could you indent it like the surrounding code, by using two spaces as indentation unit?

Also, the purpose of introducing this `promptForColumn` was to avoid the duplication of code with the "Add entity identifiers column" operation. So we should make sure that operation also switches to using this new function you introduced. You might need to add another parameter to make it possible to set the title of the dialog (since the current dialog title mentions URLs, which will not be appropriate for the other operation).

wetneb (commented):
Thanks, that's much better! It seems that by doing this refactoring you have changed the title of the dialog of the existing operation. Restoring the original title should help to get the Cypress test pass again. I would do that by adding a new argument to the `promptForColumn` function, which would be the title to display in the dialog. This way, you would still be able to use the same `promptForColumn` function in both operations, but display different titles to the user. This would let you keep the original dialog title for the existing operation and use an appropriate title (mentioning URLs) for the new operation.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -577,7 +579,8 @@
     "core-views/word-facet": "Word facet",
     "core-views/add-id-column": "Add entity identifiers column…",
     "core-views/add-id-column2": "Add a column containing entity identifiers from items",
-    "core-views/add-id-col": "Add column containing entity identifiers on \"$1\"",
+    "core-views/add-id-col": "Add column containing entity identifiers on $1",
```
wetneb, line 582: Here you have removed the double quotes around the column name, but I would keep them (to separate the user-provided text from the system text more cleanly). This should help get the Cypress test pass again.

wetneb (commented):
Great, the Cypress tests pass! Now perhaps you could re-enable the Cypress test for the new operation and we should be good to go :) I think the indentation can be improved still but I'm happy to do that at the end.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -578,6 +580,7 @@
     "core-views/add-id-column": "Add entity identifiers column…",
     "core-views/add-id-column2": "Add a column containing entity identifiers from items",
     "core-views/add-id-col": "Add column containing entity identifiers on \"$1\"",
+    "core-views/add-entity-URL-col":"Add a name for the column containing URL of matched entities",
```
wetneb, line 583: Maybe this text could be simply the same as the action name in the menu entry, "Add column with URLs of matched entities"?
```suggestion
    "core-views/add-entity-URL-col":"Add column with URLs of matched entities",
```

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -548,6 +548,8 @@
     "core-views/judg-hist": "Judgment action timestamp",
     "core-views/hist-entries": "$1: judgment action timestamp",
     "core-views/actions": "Actions",
+    "core-views/add-column-with-URLs-of-matched-entities":"Add column with URLs of matched entities",
```
wetneb, line 551: To indicate that there is a dialog showing up after one clicks this entry, we have the convention of adding "…" in English at the end.
```suggestion
    "core-views/add-column-with-URLs-of-matched-entities":"Add column with URLs of matched entities…",
```

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -55,6 +55,75 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
...
+      if (serviceUrl) {
+        service = ReconciliationManager.getServiceFromUrl(serviceUrl);
+      }
+      if (service && service.view){
+        Refine.postCoreProcess
+        (
+        "add-column",
+        {
+        baseColumnName: column.name,
+        newColumnName: columnName,
```
wetneb, line 74: The indentation can still be improved here. You can look for other calls to `Refine.postCoreProcess` for inspiration :)

wetneb (approved):

PR #5964, opened 2023-07-06
Fix: Reconciliation services can be added multiple times

### LINKED ISSUE #5926: Reconciliation services can be added multiple times
When adding a new reconciliation service in the reconciliation dialog, we do not check if the service being added is already in the list.

### To Reproduce
1. First, add a reconciliation service
2. Then, add it again

### Current Results

The service is added twice.

### Expected Behavior

The service list should not have any duplicates.

### Additional context

Reported by @fsteeg. Maybe an interesting issue for @ayushrai206?

tsukipedia:
Fixes #5926 

- Add validation to check if service URL exists in the url map of the reconciliation manager


main/webapp/modules/core/scripts/reconciliation/recon-manager.js
```diff
@@ -63,6 +63,14 @@ ReconciliationManager.registerService = function(service) {
 };
 
 ReconciliationManager.registerStandardService = function(url, f, silent) {
+
+  if (ReconciliationManager._urlMap.hasOwnProperty(url)) {
+    if (!silent) {
+      alert($.i18n('core-recon/url-already-registered'));
+    }
+    return;
```
wetneb, line 71: Instead of just returning we might need to call the callback `f` if it is defined (see further down in the function).

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -443,6 +443,7 @@
     "core-recon/cell-contains": "Each cell contains:",
     "core-recon/max-candidates": "Maximum number of candidates to return",
     "core-recon/service-title": "Services",
+    "core-recon/url-already-registered":"Service URL already exists.",
```
wetneb, line 446: Maybe the error message could be a bit more user-centric, such as:
```suggestion
    "core-recon/url-already-registered":"This reconciliation service already exists.",
```

wetneb (changes requested):
Hi @tsukipedia, thank you so much for this PR!

It's kind of you to have provided the translations of the message you introduced, but we try to make sure those messages are only edited by people who are confident in those languages. Automated translations are not always of great quality. So in a PR like this I would expect that you only insert the English translation: the other translations will be added by translators via Weblate. If your native language is not English it's fine if you also add a translation for that language in the same go.

main/webapp/modules/core/scripts/reconciliation/recon-manager.js
tsukipedia: Thanks for the quick response, I made the requested modifications. Hope I understood correctly uwu

main/webapp/modules/core/scripts/reconciliation/recon-manager.js
```diff
@@ -63,6 +63,14 @@ ReconciliationManager.registerService = function(service) {
 };
 
 ReconciliationManager.registerStandardService = function(url, f, silent) {
+
+  if (ReconciliationManager._urlMap.hasOwnProperty(url)) {
+    if (!silent) {
+      alert($.i18n('core-recon/url-already-registered'));
+    }
+    if (f) { f(url); }
```
wetneb, line 71: We probably still need your return statement here, so that the rest of the function is not executed.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -443,6 +443,7 @@
     "core-recon/cell-contains": "Each cell contains:",
     "core-recon/max-candidates": "Maximum number of candidates to return",
     "core-recon/service-title": "Services",
+    "core-recon/url-already-registered":"This reconciliation service already exists.",
```
tfmorris, line 446: Apologies for the conflicting reviews, but I think the Spanish version has better semantics. I'd suggest

```suggestion
    "core-recon/url-already-registered":"This reconciliation service is already registered.",
```

tfmorris (changes requested):

main/webapp/modules/core/scripts/reconciliation/recon-manager.js
wetneb: Yes indeed, we still need the return statement in addition to calling the callback (see below)

tfmorris (approved):

wetneb (approved):
Thanks a lot!

PR #5966, opened 2023-07-07
Refresh SECURITY.md

wetneb:
- replace mention of broken mailing list by the advisory committee. We can also consider establishing other groups or email aliases in the future - the priority for me is to have something that actually works
- update supported versions (we probably should add it as a step in the release instructions to keep this up to date)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -11,7 +11,7 @@
 
 Our core team will try their best to fix any valid vulnerability that is reported to them.
 
-You can privately report a vulnerability to us by sending a report by [creating a security advisory on GitHub](https://github.com/OpenRefine/OpenRefine/security/advisories/new). This report will be kept private while it is being assessed by the team.
+You can privately report a vulnerability to the OpenRefine team by [creating a security advisory on GitHub](https://github.com/OpenRefine/OpenRefine/security/advisories/new). This report will be kept private while it is being assessed by the team.
 
-Keep in mind that OpenRefine is designed to run locally on a users PC, while also making network calls across the internet only upon a users choice or command.
+Keep in mind that OpenRefine is designed to run locally on a user's PC, while also making network calls across the internet only upon a user's choice or command.
 As such, certain vulnerabilities might not apply to OpenRefine's design. In doubt, please submit a report anyway.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,12 +4,12 @@
 
 | Version | Supported          |
 | ------- | ------------------ |
-| 3.5.x   | :white_check_mark: |
-| <= 3.4   | :x:                |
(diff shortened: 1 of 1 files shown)


wetneb:
We could also consider enabling a new GitHub feature to let people submit security advisories via the platform:
https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability

tfmorris:
The mailing list looks operational to me: https://groups.google.com/g/openrefine-coredev/c/ZDw76iRPI8Q

I don't think replacing a technical group with an administrative group is the right equivalence in this context.

The Github reporting mechanism sounds useful because it integrates well with the rest of the bug lifecycle.

wetneb:
If we are able to configure the mailing list so that non-subscribers are able to post to it, I am fine with keeping it.

tfmorris:
I should have replied in importance order, rather than chronologically.

I think that using Github for security notifications has advantages over using an email address (whatever that address might be).

wetneb:
Let's try it out then! I have enabled it.

SECURITY.md
```diff
@@ -4,15 +4,14 @@
...
+| <= 3.6   | :x:                |
 
 ## Reporting a Vulnerability
 
-You can privately report a vulnerability to us by sending a report to this private mailing list [mailto:openrefine-coredev@googlegroups.com](mailto:openrefine-coredev@googlegroups.com)
-
 Our core team will try their best to fix any valid vulnerability that is reported to them.
 
-Keep in mind that OpenRefine is designed to run locally on a users PC, while also making network calls across the internet only upon a users choice or command.
+You can privately report a vulnerability to us by sending a report by [creating a security advisory on GitHub](https://github.com/OpenRefine/OpenRefine/security/advisories/new). This report will be kept private while it is being assessed by the team.
```
tfmorris, line 14:
```suggestion
You can privately report a vulnerability to the OpenRefine team by [creating a security advisory on GitHub](https://github.com/OpenRefine/OpenRefine/security/advisories/new). This report will be kept private while it is being assessed by the team.
```

SECURITY.md
```diff
@@ -4,15 +4,14 @@
...
 
-As such, certain vulnerabilities might not apply to OpenRefine's design.
+Keep in mind that OpenRefine is designed to run locally on a users PC, while also making network calls across the internet only upon a users choice or command.
```
tfmorris, line 16:
```suggestion
Keep in mind that OpenRefine is designed to run locally on a user's PC, while also making network calls across the internet only upon a user's choice or command.
```

tfmorris (approved):
Just a few minor copy edits

PR #5973, opened 2023-07-14
fix custom metadata edit action

### LINKED ISSUE #5772: Error when editing CustomMetadata field
When trying to edit the customMetadata field in the about section of a project to add new values, the OpenRefine server crashes and throws a ``` java.lang.IllegalArgumentException: Can not set java.util.Map field com.google.refine.ProjectMetadata._customMetadata to java.lang.String```. The issue is the same as well when interacting with the OpenRefine API through ```set-project-metadata``` endpoint as well.

### To Reproduce
1. First, open the Open Project section to the left of the screen
2. Then, choose any project (with empty customMetadata) and press on about.
3. Finally, next to the customMetada field, press the edit button and add any JSON-like format.

### Current Results
The newly added values in the customMetadata field is not saved when the page is reloaded and the server crashes (error shown below).  

### Expected Behavior
The newly added values in the customMetadata field should be saved when the page is reloaded.

### Screenshots
image

### Versions
 - Browser Version: Chrome Version 111.0.5563.147
 - JRE or JDK Version: openjdk 1.8.0_345 
 - OpenRefine: 3.7

tsukipedia:
Fixes #5772 

- edit setAnyField() method in ProjectMetadata class
- add case to edit customMetadata field
- parse given String to JSON and JSON to HashMap

### DIFF
--- a/main/src/com/google/refine/ProjectMetadata.java
+++ b/main/src/com/google/refine/ProjectMetadata.java
@@ -53,8 +53,6 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.json.JsonReadFeature;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonView;
-import com.fasterxml.jackson.databind.JsonMappingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.refine.preference.PreferenceStore;
@@ -372,21 +370,23 @@ public class ProjectMetadata {
         }
     }
 
-    public void setAnyField(String metaName, String valueString) throws JsonMappingException, JsonProcessingException {
+    public void setAnyField(String metaName, String valueString) {
         Class<? extends ProjectMetadata> metaClass = this.getClass();
         try {
             Field metaField = metaClass.getDeclaredField("_" + metaName);
             if (metaName.equals("tags")) {
                 metaField.set(this, valueString.split(","));
             } else if (metaName.equals("customMetadata")) {
-                ObjectMapper customMetadataMapper = new ObjectMapper();
-                customMetadataMapper.enable(JsonReadFeature.ALLOW_UNQUOTED_FIELD_NAMES.mappedFeature());
-                customMetadataMapper.enable(JsonReadFeature.ALLOW_BACKSLASH_ESCAPING_ANY_CHARACTER.mappedFeature());
-                Map<String, Object> mapping = customMetadataMapper.readValue(valueString, HashMap.class);
-                metaField.set(this, mapping);
+                ParsingUtilities.mapper.enable(JsonReadFeature.ALLOW_BACKSLASH_ESCAPING_ANY_CHARACTER.mappedFeature());
+                Map<String, Object> map = ParsingUtilities.mapper.readValue(valueString, HashMap.class);
+                metaField.set(this, map);
             } else {
                 metaField.set(this, valueString);
             }
+        } catch (JsonProcessingException e) {
+            String errorMessage = "Error reading JSON: " + e.getOriginalMessage();
+            logger.error(errorMessage);
+            throw new RuntimeException(errorMessage);
         } catch (NoSuchFieldException e) {
             updateUserMetadata(metaName, valueString);
         } catch (SecurityException | IllegalArgumentException | IllegalAccessException e) {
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -357,6 +357,11 @@
     <!-- dependencies are stored in the relevant submodules:
           see main/pom.xml, server/pom.xml and each extension.
     -->
+      <dependency>
+        <groupId>org.json</groupId>
```
wetneb, line 361: We cannot add this dependency to the project because it has an incompatible license. Jackson (which we already depend on) should have all the necessary features for our JSON processing needs. Also, it does not look like you are using this new dependency in your changes at all.

main/src/com/google/refine/ProjectMetadata.java
```diff
@@ -368,12 +372,18 @@ private void updateUserMetadata(String metaName, String valueString) {
...
 
-    public void setAnyField(String metaName, String valueString) {
+    public void setAnyField(String metaName, String valueString) throws JsonMappingException, JsonProcessingException {
         Class<? extends ProjectMetadata> metaClass = this.getClass();
         try {
             Field metaField = metaClass.getDeclaredField("_" + metaName);
             if (metaName.equals("tags")) {
                 metaField.set(this, valueString.split(","));
+            } else if (metaName.equals("customMetadata")) {
+                ObjectMapper customMetadataMapper = new ObjectMapper();
```
wetneb, line 382: Instead of defining a new `ObjectMapper` (which is costly), it could be simpler to reuse `ParsingUtilities.mapper`.

wetneb (commented):
Thanks for having a look at this! It could be nice to also have a unit test for this fix.

pom.xml
tsukipedia: You're right, I will remove it. I was sloppy leaving it there :c

tfmorris:
You can use `mvn formatter:format` to resolve the linting issue.

main/tests/server/src/com/google/refine/commands/project/SetProjectMetadataCommandTests.java
```diff
@@ -211,4 +213,51 @@ public void doPostThrowsIfCommand_getProjectReturnsNull() {
...
+    private String getInvalidCustomMetadata() {
+        return "{\n" + //
+                "  \"name\":test,\n" + //
+                "  \"companies\":[\"Google\", \"IBM\", \"meta\"],\n" + //
+                "  \"isValid\": false\n" + //
+                "  \"information\": {\n" + //
+                "    \"text-field\":\"value\",\n" + //
+                "    \"numeric-field\": 1\n" + //
+                "  },\n" + //
+                "  \"id\":2,\n" + //
```
tfmorris, line 260:
```suggestion
```

The extra stuff here isn't really needed and just adds noise.

main/tests/server/src/com/google/refine/commands/project/SetProjectMetadataCommandTests.java
```diff
@@ -211,4 +213,51 @@ public void doPostThrowsIfCommand_getProjectReturnsNull() {
...
+    public void setCustomMetadataTest() {
+        String customMetadata = this.getValidCustomMetadata();
+        ProjectMetadata meta = proj.getMetadata();
+        meta.setAnyField("customMetadata", customMetadata);
+
+        Assert.assertEquals(meta.getCustomMetadata("name"), "test");
+        Assert.assertEquals(meta.getCustomMetadata("isValid"), true);
+        Assert.assertEquals(meta.getCustomMetadata("id"), 1);
+        Assert.assertNotNull(meta.getCustomMetadata("companies"));
+        Assert.assertNotNull(meta.getCustomMetadata("information"));
```
tfmorris, line 227:
```suggestion
        Assert.assertTrue(meta.getCustomMetadata("companies") instanceof List);
        Assert.assertTrue(meta.getCustomMetadata("information") instanceof Map);
```
This will make the test a little bit stronger.

main/tests/server/src/com/google/refine/commands/project/SetProjectMetadataCommandTests.java
```diff
@@ -62,6 +62,8 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
 import com.google.refine.model.Project;
 import com.google.refine.util.ParsingUtilities;
 
+import net.arnx.jsonic.JSONException;
+
```
tfmorris, line 66:
```suggestion
```

Extraneous noise added by your IDE

tfmorris (changes requested):
Looking better! Just a couple of tiny things remaining.

main/tests/server/src/com/google/refine/commands/project/SetProjectMetadataCommandTests.java
tsukipedia: this is great!  thank you 🙏

tfmorris (approved):

tfmorris:
Looks great! Thanks for the prompt turnaround on the review feedback.

PR #5974, opened 2023-07-14
Added preference for auto cluster

elebitzero:
Follow-up to #5898

- Added auto cluster preference (default: false).
- When the Cluster & Edit dialog is opened when auto is false, a message will be shown: "Click Cluster to find clusters on column "X" using the parameters above. [Cluster]"
- After clustering, if the user changes any parameters and auto is false, the cluster results will be cleared, and the message will reappear.

image


wetneb:
I think it is a bit misleading that when the dialog is first opened, the main area shows a big "No clusters were found" message. At that point, that message is wrong: the clusters haven't been searched for yet.
I think there is a high chance that existing users who upgrade to this version will not notice the new cluster button and will just take it for granted that clustering has run and did not find any clusters.

This suggests a different UI: just move the cluster button to the center of the main area, replacing this "No clusters were found" message. That would make it clear that an interaction from the user is required to see the clusters.
I think this is a change that would make sense even for further clustering actions: whenever the clustering settings change, the existing clusters could be removed and replaced by the "Cluster" button.

Perhaps @OpenRefine/designers have thoughts about that?

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -193,7 +193,7 @@
     "core-dialogs/use-this-val": "Use this value",
     "core-dialogs/browse-only-these": "Browse only these values",
     "core-dialogs/browse-this-cluster": "Browse this cluster",
-    "core-dialogs/no-cluster-found": "No clusters were found with the selected method",
+    "core-dialogs/no-cluster-found": "Click Cluster to find clusters on column \"$1\" using the parameters above.",
```
wetneb: I think we need two different messages: one for the initial state of the dialog, where no clusters have been computed yet, and another one for the case where clusters have been computed, but none were found.

elebitzero: Yes. Sorry, I checked in a WIP. I have updated it again and it is now complete. Please review. Thanks.

wetneb (approved):
@OpenRefine/designers last chance to give an opinion on the design!

PR #5976, opened 2023-07-15
Add numbering to the list of Wikibase schemas

### LINKED ISSUE #5893: Manage Wikibase schemas should use a list (auto numbered) to improve display
Currently, the Manage Wikibase schemas dialog shows slight grey rows for each schema to be managed.
If one row only appears, then this looks awkward because the grey row with a schema name looks like some kind of header for the dialog, instead of a row element.  It would be better to tweak the display so that rows get numbered automatically.

### Current Results
image

### Expected Behavior
Rows are numbered as in this example (number just before the schema row for `test`).
If a 2nd schema was defined then I would expect numbered rows 1. and 2.

image

### Versions
 - Operating System: 
 - Browser Version: 
 - JRE or JDK Version: 
 - OpenRefine: 

### Additional context
From Forum support topic https://forum.openrefine.org/t/where-is-export-import-schema-option-in-or-wikidata-extension-now/784

Hisiste:
Fixes #5893

- JavaScript code was edited to append a number to the schema name.
- The change should be purely visual. The schema name variables are not edited.


Hisiste:
The change is not exactly what was shown in the screenshot of the issue. But it is solved in the simplest way that I could think of. I hope this is ok. :)

image

wetneb (commented):
Thanks a lot for the PR and welcome to the project!
I wonder if you considered using numbered list items instead? So that the number would be added by the browser directly. I suspect that might be more accessible and better localized.
https://developer.mozilla.org/fr/docs/Web/HTML/Element/ol

Hisiste:
Hi @wetneb! And thank you very much for the warming welcome. :)

I did consider using numbered list items using the `<ol>` tag. The problem was that the items are already inside a `<ol>` tag[^1] and each item inside a `<li>` tag[^2], but the number does not show. The only hint that I could find to this was in this CSS file: https://github.com/OpenRefine/OpenRefine/blob/9042ff708e37291ec2072c837cc82ceea7c5fa60/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css#L1-L5

The thing is: no matter what `list-style` is, it never shows the list number. That's why I edited the name directly instead.

Ow. Can you give me a hint on how to do this using the `<ol>` tag? Please. :)

[^1]: https://github.com/OpenRefine/OpenRefine/blob/9042ff708e37291ec2072c837cc82ceea7c5fa60/extensions/wikibase/module/scripts/dialogs/schema-management-dialog.html#L4-L8
[^2]: https://github.com/OpenRefine/OpenRefine/blob/9042ff708e37291ec2072c837cc82ceea7c5fa60/extensions/wikibase/module/scripts/dialogs/schema-list-item.html

Hisiste:
UPDATE:
I found why the numbers weren't showing. It was because of this block of code: https://github.com/OpenRefine/OpenRefine/blob/9042ff708e37291ec2072c837cc82ceea7c5fa60/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css#L11-L15
We need to have `display: list-item` instead.[^1] The problem is that this display doesn't support the `justify-content` element. Everything will be aligned to the left, so I will be searching for a solution.

Lastly, for the numbers to show up, we can use `list-style-position: inside;`. Currently, the patch should look like this:
```diff
diff --git a/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css b/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css
index 01f4618e0..890e3fec2 100644
--- a/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css
+++ b/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css
@@ -1,5 +1,6 @@
 .schema-template-list {
-  list-style: none;
+  list-style: decimal;
+  list-style-position: inside;
   max-height: 450px;
   overflow-y: auto;
 }
@@ -10,8 +11,7 @@

 .schema-template-list li {
   padding: .5em;
-  display: flex;
-  justify-content: space-between;
+  display: list-item;
 }
```
And the result is this:
image
(The changes I made to the `schema-management-dialog.js` file were reverted for the screenshot.)

The closest that I could get to the normal spacing was this:
```diff
diff --git a/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css b/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css
index 01f4618e0..890e3fec2 100644
--- a/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css
+++ b/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css
@@ -1,5 +1,6 @@
 .schema-template-list {
   list-style: decimal;
   list-style-position: inside;
   max-height: 450px;
   overflow-y: auto;
+  text-align-last: justify;
 }
```
But the result is this:
image
I will keep searching for the correct approach. If someone has an idea on how this can be done, I will be more than happy to listen. :D

[^1]: https://stackoverflow.com/a/35879429

Hisiste:
At the end I decided to just use `float: right;` on the `<div>` of `Rename | Export | Delete`. This gives us the following result:
image

I hope this solves the problem in a better way. :D

Abbe98:
Rather than styling the inner div using `list-style-*` can't the current list be converted to an `<ol>` element and (some of) the current `li` styling be moved to an inner element? Otherwise we still have a semantics issue.

Hisiste:
@Abbe98 Oh! It is already an `<ol>` element. https://github.com/OpenRefine/OpenRefine/blob/9042ff708e37291ec2072c837cc82ceea7c5fa60/extensions/wikibase/module/scripts/dialogs/schema-management-dialog.html#L5-L7

Because its class is `schema-template-list`, I edited the CSS file for this class only.
```diff
diff --git a/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css b/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css
index 01f4618e0..890e3fec2 100644
--- a/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css
+++ b/extensions/wikibase/module/styles/dialogs/schema-management-dialog.css
@@ -1,5 +1,6 @@
 .schema-template-list {
-  list-style: none;
+  list-style: decimal;
+  list-style-position: inside;
   max-height: 450px;
   overflow-y: auto;
 }
```

The only styling that I added to the `<div>` was that I could move the `Rename | Export | Delete` block to the right side.
```diff
diff --git a/extensions/wikibase/module/scripts/dialogs/schema-list-item.html b/extensions/wikibase/module/scripts/dialogs/schema-list-item.html
index 8ce783b1e..334f625b5 100644
--- a/extensions/wikibase/module/scripts/dialogs/schema-list-item.html
+++ b/extensions/wikibase/module/scripts/dialogs/schema-list-item.html
@@ -2,7 +2,7 @@
     <div class="wikibase-template-info">
         <span class="wikibase-name" bind="templateName"></span>
     </div>
-    <div class="wikibase-template-actions">
+    <div class="wikibase-template-actions" style="float: right;">
         <a bind="renameTemplate"></a> |
         <a bind="downloadTemplate"></a> |
         <a bind="deleteTemplate"></a>
```

wetneb (approved):
Thank you very much!

PR #5981, opened 2023-07-19
Expanding collapsed columns

### LINKED ISSUE #5879: Add a visual indicator/button for expanding collapsed columns
### Proposed solution

@wetneb [pointed out](https://github.com/OpenRefine/OpenRefine/pull/5869#issuecomment-1567269708) that there is no longer an indicator that one can click the header of a collapsed column to expand it. We should bring this back but rather than just have the mouse cursor change lets add a proper button(semantically not necessarily visually) so it works from the keyboard this time around.

### Alternatives considered

The cheap solution of making sure the cursor changes one again on hover when a column is collapsed.

### Additional context

I removed the cursor effect in #5869 and I think that's still a good change given the odd behavior of changing the cursor on non-clickable expanded columns as well.

VhugoJc:
Fixes #5879 

Screenshot 2023-07-19 at 12 22 14


VhugoJc:
I fixed the Issue mentioned by @wetneb in #5869 (:

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -371,12 +371,20 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
...
+    .addClass("column-header")
+    .attr('title', column.name);
     if (self._collapsedColumnNames.hasOwnProperty(column.name)) {
-      $(th).html("&nbsp;").on('click',function(evt) {
-        delete self._collapsedColumnNames[column.name];
-        self.render();
-      });
+      DOM.bind( 
+        $(th)
+        .html("<button class='column-header-menu column-header-menu-expand' bind='expandColumn' ></button>")
```
Abbe98, line 380: A title like "Expand this column" would be nice.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -371,12 +371,20 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
   this._columnHeaderUIs = [];
   var createColumnHeader = function(column, index) {
     var th = trHead.appendChild(document.createElement("th"));
-    $(th).addClass("column-header").attr('title', column.name);
+    $(th)
```
Abbe98, line 374: Unrelated change?

main/webapp/modules/core/styles/views/data-table-view.css
```diff
@@ -188,6 +188,10 @@ table.data-table td.column-header, table.data-table th.column-header {
   border: none;
 }
 
+.column-header-menu-expand{
```
Abbe98, line 191: Nitpick: missing space.

Abbe98 (dismissed):
Looks solid to me! ⭐

VhugoJc:
I changed the label to "Expand this column" and changed the requested changes. 
Screenshot 2023-07-24 at 15 58 18
I hope this is what you were looking for! Let me know if it's good or if you need anything else. Thanks!

wetneb (approved):

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -373,10 +373,16 @@ DataTableView.prototype._renderDataTables = function(table, tableHeader) {
     var th = trHead.appendChild(document.createElement("th"));
     $(th).addClass("column-header").attr('title', column.name);
     if (self._collapsedColumnNames.hasOwnProperty(column.name)) {
-      $(th).html("&nbsp;").on('click',function(evt) {
-        delete self._collapsedColumnNames[column.name];
-        self.render();
-      });
+      DOM.bind( 
+        $(th)
+        .attr('title',"Expand this column")
```
wetneb, line 378: Oops, here is a problem I oversaw: we'd need this text to be localized, with our `$.i18n` utility.

wetneb (changes requested):

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
VhugoJc: I have changed the text for `$.i18n` utility and I added the text in all the supported languages. Please, tell me if you need anything else 🚀

wetneb (commented):
Thank you so much for the swift reaction! It is also kind of you to have filled all translations, but we prefer to avoid relying on machine translation for this and we normally only add the English strings when making changes like this one. Translators will be duly notified that this new string is inserted and will have the opportunity to translate it on their own. Could you remove those new translations and only keep the changes to `translation-en.json`? I am really sorry you went through all this work - I should have explained our practices in the first place.

VhugoJc:
Don't worry, I perfectly understand!

wetneb (approved):
Thanks again!

wetneb:
Ah, of course the Cypress tests need to be updated to reflect the new representation of collapsed columns. This test suite is documented here: https://openrefine.org/docs/technical-reference/functional-tests
Let me know if you need help with that.

wetneb:
@VhugoJc are you still on this? It's nearly good to go, we just need to update the Cypress tests to make sure they still pass after your change.

VhugoJc:
ok, let me try again. sorry for the delay.

wetneb:
I have changed the tooltip to be "Expand column 'foo'" instead of "Expand this column", so that users are able to tell which column they are about to expand. This also made it easier to adapt the Cypress tests, since they were relying on this tooltip to click the column headers to expand the column.

Updating the Cypress tests is definitely some work, but it's nice that doing that encourages us to improve the UX! And it's not the first time it happens to me.

PR #5992, opened 2023-07-27
#5603 changes the batch size as per condition

### LINKED ISSUE #5603: Respect "Batch size" if specified by reconciliation manifest
Reconciliation manifests can sometimes have a [ `batchSize` property set in their manifest](https://reconciliation-api.github.io/specs/0.2/#service-manifest) indicating the maximum number of reconciliation queries in a reconciliation query batch.

OpenRefine currently sends 10 reconciliation queries per batch rather than the value specified by the `batchSize` property. By respecting the `batchSize` property many users would likely see less HTTP-requests and faster reconciliation.

Subset of #2549

ayushrai206:
Fixes #5603 

- This tweaks the batchsize to 1% of the dataset if batchsize in service manifest allows for it 
- if there is no batchsize given in the service manifest or 1% of dataset is already less than the 10(the default batch size) we keep the batch size to 10.

### DIFF
--- a/main/src/com/google/refine/model/recon/StandardReconConfig.java
+++ b/main/src/com/google/refine/model/recon/StandardReconConfig.java
@@ -166,7 +166,7 @@ public class StandardReconConfig extends ReconConfig {
         this(service, identifierSpace, schemaSpace,
                 type != null ? type.id : null,
                 type != null ? type.name : null,
-                autoMatch,batchSize, columnDetails, limit);
+                autoMatch, batchSize, columnDetails, limit);
     }
 
     public StandardReconConfig(
@@ -179,7 +179,7 @@ public class StandardReconConfig extends ReconConfig {
             boolean autoMatch,
             int batchSize,
             List<ColumnDetail> columnDetails) {
-        this(service, identifierSpace, schemaSpace, typeID, typeName, autoMatch,batchSize, columnDetails, 0);
+        this(service, identifierSpace, schemaSpace, typeID, typeName, autoMatch, batchSize, columnDetails, 0);
     }
 
     /**
@@ -210,7 +210,7 @@ public class StandardReconConfig extends ReconConfig {
         this.typeID = typeID;
         this.typeName = typeName;
         this.autoMatch = autoMatch;
-        this.batchSize=batchSize;
+        this.batchSize = batchSize;
         this.columnDetails = columnDetails;
         this.limit = limit;
     }
@@ -223,26 +223,22 @@ public class StandardReconConfig extends ReconConfig {
         }
         return null;
     }
-        @Override
-        @JsonIgnore
-        public int getBatchSize (Project project) {
-        if((project.rows.size()/100)>10)
-        {
-            if((project.rows.size()/100)<batchSize)
-            {
-                return (project.rows.size()/100);
-            }
-            else
-            {
-                if(batchSize!=0)
-                {
+
+    @Override
+    @JsonIgnore
+    public int getBatchSize(int rowCount) {
+        if ((rowCount / 100) > 10) {
+            if ((rowCount / 100) < batchSize) {
+                return (rowCount / 100);
+            } else {
+                if (batchSize != 0) {
                     return batchSize;
                 }
             }
 
         }
         return 10;
-        }
+    }
 
     @Override
     public String getBriefDescription(Project project, String columnName) {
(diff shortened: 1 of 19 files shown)


main/tests/server/src/com/google/refine/model/recon/StandardReconConfigTests.java
ayushrai206, line 1: I have changed this in the next commit.

ayushrai206, line 1: This was mistakenly changed , i have fixed this in the next commit.

main/src/com/google/refine/model/recon/StandardReconConfig.java
```diff
@@ -217,12 +223,26 @@ public ReconType getReconType() {
         }
         return null;
     }
+        @Override
+        @JsonIgnore
+        public int getBatchSize (Project project) {
```
wetneb, line 228: Instead of taking the entire `Project` as argument, perhaps you could pass only the number of rows, as this is all we need here? This would make sure that other implementations of `ReconConfig` also don't rely on too exotic aspects of the project to determine their batch size.

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -345,6 +345,7 @@ ReconStandardServicePanel.prototype.start = function() {
         schemaSpace: this._service.schemaSpace,
         type: (type) ? { id: type.id, name: type.name } : null,
         autoMatch: this._elmts.automatchCheck[0].checked,
+        batchSize:this._service.batchSize,
```
wetneb, line 348: Technically, the manifest is provided by a third party so the `batchSize` attribute may contain basically anything. It could be useful to add a bit more checks here, only passing it to the backend if it is a positive number.

wetneb (commented):
Very nice!
You could run `mvn formatter:format` to format your Java code according to our conventions.

main/src/com/google/refine/model/recon/StandardReconConfig.java
```diff
@@ -217,12 +223,26 @@ public ReconType getReconType() {
...
+            {
+                return (project.rows.size()/100);
+            }
+            else
+            {
+                if(batchSize!=0)
+                {
+                    return batchSize;
+                }
+            }
```
tfmorris, line 241: I think judicious use of `Math.max()` & `Math.min()` might make this more concise and easier to understand. Something along the lines of this (untested) code:
```suggestion
        return Math.min(Math.max(row_count/100, 10), manifest_batch_size);
```

ayushrai206:
I have shared the progress till now to gets some reviews , I am yet to write the Unit tests.

main/src/com/google/refine/model/recon/StandardReconConfig.java
```diff
@@ -220,8 +226,10 @@ public ReconType getReconType() {
 
     @Override
     @JsonIgnore
-    public int getBatchSize() {
-        return 10;
+    public int getBatchSize(int rowCount) {
+
+        return Math.min(Math.max(rowCount / 100, 10), batchSize);
+
```
wetneb, line 232: Cosmetic changes: I'd remove the blank lines before and after the return statement (not sure if our linter could take care of that…)

main/tests/server/src/com/google/refine/model/recon/StandardReconConfigTests.java
```diff
@@ -164,6 +165,27 @@ public void testReconstructNoType() throws IOException {
...
+                + "\"autoMatch\":true,"
+                + "\"batchSize\":50,"
+                + "\"columnDetails\":["
+                + "    {\"column\":\"_ - id\","
+                + "     \"property\":{\"id\":\"P3153\",\"name\":\"Crossref funder ID\"}}"
+                + "],"
+                + "\"limit\":0}";
+        StandardReconConfig c = StandardReconConfig.reconstruct(json);
+        assertEquals(c.getBatchSize(10), 10);
+        assertEquals(c.getBatchSize(1200), 12);
```
wetneb, line 185: I am asking myself if for a project with 1200 rows and a service that supports batches of 50 queries, we really want to use a batch size of 12.
I think it's not so crucial that the progress report is granular down to 1%. Personally I am okay to accept progress jumps of up to 10%  if it helps speed up the entire process. I'll ask on the issue if people would agree with using ` / 10` instead of ` / 100`.

ayushrai206: Yes,I have personally tested this on a data set and 12 is very slow!

wetneb (approved):
Well done!

PR #5997, opened 2023-07-28
Add support for JSON Objects to forEachIndex() control

### LINKED ISSUE #3147: Implement `forEachItem` for JSON object

tfmorris:
Fixes #3147

I'm on the fence as to whether this should be a separate control (name TBD), as I originally spec'd, or if rolling the functionality into the existing `forEachIndex` is close enough semantically given that the array index and the object key are both performing essentially the same function. This pull request adopts the latter approach. Doing it the other way would basically involve cloning the code, giving it a new function name, and making the error messages more specific.

- add tests for `forEachIndex`
- enhance error checking to reject using the same variable name for both index and value
- add support for JSON objects withe same structure, but the object key instead of the array index returned in the index/key variable

### DIFF
--- a/main/resources/com/google/refine/grel/ControlEvalError.properties
+++ b/main/resources/com/google/refine/grel/ControlEvalError.properties
@@ -1,5 +1,5 @@
 filter=First argument is not an array
-foreach_index=First argument to forEach is not an array
+foreach_index=First argument to forEachIndex is not an array
 foreach=First argument to forEach is not an array or JSON object
 for_range=First, second, and third arguments of forRange must all be numbers
 
@@ -12,3 +12,4 @@ expects_five_args={0} expects five arguments
 expects_second_arg_var_name={0} expects second argument to be a variable name
 expects_second_arg_index_var_name={0} expects second argument to be the index's variable name
 expects_third_arg_element_var_name={0} expects second argument to be the element's variable name
+expects_second_third_args_different={0} expects the second and third arguments to be different variable names
--- a/main/resources/com/google/refine/grel/ControlEvalError.properties
+++ b/main/resources/com/google/refine/grel/ControlEvalError.properties
@@ -1,5 +1,5 @@
 filter=First argument is not an array
-foreach_index=First argument to forEachIndex is not an array
+foreach_index=First argument to forEachIndex is not an array or JSON object
 foreach=First argument to forEach is not an array or JSON object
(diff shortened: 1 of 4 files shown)


tfmorris:
What do folks think about the approach of extending `forEachIndex()` as opposed to creating a separate `forEachEntry()` or `forEachItem()` or some such?

I'm not wedded to this approach, but it does allow reuse of a lot of the existing machinery and doesn't seem like a hugely awkward extension.

wetneb:
Maybe @ostephens has thoughts on the matter?

ostephens:
`Added comment on #3147 but in summary I'm fine with this approach

tfmorris:
@ostephens Thanks for the review and the feedback.
@wetneb If you are satisfied with the `hashCode()` update, can you please approve?

wetneb (approved):

PR #6021, opened 2023-08-21
Issue 5998 entity identifiers dialog improvements

### LINKED ISSUE #5998: UI improvements for the "Add entity identifiers column" action
When using the "Add entity identifiers column action", the UI currently looks like this:
image

### Proposed improvements

* the input field for the column name should be focused when the dialog opens
* the input field could be pre-filled with an automatically generated column name which does not conflict with any existing column name (for instance, if the base column is called "Name", the default column name could be "matched id for Name", perhaps?)
* the "Submit Query" label for the validation button does not make sense, as we are not submitting a query. It should be "OK" or "Create column", I would say. It should be marked as primary (default) button, which generally appears in bold I think
* the title of the dialog does not clearly separate the user-provided data (column name) from the rest. Instead of being `Add column containing entity identifiers on Name`, it could be `Add column of entity identifiers from column "Name"` (note the double quotes)

### Additional context

Discovered while reviewing #5960.
Maybe @OpenRefine/designers have more thoughts about this?

elebitzero:
Fixes #5998

- Add quotes (") around column name in title.
- Focus on text input when dialog opens.
- Change "Submit Query" to "OK."

After:

Add entities column dialog

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
+++ b/main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
@@ -243,7 +243,7 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
       DOM.loadHTML("core", "scripts/views/data-table/add-q-column-dialog.html"));
 
     var elmts = DOM.bind(frame);
-    elmts.dialogHeader.text($.i18n('core-views/add-id-col', "\"" + column.name + "\""));
+    elmts.dialogHeader.text($.i18n('core-views/add-id-col', column.name));
     
     elmts.or_views_newCol.text($.i18n('core-views/new-col-name'));
     elmts.okButton.html($.i18n('core-buttons/ok'));
--- a/main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
+++ b/main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
@@ -243,7 +243,7 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
       DOM.loadHTML("core", "scripts/views/data-table/add-q-column-dialog.html"));
 
     var elmts = DOM.bind(frame);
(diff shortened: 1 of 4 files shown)


main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
```diff
@@ -243,14 +243,16 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
       DOM.loadHTML("core", "scripts/views/data-table/add-q-column-dialog.html"));
 
     var elmts = DOM.bind(frame);
-    elmts.dialogHeader.text($.i18n('core-views/add-id-col', column.name));
+    elmts.dialogHeader.text($.i18n('core-views/add-id-col', "\"" + column.name + "\""));
```
wetneb, line 246: Are we certain that such double quotes will be used in all languages ? Previously, you added those double quotes to the messages and it feels better to me. If you only update the English message for it, translators will be prompted to update the translation in their language, making it possible for them to review if double quotes are also appropriate or if they want to use other punctuation signs.

wetneb (approved):
Thank you! I did not know about the `autofocus` attribute on input fields. @Abbe98 would this be okay for you?

main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js
tfmorris: The quotes definitely belong in the message. Any string concatenation in a user visible message is a "code smell," indicating something's probably wrong.

elebitzero: I agree if we are going to use quotation marks, it should be in the message, since it would look different in some other languages like German and French.

I was thinking of using square brackets which seems more universal, for example:

Add new column containing entity identifiers on [Calcium]

So I switched to concatenating the quotation marks so it could be changed without having to update the translation.  Anyway, I can change it back.

There are several other places we are not distinguishing the column/entity name in a title string.

PR #6022, opened 2023-08-22
Improve SQL bad keyword detection - fixes #6019

### LINKED ISSUE #6019: Create Project: Issue SQL Query - Query cannot contain data manipulation keyword: DELETE
### To Reproduce
1. First, create a new project
2. Then, select database and create a query where any column name contain the word delete, for example "isdeleted"
3. Finally, click preview query result

### Current Results

 Message: Query cannot contain data manipulation keyword: DELETE

### Expected Behavior

Query should run

### Screenshots

image

### Versions
 - Browser Version: Google Chrome Version 116.0.5845.96 (Official Build) (64-bit)
 - JRE or JDK Version:  JRE 1.8
 - OpenRefine: OpenRefine 3.7.4

tfmorris:
Fixes #6019

- Make SQL bad keyword detection more reliable by using regex instead of substring
- Improve localization by not using string concatenation to build messages

BUT I'm not convinced that we shouldn't just remove all this code since it's almost impossible to do reliably and could give users/administrators a false sense of security. They should be using their DB admin privs to restrict user access.

### DIFF
--- a/extensions/database/module/langs/translation-cs.json
+++ b/extensions/database/module/langs/translation-cs.json
@@ -48,7 +48,6 @@
     "database-source/alert-connection-edit": "Připojení bylo upraveno",
     "database-source/form-validation-failure": "Nová podpoba připojení není platná!",
     "database-source/alert-invalid-query-select": "Dotaz musí začínat klíčovým slovem SELECT",
-    "database-source/alert-invalid-query-keyword": "Dotaz nemůže obsahovat klíčové slovo pro manipulaci s daty:",
     "database-source/alert-query": "Musíte zadat validní dotaz",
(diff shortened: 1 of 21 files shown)


wetneb:
> BUT I'm not convinced that we shouldn't just remove all this code since it's almost impossible to do reliably and could give users/administrators a false sense of security. They should be using their DB admin privs to restrict user access.

I agree!

wetneb (approved):

tfmorris:
This is still a draft because I wasn't sure which approach we wanted to take.

@wetneb Are you approving the current draft implementation or the proposed implementation of removing the checking altogether?

PR #6025, opened 2023-08-25
CSV Byte Order Mark (BOM) support. Fixes #1241

### LINKED ISSUE #1241: CSV files with BOM don't correctly strip BOM on import
Steps to reproduce:
1) Import CSV file [testUNID.zip](https://github.com/OpenRefine/OpenRefine/files/1266702/testUNID.zip)
2) Set encoding to UTF-8
3) Open project
4) Click "Edit column -> Add column based on this column"
5) In "New column name" type "Foo"
6) In "Expression" type cells["UNID"].value

Value is not loaded because the column name is prepended with a Byte Order Mark (BOM).

tfmorris:
Fixes #1241

* UTF-8-BOM to the encoding guesser
* Add support for stripping the BOM to the importer

### DIFF
--- a/main/src/com/google/refine/importing/EncodingGuesser.java
+++ b/main/src/com/google/refine/importing/EncodingGuesser.java
@@ -1,31 +1,32 @@
 
 package com.google.refine.importing;
 
-import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.io.InputStream;
 import java.util.List;
 
-import org.apache.commons.lang.StringUtils;
-import org.mozilla.universalchardet.UniversalDetector;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.apache.commons.lang.StringUtils;
+import org.mozilla.universalchardet.UnicodeBOMInputStream;
+import org.mozilla.universalchardet.UniversalDetector;
 
 import com.google.refine.util.JSONUtilities;
 
 /**
- * This class tries to find the correct encoding based on the http://site.icu-project.org/ and the icu4j library
- * http://site.icu-project.org/home/why-use-icu4j.
+ * This class tries to find the correct encoding based on https://github.com/albfernandez/juniversalchardet which is a
+ * Java port of Mozilla's universalchardet library
+ * https://hg.mozilla.org/mozilla-central/file/tip/extensions/universalchardet/
  * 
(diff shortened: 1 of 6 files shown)


main/src/com/google/refine/importing/EncodingGuesser.java
```diff
@@ -38,9 +39,13 @@ public final static void guess(final ImportingJob job)
...
                         String location = JSONUtilities.getString(record, "location", null);
                         if (location != null) {
-                            try (InputStream is = new BufferedInputStream(
+                            try (UnicodeBOMInputStream is = new UnicodeBOMInputStream(
                                     new FileInputStream(new File(job.getRawDataDir(), location)))) {
                                 String detected = UniversalDetector.detectCharset(is);
+                                UnicodeBOMInputStream.BOM bom = is.getBOM();
+                                if (UnicodeBOMInputStream.BOM.UTF_8.equals(bom)) {
+                                    detected = UTF_8_BOM;
+                                }
```
wetneb, line 48: There are other types of BOM markers supported (such as for UTF-16, UTF-32, with different byte orders) - would it make sense to add support for those as well?

wetneb (commented):
Thanks for tackling this very important issue!

main/src/com/google/refine/importing/EncodingGuesser.java
tfmorris: Are other BOM types problematic? I've never heard any complaints about them. Given that this bug report has been open for 5 years, I'd lean towards addressing the specific problem that was reported rather than delaying for additional investigation (unless you know what needs to be adjusted).

wetneb: Ah, I had not understood that there was a specific issue in the JDK about UTF-8 and BOM handling. According to https://github.com/gpakosz/UnicodeBOMInputStream#why, it seems that UTF-16 and UTF-32 are indeed handled correctly out of the box.

tfmorris: Actually, the issue is with Microsoft applications inserting a BOM with UTF-8 in a context where it is "[neither required nor recommended](https://en.wikipedia.org/wiki/Byte_order_mark#cite_note-5)"

Where does this PR stand with regards to approval status?

wetneb (approved):

RolfBly:
I just want to say, also on behalf of several others that I know, Many thanks for the fix!

tfmorris:
You're welcome! It would have been nice if Microsoft hadn't introduced the problem in the first place, but when you're dancing with an 400kg gorilla, we kind of have to follow their lead.

PR #6026, opened 2023-08-25
Don't write newly read projects before they're modified

### LINKED ISSUE #3805: Workspace files being written unnecessarily

tfmorris:
Fixes #3805

- Repairs constructor which sets modified date on projects
- Adds test coverage to make sure we don't write newly read projects before they're modified
- initializes the lastSaved timestamp on project load (#3805)
- check for and log errors on file renames & deletes
- refactor to decide if file write is needed upfront rather than writing file and throwing it away if unneeded
- fix test from PR #4796 to correctly test project removal
- write out workspace on project removal (fixes regression which re-broke #1418)

### DIFF
--- a/main/src/com/google/refine/ProjectMetadata.java
+++ b/main/src/com/google/refine/ProjectMetadata.java
@@ -137,6 +137,7 @@ public class ProjectMetadata {
 
     public ProjectMetadata(Instant created, Instant modified, String name) {
         this(created);
+        _modified = modified;
         _name = name;
     }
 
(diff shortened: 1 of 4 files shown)


wetneb (approved):
Thank you!

PR #6043, opened 2023-09-06
fix(sec): upgrade com.google.guava:guava to 32.0.0-jre

cubxxw:
### What happened？
There are 1 security vulnerabilities found in com.google.guava:guava 31.1-jre
- [CVE-2023-2976](https://www.oscs1024.com/hd/CVE-2023-2976)

### What did I do？
Upgrade com.google.guava:guava from 31.1-jre to 32.0.0-jre for vulnerability fix

### How can we automate the detection of these types of issues?
By using the [GitHub Actions](https://github.com/murphysecurity/actions) configurations provided by murphysec, we can conduct automatic code security checks in our CI pipeline.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,7 +101,7 @@
     <opencsv-multichar.version>2.4</opencsv-multichar.version>
     <swc-parser-lazy.version>3.1.9</swc-parser-lazy.version>
     <commons-io.version>2.13.0</commons-io.version>
-    <guava.version>31.1-jre</guava.version>
+    <guava.version>32.0.0-jre</guava.version>
     <jaxb-api.version>2.3.1</jaxb-api.version>
     <juniversalchardet.version>2.4.0</juniversalchardet.version>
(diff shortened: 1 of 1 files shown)


tfmorris:
Thanks for the PR. We generally update to new versions of dependencies as they become available anyway, but I don't believe this is actually a security risk for OpenRefine since [we don't use](https://github.com/search?q=repo%3AOpenRefine%2FOpenRefine%20FileBackedOutputStream&type=code) `FileBackedOutputStream` c.f. https://github.com/google/guava/issues/2575#issuecomment-1622208055

As an aside, I've got a strong preference for linking to the canonical CVE entry (ie https://nvd.nist.gov/vuln/detail/CVE-2023-2976) as opposed to a third party site.

Reviewing the Guava release notes, it looks like we need at least 32.0.1 since 32.0 is broken on Windows, but we probably want to go straight to the latest 32.1.2.

PR #6045, opened 2023-09-06
Introducing a new variable that takes only required properties of suggest options and excludes the flyout properties

### LINKED ISSUE #3141: XSS vulnerability in suggest flyouts

ayushrai206:
Fixes #3141 

an object variable sanitizedSuggestOptions which only takes some properties used by the suggestOptions object and then I pass this new var sanitizedSuggestOptions for initializing the suggest library.

I am trying to take this approach for all the places where we have suggestOptions or suggestConfig variables.

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
+++ b/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
@@ -246,7 +246,11 @@ ReconStandardServicePanel.prototype._wireEvents = function() {
     if (this._service.ui && this._service.ui.access) {
       suggestOptions.access = this._service.ui.access;
     }
-    input.suggestT(suggestOptions);
+    var sanitizedSuggestOptions = {
+      query_param_name: suggestOptions.query_param_name,
+      access: suggestOptions.access,
+    }
+    input.suggestT(sanitizedSuggestOptions);
   }
 
   input.on("bind fb-select", function(e, data) {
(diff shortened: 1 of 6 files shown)


main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -246,7 +246,11 @@ ReconStandardServicePanel.prototype._wireEvents = function() {
     if (this._service.ui && this._service.ui.access) {
       suggestOptions.access = this._service.ui.access;
     }
-    input.suggestT(suggestOptions);
+    var sanitizedSuggestOptions = {
+      query_param_name: suggestOptions.query_param_name,
+      access: suggestOptions.access,
+    }
+    input.suggestT(sanitizedSuggestOptions);
```
wetneb, line 253: Once you have introduced a function to do the sanitization, you could simply invoke it like this:
```suggestion
    input.suggestT(sanitizeSuggestOptions(suggestOptions));
```

wetneb (commented):
That looks like a great start!

Instead of writing the same sort of code before each invocation of the suggest library, you could instead define a new function which would do this work, and then use this function in all the relevant places.

This would look like that:
```javascript
function sanitizeSuggestOptions(options) {
  return {
     // sanitized options object just like you are currently doing
  };
}
```

Concerning the fields to preserve, I think there are a couple more according to [the specs](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#suggest-metadata):
* `service_url`
* `service_path`

wetneb (approved):
Congratulations!

PR #6056, opened 2023-09-13
formatter:format changes in database module

tfmorris:
No issue.

Changes to make the linter happy. It looks like the latest security patch made it through CI without linting?

@wetneb any idea how that happened?


tfmorris:
@wetneb I automerged this since it was trivial and it'll allow me to rebase my other PR on a clean master branch, but I'm still curious how this slipped through. Something about the setup of our temporary security repos?

wetneb:
Yes indeed the CI does not run on security repos, so I should have thought about running linting on my side… sorry about that and thanks for taking care of it!

PR #6060, opened 2023-09-14
Improved property selection in reconciliation dialog

### LINKED ISSUE #6031: Property selection in reconciliation dialog is unreliable
In the reconciliation dialog, there is a panel on the right-hand side to configure which columns should be matched with which properties offered by the service:

image

There are a couple of problems with this UI.

### Redundant checkbox
The checkbox before each column name is redundant with the property field. It only makes sense to check this box if the associated property field is filled, and if the property field is filled then the box should be checked.

In other words, none of the following states are useful to the user:
* image
* image

Therefore, the value of the checkbox could be inferred directly from the state of the associated input field.

### Unclear auto-completion status in property field
The property field does not indicate clearly if the property name typed by the user has been correctly matched to a property offered by the reconciliation service.

When typing a property name in that field, the user is presented with a few options in a drop-down menu (if the service supports it):
image

If the user picks an option in the list, the property will be correctly used for reconciliation. The state of the input field after doing so looks like this:
image

If instead the user ignores the suggestions and clicks anywhere else, without picking an option, then the input field will look exactly the same, but the property will not be used for reconciliation because it is not associated with any valid property id:
image

This can be quite frustrating because once reconciliation has run, it is not necessarily easy to determine if a property was indeed used for reconciliation or not (depending on how the service uses additional properties for scoring). Moreover, one has to wait for the entire reconciliation process to finish before being able to spot this problem (in 3.x).

### Proposed solution

* Remove the checkbox
* Mark the property fields as having an error (similarly to what we do in the Wikibase schema) if they are non-empty but have not been validated with auto-completion. Refuse to start reconciliation as long as any property field is an error state.

### Alternatives considered

Find a new design for this UI, which could potentially also solve #2075 in the same go.

### Additional context

To be considered for NFDI-funded improvements to reconciliation. Cc @lozanaross @ayushrai206 and @OpenRefine/designers

elebitzero:
Fixes #6031

- Column Mapping
  * Removed checkbox (Column mapping is included if matched to a property.).
  * After selecting a property mapping from the suggest dropdown, the mapping is displayed as "`property` `(P-id)` edit" (See screenshot below).
  * Clicking edit will revert the field back to an editable textbox.

- Start Reconciling... button
  * Reconciling will only begin if each column that has a mapping was successfully mapped to a property (name, id).
  * If any column mapping contains plaintext, it is invalid so an alert window is popped up 'Column x is unmapped.' and the reconciliation is aborted.
  
image

  
- Testing
  * Tested above cases.
  * Tested that the /api call to /command/core/reconcile contains the correct information for the mapped columns. 
image

I will cover the reconcile against type part in a follow-up PR.


main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -306,32 +349,34 @@ ReconStandardServicePanel.prototype.start = function() {
...
-            column: this.getAttribute("columnName"),
-            property: {
-              id: property,
-              name: property
-            }
-          });
+        let value = jQueryTrim(this.value);
+        let hasValue = value && value.length > 0;
+        if (hasValue) {
+          alert('Column '+$(this).data("columnName")+' is not mapped.');
```
wetneb, line 370: I would localize this message with `$.i18n` and I suggest the column name gets enclosed by double quotes to make it stand out more and separate it from our own text.
Also, if multiple columns are left in an unmapped state but non-empty text field, a alert will be generated for each of them. Having sequences of alert like this is not amazing UX, so I would suggest to either only show the error for a single column (the first one the validation code encounters). Adding all the column names in a single alert could also be an option, but it's probably harder to implement and localize.

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -206,23 +206,60 @@ ReconStandardServicePanel.prototype._populatePanel = function() {
...
+    $(td1).data('id','property').css('position','relative');
+
+    let mappedColumn = $("<span>").addClass("mapped-column")
+        .css("display", "none")
+        .css("width","100%")
+        .css("padding","2px")
+        .css("position","absolute")
+        .css("top","0")
+        .css("left","0")
+        .css("background",'transparent');
```
wetneb, line 227: Perhaps it would be worth migrating most (if not all) of those CSS rules to a CSS file? And similarly below.

wetneb (commented):
Amazing! I really like it. @OpenRefine/designers any thoughts about the design?

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
elebitzero: I Agree, but I would like to keep it like this for now because there is a different error mechanism being used in the Wikibase schema editor where invalid fields are highlighted in red, so I may change it to that later.  The window.alert() is a placeholder for now.

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
elebitzero: Yes

lozanaross:
I think this is a great interim solution until a more holistic redesign of the whole dialog is developed and tested with users. I am just wondering if the "edit" button is visually consistent with other parts of the UI. But I don't necessarily have alternative suggestions.

elebitzero:
The edit button's style is based on the one from the data grid:

image

Lydiaofficial:
Alright. Will do.

wetneb (approved):

PR #6065, opened 2023-09-25
Only fix launch in snap mode, not add package

kristbaum:
Fixes #{issue number here}
Not launching browser in snap package
- Check if in snap, then open Browser


kristbaum:
As talked about here: #4484

.gitignore
```diff
@@ -66,6 +66,9 @@ refine-dev.ini
 # Ignore Node modules that might inadvertently install at current path instead of user or configured env var NODE_PATH 
 node_modules/
 
+# Ignore locally build snaps
+*.snap
+
```
wetneb, line 71: We can probably also remove this change since we are not building the Snaps in this repository

server/src/com/google/refine/Refine.java
```diff
@@ -487,7 +487,7 @@ public void actionPerformed(ActionEvent e) {
     }
 
     private void openBrowser() {
-        if (!Desktop.getDesktop().isSupported(Desktop.Action.BROWSE)) {
+        if (!Desktop.getDesktop().isSupported(Desktop.Action.BROWSE) || !(System.getenv("SNAP") == null)) {
```
wetneb, line 490: Quick simplification:
```suggestion
        if (!Desktop.getDesktop().isSupported(Desktop.Action.BROWSE) || System.getenv("SNAP") != null) {
```

wetneb (approved):

PR #6066, opened 2023-09-25
Fix selection of Wikibase statement merging strategy

### LINKED ISSUE #6064: Unable to add additional values to statements on existing Qids when uploading via OR schema
[LSEPPR BHL schema.txt](https://github.com/OpenRefine/OpenRefine/files/12712961/LSEPPR.BHL.schema.txt)
Using an schema in OpenRefine I am able to successfully add new statements to a set of existing Qids, but where a statement already exists and I want to add a second value - eg instance of, published in the second value is not added, though my reference for that value is added.
For example I experimented with [After Merkel: Germany from peace to war - Wikidata](https://www.wikidata.org/wiki/Q122664150). It already contains an instance of scholarly article statement. And I also want to add a value of academic journal article.

When my general schema didn't work I tried to isolate the issue by creating a schema simply to add the academic journal article value.   But when I upload those changes to Wikidata, although OR indicates the update has happened there is no evidence of it on the history tab of the item - and the second value is still not added.  

I understand from the OR forum that I am using the correct config settings 

image

A full description of the issue and what has been suggested so far is here https://forum.openrefine.org/t/adding-additional-statements-to-wikidata-property/306/12 but I am still unable to resolve the issue.  

I'm working in Windows 10, on Chrome, and OR 3.7.4
[vol 3 issue 1 LSEPPR for BHL upload.xls](https://github.com/OpenRefine/OpenRefine/files/12712921/vol.3.issue.1.LSEPPR.for.BHL.upload.xls)

I've attached spreadsheet of the data, and the OR schema. 

Thanks

wetneb:
Closes #6064.

There were two closely related but separate problems:
* the "Property and value" strategy (`SnakStrategy` internally) had a wrong identifier, meaning that it was not possible to select it manually (it would only work if used as the default strategy without editing the merging settings)
* the "Property, value and qualifiers" strategy was actually only a sort of "Property and qualifiers" strategy because it did not actually check the value of the main snak of each statement!

### DIFF
--- a/extensions/wikibase/module/scripts/dialogs/statement-configuration-dialog.js
+++ b/extensions/wikibase/module/scripts/dialogs/statement-configuration-dialog.js
@@ -18,7 +18,7 @@ class PropertyStrategy {
 
 class SnakStrategy {
   id() {
-    return 'property';
+    return 'snak';
   }
 
   initialState() {
--- a/extensions/wikibase/src/org/openrefine/wikibase/schema/strategies/QualifiersStatementMerger.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/schema/strategies/QualifiersStatementMerger.java
@@ -49,11 +49,14 @@ public class QualifiersStatementMerger implements StatementMerger {
 
     @Override
     public boolean match(Statement existing, Statement added) {
+        Snak existingSnak = existing.getMainSnak();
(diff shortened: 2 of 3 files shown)


wetneb:
Merging soon unless there are concerns

tfmorris:
I don't have the knowledge necessary to review this.

PR #6077, opened 2023-09-29
Update to Jetty 11 and other breaking Butterfly API changes

tfmorris:
Fixes #{issue number here}

- Update to Jetty 11 and Jakarta Servlet 5
- Update Apache commons file uploader to track Servlet API changes
- Update to Velocity 2.3

This is implemented and lightly tested and appears to work, but it's a breaking change for all extensions, so I'm not convinced it's needed until there are other breaking changes needed (and perhaps not even then).


tfmorris:
NOTE: This was tested against a local install of the proposed Butterfly 2.0, so won't run in CI until that's published publicly somewhere.

tfmorris:
Replaced by #7707

PR #6088, opened 2023-10-05
Bump max Java version to 21

### LINKED ISSUE #6087: Add support for Java 21
We currently only test up to Java 17 and issue a warning on startup if the Java version is higher. Since the Java 21 LTS release is out now, it would be good to support that.

### Proposed solution
Add CI testing for at least Java 21 (not sure we need any intermediary versions).

### Alternatives considered

Add CI testing support for all major versions in the supported range.

wetneb:
Fixes #6087

Let's see if the Temurin 21 distribution is already available through GitHub Actions.

### DIFF
--- a/.github/workflows/pull_request_server.yml
+++ b/.github/workflows/pull_request_server.yml
@@ -47,7 +47,7 @@ jobs:
   linux_server_tests:
     strategy:
       matrix:
-        java: [ 11, 20 ]
+        java: [ 11, 21 ]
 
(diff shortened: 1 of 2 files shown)


elebitzero:
The Temurin distribution of JDK 21 is not released yet. The issue below says that you can specify the version as '21-ea' if you want to test it now.

Ref: https://github.com/actions/setup-java/issues/541.

tfmorris (approved):
Looks like it's working now.

PR #6092, opened 2023-10-12
Switch to Apache commons-csv - fixes #2268

### LINKED ISSUE #2268: Migrate from old private fork of OpenCSV
We are currently using a custom snapshot of OpenCSV which is years old. The OpenCSV project is still active and a lot of releases have been published since then. We should look into upgrading to a newer version, which would also let us get rid of the locally stored .jar.

tfmorris:
Fixes #2268

- Switch from our private fork of OpenCSV to Apache commons-csv which has multi-character delimiter support as of a couple of years ago.
- Add a TODO for better error reporting to the TabularSerializer interface and remove some redundant qualifiers
- Modify a parsing test for a quote before a field separator in "ignore quotes" mode. The previous behavior was to strip the quote. The new behavior is to preserve the quote and include it in the data, which seems correct to me. (NOTE: The quote handling was modified by an OpenRefine contributor, separately from the OpenCSV team, so I'm not sure how much belief to put in it. I'm more inclined to trust the Apache Commons CSV team.)
- Modify writing test which outputs a leading empty field to accept having it quoted as being correct. Whether or not this is preferable is less clear to me, but it's a legal CSV.
- Modify e2e tests to:
  - escape quotes in test data so that they get parsed correctly during test fixture setup
  - always send importer options, not just when there's a metadata tag
  - adjust separator tests to deselect "use quotes" when switching from CSV to TSV because the interior quotes aren't correctly escaped in that case
- Three tiny unrelated changes:
  - Give a unique name to a test that had a duplicate name (presumably cut & paste error)
  - Add a comment to a flaky test that fails 50% of the time on my local system (but it's still enabled)
  - Add trailing newline at EOF to test data file
Please review the test changes carefully to make sure they're OK. I think they are, based on my research, but am happy to listen to arguments otherwise. 

According to published benchmarks, this may be slightly slower, but roughly the same order of magnitude in performance as OpenCSV. The benchmarks were presumably done with a modern version of OpenCSV, not our ancient version, so it's unclear how valid they are. Both are about 1/3 the speed of the highest performance CSV parsers available (but none of them have support for multi-character field delimiters).


tfmorris:
The test failure is for the short 3-line `food.mini.csv` parse with a TAB delimiter and quote character (") field encapsulation which makes Apache Commons CSV throw

`IOException reading next record: java.io.IOException: (line 1) invalid char between encapsulated token and delimiter`

which I don't really understand. It works with both the quote encapsulation turned off and with the larger `food.small.csv`.

tfmorris:
`food.mini.csv` and `food.small.csv` differ in more than just the number of lines that they have. The mini version has the header fields and the numeric fields quoted. It's apparently the quotes at the record delimiter boundaries which is make the CSV parser unhappy. Not sure whether it's an Apache bug or not yet.

tfmorris:
OK, this is ready to review now. The mentioned above issue was caused by the hand modifications that were made to `food.mini.csv` vs `food.small.csv` where the original

```
NDB_No,Shrt_Desc,Water,Energ_Kcal,
01001,"BUTTER,WITH SALT",15.87,717,
```
was changed to 
```
"NDB_No","Shrt_Desc","Water","Energ_Kcal"
"01001","BUTTER,WITH SALT","15.87","717"
```
The latter is fine when parsed as a CSV, but when it's parsed as a TSV with quotes enabled (the default), the Apache Commons CSV parser complains about the internal quote characters not being correctly escaped (since it's being parsed as a single data field with a value of  `01001","BUTTER,WITH SALT","15.87","717`).

While investigating I realized that Commons CSV throws an UncheckedIOException for this case, which was bypassing our error reporting, so now we correctly report that to the user so that they'll have some clue as to what is going on.

Additionally, the e2e tests make extensive use of CSV data for their test fixtures and some contains quotes, so I added quote escaping to the test fixture setup.

tfmorris:
Tagging @FWennerdahl who provided the original OpenCSV mods for quote handling and multi-character delimiters, as well as @ostephens and @thadguidry who might have a diversity of CSV/TSV files to test against (although we should also flag it to a wider audience for testing once it is merged).

tfmorris:
Please hold off on reviewing this. While generating performance comparison numbers I discovered that the highest performance CSV parser also apparently supports multi-character delimiters, so I want to investigate switching to that instead.

thadguidry:
No worries. Also sent you private Google Drive link for a large CSV file to test with.

tfmorris:
This is going to be replaced by #6098. Sorry for the noise!

Issue #5641, opened 2023-02-20
Add column by fetching URL redirection destination


Abbe98:
A common use case on my end is to update an old URI or URL and while it's possible to do so with the Jython extension it would be very nice to have it as accessible through core. Example task: `http://libris.kb.se/resource/auth/158606` -> `https://id.kb.se/term/sao/R%C3%B6kstenen`

### Proposed solution
Add a transform method for "Add column by fetching URL redirection destination" which would behave similarly to the existing "Add column by fetching URL" but utilize HTTP HEAD requests and only return the final URL.

### Alternatives considered
Three solutions currently being used: 1. fetching the content and parsing it for potential destination URLs 2. Fetching it with Jython 3. Scripting the fetching outside of OpenRefine

Alternatives for moving away from the above: implementing this but as an extension.

sn1616:
Working on this

wetneb:
I am not sure about introducing a dedicated operation for this: it could rather be an option in the existing dialogue for the URL fetching operation. Or a grel function.

sn1616:
Got it , going for an option in the existing dialogue for the url

tfmorris:
It might be worth considering extending this to support all header fields, not just `Location:` so that users can access any of them that they find interesting.

ianengelbrecht:
Any progress on this? I really need this to be able to follow redirects.

PR #6097, opened 2023-10-13
API: Add HTTP HEAD support to commands

Abbe98:
Allow extensions to register commands which can respond to HTTP HEAD requests.


wetneb (approved):

Abbe98:
@wetneb do you know if we have a god way of writing tests for this type of feature, given it isn't in use by core or bundled extensions?

wetneb:
Yes, I think one could have a unit test of the RefineServlet class itself, checking that it does call the corresponding `head` method on a command, when given a suitable HTTP request to react to. But I am not sure it would really add much.

tfmorris:
I just stumbled across #5641 which mentions the HEAD verb and could be related to this.

PR #6098, opened 2023-10-14
Switch to uniVocity CSV parser. Fixes #2268 & #1372 

### LINKED ISSUE #1372: Change to higher performance CSV parser
Why not adopt a more powerful CSV parser? I noticed we still use Bytecode OpeCSV parser for SeparatorBasedImporter and CsvExporter, that is old (last version 2.4: 13 Aug 2014 ) and slow. According to this [benchmark](https://github.com/uniVocity/csv-parsers-comparison) there are a lot of better alternatives. I also made some [tests](https://github.com/xseris/Benchmarking/tree/master/csv) with the latest versions of common CSV parsers, founding that bytecode csv parser is about 50-70% slower. I think that [Jackson](https://github.com/FasterXML/jackson-dataformats-text/tree/master/csv), [SFM](http://simpleflatmapper.org/0101-getting-started-csv.html) or [Univocity](https://github.com/uniVocity/univocity-parsers) could be great alternatives to increase a bit the importing performances.

### LINKED ISSUE #2268: Migrate from old private fork of OpenCSV

tfmorris:
Fixes #2268. Fixes #1372 

This replaces our private fork of OpenCSV with the uniVocity CSV parser which is 3x-4x faster, supports our multi-character separators, and has a bunch of other options that we can potentially take advantage of in the future. It's also 4x-5x faster than the Apache Commons CSV parser that we were considering switching to.

There are a couple of minor functional changes which are reflected in the tests:

- On import, a quote character before a separator character  is no longer stripped in "ignore quotes" mode. Instead it's
  included in the data field, which seems correct to me. I wasn't able to find an option which would mimic the previous behavior, but I think it's a quirk/bug in the legacy parser rather than a desired behavior.
- On export, CSV/TSV tests now check that the export uses the system line delimiter rather than always a Unix newline (\n). This is the exporter's default, which seems correct to me, but it would be easy to keep our current default if people think that's better.

The following changes were made to e2e tests:

- Disable quote processing option before switching separator from comma (,) to tab (\t) because the test data doesn't have legally escaped quotes when the separator isn't a comma. All the options are still tested, but the new ordering avoids a configuration which (correctly) errors.

- Force options to be sent by project creation text fixture (If a tag name wasn't present, no options were sent previously, which is a super weird side effect!)

- Escape test fixture data which contains quotes before sending it to be parsed as a CSV file. The previous parser was more permissive, but there's no reason for us not to be generating correctly escaped data for the normal case.


tfmorris:
OK, this is ready for review. Sorry for the noise with the previous PR. This will be much better!

Again tagging @FWennerdahl who provided the original OpenCSV mods for quote handling and multi-character delimiters, as well as @ostephens and @thadguidry who might have a diversity of CSV/TSV files to test against (although we should also flag it to a wider audience for testing once it is merged).

Please review the test changes especially carefully to be sure they're OK.

Two minor loose ends are:
- The new parser includes a format guesser which we could potentially switch to from our current private separator guesser, but it probably needs more testing with a larger data set first. 
- I increased the maximum cell size from the default of 4K to 256K to handle one of @thadguidry 's pathological test case files, but this may be something that we want to expose as a user settable parameter. That can be done separately though.

ostephens:
Thanks @tfmorris - this is looking good to me. Tested with a dataset of 1033 tsv files which include a wide mixture of character sets and that have been problematic to parse with the existing parser (failures look like missing/counting incorrect characters as delimiters - leading to many more columns in a the eventual project than were in the original files and some cells with multiple properties/records in a single cell).

With this branch the files were processed significantly faster, and without the errors that I've seen in OpenRefine previously (tested this commit vs 3.7.5 running linux version on mac with 8192Mb memory assigned to OR).

wetneb (approved):
Exciting! The benchmark by Owen looks very promising. And it's amazing to get rid of this old unmaintained forked dependency.

The small behavior changes look acceptable to me intuitively. I wouldn't be surprised if some issues with the new importer behavior pop up but it should not be hard to correct if needed.

PR #6102, opened 2023-10-16
Bug: Ensure the "Create project" screen is the default one on startup

### LINKED ISSUE #5980: OpenRefine opens the "Open project" tab by default
Until recently, when starting OpenRefine, one would be presented with the first panel ("Create project").
In the current development version (master branch), the "Open project" tab opens instead.

### To Reproduce
1. First, do `./refine`
2. Then, observe the panel that appears in the browser.

### Current Results
The "Open project" panel is shown.

### Expected Behavior
The "Create project" panel is shown.

### Versions
 - Operating System: Debian
 - Browser Version: both firefox and chromium
 - OpenRefine:  3.8-SNAPSHOT [9042ff7]

Abbe98:
Fixes #5980.


wetneb (approved):
Thanks a lot for taking care of it!

PR #6105, opened 2023-10-18
Delete empty test classes

wetneb:
It is not difficult to create those classes again if one wants to write tests about those GREL functions, but in the current state it feels like noise to me and gives the false impression that those functions are tested.


Abbe98 (approved):

wetneb:
That's a fun idea, but I'd prefer tests to make more meaningful assertions about the component being tested and not be there to inflate a metric. Also the use of introspection can likely cause issues (for instance due to interactions with tools like PowerMock).

PR #6118, opened 2023-10-25
Simplifies the service dialog

### LINKED ISSUE #2916: Simplify the Services dialog for reconciliation
The current slide handling for Services dialog for reconciliation is not ideal.
As mentioned in #1658 , this could probably be improved.

Currently the slider icon looks like an old product price tag to me, so not the best choice for a slider icon.  It should be an close X icon and also we should simplify the handling by introducing a visibility toggle icon as well or another button (to do the visibility toggle) called "Services..." next to the existing button "Add Standard Service...".  Instead of keeping the slide effect that retains the price tag icon and sliver of the Services panel shown on the left edge.

Something like this behavior: https://material.io/components/sheets-side#standard-side-sheet

_Originally posted by @thadguidry in https://github.com/OpenRefine/OpenRefine/issues/1658#issuecomment-656778345_

### LINKED ISSUE #5878: Show endpoint URL in reconciliation service list
When selecting a reconciliation service (in the reconciliation dialog), the user only sees the reconciliation services listed by name.
It would be useful to also show the reconciliation endpoint, for instance to:
* distinguish more easily between services provided by the official data source or third-party ones
* spot any local services, which might require starting up the local reconciliation service before using it
* inspect any other aspects of the URL which are not reflected in the name

### Proposed solution

Add the reconciliation endpoint URL underneath the name (for each service).

### Alternatives considered

There are surely other ways to present this information.

ayushrai206:
Fixes #2916 
Also Fixes #5878 
- The Service dialog is now divided into two dialogs in accordance to the proposal, which can be travelled back and forth easily by the user.
- There separate files that deal with each of the dialogs.
- The first dialog also displays the service-url  along with the service name.


main/webapp/modules/core/scripts/reconciliation/recon-dialog-2.js
```diff
@@ -0,0 +1,141 @@
...
+
+  this._createDialog();
+}
+
+ReconDialog2.prototype._createDialog = function() {
+  var self = this;
+  var dialog = $(DOM.loadHTML("core", "scripts/reconciliation/recon-dialog-2.html"));
+
+  this._elmts = DOM.bind(dialog);
+  this._elmts.dialogHeader.text($.i18n('core-recon/recon-col')+' "' + this._column.name + '"');
```
wetneb, line 48: This is something you just copied from `recon-dialog.js` and it's fine to do that, but it could be an occasion to fix the way we generate this text. We should not use concatenation here, but rather pass `this._column.name` as a parameter to the `$.i18n` function so that it can insert it in the template. The `"core-recon/recon-col"` message will need adapting accordingly, see https://openrefine.org/docs/technical-reference/translating-ui#localization-with-a-parameterized-value

main/webapp/modules/core/scripts/reconciliation/recon-dialog.html
```diff
@@ -24,7 +20,7 @@
 			<a class="button" bind="discoverServicesButton" href="https://reconciliation-api.github.io/testbench" target="_blank" ></a>
 		</div>
 		<div class="dialog-footer-bg">
-			<button class="button" bind="reconcileButton"></button>
+			<button class="button" style="font-weight: bold;" bind="nextButton"></button>
```
wetneb, line 23: Rather than manually applying a style here, I think it would be better to use the dedicated class for this: `button-primary`.
```suggestion
			<button class="button button-primary" bind="nextButton"></button>
```

main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
```diff
@@ -45,43 +46,29 @@ ReconDialog.prototype._createDialog = function() {
...
+
 };
 
-ReconDialog.prototype._registerDialogServiceOpener = function() {
-  var self = this;
 
-  $('.recon-dialog-service-opener').on('click',function() {
-    self._toggleServices();
-  });
+ReconDialog.prototype._nextDialog = function(column,selectedServiceRecordindex,serviceRecords,record) {
```
wetneb, line 67: Style suggestion: insert spaces after the `,` that separates arguments of a function.
```suggestion
ReconDialog.prototype._nextDialog = function(column, selectedServiceRecordindex, serviceRecords, record) {
```

main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
```diff
@@ -108,29 +95,49 @@ ReconDialog.prototype._cleanDialog = function() {
...
+      var mainSpan=$('<span>')
+        .addClass("recon-service-entry")
+        .appendTo(label);
+
+      var divElement=$('<div>')
+        .text(service.name)
+        .appendTo(mainSpan);
+
+      var divElementURL = $('<div>')
+        .css('color','lightgrey')
```
wetneb, line 127: In the Wikibase dialog, we use `color: grey`, so I would use that here too (for consistency and to have a slightly better contrast with the white background). Also, it would be worth storing this in a CSS class (instead of inline CSS). Ideally this color should be based on a CSS variable, which you could also apply in the Wikibase selection dialog in the same go.

wetneb (changes requested):
Fantastic! This looks overall very good to me.
Here are some suggestions of improvements for the layout:
image

image

And more suggestions below.

wetneb:
@Lydiaofficial Concerning the alignment of radio buttons, maybe my proposal to have them vertically centered is not the best one: you could instead imitate what is done in the type selection part of the second step of the dialog. There, the radio button is vertically aligned with the name of the type, which also looks good to me. You might even be able to reuse the DOM structure and the CSS classes from there, to ensure those are displayed consistently.

Lydiaofficial:
Here are modified mock ups for the corresponding issue. The major changes are the addition of radio buttons and preselction of service history.:
1. Default service UI without service selection history
image


2. Selected service / preselection history.
image

image

3. Hyperlink
image
image

4. Disabled **Next >>** button and message
image
image

Lydiaofficial:
> @Lydiaofficial Concerning the alignment of radio buttons, maybe my proposal to have them vertically centered is not the best one: you could instead imitate what is done in the type selection part of the second step of the dialog. There, the radio button is vertically aligned with the name of the type, which also looks good to me. You might even be able to reuse the DOM structure and the CSS classes from there, to ensure those are displayed consistently.

Your actual suggestion is the best for the placement of the radio buttons and I beforehand already did the same in the second step of the dialog .

main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
ayushrai206: Hi Antonin, 
I am confused about where should this be applied in the Wikibase selection dialog?

wetneb: It's in `wikibase-dialog.css`, here:
https://github.com/OpenRefine/OpenRefine/blob/ddd0137c40b6ebfd1f442f0c2a68b02f3632ba40/extensions/wikibase/module/styles/dialogs/wikibase-dialog.css#L60-L63

ayushrai206:
Hey everyone,
I have worked upon most of the reviews in my recent commit, I still need to -
1)Add a message along with the disabled next button
2)add the feature of preselecting the service according to the previous history.
3)I also noticed that there is no way right now to unselect a service, we can only select a new service to unselect the previous service, should I add a way to do this?

wetneb:
> 2)I also noticed that there is no way right now to unselect a service, we can only select a new service to unselect the previous service, should I add a way to do this?

I don't think it is needed. It does not make sense to go to the next stage without selecting a service. If the user changes their mind and no longer wants to reconcile after all, they can just close the dialog.

main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
```diff
@@ -31,11 +31,12 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
  */
 
-function ReconDialog(column, types) {
+function ReconDialog(column,previousServiceName) {
```
wetneb, line 34: Rather than using the name of a service to remember the previous selection, I think it would be better to use the endpoint. We already guarantee that endpoints are unique, whereas there could be two different services with the same name.

Also, mind the indentation: function arguments are normally not just separated by a comma, but also a space after the comma.

main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
```diff
@@ -31,11 +31,12 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
   this._column = column;
   this._serviceRecords = [];
   this._selectedServiceRecordIndex = -1;
-
+  this._record=null;
+  this.previousRecord=previousServiceName;
```
wetneb, line 39: For consistency with the rest of the style, I think it would be better to have space characters around the equal signs.

wetneb (changes requested):
Almost there! But I noticed a curious bug: when deleting a service, the selected service is changed too:
selection_change_bug

ayushrai206:
I almost forgot to generalise the pop up message, doing that asap.

Abbe98:
To be a little bold; do we need the radio buttons and the "next" button at all? It's not like the selection action enables any other action?

ayushrai206:
> Yes it would indeed save a click if selecting the service would directly take you to the next dialog. I would be happy with that too!

I tried implementing this and I am a little confused about what would happen if the radio button is already checked with the service the user used the last time to reconcile the data with as then user has to click the same already selected radio button again.

ayushrai206:
> Almost there! But I noticed a curious bug: when deleting a service, the selected service is changed too: selection_change_bug [ selection_change_bug ](https://user-images.githubusercontent.com/309908/279978371-d765e65a-898d-41bd-9fef-b0f8b25eceec.gif) [ ](https://user-images.githubusercontent.com/309908/279978371-d765e65a-898d-41bd-9fef-b0f8b25eceec.gif)

fixed this! have a look

Abbe98:
> I tried implementing this and I am a little confused about what would happen if the radio button is already checked with the service the user used the last time to reconcile the data with as then user has to click the same already selected radio button again.

I would imagine it without  radio buttons as each service would act as a button in itself.

main/webapp/modules/core/scripts/reconciliation/recon-dialog-2.js
```diff
@@ -0,0 +1,142 @@
...
+      if (i !== this._selectedServiceRecordIndex) {
+        if (this._selectedServiceRecordIndex >= 0) {
+          var oldRecord = this._serviceRecords[this._selectedServiceRecordIndex];
+          if (oldRecord.handler) {
+            oldRecord.selector.removeClass("selected");
+            oldRecord.handler.deactivate();
+          }
+        }
+
+        record.selector.addClass("selected");
```
wetneb, line 121: This code block feels overly complicated to me. It seems that you are manually re-implementing the logic of selecting a service and un-selecting any other service whenever the user clicks on a record. Could it be that this is a left over from the code from the previous dialog, which did not have radio buttons? If the radio buttons are correctly grouped together (by having the same `name` attribute) then the browser should already enforce the fact that selecting a service unselects any previously selected service). I think it would be preferable to rely on that mechanism.

wetneb:
@Abbe98 @lozanaross so we have two suggestions of changes from you both:
* going without radio buttons and turning each service into a button, sparing us the next button
* keeping radio buttons but adding a background color to the currently selected service

Because the suggestions seem to be mutually exclusive to me, is it okay for you if we treat those as follow-up improvements to be decided on and merge the PR as is?

Abbe98:
> Because the suggestions seem to be mutually exclusive to me, is it okay for you if we treat those as follow-up improvements to be decided on and merge the PR as is?

I don't consider this a blocker at all! It might, however, be nice to reach consensus in this thread. I think there should be a background color in either scenario either to indicate selection or focus/hover.

wetneb:
Ah, a background color on focus/hover would make a lot more sense to me, as it would also be of great help to line up the "delete" button with the name and URL of the service.

wetneb:
> > Almost there! But I noticed a curious bug: when deleting a service, the selected service is changed too:
> 
> fixed this! have a look

I tried this again and I still have the same behaviour.

The hover effect is nice, but I would tweak it a little perhaps. Instead of this current appearance:
image


I would prefer something like this, which would match @Lydiaofficial's design better (note the expansion of the background color to fill the entire list view):
image

This can be achieved by removing the `padding: 10px` from `.recon-dialog-service-list-container` and replacing it by a `padding: 5px 10px` on `.recon-dialog-service-list-element`.

Also, to my taste the color used for the hover background is too dark: it makes the endpoint poorly readable. I would expect something lighter, for instance the same color as the dialog header:

image

ayushrai206:
> > > Almost there! But I noticed a curious bug: when deleting a service, the selected service is changed too:
> > 
> > 
> > fixed this! have a look
> 
> I tried this again and I still have the same behaviour.
> 
> The hover effect is nice, but I would tweak it a little perhaps. Instead of this current appearance: image
> 
> I would prefer something like this, which would match @Lydiaofficial's design better (note the expansion of the background color to fill the entire list view): image
> 
> This can be achieved by removing the `padding: 10px` from `.recon-dialog-service-list-container` and replacing it by a `padding: 5px 10px` on `.recon-dialog-service-list-element`.
> 
> Also, to my taste the color used for the hover background is too dark: it makes the endpoint poorly readable. I would expect something lighter, for instance the same color as the dialog header:
> 
> image

Oh yes! this is not working indeed, following up on this.

wetneb:
Small detail: when you reply to post on GitHub, I think it's good to avoid quoting the message if it contains images, because that duplicates the images in your new post, which makes the discussion quite lengthy and hard to read. You could edit your last message and remove the quotation, so that only your reply remains.

In our call this week, we thought that it would be better not to introduce a `self._record` variable to hold the current selection in Javascript, but rather to only read it when going to the next dialog. This simplifies the state of this class and should help avoid the remaining bug mentioned above, with the selection being reset whenever a service is deleted.

wetneb (approved):

wetneb:
Great, let's merge this!

PR #6131, opened 2023-11-02
Improved type selection in recon dialog

elebitzero:
Follow-up to https://github.com/OpenRefine/OpenRefine/pull/6060

* Reconcile Against Type Mapping
   * After selecting a type mapping from the suggest dropdown, the mapping is displayed as "`type` `(Q-id)` edit" (See screenshot below).
   * Clicking edit will revert the field back to an editable textbox.

* Start Reconciling... button
   * If 'Reconcile against type' was selected reconciling will only begin if the type was successfully mapped  (name, id).  
   * If the type mapping contains plaintext, it is invalid so an alert window is popped up 'Reconcile against type xxxx is unmapped.' and the reconciliation is aborted.
   
   
image


wetneb (approved):
Looks great  to me!

PR #6136, opened 2023-11-04
Wikibase: proxy manifest requests

Abbe98:
This makes the backend proxy requests for Wikibase manifests to allow them to be hosted in MediaWiki itself(removing the CORS requirement).

MediaWiki hosted manifest example(Wikidata): https://www.wikidata.org/wiki/User:Abbe98/openrefine-manifest.json?action=raw

Setting this to draft for now but I would like to hear what others think about this approach before working out the details.

### DIFF
--- a/extensions/wikibase/src/org/openrefine/wikibase/commands/FetchManifestCommand.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/commands/FetchManifestCommand.java
@@ -1,18 +1,19 @@
 
 package org.openrefine.wikibase.commands;
 
+import static org.openrefine.wikibase.commands.CommandUtilities.respondError;
+
+import java.io.IOException;
+
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import com.google.refine.commands.Command;
 
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
 
-import static org.openrefine.wikibase.commands.CommandUtilities.respondError;
-
-import java.io.IOException;
+import com.google.refine.commands.Command;
 
 /**
  * Proxies Wikibase manifests to allow the client to bypass CORS restrictions.
--- a/extensions/wikibase/src/org/openrefine/wikibase/commands/FetchManifestCommand.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/commands/FetchManifestCommand.java
@@ -1,0 +1,43 @@
+package org.openrefine.wikibase.commands;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.refine.commands.Command;
+
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+import static org.openrefine.wikibase.commands.CommandUtilities.respondError;
+
+import java.io.IOException;
+
+/**
+ * Proxies Wikibase manifests to allow the client to bypass CORS restrictions.
+ */
+public class FetchManifestCommand extends Command {
+    
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response)
+        throws ServletException, IOException {
+        String url = request.getParameter("url");
+        try {
+            if (url == null) {
+                respondError(response, "No URL provided.");
+                return;
+            }
+
+            // fetch the contents at the url with a plain get request and return the response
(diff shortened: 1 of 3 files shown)


wetneb:
I think it makes sense to fetch the manifest from the backend instead of in the frontend.
The implementation you have looks like an open proxy to let the frontend fetch any JSON through the backend. My intuition would be to try to make this more specific to Wikibase manifests, for instance by validating the JSON structure of the manifest before returning it, and perhaps even registering it in the command in the list of known Wikibases.

GitHub identifies a security issue in the fact that we are fetching a user-supplied URL. I guess it makes sense but I am not clear on what the security implications really are. Perhaps the server could serve a maliciously crafted JSON payload which would exploit a vulnerability of our JSON parser, say (like there were similar attacks with exponential expansion of XML documents). But obviously we already do plenty of backend-side fetching of user-supplied URLs so any security issues with that would also apply there.

Abbe98:
> I think it makes sense to fetch the manifest from the backend instead of in the frontend.
The implementation you have looks like an open proxy to let the frontend fetch any JSON through the backend. My intuition would be to try to make this more specific to Wikibase manifests, for instance by validating the JSON structure of the manifest before returning it, and perhaps even registering it in the command in the list of known Wikibases.

I think that #3109 is a natural follow up, however, my intent for now is only to enable the manifest hosting use-case.

 > GitHub identifies a security issue in the fact that we are fetching a user-supplied URL. I guess it makes sense but I am not clear on what the security implications really are. Perhaps the server could serve a maliciously crafted JSON payload which would exploit a vulnerability of our JSON parser, say (like there were similar attacks with exponential expansion of XML documents). But obviously we already do plenty of backend-side fetching of user-supplied URLs so any security issues with that would also apply there.

I share your view, an exploit here would already effect other parts of the codebase. Just proxying it is probably the safest thing we can do?

Abbe98:
Maybe a user-agent string?

tfmorris:
I agree with @wetneb that some type of restriction/registration of the target domain or payload would be useful. The fact that the current security posture is poor doesn't mean that we shouldn't try to minimize, or at least not increase, the attack surface. Letting the user pick from a set of configured URLs would like resolve the security warning, but if that's not possible, validating the content (a la #3109) would be next best.

What issue # is this associated with? The PR template should have included a placeholder for the issue #.

Abbe98:
>I agree with @wetneb that some type of restriction/registration of the target domain or payload would be useful. The fact that the current security posture is poor doesn't mean that we shouldn't try to minimize, or at least not increase, the attack surface. 

In what way would this increase the attack surface?

extensions/wikibase/src/org/openrefine/wikibase/commands/FetchManifestCommand.java
```diff
@@ -0,0 +1,43 @@
...
+            }
+
+            // fetch the contents at the url with a plain get request and return the response
+            OkHttpClient client = new OkHttpClient();
+            Request req = new Request.Builder().url(url).build();
+            Response res = client.newCall(req).execute();
+            response.getWriter().write(res.body().string());
+            response.setContentType("application/json");
+            response.setStatus(200);
+            response.setCharacterEncoding("UTF-8");
```
wetneb, line 38: It would be worth cleaning up this part to just go through our generic utilities to return JSON from a command, which would probably involve parsing the JSON before (which should be fine?)
I think writing the response body before setting the encoding can cause encoding issues (if I remember correctly).

wetneb (commented):
After having a fresh look at this, I think this looks safe enough as it stands. @tfmorris do you have more thoughts about in which way this would be unsafe?

tfmorris:
I've lost the thread here. I'm happy to let you two agree on the best solution.

wetneb:
@Abbe98 I'm assuming you won't have time to make this progress, so I'd just make a few cleaning tweaks and merge it unless you have concerns?

wetneb:
I am actually tempted to make this more generic, rather than more specific, turning it into a endpoint to fetch JSON from arbitrary third-party sources and return it to the frontend. That would be a way to eliminate all uses of JSONP in the frontend, which would be a clear security win. (I think @Abbe98 has suggested something similar in the past.)

We could still have some safeguards to limit the size of the data downloaded (to make sure we don't choke up the backend) and ensure that the JSON parsing is suitable for adversarial payloads (which should be the default with Jackson, but it's always worth checking).

Abbe98:
I patched the potential issue regarding encoding that @wetneb pointed out. While I think there is more potential work in this area, I think a general purpose proxy is out of scope for now(it do indeed sound like something I might get back to!).

wetneb (approved):
Yay, let's get this merged!

PR #6142, opened 2023-11-06
Encode strings before using them as URLs for Add Column by Fetching URLs - fixes #6140

### LINKED ISSUE #6140: Correctly encode URL strings before use by Fetch URL
Currently users are required to know what URLs might need encoding and correctly encode them before using the with "Add column by fetching URLs..." I can't think of a circumstance where the user benefits from using the string directly, as is, so I would suggest that we attempt to encode it as a URL in the same way that a web browser would for addresses entered in its browser bar.

### Proposed solution
Encode the value that results from the GREL expression entered by the user in the same way that a web browser would for addresses entered in its browser bar.

### Alternatives considered
* Leaving it as is, which seems sub-optimal
* It may be desirable to have this be optional and switched on by default, so that users can disable it if need be, just to be extra cautious

tfmorris:
Fixes #6140

- Automatically encode URL before using it for fetch, using the correct encoding for each of the three different pieces of the URL which need encoding.

One question I have is whether we should be percent-decoding the string before feeding it to the URL encoder. This could potentially make it more compatible with existing recipes which uses `.encode("url")` as well as allowing the use of other sources of pre-encoded URLs, but is opening up a little bit of a can of worms.

### DIFF
--- a/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
+++ b/main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
@@ -37,6 +37,9 @@ import static com.google.common.base.Strings.isNullOrEmpty;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
```diff
@@ -321,14 +324,15 @@ Serializable cachedFetch(String urlString) {
...
-                try {
-                    return _httpClient.getAsString(urlString, headers);
-                } catch (IOException e) {
-                    return _onError == OnError.StoreError ? new EvalError(e) : null;
-                }
-            } catch (Exception e) {
-                return _onError == OnError.StoreError ? new EvalError(e.getMessage()) : null;
+            try {
+                // Do a little dance to get all the pieces of the URL correctly encoded, if possible
+                // TODO: Should we try to undo any existing percent encoding, so we don't end up double encoded?
```
wetneb, line 329: I suspect users might have got used to doing the percent encoding on their side given that OpenRefine wasn't doing it for them, so I think it'd be really valuable to avoid double encoding here.

To test this more easily, it might be convenient to extract the URL parsing logic into a method that can be called independently from the unit test, without having to fiddle with the mock test server (but it's great that you also adapted that test).

wetneb (changes requested):
Thanks for tackling this.

main/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperation.java
tfmorris: I've made the requested change.

tfmorris:
@wetneb ping. I addressed your comment a couple of weeks ago and this is ready for re-review.

main/tests/server/src/com/google/refine/util/HttpClientTests.java
```diff
@@ -27,4 +32,12 @@ public void fromHostsToPattern() {
...
         Assert.assertTrue(pattern3.matcher("myhosts.mydomain.mine").matches());
     }
+
+    @Test
+    public void urlEscaping() throws MalformedURLException, URISyntaxException {
+        final String ARABIC_PATH = "احمد";
+        final String BASE_URL = "https://example.com/";
+        final String TEST_URL = BASE_URL + ARABIC_PATH;
+        assertEquals(HttpClient.getEscapedUrl(TEST_URL), BASE_URL + UrlEscapers.urlPathSegmentEscaper().escape(ARABIC_PATH));
+    }
```
wetneb, line 42: Would it not be worth adding some test cases to demonstrate the handling of already-escaped URLs, to make sure they are not doubly escaped?

wetneb (commented):
Sorry for the delay!

PR #6146, opened 2023-11-07
Spinner alternative

tfmorris:
Refs #3017 Based on #5902

- Make spinners bigger
- Use a spinner with visible variation in the animation (from https://loading.io/css/)
- Use new spinner in Wikibase extension

There are still a couple of problems (at least):
- There's something wrong with the scaling of the small spinner so it comes out distorted
- The requirement to add 13 HTML elements is kind of clunky, but my CSS fu is not strong enough to tackle it
- A visual check should be made of all the places that the new spinners appear to make sure that they new larger size is OK, particularly for the small spinner.


tfmorris:
Continue the discussion on #5902

PR #6152, opened 2023-11-10
Display of logo of the Reconciliation service a which column is reconciled against

### LINKED ISSUE #4824: Better display of which reconciliation service a specific column is reconciled against
Currently a default OpenRefine workflow supposes users only reconcile data against a single service, e.g. Wikidata, and then proceed towards schema building and data upload with this single service in mind. However, in reality - with the introduction of the possibility to reconcile against and upload data edits to any arbitrary Wikibase (and soon Wikimedia Commons too, thanks to the work on the Commons extension) - users may wish to work across several services at once. For example enriching (or extending) a dataset from one Wikibase with data from another, in addition to also including data from various other authority control sources (e.g. GND, Getty’s ULAN, VIAF, etc). Currently, however, once a reconciliation operation is carried out — there is no easy visual cue to a user to specify what each column has been reconciled against, leading easily to making mistakes during schema building. 

## Proposed solution
The proposal here is to introduce small logos of the reconciliation service in question right underneath the column name (and column options menu), adjacent to the status bar that indicates if a column has been fully reconciled or not:
Image

In addition these logos could also be integrated in the UI during Schema building, like so: 
Image

A further complication here is that reconciliation service maintainers will need to include a logo in their manifest in order for the OpenRefine UI to be able to display it then in the two locations identified above. If no logo is provided, the name of the service may need to be spelled out in full - although this solution is not ideal due to potential of having to display long names in small font sizes (see below in Alternatives considered). The latter also won’t be practical in the schema building UI and therein column rendering should remain as it has been until now.

## Alternatives considered
Alternative considerations included a change in color scheme for each service. However this can easily introduce issues with regards to accessibility: if random colors are assigned to each service - making sure there is enough contrast between different colors and between different states (i.e fully reconciled vs partially reconciled) would add additional complexity to the UI. Furthermore, this would be a departure from currently established OpenRefine branded color palette. 

Another solution was to spell out the name of the service in type beneath the green bar indicating state of the reconciliation completion, but this would also be difficult to balance in terms of size of font (so that it's legible) vs length of the title of the service.

## Additional context
This issue concerns a UI enhancement proposed during the design process for the Wikimedia Commons extension (see the wireframes where this solution was first introduced here: https://commons.wikimedia.org/wiki/File:Structured_Data_on_Commons_file_upload_in_OpenRefine_-_wireframes_draft_v2.pdf), however it is not specific to that extension alone, but rather can be useful to the core OpenRefine UI, too.

ayushrai206:
Fixes #4824 

- We have a logo of the service against which the column is reconciled beside the recon stats bar
- We also have the logo beside the column name during schema building(if the column is reconciled against a service which has a logo )

### DIFF
--- a/extensions/wikibase/module/scripts/schema-alignment.js
+++ b/extensions/wikibase/module/scripts/schema-alignment.js
@@ -120,6 +120,9 @@ SchemaAlignment._rerenderTabs = function() {
   var schemaTab = $(DOM.loadHTML("wikidata", "scripts/schema-alignment-tab.html")).appendTo(this._schemaPanel);
   var schemaElmts = this._schemaElmts = DOM.bind(schemaTab);
   schemaElmts.targetWikibaseLabel.text($.i18n('wikibase-schema/target-wikibase-instance'));
+  schemaElmts.dialogExplanation.html($.i18n('wikibase-schema/dialog-explanation',
+      WikibaseManager.getSelectedWikibaseMainPage(),
+      WikibaseManager.getSelectedWikibaseName()));
   let editableEntityTypes = WikibaseManager.getSelectedWikibaseEditableEntityTypes();
   for (let entityType of editableEntityTypes) {
     let addButton = $('<div></div>').addClass("wbs-toolbar");
@@ -246,7 +249,7 @@ SchemaAlignment.updateColumns = function() {
      // TODO we could potentially ignore any reconciliation to a siteIRI not
      // mentioned in the manifest…
      var cell = SchemaAlignment._createDraggableColumn(column.name,
-        reconConfig && column.reconStats ? reconConfig.identifierSpace : null, reconConfig);
+        reconConfig && column.reconStats ? reconConfig.identifierSpace : null);
      this._columnArea.append(cell);
   }
 
@@ -309,6 +312,18 @@ SchemaAlignment.switchTab = function(targetTab) {
 
   if (targetTab === "#view-panel") {
     ui.dataTableView.render();
+  }
+};
+
+SchemaAlignment.resize = function() {
+  if (this._viewPanel) {
+    var panelHeight = this._viewPanel.height();
+    this._schemaPanel.height(panelHeight);
+    this._issuesPanel.height(panelHeight);
+    this._previewPanel.height(panelHeight);
+    // Resize the inside of the schema panel
+    var headerHeight = this._schemaElmts.schemaHeader.outerHeight();
+    this._schemaElmts.canvas.height(panelHeight - headerHeight - 10);
   }
 };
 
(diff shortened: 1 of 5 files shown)


extensions/wikibase/module/styles/schema-alignment.css
```diff
@@ -20,7 +20,11 @@
 .schema-alignment-tab-content {
   max-width: 70em;
 }
-
+.serviceLogo-for-schema{
+  float: left!important;
+  width: 19px;
+  height: 21px; 
+};
```
wetneb, line 27: You have an invalid syntax here: remove the `;`

main/webapp/modules/core/styles/views/data-table-view.css
```diff
@@ -192,15 +192,25 @@ table.data-table td.column-header, table.data-table th.column-header {
   background-position: -17px 0px;
 }
 
+.container {
```
wetneb, line 195: It would be better to use a more specific CSS class name.

extensions/wikibase/module/scripts/schema-alignment.js
```diff
@@ -1557,7 +1580,7 @@ SchemaAlignment._initField = function(inputContainer, mode, initialValue, change
                 initialValue.type === "wbdatevariable" ||
                 initialValue.type === "wblocationvariable" ||
                 initialValue.type === "wblanguagevariable") {
-        var cell = SchemaAlignment._createDraggableColumn(initialValue.columnName, false);
+        var cell = SchemaAlignment._createDraggableColumn(initialValue.columnName, false, null);
```
wetneb, line 1583: Instead of passing `null` here we could potentially try to look up the column by name and get its recon config. That would let us show logos on columns which are inserted in the schema.

Abbe98:
I'm a little worried about this one, should we trust reconciliation services to the extent that we inject content this way? It kinda defeats the sandboxing work done in other places.

wetneb:
Thanks for bringing this up! We already include external images in the Wikibase selection dialog, so it wouldn't be a new threat, but it's still worth considering of course.
Looking at [this stackoverflow post](https://stackoverflow.com/questions/13551372/security-concerns-of-embedding-a-third-party-image-in-html), the following concerns are brought up:
* user tracking: that's already possible whenever the user fetches the manifest of the reconciliation service or does any other actions with it, but maybe those images would be requested more often.
* 404 errors: in that case, the logo just wouldn't be displayed, and that's probably fine?
* change of content of the logo: I think that's fine
* browser vulnerability exploit: that's something that seems difficult to avoid and probably affects a lot of other places.

In general, I wonder if it would make sense to channel all calls to reconciliation services via the backend. It would have some latency consequences (for instance for suggest services) but perhaps it would be overall a good move:
* we would have a tighter control over the headers that are exposed to the reconciliation service, preventing user-agent leak for instance
* we wouldn't have to worry about services supporting CORS anymore
* the hostname of the reconciliation service would only need to be resolvable from one place, the backend. I have seen people being confused about bugs which were due to the fact that only the browser or the server were able to resolve the hostname, in situations where the server isn't running on the same machine as the browser
* in this particular case, we'd be able to sanitize (such as thumbnail) the image before serving it to the frontend, although this would likely require some pretty heavy external libraries

But of course such a move is a pretty big task and I would say it's outside the scope of this issue/PR.

Abbe98:
There is a lot of attack surface here, first one could just point to a SVG image as SVG support <script> elements and (more importantly given browser-context) various "on" events.

Secondly someone could inject something other than an image and do all kinds of nasty things.

https://security.stackexchange.com/questions/135513/what-could-an-img-src-xss-do


> We already include external images in the Wikibase selection dialog, so it wouldn't be a new threat

It wouldn't be a new threat to users of the Wikibase extension to the rest of us it would be new(AFAIK).

wetneb:
Yes I also thought about SVG, but as your link indicates, including an SVG image via `` does not execute the JS inside: it would only work if the SVG was embedded directly as an `<svg>` tag.

> Secondly someone could inject something other than an image and do all kinds of nasty things.

what are you thinking about? If it's not an image than the browser will not fall back on just including whatever is served there as HTML code, right?

Abbe98:
https://security.stackexchange.com/questions/135513/what-could-an-img-src-xss-do

All the answers apply here not only the fist one about SVG.

extensions/wikibase/module/scripts/schema-alignment.js
ayushrai206: Oh yes, fixing this asap!

wetneb:
@Abbe98 are you happy with the mitigation I proposed above? I would like to propose to @ayushrai206 that we go ahead with this, with the URL and host validation.

Abbe98:
I think passing the URL through `new URL()` is enough.

wetneb:
Makes sense. Closing this PR in favor of #6156 then

PR #6161, opened 2023-11-16
Fix DeleteProject NPE during e2e tests. Fixes #3479.

### LINKED ISSUE #3479: Not all test projects created during ui_testing are getting scraped

tfmorris:
Fixes #3479 

- refactor & DRY up tag management code
- add some tests for project related commands
- add tests for tag commands

### DIFF
--- a/main/tests/server/src/com/google/refine/commands/CommandTestBase.java
+++ b/main/tests/server/src/com/google/refine/commands/CommandTestBase.java
@@ -3,8 +3,8 @@ package com.google.refine.commands;
 
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
+import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertFalse;
-import static org.testng.Assert.assertTrue;
 
 import java.io.IOException;
 import java.io.PrintWriter;
@@ -13,9 +13,12 @@ import java.io.StringWriter;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
 import org.testng.annotations.BeforeMethod;
 
 import com.google.refine.RefineTest;
+import com.google.refine.util.ParsingUtilities;
 import com.google.refine.util.TestUtils;
 
 public class CommandTestBase extends RefineTest {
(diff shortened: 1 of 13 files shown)


main/tests/server/src/com/google/refine/commands/CommandTestBase.java
```diff
@@ -40,4 +43,13 @@ public void setUpRequestResponse() {
     protected void assertCSRFCheckFailed() {
         TestUtils.assertEqualsAsJson(writer.toString(), "{\"code\":\"error\",\"message\":\"Missing or invalid csrf_token parameter\"}");
     }
+
+    /**
+     * Convenience method to check that CSRF protection was NOT triggered
+     */
+    protected void assertErrorNotCSRF() {
+        String response = writer.toString();
+        assertTrue(response.contains("\"code\":\"error\""));
```
wetneb, line 52: This relies on the response containing JSON formatted in a specific way (without spaces around `:`), which I'd rather avoid. What motivates you not to use the helper above?

tfmorris: If "helper above" means `assertCSRFCheckFailed()`, it's testing for a different case. This helper is checking for `code: "error"` and `message: <anything but invalid csrf_token>`.

I can invest additional effort in coding it to protect against a potential future JSON formatting change if you think it's important.

wetneb: Ah right, I missed the negation. Yes I think it would be cleaner to parse the JSON and then access the fields of the resulting `JsonNode`.

main/tests/server/src/com/google/refine/commands/CommandTestBase.java
```diff
@@ -47,9 +50,11 @@ protected void assertCSRFCheckFailed() {
     /**
      * Convenience method to check that CSRF protection was NOT triggered
      */
-    protected void assertErrorNotCSRF() {
+    protected void assertErrorNotCSRF() throws JsonProcessingException {
         String response = writer.toString();
-        assertTrue(response.contains("\"code\":\"error\""));
+        ObjectMapper mapper = new ObjectMapper();
+        JsonNode node = mapper.readValue(response, JsonNode.class);
```
wetneb, line 56: Thank you! `ObjectMapper` objects are sort of expensive to instantiate so Jackson recommends to reuse a centrally-initialized instance. You could just do that by using `ParsingUtilities.mapper`.

main/tests/server/src/com/google/refine/commands/CommandTestBase.java
tfmorris: I've updated the test. I'm open to suggestions for a better method name or doc string to make its purpose clearer.

wetneb (approved):

PR #6162, opened 2023-11-16
Fix flaky Cypress test in cypress/e2e/create-project/preview_project.cy.js "Tests disabling of automatic preview" 

### LINKED ISSUE #6158: Flaky Cypress test in cypress/e2e/create-project/preview_project.cy.js "Tests disabling of automatic preview"
The test  "Tests disabling of automatic preview" in `cypress/e2e/create-project/preview_project.cy.js` seems to be flaky as it fails in two PRs with unrelated changes (#6108 and #6156):
* https://github.com/OpenRefine/OpenRefine/actions/runs/6878198239/job/18707359425
* https://github.com/OpenRefine/OpenRefine/actions/runs/6878198239/job/18707359425

elebitzero:
Fixes #6158
      
Changes:
  - Remove call to `waitForImportUpdate()`
  - Updated command description of `waitForImportUpdate()`
 
`waitForImportUpdate()` has the same problem as `waitForOrOperation()` which I have been trying to remove.  Both of those methods are waiting for a temporary loading/in-progress element to be visible and then not visible.

The problem is that the element may have been shown and hidden again before this code executes, causing it to wait forever until it times out.

The solution is to not focus on the transient loading indicators, but to just do `cy.get()` and assertions on the expected state after the operation completes. For long running operations,  the timeout on the `cy.get()` command may need to be increased if the operation could exceed the default timeout of 4 seconds.

I added @deprecated note on `waitForImportUpdate()` similar to what I added for `waitForOrOperation()`. The rest of the calls can be removed once we have fixed all the tests using them.

### DIFF
--- a/main/tests/cypress/cypress/e2e/create-project/preview_project.cy.js
+++ b/main/tests/cypress/cypress/e2e/create-project/preview_project.cy.js
@@ -181,24 +181,24 @@ describe(__filename, function () {
     // **Testing ignore feature with auto preview enabled** //
     cy.get('input[bind="ignoreInput"]').type('{backspace}1');
     cy.get('input[bind="ignoreCheckbox"]').check();
-    cy.waitForImportUpdate();
+
     // Look for automatic preview update
     cy.get('table.data-table tr').eq(1);
     cy.get('table.data-table tr').eq(1).should('to.contain', '01002');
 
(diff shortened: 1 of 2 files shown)


wetneb (approved):
Thanks a lot for taking care of this!

PR #6157, opened 2023-11-15
Replace custom formatRelativeDate with standard Intl.RelativeTimeFormat.format()

elebitzero:
Fixes #654

Changes proposed in this pull request:
- Replace custom implementation for the `formatRelativeDate()` used by Google Data importer with the standard browser API `Intl.RelativeTimeFormat`.
- Fixes https://github.com/OpenRefine/OpenRefine/issues/654
- Remove translations for 'core-util-enc/today', 'core-util-enc/yesterday', 'core-util-enc/days-ago', 'core-util-enc/weeks-ago', 'core-util-enc/months-ago', and 'core-util-enc/years-ago' which were buggy as not all the translations include the param '$1' so probably was not correct.

The bug talks about Open Project view. Maybe it used `formatRelativeDate()` at one time in 2013, but that view is now using Moment.js.  Moment.js should probably be replaced since it is deprecated.

https://github.com/OpenRefine/OpenRefine/blob/fb9cd795829296660630c2916bebb5265ad7a912/main/webapp/modules/core/scripts/index/open-project-ui.js#L152

SCREENSHOT (Existing)
image

SCREENSHOT (Intl.RelativeTimeFormat)
image

SCREENSHOT (es-MX language example)
image

PR #6169, opened 2023-11-17
Remove Datejs and Moment.js dependencies

elebitzero:
  - Change formatting of dates in Open Project table to use Intl.DateFormat instead of Moment.js.
   - Formatting the same except AM/PM is not shown because Intl.DateFormat doesn't support that when 24-hour military time is used.
  - Webapp package.json
    - Remove datejs-coolite (Unused after https://github.com/OpenRefine/OpenRefine/pull/6157).
    - Remove Moment.js (Deprecated https://momentjs.com/).
    
BEFORE (Moment.js - using 24HR time with AM/PM)
image

AFTER (Intl.DateFormat using 24HR time without AM/PM)
image

### DIFF
--- a/main/webapp/modules/core/scripts/index/open-project-ui.js
+++ b/main/webapp/modules/core/scripts/index/open-project-ui.js
@@ -134,6 +134,11 @@ Refine.OpenProjectUI.prototype._fetchProjects = function() {
     });
 };
 
+const dateOptions = { year: 'numeric', month: '2-digit', day: '2-digit' };
+const timeOptions = { hour: '2-digit', minute: '2-digit', hour12: true };
+const dateFormatter = new Intl.DateTimeFormat(navigator.language, dateOptions);
+const timeFormatter =  new Intl.DateTimeFormat(navigator.language, timeOptions);
+
 Refine.OpenProjectUI.prototype._renderProjects = function(data) {
   var self = this;
   var projects = [];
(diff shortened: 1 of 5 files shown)


main/webapp/modules/core/scripts/index/open-project-ui.js
```diff
@@ -134,6 +134,11 @@ Refine.OpenProjectUI.prototype._fetchProjects = function() {
     });
 };
 
+const dateOptions = { year: 'numeric', month: '2-digit', day: '2-digit' };
+const timeOptions = { hour: '2-digit', minute: '2-digit', hour12: true };
+const dateFormatter = new Intl.DateTimeFormat(navigator.language, dateOptions);
+const timeFormatter =  new Intl.DateTimeFormat(navigator.language, timeOptions);
```
tfmorris, line 140:
```suggestion
const options = { dateStyle: 'medium', timeStyle: 'medium' };
const dateFormatter = new Intl.DateTimeFormat(navigator.language, options);
```

main/webapp/modules/core/scripts/index/open-project-ui.js
```diff
@@ -149,7 +154,20 @@ Refine.OpenProjectUI.prototype._renderProjects = function(data) {
...
+      const findPart = (parts, type) => parts.find(part => part.type === type)?.value;
+
+      const year = findPart(dateParts, 'year');
+      const month = findPart(dateParts, 'month');
+      const day = findPart(dateParts, 'day');
+      const hour = findPart(timeParts, 'hour');
+      const minute = findPart(timeParts, 'minute');
+      const dayPeriod = findPart(timeParts, 'dayPeriod'); // AM or PM
+
+      project.date = `${year}-${month}-${day} ${hour}:${minute} ${dayPeriod}`;
```
tfmorris, line 170:
```suggestion
      project.date = dateFormatter.format(date);
```

I suggest we let the internationalization code do *all* of the formatting

tfmorris (changes requested):
This is great! The current implementation is clearly wrong. I never noticed that before.

Rather than switching to a different, slightly better, fixed format, I suggest that we punt the entire problem to the internationalization package. This will use 24 hour clock in the UK and AM/PM in the U.S., let the Arabic users read their dates & times from right to left, etc. It gets us out of the business of date & time formatting altogether.

elebitzero:
@tfmorris

I initially used the default formatting based on user's language. For 'en-US' that resulted in '11/15/2023, 09:43 PM'.

Because the timestamps are sortable, it makes sense to use a hardcoded datetime format which orders the elements from large to small, e.g. YYYY MM DD HH MM. While thinking about it further today, I realize that is why they used military time also.

So, I have updated the PR to keep the hardcoded formatting, except remove the AM/PM, which is not supported by Intl.DateFormat when 24-hour format is selected.


- '2023-11-15 12:50 PM'  current
- '11/15/2023, 09:43 PM'  (Intl.Dateformat default for 'en-US' language
- '2023-11-15 12:50' PR  

If you really think that the datetime should be formatted per user's language, then please file another issue so it can be discussed.

tfmorris:
OK, I created #6172. It looks like it couple be as simple as prepending a hidden span with a timestamp (or the current ISO 8601 date) which could form the basis of the sorting.

tfmorris (approved):

tfmorris:
I've put up #6173 which can be reviewed once this PR is merged.

tfmorris:
@Abbe98 since you gave a thumb's up to the internationalization idea, you might be interested in reviewing #6173. For some reason I'm unable to request you as a reviewer on the PR.

PR #6175, opened 2023-11-19
Defer initialization of Jython interpeter from startup. Fixes #6174

### LINKED ISSUE #6174: Defer Jython initialization from startup time
Currently Jython initialization takes 1/4 to 1/3 of the OpenRefine startup time (~2 seconds on my aging laptop) which seems wasteful since Jython is rarely used by users.

### Proposed solution
Defer initialization to the first time that Jython is needed.

tfmorris:
Fixes #6174

- Initialize Python interpreter when first needed, rather than unconditionally at startup.


wetneb (approved):
Makes a lot of sense to me! Tested and it works as expected on my side.

PR #6176, opened 2023-11-20
Switch HistoryEntry to Instant from OffsetDateTime(at UTC)

### LINKED ISSUE #3047: Fix project metadata dates

tfmorris:
Fixes #3047 This should have been included as part of the project metadata changes.

- Changes internal timestamp data type for HistoryEntry from OffsetDateTime (with a business logic enforced Zulu timezone) to the natural type Instant. 

Serialization is unchanged and no tests were changed.

### DIFF
--- a/main/src/com/google/refine/history/HistoryEntry.java
+++ b/main/src/com/google/refine/history/HistoryEntry.java
@@ -1,6 +1,6 @@
 /*
 
-Copyright 2010, Google Inc.
+Copyright 2010, 2023 Google Inc. & OpenRefine contributors
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Looks safe enough to me.

Issue #6179, opened 2023-11-22
Flaky tests in ListFacetTests.java


ThugJudy:
## Description of the bug
The following test cases 
- [com.google.refine.browsing.facets.ListFacetTests.serializeListFacet](https://github.com/ThugJudy/OpenRefine/blob/d6a00af85a96b389d49e878e6e12765db76ad6b0/main/tests/server/src/com/google/refine/browsing/facets/ListFacetTests.java#L96-L108)
- [com.google.refine.browsing.facets.ListFacetTests.testSelectedEmptyChoice](https://github.com/ThugJudy/OpenRefine/blob/d6a00af85a96b389d49e878e6e12765db76ad6b0/main/tests/server/src/com/google/refine/browsing/facets/ListFacetTests.java#L122-L133)
fail due to the non-deterministic behavior of the hashmap that is being implemented in the choices object.

https://github.com/OpenRefine/OpenRefine/blob/6afbd94843a5ce54fe40f565927235f2dc9b1605/main/src/com/google/refine/browsing/util/ExpressionNominalValueGrouper.java#L80

HashMap does not maintain the order and hence results in the test to be flaky

I found and confirmed the flaky behavior using an open-source research tool [NonDex](https://github.com/TestingResearchIllinois/NonDex), which shuffles implementations of nondeterminism operations.

### To Reproduce
The following command can be used to reproduce assertion failures and verify the fix:
```
mvn -pl main edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=com.google.refine.commands.recon.GuessTypesOfColumnCommandTests#testGuessTypes
```
mvn -pl main edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=com.google.refine.browsing.facets.ListFacetTests#serializeListFacet
### Current Results
The test fails upon running the nondex command due to the shuffling of the JSON elements inside the query.

### Expected Behavior
The test should pass consistently regardless of the shuffling of the elements.

Screenshots
Versions
Operating System: MacOS Venture Version 13.4.1 (22F82)cOs
JRE or JDK Version: "11.0.20"

PR #6180, opened 2023-11-22
Fix flaky tests

ThugJudy:
Fixes [6179](https://github.com/OpenRefine/OpenRefine/issues/6179)

## Description of the fix

As mentioned in the Issue, the ordering is not guaranteed in hashmap, so I sort the choices inside the test so that the order remains constant.

### DIFF
--- a/main/tests/server/src/com/google/refine/browsing/facets/ListFacetTests.java
+++ b/main/tests/server/src/com/google/refine/browsing/facets/ListFacetTests.java
@@ -28,9 +28,14 @@
 package com.google.refine.browsing.facets;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
 
 import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.testng.annotations.Test;
 
 import com.google.refine.RefineTest;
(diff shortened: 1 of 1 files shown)


wetneb (changes requested):
Thanks for tackling this! I agree that the order of facet choices in the JSON representation of the facet is unspecified.

main/tests/server/src/com/google/refine/browsing/facets/ListFacetTests.java
```diff
@@ -126,8 +128,9 @@ public void testSelectedEmptyChoice() throws IOException {
         Engine engine = new Engine(project);
 
         ListFacetConfig facetConfig = ParsingUtilities.mapper.readValue(jsonConfig, ListFacetConfig.class);
-        Facet facet = facetConfig.apply(project);
+        ListFacet facet = (ListFacet) facetConfig.apply(project);
         facet.computeChoices(project, engine.getAllFilteredRows());
+        facet.getChoices().sort(Comparator.comparing(choice -> String.valueOf(choice.decoratedValue.label)));
```
wetneb, line 133: In general, `facet.getChoices()` might return an immutable list: as a user of this class, I wouldn't assume I can sort the results in place. So I think it would be a cleaner to take a copy of the JSON the facet is serialized to, and do the sorting on the JSON list. After that, you'd be able to assert that the resulting JSON is equal to the expected normalized JSON.

ThugJudy: Made the changes as per the suggestions. Let me know if you have any questions.

wetneb (approved):
Thanks a lot!

PR #6184, opened 2023-11-24
fix: reconciliation documentation link display type

Abbe98:
`inline` is the default type for <a> elements and `block` causes the links to be stacked vertically.

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
+++ b/main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
@@ -103,7 +103,7 @@ ReconStandardServicePanel.prototype._constructUI = function() {
   if(this._service.documentation) {
     this._elmts.documentationLink.attr("href", this._service.documentation);
     // Show the documentation link if documentation is available
-    this._elmts.documentationLink.css("display", "block");
+    this._elmts.documentationLink.css("display", "inline");
   } 
   
   this._elmts.againstType.on('change', function() {


wetneb:
I think stacking the links vertically was intentional, no? Could you maybe add some screenshots of how it looks before and after?

Abbe98:
I don't think it is intentional given how it pushes the main content further down.

Vertically:
Screenshot 2023-11-30 at 14-49-34 My Project - OpenRefine

Horizontally:
Screenshot 2023-11-30 at 14-49-19 My Project - OpenRefine

tfmorris:
I don't love either version. The horizontal layout doesn't really parse as two separate links to my eye and the vertical version adds unnecessary whitespace (when there's a ton of space available in the header above).

magdmartin:
Can the @OpenRefine/designers team help with?

wetneb:
I'd be in favor of removing the "Access service API" link anyway, since it's not designed to be viewed by users directly. We will be displaying the URLs (#5878) with #6118. Then the "Service documentation" link would be alone and it would solve the dilemma.

wetneb:
@Abbe98 would it make sense for you to simply remove the service API link now that #6118 is merged (meaning that the link is displayed earlier)?

Abbe98:
@wetneb no I consider that covering a rather different usecase and I think such a change deserves its own thread independently from this.

wetneb:
Okay, so here is a PR for that change: #6284.

wetneb:
@Abbe98 now that we removed the API URL in #6284, this change will not have an effect anymore, although the underlying problem is still there (although it only pushes content down as much as the API URL link used to).

Do you want to revisit this (moving the documentation link elsewhere, perhaps) or should we close the PR?

wetneb:
Closing per above, feel free to reopen to move the link somewhere else.

Abbe98:
The display type is unrelated to the above and should be fixed anyway.

wetneb:
Sorry about that then! As far as I can tell the PR does not make a visual difference anymore at least. Can you expand on why the current display type is wrong and should be changed?

Abbe98:
The browser default for `a` elements is `inline`, suddenly switching it to `block` might cause side-effects down the road or just cause pain for someone trying to debug the layout.

wetneb:
Ok, if the motivation is to clean this up then I'd suggest migrating to using jQuery's `.hide()` and `.show()` instead of setting inline CSS explicitly.

Abbe98:
Yes that would be the better approach with the same effect.

wetneb (approved):

PR #6186, opened 2023-11-24
project client: get the project-id from the URL rather than from Velocity injection

Abbe98:
Part of a downstream attempt to nuke Velocity usage.


wetneb (commented):
I think it's a worthy goal! Not approving immediately to let others chime in.

tfmorris:
No objections from my point of view, although I still think it'd be worthwhile having PRs attached to an issue whether that be for migrating from Velocity to something else or just removing it or ...

Also, weirdly, I didn't get notified until @wetneb's post, so I'm wondering if @Abbe98 has me blocked on Github.

tfmorris (approved):

wetneb:
For the record: I have followed up with @tfmorris and @Abbe98 (privately and independently) about this blocking. This is not a good sign for the social health of the project and I am determined to address this problem once and for all.

PR #6191, opened 2023-11-26
Improve error messages in reconciliation operation#6111

### LINKED ISSUE #6111: Improve error messages in reconciliation operation
The errors that are shown to the user when reconciliation fail are wordy and not very helpful to understand the source of the error.
Currently, the error message contains the JSON serialization of the entire batch. This was perhaps helpful when those error messages were written to the server logs, as they were necessary to figure out which part of the dataset was responsible for the error, but now that those errors are displayed in the grid, the context is already clear from the location of the error.

### Proposed solution

Review all the places where `recon.error` is set and make sure a helpful error message is set.

For instance:
* if we cannot connect to the reconciliation service, I would expect a message such as `Connecting to my-recon-service.com failed`
* if the reconciliation service returned an HTTP error, I would expect a messages such as `The reconciliation service returned HTTP error 403`
* if the reconciliation service returns an invalid JSON response, we could have something like `The reconciliation service returned an invalid response: expected ':' after field "queries"`

### Alternatives considered

When it is worth adding more information to the error, but that information can potentially take up a lot of space, then perhaps we could still use the server logs for that.

### Additional context
Follow up to #6006.

ayushrai206:
Fixes #6111

- Improved the error messages at the places mentioned in the pull request with a message that is concise as well as easily understood by the user
- Introduced some tests to check error due to reasons such as connection to service failed and Domain name not found 
- Also improved the error messages for the errors of the newly added tests.


main/src/com/google/refine/model/recon/StandardReconConfig.java
```diff
@@ -460,48 +455,58 @@ public List<Recon> batchRecon(List<ReconJob> jobs, long historyEntryID) {
...
-                    }
+                        recon.error = "The service returned a JSON response without \"result\" field for query " + key;
 
-                    if (recon != null) {
-                        recon.service = service;
                     }
-                    recons.add(recon);
+                } else {
+                    recon = new Recon(historyEntryID, identifierSpace, schemaSpace);
+                    recon.error = "The service returned a JSON response without \"key\" field ";
```
wetneb, line 495: This error message is a bit misleading: we are not expecting a `"key"` field, but rather a field whose name is equal to the contents of the `key` Java variable. So it should rather be something like:
```suggestion
                    recon.error = "The service returned a JSON response without \"" + key + "\" field ";
```

main/src/com/google/refine/operations/recon/ReconOperation.java
```diff
@@ -298,6 +298,8 @@ public void run() {
                             // TODO add EvalError instead? That is not so convenient
                             // for users because they would lose the cell contents.
                             // Better leave the cell unreconciled so they can be reconciled again later.
+                            Cell oldCell = entry.cell;
```
wetneb, line 301: This variable declaration does not do anything and can be removed

main/tests/server/src/com/google/refine/model/recon/StandardReconConfigTests.java
```diff
@@ -564,11 +566,94 @@ public void BatchReconTestError() throws Exception {
...
+        StandardReconConfig config = StandardReconConfig.reconstruct(configJson);
+        StandardReconConfig.StandardReconJob job = new StandardReconConfig.StandardReconJob();
+        job.text = "david lynch";
+        job.code = "{\"query\":\"david lynch\",\"type\":\"Q11424\",\"properties\":[{\"pid\":\"P57\",\"v\":\"david lynch\"}],\"type_strict\":\"should\"}";
+        List<ReconJob> jobList = new ArrayList<ReconJob>();
+        jobList.add(job);
+
+        List<Recon> returnReconList = config.batchRecon(jobList, 1000000000);
+        // assertions
+        // assertEquals(returnReconList.get(0).error, url.host() + ": nodename nor servname provided, or not known");
```
wetneb, line 651: Here we can have a comment explaining that the error message is unstable and system-dependent, so we are not asserting for its exact contents.

main/src/com/google/refine/operations/recon/ReconOperation.java
ayushrai206: I think it is being used in the next line,should we use entry.cell directly?

wetneb: It looks like it's not there anymore, great.

wetneb (approved):

PR #6194, opened 2023-11-28
Fix flaky Cypress test: Copy reconciliation data from species to species_copy

elebitzero:
The copy reconciliation data test copies a column, then counts 'Choose new match' in all TDs, including the original column.  On slower computers the count is 4.  On faster computers the count is 8.

- Fix flaky Cypress test that fails on faster computers.

- Change to first wait for 'duplicated_column' to show up, then count 'Choose new match' occurrences only in the last column.


wetneb (approved):
Thanks a lot!

PR #6195, opened 2023-11-30
Added 10s timeout for Jython tests to fix flakiness

elebitzero:
The Cypress tests related to Jython have become flaky due to the recent change to [using lazy initialization for the Jython interpreter](#6175), resulting in the first server request related to Jython sometimes exceeding the default 4 second timeout.

image
Ref: https://cloud.cypress.io/projects/s5du3k/analytics/flaky-tests

Changes:
- Added 10 second timeout for Jython related tests, so that the `cy.get()` calls will wait up to 10 seconds.


tfmorris (approved):
Thanks for looking into this. I should have thought to dig further into the newly flaky test, but didn't realize they were Jython related. Nominal initialization time is 2-3 seconds, so the 10 second timeout should have plenty of headroom.

PR #6198, opened 2023-12-01
Made text selectable in alert dialog

### LINKED ISSUE #6187: Text in error dialogs should be selectable
In the error dialog shown in #5444 none of the text is selectable, making it difficult to quote or reference in error reports.

I think we moving away from browser alerts has been mentioned in other issues, which would probably make it easier to fix this.

### Current Results
Text is not selectable.

### Expected Behavior
Error text is selectable

### Screenshots
See #5444 

### Versions
 - Operating System: macOS X
 - Browser Version: Chrome Version 119.0.6045.105 (Official Build) (x86_64)
 - OpenRefine: 3.6.2

elebitzero:
Fixes #6187

Changes:
* Add `DialogSystem.alert()` to use in place of `alert()`.
* Change all uses of `alert()` and `window.alert()` in gdata extension to use `DialogSystem.alert()`.

Notes:
* The text displayed in `alert()` in Chromium browsers is not selectable if the text contains a newline character.

Before:
image

After:
image

### DIFF
--- a/extensions/gdata/module/authorized.vt
+++ b/extensions/gdata/module/authorized.vt
@@ -51,7 +51,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
       try {
         callback();
       } catch (e) {
-        alert(e.message);
+        DialogSystem.alert(e.message);
       }
     } else {
       w.close();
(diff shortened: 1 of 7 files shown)


tfmorris (approved):
Looks great, thank you! I added a couple of missing translations. Please double check that I didn't mess anything up. I verified that the text is selectable in Chrome.

We shouldn't have been using `alert()` in the first place, so this will be a great first step to getting us a proper error dialog.

PR #6208, opened 2023-12-04
Fix for workflows running on forks

elebitzero:
I am seeing the snapshot release GitHub action workflow being triggered for each cloned repo when a user syncs their master branch with OpenRefine's master branch. The snapshot build fails on these runs and the e2e tests are published to Cypress Dashboard as duplicate runs.

Changes:
- Don't run the following workflows on forks:
  - snapshot_release.yml - (Should not be run on forks).
  - codeql_java.yml - (Shouldn't be run on forks as these already run on OR's master branch when you commit).
  - codeql_javascript.yml - (Shouldn't be run on forks as these already run on OR's master branch when you commit).
  - unassign_issues.yml - (Should not be run on forks as well).
  
The following screenshot shows 4 workflow runs on Cypress Dashboard that have the same commit from when developers sync their master with the latest OR master branch:

image

The solution is to check `github.repository == 'OpenRefine/OpenRefine'` for each job in any workflows that should only be run in the main repo.

Example from Microsoft's TypeScript repo:
image

### DIFF
--- a/.github/workflows/codeql_java.yml
+++ b/.github/workflows/codeql_java.yml
@@ -12,6 +12,7 @@ on:
 
 jobs:
   analyze:
+    if: github.repository == 'OpenRefine/OpenRefine'
     name: Analyze Java
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/codeql_javascript.yml
+++ b/.github/workflows/codeql_javascript.yml
@@ -12,6 +12,7 @@ on:
 
 jobs:
   analyze:
+    if: github.repository == 'OpenRefine/OpenRefine'
     name: Analyze JavaScript
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/snapshot_release.yml
+++ b/.github/workflows/snapshot_release.yml
@@ -12,6 +12,7 @@ on:
 
 jobs:
   prepare_e2e_test_matrix:
+    if: github.repository == 'OpenRefine/OpenRefine'
     runs-on: ubuntu-latest
     outputs:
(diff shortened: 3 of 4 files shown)


elebitzero:
Sorry, I was in a hurry and didn't do well explaining the PR. I have updated the summary with more explanation.

> I think it makes a lot of sense not to publish Cypress runs made on forks to our central dashboard. But isn't it useful for developers to have the test suite run on their forks as well?

The Cypress dashboard issue is just a side effect.  The E2E Cypress tests are run when snapshot_release.yml is run on a push to "master" branch. In forks this gets triggered when a user syncs the fork's master branch with the main repo's master branch.

The snapshot workflow should not be run in forks.

What you are suggesting sounds like something else like an e2e_test.yml workflow that a user could manually trigger on any branch in their fork, but I'm not sure how useful that would be. If you are working on Cypress tests, you will have to figure out how to run Cypress locally. Then, you will have already tested your changes, and when you submit the PR in the main repo, pull_request_e2e.yml will run the Cypress tests.

tfmorris (approved):
These all look fine to me. There's no question in my mind that `snapshot_release.yml` should not be running in forked repos. We could create a new feature request for an action which runs e2e tests on non-master branches in forks, but I'm going to have it disabled in my fork.

I'm approving, but suggest letting @wetneb weigh back in before merging.

wetneb (approved):
Ok - I think it would still make sense to run the Cypress tests on forks because it takes a lot of time to run those locally, but I guess people can re-activate that on their side if they want.

PR #6227, opened 2023-12-07
Widen regex of allowed characters in Commons filenames.

### LINKED ISSUE #5656: Wikimedia Commons Wikibase not detecting non-Latin script characters

wetneb:
I can't find authoritative documentation about this so this is pretty much example driven. To reflect that, I have lowered the severity of the check so that users are not prevented from uploading if the check incorrectly fails.

Closes #5656.

### DIFF
--- a/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizer.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/FileNameScrutinizer.java
@@ -22,7 +22,7 @@ public class FileNameScrutinizer extends EditScrutinizer {
     // see https://commons.wikimedia.org/wiki/Commons:File_naming
     public static final int maxFileNameBytes = 240;
     public static final Pattern forbiddenFileNameChars = Pattern.compile(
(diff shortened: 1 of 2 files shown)


elebitzero (approved):
Looks good 👍

PR #6239, opened 2023-12-13
Make UI update when matching cell

### LINKED ISSUE #6236: 'Match this cell' button not working in preview
'Match this cell' button in the preview overlay during reconciliation workflow doesn't do anything, the single tick box in the column works though. 

### To Reproduce
1. Reconcile some data to Wikidata, or Wikibase, or some other service (where you know the data might have multiple items of a similar name).
2. Try selecting a match from the preview overlay.
3. Check if a match is produced, or if you need to use the tick box instead.

### Current Results
Currently when tested with several services (Wikidata, Wikibase, Getty ULAN, etc) only the notification that the item was matched appears, but a match is not performed. 

### Expected Behavior
The item should be matched, just like when ticking the single tick box. 

### Screenshots
Step 1: Identifying the correct candidate

Image

Step 2: Clicking the 'Match this cell' button and getting a positive notification, but no resulting match
Image

Step 3: Clicking the tick box instead results in a match

Image

Image

### Versions
 - Operating System: MacOs Monterey 12.7.1
 - Browser Version: Chrome 119.0.6045.159
 - JRE or JDK Version: 17.0.4.1
 - OpenRefine: Version 3.7.6

### Datasets
N/A

### Additional context
N/A

elebitzero:
Fixes #6236

I was able to reproduce the issue by clicking the 'Match this cell' button in the preview overlay. Before the change, refreshing the page after clicking 'Match this cell' updates that cell's contents.

The UI just needed to update when clicking the 'Match this cell' button.

After:
match this cell ui fix

- Make UI update when clicking the 'Match this cell' button.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-renderers/recon-renderer.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-renderers/recon-renderer.js
@@ -360,7 +360,7 @@ class ReconCellRenderer {
       command,
       params,
       bodyParams,
-      { columnStatsChanged: columnStatsChanged },
+      { cellsChanged: true, columnStatsChanged: columnStatsChanged },
       {
         onDone: function(o) {
           if (o.cell.r) {


tfmorris (approved):
LGTM

PR #6244, opened 2023-12-15
Improves the display of reconciliation results

### LINKED ISSUE #6154: Improve display of reconciliation results
### Problem Description
During user interviews conducted in September on the redesign of the reconciliation dialog feature, a common pain point identified by reconciliation was the challenge of dealing with a large number of results displayed after selecting a reconciliation service. The feedback highlighted issues with both the accuracy and the overwhelming nature of the displayed results. Users expressed the need for improved visibility and control as they often had to sort through numerous irrelevant results and this also causes a distatsteful UI by expanding cells thereby creating too much whitespace.

### Proposed solution
To address the identified problem, I propose an enhancement to the reconciliation feature in OpenRefine. The suggested solution involves introducing a more user-friendly way for managing the display of reconciliation candidates. 
- We start of displaying limited number of resuts, say limited to 2 by defaut.
- A "See More" button: Users can reveal additional candidates beyond the initially specified limit.
- A "See less" buton (link) to reverse the previous action and return back to the default display of 2 results.

This solution aims to provide users with greater control over the display of results, making the reconciliation process more efficient and user-friendly.

### Alternatives considered
An alternative is a straightforward "truncate results" option added to the dropdwon options of each reconciled cell. On click, a pop-up containing an input field appears where users can input the number of results they want displayed.

### Additional context
The proposed solution is a response to direct user feedback about the challenges faced when reconciling data, particularly in scenarios where a large number of candidates are presented. By allowing users to control the visibility of results, the proposed feature seeks to enhance the overall user experience, improving accuracy, and reducing the time and effort required for reconciliation tasks.

**Proposed defaukt display of results**
image

When user clicks on "See more"
image

**Alternative proposed**

image

image

ayushrai206:
Fixes #6154 

- Now, the no. of matched candidates that are visible after reconciliation is performed is set to a default of 3(which can be changed easily by some minor code changes as I have introduced a visibleCandidate variable in the code)
- The user can see all the matched candidates by clicking on see more option, and go back to the previous no. by clicking on the see less option.
- The see more and see less option would only be shown to the user if the no. of matched candidates are greater than 3(visibleCandidates).


wetneb (commented):
Very nice!
I think it would be worth having a Cypress test for this. Luckily, we already have a sample test project which has more than three reconciliation candidates in many cells (stored in `main/tests/cypress/cypress/fixtures/reconciled-project-no-automatch.zip`).
You can take inspiration from the `project/grid/column/reconcile/actions/match_to_best_candidate.cy.js` test to load it.

ayushrai206:
The cypress tests fail weirdly, @wetneb can you look into this please-
1)After inspecting the tests, I found out that the test expects only 1 recon candidate in the 1st cell and 5 recon candidates in the last cell(but actually the no. of recon candidates is 6 in the first and last cells), while 6 candidates for 2nd to 5th candidates work perfectly fine.

wetneb (commented):
Very nice!

I think the failure is due to the fact that because you are clicking so many "See more" links throughout the project, this causes the project view to scroll down (so that the last such link is visible). Then, you are making an assertion about visible reconciliation candidates in the first cell, but this cell is no longer fully visible, so not all of its candidates are visible.

I would propose you only do your test within the first reconciled cell and ignore all others. The other cells don't really have any particular aspect that makes them different from the first one, so I don't think those assertions add a lot of value to the test suite. It will also make the test faster.

wetneb (changes requested):
I tried it out interactively and there is a couple of problems:

* When clicking on "See less", I get the error: "Uncaught ReferenceError: visibleCandidate is not defined".
* The font sizes for "See more" and "Search for match" are inconsistent, I would expect them to be the same as proposed in @Lydiaofficial's design
* The "Search for match" link has changed position in matched cells, I would expect it to remain in its original position

screenshot illustrating the problems

wetneb (approved):
This works as expected now and looks great to me, thanks! I would merge it soon unless people see more things to improve.

PR #6260, opened 2023-12-21
Qualify selectors with metadata dialog ID. Fixes #6259

### LINKED ISSUE #6259: Cypress tests fail if a project name contains the word "Subject"
If you run the e2e tests using a data directory containing a project the word "Subject" (or "Tags" or any other metadata test target), the project_metadata tests fail.

### To Reproduce
1. Start OpenRefine server & create project containing the word "Subject"
2. Start Cypress and run the `project_metadata` e2e test

### Current Results
Test fails

### Expected Behavior
Test passes

### Additional context
Although the tests should nominally be run in a clean directory, this is bound to cause trouble in the future (and also shows the test isn't strictly qualified enough).

The facet by choice count test is also flaky on my system, but, annoyingly, only when run with the full test suite, not when run in isolation. Perhaps it's on the edge of a timing boundary.

tfmorris:
Also mark flaky test

Fixes #6259

### DIFF
--- a/main/tests/cypress/cypress/e2e/project/grid/column/facet/facets.cy.js
+++ b/main/tests/cypress/cypress/e2e/project/grid/column/facet/facets.cy.js
@@ -330,7 +330,7 @@ describe(__filename, function () {
       .find('.facet-body-controls')
       .contains('Facet by choice counts')
       .click();
-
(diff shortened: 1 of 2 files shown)


elebitzero (approved):
Looks good 👍

PR #6267, opened 2023-12-29
delays the disappearance of preview by 250 milliseconds

### LINKED ISSUE #3144: Unhelpful placement of reconciliation previews
Sometimes, previews of reconciliation candidates appear in locations which hide parts of the reconciliation candidate itself or its matching buttons:

### To Reproduce
1. First, reconcile some cells
2. Then, reduce the size of the window
3. Finally, hover the reconciliation candidate

### Current Results

image

### Expected Behavior

The reconciliation preview should be placed somewhere which does not hide the reconciliation candidate itself.

### Versions
 - Operating System: Debian
 - Browser Version: Firefox 68
 - JRE or JDK Version: irrelevant
 - OpenRefine: 3.4 beta2

ayushrai206:
Fixes #3144 

- Now the preview does not disappear quickly and gives enough time to the user to select any option in the preview if they want.
- The preview stays on the screen for 250ms which can be changed easily as a previewDelay variable is added.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/cell-renderers/recon-renderer.js
+++ b/main/webapp/modules/core/scripts/views/data-table/cell-renderers/recon-renderer.js
@@ -443,25 +443,39 @@ class ReconCellRenderer {
    * Sets up a preview widget to appear when hovering the given DOM element.
    */
   previewOnHover(service, candidate, hoverElement, coreElement, showActions, rowIndex, cellIndex, cell, cellUI) {
-      var self = this;
-      var preview = null;
-      if ((service) && (service.preview)) {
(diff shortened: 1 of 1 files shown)


wetneb:
Thanks a lot! I have not tested it yet, but my understanding of the intention behind @Abbe98's proposal to introduce a delay is that the delay in itself would not fix the issue: it would just be that having a delay makes it possible to position the preview further away from the reconciliation candidate, in a way which does not hide buttons. So I think it would be worth updating this positioning too.

wetneb:
@ayushrai206 are you still working on this or can we close this PR?

PR #6284, opened 2024-01-05
Remove 'Access service API' link from reconciliation dialog

wetneb:
Follow-up to #6118 suggested by #6184.

Because the accessing the API directly with the browser is rarely useful for users, I think we don't need to show this link in this dialog. The full URL is visible in the previous step (service selection), which enables users to copy the link (for instance to share it with someone else who would like to install the same service).

@Abbe98 note that the problem you want to solve in #6184 (the dialog being pushed down by the documentation link when it is present) is still here - it might make sense to put this documentation link in a totally different place so that this does not happen. But then the solution will be different from the one you proposed in #6184.

### Before
image

### After
image

### Before
image

### After
image

### DIFF
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -425,7 +425,6 @@
     "core-project/continue-remaining": "Continue with the remaining operations?",
     "core-project/undo": "Undo",
     "core-recon/add-std-srv": "Add standard service",
-    "core-recon/access-service": "Access service API",
     "core-recon/service-documentation":"Service documentation",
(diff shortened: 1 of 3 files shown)


elebitzero (approved):
I tried out the change, it looks good to me 👍

elebitzero (changes requested):
Hi,
 
  Thank you for the contribution.  The smooth scroll and center model look good.
 
  For the icons, I didn't want to use ChatGPT icons since I didn't know where they got them, so I was just using the Heroicons for now.
 
  Please update the PR not including the icons, and I'll merge it.   Otherwise, I'll incorporate the changes at some point.
 
  Thanks!

PR #6285, opened 2024-01-05
Improve how to indicate the data source following data extension#5130

### LINKED ISSUE #5130: Improve how to indicate the data source following data extension
Currently a default OpenRefine-to-Wikibase workflow works well if users only reconcile data against a single service, e.g. Wikidata, and then proceed towards schema building and data upload with this single service. However, in reality - with the introduction of the possibility to reconcile against and upload data edits to any arbitrary Wikibase (and Wikimedia Commons, too) - users may wish to work across several services at once. For example, enriching (or extending) a dataset from one Wikibase with data from another, in addition to also including data from various other authority control sources (e.g. GND, Getty’s ULAN, VIAF, etc). Currently, however, once a data extension operation is carried out — there is no easy visual cue to a user to specify what source each column has been pulled from and - depending on the data type - also reconciled against, leading easily to making mistakes during schema building when users only see column titles and a green bar.

If data is pulled from Wikidata via reconciled values - the data can be shown to be reconciled to Wikidata with a little logo as per suggestion in this issue: #4824, but there is no way to tell users the **source** was **also** Wikidata. For that users proposed that prefixes are added to the column headers.

### Proposed solution
Add prefixes to column names to indicate where the data was pulled from. E.g. wmc: for Wikimedia Commons; wd: for Wikidata; wb: for any other Wikibase, etc. 

Example where some artworks are reconciled with Wikidata and then the 'collection' property values are pulled from Wikidata via data extension:
Screenshot 2022-07-30 at 12 04 19

Example with data pulled from Commons:
Screenshot 2022-07-24 at 00 28 17

### Alternatives considered
Adding small service logos next to the reconciled status green bar, but that solution does not always translate 1:1 with the data extension use case. 

### Additional context
This feature request is most useful during schema building, so the prefixes should naturally be carried over to the drag / drop elements in that interface that contain the column names.

ayushrai206:
Fixes #5130

- Adds a variable sourceReconConfig that contains the information about the reconconfig of the column from which data is extended from
- Also updated the tests to test for sourceReconConfig
- Finally, a tooltip is added to the new column which contains the extended data.


main/webapp/modules/core/scripts/views/data-table/menu-edit-column.js
```diff
@@ -185,7 +185,7 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
             "extend-data", 
             {
               baseColumnName: column.name,
-	      endpoint: endpoint,
+	          endpoint: endpoint,
```
wetneb, line 188:
```suggestion
	      endpoint: endpoint,
```

main/webapp/modules/core/scripts/views/data-table/menu-edit-column.js
```diff
@@ -596,6 +596,7 @@ DataTableColumnHeaderUI.extendMenu(function(column, columnHeaderUI, menu) {
         id: "core/add-column-by-reconciliation",
         label: $.i18n('core-views/add-col-recon-val'),
         click: doAddColumnByReconciliation
+
```
wetneb, line 599:
```suggestion
```

main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
```diff
@@ -94,6 +94,18 @@ DataTableColumnHeaderUI.prototype._render = function() {
...
   }
+  if("sourceReconConfig" in this._column) {
+    if(this._column.sourceReconConfig.service){
+     var service = ReconciliationManager.getServiceFromUrl(this._column.sourceReconConfig.service);
+     var serviceName;
+     if(service) {
+       serviceName=service.name;
+     }
+     if(serviceName){
+      elmts.nameContainer.attr("title","Data extended from "+service.name);
```
wetneb, line 105: We'll need to localize the tooltip with `$.i18n`.

wetneb (commented):
It looks great to me! I have pinged people on the issue to make sure everyone is happy with the phrasing of the tooltip. Once that is cleared (and the localization issue below), I would say we can merge.

wetneb (approved):

PR #6320, opened 2024-01-22
The reconciliation errors are now selectable by facets

### LINKED ISSUE #6232: Make reconciliation errors selectable by facets
Consider a reconciliation process where some cells could not reconcile properly, resulting in errors stored in cells.
As a user, I want to:
* be notified that those errors happened, without manually scrolling through the dataset to find red text
* be able to easily select all the rows where errors happened, to either fix them manually or re-run reconciliation on this subset only

### Proposed solution

Adapt the reconciliation judgment facet to have a new category, "error", corresponding to the cells with a non-null reconciliation error.

This could be done either by changing the expression that defines the facet (such as `if(isNonBlank(cell.recon.error), "error", forNonBlank(cell.recon.judgment, v, v, if(isNonBlank(value), "(unreconciled)", "(blank)")))`),
or by introducing a new judgment value in the backend. I think the latter would be cleaner.

### Alternatives considered

Introduce another facet for that? It seems a bit overkill to me, and would take up more screen space.

### Current screenshot

Currently, cells with reconciliation errors have status "none":
image

After this change, it would look like this instead:
image

ayushrai206:
Fixes #6232 

- The reconciliation errors can now be selected through facets 
- A test is also added in the StandardReconConfigTests file that checks that the judgment has been set to error in case of an error.


main/src/com/google/refine/commands/recon/ReconJudgeOneCellCommand.java
```diff
@@ -190,10 +190,19 @@ protected HistoryEntry createHistoryEntry(long historyEntryID) throws Exception
...
+                }
                 newCell.recon.match = null;
-
                 description = "Discard recon judgment for " + cellDescription;
+
+            } else if (judgment == Judgment.Error) {
+                newCell.recon.judgment = Recon.Judgment.Error;
+                newCell.recon.match = null;
+
+                description = "Errors in recon judgment for " + cellDescription;
```
wetneb, line 205: This case would correspond to the user manually setting the recon object to an error state, but I don't think that makes sense. Errors are meant to be generated by the reconciliation process itself, not added manually by users.
So I would say we should throw an exception in this case (such as `InvalidArgumentException`).

main/src/com/google/refine/model/Recon.java
```diff
@@ -69,7 +69,8 @@ static public enum Judgment {
         @JsonProperty("none")
         None, @JsonProperty("matched")
         Matched, @JsonProperty("new")
-        New
+        New, @JsonProperty("errors")
```
wetneb, line 72: I'd adjust the JSON serialization of this field to be just `error`, not `errors`
```suggestion
        New, @JsonProperty("error")
```

main/src/com/google/refine/model/ReconStats.java
```diff
@@ -51,21 +51,26 @@ public class ReconStats {
...
+    @JsonProperty("errorTopics")
+    final public int errorsTopics;
 
     @JsonCreator
     public ReconStats(
             @JsonProperty("nonBlanks") int nonBlanks,
             @JsonProperty("newTopics") int newTopics,
-            @JsonProperty("matchedTopics") int matchedTopics) {
+            @JsonProperty("matchedTopics") int matchedTopics,
+            @JsonProperty("errorsTopics") int errorsTopics) {
```
wetneb, line 62: To be consistent with the `@JsonProperty` annotation above:
```suggestion
            @JsonProperty("errorTopics") int errorsTopics) {
```

wetneb (changes requested):
Nice! Getting there :)

main/src/com/google/refine/commands/recon/ReconJudgeOneCellCommand.java
```diff
@@ -199,10 +199,9 @@ protected HistoryEntry createHistoryEntry(long historyEntryID) throws Exception
                 description = "Discard recon judgment for " + cellDescription;
 
             } else if (judgment == Judgment.Error) {
-                newCell.recon.judgment = Recon.Judgment.Error;
-                newCell.recon.match = null;
 
-                description = "Errors in recon judgment for " + cellDescription;
+                throw new Exception("Invalid Argument Exception");
```
wetneb, line 203:
```suggestion
                throw new InvalidArgumentException("Cannot manually set judgment to 'error'");
```

main/src/com/google/refine/operations/recon/ReconJudgeSimilarCellsOperation.java
```diff
@@ -222,10 +222,10 @@ public boolean visit(Project project, int rowIndex, Row row) {
                                     recon.judgment = Recon.Judgment.New;
                                     recon.match = null;
                                 } else if (_judgment == Judgment.None) {
-
-                                    recon.judgment = recon.error == null ? Recon.Judgment.None : Recon.Judgment.Error;
-
+                                    recon.judgment = Recon.Judgment.None;
```
wetneb, line 225: I think it is still worth having this logic here, so that if the user discards a match, we still set the judgment to "error" if the recon's error is not null. It's a different case, independent from the `else if` block you added below
```suggestion
                                    recon.judgment = recon.error == null ? Recon.Judgment.None : Recon.Judgment.Error;
```

wetneb (approved):

PR #6327, opened 2024-01-27
Fix wiring of event listeners in recon dialog

### LINKED ISSUE #6326: Recon dialog: no auto-completion for types and properties until types have been guessed
This is a rather subtle bug, but which might cause real problems for slow reconciliation services.

### To Reproduce
1. First, open the reconciliation dialog on a column
2. Before the type guessing completes, try to input a property in the right-hand panel

### Current Results

Auto-completion does not happen

### Expected Behavior

Auto-completion should be available already

### Versions
 - master branch

wetneb:
Closes #6326.

This does the wiring of event listeners as soon as we set up the dialog, not just once we get the types back.


ayushrai206 (approved):
Useful amend, go for it!

PR #6342, opened 2024-02-02
Improve GREL operators. Fixes #6341 Closes #6340

### LINKED ISSUE #6340: GREL comparison operators should support strings
Currently the only GREL comparison operators supported for strings are "==" and "!=" and those only do comparisons of Unicode code points, so e.g. composed and non-composed versions of É aren't equal to each other. The other comparison operators ">", "<", ">=", and "<=", aren't supported at all.

### Proposed solution
All of the above operators should be supported and the comparisons should be done in a locale sensitive manner with canonical decomposition normalization.

### Alternatives considered
Comparison could be done using Unicode code points instead of a locale-specific collating sequence, but that wouldn't generate the results expected by users for non-ASCII characters.

### LINKED ISSUE #6341: GREL string concatentation (+) overly aggressive
Currently the string concatenation operator (+) is used in cases where it isn't appropriate so that, for example `"1/1/1990".toDate() + 1` converts the values to strings and concatenates them, which is almost certainly not what the user expects.

### Current Results
The binary addition operator (+) results in string concatenation when used with anything except two numbers which is overly greedy behavior. 

### Expected Behavior
String concatenation should be limited to cases where at least one operand is a string. Unsupported cases (e.g. date + integer) should be an error (which just returns `null` in current GREL conventions).

tfmorris:
Fixes #6341. Closes #6340

- restrict string concatenation (+) to cases where at least one operand is a string (to avoid things like concatenating two dates as strings)
- Add support for dates and strings to comparison operators Strings use a collator for the default locale with normalized decomposition for the comparisons.
- Above implementation uses Comparable, so any future data types that implement that interface should get supported for free
- Add a bunch more tests

### DIFF
--- a/main/tests/server/src/com/google/refine/grel/GrelTests.java
+++ b/main/tests/server/src/com/google/refine/grel/GrelTests.java
@@ -114,6 +114,10 @@ public class GrelTests extends RefineTest {
         }
     }
 
+    static private String COMPARISON_OPERATORS[] = { "==", "!=", ">", "<", ">=", "<=", };
+    static private String INVALID_OPERATORS[] = { "=<", "=<", "**", "^", "!", };
+    static private String NUMERIC_OPERATORS[] = { "+", "-", "*", "/", "%", };
+
     @Test
     public void testMath() throws ParsingException {
         String tests[][] = {
@@ -131,8 +135,12 @@ public class GrelTests extends RefineTest {
                 { "3%2", "1" },
                 { "3/2", "1" },
                 { "3.0/2", "1.5" },
+                // integer comparisons
                 { "1==1", "true" },
                 { "1==2", "false" },
+                { "1!=2", "true" },
+                { "1!=1", "false" },
+//                { "1<>2", "true" },  // Scanner considers this an operator, but not the parser
                 { "1>2", "false" },
                 { "1<2", "true" },
                 { "1>1", "false" },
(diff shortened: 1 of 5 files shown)


main/tests/server/src/com/google/refine/grel/GrelTests.java
```diff
@@ -165,18 +188,30 @@ public void testPI() throws ParsingException {
...
+                { "'1/1/1900'.toDate() + ' foo'", "1900-01-01T00:00Z foo" },
+                { "1 + ' foo'", "1 foo" },
+                { "1.0 + ' foo'", "1.0 foo" },
+                { "2 * 3.0 + ' foo'", "6.0 foo" },
+                { "'a' > 'b'", "false" },
+                { "'a' < 'b'", "true" },
+                { "'a' == 'a'", "true" },
+                { "'a' == 'b'", "false" },
+                { "'a' != 'b'", "true" },
+                { "'E\u0301' == 'É'", "true" }, // combining accent equivalent to single character form
```
thadguidry, line 213: @tfmorris So how does a user control the CANONICAL_DECOMPOSITION itself?  Like for instance, if a power user does not want to accept accent equivalency and only wants to accept String literal checking for equality?  Do we need to add some other parameter to control their preference for that?

wetneb (approved):
Looks good to me!

main/tests/server/src/com/google/refine/grel/GrelTests.java
tfmorris: Obviously operators can't take additional parameters. I chose this as the default because I think it meets the principal of least surprise, but it's not out of the question to introduce a preference setting for the "strength" of string equality to ignore composition, accents, or even case. If you think this is an important feature, please create an issue for it.

Also, if you have a strong argument for why the default should be different, I'm happy to discuss. In my experience, the majority of users want things that look the same to be considered equal.

PR #6343, opened 2024-02-02
Add more HTTP headers. Fixes #6334

### LINKED ISSUE #6334: Adding header: 'Accept-Language"
I must add header: "Accept-Language: en" while I'm adding a column by fetching URLs. Normally, my GREL code looks like this: "https://api-zbiory.mnk.pl/api/object/"+value, but to get data in English I must add this header... 
I know that I can use Python instead of GREL to create a request with this header, but it isn't simple and create some problems with code, many errors and stress.

### Proposed solution

Add header "Accept-Language" to the list of available headings in window: "Add column by fetching URLs based on column ..."

tfmorris:
Fixes #6334

- Add support for Accept-Language, Accept-Charset, Accept-Encoding and If-Modified-Since.
- Also switch to using constants instead of free form strings.

### DIFF
--- a/main/src/com/google/refine/commands/HttpHeadersSupport.java
+++ b/main/src/com/google/refine/commands/HttpHeadersSupport.java
@@ -39,6 +39,7 @@ import java.util.Set;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.http.HttpHeaders;
 
 import com.google.refine.RefineServlet;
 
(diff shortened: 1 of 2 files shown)


wetneb:
Pinging @ostephens as the original author of this feature, if he has thoughts about the sets of headers to offer.

tfmorris:
I'm happy to restrict it to just `Accept-Language` if folks are opposed to the additional three items, but let's make a decision and move on.

ostephens:
Very happy to have all four added

ostephens (approved):

PR #6344, opened 2024-02-03
Use real TSV exporter/importer. Fixes #2071

### LINKED ISSUE #2071: TSV export escapes double quotes (") unnecessarily

tfmorris:
Fixes #2071

- Don't escape quotes when writing TSV format
- Escape \n and \r as is common practice
- Adjust tests to match desired behavior (they were written over a decade ago to match the observed behavior of the CSV library rather than the industry practice)
- Switch to real TSV parser for case where \t is the separator and there are no "exotic" options (so users can still use the CSV importer with \t if they need fancy quote support or something similar)
- Add test for unescaping of \t, \n, \r

### DIFF
--- a/main/src/com/google/refine/importers/SeparatorBasedImporter.java
+++ b/main/src/com/google/refine/importers/SeparatorBasedImporter.java
@@ -51,12 +51,15 @@ import java.util.Map;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.base.CharMatcher;
+import com.univocity.parsers.common.AbstractParser;
 import com.univocity.parsers.common.TextParsingException;
 import com.univocity.parsers.common.record.Record;
 import com.univocity.parsers.csv.CsvFormat;
 import com.univocity.parsers.csv.CsvParser;
 import com.univocity.parsers.csv.CsvParserSettings;
 import com.univocity.parsers.csv.UnescapedQuoteHandling;
+import com.univocity.parsers.tsv.TsvParser;
+import com.univocity.parsers.tsv.TsvParserSettings;
 import org.apache.commons.text.StringEscapeUtils;
 
 import com.google.refine.ProjectMetadata;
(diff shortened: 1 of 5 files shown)


main/src/com/google/refine/importers/SeparatorBasedImporter.java
```diff
@@ -136,21 +142,28 @@ public void parseOneFile(
...
-        format.setQuote(quote);
-        format.setLineSeparator("\n");
-        settings.setIgnoreLeadingWhitespaces(false);
-        settings.setIgnoreTrailingWhitespaces(false);
-        if (strictQuotes) {
-            settings.setUnescapedQuoteHandling(UnescapedQuoteHandling.RAISE_ERROR);
+        AbstractParser parser;
+        if (tsv) {
+            TsvParserSettings settings = new TsvParserSettings();
+            settings.setMaxCharsPerColumn(256 * 1024); // TODO: Perhaps use a lower default and make user configurable?
```
wetneb, line 148: I think it makes sense to have different default settings for TSV vs CSV but with this change we'd be ignoring quote settings (the quote character, and the setting to preserve leading/trailing whitespace), as soon as the user uses tabs as separator, right?
As a user I'd find it confusing to see that options I am offered aren't actually honored.

Would it not be preferable to integrate this logic in `createParserUIInitializationData`, so that the default options for those parameters are set to sensible values, and then we can stick to passing them to `CsvFormat` regardless of the separator, as it used to be?

wetneb (commented):
Thanks! I have to say I find it hard to judge whether people prefer escaping to quoting… Is this mandated by some RFC or other type of spec? Is this well understood by Excel/LibreOffice/Google Sheets?

I have no strong feelings on this but I suspect this can be a disruptive change for some folks so better err on the side of caution.

main/src/com/google/refine/importers/SeparatorBasedImporter.java
wetneb: Ok, I see the options are honored, thanks for the clarification. I think the UI is fine as such.

wetneb (approved):

Issue #6352, opened 2024-02-08
Checkbox UX


cooperzoe:
In the initial data import phase, when I go to specify the number of rows I'd like to load (or discard), I notice that I need to both fill in the number in the blank space and check the box. If I don't check the box, the limitation I set won't work. From a UX perspective, this feels redundant and unclear. 

I think it would be more intuitive for the user to have that limitation automatically confirmed upon entering the number. 
<img width="1233" alt="Screenshot 2024-02-08 at 12 27 27 PM" src="https://github.com/OpenRefine/OpenRefine/assets/52141007/7e982429-2fb3-4f1c-8331-46adbf32c690">

tfmorris:
I'm not sure I understand "limitation automatically confirmed" but if the intent is that the checkbox just track whether or not the field is populated with a number, perhaps the checkbox is not needed at all.

thadguidry:
The checkbox is not needed at all, but there should be a visual highlight somehow that the input box is filled and that this has **Enabled** the option.  Not sure what the best way to present "Enabled" is.  A checkbox works, but we don't want the user to have the sense that it's interactive.  We need some other visual indicator to show "Enabled" without the feeling of interactivity needed.  I've often seen Red color used around border for required input boxes not filled in.  But what about the opposite?  Wonder what other frameworks do?

A blue border perhaps when it's filled and effectively "Enabled" with a blue checkbox indicator at the end of the input box, might work well?
Example:  https://codepen.io/razieglow/pen/qdRQzw

wetneb:
Removing "good first issue" because there is still an ongoing discussion about the design.

tfmorris:
A similar argument could be made for two nearby fields which default to 0:

  | Ignore first | 0 | line(s) at beginning of file
-- | -- | -- | --

  | Discard initial | 0 | row(s) of data
-- | -- | -- | --

In my opinion, a better UI would be to:
1. remove the checkboxes
2. make the entry fields numeric with a default value of 0

This would enable the up/down errors to increment the field value which would make it easy for the user to select the, typically small, values that they want.

The max lines field probably needs to stay a string with a default value of empty which gets validated and parsed to an integer on user entry (unless there's a way to have a numeric field with an empty, non-zero, default).

[...]

PR #6358, opened 2024-02-11
Deal with localized numbers. Fixes OpenRefine#6351

### LINKED ISSUE #6351: Localized numbers not understood correctly in parsing options
The form inputs which expect numbers in the parsing panel (for instance for CSV/TSV) do not correctly understand numbers written with thousands separators such as "1,000" in English or "1 000" in French.

### To Reproduce
1. First, start a new project by uploading a CSV file with more than 1,000 rows
2. In the parsing options, input "1,000" in the "Load at most X lines" field

### Current Results

Observe that only one line is loaded in the preview, not one thousand.

### Expected Behavior

Numbers input with decimal separators should be understood properly, in all fields where a number is input.

### Versions
 - Operating System: Debian
 - Browser Version: Firefox 115.7.0esr
 - JRE or JDK Version: 17
 - OpenRefine: master branch (3.8-snapshot)

### Additional context

Discovered in a testing session with @cooperzoe.

Those input fields are currently marked as `type="text"`, but `type="number"` would feel more appropriate and likely solve this issue.

Xiayucheng1212:
Fixes #6351 

This change helps all input fields under the parsing options to accept any kind of number input format, such as "1,000" or "1 000".
- Adapt the "parseIntDefault" function to different localized numbers.
  1. Remove the numbers after the decimal point
  2. Extract all the numbers remained
  3. Join them together

### DIFF
--- a/main/webapp/modules/core/scripts/index/parser-interfaces/excel-parser-ui.html
+++ b/main/webapp/modules/core/scripts/index/parser-interfaces/excel-parser-ui.html
@@ -24,19 +24,19 @@
       <tr>
         <td width="1%"><input type="checkbox" bind="ignoreCheckbox" id="$ignore" /></td>
         <td><label for="$ignore" id="or-import-ignore"></label></td>
-        <td><input bind="ignoreInput" spellcheck="false" type="text" class="lightweight" size="2" value="0" />
+        <td><input bind="ignoreInput" type="number" class="lightweight" size="7" value="0" />
           <label for="$ignore" id="or-import-lines"></label></td></tr>
       <tr><td width="1%"><input type="checkbox" bind="headerLinesCheckbox" id="$headers" /></td>
         <td><label for="$headers" id="or-import-parse"></label></td>
-        <td><input bind="headerLinesInput" spellcheck="false" type="text" class="lightweight" size="2" value="1" />
+        <td><input bind="headerLinesInput" type="number" class="lightweight" size="7" value="1" />
           <label for="$headers" id="or-import-header"></label></td></tr>
       <tr><td width="1%"><input type="checkbox" bind="skipCheckbox" id="$skip" /></td>
         <td><label for="$skip" id="or-import-discard"></label></td>
-        <td><input bind="skipInput" spellcheck="false" type="text" class="lightweight" size="2" value="0" />
+        <td><input bind="skipInput" type="number" class="lightweight" size="7" value="0" />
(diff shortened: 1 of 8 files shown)


wetneb (commented):
That looks a little custom: although it probably works for all the number formatting conventions I am aware of, it can well be that in some languages it breaks, no?

What do you think of my suggestion to letting the browser handle that for us by setting those input fields to `type="number"` instead of `type="text"`?

Xiayucheng1212:
@wetneb  Yes, I think that will prevent potential problems. Thank you, I will try!

wetneb:
@Xiayucheng1212 are you still working on this? See also the relevant discussion in #6352: moving to `type="number"` could mean that the field is not empty by default, but contains 0 instead (I haven't tried). Obviously we need to preserve the default behaviour of not limiting the import at all, ideally with a blank field.

Xiayucheng1212:
@wetneb, Yes, I have tried the type="text", and it worked well. The defult value problem is also properly handled. I think it should be good to move on. Thank you!

wetneb (changes requested):
Thanks a lot! Interestingly, the results are different in Chrome and Firefox.

In Chrome it looks like this, which is probably fine although the number inputs could be shorter:
image

In Firefox it looks like this:
image
As you can see, it's not very convenient to use such narrow input fields to input numbers with more than one digit. I would propose to bump `size="2"` to `size="7"` on those input fields, which would give something like this:
image

wetneb:
@Xiayucheng1212 are you still working on this? I think this almost mergeable, we just need to adapt the width of the input fields for Firefox.

Xiayucheng1212:
> @Xiayucheng1212 are you still working on this? I think this almost mergeable, we just need to adapt the width of the input fields for Firefox.

Yes! Thank you for your notice, I edited the input size to 7. It's ready to go now.

wetneb (approved):
Thanks a lot!

wetneb (changes requested):
Sorry I just realized that there are a lot more such panels that we should probably fix in the same go, such as `excel-parser-ui.html`, `fixed-width-parser-ui.html` and so on.
Would you be willing to do that too?
Apologies again for realizing that so late.

Also, the `spellcheck="false"` attribute can probably be removed on those fields, as I don't think browsers will try enabling spellchecking on number inputs.

Xiayucheng1212:
Sure, no problem! I will check it later!

wetneb (approved):
Thanks a lot!

PR #6359, opened 2024-02-12
#5452 Removing missing unit warning

### LINKED ISSUE #5452: Warning for missing unit when making schema to Wikidata
### To Reproduce
1. Have a dataset ready to upload to Wikidata
2. Make a schema for a quantity property, but where Wikidata does not expect units, for example students count (P2196)

This video shows this error (link is time stamped to when I discover it): https://youtu.be/oSXxraHgbLs?t=1594

### Current Results

Under issues a warning like this is issued:
> **Units missing for students count (P2196)**
> Values such as `whatever number you had` on [Wikidata item] are expected to have units.

### Expected Behavior

No warning should be given.

### Screenshots

Skärmbild från 2022-11-18 18-20-40

### Versions
This was in Chrome Version 107.0.5304.87 on Ubuntu 22.04 using OpenRefine in Wikimedia PAWS, 28 October, 2022.

### Datasets

[Studenter-universitet.csv](https://github.com/OpenRefine/OpenRefine/files/10043481/Studenter-universitet.csv)

payalsaraljain:
Fixes #5452 

Fix for removing the warning for Statements which do not have any unit like student count
Added the logic for not showing warning in case Statement does not have any unit

- 
-
-

### DIFF
--- a/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/QuantityScrutinizer.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/QuantityScrutinizer.java
@@ -19,7 +19,7 @@ import org.openrefine.wikibase.qa.QAWarning;
 
 /**
  * Scrutinizer checking for units and bounds in quantities.
- * 
+ *
  * @author Antonin Delpeuch
  *
  */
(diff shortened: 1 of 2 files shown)


wetneb (changes requested):
Thanks a lot, that looks like a valid fix!
Do you think you could write an associated unit test? It should go in the corresponding test file, `extensions/wikibase/tests/src/org/openrefine/wikibase/qa/scrutinizers/QuantityScrutinizerTest.java`, where you should be able to take inspiration from the existing test cases (such as `testNoUnit`).

payalsaraljain:
Sure, will add the unit test case

payalsaraljain:
Hi @wetneb , I have added the test case, please let me know if any other changes are required

wetneb (changes requested):
Thank you! However, the test you have added seems to be identical to `testNoUnit`, no?

Given that `testNoUnit` was passing before your fix, I don't think that your new test (`testNoUnitRequired`) tests the case that you fixed.

Looking at the constraint defined on https://www.wikidata.org/wiki/Property:P2196 (mentioned in #5452), it seems that what we need is a constraint definition that uses a "no value" snak as qualifier:
image

wetneb (approved):
That looks great to me, thank you!

PR #6360, opened 2024-02-13
fixes #6330 Trim whitespace when converting stream to string in Impor…

### LINKED ISSUE #6330: Creating a project from a URL with a trailing space fails
### To Reproduce
1. First, launch OpenRefine and go to "Create Project" -> "Web Adresses (URL)"
2. Then, type in a URL with a trailing space at the end, such as "https://datasets.imdbws.com/name.basics.tsv.gz "
3. Finally, press "Next »"

### Current Results

Error: 
```
Error uploading data
Illegal character in path at index 46: https://datasets.imdbws.com/name.basics.tsv.gz "
```

### Expected Behavior

The whitespace is trimmed and the project is successfully created.

### Versions
 - OpenRefine: master branch

surajbora59:
…tingUtilities.java

Fixes #6330

Remove trailing space for url before assign to urlString

### DIFF
--- a/main/tests/server/src/com/google/refine/importing/ImportingUtilitiesTests.java
+++ b/main/tests/server/src/com/google/refine/importing/ImportingUtilitiesTests.java
@@ -500,12 +500,14 @@ public class ImportingUtilitiesTests extends ImporterTest {
     }
 
     /**
-     * This test method is designed to validate the behavior of the system when a URL with a trailing space is used.
-     * It simulates a scenario where a URL with a trailing space is used to retrieve content from a POST request.
-     * The expected behavior is that the system should trim the URL and proceed with the request as normal.
+     * This test method is designed to validate the behavior of the system when a URL with a trailing space is used. It
+     * simulates a scenario where a URL with a trailing space is used to retrieve content from a POST request. The
+     * expected behavior is that the system should trim the URL and proceed with the request as normal.
      *
-     * @throws IOException if an I/O error occurs during the test
-     * @throws FileUploadException if a file upload error occurs during the test
+     * @throws IOException
+     *             if an I/O error occurs during the test
+     * @throws FileUploadException
+     *             if a file upload error occurs during the test
      */
     @Test
     public void testTrailingSpaceInUrl() throws IOException, FileUploadException {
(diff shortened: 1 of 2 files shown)


wetneb (changes requested):
Thank you! That looks like a valid fix.

Do you think you could write an accompanying test for it? You should be able to take inspiration from existing tests in `ImportingUtilitiesTests.java`.

surajbora59:
@wetneb Have added the test case for this also, please have a look if test case looks fine.

wetneb:
@surajbora59 nice! It looks like the CI is failing because of some style changes. You could fix that by running `./refine lint` and committing the results.

surajbora59:
@wetneb pushed the changes from by running `./refine lint` command.

main/tests/server/src/com/google/refine/importing/ImportingUtilitiesTests.java
```diff
@@ -497,4 +499,61 @@ public void testIsCompressedFile() throws IOException {
...
+                                JSONUtilities.safePut(progress, "message", message);
+                            }
+                            JSONUtilities.safePut(progress, "percent", percent);
+                        }
+
+                        @Override
+                        public boolean isCanceled() {
+                            return job.canceled;
+                        }
+                    });
```
wetneb, line 553: Maybe we could do with an `Assert.fail()` here to make sure an exception was indeed thrown, otherwise the assertion in the catch block will be silently ignored (similarly to the test case you copied this from).

Alternatively you could start from the test case for a successful import, it would perhaps be a bit more natural.

surajbora59: I have added the assert in try block, also from one of test case I have tried to start test case for successful import not sure if it is correct.

surajbora59: Also for successful import I would be needing a file hosted somewhere that will be the issue here for this test case.

wetneb:
My bad, I thought we had a test case for a successful URL import, but it does look like we haven't got that yet (we could add one without relying on an online location, we could use HTTP mocking like in the test case you wrote). That's fine, the way you do it is a valid way to test your fix.

We probably only need to `./refine lint` again and it should be good to go. Thanks for your perseverance!

surajbora59:
image


No changes at my end after formatting the code.

I noticed that above workflow failed due to not able to resolve some dependencies
PFB, screenshot for that workflow


image

surajbora59:
Re-running the workflow is likely to resolve the problem.

wetneb (approved):
Thank you very much!

PR #6368, opened 2024-02-15
Improve error reporting when parsing incomplete GREL expressions

### LINKED ISSUE #6364: Internal error when parsing incomplete GREL expressions
Some incomplete GREL expressions generate an internal error in the backend, instead of a proper syntax error that can be reported to the user.

### To Reproduce
1. First, open the "Transform…" dialog on any column
2. Then, type the expression `value.replace(/\`

### Current Results

"Internal error" is displayed

### Expected Behavior

A parsing error should be displayed instead. For instance, if you remove the last character, you get:
"Parsing error at offset 14: Missing number, string, identifier, regex, or parenthesized expression".

### Screenshots

image

image

### Versions
 - Operating System: irrelevant
 - Browser Version: irrelevant
 - JRE or JDK Version: OpenJDK 17
 - OpenRefine: 3.8-SNAPSHOT (master branch)

surajbora59:
…imit

Fixes #6364 

Currently there was no check to maintain that index should be always be less than limit added that to avoid OutOfBoundsExpections 

image


wetneb (approved):
Updates to this parsing code are always a bit fiddly but in this case it looks good to me. Thank you!

PR #6370, opened 2024-02-19
Report importer initialization errors. Fixes #6058

### LINKED ISSUE #6058: XML import preview should report errors to the user
When attempting to import the Getty AAT vocabulary [XML file](https://web.archive.org/web/20220418022604/http://aatdownloads.getty.edu/VocabData/aat_xml_1121.zip) the XML parser throws an error about the `xsi` namespace prefix being undefined, but this error is swallowed and never shown to the user (or even logged on the console)

### To Reproduce
1. Create a project using the URL above

### Current Results

Create Project preview page shows an empty display with the message "Click on the first XML element corresponding to the first record to load." (even though there are no XML elements displayed) and the console logs
```
13:49:52.532 [..e.importers.XmlImporter] Error generating parser UI initialization data for XML file (474ms)
```

### Expected Behavior

Console displays the exception which is currently being swallowed
```
ParseError at [row,col]:[1,205]
Message: http://www.w3.org/TR/1999/REC-xml-names-19990114#AttributePrefixUnbound?Vocabulary&xsi:noNamespaceSchemaLocation&xsi
```

User is shown an interpretable (hopefully) error message, along with the line and column where the error occurred, and, ideally, the text of the offending line.

### Versions
 - OpenRefine: 3.7.5

### Datasets
https://web.archive.org/web/20220418022604/http://aatdownloads.getty.edu/VocabData/aat_xml_1121.zip

tfmorris:
Fixes #6058

- check for errors during createParserUIInitializationData phase
  (e.g. XML or JSON parse errors) and pass them to the front end
- DefaultImportingController now extends Command since it's really
  a command and it provides direct access to HTTP response methods
- DRY up Command and HttpUtilities redundant code
- add front end error handling for when initial preview parse fails
  and display error to the user
- use new error dialog instead of alert()s everywhere
- locally detabify Javascript near diffs to make code more readable
- add XML import e2e tests with both success and error cases
  (no XML tests previously & still missing most formats except CSV)

### DIFF
--- a/main/src/com/google/refine/commands/Command.java
+++ b/main/src/com/google/refine/commands/Command.java
@@ -36,8 +36,8 @@ package com.google.refine.commands;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.io.Writer;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
@@ -46,8 +46,6 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.core.JsonGenerator;
-import org.apache.velocity.VelocityContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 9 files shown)


tfmorris:
Looks like I missed a piece when separating this out from the rest of the error reporting fixes. Marking as Draft for now until I can investigate.

wetneb (approved):

tfmorris:
This may have improved #6454 also, so double check if that's still an issue.

tfmorris:
> This may have improved https://github.com/OpenRefine/OpenRefine/issues/6454 also, so double check if that's still an issue.

Turns out this was necessary, but not sufficient, but it's fixed now.

PR #6371, opened 2024-02-19
Avoid specifying test projects in CSV format

wetneb:
By specifying their columns and rows as arrays, we:
- avoid the dependency on the CSV parser
- make it easier to specify cell values of various datatypes
- improve readability

This is a preparation move towards backporting more of my testing improvements from 4.0

### DIFF
--- a/extensions/wikibase/tests/src/org/openrefine/wikibase/testing/WikidataRefineTest.java
+++ b/extensions/wikibase/tests/src/org/openrefine/wikibase/testing/WikidataRefineTest.java
@@ -2,8 +2,11 @@
 package org.openrefine.wikibase.testing;
 
 import static org.mockito.Mockito.mock;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.fail;
 
 import java.io.File;
+import java.io.Serializable;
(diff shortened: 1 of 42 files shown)


extensions/wikibase/tests/src/org/openrefine/wikibase/testing/WikidataRefineTest.java
```diff
@@ -29,10 +36,18 @@ public class WikidataRefineTest {
     private List<Project> projects = new ArrayList<Project>();
     private List<ImportingJob> importingJobs = new ArrayList<ImportingJob>();
 
+    /**
+     * @deprecated use {@link #createProject(String[], Serializable[][])}
+     */
+    @Deprecated
     public Project createCSVProject(String input) {
         return createCSVProject("test project", input);
     }
```
wetneb, line 45: This duplication with `RefineTest.java` will be addressed by the later modularization, isolating test utilities in an independent Maven module that extensions (be they in this repository or elsewhere) can depend on, and avoid having to reimplement our test utilities to write their own tests.

tfmorris (approved):

PR #6374, opened 2024-02-19
Fixes #6367: replace doesn't work

### LINKED ISSUE #6367: replace menu option with double quote (") doesn't work
I have several lines that use “”, but I should replace it with "".
When going to "Edit cells --> Replace" and replacing “ with " the following message appears:
Text transform on 0 cells in column name: value.replace("?",""") Undo

In "Transform", when I use GREL value.replace('“', '"') works normally.

Estação “Tatuca”, Centro de Memória - Unicamp (APS 1.2.046).jpg
Praça “Voluntários de 32”, Centro de Memória - Unicamp (APS 1.1.0297).jpg
Fabrica de Chapéos “Cury”, Centro de Memória - Unicamp (APS 1.1.0364).jpg

### Versions
 - Browser Version: Chrome 121.0.6167.184
 - JRE or JDK Version: java version "18.0.2"
 - OpenRefine: OpenRefine 3.7.9

surajbora59:
Fixes #6367 

While using replace from "”" to """, value.replace("”",""") grel expression.

Added code to escape double quotes(") when using replace.

image

### DIFF
--- a/main/src/com/google/refine/grel/Scanner.java
+++ b/main/src/com/google/refine/grel/Scanner.java
@@ -174,13 +174,15 @@ public class Scanner {
             while (_index < _limit) {
                 c = _text.charAt(_index);
                 if (c == delimiter) {
-                    _index++; // skip closing delimiter
-
-                    return new Token(
-                            start,
-                            _index,
-                            TokenType.String,
-                            sb.toString());
+                    if ((_index < _limit - 1 && _text.charAt(_index + 1) != delimiter) || (_index == _limit - 1)) {
+                        _index++; // skip closing delimiter
+                        return new Token(
+                                start,
+                                _index,
+                                TokenType.String,
+                                sb.toString());
+                    }
+                    sb.append(c);
                 } else if (c == '\\') {
                     _index++; // skip escaping marker
                     if (_index < _limit) {
--- a/main/src/com/google/refine/grel/Scanner.java
+++ b/main/src/com/google/refine/grel/Scanner.java
@@ -174,12 +174,13 @@ public class Scanner {
             while (_index < _limit) {
                 c = _text.charAt(_index);
                 if (c == delimiter) {
-                        _index++; // skip closing delimiter
-                        return new Token(
-                                start,
-                                _index,
-                                TokenType.String,
-                                sb.toString());
+                    _index++; // skip closing delimiter
+
+                    return new Token(
+                            start,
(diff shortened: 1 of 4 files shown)


tfmorris:
Thanks for tackling this! I think the problem is actually in the front end, not the back end, so I'd suggest looking at the Javascript code which constructs the GREL expression. It probably needs to escape any occurrence of the double quote (") character in either the search string or the replacement string provided by the user. The fix shouldn't require any changes to the GREL parser.

surajbora59:
@tfmorris , as advised, I have removed the modifications from the backend and added code to escape any occurrence of the double quote (").

tfmorris:
Thank you! Looks great. Would you mind adding a front end test to make sure that we don't break this again in the future?

https://github.com/OpenRefine/OpenRefine/blob/master/main/tests/cypress/cypress/e2e/project/grid/column/edit-cells/replace.cy.js

tfmorris:
One other request, please don't use the number sign (#) in the names of branches in the future. Since it's the shell comment character, it causes the command 

```
git checkout #6367-replace-not-working
```

to do nothing unless you quote the branch name.

surajbora59:
> Thank you! Looks great. Would you mind adding a front end test to make sure that we don't break this again in the future?
> 
> https://github.com/OpenRefine/OpenRefine/blob/master/main/tests/cypress/cypress/e2e/project/grid/column/edit-cells/replace.cy.js

I don’t have much idea about cypress but will try to look into it and add test case for the same.

wetneb:
@surajbora59 you could probably modify an existing test case in in [`replace.cy.js`](https://github.com/OpenRefine/OpenRefine/blob/9a70da0c6743ff188be9e5c198c748a30bae4c89/main/tests/cypress/cypress/e2e/project/grid/column/edit-cells/replace.cy.js#L4) so that the string replaced with contains a double quote.

To run the Cypress tests, see: https://openrefine.org/docs/technical-reference/functional-tests

surajbora59:
@wetneb , few observation in the doc which you shared I think it needs to be updated I noticed that there is no integration folder, adding the tests in the `*cy.js` are files automatically available in the list in cypress.

image

 Also need to updated the port  "333" instead of "3333"
image

surajbora59:
Test Results:

image

tfmorris (approved):
Looks perfect! Thank you!!

PR #6383, opened 2024-02-21
Introduce test utility to compare project contents

### LINKED ISSUE #6372: Java tests: ease comparison of project grid with expected state
At the moment, many of our Java tests follow a pattern like this:

https://github.com/OpenRefine/OpenRefine/blob/98d01b52de42b853cb5947114d99e036cf2f61ec/main/tests/server/src/com/google/refine/operations/cell/KeyValueColumnizeTests.java#L215-L227

This is cumbersome, as it is very verbose and does not let the reader quickly grasp the shape of the expected result.

### Proposed solution

Specify the expected state of a project after some operation by creating another project, in the expected state.
Introduce a test utility method to assert that two projects have equal contents.
Update the tests so that they go through such methods.

### Alternatives considered

We could also make an assertion on an exported version of the project (such as CSV), but that would not be very readable and would introduce a dependency on an exporter which has nothing to do with the code being tested.

### Additional context
Follow-up to #6371.

This will also help abstract away the way cell values are stored and accessed in a project, reducing the number of changes we need to make on the test suite for the new architecture developed in the `4.0` branch, as proposed here:
https://forum.openrefine.org/t/reproducibility-project-november-2023-report/1100/5

wetneb:
Closes #6372. Follow-up to #6371.

This is a preparation before backporting more tests from the 4.0 branch.

There are a couple of places which I didn't migrate because checking for equality on reconciliation objects is currently inconvenient (as `equals(Object)` is not overridden on the `Recon` class). I'll follow-up with a fix for that.

The only change in non-test classes is making `Cell` serializable.

### DIFF
--- a/main/tests/server/src/com/google/refine/RefineTest.java
+++ b/main/tests/server/src/com/google/refine/RefineTest.java
@@ -86,6 +86,7 @@ import com.google.refine.util.TestUtils;
  */
 public class RefineTest {
 
+    public static final double EPSILON = 0.0000001;
     protected static Properties bindings = null;
 
     protected Logger logger;
@@ -485,7 +486,7 @@ public class RefineTest {
                     assertEquals(
                             (double) actualCell.value,
                             (double) expectedCell.value,
-                            0.0000001,
+                            EPSILON,
                             String.format("mismatching cells in row %d, column '%s'", i, actual.columnModel.columns.get(j)));
                 } else {
                     assertEquals(
--- a/main/tests/server/src/com/google/refine/RefineTest.java
+++ b/main/tests/server/src/com/google/refine/RefineTest.java
@@ -47,6 +47,7 @@ import java.io.StringReader;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
+import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.BooleanNode;
@@ -70,6 +71,7 @@ import com.google.refine.importers.SeparatorBasedImporter;
 import com.google.refine.importing.ImportingJob;
 import com.google.refine.importing.ImportingManager;
 import com.google.refine.io.FileProjectManager;
+import com.google.refine.messages.OpenRefineMessage;
 import com.google.refine.model.Cell;
 import com.google.refine.model.Column;
 import com.google.refine.model.ModelException;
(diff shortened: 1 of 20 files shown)


main/tests/server/src/com/google/refine/RefineTest.java
```diff
@@ -449,4 +451,58 @@ protected void runAndWait(ProcessManager processManager, Process process, int ti
...
+    /**
+     * Checks that the grid contents are equal in both projects. Differences in "cell indices" are not taken into
+     * account as this is an internal detail: the goal is to assert equality of the user-facing parts of project data.
+     * 
+     * @param actual
+     *            the actual project state
+     * @param expected
+     *            the expected project state
+     */
+    public static void assertProjectEquals(Project actual, Project expected) {
```
tfmorris, line 464: For the test cases included in this PR, it seems like the `expected` project is always something which has just been created by passing an array of column names and a grid of values to the `createProject()` method and this comparison only considers the grid, not other aspects of a `Project`, so I'm wondering if it wouldn't be simpler and more direct to just compare directly against the column names & grid. This would allow you to skip half of the cell index redirection. This would also avoid the semantic mismatch if we want to introduce a real projectEquals assertion in the future.

Are there other places where the project creation is further removed from the test assertion?

main/tests/server/src/com/google/refine/importers/ExcelImporterTests.java
```diff
@@ -266,41 +235,28 @@ public void readXlsxAsText() throws IOException {
...
-
-        assertEquals(project.rows.get(1).getCellValue(9), "0001");
-        assertEquals(project.rows.get(2).getCellValue(9), "0002");
-
-        assertEquals(project.rows.get(ROWS - 1).getCellValue(10), "(617) 235-1322");
-
-        assertEquals(project.rows.get(ROWS - 1).getCellValue(11), NOW_STRING.substring(0, 10)); // date only
-
-        assertEquals(project.rows.get(ROWS - 1).getCellValue(12), "$" + numberFormat.format(1234.56));
+        final NumberFormat numberFormat = DecimalFormat.getInstance(Locale.getDefault());
```
tfmorris, line 240: Not really related to this PR, but I think it would be better to choose a non-US locale and pin it rather than allowing the tests to vary based on where they are run.

main/tests/server/src/com/google/refine/importers/MarcImporterTests.java
```diff
@@ -104,18 +103,44 @@ public void readMarcFileWithUnicode() throws FileNotFoundException, IOException
...
+                        { null, null, null, "300", null, null, "a", "p. cm." },
+                        { null, null, null, "440", "4", null, "a", "The World's classics" },
+                        { null, null, null, "651", "0", null, "a", "Florence (Italy)" },
+                        { null, null, null, null, null, null, "x", "Drama." },
+                        { null, null, null, "650", "0", null, "a", "Runaway husbands" },
+                        { null, null, null, null, null, null, "x", "Drama." },
+                        { null, null, null, "650", "0", null, "a", "Married women" },
+                        { null, null, null, null, null, null, "x", "Drama." },
+                        { null, null, null, "655", "7", null, "a", "Comedies." },
+                        { null, null, null, null, null, null, "2", "gsafd" },
```
tfmorris, line 140: Seems like it might be useful to figure out a way to only compared a sub-grid or to be able to specify rows to be skipped or something which would make cases like this less tedious to construct.

tfmorris (commented):
I haven't made it through all the importers yet, but wanted to at least return the first batch of comments.

I missed your comment about skipping things that depended on `equals()` & `hashCode()` when deciding the ordering to review in, so hopefully it wasn't important for this one to be done first.

main/tests/server/src/com/google/refine/RefineTest.java
```diff
@@ -449,4 +451,58 @@ protected void runAndWait(ProcessManager processManager, Process process, int ti
...
+                    assertEquals(
+                            (double) actualCell.value,
+                            (double) expectedCell.value,
+                            0.0000001,
+                            String.format("mismatching cells in row %d, column '%s'", i, actual.columnModel.columns.get(j)));
+                } else {
+                    assertEquals(
+                            actualCell == null ? null : actualCell.value,
+                            expectedCell == null ? null : expectedCell.value,
+                            String.format("mismatching cells in row %d, column '%s'", i, actual.columnModel.columns.get(j)));
```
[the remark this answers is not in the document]
wetneb: This is something that is already handled by `assertEquals` itself.
If you try executing `assertEquals(1, 2, "incorrect value in row X column Y")`, you will get the following failure:
`java.lang.AssertionError: incorrect value in row X column Y expected [2] but found [1]`.

main/tests/server/src/com/google/refine/RefineTest.java
wetneb: I prefer keeping the assertion separate from the creation of the expected project, as it feels architecturally cleaner to me. That's primarily motivated by my further changes that this prepares for: introducing the `Grid` class, an immutable class which represents the state of the project at a given point in the history. Once that is done, this test utility will be adapted to compare Grid objects instead. This would let you do things such as:
* testing that an operation has no effect on a grid: just check that the original grid is equal to the one that is returned by the operation. If we were to inline grid creation inside the test utility, you'd need to supply again the test data the grid was created from.
* even without introducing `Grid`, there are sometimes aspects of the project's state that aren't so easy to set up via `createProject` and which should rather be added manually afterwards (say, defining column groups). Keeping the assertion separate from the creation makes it possible to add those details and then do the assertion.

> This would also avoid the semantic mismatch if we want to introduce a real projectEquals assertion in the future.

What do you mean by that? What would a real `projectEquals` assertion look like?

main/tests/server/src/com/google/refine/RefineTest.java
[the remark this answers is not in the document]
tfmorris: Ahhh, ok! I guess I was (mis)remembering cases that used a different assert.

main/tests/server/src/com/google/refine/RefineTest.java
tfmorris: I wasn't proposing creating a project, but rather just using the test data grid directly. It seems like String[] columnNames, Serializable[][] values is effectively your grid anyway, but I don't feel strongly enough to argue about it (and I don't know what this future state looks like).

> What would a real `projectEquals` assertion look like?

It would compare other non-grid aspects of the project like its metadata, change history, etc.

wetneb: Ah yes I see. Then I think the introduction of the `Grid` class will go in the direction you are aiming at, since that will be an object that represents just the current project state (without history, project id and other things like that).

Checking for equality on entire projects (including the history) could be useful in certain cases but I don't think it would be convenient for operation tests to rely on this, typically because history metadata contains timestamps and localized information such as descriptions that would be annoying to equate exactly.

main/tests/server/src/com/google/refine/importers/ExcelImporterTests.java
wetneb: I have thought about proposing that indeed… One downside I would see with it is that it would let us bake localized strings into test assertions, which would mean the test suite could break if a translator decides to change the translation to something else.
In any case, moving to something like that would have impacts in quite a few other places in the test suite I expect, so better leave that for another PR.

tfmorris (approved):
Github seems to think I still have a pending comment or two, but since you've addressed the epsilon issue and this has been dangling, please go ahead and merge.

tfmorris:
p.s. and apologies for the delay. I discovered this while cleaning up stale branches in my fork.

PR #6388, opened 2024-02-22
Add equals(), hashCode() and toString() for Cell and subcomponents

wetneb:
To ease testing and debugging.
Follow-up to #6383.

### DIFF
--- a/main/src/com/google/refine/model/Cell.java
+++ b/main/src/com/google/refine/model/Cell.java
@@ -40,6 +40,7 @@ import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
+import java.util.Objects;
 import java.util.Properties;
 
 import com.fasterxml.jackson.annotation.JacksonInject;
@@ -193,6 +194,29 @@ public class Cell implements HasFields {
 
     @Override
     public String toString() {
+        // TODO this is kept like this for now, but it should rather be a string which makes the difference
+        // between the cell and the value clear (i.e. not be the same string representation as the value)
+        // and include a representation of the recon field too.
+        // (toString() should not be exposed to the end user and is only meant to be used for logging/debugging
+        // purposes)
         return StringUtils.toString(value);
     }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(recon, value);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        Cell other = (Cell) obj;
+        return Objects.equals(recon, other.recon) && Objects.equals(value, other.value);
+    }
+
 }
--- a/main/src/com/google/refine/model/Recon.java
+++ b/main/src/com/google/refine/model/Recon.java
@@ -34,10 +34,12 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package com.google.refine.model;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.Random;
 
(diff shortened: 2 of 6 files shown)


main/src/com/google/refine/model/Cell.java
```diff
@@ -193,6 +194,29 @@ static public Cell deserialize(
 
     @Override
     public String toString() {
+        // TODO this is kept like this for now, but it should rather be a string which makes the difference
+        // between the cell and the value clear (i.e. not be the same string representation as the value)
+        // and include a representation of the recon field too.
+        // (toString() should not be exposed to the end user and is only meant to be used for logging/debugging
+        // purposes)
```
tfmorris, line 201: If this is an implementation effort thing, perhaps one of the following would help (we already depend on both Apache lang3 and Guava):

- https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/builder/ReflectionToStringBuilder.html
- https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/builder/RecursiveToStringStyle.html
- https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/builder/ToStringBuilder.html
- https://javadoc.io/static/com.google.guava/guava/31.0.1-jre/com/google/common/base/MoreObjects.html#toStringHelper(java.lang.Class)

main/src/com/google/refine/model/Recon.java
```diff
@@ -421,4 +423,42 @@ public Recon(
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        Recon other = (Recon) obj;
+        return Objects.equals(candidates, other.candidates) && Objects.equals(error, other.error)
```
tfmorris, line 446: Is this equivalent to `Objects.deepEquals()`? If yes, please simplify. If no, please add a comment about which fields are (intentionally) left out. I think all fields are included, but I easily could have missed something.

tfmorris (changes requested):
A couple of minor suggestions. Approved without re-review when you've addressed them.

main/src/com/google/refine/model/Recon.java
wetneb: Do you mean [this Objects.deepEquals()](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#deepEquals-java.lang.Object-java.lang.Object-)? It's not clear to me how using this would simplify this method. I don't think I can replace the entire method code by just a call to `deepEquals` since according to its documentation, it would just call back this very `equals` method, resulting in an infinite recursion. What sort of simplification did you have in mind?

main/src/com/google/refine/model/Cell.java
wetneb: I didn't leave this out because I thought it would be complicated to re-implement (the methods added in this PR are all generated by Eclipse), but rather because I am worried that the return value of this `toString()` method can currently make it to the end user in one way or another (for instance, maybe in expression preview). Therefore, changing this method could have a noticeable impact for users, which would be unintended.

Independently, I'd rather avoid relying on helpers which rely on reflection, as it probably comes with a performance penalty, or at least with a complexity that can be avoided easily in such a case.

Does that make sense?

tfmorris: Wouldn't this be showing up as "Cell@1234" if the default `Object.toString()` were being invoked?

There are implementations in that list which don't use reflection. Or you could roll you own by hand.

We can circle back and fix it another time if you'd rather not do it now.

main/src/com/google/refine/model/Recon.java
tfmorris: Yes, that was dumb of me! That method doesn't do what I thought it did.

tfmorris (approved):

main/src/com/google/refine/model/Cell.java
wetneb: > Wouldn't this be showing up as "Cell@1234" if the default Object.toString() were being invoked?

No, at the moment, because we are already overriding `toString()` (before this PR), this would simply show up as the cell value - which is why I suspect there could be places where we do that without noticing.

PR #6389, opened 2024-02-22
Introduce a test helper to run an operation on a project

wetneb:
This simplifies somewhat writing tests for operations, abstracting away all the process management business.
It should speed test execution a bit by avoiding to rely on a periodic polling for the process' status and avoiding the use of an extra thread.

The goal is to be able to write cleaner tests, running an operation on a project and checking the state of the grid afterwards.


main/tests/server/src/com/google/refine/RefineTest.java
```diff
@@ -431,6 +434,29 @@ protected ButterflyModule getCoreModule() {
...
+     * @returns the duration of the operation in milliseconds
+     */
+    protected long runOperation(AbstractOperation operation, Project project) throws Exception {
+        long start = System.currentTimeMillis();
+        Process process = operation.createProcess(project, new Properties());
+        if (process.isImmediate()) {
+            process.performImmediate();
+        } else {
+            LongRunningProcessStub longRunning = new LongRunningProcessStub((LongRunningProcess) process);
+            longRunning.run();
```
tfmorris, line 450: Because this is running on the main thread an unhandled exception should propagate, what happens if this runs forever (or even above some upper bound)?

If we don't have one already, it might be worth adding a test to make sure that exception handling works the way we expect. Also some separate explicit tests of the long running process lifecycle (isRunning, etc).

main/tests/server/src/com/google/refine/RefineTest.java
```diff
@@ -431,6 +434,29 @@ protected ButterflyModule getCoreModule() {
...
+        }
+        long end = System.currentTimeMillis();
+        return end - start;
+    }
+
+    /**
+     * @deprecated use {@link #runOperation(AbstractOperation, Project)}
+     */
+    @Deprecated
```
tfmorris, line 459: How much would it speed things up to drop the polling interval from 200 to 100 (or 50 or lower) ms? Test efficiency is good, but it's also valuable to stay as close as possible to how the production code works.

main/tests/server/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperationTests.java
```diff
@@ -162,14 +154,11 @@ public void testUrlCaching() throws Exception {
...
-                    500,
+                    5,
                     true,
                     null);
 
-            // We have 100 rows and 500 ms per row but only two distinct
-            // values so we should not wait much more than ~1000 ms to get the
-            // results.
-            runAndWait(op, 1500);
+            runOperation(op, project);
```
tfmorris, line 161: Is this testing the functionality implied by the previous comment about collapsing identical values?

main/tests/server/src/com/google/refine/RefineTest.java
```diff
@@ -431,6 +434,29 @@ protected ButterflyModule getCoreModule() {
...
 
+    /**
+     * Runs an operation on a project. If it's a long-running operation, its process is run in the main thread until
+     * completion.
+     * 
+     * @returns the duration of the operation in milliseconds
+     */
+    protected long runOperation(AbstractOperation operation, Project project) throws Exception {
+        long start = System.currentTimeMillis();
+        Process process = operation.createProcess(project, new Properties());
```
tfmorris, line 445: Can we align with how the real code does it?

```suggestion
        Process process = operation.createProcess(project, ExpressionUtils.createBindings(Project project));
```

tfmorris (commented):
I like the simplification and the redundancy elimination. I'm on the fence about skipping thread creation. Did you quantify how much time it saved vs something like just reducing the polling interval?

main/tests/server/src/com/google/refine/RefineTest.java
wetneb: If the operation runs forever, then the test will run forever indeed. That's a downside of this approach.
For me, the main motivation is to reduce the resource usage of the test suite, by avoiding the creation of many threads (which are not even reused via a thread pool, as is common practice), and simplify debugging.

If you prefer keeping it in a separate thread, I'd be happy revert back to that. My main motivation here is to introduce this test utility to simplify tests and minimize the changes I need to do later on. The specifics of how operations are executed is not so important to me at this stage because that will change completely with the new architecture anyway.

main/tests/server/src/com/google/refine/RefineTest.java
wetneb: I think the speed up will be marginal in most cases. Same as above: if you prefer creating another thread and polling on it, we can revert to that.

main/tests/server/src/com/google/refine/RefineTest.java
wetneb: Can you point to the place in the real code where we do this? I'd find it surprising, because `ExpressionUtils.createBindings` is creating the environment for evaluating expressions, which is something completely different from the creation of a `Process`.

tfmorris: Oops!  I was confusing this with the variable bindings. As a matter of fact, it doesn't look the options are ever used at all, so we should probably just remove that parameter (separately).

main/tests/server/src/com/google/refine/RefineTest.java
tfmorris: Let's put it back to using a separate thread then, if you're OK with that. It keeps things closer aligned with the production code and we're talking hundreds or perhaps thousands of threads for the test suite, not millions.

My rough timings indicate that the savings is on the order of 5% and dropping the polling interval would get some of that back.

main/tests/server/src/com/google/refine/RefineTest.java
tfmorris: Feel free to drop the polling interval if you want.

tfmorris (approved):
LGTM. One small suggestion on a comment, but feel free to merge.

PR #6393, opened 2024-02-24
Remove unneccessary "USE" statement in psql test fixture

### LINKED ISSUE #2637: Fix a test with SQL engine and Travis & Appveyor for the Database Import
**Describe the bug**
It would be really nice to have both Travis and Appveyor, actually test create the sqlite and test... as well as fix the `travis-pgsql.sql` failure on Appveyor.

This looks like a syntax error, we probably don't need that USE line any longer in `extensions\database\tests\conf\travis-mysql.sql` ?  

Since we already state on Travis config that we are going to use the test_db through the `before_install` Line 26 here: https://github.com/OpenRefine/OpenRefine/blob/master/.travis.yml#L26

Why is this comment still here?  https://github.com/OpenRefine/OpenRefine/blob/master/extensions/database/tests/conf/travis_tests.xml#L32  

Did something not get put back into place correctly?  https://github.com/OpenRefine/OpenRefine/commit/a6424e53a092abd90a72a24d60cae5d8b20eca8f  

Groups run nothing? Is that a problem? So let's ensure the integrity of our database extension with tests by Travis and Appveyor that are healthy again and working correctly fully.

 - Version 3.4

**Screenshots**
On GitHub.  The PR when viewed will show near the bottom the checks passing/failing.
image

When you click on Details blue link it will take you to Appveyor or Travis and there you can see the logs from each respectively:
image

Our tests should actually use those various SQL engine instances that we have wired up to spin up in the CI (Continuous Integration) vendors like Appveyor and Travis.  I am unsure if all the DB extension tests are using those instances...and if we are in fact still testing against those 3 DB instances we spin up on Travis and Appveyor.  So double-check all of that.  I want to make sure we didn't accidentally turn off testing for some, or skipping tests for some (we did skip some things in the past because of problems with Appveyor but should no longer be skipping tests AT ALL, so that we can ensure regression testing properly going forward with all these changes and expanding our contributor base such as with GSoC and yourself ;-)  I.E. I don't want our quality to go down and for us to get lazy.

**Additional context**
Recommandations: Iterate on it until db tests are working well on Appveyor and Travis.  Looks like mostly just some .yml file updates and .sql file updates in various places are all that's needed.  You can see the previous pqsql and mysql as examples...oh wait, we have sqlite already by that fellow a few months ago...so just ensure Appveyor and Travis are indeed testing the DB extension against those 3, postgres, mysql, sqlite properly.

Our tests should actually use those various SQL engine instances that we have wired up to spin up in the CI (Continuous Integration) vendors like Appveyor and Travis. I am unsure if all the DB extension tests are using those instances...and if we are in fact still testing against those 3 DB instances we spin up on Travis and Appveyor. So double-check all of that. I want to make sure we didn't accidentally turn off testing for some, or skipping tests for some (we did skip some things in the past because of problems with Appveyor but should no longer be skipping tests AT ALL, so that we can ensure regression testing properly going forward with all these changes and expanding our contributor base such as with GSoC and yourself ;-) I.E. I don't want our quality to go down and for us to get lazy.

_This issue is mostly written from discussion in the previous issue: « Only show the database login fields when necessary #2615 »_

wetneb:
Closes #2637.

### DIFF
--- a/extensions/database/tests/conf/test-pgsql.sql
+++ b/extensions/database/tests/conf/test-pgsql.sql
@@ -1,5 +1,3 @@
-USE test_db;
-
 CREATE TABLE IF NOT EXISTS test_table (
   id integer NOT NULL,
   ue_id char(8) NOT NULL,


tfmorris (approved):
RSLGTM - tests pass

PR #6394, opened 2024-02-24
Draft: 6270 more db connections

### LINKED ISSUE #6270: Let's have more connection to db (odbc, jdbc, etc..)
Hello all,

I just would like to connect to my databases. As of today, only 4 are proposed : 
image

### Proposed solution
There are a few connection standards such as odbc, oledb jdbc, db-api... that you can use in order to extend supported databases without developing a totally specific solution

### Alternatives considered
None

### Additional context
in corporate environment, data is mainly stored in databases..

Best regards,

Simon

surajbora59:
Fixes #6270 

added generic class dbservice and dbmanager class

### DIFF
--- a/extensions/database/src/com/google/refine/extension/database/SQLType.java
+++ b/extensions/database/src/com/google/refine/extension/database/SQLType.java
@@ -30,6 +30,7 @@
 package com.google.refine.extension.database;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 public final class SQLType {
@@ -41,12 +42,22 @@ public final class SQLType {
         this.driverContainer = container;
     }
 
-    public static SQLType forName(String name) {
-        for (SQLType sqlType : jdbcDriverRegistry.values()) {
-            if (sqlType.getIdentifier().equalsIgnoreCase(name)) {
-                return sqlType;
+    public static SQLType forName(List<String> names) {
+        for(String name : names){
+            for (SQLType sqlType : jdbcDriverRegistry.values()) {
+                if (sqlType.getIdentifier().equalsIgnoreCase(name)) {
+                    return sqlType;
+                }
             }
         }
+        return null;
+    }
+    public static SQLType forName(String name) {
+            for (SQLType sqlType : jdbcDriverRegistry.values()) {
+                if (sqlType.getIdentifier().equalsIgnoreCase(name)) {
+                    return sqlType;
+                }
+            }
         return null;
     }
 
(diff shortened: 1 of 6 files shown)


extensions/database/src/com/google/refine/extension/database/SQLType.java
```diff
@@ -41,14 +42,24 @@ private SQLType(DriverContainer container) {
...
 
-    public static SQLType forName(String name) {
-        for (SQLType sqlType : jdbcDriverRegistry.values()) {
-            if (sqlType.getIdentifier().equalsIgnoreCase(name)) {
-                return sqlType;
+    public static SQLType forName(List<String> names) {
+        for(String name : names){
+            for (SQLType sqlType : jdbcDriverRegistry.values()) {
+                if (sqlType.getIdentifier().equalsIgnoreCase(name)) {
+                    return sqlType;
```
wetneb, line 49: The signature of this method looks odd to me - is this actually used anywhere?

wetneb (commented):
Thanks a lot!

I think in the given state, this looks like a sensible refactoring. I am assuming you would delete all the vendor-specific classes such as MySQLDatabaseService.
However I am not clear on how it would solve #6270 (which is not very precisely specified!). I saw you added a few more databases: we would probably need to add new Maven dependencies for their JDBC drivers to be in the classpath. Maybe that's an acceptable way of doing things: we'd need to check the license of those drivers and how much they weigh.

But intuitively we could go more in the direction of #1909: letting the user supply a JDBC driver to connect to whatever other DB. Arguably that's more work and it's not even clear to me what the expected UX would be like, but it would be interesting to have this use case in mind and gear the internal architecture towards that a bit more. In a sense, I feel like this refactoring is going further away from this use case.

extensions/database/src/com/google/refine/extension/database/SQLType.java
surajbora59: image

it is used to check if the jdbc driver is present in `jdbcDriverRegistry` map.

image


same is being used in other vendor-specific classes.

surajbora59:
> But intuitively we could go more in the direction of #1909: letting the user supply a JDBC driver to connect to whatever other DB. Arguably that's more work and it's not even clear to me what the expected UX would be like, but it would be interesting to have this use case in mind and gear the internal architecture towards that a bit more. In a sense, I feel like this refactoring is going further away from this use case.

Ref: **DBeaver Community Edition**
image

New fields/UX could be something like above with an extra field to give driver from the user. Also we can remove this className: field if somehow we can fetch it from the driver jar provided(not sure about this).

PR #6397, opened 2024-02-24
Fix recon dialog suggested types radio button

### LINKED ISSUE #6328: Recon dialog: suggested types should be clickable (missing <label> DOM element)
In the recon dialog, the suggested types are not clickable: one needs to click the radio button next to them.
This is bad accessibility.

### Proposed solution

Enclose the type name and id in a `<label>` element associated to the radio button, so that one can click on the text as well.

### Alternatives considered
none

### Additional context
image

Redeem-Grimm-Satoshi:
Fixes #6328

-  label element added for the `type name` and `id` , associated with the radio button making it clickable


main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -169,6 +169,7 @@ ReconStandardServicePanel.prototype._populatePanel = function() {
       var radio = $('<input type="radio" name="type-choice">')
       .val(typeID)
       .attr("typeName", typeName)
+      .attr("id","type-choice"+ typeID)
```
wetneb, line 172: The `typeID` is provided by the reconciliation and could be any string. For instance, it could contain whitespace, which is forbidden by the specifications (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id). So perhaps it would be better to use the index of the type in the list. You would need to add it as a parameter to the `createTypeChoice` function, but that looks doable.

wetneb (changes requested):
Well done, that looks very good to me! I haven't tried it out yet but I noticed one potential problem, see below.

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
wetneb: Another approach, which does not require introducing a unique `id`, is to make sure the `<label>` tag wraps both the `input` element and its text.
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label.

Redeem-Grimm-Satoshi: You're right, An easier fix to the already proposed solution could be to add `type.trim()` to this `var typeID = typeof type == "string" ? type : type.id;` for that will fix the white-space issue. I've tested it and it works quite alright

@wetneb what do you think?

wetneb: I don't think trimming will help because the whitespace can be anywhere in the string. Transforming the type ids can also make them non-unique. I'd prefer any of the two solutions above.

Redeem-Grimm-Satoshi: I just pushed the code for the second approach

main/webapp/modules/core/scripts/reconciliation/standard-service-panel.js
```diff
@@ -166,43 +166,52 @@ ReconStandardServicePanel.prototype._populatePanel = function() {
...
-        $(td1).html(
-            typeName + 
-            '<br/>' +
-            '<span class="type-id">' + typeID + '</span>');
+      // Append the label to the second cell
+      $(td1).append(label);
+
+      // Add the type ID if it's different from the type name
+      if (typeName !== typeID) {
+        $(td1).append('<br/><span class="type-id">' + typeID + '</span>');
```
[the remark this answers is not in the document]
Redeem-Grimm-Satoshi: I wrapped the entirec` <tr>` element inside the `<label>` and it worked quit well; 

second approach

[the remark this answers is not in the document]
Redeem-Grimm-Satoshi: @wetneb

wetneb:
Thanks a lot for your efforts!

I would be reluctant to introduce a `<table>` element within each `<label>`, because using `<table>` for layout purposes is generally frown upon.

I am far from an HTML expert and might not be so up to date on what the latest best practices are, but one DOM structure you could take inspiration from is the one we used in the reconciliation service selection dialog (just before the dialog you are working in), where it looks like this:
```html
<label>
    <input type="radio" value="..." />
    <span>
         <div>Readable name</div>
         <div>https://..... </div>
    </span>
</label>
```
With the label having `display: block` and the span having `display: inline-block`.

It's not perfect, because it's probably a bit weird to have divs into a span instead of the other way around, but that's probably still better than a table containing labels containing tables.

Of course, those things should just be components that we implement once and reuse in many places, but as you can see we are not quite there yet!

Alternatively, I still think using the indices to generate the DOM ids would be pretty easy to implement…

Redeem-Grimm-Satoshi:
@wetneb Thanks for the feedback. I adjusted it according using the `indices` to generate the` DOM ids` for suggested types radio button

Probably the first approach you mentioned earlier

wetneb (approved):
That works well, thank you!

PR #6398, opened 2024-02-25
Fix benign "bug"

aminnaggar:
In the `Encode_Non_Initial_G_Front_Vowel` method inside  `Metaphone3.java`. The negation is redundant since if `m_current` equaled 3 and it was indeed prefixed with "MAC" the first condition would fail, and we'd land where we need to anyway.

```java
if(
  ! ((m_current == 2) && StringAt(0, 2, "MC", "")) 
  || ((m_current == 3) && StringAt(0, 3, "MAC", ""))
)
```

### DIFF
--- a/main/src/com/google/refine/clustering/binning/Metaphone3.java
+++ b/main/src/com/google/refine/clustering/binning/Metaphone3.java
@@ -3259,7 +3259,7 @@ public class Metaphone3 {
 				if(Internal_Hard_G())
 				{
 					// don't encode KG or KK if e.g. "mcgill"
-					if(!((m_current == 2) && StringAt(0, 2, "MC", "")) 
+					if(((m_current == 2) && StringAt(0, 2, "MC", "")) 
 						   || ((m_current == 3) && StringAt(0, 3, "MAC", "")))
 					{
 						if(SlavoGermanic())


tfmorris (commented):
Thanks for the contribution -- and welcome to the project! 

Given that this change is in a rather complex piece of contributed code, for which we don't have test coverage, I think it might be better to just leave it as it is, but I'll let @wetneb weigh in for the final decision.

wetneb:
Thanks! I am unclear on if the change causes any visible difference for users in the end - it seems like it does not, right?

I had a cursory look at this class and there seems to be quite a few other places where some conditions cannot be reached because of the surrounding context. I don't think this is really worth cleaning up to be honest. As @tfmorris said, it's not tested, and I think it's also not worth investing too much effort into getting a decent coverage (see #6400).

I'd be interested to know what brought you to submitting this change in the first place. How did you notice this "bug"? What was your motivation to work on it?

wetneb:
Closing per the above, feel free to reopen @aminnaggar if you feel like we have misunderstood your motivation here.

aminnaggar:
That's super fair, I'd do the same. I wasn't aware of the poor test coverage here, and it's not the kind of thing worth investing a bunch of time on.

re: how I found it: I was porting it to typescript and my linter started yelling at me lol.

Nice meeting both of you, apologies for the distraction.

PR #6400, opened 2024-02-26
Metaphone3 testing improvements

wetneb:
- converts some dead code (`main` method from Metaphone3) into basic unit test for the Metaphone3 heuristic
- exclude Metaphone3 from our coverage results, because it has a few orders of magnitude more instructions than the rest of our code, dominating the statistics reported in coverage results. It is a heuristic anyway (it will never be correct).

I also considered:
* extracting Metaphone3 as an external library, either as a separate Maven module or moving it to a separate repository, but I am not sure anyone would want to reuse it
* auto-generating many test cases for it for the sake of increasing coverage, but that's a bit pointless given that it will not actually add any testing value in my opinion. It's not worth the effort to write the tests nor the resources to run the tests as part of our test suite, I would say


tfmorris:
> extracting Metaphone3 as an external library, either as a separate Maven module or moving it to a separate repository, but I am not sure anyone would want to reuse it

Our version is a stranded fork of a now commercial product, so I agree this doesn't make sense.
https://en.wikipedia.org/wiki/Metaphone#Metaphone_3

tfmorris (approved):

PR #6407, opened 2024-03-01
Remove use of Properties for URL query parsing, fixes #6403

### LINKED ISSUE #6403: Remove use of java.util.Properties for URL query parsing utility
As part of the epic #6391 to minimize our use of `java.util.Properties` we want our URL parameter parsing to not use this as a return type.

### Proposed solution
- deprecate all methods in `ParsingUtilities` which return java.util.Properties`
- create a replacement for `static public Properties parseUrlParameters(HttpServletRequest request)` with the signature `static public Map<String, String> parseParameters(HttpServletRequest request)` which uses available functionality for its implementation like `request.getParameterMap()`
- modify all uses of `parseUrlParameters` to use the new method and adjust all necessary code to accommodate the new return type. This primarily affects our bundled importers.
- be careful to preserve any public APIs (ie you can add new methods, but not remove or modify existing ones). I think that only `ImportingUtilities` has public APIs which are affected, but double check.

### Alternatives considered

We could use `request.getParameterMap` directly, but since it returns `Map<String, String[]>` instead of Map<String, String> it would require more extensive modifications to the existing code and none of our current parameters can be repeated, so the array support is unnecessary.

EliasStihl:
Fixes #6403

- Deprecate all methods in `ParsingUtilities` that return `java.util.Properties`.
- Introduce a new method, `static public Map<String, String> parseParameters(HttpServletRequest request)`, as a replacement for `parseUrlParameters`. This new method utilizes the `request.getParameterMap()` functionality for its implementation.
- Modify all uses of `parseUrlParameters` to use the new method.
- Adjust all necessary code to accommodate the new return type `(Map<String, String>)`.
- Deprecate all public methods `ImportingUtilities` that use `java.util.Properties` and make new versions of them using `Map<String, String>` instead.

### DIFF
--- a/extensions/database/src/com/google/refine/extension/database/DatabaseImportController.java
+++ b/extensions/database/src/com/google/refine/extension/database/DatabaseImportController.java
@@ -33,7 +33,7 @@ import java.io.IOException;
 import java.io.Writer;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Properties;
+import java.util.Map;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
@@ -83,9 +83,9 @@ public class DatabaseImportController implements ImportingController {
             logger.debug("doPost Query String::{}", request.getQueryString());
         }
         response.setCharacterEncoding("UTF-8");
-        Properties parameters = ParsingUtilities.parseUrlParameters(request);
+        Map<String, String> parameters = ParsingUtilities.parseParameters(request);
 
-        String subCommand = parameters.getProperty("subCommand");
+        String subCommand = parameters.get("subCommand");
 
         if (logger.isDebugEnabled()) {
             logger.info("doPost::subCommand::{}", subCommand);
@@ -131,7 +131,8 @@ public class DatabaseImportController implements ImportingController {
      * @throws ServletException
      * @throws IOException
      */
-    private void doInitializeParserUI(HttpServletRequest request, HttpServletResponse response, Properties parameters)
+    private void doInitializeParserUI(HttpServletRequest request, HttpServletResponse response,
+            Map<String, String> parameters)
             throws ServletException, IOException {
         if (logger.isDebugEnabled()) {
             logger.debug("::doInitializeParserUI::");
@@ -164,13 +165,13 @@ public class DatabaseImportController implements ImportingController {
      * @throws DatabaseServiceException
      */
     private void doParsePreview(
-            HttpServletRequest request, HttpServletResponse response, Properties parameters)
+            HttpServletRequest request, HttpServletResponse response, Map<String, String> parameters)
             throws ServletException, IOException, DatabaseServiceException {
         if (logger.isDebugEnabled()) {
-            logger.debug("JobID::{}", parameters.getProperty("jobID"));
+            logger.debug("JobID::{}", parameters.get("jobID"));
         }
 
-        long jobID = Long.parseLong(parameters.getProperty("jobID"));
+        long jobID = Long.parseLong(parameters.get("jobID"));
         ImportingJob job = ImportingManager.getJob(jobID);
         if (job == null) {
             HttpUtilities.respond(response, "error", "No such import job");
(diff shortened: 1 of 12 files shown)


EliasStihl:
When I run the tests locally with `./refine test`, all tests pass, any idea why the tests fail on the CI server @tfmorris ?

tfmorris:
The tests pass locally for me as well, but a re-run of the CI failed as
well, so it's not just a flaky test. Let me investigate...

tfmorris:
Three of the Database tests fail when run in my IDE as well, so looking
into what's different among the CI, ./refine test, and IntelliJ
environments.

tfmorris:
OK, I'm not sure why the test isn't failing locally (perhaps some difference in the test runner environment?), but the failure is real. I think what's needed is to update the mocks to implement `HttpServletRequest.getParameterMap()` rather than just `HttpServletRequest.getParameter()` as they do now 
https://github.com/OpenRefine/OpenRefine/blob/cdbdf362952914f7284dc182084293f12bc9f3b9/extensions/database/tests/src/com/google/refine/extension/database/DatabaseImportControllerTest.java#L171-L178

wetneb:
The local test suite does not have all tests of the database extension enabled, because those require having a running SQL server (of different flavors).
You can see in the CI configuration how they are enabled:
https://github.com/OpenRefine/OpenRefine/blob/e6dbcfcd2619dc5a8328daf7d1de522a8369e276/.github/workflows/pull_request_server.yml#L99-L110

EliasStihl:
Thank you for your responses, but unfortunately, I am not sure how I should move forward with this? I have tried, but I don't know how I should replicate the failures locally so it becomes very hard for me to test.  

One small difference between the old URL parser and the new parser is that the old parser took the last value if the key was provided multiple times. However, @tfmorris explained in the issue that parameters could not be repeated. Therefore it shouldn't be a problem to return the first value like this  `return request.getParameterMap().entrySet().stream()
                .collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue()[0]));` ?

tfmorris:
@EliasStihl My apologies. The situation is definitely sub-optimal. The method I used to recreate the failure was to directly run the tests in [DatabaseImportControllerTest](https://github.com/OpenRefine/OpenRefine/blob/cdbdf362952914f7284dc182084293f12bc9f3b9/extensions/database/tests/src/com/google/refine/extension/database/DatabaseImportControllerTest.java) using my IDE's test runner. I was using IntelliJ, but Eclipse should work as well. Without MySQL installed they won't all succeed, but they should give you enough insight to see what needs to be fixed.

The problem is not with your code, but with the test itself. It mocks `getParameter()`, but does not mock `getParameterMap()` which is what the new query parameter parser needs. Because there is other code which still depends on `getParameter()` they actually both need to be mocked and return the same data, so I'd suggest one or more helper methods be added to the test code.

We apparently don't have the database test environment setup documented for developers, which I will work on now, but you should be able to get away with just MySQL. Basically, you want to:
1. Install MySQL, then run the two commands that @wetneb mentioned above:
3. `mysql -u root -h 127.0.0.1 -P <port, if non default> -proot -e 'CREATE DATABASE test_db;' `
4. `mysql -u root -h 127.0.0.1 -P <port if non default>  -proot < extensions/database/tests/conf/test-mysql.sql 
5. Finally, remove this line from the test configuration file https://github.com/OpenRefine/OpenRefine/blob/cdbdf362952914f7284dc182084293f12bc9f3b9/extensions/database/tests/conf/tests.xml#L42

If you get stuck, or this is too much hassle, let me know and I can fix up the PR so that we can get it merged.

tfmorris:
Actually, I've got an easier solution. We'll just switch that test to use SQLite instead of MySQL. It doesn't look like I can push to your branch, so I've put the fix up in a branch on my repo. If you `git cherry-pick` this commit unto your branch, you should get test failures locally. https://github.com/tfmorris/OpenRefine/commit/89ee6050d08123155b927605827317c4a442ae2a

EliasStihl:
Once again, thank you for the incredible support! I think all tests pass now, but please let me know if there is anything you want me to change.

tfmorris (approved):
Looks great! Thank you for sticking with it! I'll merge it after #6412 gets merged so I can rebase it on top. I also have a couple of test helper methods for DatabaseImportControllerTest which will remove some of the duplication.

For future reference, it makes it easier for us to review if you push incremental fixes in separate commits rather than force pushing a completely new commit. That way we can easily separate stuff that's been reviewed already from the new stuff. All the commits will get squashed when we merge the PR, so there's no need to worry about the extra commits.

EliasStihl:
Okay, sorry about that, but thanks for telling me so I know how to do it properly in the future!

PR #6415, opened 2024-03-04
Fix swapped button labels in wikibase schema editor

### LINKED ISSUE #6329: Swapped button labels in Wikibase schema editor
The labels of the "Save schema changes" and "Save schema" buttons do not correspond to what the buttons actually do.

### Current Results

Clicking on the "Save schema changes…" button brings up a dialog to save the schema in the library (to be used in other projects).
The "Save schema" button at the bottom just saves the changes made to the schema into the project.

### Expected Behavior

The labels of the buttons should be swapped to match their actual behavior.
The button at the bottom saves changes in the project, and is logically located next to the button that discards those changes.

The button at the top is not meant to be used for saving incremental changes but rather saving a schema to be able to reuse it later: as such, it does not make sense to me to have the "changes" word in its label.

### Screenshots

image

### Versions
 - master branch

### Additional context
Introduced by #5935.

cc @Abbe98 @lozanaross

Redeem-Grimm-Satoshi:
Fixes #6329

- The labels of the `"Save schema changes"` and `"Save schema" `buttons have been swapped so they correspond to their actual use

@wetneb


wetneb (commented):
I think it would be simpler to swap the message ids in the Javascript code: this would let us retain all existing translations of those messages, without bothering translators to re-translate those messages again. What do you think about that?

Redeem-Grimm-Satoshi:
> I think it would be simpler to swap the message ids in the Javascript code: this would let us retain all existing translations of those messages, without bothering translators to re-translate those messages again. What do you think about that?

Yes you're right. I was just about to ask about the translation issue.

Swapping the message ids is the most feasible solution

Redeem-Grimm-Satoshi:
@wetneb are you talking about this message ID? or a different one..

id

wetneb:
I mean that instead of swapping the texts that are associated with message ids, you could rather swap the usages of the message ids. The message ids that I mean are the ones that are affected by your current changes (see https://github.com/OpenRefine/OpenRefine/pull/6415/files). Does that help?
At this point, I don't know how to describe this approach more precisely than by making the changes myself, but let me know if any particular aspect of this is unclear.

Redeem-Grimm-Satoshi:
@wetneb Thanks for your feedback, I have been looking at it and still having a hard time understanding this **" The message ids that I mean are the ones that are affected by your current changes "**

Because from the code above, what I know is` wikibase-schema/save-new-schema` is the message id for **"Save schema"**

Am I getting it all wrong?

wetneb:
Yes. You have correctly identified the two message ids affected by this change: `wikibase-schema/save-new-schema` and `wikibase-schema/save-button`.

What you have done so far is change the text that is associated to them in English.
I would prefer if you kept the text unchanged (so, revert your current changes). Instead go to the `.js` files where the message ids are used and swap the usages there. This means, you should swap `$.i18n("wikibase-schema/save-new-schema")` and `$.i18n("wikibase-schema/save-button")` in the Javascript files.

Redeem-Grimm-Satoshi:
@wetneb Thanks for clarifying, I understand now

Redeem-Grimm-Satoshi:
@wetneb I've swapped the message ids usage

wetneb (approved):

PR #6421, opened 2024-03-06
Add module names to About page. Fixes #5546

### LINKED ISSUE #5546: Show installed extensions in the GUI
When using a hosted or managed OpenRefine instance it isn't necessarily obvious which extensions are installed, and thus it can be tricky to know the available features or where to find relevant documentation.

### Proposed solution

It would be nice if the about page listed all installed extensions.

### Alternatives considered

Maybe there could be cases where a host doesn't want to show the installed extensions?

tfmorris:
Fixes #5546

- add Butterfly's list of loaded modules to the info returned by the GetVersion command
- Add that information to the About page HTML & update the Javascript to populate it
- also repair broken OpenRefine & JVM version display on About page

### DIFF
--- a/main/src/com/google/refine/commands/GetVersionCommand.java
+++ b/main/src/com/google/refine/commands/GetVersionCommand.java
@@ -62,10 +62,11 @@ public class GetVersionCommand extends Command {
         @JsonProperty("java_runtime_name")
         public String java_runtime_name = System.getProperty("java.runtime.name", "?");
         @JsonProperty("java_runtime_version")
-        // public String java_runtime_version = Runtime.getRuntime().version(); // Java 9 or later
-        public String java_runtime_version = System.getProperty("java.runtime.version", "?");
+        public String java_runtime_version = Runtime.getRuntime().version().toString();
         @JsonProperty("display_new_version_notice")
         public String display_new_version_notice = System.getProperty("refine.display.new.version.notice");
+        @JsonProperty("module_names")
+        public String[] modules = servlet.getModuleNames();
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


Abbe98:
Thank you @tfmorris! This is great for extension developers who can now use the API to see what other extensions are installed.

main/src/com/google/refine/commands/GetVersionCommand.java
```diff
@@ -62,10 +62,11 @@ protected class VersionResponse {
         @JsonProperty("java_runtime_name")
         public String java_runtime_name = System.getProperty("java.runtime.name", "?");
         @JsonProperty("java_runtime_version")
-        // public String java_runtime_version = Runtime.getRuntime().version(); // Java 9 or later
-        public String java_runtime_version = System.getProperty("java.runtime.version", "?");
+        public String java_runtime_version = Runtime.getRuntime().version().toString();
         @JsonProperty("display_new_version_notice")
         public String display_new_version_notice = System.getProperty("refine.display.new.version.notice");
+        @JsonProperty("module_names")
```
Abbe98, line 68: Given that "extensions" are the terminology used in the interface would it make sense to use the same in the API?

wetneb: The `main` module is a Butterfly module but not an extension, I would say, so the current serialization seems to be more faithful to what is being returned.

Abbe98: @wetneb, good point.

wetneb (approved):

tfmorris:
@Abbe98 glad you like it! -- and thanks for fixing things so that I get notification of your updates.

PR #6438, opened 2024-03-11
Added a new feature to import files using drag and drop. Requested in issue #6250

### LINKED ISSUE #6250: Support drag and drop when creating a project from local files
In the project creation screen, the user needs to click the "Browse" button to select files from their computer. Dragging and dropping a file there does not work.

### Proposed solution

Add support for dragging and dropping a file, as an alternative to clicking the "Browse" button. The drag and drop could be enabled for a reasonably wide rectangular area around the button (likely an existing `<div>` element). For instance, the area circled in red in this screenshot could be a candidate:

image

On top of this, the "Next" button is perhaps redundant: one could go to the next stage as soon as a file has been selected / dropped.

### Alternatives considered

We could also make it visually explicit that drag and drop is supported. Perhaps @cooperzoe has suggestions?

### Additional context

Brought up during @cooperzoe's onboarding with the tool.

Ahmed-Elgamel:
Added a new way to import files using drag and drop

Fixes #6250

Here is how the UI will look like after adding the drag and drop feature

Before dragging any file:

Screenshot 2024-03-11 at 8 38 46 PM

After dragging and dropping the files you need:

Screenshot 2024-03-11 at 8 39 45 PM

- Added new text below the choose files button
- Added a box to indicate the area to which the files will be dragged to
- Added some css to polish the look of this feature
- Modified some javascript to take the new file input feature into consideration

### DIFF
--- a/main/webapp/modules/core/scripts/index/default-importing-sources/import-from-computer-form.html
+++ b/main/webapp/modules/core/scripts/index/default-importing-sources/import-from-computer-form.html
@@ -1,17 +1,15 @@
 <form bind="form"><div class="grid-layout layout-normal"><table>
   <tr><td id="or-import-locate-files"></td></tr>
-  <tr><td><input type="file" multiple bind="fileInput1" name="upload" /></td></tr>
-  <tr><td id="or-drag-locate-files"></td></tr>
-  <tr><td>
-      <div class="file-area">
-        <input type="file" multiple bind="fileInput2" name="upload" required/>
+  <tr>
+    <td>
 
-        <div class="file-dummy">
-          <span class="default">Click to select a file, or drag it here</span>
-          <span class="success">Great, your file is selected</span>
-        </div>
+      <div class="file-input-container">
+        <input type="file" multiple id="fileInput" name="upload" class="file-input" />
+        <span class="file-input-text">Drag files here</span>
       </div>
 
-    </td></tr>
+    </td>
+  </tr>
   <tr><td><button bind="nextButton" class="button button-primary" type="button"></button></td></tr>
 </table></div></form>
+
--- a/main/webapp/modules/core/scripts/index/default-importing-sources/import-from-computer-form.html
+++ b/main/webapp/modules/core/scripts/index/default-importing-sources/import-from-computer-form.html
@@ -4,7 +4,7 @@
     <td>
 
       <div class="file-input-container">
-        <input type="file" multiple id="fileInput" name="upload" class="file-input" />
+        <input type="file" multiple bind="fileInput" name="upload" class="file-input" />
         <span class="file-input-text">Drag files here</span>
       </div>
 
--- a/main/webapp/modules/core/scripts/index/default-importing-sources/import-from-computer-form.html
+++ b/main/webapp/modules/core/scripts/index/default-importing-sources/import-from-computer-form.html
@@ -1,5 +1,17 @@
 <form bind="form"><div class="grid-layout layout-normal"><table>
   <tr><td id="or-import-locate-files"></td></tr>
-  <tr><td><input type="file" multiple bind="fileInput" name="upload" /></td></tr>
+  <tr><td><input type="file" multiple bind="fileInput1" name="upload" /></td></tr>
+  <tr><td id="or-drag-locate-files"></td></tr>
+  <tr><td>
+      <div class="file-area">
+        <input type="file" multiple bind="fileInput2" name="upload" required/>
+
+        <div class="file-dummy">
+          <span class="default">Click to select a file, or drag it here</span>
+          <span class="success">Great, your file is selected</span>
+        </div>
+      </div>
+
+    </td></tr>
   <tr><td><button bind="nextButton" class="button button-primary" type="button"></button></td></tr>
 </table></div></form>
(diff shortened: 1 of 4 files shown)


Ahmed-Elgamel:
Hi @wetneb ,is this how the feature is expected to be like.Glad to hear your feedback and do any necessary changes for the pr to be merged :)

wetneb:
Thank you very much!

If we are going for adding a big area like this, then I would be in favour as keeping the original "Browse" file input in it, so that people easily recognize it as being just one input field. In that case, I'd be tempted to go for a pure CSS solution as suggested by @Abbe98 in the issue. For instance, in your second screenshot above, it strikes me that the original file input still shows "No file selected" even though you have dropped a file already, which feels inconsistent. This is due to the use of two `<input />` elements, which seems not ideal for accessibility.

We should strive to make the area stick to the current look and feel of the interface. I don't think we have any similar drag-and-drop area to take inspiration from, but the border color used when the file has been selected strikes me as not being part of our existing color palette. If a change in color is needed (which I am not really sure of, once the two file inputs are merged together), then it would be good to pick something that is already used elsewhere. Perhaps one other shade of blue (see our [design system](https://www.figma.com/file/2fGTsMdExFEf4ReUNFzB0V/OpenRefine-Outreachy?type=design&node-id=131-315&mode=design)).

Another approach would be not to change the UI and simply enable drag and drop in an existing area, without highlighting it visually. This would cater for the cases of users spontaneously using drag and drop because they have been used to that from many other tools. This is what @cooperzoe did intuitively and prompted the issue #6250.
I think that would also be a good solution, which could simply follow @cooperzoe's mockup in the issue.

Ahmed-Elgamel:
Thank you for your feedback!
I do think that putting two input fields is not ideal for accessibility.Incorporating a pure CSS solution, as suggested by @Abbe98, sounds like a good approach to keep the "Browse" file input recognizable while also enabling drag-and-drop functionality.

Ahmed-Elgamel:
Now the UI looks like this.

Screenshot 2024-03-15 at 1 04 16 PM

main/webapp/modules/core/scripts/index/default-importing-sources/import-from-computer-form.html
```diff
@@ -1,5 +1,15 @@
 <form bind="form"><div class="grid-layout layout-normal"><table>
   <tr><td id="or-import-locate-files"></td></tr>
-  <tr><td><input type="file" multiple bind="fileInput" name="upload" /></td></tr>
+  <tr>
+    <td>
+
+      <div class="file-input-container">
+        <input type="file" multiple id="fileInput" name="upload" class="file-input" />
```
wetneb, line 7: I noticed that clicking the "Next »" button has no effect anymore as far as I can tell. This is likely related to your removal of the `bind` attribute on the `<input />`. Obviously it's quite critical to make sure this button is still functional :)

main/webapp/modules/core/styles/index/default-importing-sources.css
```diff
@@ -40,3 +40,26 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
+.file-input-container {
+    position: relative;
+}
+
+.file-input-text {
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-110%, -30%);
```
wetneb, line 61: How did you come up with such percentages? Will the text stay centered if the width of the box increases? I wonder if there would be a more reliable solution.

wetneb (commented):
Thanks a lot! That's a clear improvement in my opinion.
I tried it out and on my side (Firefox on Linux) it looks like this:
image

Anticipating that the "no files selected" text will vary depending on the user's locale and system settings, I would enlarge the frame to make sure it fits. I think using `min-width` instead of `width` would be safer, as it would let the area expand as needed.

Given that it's such a visible part of the UI, I would leave extra time for more people to chime in about the appearance. Pinging @OpenRefine/designers.

thadguidry:
@wetneb Agree that we need to ensure the frame adjusts well for localized description text within it.
Other than that, I think the design looks fine.

main/webapp/modules/core/styles/index/default-importing-sources.css
Ahmed-Elgamel: I came up with these values using trial and error.Will try to find a more reliable solution.

Ahmed-Elgamel: The commit 00f4dff makes the text centered along both x and y axis even when the box size increases.

main/webapp/modules/core/scripts/index/default-importing-sources/import-from-computer-form.html
```diff
@@ -1,5 +1,15 @@
 <form bind="form"><div class="grid-layout layout-normal"><table>
   <tr><td id="or-import-locate-files"></td></tr>
-  <tr><td><input type="file" multiple bind="fileInput" name="upload" /></td></tr>
+  <tr>
+    <td>
+
+      <div class="file-input-container">
+        <input type="file" multiple bind="fileInput" name="upload" class="file-input" />
+        <span class="file-input-text">Drag files here</span>
```
wetneb, line 8: The "Drag files here" text should be localized, via our `$.i18n()` utility.

wetneb (commented):
Aside from the small improvement required below, I would merge this soon.

wetneb (approved):

PR #6439, opened 2024-03-12
Fix signing of released artifacts in GitHub Actions

wetneb:
Following up to the upgrate of maven-gpg-plugin to 3.2.0 (#6435). See https://issues.apache.org/jira/projects/MGPG/issues/MGPG-90?filter=allopenissues.


cstamas (approved):
LGTM

PR #6441, opened 2024-03-13
First batch of importer and operation tests

wetneb:
This is backport of tests I wrote for importers and operations as part of my work on the new architecture in the 4.0 branch.

The goal is to improve test coverage of our core components as Java tests (and not just Cypress tests) and make it easier to validate that operation behavior is preserved in the new architecture. The tests will not be exactly the same, but the diffs should be minimal enough.

More tests are to come.


tfmorris (approved):
Since it's all test code, I only did a cursory review.

PR #6442, opened 2024-03-13
Fixes #6406. Corrects behavior of cell editing when confirming single cell changes with "Enter."

### LINKED ISSUE #6406: beta: deleting single cell contents and pressing enter repopulates textbox
The input box repopulates with cell's previous data after using the "edit" button to delete contents and then press enter to commit.

### To Reproduce
1. First, open a project and click "edit" on a cell that has text contents.
2. Then, delete the contents of the cell.
3. Finally, press enter to commit the changes to the cell

### Current Results
The cell contents are deleted, but the "edit" box immediately reappears with the cell's previous contents populating it. Pressing "escape" or clicking "cancel" will close the "edit" input box without re-entering the cell's content.

### Expected Behavior
Optimal: The cell's contents are deleted and the edit box closes.
Also acceptable: The cell's contents are deleted and the edit box remains open.

### Versions
 - Operating System: Windows 11 Pro 23H2, build 22631.3155
 - Browser Version: Version 122.0.6261.71 (Official Build) (64-bit)
 - JRE or JDK Version: 
 - OpenRefine: Version 3.8-beta1 [7fd4077]

### Datasets
 Contact me for a small sample set, but otherwise I have tried it with multiple projects and the issue is present for each

WR-Smiley:
Fixes #6406. Adds _focusBeforeEdit.focus() and MenuSystem.DismissAll() to onDone function in cell-ui.js to correct behavior.

_focusBeforeEdit and DismissAll also kept in main body of commit to maintain apply-to-all functionality; reordered for consistency throughout function. Specific placement in onDone required to maintain correct focus tracking.

### DIFF
--- a/main/src/com/google/refine/importers/JsonImporter.java
+++ b/main/src/com/google/refine/importers/JsonImporter.java
@@ -44,6 +44,7 @@ import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonParser.NumberType;
 import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.json.JsonReadFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.BooleanNode;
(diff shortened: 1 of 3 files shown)


wetneb (approved):
Thanks a lot!

For front-end regressions like this one, it is often interesting to have a [Cypress test](https://openrefine.org/docs/technical-reference/functional-tests) to make sure it does not break again in the future, but it seems that the bug isn't so easily reproducible in Cypress (perhaps due to some subtleties of focus handling in that environment?). So let's just merge it like this.

PR #6450, opened 2024-03-14
Fix inconsistent Extract/Apply button location

### LINKED ISSUE #6432: Inconsistent location of Extract / Apply button
The button Extract / Apply are not consistently positioned between the project creation and after doing some operations.

After the project creation, the buttons are located under the information module 
image

After the user performs at least one operation, the button moves up. 

image

I would expect the button to always be at the top.

Redeem-Grimm-Satoshi:
Fixes #6432

- Fix inconsistent Extract/Apply button location by making it stick to the top right

### DIFF
--- a/main/webapp/modules/core/scripts/project/history-panel.html
+++ b/main/webapp/modules/core/scripts/project/history-panel.html
@@ -6,9 +6,6 @@
   <p bind="or_proj_mistakes"></p>
   <p><a href="https://openrefine.org/docs/manual/running#history-undoredo" target="_blank"><b><span bind="or_proj_learnMore"></span></b></a></p>
 </div>
-<div class="history-panel-controls" bind="controlsDiv">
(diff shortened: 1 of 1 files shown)


Redeem-Grimm-Satoshi:
@wetneb

Redeem-Grimm-Satoshi:
@wetneb  I simply just moved the `controls div `to the top of the `helper div`, I worked at my end as you can see below;
https://github.com/OpenRefine/OpenRefine/assets/45304978/a4546b78-1d45-4829-90d9-b4e76b467dd0

Redeem-Grimm-Satoshi:
@wetneb I just noticed I didn't delete the previous `history-panel-controls` div and that is why those empty buttons were present at the top, I just removed it now and it works fine

wetneb (approved):
Thank you!

PR #6460, opened 2024-03-18
 Replace Properties with Map<String, String> in Exporters interfaces

### LINKED ISSUE #6419: Replace Properties with Map<String, String> in Exporters interfaces
This should be considered experimental. It's not clear that benefit will outweigh the cost of the API churn, but it depends on the proposed solution.

We would like to phase out usage of java.util.Properties in places where it's only being used because it was the only option at the time. One candidate is the Exporter API, specifically it's `StreamExporter` and `WriterExporter` subinterfaces (`UrlExporter` can be ignored since it isn't used anywhere). Each has a single method [`export()`](https://github.com/OpenRefine/OpenRefine/blob/f4878804960162ba6145941bfbe0d14e60133d9b/main/src/com/google/refine/exporters/StreamExporter.java)  which includes a `Properties`-typed `options` parameter in its API. These should be replaced with a Map<String, String> https://github.com/OpenRefine/OpenRefine/blob/f4878804960162ba6145941bfbe0d14e60133d9b/main/src/com/google/refine/exporters/StreamExporter.java#L45

### Alternatives considered
There are (at least) two possible ways to do this:
1. Introduce new StreamExporter2 and WriterExporter2 interfaces which are independent of the current interfaces
2. Have the new interfaces extend StreamExporter and WriterExporter
 
### Additional context

Acceptance criteria:
- all existing exporters, both bundled and those provided by extensions, continue to work without being modified or recompiled
- a clear migration path is defined from the old API to the new API for exporters which are not bundled, but are part of extensions

karindev:
Fixes #6419

This makes changes to the 4.0 branch since the interface is already modified due to other changes.

- Uses of `Properties` are replaced with `Map<String, String>` in the `export()` method of Exporters interfaces
- The classes implementing the interfaces are updated to match
- Methods `exportRows()` and `countColumnRows()` CustomizableTabularExporterUtilities are also changed to use `Map<String, String>` to accommodate the change
- Unit tests are updated to match all changed code and mocking is removed where unnecessary


wetneb:
Thanks a lot for this!
I think I intuitively agree with the idea that it's worth doing this sort of change on the occasion of a larger change of interface, so adding it on top of the data model changes feels right in that sense. On the other hand there is also a case for keeping breaking changes as atomic as possible and giving people the opportunity to migrate gradually. We are currently discussing how to schedule such changes on the forum:
https://forum.openrefine.org/t/scheduling-breaking-changes-we-have-on-our-radar/1410/4

In the light of this discussion, I would keep this PR open (probably for a long time) until we have some clarity on how we proceed.
From my side, I would like to express my support to your approach and my gratitude for your work - even if it might not be reflected in the formal status of this PR as soon as I would wish!

extensions/wikibase/tests/src/org/openrefine/wikibase/exporters/QuickStatementsExporterTest.java
```diff
@@ -89,7 +90,7 @@ public void testSimpleProject()
         Engine engine = new Engine(grid, EngineConfig.ALL_ROWS, 1234L);
 
         StringWriter writer = new StringWriter();
-        Properties properties = new Properties();
+        Map<String, String> properties = new HashMap<String, String>();
```
tfmorris, line 93:
```suggestion
        Map<String, String> properties = new HashMap<>();
```

extensions/wikibase/tests/src/org/openrefine/wikibase/exporters/QuickStatementsExporterTest.java
```diff
@@ -232,7 +233,7 @@ public void testNoSchema()
                 new Serializable[][] { { "c", "d" } });
         Engine engine = new Engine(grid, EngineConfig.ALL_ROWS, 1234L);
         StringWriter writer = new StringWriter();
-        Properties properties = new Properties();
+        Map<String, String> properties = new HashMap<String, String>();
```
tfmorris, line 236:
```suggestion
        Map<String, String> properties = new HashMap<>();
```

extensions/wikibase/tests/src/org/openrefine/wikibase/exporters/SchemaExporterTest.java
```diff
@@ -29,7 +30,7 @@ public void testNoSchema()
         Engine engine = new Engine(grid, EngineConfig.ALL_ROWS, 1234L);
 
         StringWriter writer = new StringWriter();
-        Properties properties = new Properties();
+        Map<String, String> properties = new HashMap<String, String>();
```
tfmorris, line 33:
```suggestion
        Map<String, String> properties = new HashMap<>();
```

main/src/main/java/org/openrefine/commands/project/ExportRowsCommand.java
```diff
@@ -68,9 +69,8 @@ public class ExportRowsCommand extends Command {
      * This command uses POST but is left CSRF-unprotected as it does not incur a state change.
      */
 
-    @SuppressWarnings("unchecked")
-    static public Properties getRequestParameters(HttpServletRequest request) {
-        Properties options = new Properties();
+    static public Map<String, String> getRequestParameters(HttpServletRequest request) {
+        Map<String, String> options = new HashMap<String, String>();
```
tfmorris, line 73:
```suggestion
        Map<String, String> options = new HashMap<>();
```

modules/refine-workflow/src/main/java/org/openrefine/exporters/StreamExporter.java
```diff
@@ -35,15 +35,15 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
...
+import java.util.Map;
 
 import org.openrefine.ProjectMetadata;
 import org.openrefine.browsing.Engine;
 import org.openrefine.model.Grid;
 
 public interface StreamExporter extends Exporter {
 
-    public void export(Grid grid, ProjectMetadata projectMetadata, long projectId, Properties options, Engine engine,
+    public void export(Grid grid, ProjectMetadata projectMetadata, long projectId, Map<String, String> options, Engine engine,
```
tfmorris, line 46: Compatibility can be preserved here by retaining the original method and adding the new method, both with default implementations.

modules/refine-workflow/src/test/java/org/openrefine/exporters/OdsExporterTests.java
```diff
@@ -115,7 +114,7 @@ public void exportSimpleOds() throws IOException {
 
         try {
             OdfDocument odfDoc = OdfDocument.loadDocument(new ByteArrayInputStream(stream.toByteArray()));
-            List<OdfTable> tables = odfDoc.getTableList();
+            List<OdfTable> tables = odfDoc.getTableList(true);
```
tfmorris, line 117: We usually try to avoid mixing unrelated changes together, so we'd normally want this deprecation update in a separate PR.

tfmorris (commented):
I want to reiterate @wetneb 's thanks for this work! I'm in the process of porting this to the current master and will preserve your commits for the bulk of the work. 

I was remiss in the original issue in not pointing out the best implementation strategy which is to use Java 8's default method implementations to fully preserve API compatibility.

No action is needed on these review comments, but I wanted to provide you with the feedback in case it's useful to you. I've already addressed them in the port to `master`. I've also split the PR in two so that we can be assured that the interface changes are 100% compatible before updating the exporters in a separate PR.

tfmorris:
I've adapted this to the `master` branch and split it into two pieces: API (#6796) & Implementation (#6797) which supersede this PR.

Thanks again for your contribution @karindev ! I've used your commits as the basis for those, so you'll receive credit when they're reviewed and merged.

PR #6461, opened 2024-03-18
Adds command for adding rows

### LINKED ISSUE #1855: Implement "Prepend rows" command from gokbutils
The gokbutils extension includes a command which adds blank rows to the start of a project. This is a request to implement that command into the core product using the same implementation as currently in the gokbutils

 https://github.com/ostephens/refine-gokbutils

steve-kasica:
Fixes #1855

- Adds a command for adding rows by prepending rows to an existing project.
- Adds a new model change for inserting new rows into an existing project.


wetneb:
Exciting! Congratulations on your first PR to OpenRefine :) Since that's part of a conversation with @tfmorris I'll let him review first. I am withdrawing the corresponding GSoC project proposal since you are quite close to having completed it already. It was probably a bit too light for GSoC anyway.

main/tests/server/src/com/google/refine/commands/row/AddRowsCommandTests.java
```diff
@@ -0,0 +1,26 @@
...
+    public void setUpCommand() {
+        command = new AddRowsCommand();
+    }
+
+    @Test
+    public void testCSRFProtection() throws ServletException, IOException {
+        command.doPost(request, response);
+        assertCSRFCheckFailed();
+    }
+
```
tfmorris, line 25: Would it be possible to add a test which demonstrates / tests the REST API for this command?

That would also help document what the format is.

tfmorris (changes requested):
This is awesome! Thank you very much.

I've made a few comments and suggestions, but my biggest question concerns the front end. Do you also have the UI pieces for this or should we create a separate issue to cover those?

For the backend code that's here, most of my comments are minor, but probably the most important is test coverage for the new command.

main/src/com/google/refine/commands/row/AddRowsCommand.java
```diff
@@ -0,0 +1,77 @@
...
+            Project project = getProject(request);
+
+            String rowDataParam = "rows";
+            if (!request.getParameterMap().containsKey(rowDataParam) || request.getParameter(rowDataParam) == null) {
+                throw new ServletException(String.format("Parameter \"%s\" is required", rowDataParam));
+            }
+            String[] rowData = request.getParameterValues(rowDataParam);
+
+            // TODO: load Rows with existing pool. Current implementation requires users to re-reconcile values
+            //   in a previously reconciled column.
```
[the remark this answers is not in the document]
steve-kasica: I don't think it'd be too hard to address. But, I'm unsure how I'd deserialize a Row with an existing Pool. Recon is one of those areas of the code that I'm still don't know very well. As it stands now, a POST request to insert a row into a project with a reconciled column results in an unreconciled string in that column. The user then has to re-reconcile the column to address the new row, which I think is undesirable from a UI/UX perspective. There's also an edge case to consider where a user insert a new row where the recon entity is not in the existing pool. 

I'll work on it, but @wetneb can advise on how the best way to access an existing Pool in this scenario?

main/src/com/google/refine/model/changes/RowAdditionChange.java
```diff
@@ -0,0 +1,96 @@
...
+import com.google.refine.util.Pool;
+
+public class RowAdditionChange implements Change {
+
+    private final List<Row> _additionalRows;
+
+    private final int _insertionIndex;
+    private static String _countField = "count"; // Field name for sizes of `_additionalRows` in `change.txt`
+    private static String _insertionIndexField = "index"; // Field name for index to insert new rows in `change.txt`
+    private static String eoc = "/ec/"; // end of change marker, not end of file, in `change.txt`
```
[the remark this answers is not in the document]
steve-kasica: Ok will change.

main/src/com/google/refine/commands/row/AddRowsCommand.java
```diff
@@ -0,0 +1,77 @@
...
+//            EngineConfig engineConfig) throws ServletException {
+//
+
+//
+
+//
+//        return
+//
+//    }
+
```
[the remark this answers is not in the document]
steve-kasica: 🤦 Ooops, will do.

main/src/com/google/refine/commands/row/AddRowsCommand.java
```diff
@@ -0,0 +1,77 @@
...
+                    .map(rowStr -> {
+                        try {
+                            return Row.load(rowStr, pool);
+                        } catch (Exception e) {
+                            throw new RuntimeException(e);
+                        }
+                    })
+                    .collect(Collectors.toList());
+
+            AbstractOperation op = new RowAdditionOperation(rows);
```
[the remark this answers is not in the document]
steve-kasica: Not really, I originally thought that this new model change would be used for row insertion and row interpolation operations. So a row interpolation operation would need to insert a new row block at any index in the project. As I think through the UI/UX part of this, I think it might be helpful to also expose the insertion index at the command. It might address an edge case where the user inserts a new row in paginated results. Ok will change.

main/tests/server/src/com/google/refine/commands/row/AddRowsCommandTests.java
steve-kasica: Completely agree, but I'm not sure how to do that with Mockito. I've only done API testing with [REST Assured](https://rest-assured.io/) / Postman. Is this something I'd write a Cypress test for?

steve-kasica:
I don't have UI code written yet, but I'm happy to work on it since I've already started thinking about some edge cases to handle w.r.t the UI. I don't mind keeping the UI work in this PR, and not creating a separate issue since it goes hand-in-hand.

main/src/com/google/refine/commands/row/AddRowsCommand.java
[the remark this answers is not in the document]
wetneb: Before we think about all that I would like to be clear on what is the use case (for an end user, using the web interface) behind supplying row data in this way. So far it's not clear to me how this is meant to be used. I think it would be simpler to concentrate on the feature of just adding blank rows first, for which this part is not needed.

If we do know we want to add rows with certain contents, determining how the user is meant to specify them (from the same project, duplicating them? from another project? input manually in some sort of dialog?) will help determine the right approach to specify this data in the API and therefore handle the recon part.

Ahmed-Elgamel:
Hi @steve-kasica ,I have the UI finished if you want I can push it to this pr. Also, after the UI it seems that whatever number of rows I give to the server it only add one row and it adds it to the beginning of the data not at the end.Is this behavior intended?

main/src/com/google/refine/commands/row/AddRowsCommand.java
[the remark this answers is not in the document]
tfmorris: If we're adding blank rows and the only unreconciled cells are the new blank ones, I think the current solution is fine. The wording of the TODO made me thing that the entire column would become unreconciled, which I don't believe is the case.

main/tests/server/src/com/google/refine/commands/row/AddRowsCommandTests.java
tfmorris: You can see an example of how to set up a request parameter for Mockito here: 
https://github.com/OpenRefine/OpenRefine/blob/2c7b5007e3a2dc496bf85c10a28cad864cff7651/main/tests/server/src/com/google/refine/commands/row/GetRowsCommandTest.java#L107-L109

You could use Cypress, but it's more straightforward to do it as a unit test.

tfmorris:
You can address the linter failures with the command `./refine lint`.

For the UI, let's keep things simple to start and just do "Add N rows at the beginning" (which was the original use case) and perhaps "Add N rows at the end." More complicated scenarios (other insertion points, non-blank lines, etc) we can leave for future enhancements.

main/src/com/google/refine/commands/row/AddRowsCommand.java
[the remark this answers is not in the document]
wetneb: I'd be in favour of not having support for loading non-blank rows in the back-end, if that feature isn't exposed in the UI and is known to be unfinished. That helps us avoid dead code.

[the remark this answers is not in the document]
steve-kasica: This demo that illustrates how new rows are typed as strings by default in a column that include reconciled cells. The same concept extends to other cell types: dates, numbers, null cells, booleans. 

https://github.com/OpenRefine/OpenRefine/assets/2933702/1c5bfc97-e4a3-4c23-9528-69defaebc9f2

steve-kasica:
Sorry I must have my GitHub notification settings not setup correctly because I wasn't aware of the activity on this PR from the last couple weeks. I was held up on trying to support multiple cell types in the UI, but @wetneb convinced me to ship what I have at the Meetup today.  Anyway, I've pushed those UI changes and will tackle the remaining unresolved changes from @tfmorris tomorrow.

steve-kasica:
> Hi @steve-kasica ,I have the UI finished if you want I can push it to this pr. Also, after I added the UI it seems that whatever number of rows I give to the server it only add one row and it adds it to the beginning of the data not at the end.Is this behavior intended?

@Ahmed-Elgamel Sorry I didn't see this until now. Is there somewhere I can see the UI you've built? I didn't run into those issues with the UI I just pushed, but I bet they stem from serializing new row data from the client and server.

steve-kasica:
Here's an example for adding blank rows, with the updated input suggested by @magdmartin during the meetup today.

https://github.com/OpenRefine/OpenRefine/assets/2933702/1ba79e5d-043d-47ec-b5c3-5885ce5f3b9d

wetneb:
Hi @steve-kasica, I wonder if your "thumbs up" reaction on the post above means a preference for:

1. sticking to the current scope of the PR (including editing), in which case you would like us to continue reviewing it in this state

OR

2. scaling down to only adding blank rows?

steve-kasica:
👍  to rolling back to only adding blank rows.

steve-kasica:
I've added a UI for adding blank rows and finished increasing test coverage. I found a bug when reverting changes when row are appended to the project. 

Here's a short demo: https://youtu.be/j4iEtsR0JdM

main/tests/server/src/com/google/refine/commands/row/AddRowsCommandTests.java
```diff
@@ -0,0 +1,210 @@
...
+        when(request.getParameter(AddRowsCommand.INDEX_PARAMETER)).thenReturn("0");
+        when(request.getParameterValues(AddRowsCommand.ROWS_PARAMETER)).thenReturn(additionalRows);
+
+        command.doPost(request, response);
+
+        JsonNode node = ParsingUtilities.mapper.readValue(writer.toString(), JsonNode.class);
+        assertNotNull(node.get("historyEntry"));
+        assertNotNull(node.get("historyEntry").get("id"));
+        assertNotNull(node.get("historyEntry").get("description"));
+        assertNotNull(node.get("historyEntry").get("time"));
```
wetneb, line 104: Would it make sense to merge the two tests `testSuccessResponseCode` and `testSuccessResponseHistoryEntrySchema` since they seem to be making different assertions in the same testing scenario? You could move your assertions about `historyEntry` to the first test above. Or is it intentional that they are separate?

main/tests/server/src/com/google/refine/operations/row/RowAdditionOperationTests.java
```diff
@@ -0,0 +1,65 @@
...
+    @BeforeTest
+    public void init() {
+        logger = LoggerFactory.getLogger(this.getClass());
+    }
+
+    @Test
+    public void testSerialization() throws IOException {
+        String json = "{"
+                + "\"op\":\"core/row-addition\","
+                + "\"description\":\"Add rows\"}";
```
wetneb, line 59: I'd expect the number of rows and the insertion index to be part of the JSON serialization of the operation, to make sure it can be executed again based on the operation metadata.

main/webapp/modules/core/scripts/dialogs/add-rows-dialog.js
```diff
@@ -0,0 +1,101 @@
...
+        "rows[]": data,
+        "index": index,
+      },
+      null,
+      {
+        "onDone": (o) => Refine.fetchRows(
+          theProject["rowModel"]["start"],
+          null,
+          () => ui.dataTableView.update(onDone),
+          null),
```
wetneb, line 96: Instead of manually triggering the update of the UI, I think it would be simpler to use the existing utilities we have to signal what the operation changed, and let the core system schedule those updates. This means you shouldn't need an `onDone` callback but simply pass flags to indicate which parts of the UI should be refreshed, like here:
https://github.com/OpenRefine/OpenRefine/blob/183246f6b0fde965d4d2d7bdcf6595dc97e857a9/main/webapp/modules/core/scripts/dialogs/column-reordering-dialog.js#L95
or here:
https://github.com/OpenRefine/OpenRefine/blob/183246f6b0fde965d4d2d7bdcf6595dc97e857a9/main/webapp/modules/core/scripts/views/data-table/menu-reconcile.js#L399

In this case I think `modelsChanged: true` should be necessary because you are changing the number of rows in the project. It should be sufficient because that triggers an update of basically everything.

That should have the effect of updating other parts of the UI: for instance, the facets.

main/webapp/modules/core/scripts/dialogs/add-rows-dialog.js
```diff
@@ -0,0 +1,101 @@
+const AddRowsDialog = (function (path) {
+  let level;
+  const $dialog = $(DOM.loadHTML("core", path));
+
+  class Row {
+    #isFlagged;
+    #isStarred;
+    #cells;
```
wetneb, line 8: I wonder if this class is necessary because since we are only supporting adding blank rows, we should be able to only send the number of blank rows (and not their JSON serialization) to the backend, no?

main/webapp/modules/core/scripts/dialogs/add-rows-dialog.js
```diff
@@ -0,0 +1,101 @@
+const AddRowsDialog = (function (path) {
```
wetneb, line 1: What is the purpose of declaring this as a function that gets applied immediately? Does it delay the evaluation of some instructions to later?

main/src/com/google/refine/model/changes/RowAdditionChange.java
```diff
@@ -0,0 +1,122 @@
...
+public class RowAdditionChange implements Change {
+
+    private final List<Row> _additionalRows;
+
+    private final int _insertionIndex;
+    private static String COUNT_FIELD = "count"; // Field name for sizes of `_additionalRows` in `change.txt`
+    private static String INDEX_FIELD = "index"; // Field name for index to insert new rows in `change.txt`
+    private static String EOC = "/ec/"; // end of change marker, not end of file, in `change.txt`
+
+    public RowAdditionChange(List<Row> additionalRows, int insertionIndex) {
```
wetneb, line 52: I wonder if it would make sense to only store the number of additional rows, since they are known to be blank. That would be simpler and would save memory.

On the other hand you could say that it's useful to make it possible to support non-blank rows at the change level for later, so that projects created with later versions of OpenRefine could also be read by earlier versions, but that's generally something that's too hard to support reliably to advertise it to users. Also, if we are to support arbitrary rows, we would probably need to serialize the recon pool too, so that rows can contain recon values.

No strong feelings either way.

wetneb (changes requested):
Thanks a lot! That looks a lot more manageable to get merged.

Here are a few screenshots to make it easier for others to chime in about the UI:
image
image

For the dialog, I'd change the "Submit Query" button to "OK", marking it as primary by adding the `button-primary` class and position it next to "Cancel" in the same way that we have normalized other dialogs in https://github.com/OpenRefine/OpenRefine/pull/6482 (it might be sufficient for that to just rebase your branch on top of the latest master branch).

main/tests/server/src/com/google/refine/commands/row/AddRowsCommandTests.java
steve-kasica: I see what you mean. Ok I'll merge those to one test to cover the response body of a successful response.

I initially separated them because I wanted `testSuccessResponseCode` to test the HTTP response status code, after past confusion about status codes (#5772). But, the mock for HttpServletResponse doesn't provide a real status code.

main/tests/server/src/com/google/refine/operations/row/RowAdditionOperationTests.java
steve-kasica: Good catch!

main/webapp/modules/core/scripts/dialogs/add-rows-dialog.js
steve-kasica: Awesome, I wish I knew about this sooner. Are these existing utilities documented anywhere? I really struggled figuring out how to maintain state between the client and server when writing JavaScript. I'll also write a Cypress test for an update to facets.

main/webapp/modules/core/scripts/dialogs/add-rows-dialog.js
steve-kasica: It's an older designed pattern in JavaScript called [immediately invoked function expressions (IIFE)](https://developer.mozilla.org/en-US/docs/Glossary/IIFE). JavaScript modules arguably make it obsolete now, but it actually works well with how scripts are bundled in OpenRefine. It prevents pollution of the global namespace and create private and public variables and methods as in the module pattern.

main/src/com/google/refine/model/changes/RowAdditionChange.java
steve-kasica: I have strong feelings in favor of the current implementation, sending the server rows that are blank. This design decision setups future development for supporting non-blank rows. I think adding rows of new data into an existing project is the primary task that this feature will address and padding a table with blank rows is an edge case, although [supporting expressive use of data tables is still important](https://www.tableau.com/research/publications/untidy-data-unreasonable-effectiveness-tables) and I understand the pragmatic importance of incremental enhancements in a widely used application such as OpenRefine.

I also think OpenRefine benefits from a new model change for a small subset of new rows, as per this TODO item in `MassRowChange`. `RowAdditionChange` begins to address it.
 
https://github.com/OpenRefine/OpenRefine/blob/183246f6b0fde965d4d2d7bdcf6595dc97e857a9/main/src/com/google/refine/model/changes/MassRowChange.java#L49-L50

main/webapp/modules/core/scripts/dialogs/add-rows-dialog.js
steve-kasica: Yes, we could do that, but I think in the future we would wish that designed this feature from the beginning to send rows that are blank, as opposed to the number of blank rows, see   https://github.com/OpenRefine/OpenRefine/pull/6461#discussion_r1576775598.

main/src/com/google/refine/model/changes/RowAdditionChange.java
tfmorris: I'm OK with row data which is required to be blank as long we we're clear about that restriction in the code comments and that the tests match whatever we're claiming to support (I haven't reviewed the tests yet).

tfmorris:
The dialog button changes look good. I've only done cursory testing so far, but my first attempt generated something weird which I wanted to report. I did: 1) prepend, 2) append, 3) append and got the following change history:
```json
[
  {
    "op": "core/row-addition",
    "rows": [
      {
        "starred": false,
        "flagged": false,
        "cells": []
      }
    ],
    "index": 0,
    "description": "Add rows"
  },
  {
    "op": "core/row-addition",
    "rows": [
      {
        "starred": false,
        "flagged": false,
        "cells": []
      }
    ],
    "index": 1623,
    "description": "Add rows"
  },
  {
    "op": "core/row-addition",
    "rows": [
      {
        "starred": false,
        "flagged": false,
        "cells": []
      }
    ],
    "index": 1629,
    "description": "Add rows"
  }
]
```

I suspect that this was because I was initially in record mode and switched to row mode between steps 2 & 3, since the project ended up with 1625 records and 1630 rows. Obviously append should use the row count, not the record count.

I'll do some more testing and code review shortly...

steve-kasica:
Thanks for catching this bug @tfmorris. I can reproduce it. I assumed `theProject.rowModel.total` would always provide the total *rows* in a project, even in records mode. The food dataset I was testing on before isn't a good example records-mode data. 

I also found another related issue. If I append a row in records mode, it inserts it in the middle of project because it's using record count instead of row count.

image

If I just replace `theProject.rowModel.total` with `theProject.metadata.rowCount`, would that fix that issue?

wetneb:
My understanding of the status of this PR is that @steve-kasica has responded to reviews. @tfmorris do you still feel responsible for reviewing this or should someone take over?

tfmorris:
@steve-kasica @wetneb Sorry for the delay! Reviewing now...

tfmorris (changes requested):
Looks great! I think we're really close. I ran lint on the code and pushed the diffs (you didn't see the failure because we hadn't approved the CI workflows to run).

I think the things that remain are:
- internationalization of a couple of strings in the pull down menu
- require (and test that) rows be empty

tfmorris:
@steve-kasica Are you still working on this? Are you able to address the test comments above? I don't think they'll be a ton of work, so I can probably add them to my todo list if necessary to get this pushed over the line.

steve-kasica:
Hi @tfmorris, 

Sorry for the delayed response. My expectations for how much time I can devote to this kind of work shifted in May, and I'm unfortunately not able to devote as much time over the summer to this kind of work. But, I'll try to find some time this week to implement these easy changes. If they're still unresolved next week, then please feel free to take the reins. Thanks!

tfmorris:
Awesome. Thank you for sticking with it!

main/webapp/modules/core/scripts/dialogs/add-rows-dialog.html
```diff
@@ -0,0 +1,50 @@
...
+                <div style="display:flex;">
+                    <label for="add-rows-position"
+                           bind="addRowsPositionLabel"
+                           style="flex:1"></label>
+                    <select name="add-rows-position"
+                            id="add-rows-position"
+                            data-cy="insert-position"
+                            style="flex:2;max-width:83px;">
+                        <option bind="optionBeginning"
+                                value="">Beginning</option>
```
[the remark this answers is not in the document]
steve-kasica: ✅ See [083f502](https://github.com/OpenRefine/OpenRefine/pull/6461/commits/083f5028c5597d9dcc22ad5bb227c9a5e0059445)

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -806,6 +806,14 @@
...
     "core-views/column-join-dont-escape": "In separator and nulls substitutes, use \\n for new lines, \\t for tabulation, \\\\n for \\n, \\\\t for \\t.",
     "core-views/cell-edit-date-help": "Input in ISO 8601 date format such as 1987-03-04T03:43:30Z",
+    "core-views/add-rows": "Add blank rows",
+    "core-views/add-rows/prepend-blank": "Prepend one row",
+    "core-views/add-rows/append-blank": "Append one row",
+    "core-views/add-rows/open-dialog": "Add multiple rows",
+    "core-views/add-rows/count-label": "Number of blank rows",
+    "core-views/add-rows/insertion-label": "Insertion point",
+    "core-views/add-rows/description": "Use the form to prepend new rows",
+    "core-views/add-rows/label": "Create more rows",
```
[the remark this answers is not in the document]
steve-kasica: ✅ See [083f502](https://github.com/OpenRefine/OpenRefine/pull/6461/commits/083f5028c5597d9dcc22ad5bb227c9a5e0059445)

main/tests/server/src/com/google/refine/commands/row/AddRowsCommandTests.java
```diff
@@ -0,0 +1,199 @@
...
+    // If rows parameter contains malformed JSON, `getRowData` throws JsonParseException
+    public void testMalformedJSONRows() throws Exception {
+        when(request.getParameter("csrf_token")).thenReturn(Command.csrfFactory.getFreshToken());
+        when(request.getParameter(AddRowsCommand.INDEX_PARAMETER)).thenReturn("0");
+        when(request.getParameterValues(AddRowsCommand.ROWS_PARAMETER))
+                .thenReturn(new String[] { "{ \"c\", \"3\" " }); // Intentionally missing closing brace
+
+        command.getRowData(request);
+    }
+
```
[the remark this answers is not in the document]
steve-kasica: Good catch, I assumed that syntactically-valid, semantically-invalid JSON would be something that the [`Row.load`](https://github.com/OpenRefine/OpenRefine/blob/abf4c991e7ab1da0009aa9da994615bc41ced9b1/main/src/com/google/refine/model/Row.java#L206) would handle. But it seems like you could pass anything valid JSON to `Row.load` and it'd give you a Row instance. I agree that just mandating that empty rows are "`{}`" is the fastest way forward.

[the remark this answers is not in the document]
steve-kasica: ✅ See [0abaa71](https://github.com/OpenRefine/OpenRefine/pull/6461/commits/0abaa71ddc4d80543891a8eff7b442aa5b6a3727)

main/tests/server/src/com/google/refine/commands/row/AddRowsCommandTests.java
```diff
@@ -0,0 +1,199 @@
...
+import com.google.refine.commands.Command;
+import com.google.refine.commands.CommandTestBase;
+import com.google.refine.model.Project;
+import com.google.refine.util.ParsingUtilities;
+
+public class AddRowsCommandTests extends CommandTestBase {
+
+    private Project project = null;
+    private String[] additionalRows = {
+            "{ starred: false, flagged: false, cells: []}",
```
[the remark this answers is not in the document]
steve-kasica: ✅ See [3d6a364](https://github.com/OpenRefine/OpenRefine/pull/6461/commits/3d6a36420a06aa8578240f4a6c1762280936bc0c)

main/src/com/google/refine/commands/row/AddRowsCommand.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public List<Row> getRowData(HttpServletRequest request) throws Exception {
+        String[] data = request.getParameterValues(ROWS_PARAMETER);
+        if (data.length == 0) {
+            throw new IllegalArgumentException("Parameter " + ROWS_PARAMETER + " is empty");
+        }
+        List<Row> rows = new ArrayList<>(data.length);
+        Pool pool = new Pool();
+        for (String rowStr : data) {
+            rows.add(Row.load(rowStr, pool));
```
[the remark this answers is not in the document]
steve-kasica: ✅ See [0abaa71](https://github.com/OpenRefine/OpenRefine/pull/6461/commits/0abaa71ddc4d80543891a8eff7b442aa5b6a3727)

steve-kasica:
@tfmorris Ok I've implemented those changes and rebased.

tfmorris (approved):
Great work! Thank you very much.

PR #6466, opened 2024-03-19
Second batch of operation tests

wetneb:
Follow-up to #6441, with the same motivation. This covers the remaining operations (recon-related and row-based).


tfmorris (approved):
RSLGTM

PR #6473, opened 2024-03-19
Removed grey background from button tag

### LINKED ISSUE #6440: Column menu button icon has acquired a hard grey box
This is pretty trivial and may have been a deliberate aesthetic choice so apologies if it doesn't constitute a bug, but while assessing OR 3.8-beta1 I noticed that the menu buttons next to the column names now have a hard grey box around them which, in my opinion, looks a little inelegant (at least compared to what it used to be).

3.8-beta1:
Screenshot 2024-03-13 at 11 25 50

3.7.9:
Screenshot 2024-03-13 at 11 26 08

abhishekkujur1307:
Fixes #6440

- `padding: 0` - Removes unwanted padding
- `background-color: transparent` - Converts grey background into transparent color

### DIFF
--- a/main/webapp/modules/core/styles/views/data-table-view.css
+++ b/main/webapp/modules/core/styles/views/data-table-view.css
@@ -186,6 +186,8 @@ table.data-table td.column-header, table.data-table th.column-header {
   background-repeat: no-repeat;
   cursor: pointer;
   border: none;
+  padding: 0;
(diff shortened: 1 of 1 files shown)


wetneb (approved):
It seems to work on my side, thank you! Tested both with Firefox and Chromium.

PR #6478, opened 2024-03-20
Provide slf4j-api & jackson to extension modules. Fixes #6475

### LINKED ISSUE #6475: Bundled extensions should declare slf4j and Jackson as "provided"
As described in https://github.com/OpenRefine/simile-butterfly/issues/15#issuecomment-2008558262 core facilities that are used by extensions should be declared as "provided" in the extensions POMs, so that they don't get re-imported with a different classloader hierarchy causing LinkageErrors. In particular, this applies to slf4j and jackson-databind.

### Current Results
Butterfly fails to start with OpenRefine extension modules unless its classloading isolation mechanism is disabled.

### Expected Behavior
All bundled extensions load without issue even with Butterfly's isolation enabled.

tfmorris:
Fixes #6475

- Add SLF4J & Jackson modules as "provided" to extension POMs

### DIFF
--- a/extensions/database/pom.xml
+++ b/extensions/database/pom.xml
@@ -109,6 +109,12 @@
     </dependency>
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>${jackson.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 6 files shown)


wetneb (approved):
I realize we could also do this for any transitive dependencies of an extension's dependencies that are common with `main`'s dependencies, assuming the versions match. That would help us avoid shipping the same jar multiple times.

PR #6480, opened 2024-03-21
Fix class mapping mechanism broken by the migration to Jackson

wetneb:
Follow-up of a discussion with @tfmorris.

### DIFF
--- a/main/tests/server/src/com/google/refine/ClassResolverTests.java
+++ b/main/tests/server/src/com/google/refine/ClassResolverTests.java
@@ -1,0 +1,29 @@
+
+package com.google.refine;
+
+import org.testng.annotations.Test;
+
+import com.google.refine.preference.TopList;
+import com.google.refine.util.ParsingUtilities;
+import com.google.refine.util.TestUtils;
+
+public class ClassResolverTests {
+
+    String oldJson = "{\"class\":\"com.google.gridworks.preference.TopList\",\"top\":100,\"list\":[]}";
+    String newJson = "{\"class\":\"com.google.refine.preference.TopList\",\"top\":100,\"list\":[]}";
+
+    @Test
+    public void deserializeClassWithOldName() throws Exception {
+        TopList topList = ParsingUtilities.mapper.readValue(oldJson, TopList.class);
+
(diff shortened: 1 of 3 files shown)


wetneb:
@tfmorris any thoughts on this one?

tfmorris:
@wetneb Sorry for the delay. I'll try to review this shortly.

main/tests/server/src/com/google/refine/ClassResolverTests.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.testng.annotations.Test;
+
+import com.google.refine.preference.TopList;
+import com.google.refine.util.ParsingUtilities;
+import com.google.refine.util.TestUtils;
+
+public class ClassResolverTests {
+
+    String oldJson = "{\"class\":\"com.google.gridworks.preference.TopList\",\"top\":100,\"list\":[]}";
```
tfmorris, line 12: Are there actually projects in the wild with this embedded class name? I didn't think that the "Google Refine" era and Jackson JSON era overlapped at all (and I thought there was no embedded class name prior to Jackson).

I'm going to see if I can dig up one or more vintage Google Refine / OpenRefine projects to test with.

tfmorris (commented):
Yet another apology, but I *am* actively reviewing it now -- and I *do* appreciate the effort you put into researching and implementing a solution. Just a question (above) to start while I hunt for some legacy projects. 

It looks like this mechanism would help with the com.google.refine -> org.openrefine case (assuming that we can't get rid of the embedded class names), so that's useful, but I'd really prefer not to have classnames embedded at all since it seems inherently fragile (and I don't think legacy projects have them).

main/tests/server/src/com/google/refine/ClassResolverTests.java
tfmorris: My bad! It looks like that class name has been buried in there since 2010. I still don't love it and would want to think through the various potential failure modes, particularly if it was used by a custom extension, as appears to have been the intention.

tfmorris (approved):

Issue #6353, opened 2024-02-08
OK & CANCEL button placement


cooperzoe:
In the newest version of OpenRefine, I notice the OK button is on the left hand side and the CANCEL button on the right in this window. I expected the OK button to be on the right hand side and found it's unusual placement counterintuitive.

I propose we either switch their placements, or put OK in the righthand corner and keep cancel next to it. If we choose the latter option, perhaps we could make the OK button blue, so they don't look too similar to one another.

<img width="1177" alt="Screenshot 2024-02-08 at 12 31 38 PM" src="https://github.com/OpenRefine/OpenRefine/assets/52141007/bd5e1f3b-0684-4b57-9db4-13e56c6ab3ba">

wetneb:
I would be in favor of agreeing one one convention and normalize all dialogs to follow that convention.

My vote goes to have both buttons on the right-hand side and the OK button to be marked as "primary" (meaning that its text will appear in bold).

tfmorris:
I agree that the current placement is unusual. I think the application should follow the UI conventions of the underlying platform. They all have style guides and standard dialogs which implement those style guides (and perhaps they're all the same, but if not I think we should track the differences).

areebqazi:
Hello is anyone doing this issue or can I work on it ?

tfmorris:
@areebqazi Thanks for you interest, but we still need to come up with a firm design for this before it's ready for work.

tfmorris:
Although Jakob Nielsen says to follow platform conventions, except for web interfaces, I'd argue that we actually have enough information available about the user environment that we're closer to the platform case than the "web" case. He also has some input on button styling.

https://www.nngroup.com/articles/ok-cancel-or-cancel-ok/
https://www.lukew.com/ff/entry.asp?571
https://ux.stackexchange.com/questions/1072/ok-cancel-on-left-right

vr-varad:
so are the changes required in the buttons??

wetneb:
@tfmorris do you mean we should detect the operating system and adapt the button placement based on that?
I'd rather go with a uniform style regardless of the OS, for simplicity.

tfmorris:
> do you mean we should detect the operating system and adapt the button placement based on that? 

Yes, I think that would give the best user experience.

> I'd rather go with a uniform style regardless of the OS, for simplicity.

[...]

PR #6482, opened 2024-03-24
Issue 6353 implement Windows ordering buttons ("OK" / "Cancel")

zyadtaha:
Fixes [#6353](https://github.com/OpenRefine/OpenRefine/issues/6353)

- Implemented Windows ordering buttons ("OK" / "Cancel")
- Marked the OK button  as "primary" 

Here's how it looks like :
Screenshot from 2024-03-24 13-29-28

### DIFF
--- a/extensions/wikibase/module/scripts/dialogs/wikibase-dialog.html
+++ b/extensions/wikibase/module/scripts/dialogs/wikibase-dialog.html
@@ -14,7 +14,7 @@
       <a class="button" bind="discoverManifestsButton" href="https://github.com/OpenRefine/wikibase-manifests" target="_blank"></a>
     </div>
     <div class="dialog-footer-bg">
-      <button class="button" bind="cancelButton"></button>
+      <button class="button okButton" bind="cancelButton"></button>
       <button class="button" bind="okButton"></button>
     </div>
   </div>
--- a/extensions/wikibase/module/scripts/dialogs/wikibase-dialog.html
+++ b/extensions/wikibase/module/scripts/dialogs/wikibase-dialog.html
@@ -14,8 +14,8 @@
       <a class="button" bind="discoverManifestsButton" href="https://github.com/OpenRefine/wikibase-manifests" target="_blank"></a>
     </div>
     <div class="dialog-footer-bg">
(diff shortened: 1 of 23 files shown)


wetneb (commented):
Very nice, thank you! I wonder if you considered re-using the existing CSS class `button-primary`? If so, why did you choose to introduce `okButton` instead?

zyadtaha:
Yes, you are right it's better to reuse the existing class.
Sorry for the mistake, had fixed it.

wetneb (commented):
I have had a quick look and the migration looks complete (although there can be places I have overlooked too).
I like that the order we went for seems to be consistent with the majority of existing dialogs so it shouldn't come at a big surprise to users.

extensions/wikibase/module/scripts/dialogs/wikibase-dialog.html
```diff
@@ -15,7 +15,7 @@
     </div>
     <div class="dialog-footer-bg">
       <button class="button" bind="cancelButton"></button>
-      <button class="button" bind="okButton"></button>
+      <button class="button button-primary" bind="okButton"></button>
```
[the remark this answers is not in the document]
zyadtaha: Done.

wetneb (approved):

main/webapp/modules/core/styles/reconciliation/recon-dialog.css
```diff
@@ -104,7 +104,6 @@ a.recon-dialog-service-selector-remove {
 }
 .dialog-footer {
```
tfmorris, line 105: Sorry I missed the review period on this, but I don't think CSS for a reconciliation dialog should be modifying *any* of the common styles. In my opinion, none of the modifications to `.dialog-footer`, `.button-left`, and `.button-right` don't belong here. The problem is that this file is always loaded, so the definitions here are going to overwrite the intended definitions where they are first defined.

Perhaps @ayushrai206 can clarify the intent for the original changes.

wetneb: Sorry that it escaped me in the review, it's worth fixing indeed!

PR #6484, opened 2024-03-26
Drop 'sample' extension.

### LINKED ISSUE #2300: Drop the sample extension
We have a `sample` extension which is meant to demonstrate the basic scaffolding to get a minimalistic extension to work.

This is not super useful since we already have quite a lot other extensions in the repository, which are actually maintained since they expose functionality that people use. There is a risk is that this extension does not get updated (because breaking it does not actually break anything for users).

Also, as an extension developer, I generally want to develop my extension in its own repository, with its own build system. The Maven set up for extensions included in this repository is not easy to translate to extensions developed outside the repository, since our own set up relies on the main OpenRefine classes being present in the reactor at build time.

Once #2254 is done, it should be easier to propose a clean Maven setup for external extensions. We could move the sample extension to a dedicated repository, which was created a long time ago but remained empty:
https://github.com/OpenRefine/refine-sample-extension

wetneb:
Closes #2300.

Marked as draft for now because it will require adaptation to the extension writing guide at https://openrefine.org/docs/technical-reference/writing-extensions


wetneb:
Now that there is an accompanying documentation change, this is ready:
https://github.com/OpenRefine/openrefine.org/pull/333

tfmorris (approved):

PR #6486, opened 2024-03-26
Drop legacy Keyer classes not registered as clustering methods

### LINKED ISSUE #6485: Drop keying methods which are not exposed to users
In the backend we have a few classes which implement clustering methods that are purposefully not registered, because they are known to perform worse than other ones (see #1906).

I propose we drop this dead code.

### LINKED ISSUE #6495: loosing rows in a project
I have a proyect of 24078 rows i am working with. 
image

But somehow, after moving backwards unintentionally with the touch pad of my mac, I went back to the OpenRefine hub and when I enter again into the project it has less rows. 
image

And it continues to be like that. There is no sing in the history of such a delete of rows...
image

Do not undestand what happened. I am loosing hours of work here.

wetneb:
Closes #6485.

Those classes were still used in the phonetic() GREL function so I have replaced that by direct calls to the underlying Commons encoding library. We could probably drop support for those encoding methods in this GREL function too, but I have kept them for now as I don't have visibility of their usage and the problems affecting those encodings.

This is a preparation for Maven modularization changes, weakening the dependency between GREL and clustering.


main/src/com/google/refine/expr/functions/strings/Phonetic.java
```diff
@@ -71,32 +93,27 @@ public Object call(Properties bindings, Object[] args) {
...
+                return getMetaphone().metaphone(str);
             } else if ("soundex".equalsIgnoreCase(encoding)) {
-                return soundex.key(str);
-            } else if ("cologne".equalsIgnoreCase(encoding)) {
-                return cologne.key(str);
+                return getSoundex().soundex(str);
             } else {
-                // + " doesn't know how to handle the '"
-                // + encoding + "' encoding.");
+                Keyer keyer = KeyerFactory.get(encoding.toLowerCase());
```
tfmorris, line 110: This is going to return some keyers which aren't really phonetic in nature (n-gram, fingerprint). Not sure if we want to exclude them, but it's probably not a big deal either way.

main/src/com/google/refine/expr/functions/strings/Phonetic.java
```diff
@@ -48,12 +49,33 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
 
 public class Phonetic implements Function {
 
-    // TODO: We could probably lazily initialize these when needed for efficiency
-    static private Metaphone3Keyer metaphone3 = new Metaphone3Keyer();
-    static private DoubleMetaphoneKeyer metaphone2 = new DoubleMetaphoneKeyer();
-    static private MetaphoneKeyer metaphone = new MetaphoneKeyer();
-    static private SoundexKeyer soundex = new SoundexKeyer();
-    static private ColognePhoneticKeyer cologne = new ColognePhoneticKeyer();
+    // TODO deprecate and drop those legacy encodings?
```
tfmorris, line 52: Soundex is pretty widely used historically (e.g. US Census indexes). I'm not sure how popular it is among OpenRefine users.

main/src/com/google/refine/expr/functions/strings/Phonetic.java
```diff
@@ -48,12 +49,33 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
 
 public class Phonetic implements Function {
```
tfmorris, line 50: This might be a good time to make sure that we've got at least cursory test coverage of the `phonetic` function. All we've got currently are a few invalid argument tests.

tfmorris (changes requested):
I agree in principle. Just a few minor comments. The only thing that really needs to be addressed is to add a few lines of test coverage. Doesn't need to be exhaustive, but just demonstrate that the basic functionality works.

main/src/com/google/refine/expr/functions/strings/Phonetic.java
wetneb: Yes… it also gives access to any keyer that an extension registers, which is nice I would say, no? People would still need to discover its identifier though.

main/src/com/google/refine/expr/functions/strings/Phonetic.java
wetneb: done

tfmorris (approved):

PR #6502, opened 2024-04-03
Decouple expression languages from core infrastructure

wetneb:
Migrates the registration of expression languages provided by default to the controller.js, and removes references to GREL from functionally independent areas.

This is a first step towards having GREL in its own Maven module.

### DIFF
--- a/main/src/com/google/refine/LookupCacheManager.java
+++ b/main/src/com/google/refine/LookupCacheManager.java
@@ -37,7 +37,6 @@ import java.util.Map;
 import com.google.refine.expr.ExpressionUtils;
 import com.google.refine.expr.HasFieldsListImpl;
 import com.google.refine.expr.WrappedRow;
-import com.google.refine.expr.functions.Cross;
 import com.google.refine.model.Column;
 import com.google.refine.model.Project;
 import com.google.refine.model.Row;
@@ -49,6 +48,8 @@ import com.google.refine.util.LookupException;
  * @author Lu Liu
  */
 public class LookupCacheManager {
+
+    public static final String INDEX_COLUMN_NAME = "_OpenRefine_Index_Column_Name_";
 
     protected final Map<String, ProjectLookup> _lookups = new HashMap<>();
 
@@ -111,7 +112,7 @@ public class LookupCacheManager {
         }
 
         // if this is a lookup on the index column
-        if (lookup.targetColumnName.equals(Cross.INDEX_COLUMN_NAME)) {
+        if (INDEX_COLUMN_NAME.equals(lookup.targetColumnName)) {
             for (int r = 0; r < targetProject.rows.size(); r++) {
                 lookup.valueToRowIndices.put(String.valueOf(r), Collections.singletonList(r));
             }
--- a/main/src/com/google/refine/model/recon/ReconciledDataExtensionJob.java
+++ b/main/src/com/google/refine/model/recon/ReconciledDataExtensionJob.java
@@ -56,7 +56,8 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
-import com.google.refine.expr.functions.ToDate;
+import com.google.refine.expr.util.CalendarParser;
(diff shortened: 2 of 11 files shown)


wetneb:
@tfmorris any interest in this? I'd like to merge it to be able to start the Maven modularization PR.

main/src/com/google/refine/LookupCacheManager.java
```diff
@@ -50,6 +49,8 @@
  */
 public class LookupCacheManager {
 
+    public static final String INDEX_COLUMN_NAME = "_OpenRefine_Index_Column_Name_";
```
tfmorris, line 52: This seems unrelated. Is it intended to be included here? The change itself seems fine, but seems odd to have here.

main/tests/server/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperationTests.java
```diff
@@ -101,9 +101,8 @@ public class ColumnAdditionByFetchingURLsOperationTests extends RefineTest {
             "    \"status\" : \"pending\"\n" +
             " }";
 
-    @Override
     @BeforeTest
-    public void init() {
+    public void initOperation() {
```
tfmorris, line 105: This seems unrelated

main/src/com/google/refine/model/recon/ReconciledDataExtensionJob.java
```diff
@@ -250,11 +251,13 @@ protected ReconciledDataExtensionJob.DataExtension collectResult(
...
                     storeCell(rows, rowindex, colindex, v);
                 } else if (val.has("date")) {
-                    ToDate td = new ToDate();
-                    String[] args = new String[1];
-                    args[0] = val.get("date").asText();
-                    Object v = td.call(null, args);
-                    storeCell(rows, rowindex, colindex, v);
+                    Object date;
+                    try {
+                        date = CalendarParser.parseAsOffsetDateTime(val.get("date").asText());
```
tfmorris, line 256: This also seems unrelated, but even if it gets moved to another PR, I don't think we should be introducing direct calls to CalendarParser which we should be targeting to get rid of. Who's date format is this? If it's something like an ISO date, we should just use the Java classes directly. If it's a human date, it should probably stay wrapped in `ToDate()`

tfmorris (changes requested):
This seems like 3 or 4 unrelated sets of changes lumped together, but perhaps I'm not fully understanding what's going on. The core parser changes seem fine (with the addition of the `since = ` on the deprecation annotation).

main/src/com/google/refine/LookupCacheManager.java
wetneb: It looks like I should have explained the whole thing better, sorry. This change is intentional. Having this constant declared in `Cross` constitutes a dependency from `LookupCacheManager` (a core class) to `Cross` (a GREL function), which will prevent the separation of the GREL implementation as a separate Maven module which would depend on a core module.

main/tests/server/src/com/google/refine/operations/column/ColumnAdditionByFetchingURLsOperationTests.java
wetneb: If I remember correctly, it's a necessary change for the tests to pass after adding the GREL registration to `RefineTest` instead of being done statically (otherwise, this `init` method overrides that of `RefineTest` which doesn't get executed for this test).

main/src/com/google/refine/model/recon/ReconciledDataExtensionJob.java
wetneb: The motivation is still the same: removing a dependency from a core class to a GREL function. Using `CalendarParser` means that we support almost the same sort of date formats as before, which could be useful for services which rely on that.
But that being said now that the specs mandate ISO dates (although only in the JSON schemas so far, we should fix that) it should probably be safe to switch to supporting ISO dates only.

tfmorris: Sorry for the delay in circling back on this. I think using Java classes to support ISO dates (only) would be a better approach.

tfmorris (approved):
Thanks. LGTM

Issue #4303, opened 2021-11-17
Enable Chunked Uploads for Wikimedia Commons uploads, enabling >100MB file uploads


trnstlntk:
Wikimedia Commons deals with large files in a variety of ways; see https://commons.wikimedia.org/wiki/Commons:Maximum_file_size for some context and pointers.

Some end users of Wikimedia Commons batch upload tools have pointed out that they want to upload batches of larger files (e.g. book scans, video files), where individual files can be 1 to 3 GB or more in size. Such large file sizes are currently not handled very well or straightforwardly in existing upload tools like UploadWizard or Pattypan. 
It would be desirable if OpenRefine would handle these better, and/or at least communicate clearly which issues exist around such files during upload.

trnstlntk:
Status update while going through older issues and looking at our current [timeline for the Wikimedia-funded project on SDC support](https://meta.wikimedia.org/wiki/Grants:Project/Structured_Data_on_Wikimedia_Commons_functionalities_in_OpenRefine/Timeline). 

We are working towards a Minimum Viable Product before end June and end October 2022. At this point it is still quite unclear to us how OpenRefine will deal with uploading large files to Wikimedia Commons. As I see it now: before the Wikimedia grant concludes (October 2022), I think it is feasible to do some stress testing with uploading large files. However, if we discover that they pose major problems, it's unclear if we'll be able to address these before October 2022. It's quite possible that additional work will be needed after that.

wetneb:
This is a follow-up task to #4682.

Abbe98:
This likely requires upstream work as the reason why UploadWizard, Pattypan, etc does not handle this very well is that Wikimedia Commons does not report the progress of chunked uploads(T309094) and that it can take several hours for Commons to publish large files.

The option I have considered for Pattypan is to display a warning suggesting that one request server-side uploads or snail-mails a disk of files to Wikimedia.

trnstlntk:
@Vesihiisi has done some uploads of TIFF files some time ago with OpenRefine and was unsuccessful, although they could upload them with UploadWizard. Soon (August-September 2022?) they will have more files to test with, if we want to work on this. It may involve doing something with [Chunked Uploads](https://commons.wikimedia.org/wiki/Commons:Chunked_uploads), as I understand it correctly, but I encourage others to chime in with more feedback and pointers.

trnstlntk:
[...]

PR #6509, opened 2024-04-04
Upload files in chunks

sebastian-berlin-wmse:
Adds function for uploading a file in chunks as described on https://www.mediawiki.org/wiki/API:Upload#Example_3:_Upload_file_in_chunks. The chunks are created as temporary files and are removed once uploaded.

For #4303


sebastian-berlin-wmse:
Replaced by #6510.

PR #6510, opened 2024-04-04
Upload files in chunks

sebastian-berlin-wmse:
Adds function for uploading a file in chunks as described on https://www.mediawiki.org/wiki/API:Upload#Example_3:_Upload_file_in_chunks. The chunks are created as temporary files and are removed once uploaded.

For #4303

### DIFF
--- a/extensions/wikibase/pom.xml
+++ b/extensions/wikibase/pom.xml
@@ -164,7 +164,12 @@
       <scope>test</scope>
     </dependency>
 
+  
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>2.6</version>
+    </dependency>
   </dependencies>
 
 </project>
-
--- a/extensions/wikibase/pom.xml
+++ b/extensions/wikibase/pom.xml
@@ -168,7 +168,7 @@
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
-      <version>2.6</version>
+      <version>${commons-io.version}</version>
     </dependency>
   </dependencies>
 
--- a/extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
@@ -349,7 +349,8 @@ public class MediaFileUtils {
                         "The file upload action returned the '" + result + "' error code. Warnings are: " + Objects.toString(warnings));
             }
             if (filename == null && filekey == null) {
-                throw new MediaWikiApiErrorException(result, "The MediaWiki API did not return any filename for the uploaded file");
+                throw new MediaWikiApiErrorException(result,
+                        "The MediaWiki API did not return any filename or filekey for the uploaded file");
             }
         }
 
--- a/extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
@@ -411,7 +411,10 @@ public class MediaFileUtils {
             }
 
             String fileName = "chunk-" + chunksRead + "-";
-            BoundedInputStream inStream = new BoundedInputStream(stream, chunkSize);
+            BoundedInputStream inStream = BoundedInputStream.builder()
+                    .setInputStream(stream)
+                    .setMaxCount(chunkSize)
+                    .get();
             File chunk = Files.createTempFile(fileName, getExtension()).toFile();
             OutputStream outStream = new FileOutputStream(chunk);
             bytesRead += inStream.transferTo(outStream);
--- a/extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
@@ -4,10 +4,11 @@ package org.openrefine.wikibase.editing;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.net.URL;
 import java.nio.file.Files;
-import java.nio.file.Path;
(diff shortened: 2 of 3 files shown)


extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -306,4 +370,55 @@ public MediaInfoIdValue getMid(ApiConnection connection, String siteIri) throws
...
+         */
+        public File readChunk() throws IOException {
+            if (bytesRead >= path.length()) {
+                return null;
+            }
+
+            // Read at most the remaining bytes.
+            int bytesToRead = (int) Math.min(path.length() - bytesRead, chunkSize);
+            byte[] bytes = new byte[bytesToRead];
+            int chunkBytesRead = stream.read(bytes);
```
wetneb, line 406: With this workflow you are reading the entire chunk in memory, before writing it back to the file. I think we should be able to avoid the memory usage, no?
Intuitively, if you wrap `stream` inside a [`BoundedInputStream`](https://commons.apache.org/proper/commons-io/javadocs/api-2.4/org/apache/commons/io/input/BoundedInputStream.html) and then call [`transferTo()`](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/io/InputStream.html#transferTo-java.io.OutputStream-) on it, it should do just that. But I haven't tried it, so it can well be that I am missing a detail there.

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -102,6 +106,62 @@ public MediaUploadResponse uploadLocalFile(File path, String fileName, String wi
...
+                // In the first request we don't have offset or file key.
+                parameters.put("offset", "0");
+            } else {
+                parameters.put("offset", String.valueOf(response.offset));
+                parameters.put("filekey", response.filekey);
+            }
+            Map<String, ImmutablePair<String, java.io.File>> files = new HashMap<>();
+            String chunkName = "chunk-" + i + ".png";
+            files.put("chunk", new ImmutablePair<String, File>(chunkName, chunk));
+            response = uploadFile(parameters, files);
```
wetneb, line 148: After calling `uploadFile` it's probably worth calling `response.checkForErrors()` so that we can abort the upload early if uploading any of the chunk fails.

wetneb (commented):
Looking great! Excited that this is going forward :)

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -102,6 +106,62 @@ public MediaUploadResponse uploadLocalFile(File path, String fileName, String wi
...
+            parameters.put("filesize", String.valueOf(path.getLength()));
+            if (response == null) {
+                // In the first request we don't have offset or file key.
+                parameters.put("offset", "0");
+            } else {
+                parameters.put("offset", String.valueOf(response.offset));
+                parameters.put("filekey", response.filekey);
+            }
+            Map<String, ImmutablePair<String, java.io.File>> files = new HashMap<>();
+            String chunkName = "chunk-" + i + ".png";
```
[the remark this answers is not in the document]
sebastian-berlin-wmse: I don't know if the file extension is necessary, but I've seen it in the examples. Now it uses the same as the original file at least.

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
sebastian-berlin-wmse: I did as you suggested and it seems to work fine.

To only remaining thing is that if the BoundedInputStream should be closed according to my IDE. I've tried with both calling `closed` and adding a `try` for it, but it looks like that closes the wrapped FileInputStream; I get "java.io.IOException: Stream Closed" when calling `transferTo()` for the second chunk.

An ideas how to solve this?

wetneb (commented):
Nice, that looks great to me!

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
wetneb: I think we just need to make sure that the `FileInputStream` gets closed once we are done, but not the `BoundedInputStream`.
I would make `ChunkedFile` implement the `Closeable` interface, delegating its `close()` method to the FileInputStream. Then it will be up to whoever creates a `ChinkedFile` to also close it afterwards (for instance with a try-with-resources block).

extensions/wikibase/pom.xml
```diff
@@ -164,7 +164,12 @@
       <scope>test</scope>
     </dependency>
 
+  
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>2.6</version>
```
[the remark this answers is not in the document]
sebastian-berlin-wmse: Thanks. This was added by the IDE, I don't really know how it works🙂

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
sebastian-berlin-wmse: After https://github.com/OpenRefine/OpenRefine/pull/6510/commits/4d5614e080ac5b6069b891ce399b14afd5f2efb6 I no longer have a warning about closing the BoundedInputStream. Do you know if I still need to do that or if it's handled by the new way of creating it?

wetneb: Yes I still think it would be worth making `ChunkedFile` implement the `Closeable` interface, so that we can properly close the `FileInputStream` that it allocates in its constructor. Otherwise, it can be that the underlying file descriptor does not get closed, which would constitute a resource leak.

sebastian-berlin-wmse: done

wetneb (approved):
Great! All good for me :)

PR #6511, opened 2024-04-04
Switch release notification to openrefine.org and display other events

### LINKED ISSUE #6017: Change notification to download latest version to official download page
When a user runs a version of OpenRefine that is not the current version, they are notified in the UI that there is a more up to date version an encouraged to download it via a link.

The code that displays this information to the user is at
https://github.com/OpenRefine/OpenRefine/blob/3dabe5130b43d6590faaaf37fae541b655f4e713/main/webapp/modules/core/scripts/index.js#L85

The code retrieves information on the latest version from https://api.github.com/repos/openrefine/openrefine/releases/latest and extracts the `data.html_url` property from the response to get a link to the download. However, this contains the URL of the latest tagged release on Github (e.g. https://github.com/OpenRefine/OpenRefine/releases/tag/3.7.4) rather than linking to the official download page. From the Github tagged release page it isn't immediately obvious how to download the latest version, and those users not familiar with Github are presented with an unfamiliar interface.

Suggest that instead the user notification should provide a link to the official download page https://openrefine.org/download so that the user is directed to the official download page with more obvious options and a more user-friendly interface

### LINKED ISSUE #6290: Support showing in-tool notifications about user surveys or other noteworthy things
We currently show a notification on the main page of the tool if a new version is available. It would be useful if we could use a similar mechanism to advertise other things as well, such as when we run a user survey or when we organize an event of some sort.

### Proposed solution

In the JSON file that is queried by OpenRefine to retrieve the latest version, also add another field for an arbitrary notification, like this:
```
{
     "versions": …
     "notification": {
         "text": "Take part in our user survey",
         "link": "https://limesurvey.org…"
     }
}
```
The notification would be displayed in priority over new versions (so that we also reach people who haven't upgraded for a while):

image

### Alternatives considered

The notification could also be displayed elsewhere. Maybe @OpenRefine/designers have suggestions?

### Additional context

I would first migrate to stop querying the GitHub API and use a JSON file hosted on openrefine.org, as suggested by https://github.com/OpenRefine/OpenRefine/issues/6017 and #4463.

wetneb:
This also makes sure we first prompt the user before making a request to an external web service.
Closes #6290. Closes #6017.

### How to test this PR

* To test the workflow of getting notified about a new release, change the `ASSIGNED_VERSION` in `main/src/com/google/refine/RefineServlet.java` from `3.9-SNAPSHOT` to some older version, such as `3.5.0`
* To test the workflow of getting notified about another sort of event, change `https://openrefine.org/versions` to `https://deploy-preview-316--openrefine-website.netlify.app/versions.json` in `main/webapp/modules/core/scripts/index.js`

### Screenshots

screenshot of the prompt for notifications

screenshot of the notification about new version

screenshot of the notification about a survey


tfmorris (changes requested):
I think we need to be more forthcoming about the implications of answering "Yes" to the notifications question. Perhaps include a link to a (to be created) web page which explains the implications (ie tracking) and includes (or links to) the privacy policy for the collected data?

We want them to say Yes, but we want it to be an informed Yes.

wetneb:
I'm not sure what this would look like. Would you like to propose a new design in the issue?

magdmartin:
We can update the existing https://openrefine.org/privacy page on the website with details on what will be stored on OpenRefine when the user subscribe to the notification.

tfmorris:
Yes, as @magdmartin suggests, either create a new section or extend https://openrefine.org/privacy#which-openrefine-operations-transmit-data-to-third-parties to include information about the privacy implications of opting in to polling for updates and add something like "(click [here](https://openrefine.org/privacy#which-openrefine-operations-transmit-data-to-third-parties) for privacy info)" to the question.

wetneb:
Here is another version with a link to the privacy notice:
image

I went for a different wording because "click here" links should generally be avoided (see for instance https://granicus.com/blog/why-click-here-links-are-bad/)

wetneb:
I'd like to merge this soon and release it in 3.8 unless anyone has more concerns.

tfmorris (commented):
I gave it a quick look and didn't see any obvious problems, but my eyes kind of glaze over when presented with that much nested Javascript, so it might be useful to get a front end dev to review, if available.

tfmorris (approved):
The previously requested changes look fine, so approved from that point of view (so Github doesn't keep flagging it red).

PR #6513, opened 2024-04-08
Package Java classes into jars

### LINKED ISSUE #6257: Maven Linux build: please provide jar files instead of single class files
The Linux binary distribution of OpenRefine distributes hundreds of single class files instead of merging them into a few jar files. This can be observed by downloading the latest tar.gz distribution of OpenRefine from Github. This is only a minor issue but would possibly help to improve some reproducible build problems in Debian. 

### Proposed solution

Modify the maven build to install jar files instead of single class files.

wetneb:
Closes #6257.

I tested running it both from the development version (`./refine` in a git clone) and from the packaged version on Linux. Let's see if the CI agrees.

I'd be tempted to include this in 3.8 but because it is likely that this breaks some things, I'd make it another beta instead of releasing 3.8.0 already.

### DIFF
--- a/extensions/database/pom.xml
+++ b/extensions/database/pom.xml
@@ -23,7 +23,6 @@
     </resources>
     <sourceDirectory>src</sourceDirectory>
     <testSourceDirectory>tests/src</testSourceDirectory>
-    <outputDirectory>module/MOD-INF/classes</outputDirectory>
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/extensions/gdata/pom.xml
+++ b/extensions/gdata/pom.xml
@@ -27,7 +27,6 @@
     </resources>
     <sourceDirectory>src</sourceDirectory>
     <testSourceDirectory>tests/src</testSourceDirectory>
-    <outputDirectory>module/MOD-INF/classes</outputDirectory>
     <plugins>
       <plugin>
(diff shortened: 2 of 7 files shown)


tfmorris (changes requested):
`./refine dist 3.8.9` says "must provide a version number", so tested with
```
./refine test
mvn package
```

It looks like the bundled extensions have both a jar and the individual class files, e.g.
```
openrefine-3.8.99/webapp/extensions/database/module/MOD-INF/lib/openrefine-database.jar
openrefine-3.8.99/webapp/extensions/database/module/MOD-INF/classes/com/google/refine/extension/database/DatabaseServiceException.class
```

wetneb:
Good catch, there were some leftovers in the jython extension. The ones for the database extension you mention are likely left over from a previous build: it's worth doing `mvn clean` before.

wetneb:
I thought about it more and actually you probably need to do `mvn clean` twice:
* while on the master branch
* and then while on this PR branch
This is because of the POM changes I made: in my branch, the POM no longer dictates to output classes in a particular directory, so invoking `mvn clean` no longer cleans that directory either.

tfmorris (approved):
LGTM There are 3 class files left for the sample and phonetic extensions, but we can clean that up when we clean up the sample extension (and the phonetic extension is being merged into main).

PR #6514, opened 2024-04-08
Avoid crashing when a cell renderer throws an exception

wetneb:
This will let the next cell renderers attempt to render the cell instead of aborting the rendering of the entire grid.

For https://github.com/OpenRefine/CommonsExtension/issues/99.


tfmorris (approved):

wetneb:
I'm adding this to 3.8 as well.

PR #6528, opened 2024-04-11
Handle Byte Order Mark (BOM) correctly for CSVs. Fixes #6527

### LINKED ISSUE #6527: CSV/TSV separators not guessed correctly for files with Byte Order Marks
It looks like the fix for #1241 is incomplete in that not all places which need to handle the new pseudo encoding which was introduced to handle UTF-8 with Byte Order Marks (BOM) were updated.

### To Reproduce
1. Create a project using this url: http://www.biodiversitylibrary.org/data/TSV/hosted/creator.txt

### Current Results
An exception is logged on the console for an unsupported character encoding and the separator guessing process aborts.

By inspection, the fixed width importer is also susceptible to the same problem.

### Expected Behavior
The separators are guessed correctly for CSV/TSV files

tfmorris:
DRAFT - still needs test updates

Fixes #6527

- Create utility method to handle creating an InputStreamReader with a character encoding which is null, our special BOM encoding, or an actual Java character encoding to centralize handling of this
- Replace all conditional code with calls to the new utility
-

### DIFF
--- a/main/src/com/google/refine/importers/FixedWidthImporter.java
+++ b/main/src/com/google/refine/importers/FixedWidthImporter.java
@@ -27,11 +27,12 @@
 
 package com.google.refine.importers;
 
+import static com.google.refine.importing.ImportingUtilities.getInputStreamReader;
+
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
 import java.io.LineNumberReader;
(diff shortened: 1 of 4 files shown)


tfmorris:
Closing this in favor of #6596. The piece that replaces this is in commit https://github.com/OpenRefine/OpenRefine/pull/6596/commits/076f836b939d47b7fe434212c68a7f35b1c02882

PR #6535, opened 2024-04-14
Closes #5702, wikibase old looking UI on warnings.

### LINKED ISSUE #5702: Improve look and feel of the issues tab of the Wikibase extension
In the Wikibase extension, we offer a tab which displays various issues detected in the edits generated by the current Wikibase schema.

image

I made this UI but I find it pretty ugly:
* the icons for the warnings are too big, sort of pixelated (I think) and their style is inconsistent with OpenRefine's look and feel (we do not have anything similar anywhere else)
* it feels quite tight overall, and not very readable.

How could this be improved? Perhaps there are other changes we could think of to make this part of the UI more inviting?

t8210103:
Closes #5702.

- Changed the icons of the warnings according to the proposed ones.
- Changed the size of the icons from 60px to 30px.
- Increased the gap between the first line of text (header) and the second line of text, just as proposed for visibility reasons.
- Reduced the opacity of the second line of text, just as proposed.
- Moved the 'Locate rows' button to top right next to the first text (header), just as proposed.
- Images indicating the changes both in the case where the button 'Locate rows' appears and the case where the button doesn't appear:

Screenshot 2024-04-14 210118

Screenshot 2024-04-14 210156


wetneb (commented):
Fantastic, thank you so much! That looks really great to me.
I wonder if it would be doable to use vector images (in SVG format) instead of the PNG ones, now that the icons are so much simpler? We are trying to migrate away from our reliance on raster graphics, to make scaling and styling easier.

t8210103:
I pushed a commit that adds the new svg format for the Infromation, Important, Warning and Critical icons.

tfmorris (commented):
Test failure was due to flakiness and is passing after being rerun

PR #6546, opened 2024-04-18
Preserve pagination after actions which trigger a grid update

### LINKED ISSUE #33: Manually selecting a reconciliation match returns the grid to page 1.

### LINKED ISSUE #570: Editing records and advancing forward page by page- how the list refreshes and displays the records

### LINKED ISSUE #572: Undo/Redo page reset problem

wetneb:
This is a backport of #5411 to the master branch. Closes #33, closes #570, closes #572 (already marked as such by #5411).

For this one, although the change was sort of forced on me by the new architecture, it can be backported with reasonable effort to the master branch.

**Summary:** this changes the grid pagination so that users to not input a page number, but the index of the first row in the view. Why do this?
* it solves #33 and others, because starting row number can be meaningfully preserved for all row-wise operations. In comparison, preserving the page number is not guaranteed to keep the view in the same place because changes to earlier rows can shift page boundaries
* it's generally a better practice for pagination, as is explained by various documents (such as https://medium.com/swlh/sql-pagination-you-are-probably-doing-it-wrong-d0f2719cc166)
* it's required for the new architecture, which is what pushed me to working on this.

It's a change that has implications for the UI and the web API (which is not meant to be stable).
It relies on manually annotating all operations with row/record preservation. In this PR it is done in the frontend, but further backend changes will move this information to the backend (making it possible to detect the row/record preservation of sequences of operations run via the `apply-operations` command, for instance).

One oddity of this change is that in records mode, the pagination is still done based on row ids and not on record ids. This is intentional, because:
* it would make it possible to switch between rows and records mode while preserving pagination (although this is not enabled so far)
* so far, the new architecture drops the notion of record ids and uses the starting row index instead. This is primarily for performance reasons, as it enables more efficient record grouping.

### DIFF
--- a/main/src/com/google/refine/commands/row/GetRowsCommand.java
+++ b/main/src/com/google/refine/commands/row/GetRowsCommand.java
@@ -144,18 +144,18 @@ public class GetRowsCommand extends Command {
         /**
          * The value to use as 'end' when fetching the page before this one. Can be null if there is no such page.
          */
-        @JsonProperty("previousPageId")
+        @JsonProperty("previousPageEnd")
         @JsonInclude(Include.NON_NULL)
-        protected final Integer previousPageId;
+        protected final Integer previousPageEnd;
         /**
          * The value to use as 'start' when fetching the page after this one. Can be null if there is no such page.
          */
-        @JsonProperty("nextPageId")
+        @JsonProperty("nextPageStart")
         @JsonInclude(Include.NON_NULL)
-        protected final Integer nextPageId;
+        protected final Integer nextPageStart;
 
         protected JsonResult(Mode mode, List<WrappedRow> rows, int filtered,
-                int totalCount, int totalRows, int start, int end, int limit, Pool pool, Integer previousPageId, Integer nextPageId) {
+                int totalCount, int totalRows, int start, int end, int limit, Pool pool, Integer previousPageEnd, Integer nextPageStart) {
             this.mode = mode;
             this.rows = rows;
             this.filtered = filtered;
@@ -165,8 +165,8 @@ public class GetRowsCommand extends Command {
             this.end = end == -1 ? null : end;
             this.limit = limit;
             this.pool = pool;
-            this.previousPageId = previousPageId;
-            this.nextPageId = nextPageId;
+            this.previousPageEnd = previousPageEnd;
+            this.nextPageStart = nextPageStart;
         }
     }
 
@@ -270,26 +270,26 @@ public class GetRowsCommand extends Command {
             List<WrappedRow> wrappedRows = rwv.results;
 
             // Compute the indices of the previous and next pages
-            Integer previousPageId = null;
-            Integer nextPageId = null;
+            Integer previousPageEnd = null;
+            Integer nextPageStart = null;
             if (start != -1) {
                 if (start > 0) {
-                    previousPageId = start;
+                    previousPageEnd = start;
                 }
                 if (!wrappedRows.isEmpty()) {
-                    nextPageId = wrappedRows.get(wrappedRows.size() - 1).paginationIndex + 1;
+                    nextPageStart = wrappedRows.get(wrappedRows.size() - 1).paginationIndex + 1;
                 }
             } else {
                 if (!wrappedRows.isEmpty() && wrappedRows.get(0).paginationIndex > 0) {
-                    previousPageId = wrappedRows.get(0).paginationIndex;
+                    previousPageEnd = wrappedRows.get(0).paginationIndex;
                 }
-                nextPageId = end;
+                nextPageStart = end;
             }
 
             JsonResult result = new JsonResult(engine.getMode(),
                     rwv.results, rwv.total,
                     engine.getMode() == Mode.RowBased ? project.rows.size() : project.recordModel.getRecordCount(),
-                    rwv.totalRows, start, end, limit, pool, previousPageId, nextPageId);
+                    rwv.totalRows, start, end, limit, pool, previousPageEnd, nextPageStart);
 
             respondJSON(response, result);
         } catch (Exception e) {
--- a/main/src/com/google/refine/commands/row/GetRowsCommand.java
+++ b/main/src/com/google/refine/commands/row/GetRowsCommand.java
@@ -371,7 +371,7 @@ public class GetRowsCommand extends Command {
             return false;
         }
 
-        public boolean internalVisit(Project project, int rowIndex, Row row, int paginationIndex) {
+        protected boolean internalVisit(Project project, int rowIndex, Row row, int paginationIndex) {
             results.add(new WrappedRow(row, rowIndex, null, paginationIndex));
             return false;
         }
--- a/main/src/com/google/refine/commands/row/GetRowsCommand.java
+++ b/main/src/com/google/refine/commands/row/GetRowsCommand.java
@@ -34,8 +34,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package com.google.refine.commands.row;
 
 import java.io.IOException;
-import java.io.PrintWriter;
-import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.List;
 
 import javax.servlet.ServletException;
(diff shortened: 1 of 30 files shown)


tfmorris (commented):
This is awesome! I took a pass through the code and didn't see any significant issues.

I need to think more about the implications of starting row vs page # from a UX point of view. Have you asked testers for feedback on the change?

Unfortunately the first thing that I tried, clear recon choice, reset the grid back to the beginning. I'm not sure if I was just unlucky and chose an overlooked operation or there's something the matter with my build. I was clearing the reconciliation choice in preparation for what I really wanted to test which was reconcile a single item manually.

tfmorris:
p.s. I had rebased your branch onto `master` so that I could test with the latest code base, but perhaps something get messed up, so I'll retest with an untouched version of the branch.

main/src/com/google/refine/commands/row/GetRowsCommand.java
```diff
@@ -257,33 +327,60 @@ public void end(Project project) {
...
+                internalVisit(project, rowIndex, row, sortedRowIndex);
             }
             total++;
+            totalRows++;
 
             return false;
         }
 
         @Override
         public boolean visit(Project project, Record record) {
```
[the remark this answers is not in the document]
wetneb: It's deprecated in the interface already, so I don't think it's necessary to add the `@Deprecated` annotation on the implementation again, is it?

wetneb:
Thanks for the review! Good catch, I simply hadn't enabled the preservation of pagination for the 'Choose new match' button. It should work now.

arthurpsmith:
Hi - I've been trying this, love that it mostly fixes the problem! However if you navigate to a page while reconciling using the "Previous" link (going back in pages), then the first match done on a page removes all reconciliation links and data on the page. Navigation back one more page (Previous) and then forward (Next) restores the reconciliation data, and in general when you navigate via "Next" it always works fine.

wetneb:
Thanks a lot for trying it out and finding this bug! That was an oversight on my part which [should be fixed](https://github.com/OpenRefine/OpenRefine/pull/6546/commits/afb4b37eae14c56236ce7ea4d35fb028db2ee479).

wetneb:
@tfmorris do you intend to review this PR again or is it good to go on your side? I'd be interested in merging it to unblock #6563.

main/src/com/google/refine/commands/row/GetRowsCommand.java
```diff
@@ -63,54 +62,111 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
...
+        protected final Integer end;
         @JsonProperty("limit")
         protected final int limit;
         @JsonProperty("pool")
         protected final Pool pool;
 
+        /**
+         * The value to use as 'end' when fetching the page before this one. Can be null if there is no such page.
+         */
+        @JsonProperty("previousPageId")
```
tfmorris, line 147: Calling this a page ID seems a little unusual. Is the name left over from a previous iteration?

main/src/com/google/refine/commands/row/GetRowsCommand.java
```diff
@@ -63,54 +62,111 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
...
+        @JsonInclude(Include.NON_NULL)
+        protected final Integer previousPageId;
+        /**
+         * The value to use as 'start' when fetching the page after this one. Can be null if there is no such page.
+         */
+        @JsonProperty("nextPageId")
```
tfmorris, line 153: ditto page id

tfmorris (approved):
I did another quick review. Sorry for the delay.

I haven't looked in detail at the internals of the implementation, but I was kind of hoping for more viewport stability for record/row switches (which you mention is explicitly excluded for now), but also for things like facet selection changes. 

I was kind of envisioning a world where the first row on the screen would stay stable (or the closest one to it if it was filtered). Is that a possibility or too hard?

I'd like to see the deprecated and new APIs marked with versions, but other than that it's good to merge.

main/src/com/google/refine/commands/row/GetRowsCommand.java
wetneb: Makes sense, I have renamed it to `previousPageEnd`

main/src/com/google/refine/commands/row/GetRowsCommand.java
wetneb: Same, renamed to `nextPageStart`

main/src/com/google/refine/commands/row/GetRowsCommand.java
```diff
@@ -149,27 +205,19 @@ protected void internalRespond(HttpServletRequest request, HttpServletResponse r
             }
 
             Engine engine = getEngine(request, project);
-            String callback = request.getParameter("callback");
```
tfmorris, line 152: I missed this during my review, but the removal of JSONP support will break existing callers (like the GData extension which doesn't currently work). Instead of being silently ignored, the server should return a clear error if JSONP is not being supported.

PR #6552, opened 2024-04-22
wikibase: Better handling of the 'badtags' error

### LINKED ISSUE #6551: Tag management of the Wikibase editing operation
The Wikibase editing operation currently adds a tag `openrefine-3.9` (in the master branch) to the edits it makes on Wikidata.
Because this tag does not exist on Wikidata yet, editing fails.

In the past, I have taken care of requesting the creation of the tag from Wikidata administrators every time we published a new minor release and I could do it again this time. But it does not feel great, because:
* it's one more manual step that needs to be done after releases, so I'd be keen to eliminate it to keep the release process as simple as possible
* this tag creation is one of the hurdles people encounter when setting up a Wikibase instance to work with OpenRefine, or when we publish a new release and the users of this Wikibase instance upgrade to it.
* The versioning of the Wikibase tags was originally introduced to make it easier to spot problematic edits which would have been introduced by a faulty OpenRefine version, so that they could be fixed. I am not aware of any such instance happening for the past few years, so I don't think this is worth the overhead.

So I would be keen to remove this versioning. One problem is that the tag that is applied on edits is stored in the Wikibase manifests that users have stored in their workspace preferences, in the form of `openrefine-${version}` by default for Wikidata. So it's difficult to change those. I can think of the following approaches:
* Drop support for tagging our edits completely. Editing just works out of the box on any Wikibase instance. Arguably the "(details)" link with the EditGroups batch id is more useful and already identifies the edit as being made via OpenRefine. But if people rely on those tags, they will be annoyed (I remember @lydiapintscher saying that it was important on Wikidata)
* Go for a slightly more careful approach: try adding the tag, but if the tag isn't present on the Wikibase instance, edit without adding the tag. This would make Wikibase upload work out of the box, but will silently create untagged batches, which could be annoying if people do rely on OpenRefine edits being tagged. When they discover that, it's very hard or impossible to add back the tags to the edits.
* Or a slightly more elaborate version: if the tag template contains a version number, try using the version-specific tag. If that fails, try using a version agnostic tag (such as `openrefine`). If that fails, edit without tag. If we request the creation of the `openrefine` tag on Wikidata it would ensure edits remain tagged while still providing out-of-the-box editing support for other Wikibase instances (who might or might not care about tags)

Maybe the last version is good enough?

wetneb:
This falls back on editing with a non-versioned `openrefine` tag instead of `openrefine-${version}` by default, and on editing without a tag if `openrefine` isn't available.

Closes #6551.

### DIFF
--- a/extensions/wikibase/module/scripts/wikibase-manager.js
+++ b/extensions/wikibase/module/scripts/wikibase-manager.js
@@ -51,7 +51,7 @@ WikibaseManager.getSelectedWikibaseMaxlag = function() {
 
 WikibaseManager.getSelectedWikibaseTagTemplate = function() {
   let tag = WikibaseManager.getSelectedWikibase().wikibase.tag;
-  return tag === undefined ? 'openrefine-${version}' : tag;
+  return tag === undefined ? 'openrefine' : tag;
 };
 
 WikibaseManager.getSelectedWikibaseMaxEditsPerMinute = function() {
--- a/extensions/wikibase/src/org/openrefine/wikibase/editing/EditBatchProcessor.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/editing/EditBatchProcessor.java
@@ -27,6 +27,7 @@ package org.openrefine.wikibase.editing;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -65,8 +66,9 @@ public class EditBatchProcessor {
     private NewEntityLibrary library;
     private List<EntityEdit> scheduled;
     private String summary;
-    private List<String> tags;
+    private LinkedList<String> tagCandidates;
 
+    private String currentTag;
     private List<EntityEdit> remainingUpdates;
     private List<EntityEdit> currentBatch;
     private int batchCursor;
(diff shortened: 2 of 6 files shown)


wetneb:
@sebastian-berlin-wmse and @Vesihiisi have tested it and it works, but in the case of media upload this has the bad side effect of uploading the file twice if the first upload fails because of a bad tag. It would be more efficient if we could avoid this double upload. Note that this would only happen for the first upload in a batch, because OpenRefine keeps using the latest tag that worked afterwards.

One approach would be to upload without tags and then apply the tags manually. Perhaps that would be a better architecture, although it might also have downsides:
* the edit isn't tagged immediately, meaning that it might get consumed in this untagged state by eager consumers (recent changes pollers, or Wikimedia event stream listeners…)
* in general, that would add one HTTP request to each edit made

Also Wikidata-Toolkit doesn't have a wrapper for applying tags to an edit so we'd need to make the request ourselves (but that wouldn't be the first time).

At least we should update the manifests so that new users get the right (unversioned) tag.

wetneb:
Actually, the `openrefine-${version}` tag we were using so far doesn't seem to be hard-coded in the manifests, so we should be able to let users upgrade seamlessly to the new tag. This should solve the double upload problem for Commons and Wikidata.

tfmorris (commented):
I don't know enough about Wikibase to offer an opinion.

PR #6555, opened 2024-04-22
wikibase: report editing results in the grid

### LINKED ISSUE #1782: Find a way to report failures of item creations for new editors in Wikidata extension

### LINKED ISSUE #5166: Make error messages more useful
When an error occurs it's not clear which edit gives a problem, I'm trying to correct 73 records and everytime at 40% it gets stuck. In the console I get an error, but no idea which edit exactly gives the problem.

### Proposed solution
Add the statement that is being sent to Wikidata in the error message, so it's easier to debug where the error could be.

### Alternatives considered
Add the Qid of the statement that gives the problem.

### Additional context
```
Exception in thread "Thread-8" java.lang.IllegalArgumentException: Statements describe different subject.
        at org.apache.commons.lang3.Validate.isTrue(Validate.java:158)
        at org.wikidata.wdtk.datamodel.implementation.StatementDocumentUpdateImpl.<init>(StatementDocumentUpdateImpl.java:68)
        at org.wikidata.wdtk.datamodel.implementation.LabeledDocumentUpdateImpl.<init>(LabeledDocumentUpdateImpl.java:64)
        at org.wikidata.wdtk.datamodel.implementation.TermedDocumentUpdateImpl.<init>(TermedDocumentUpdateImpl.java:78)
        at org.wikidata.wdtk.datamodel.implementation.ItemUpdateImpl.<init>(ItemUpdateImpl.java:91)
        at org.wikidata.wdtk.datamodel.implementation.DataObjectFactoryImpl.getItemUpdate(DataObjectFactoryImpl.java:405)
        at org.wikidata.wdtk.datamodel.helpers.Datamodel.makeItemUpdate(Datamodel.java:1219)
        at org.openrefine.wikidata.updates.ItemEdit.toEntityUpdate(ItemEdit.java:165)
        at org.openrefine.wikidata.editing.EditBatchProcessor.performEdit(EditBatchProcessor.java:173)
        at org.openrefine.wikidata.operations.PerformWikibaseEditsOperation$PerformEditsProcess.run(PerformWikibaseEditsOperation.java:262)
        at java.base/java.lang.Thread.run(Unknown Source)
```

wetneb:
Closes #1782. Backports #5944 (see the discussion there).

<s>Depends on #6552 (which is included here, I can rebase once it gets merged), hence the draft status.</s>

This follows the design proposed [on the forum](https://forum.openrefine.org/t/partial-results-of-long-running-operations/482).
The Wikibase upload operation creates a new column (or reuses the column, if it already exists) to store the editing errors of edits related to a given row:
image
It also adds a facet to easily locate rows containing errors:
image

For successful edits it stores a URL to the diff view of the edit (such as https://www.wikidata.org/w/index.php?diff=1920784148&oldid=1919995661).

This also closes #5166, since it makes it possible to relate the errors to the offending lines better.

One could also make the creation of the error-reporting column optional and potentially with a configurable name. It would also be good to make it possible to create the facet manually (just like the judgment / score facets of the reconciliation operation) if it has been closed manually or not restored after reopening the project, but I am not sure where to add the button for that.

### DIFF
--- a/extensions/wikibase/module/langs/translation-en.json
+++ b/extensions/wikibase/module/langs/translation-en.json
@@ -150,6 +150,7 @@
     "perform-wikibase-edits/perform-edits": "Upload edits",
     "perform-wikibase-edits/cancel": "Cancel",
     "perform-wikibase-edits/analyzing-edits": "Analyzing your edits…",
+    "perform-wikibase-edits/wikibase-editing-results": "Wikibase editing results",
     "import-wikibase-schema/dialog-header": "Import Wikibase schema",
     "import-wikibase-schema/file-label": "From JSON file: ",
     "import-wikibase-schema/schema-label": "Or from JSON text:",
--- a/extensions/wikibase/module/scripts/dialogs/perform-edits-dialog.js
+++ b/extensions/wikibase/module/scripts/dialogs/perform-edits-dialog.js
@@ -58,9 +58,10 @@ PerformEditsDialog.launch = function(logged_in_username, max_severity) {
         maxlag: elmts.maxlag.val(),
         editGroupsUrlSchema: WikibaseManager.getSelectedWikibaseEditGroupsURLSchema(),
         tag: WikibaseManager.getSelectedWikibaseTagTemplate(),
-        maxEditsPerMinute: WikibaseManager.getSelectedWikibaseMaxEditsPerMinute()
+        maxEditsPerMinute: WikibaseManager.getSelectedWikibaseMaxEditsPerMinute(),
+        resultsColumnName: $.i18n('perform-wikibase-edits/wikibase-editing-results')
       },
-      { includeEngine: true, cellsChanged: true, columnStatsChanged: true },
+      { includeEngine: true, modelsChanged: true, cellsChanged: true, columnStatsChanged: true },
       { onDone: function() { dismiss(); } }
     );
   };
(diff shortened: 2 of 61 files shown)


wetneb:
Merging this soon.

wetneb:
Given @lokal-profil's interest in error reporting (in https://github.com/OpenRefine/CommonsExtension/issues/103#issuecomment-2178657843) I had a closer look at the UX for media files with this PR. So far, when creating new entities (such as when uploading new files), the "Wikibase editing results" column was left empty, because the link to the created entity was already included in the reconciled cell. But leaving such cells blank stroked me as counter-intuitive: if users are trained to recognize a successful edit by a cell containing a link to Wikibase instance in this column, I think it makes sense to also include a link to the created entity in this case. So I have made this change accordingly.

tfmorris (commented):
My apologies! I apparently started reviewing this and never clicked "Submit."

I think the changes to return values are probably backward compatible, but the changes to public constructors should probably be evaluated for their compatibility impact.

I'll see if there's a way to find any other partially completed Github reviews that I have outstanding.

wetneb:
Thanks for the review. I think it's generally useful to pay attention to changes to public APIs, but in this case they are all within the Wikibase extension.

As far as I am concerned, I don't have any stability expectations for this module, as the extension is not designed to be extensible itself, nor to be used as a library.

The `public` nature of those APIs is because classes get reused in different packages within the extension, not because they are meant to be reused in other code bases. For instance, in this case, `org.openrefine.wikibase.editing.EditBatchProcessor` needs to be public because it is used by `org.openrefine.wikibase.operationsPerformWikibaseEditsOperation`. Preventing its reuse outside of the Wikibase extension would likely require relying on the [JPMS](https://www.oracle.com/corporate/features/understanding-java-9-modules.html) (but it would only be worth looking into that if the extension was designed to be reused at all).

 Do you have different expectations about the stability of the API exposed by this module? Or are you aware of any reuse of this module somewhere? The Commons extension does not depend on it. I wonder if others could have similar expectations as yours about the stability of this module and how to avoid such misunderstandings.

PR #6563, opened 2024-04-24
Resizeable columns

### LINKED ISSUE #4806: Possibility to adjust column width manually

wetneb:
Closes #4806. Backports #5840.

<s>Depends on #6546, therefore marked as draft until this one is merged.</s>

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
@@ -59,17 +59,6 @@ DataTableColumnHeaderUI.prototype.getColumn = function() {
   return this._column;
 };
 
-// global state for the resizing of column headers
-DataTableColumnHeaderUI.resizingState = {
-  dragging: false, // whether we are currently resizing any column
-  col: null, // the column being resized
-  columnName: null, // the name of the column being resized
-  originalWidth: 0, // the original width of the header when the dragging started
-  originalPosition: 0, // the original position of the cursor when the dragging started
-  moveListener: null, // the event listener for mouse move events
-  releaseListener: null, // the event listener for mouse release events
-};
-
 DataTableColumnHeaderUI.prototype._render = function() {
   var self = this;
   var td = $(this._td);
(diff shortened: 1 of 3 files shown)


tfmorris (commented):
This will be a very nice addition! I did some functional testing and things look good. I did glance at the Javascript, but it would be better if @elebitzero and/or @Abbe98 can spare the time to review it.

One weird thing I noticed is that the vertical border between some header cells was missing:
Screenshot 2024-07-29 at 5 43 23 PM

wetneb:
Good catch! This was caused by some fractional widths being set, which should be fixed by [d13a68a](https://github.com/OpenRefine/OpenRefine/pull/6563/commits/d13a68a0a57525a9979b25fbeaefcee8ebd83a52).

tfmorris:
Thanks for the update, but, unfortunately, it doesn't seem to have fixed the problem. The initial size of many columns is fractional, but they display fine, so I poked a little bit at the CSS. When I increased the right border from 1px to 2px, the display looks like this:
Screenshot 2024-08-14 at 3 29 18 PM
so the border on the "Profession" column is there, but partially occluded by something. The only other hint that I can offer is that removing `width: max-content` appears to fix the border rendering (but also collapses all the column widths to their default).

Screenshot 2024-08-14 at 3 32 24 PM
Screenshot 2024-08-14 at 3 32 36 PM

wetneb:
Sorry about this, I am not sure why I thought my commit had fixed it: I could indeed still reproduce the problem on my side.

After more investigation it seems that the `will-change: transform` CSS attribute was the culprit. I am not completely sure why it was introduced back when @lisa761 worked on infinite scrolling. I suspect it was to anticipate that the width of a column could change when scrolling, as new content would be added in the newly-fetched rows. This isn't the case here as we don't have any infinite scrolling.
Also this property is meant to be added in a transitional fashion via a script rather than permanently, so I think it's safe to remove. And it seems to reliably fix the issue as far as I can tell (tested with Firefox and Chromium).

tfmorris (changes requested):
That seems to have done the trick! Borders are now rendered reliably.

Apparently I was testing on a project without column groups before and I noticed when testing with column groups, that the resize affordance is only shown on the bottommost row of the column header. It would be nicer if the full vertical height of the header, including all rows, were selectable. Because column groups are relatively rare, this is low priority, but it would be nice to include it if it's not too complicated.

Screenshot 2024-09-18 at 3 43 56 PM

wetneb:
I have expanded the resizing controls so that they also cover the column groups area

tfmorris (approved):
Looks great!

PR #6567, opened 2024-04-25
refine script: Abort build early if NPM is missing

### LINKED ISSUE #6566: ./refine script should fail properly if npm is not available
When running `./refine` on Linux, if OpenRefine hasn't been built yet the script will attempt to do so.
If `npm` isn't available, downloading frontend dependencies will be skipped without failing the build, resulting in a dysfunctional frontend.
We should instead abort the build and clearly report the problem to the user.

wetneb:
Closes #6566.


tfmorris (approved):
RSLGTM

PR #6571, opened 2024-04-28
grel: Prepare tests for modularization

wetneb:
- introduce a new base class for GREL tests, which extends our current `RefineTest`. This makes it possible to remove the test utilities that depend on GREL from the RefineTest class, in anticipation for it being in a core Maven module where GREL isn't available
- also splits out some tests that were in `EvalErrorTests` but really belong to the test classes for the functions they actually test. This is also necessary because `EvalError` will live in the core Maven module (being part of the core data model, as they can be stored in Cell), whereas the corresponding functions will be in the GREL module

### DIFF
--- a/main/tests/server/src/com/google/refine/expr/functions/CoalesceTests.java
+++ b/main/tests/server/src/com/google/refine/expr/functions/CoalesceTests.java
@@ -33,37 +33,15 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package com.google.refine.expr.functions;
 
-import java.util.Properties;
-
-import org.slf4j.LoggerFactory;
 import org.testng.Assert;
-import org.testng.annotations.AfterMethod;
-import org.testng.annotations.BeforeMethod;
-import org.testng.annotations.BeforeTest;
 import org.testng.annotations.Test;
 
-import com.google.refine.RefineTest;
 import com.google.refine.expr.EvalError;
+import com.google.refine.grel.GrelTestBase;
 
-public class CoalesceTests extends RefineTest {
+public class CoalesceTests extends GrelTestBase {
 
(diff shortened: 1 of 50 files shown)


main/tests/server/src/com/google/refine/expr/functions/CoalesceTests.java
```diff
@@ -33,38 +33,16 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
...
-    @BeforeMethod
-    public void setUp() {
-        bindings = new Properties();
-    }
-
-    @AfterMethod
-    public void tearDown() {
-        bindings = null;
-    }
-
```
wetneb, line 67: I have also used this opportunity to clean up unnecessary set up functions which are already included in the base class.

main/tests/server/src/com/google/refine/expr/functions/strings/DetectLanguageTest.java
```diff
@@ -1,25 +1,20 @@
...
-
-    String functionName;
+public class DetectLanguageTest extends GrelTestBase {
 
-    @Override
-    @BeforeTest
-    public void init() {
-        functionName = "detectLanguage";
-    }
+    String functionName = "detectLanguage";
```
tfmorris, line 17: This seems a little odd, but I guess it comes from the previous code.

tfmorris (approved):
LGTM. Sorry it took me so long to review. The change summary in the PR was very helpful in wading through the volume of changes to focus on the critical parts, so thanks for that.

wetneb:
Merging this broke the build because of a missing import, so I am adding it directly to master with [6a86001](https://github.com/OpenRefine/OpenRefine/commit/6a86001ba7e9589f6f0374660a3e65980fd8f6ce)

PR #6572, opened 2024-04-28
tests: remove spurious dependencies to ImporterTest

wetneb:
Another small test change that's in the way of Maven modularization.


wetneb:
No, that's a cherry-picking error on my side.

PR #6574, opened 2024-04-28
Remove grel registration from RefineTest

wetneb:
This removes the final dependency from RefineTest to GREL by not registering GREL in it by default.
Instead, we register and unregister it in each test that relies on it.

~Draft because it depends on #6502 and #6571. Once those are merged, the diffs will be smaller.~


tfmorris (approved):
Based on the number of identical changes, it feels like there's probably an opportunity to DRY this up, but that's more an artifact of our current test structure than this PR, so we can refactor things separately.

PR #6591, opened 2024-05-07
wikibase: Alert the user when they do not provide an edit summary

### LINKED ISSUE #6306: Make it clear that summary is mandatory for Wikibase upload
When I was close to finishing my Wikibase edit and needed to add a summary in the upload menu, it wasn't clear to me that the **summary** was mandatory for the upload to work.

Let's add a visual cue - for example, an asterisk - to indicate the field is required.
Upload edits

t8210103:
Closes #6306 

- Unfortunately, i couldn't configure the request for a working "required" tag inside the HTML code. 
- Directly implementing the "required" tag in the appropriate HTML "input" tag did not work, since the form isn't directly submitted from the HTML code, but from a JavaScript code that doesn't recognize the "required" tag.
- Although, as i noticed in different parts off the project, problems as such are dealt with the pop-up window that follows: 
Screenshot 2024-05-07 202645
-(The previous message pops-up when the user clicks "next" without choosing a file in the "create project" section).
-What this PR changes, is instead of focusing to the box next to "summary", it pops-up the following message:
Screenshot 2024-05-07 202618
-(This validation message was already implemented inside the code but never reached).

### DIFF
--- a/extensions/wikibase/module/scripts/dialogs/perform-edits-dialog.js
+++ b/extensions/wikibase/module/scripts/dialogs/perform-edits-dialog.js
@@ -33,11 +33,6 @@ PerformEditsDialog.launch = function(logged_in_username, max_severity) {
     var formCopy = hiddenIframe.find("#wikibase-perform-edits-form");
     formCopy.trigger('submit');
 
-    if (elmts.editSummary.val().length === 0) {
-      elmts.editSummary.trigger('focus');
-      return;
-    }
-
     if (elmts.maxlag.val().length === 0) {
       elmts.maxlag.val(WikibaseManager.getSelectedWikibaseMaxlag());
     }


wetneb (commented):
Thank you for having a look!

Reusing the validation logic from the backend is not a bad idea, but I am reluctant to have it include a Java exception class name because it's an internal implementation detail that isn't very readable and therefore shouldn't be presented to users.

Also, opening a modal dialog strikes me as rather "obstructive" since it requires the user to first close the modal, then manually focus the summary field, before they can start describing their changes. Don't you like pointing the user to the field directly? It feels like a widespread practice to me.

t8210103:
Absolutely, I understand your concerns. I will close this pull request for now, and if I manage to address these points, I'll open a new one.

PR #6592, opened 2024-05-08
feat: openrefine icon compliant with new MacOS guidelines

### LINKED ISSUE #6399: Add an icon in line with the new visual guidelines of MacOS
- The icon of OpenRefine is out of place with the new graphic charter for m

### Proposed solution
- Add an icon in line with the new visual guidelines of MacOS

image

### Available solution
- https://macosicons.com/#/open%20refine

teolemon:
Fixes #6399 

### What
- feat: openrefine icon compliant with new MacOS guidelines


tfmorris:
Thanks! 

Pinging @OpenRefine/designers to see if they have any feedback.

wetneb:
When pinging designers it's generally helpful to provide a screenshot for them to be able to judge the changes easily - especially since GitHub doesn't provide previews for `.icns` files. Here is one:
Screenshot 2024-05-21 at 10 06 51
(the logo on the left-hand side is the new one, the old one is on the right-hand side)

thadguidry:
still think we can go _just_ a bit larger... to the maximum that's allowed for the icon grid? as I mentioned in this comment https://github.com/OpenRefine/OpenRefine/issues/6399#issuecomment-2103960061

wetneb (approved):
@thadguidry I don't think we can go much larger because we are limited on the width by the white frame.

I'd merge this soon unless @teolemon wants to do further tweaks.

tfmorris (approved):

wetneb:
> I have yet to see the designers reply to a ping.

Yes we should delete this GitHub team, especially because the recommended way is to add issues/PRs to a GitHub project instead:
https://forum.openrefine.org/t/design-workflows-in-github/1396
I have to say that I have doubts about how effective this workflow is (it doesn't seem to bring a lot of designer attention to issues either), but if that's the one that designers are asking for, well, it's worth trying…

> @wetneb have you been successful in getting replies by including screenshots?

It looks like in this case it helped @thadguidry chime in, no? (Especially knowing that Thad generally uses Windows)
I mean, it does save people the trouble from pulling a branch via git and compiling (and in this case even packaging), which is time consuming even for people who are totally comfortable with the process, so it's hard for me to imagine how that would not help…

teolemon:
No tweaks from me, but I'm not a designer by training 🫠

Issue #5792, opened 2023-04-18
Read files to import directly from .zip archive without extracting them first


wetneb:
When importing a file supplied as a `.zip` archive (which can contain multiple tables to import), OpenRefine uploads the archive to the import directory, extracts the archive to disk, and then goes on to read those files.

For large archives, the time spent unzipping the archive is actually noticeable in the import process. This step has been identified earlier as being wasteful (by @tfmorris):
https://github.com/OpenRefine/OpenRefine/blob/c4f4a11aa9cad3797d19a2dd1b06d89fc690a827/main/src/com/google/refine/importing/ImportingUtilities.java#L699-L705

### Proposed solution

We could skip that step and just let the importers run off the zip file itself. It is fairly straightforward to obtain an `InputStream` from an entry inside a zip file. This would have the following benefits:
* skipping a step in the import process
* using less temporary disk storage during the import process

It is likely that reading from inside a zip file is not necessarily much slower than reading from the uncompressed counterpart (since the decompression overhead can be compensated by the lower disk IO load). Generally speaking it is standard practice to decompress on the fly while processing a file.

Some archive types might not support decompression on the fly: for those formats, we would keep the existing system.

### Additional context

I think this is not so relevant for 3.x (where large projects would not really be workable downstream anyway), but for 4.x it would make sense because it could make a noticeable difference in the import user experience.

WR-Smiley:
I'd like to be assigned to take a crack at this one. Any other specific archive types that we want to be read while still compressed?

wetneb:
Nice! I'd be curious if this change can be implemented without changing the interface that importers need to implement. If such a change is needed then it would be nice to make it backwards-compatible, if doable.

I think it could make sense to support this for any archives where it's possible to do so, unless it poses some architectural concerns.

WR-Smiley:
I'll do some research on maintaining the interface and let you know if I find a change is likely needed. I'll try to implement general compatibility for all archives where possible. Thanks!

thadguidry:
Will the proposed implementation give the user a good sense of the memory being used and growing larger as the importer processes the files?  Hopefully our existing percentage progress bar will be fairly accurate, or do it need some tweak in it's code also for this issue?

WR-Smiley:
[...]

PR #6593, opened 2024-05-08
fix: Opening all files inside .tar.gz

DaxServer:
Simplified the opening of tar archives by removing unnecessary check for the first entry and directly returning the `TarArchiveInputStream` in the `tryOpenAsArchive` method. Also updated the usage of deprecated `getNextTarEntry` to `getNextEntry` in ImportingUtilities.java.

Supports #5792


wetneb (approved):
Sorry about the long delay in reviewing this!
On the surface I was worried that this would break the ability to read Gzipped files which are not TAR archives but I checked that it isn't the case.

PR #6596, opened 2024-05-10
Fix encoding guesser for non-UTF-8 BOM-based files. Fixes #6595

### LINKED ISSUE #6595: Encoding issue regression for files imported into version 3.8.0
In OpenRefine 3.8.0, extra NUL chars are being introduced between letters in a dataset, where it is not happening with previous version 3.7.9

### To Reproduce
1. First, download dataset zip from below
2. Open using OpenRefine 3.8.0
3. Notice NUL chars between letters
4. Now open using OpenRefine 3.7.9
5. Notice no NUL chars between letters

### Current Results
NUL chars are being introduced erroneously between letters.  Appears to be some kind of errant decoding/encoding issue?

### Expected Behavior
Should work as it did before with OpenRefine 3.7.9 , without the extra NUL chars being introduced.

### Screenshots

With OpenRefine 3.8.0
image
With OpenRefine 3.7.9
image

### Versions
 - Operating System: Mac OS and Windows 11
 - Browser Version: Edge and Safari
 - OpenRefine: 3.8.0

### Datasets

Download the ZIP file from https://www.cssf.lu/en/Document/identifiers-of-aifms/
direct link: https://www.cssf.lu/wp-content/uploads/IDENTIFIANTS_AIFM.zip

### Additional context
Opening the ZIP file directly in OpenRefine or extracting it first and then opening in OpenRefine, causes the same encoding issue and appearance of NUL char between letters.

tfmorris:
Fixes #6595

- refactor InputStreamReader handling to DRY up handling of null encoding and our fake UTF-8 + BOM encoding
- fix user override of file encoding
- fix encoding guessing by not skipping BOM for non-UTF-8 cases so that it is still available for character encoding guesser
- Add tests for UTF-16LE and UTF-16BE with BOMs
- Refactor encoding guesser so that it can be used for a single file
- Enhanced format guessing tests to not assume UTF-8 and instead guess encoding using refactored encoding guesser


wetneb (approved):

wetneb:
This looks like something worth backporting to 3.8, perhaps?

tfmorris:
Yes, I think so. Apparently the UTF-16LE encoding is not uncommon in the Windows ecosystem, even though most of the rest of the world principally uses UTF-8.

The core fix for #6595 is simply adding the second parameter to `new UnicodeBOMInputStream(is, true)`, so I could generate a more minimal fix, but I think all 3 fixes are useful and a bunch of the changed code in the PR are things like additional tests, added TODOs, etc.

PR #6600, opened 2024-05-13
recon: Add back missing 'search for match' button

### LINKED ISSUE #6599: 'Search for match' link not shown when no reconciliation candidates are present
Quoting [E11 on the forum](https://forum.openrefine.org/t/search-for-match-missing-during-reconciliation-3-8-0/1513):

> I'm attempting to reconcile values to Wikidata using verstion 3.8.0, against no particular type, no auto-matching. When the service offers one or more potential matches, I do see the "Search for match" link in that cell. But when the service offers no potential matches, I do not see the "Search for match" link. Is this a bug, or am I missing something?

I can indeed reproduce this.

### Screenshots
image

### Versions
 - OpenRefine: 3.8.0, master branch too

wetneb:
Closes #6599.

A lot of the recon rendering code was guarded by an `if` statement which checked that there was at least one candidate, but this was not necessary as the buttons should be available even if there isn't any candidate. The 'Search for match' button is in fact particularly useful when the recon service does not return any candidate.

I would release this in 3.8.1.


tfmorris (approved):
RSLGTM based on e2e end test passing

PR #6606, opened 2024-05-15
Fix NullPointerException in forEachIndex

### LINKED ISSUE #6598: forEachIndex with array containing null values throw NullPointerException
## Summary
`forEachIndex(["a","b",null], i, x, x)` should return the original array rather than throwing an error as it currently does.

**>>> OK**

image

image

image

image

.
.
.
**>>> KO**

image

image

image

### Expected Behavior
In my opinion the code shoud always output the original array

### Versions
 - OpenRefine: 3.7.9 with builtin java

tfmorris:
Fixes #6598 

- Fix NPE when `forEachIndex` is used with an array containing `null`s
- Make sure tests don't use `join()` which has broken handling for `null`s

Note: this is based off branch for PR #6605 so that should be review/merged first

### DIFF
--- a/main/src/com/google/refine/expr/functions/arrays/Join.java
+++ b/main/src/com/google/refine/expr/functions/arrays/Join.java
@@ -59,13 +59,16 @@ public class Join implements Function {
 
                 if (v.getClass().isArray() || v instanceof List<?> || v instanceof ArrayNode) {
                     StringBuffer sb = new StringBuffer();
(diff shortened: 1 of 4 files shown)


wetneb (approved):

wetneb:
Sorry, of course I merged it in the wrong order… But both of your fixes are in.

tfmorris:
Thanks for the quick review!

PR #6607, opened 2024-05-15
Changed aria-role=dialog to role=dialog

### LINKED ISSUE #6602: The dialog system uses an incorrect WAI ARIA Role attribute
The WAI-ARIA element attribute `aria-role` is deprecated and incorporated into browsers via only `role` [1](https://stackoverflow.com/questions/10403138/what-is-the-purpose-of-the-role-attribute-in-html) [2](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/WAI-ARIA_basics)

We should replace `aria-role=` with only `role=`

`<div class="dialog-frame" style="min-width: 600px; visibility: visible;" aria-role="dialog" tabindex="-1">`
to
`<div class="dialog-frame" style="min-width: 600px; visibility: visible;" role="dialog" tabindex="-1">`

### To Reproduce
1. Start OpenRefine
2. Open console (or browser debugging tools)
3. Notice error of `aria-role` 

### Current Results
Error in browser consoles while debugging various parts of OpenRefine's frontend.

### Expected Behavior
no error about "aria-role" being unrecognized.

### Versions
 - JRE or JDK Version: JDK 21
 - OpenRefine: master

t8210103:
Fixes #6602 

- Changed **aria-role=dialog** to **role=dialog** just as suggested in the issue.
- Expected behavior is satisfied.

### DIFF
--- a/main/webapp/modules/core/scripts/util/dialog.js
+++ b/main/webapp/modules/core/scripts/util/dialog.js
@@ -76,7 +76,7 @@ DialogSystem.showDialog = function(elmt, onCancel) {
 
   DialogSystem.setupEscapeKeyHandling();
 
-  elmt.attr("aria-role", "dialog");
+  elmt.attr("role", "dialog");
   var dialogHeader = elmt.find(".dialog-header");
   if (dialogHeader.length && dialogHeader[0].id) {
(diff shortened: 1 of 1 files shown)


wetneb:
Isn't this a duplicate of #6603?

Issue #4301, opened 2021-11-15
Cluster values by distance or bins defined by an expression


wetneb:
We offer two distances to cluster values together: Levenshtein and PPM. Sometimes, with knowledge of the data I could come up with a better distance, and express it as a GREL function (or another expression language). OpenRefine currently does not allow such user-defined functions.

### Proposed solution

Offer the possibility to use a custom function that takes two arguments (say `a` and `b`) and returns their distance. Then, do clustering based on that distance. For instance, I could use the following function: `1 - len(LCP(s,t)) / avg(len(s), len(t))` (taken from the mailing list, see below).

### Alternatives considered

Currently, users would need to develop their own OpenRefine extension to define such a distance.

### Additional context

Requested on the mailing list: https://groups.google.com/g/openrefine/c/ABxaEFQS8mU/m/DKGkYaFKAgAJ

--------------------------
_This is a [proposed Outreachy project in 2022](https://github.com/OpenRefine/OpenRefine/wiki/GSoC-Outreachy-2022%C2%A0Ideas). If you are **not** planning to apply for an internship via Outreachy/GSoC, we kindly ask that you do not work on this task yet, in order to leave the floor to potential interns._

wetneb:
This could also be considered for our other binning strategy: binning. In this case, the expression would take a single argument (the value being clustered) and return its corresponding bin. Perhaps that is actually easier to start with.

thadguidry:
Note: Our Key Collision method to users is also known as our Binning method.

wetneb:
I have proposed this as a potential Outreachy project (and maybe GSoC if we participate to that as well). This means I think it should be a good topic for a 3 months internship. This is therefore not suitable as a first issue: have a look at [our list of good first issues](https://github.com/OpenRefine/OpenRefine/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) instead.

gitonthescene:
Since clustering is really a [special case](https://github.com/OpenRefine/OpenRefine/issues/2003#issuecomment-792545880) of reconciliation, tools which do this sort of reconciliation can be repurposed to do clustering.

ostephens:
@gitonthescene - can you clarify what you mean when you say

> Since clustering is really a https://github.com/OpenRefine/OpenRefine/issues/2003#issuecomment-792545880 of reconciliation

[...]

PR #6612, opened 2024-05-18
clustering: User defined distances and keyers

zyadtaha:
The first pr in the GSoC project #4301 

Changes in this pull request:
- Added `UserDefinedKeyer` class + unit tests for it
-  Added `UserDefinedDistance` class + unit tests for it

### DIFF
--- a/main/src/com/google/refine/clustering/binning/UserDefinedKeyer.java
+++ b/main/src/com/google/refine/clustering/binning/UserDefinedKeyer.java
@@ -1,0 +1,68 @@
+/*
+
+Copyright 2010, Google Inc.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+    * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+    * Neither the name of Google Inc. nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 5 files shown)


wetneb (commented):
That looks really great!
The next step would be to register the keyer and distance so that they can be used in the compute-clusters HTTP command. This registration is done in the `controller.js` file (`main/webapp/modules/core/MOD-INF/controller.js`).
One problem that you will encounter is how to pass the clustering expression to the keyer. You can have a look at how it's currently done for the ngram clusterer for instance: it's quite ad-hoc and not extensible. It's worth thinking about ways to improve that, or keep using the same strategy if we think it's still the best approach.

tfmorris (commented):
Was it intentional that this branch was created in the organization's repository? I think it would be preferable to do the development on a private fork of the repo and submit the PR from there, but I don't know if @wetneb told you to do it this way for some reason.

It would also be useful to reference the issue that this addresses so that the PR and the issue are linked together.

zyadtaha:
@tfmorris no, it wasn't intentional. Sorry for that mistake. I will work on my private fork in the future PRs.

zyadtaha:
Thank you so much. You've saved me time.

zyadtaha:
I thought about the idea of passing the clustering expression to the keyer like how it's currently done for the n-gram clusterer but this will need adapting to:
- `BinningClusterer` and `kNNClusterer` (add the expression to the parameter field of the classes and when initializing the visitor use it)
- `UserDefinedKeyer` and `UserDefinedDistance` (not passing expression through constructor but the keying/distance function)

So I have decided to only check if there is an expression in the request. If yes, then pass it to the constructors of the Keyer or Distance. The changes are only made in `ComputeClustersCommand`

I also tested passing expressions in the request manually and it works fine. Should we have a unit test for it ?

wetneb:
I think the approach you used to configure the keyer / distance is fine. Our existing architecture does not make it possible to register configurable keyers or distances, so some workaround like you used is necessary.

It would be good to clean this up in the future, but I would say this can be part of a separate refactoring, especially since it will impact other clustering methods (such as ngrams).

wetneb:
I'm changing the title of this pull request to make its topic clearer (there is nothing in it that indicates that it's about clustering so far).

wetneb (approved):
Let's merge this already and iterate in follow-up PRs.

PR #6618, opened 2024-05-20
Revised allocateID function.

### LINKED ISSUE #6471: Generate deterministic project IDs correlated with creation date
Currently project IDs are generated using the formula:

```
        return System.currentTimeMillis() + Math.round(Math.random() * 1000000000000L);
```

This suffers from the problem that there's no way to recover the original Unix timestamp to know when the project was created if the ID is the only information available. In data recovery and debugging scenarios, it can be useful to know when a project was created.

### Proposed solution
Multiply the timestamp by a fixed amount and then add the random component in the lower order bits, e.g.

```
        return (System.currentTimeMillis() * 1000) + Math.round(Math.random() * 1000);
```

### Alternatives considered
Powers of 2 are easier for shifting/masking bits, but harder for humans to process
```
        return (System.currentTimeMillis() << 10) + Math.round(Math.random() * 1024);
```

### Additional context
Throughout the system the identifier is treated as opaque, so changing how it's generated shouldn't affect anything.

The 64-bit value returned by `System.currentTimeMillis()` has the high order 23 bits clear, so there's plenty of headroom to shift it up without affecting longevity of the IDs.

Sophia3287:
Fixes #6471

- This pull request updates the allocateID() method to use higher-order random numbers. 
- By shifting the current timestamp 20 bits left, it adds a random number up to 1 million (2^20) to the lower bits. 
-


wetneb:
Superseded by #6621

PR #6626, opened 2024-05-23
Checking running status of OpenRefine with wget will not work correct…

### LINKED ISSUE #6622: Checking running status of OpenRefine with wget will not work correctly
The PR #6361 which was meant to fix #2000 changes the `check_running()` code in the `refine` script. It introduces a variable `URL_LOCAL="http://127.0.0.1"` alongside the existing variable `URL="http://${REFINE_HOST_INTERNAL}:${REFINE_PORT}/".`

- if cURL is used to check for a running OpenRefine instance, it checks using the URL variable
- if wget is used to check for a running OpenRefine instance, it checks using the URL_LOCAL variable

This means wget will never fetch the OpenRefine page because it doesn't use the specified host or port that OpenRefine is/should be expected to be running on.

I believe that the lines for wget 
https://github.com/OpenRefine/OpenRefine/blob/24947a3df0ac806024a726bf4843f1614fbfab22/refine#L106
https://github.com/OpenRefine/OpenRefine/blob/24947a3df0ac806024a726bf4843f1614fbfab22/refine#L116

should check the `URL` not `URL_LOCAL` and `URL_LOCAL` variable can be removed as it is redundant

dino2580:
…ly #6622

Fixes #6622

- Removed Use of Local Url
- Added URL(global) in place

### DIFF
--- a/refine
+++ b/refine
@@ -94,7 +94,6 @@ check_running() {
     check_downloaders
     URL="http://${REFINE_HOST_INTERNAL}:${REFINE_PORT}/"
     CHECK_STR="<title>OpenRefine</title>"
-    URL_LOCAL="http://127.0.0.1"
 
     if [ "$CURL" ] ; then
         curl --noproxy 127.0.0.1 -s -S -f $URL > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thank you!

wetneb:
I would add this to 3.8.2

PR #6631, opened 2024-05-26
Update pom.xml Adding David as Developer Emeritus

thadguidry:
- several other open source projects assign the role Developer Emeritus, which we should do properly for David to avoid confusion.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,11 +47,11 @@
 
   <developers>
     <developer>
-      <id>dfhuynh</id>
-      <name>David F. Huynh</name>
-      <url>https://github.com/dfhuynh</url>
+      <id>contributors</id>
+      <name>OpenRefine Contributors</name>
+      <url>https://github.com/OpenRefine/OpenRefine/graphs/contributors</url>
       <roles>
-        <role>Developer Emeritus</role>
+        <role>Contributor</role>
       </roles>  
     </developer>
   </developers>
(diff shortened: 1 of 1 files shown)


magdmartin:
Would it be something we track via the README or the GOVERNANCE.md document instead? I am worried that we will maintain a list of contributors in yet another file that will quickly become out of date.

thadguidry:
POM list of contributors is something that other projects do.
But I am not asking for that here.
We had decided to only add David into the POM as original source developer emeritus.
But we didn't add that role to make it clear.
That's what this PR does.

As to where other lists of contributors go... It should NOT be the POM.
Over the years I have changed my stance and I am comfortable with provenance coming from mailing lists, forum, and Git history, which means we don't have to create any contributor listings anywhere.

tfmorris:
David definitely deserves all the recognition that we can give him as the founder and primary developer, but it seems odd to have him listed in the POM as the only developer, even with the emeritus tag. The POM documentation [says](https://maven.apache.org/pom.html#Developers):

> A good rule of thumb is, if the person should not be contacted about the project, they do not need to be listed here.

For the purposes of the POM, I think I'd either go with just referencing the organization, which can be used to find all the developers, or expanding to some set of core developers and listing David as emeritus. I think I'd lean towards the former (and hopefully there are other places where we give him his due).

thadguidry:
@tfmorris There are other places, in our website docs.

Ok, I think I will change this to "OpenRefine Contributors" as you suggest.

tfmorris (approved):

PR #6628, opened 2024-05-24
Add new GREL function to calculate the edit distance #6613

dino2580:
Fixes #[6613](https://github.com/OpenRefine/OpenRefine/issues/6613)

Changes proposed in this pull request:
- Used Apache LevenshteinDistance.java
- Created editDistance grel

PR #6634, opened 2024-05-27
Add testcases for Pull request#6628

xinluz6:
Fixes #6628 

- Added test cases for the EditDistance function after modifying the code based on the improvements suggested in pull request #6628
- 
-


wetneb:
Could you perhaps make this PR against the branch `patch-3` of @dino2580's fork?
This way @dino2580 could accept your PR on their side, which would then update their PR in OpenRefine's official repository.
If we run the CI on your existing PR here, it will fail because the tests will not pass without the implementation being there.

xinluz6:
Hi @wetneb，I have made the PR against the branch patch-3 of @dino2580's fork. Thank you for the suggestion and guidance! I apologize for any inconvenience caused.

wetneb:
Thank you! Then I believe we can close this one, as it will be merged via #6628.

PR #6641, opened 2024-05-27
Add test case for save method.

Sophia3287:
Fixes #{issue number here}

- Add test case for save method.
- 
-

### DIFF
--- a/main/tests/server/src/com/google/refine/io/ProjectMetadataUtilitiesTest.java
+++ b/main/tests/server/src/com/google/refine/io/ProjectMetadataUtilitiesTest.java
@@ -60,4 +60,33 @@ public class ProjectMetadataUtilitiesTest {
             file.delete();
         }
     }
+    /**
+     * Check if the save method can correctly save the metadata to the target directory.
+     **/
+    @Test
+    public void testSave() throws Exception {
+        File projectDir = new File(System.getProperty("java.io.tmpdir"), "projectDir");
+        projectDir.mkdir();
+
+        try {
+            ProjectMetadataUtilities.save(actualMetadata, projectDir);
+            File metadataFile = new File(projectDir, "metadata.json");
+
+            expectedMetadata = ProjectMetadataUtilities.loadFromFile(metadataFile);
+            Assert.assertEquals(expectedMetadata.getName(), actualMetadata.getName());
+        } finally {
+            deleteDirectory(projectDir);
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


main/tests/server/src/com/google/refine/io/ProjectMetadataUtilitiesTest.java
```diff
@@ -60,4 +60,33 @@ public void testSaveAndLoadFile() throws Exception {
...
+            File metadataFile = new File(projectDir, "metadata.json");
+
+            expectedMetadata = ProjectMetadataUtilities.loadFromFile(metadataFile);
+            Assert.assertEquals(expectedMetadata.getName(), actualMetadata.getName());
+        } finally {
+            deleteDirectory(projectDir);
+        }
+    }
+
+    private void deleteDirectory(File directoryToBeDeleted) {
```
wetneb, line 82: Why note use `FileUtils.deleteDirectory` directory?

main/tests/server/src/com/google/refine/io/ProjectMetadataUtilitiesTest.java
```diff
@@ -60,4 +60,33 @@ public void testSaveAndLoadFile() throws Exception {
             file.delete();
         }
     }
+    /**
+     * Check if the save method can correctly save the metadata to the target directory.
+     **/
+    @Test
+    public void testSave() throws Exception {
+        File projectDir = new File(System.getProperty("java.io.tmpdir"), "projectDir");
```
wetneb, line 68: Why not use `TestUtils.createTempDirectory` which we use in other places in tests?

wetneb (commented):
Hi, thanks for the contribution!

First off, I'd be interested to know what brought you to submitting this change: what is your motivation, how did you settle on adding a test case for this particular method? It's rather rare to get PRs focused on tests only without referring to a particular issue to solve.

Second, the test scenario is quite similar to the other test case above (`testSaveAndLoadFile`) so perhaps the added value isn't super high, no? Looking at the method you are testing, there is quite some logic to dance around the existing file to be overwritten (not sure to what extent it's really needed by the way), so perhaps you could tweak your test case so it demonstrates some of that? For instance by creating a `metadata.json` file earlier in the test.

Finally, the CI is failing because of linting issues: you should be able to solve them with `./refine lint`.

wetneb:
This PR looks abandoned to me so I'm closing it. Happy to reopen if you want to go back to it.

PR #6648, opened 2024-05-28
Improve localization of row/record count bar

### LINKED ISSUE #6647: Incorrect localization for row/record count in main summary bar
Our UI currently looks like this in Italian:
image
I find it a bit embarrassing to have "matching" in there: it's in fact because we don't let translators localize this word.

wetneb:
Fixes #6647.

This fixes non-localizable words in row/record count bar ("matching", "total").
This also renders the row/record counts in the user locale (such as 12,345 instead of 12345 in English).

Screenshot:
image


main/tests/cypress/cypress/e2e/tutorial/layout-of-openrefine.cy.js
```diff
@@ -25,13 +25,13 @@ describe(__filename, function () {
     cy.get('.dialog-container input[bind="separatorInput"]').type('|');
     cy.confirmDialogPanel();
 
-    cy.get('#summary-bar').should('to.contain', '4009 rows');
+    cy.get('#summary-bar').should('to.contain', '4,009 rows');
```
tfmorris, line 28: Adding a thousands separator seems like an unrelated change. We probably want to be consistent in the UI about whether or not we use a thousands separator and there are likely some places where space is tighter than here.

tfmorris (approved):

main/tests/cypress/cypress/e2e/tutorial/layout-of-openrefine.cy.js
wetneb: No strong feelings about keeping it or not - I introduced it in the new architecture because I find it visually difficult to parse large row/record counts in the 10M ballpark (so it's perhaps not so useful without those capabilities). While it could indeed be used in other places, this is the place where such large counts are displayed the most, and in other places (like facet statistics) numbers tend to be smaller and need to be more compact.

tfmorris: If the designers are OK with using thousands separators some places and not others, I'm fine with this (and already approved the PR).

PR #6651, opened 2024-05-29
Separate 'core' and 'grel' Maven modules

### LINKED ISSUE #6556: Reusable java test utilities for extensions
As an extension developer, I want to be able to easily write unit tests for the java classes I write.
The `RefineTest` and `TestUtils` classes that we use in the `main` module are test classes that are not designed for public consumption.

We could have a separate Maven module which would offer basic utilities (covering things such as workspace initialization, teardown and project creation) so that extensions can easily depend on such utilities for their tests.

We could also investigate how to do the same for Cypress tests, but I would keep that for a separate issue.

I intend to tackle this as part of the Maven modularization effort.

wetneb:
This introduces new Maven modules:
* a `core` module (which could also be named differently to avoid the confusion with the current `main`), containing the core classes defining OpenRefine's data model and main interfaces (what is a project, an operation, an importer, an expression language…). This module has  minimal dependencies to external libraries (such as Butterfly, Jackson and so on)
* a `grel` module which contains the implementation of our default expression language.

The main motivation is to prepare for the new architecture, primarily with the separation between `core` and `main`, so that the `core` module can contain various interfaces whose implementations are provided by other modules (Spark-based runner, local runner, test runner). But it also has other advantages:
* making it possible for extensions to narrow down the scope of classes from OpenRefine that they depend on. At the moment extensions generally depend on our `main` artifact, meaning that they are pulling a ton of dependencies in their classpath. After this move, they can still do that, but many of them should be able to move to only depending on `core` (and possibly `grel`) , which reduces the amount of libraries they can start depending on accidentally (because those libraries will not be visible at all during development). Down the line, we could consider somehow enforcing that extensions indeed do not depend on `main` (for instance by not publishing it in Maven Central) and having the required classpath isolation at runtime.
* provide the same sorts of isolation benefits within our own code base, avoiding things like https://github.com/OpenRefine/OpenRefine/pull/6557
* provide smaller test utilities (Closes #6556) - the current solution makes extensions download not just our test utilities, but our test suite as a whole. In earlier attempts, I have introduced a dedicated `testing` module for this purpose, but it seems to me that the `test-jar` approach is much more natural and flexible.
* perhaps provide a path towards versioning GREL independently of OpenRefine, if we want to get serious about reproducibility and backwards-compatibility, as we regularly introduce changes to GREL behaviour which are technically breaking… but I am really not sure it's worth the effort. Also, this PR intentionally avoids making changes to the Java classes so that the renames are reliably detected by git, making it easy to merge/rebase things on top of this change.

Caveats:
* this will likely have some negative impact on coverage statistics, but it does not change the actual overall coverage (all tests are retained)
* this will require updating paths to localization files in Weblate after merging
* it could be worth adapting the package names so that they reflect which module they are part of. That's a breaking change though.
* I would recommend re-importing the project in IDEs after the change (unless the IDEs are clever enough to pick up the new modules automatically)
* It's worth rebasing this PR just before merging to make sure any newly introduced Java file ends up in the right place, to avoid build errors


tfmorris:
Generally I like the idea of more modularity and smaller modules. I've started looking at this PR, but don't think I understand it well enough to have an opinion yet. ~It doesn't currently build.~ I ~may try to~ fixed that as an exercise in better understanding things.

I'm going to write down some notes here on things I've discovered as I review:
- remaining `main` module - 146 Java source files, 108 Java test files, 294 total files
- new `core` module - 204 Java source modules, 58 Java test modules, 311 total files
- new `grel` module - 137 Java source modules, 66 Java test modules, 336 total files
- the new UserDefinedKeyer tests conflict with the modularization. It might be easily fixable, but I just commented them out to get things to build.
- the `core` module includes all the Java interfaces, but also a surprisingly large (at least on the surface) number of concrete implementation classes. I'm not sure if this is a sign that the modularity could be improved further.
- another potential change would be aligning the src/test directories with the Maven naming convention. It's confusing to have different parts of the directory tree following different naming conventions.

I'll try to continue to poke around at the new structure to better understand what's where and what the dependency structure looks like.

tfmorris:
p.s. I pushed the rebased version to @wetneb 's branch, but I may not have caught all the new things which have been added.

wetneb:
Thanks for having a look!

> the core module includes [...] a surprisingly large (at least on the surface) number of concrete implementation classes. I'm not sure if this is a sign that the modularity could be improved further.

If you have ideas of things that should be abstracted away behind interfaces, I'm happy to hear that. But I think there is value in keeping this PR to be one that only moves files around and not mix it up with other refactorings.

Also I am happy to take care of rebasing. Force-pushing to an actively maintained fork can be a bit tricky because it risks losing commits if I happen to push to it in the meantime. So let me uncheck the "Maintainers are allowed to edit this pull request" checkbox.

tfmorris:
Sorry about that. Since Github records the before and after commit hashes for the force push, I considered it safe and recoverable. I'll keep my rebases private in the future.

tfmorris (commented):
- Can you summarize what the binary and source compatibility implications of this refactoring are?
- The new LevenshteinDistance function needs to be moved when this is rebased. I think you may also need #6792 to get rid of the JUnit usages (or at least I did, but I didn't investigate too deeply).
- Clojure seems to be a bit of an odd man out with Jython & GREL in their own modules

The build seems to produce:

- ./main/webapp/WEB-INF/lib/grel-3.9-SNAPSHOT.jar
- ./main/webapp/WEB-INF/lib/core-3.9-SNAPSHOT.jar
- ./modules/grel/target/openrefine.jar
- ./modules/grel/target/openrefine-tests.jar
- ./modules/core/target/openrefine.jar
- ./modules/core/target/openrefine-tests.jar

I would expect the jars to be named something like openrefine-core and openrefine-grel.

wetneb:
> Can you summarize what the binary and source compatibility implications of this refactoring are?

I think this PR in itself should not have any consequence to runtime code: any extension which depends of our current `main` artifact will still get access to all the classes moved to `core` and `grel` since they are dependencies. That being said, for extensions which depend on the test-jar of the `main` artifact, there could be a need to add dependencies to the `core` or `grel` test-jar.
If anything, it should be the preliminary refactorings (already merged) which could cause some breakages, since those contain actual code changes.

> Clojure seems to be a bit of an odd man out with Jython & GREL in their own modules

Maybe it's worth moving it to a bundled extension like `jython`?

> I would expect the jars to be named something like openrefine-core and openrefine-grel.

Good catch.

wetneb:
@tfmorris if you feel like merging this, it's just freshly rebased

tfmorris:
I'm starting to regret not having a deeper (and more public) conversation about the compatibility of this restructuring. I'm not confident that there are no downstream implications given the repackaging into separate JARs. It seems like it's going to affect all packagers.

I also failed to notice earlier that some packages are split across modules. This is something that JPMS disallows, so we would need to change again for that.

I wonder if we shouldn't consider deferring this to post-3.9? I know it'll be a mess to try and back out. :(

wetneb:
> I'm not confident that there are no downstream implications given the repackaging into separate JARs. It seems like it's going to affect all packagers.

Can you explain a bit more in which scenario you fear this is going to cause problems?

I would assume that all packagers of OpenRefine rely on our existing Maven-based build system. If that's the case, then continuing to invoke the Maven build will continue to produce jars for the `main` and `server` artifacts as before. The `main` artifact will have less classes in itself and two more dependencies, `core` and `grel`, which Maven builds and packages on its own and includes alongside the other dependencies. Adding or removing dependencies between versions is a routine operation, so I would find it unlikely that this causes a problem on its own. Where do you see this failing?

If some OpenRefine distribution is done by rewriting the build with another platform (say Gradle instead of Maven) then they will indeed need to update the build configuration files to reflect the new module structure, but that's something they would need to do at each release anyway (to update dependencies for instance).

> I also failed to notice earlier that some packages are split across modules. This is something that JPMS disallows, so we would need to change again for that.

It would clearly be better to have package names that reflect the separation into Maven artifacts, even without relying on JPMS, but any change of package names is breaking, so it's not clear to me why it would be better to do it before. Perhaps one approach could be to change the package names and adapt the class loader with custom code so that the old package names still work… it feels pretty hacky but if we only care about compatibility with extensions, it could potentially work as we have control over which class loader is used to load them. Whether that can play out well with JPMS down the line, I have no idea.

To me, this separation is a way for us to start isolating functionality internally without having adverse effects on reusers. As long we don't do that, we'll continue to develop accidental inter-dependencies in the `main` monolith, making it harder to separate things out later.

tfmorris:
I should have been clearer that my comments are specifically in relation to the point release 3.9 where we want to preserve compatibility. I don't disagree with moving towards increased modularity.

The packaging concern is that `openrefine-main.jar` no longer contains all the classes of the API, something that was flagged by the API compatibility checker when I ran it. At a minimum it seems like downstream folks are going to need to adjust their classpaths, JAR copy scripts, etc.

In retrospect, it seems like a risky change for a point release. I'd rather address our regrets now than after the release -- or convince ourselves that there is no risk of regrets.

wetneb:
What I'd like to understand is what sorts of "downstream folks" you are worried about exactly. Do you have any example to share (even a made up scenario - it doesn't need to be a known extension or packaging)? The notion of "preserving compatibility" needs to be qualified to specify what classes of uses are expected to be preserved. Someone could decide to build something that depends of the exact binary layout of our DMG file and we would almost surely break compatibility at each release for that.

From what I can tell, the `openrefine-main.jar` was never self-contained, as it required additional jars for the dependencies it relies on. Some of which are even part of the extension API (jackson, jetty, probably others).
If anyone packages OpenRefine by copying around a fixed list of jars, this workflow already breaks every time we change (or likely even update) our dependencies, so I find that rather unlikely.

wetneb:
Thanks. I would still like to understand why any such packager would choose to copy a fixed list of jar files into their package.
Our `main/webapp/WEB-INF/lib` contains 91 different jar files at the moment:
```
annotations-12.0.jar
arithcode-1.2.jar
butterfly-1.2.6.jar
caffeine-3.1.8.jar
…
core-3.9-SNAPSHOT.jar
grel-3.9-SNAPSHOT.jar
…
openrefine-main.jar
```

In the past, we've routinely added, removed and updated files in this directory.
Any packaging pipeline which breaks when such changes happen would have been broken by the past minor releases, so I am not sure why one would rely on such a pipeline.

Recently, we have also changed the directory structure of our packages by bundling our own `.class` files into jars (https://github.com/OpenRefine/OpenRefine/issues/6257, released in 3.8). Do you also regret this move in retrospect?

I would be happy to analyze a few examples (Docker images, Debian package, AUR package, Chocolatey package, Homebrew package), showing how each of them copy over the entire contents of the `main/webapp/WEB-INF/lib`  directory into their package. But I don't know if I should do that, as my survey wouldn't be comprehensive and that might not actually do anything to reassure you?

More generally, I would argue that our semantic versioning should reflect compatibility with extensions, and not the specifics of the archive contents of our released artifacts. OpenRefine is an end user application, not a library.

PR #6653, opened 2024-05-31
Restore public API. Fixes #6652

### LINKED ISSUE #6652: Restore deleted constructor to StandardReconConfig
Per https://github.com/OpenRefine/OpenRefine/issues/6581#issuecomment-2138895162 fix incompatible public API changes that cause compilation errors:
```
no suitable constructor found for StandardReconConfig(java.lang.String,java.lang.String,java.lang.String,java.lang.String,java.lang.String,boolean,java.util.ArrayList<com.google.refine.model.recon.StandardReconConfig.ColumnDetail>,int)
```

\
### Current Results
Existing extensions don't compile.

### Expected Behavior
Both binary and source compatibility are preserved for extensions

tfmorris:
Draft while I figure out the intricacies of Jackson annotations...

Fixes #6652

- Restore deleted public constructors & getters

### DIFF
--- a/extensions/wikibase/tests/src/org/openrefine/wikibase/editing/NewEntityLibraryTest.java
+++ b/extensions/wikibase/tests/src/org/openrefine/wikibase/editing/NewEntityLibraryTest.java
@@ -62,7 +62,7 @@ public class NewEntityLibraryTest extends WikidataRefineTest {
         Project project = createProject(TestingData.inceptionColumns,
                 TestingData.inceptionProjectGridWithNewItem);
         StandardReconConfig config = new StandardReconConfig("http://my.endpoint",
-                "http://my.schema", "http://my.schema", "Q5", "human", true, 10, Collections.emptyList());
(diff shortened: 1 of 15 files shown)


tfmorris:
I'm not 100% sure of that the JSON serializations are compatible with the previous, but I think this is close enough for an initial review. I think the only potential issue is with mixed version client/server, which we don't support.

- restores deleted public constructors and methods
- restores original unit tests
- reverts unnecessary changes to calling code (e.g. places where it's just passing the default for the new batchSize parameter)
- marks two new constructors as being deprecated (we don't need the combinatorial explosion). They were only introduced in 3.8 in April, so I could be convinced to be more aggressive and just delete them outright (with a release note).

wetneb (approved):
We could consider introducing builder classes in such a case, to make it easier to skip optional parameters.

wetneb:
Also I wouldn't delete the methods outright: the Commons extension has already been adapted to use the new constructor.

PR #6658, opened 2024-06-03
Disable dependabot grouping for java dependencies

wetneb:
Given the failures of #6650, #6656, let's abandon grouping for java dependencies.

I think the grouping for JS dependencies seems to be worth retaining: #6655 is nice.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,12 +19,7 @@ updates:
       - ">= 10.0.0"
   - dependency-name: "formatter-maven-plugin"
   target-branch: "master"
-  groups:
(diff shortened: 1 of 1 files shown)


tfmorris (approved):
I'm OK with giving up on this, but it looks like we could probably used the `exclude-patterns` option to the `groups` directive to exclude Jena and require dependabot to generate separate PRs (which would presumably honor the the ignore-via-closing that we have in place already).

wetneb:
Ok! Happy to experiment more, anyway it's not like it's critical that we update as soon as new versions are published in general.

PR #6659, opened 2024-06-04
Add dialog to warn of history entry deletion

### LINKED ISSUE #3184: Add warnings for Undo/Redo history changes

wetneb:
Backports #5577 (see the discussion there) to master. Fixes #3184.


main/webapp/modules/core/scripts/project.js
```diff
@@ -436,20 +436,29 @@ Refine.postProcess = function(moduleName, command, params, body, updateOptions,
...
-    body,
-    onDone,
-    "json"
-  );
+    window.setTimeout(function() {
+        if (!done) {
+        dismissBusy = DialogSystem.showBusy();
+        }
+    }, 500);
+  }
```
tfmorris, line 454: Indentation looks off here

```suggestion
    window.setTimeout(function() {
      if (!done) {
        dismissBusy = DialogSystem.showBusy();
      }
    }, 500);
  }
```

tfmorris (changes requested):
A couple of minor style/wording comments, but it can be merged without an additional review if those are addressed.

tfmorris (approved):
Thanks! LGTM

Issue #6613, opened 2024-05-20
Add new GREL function to calculate the edit distance


zyadtaha:
It would be easier if we could calculate number of edits required to make one value perfectly match another using GREL.

### Proposed solution
Provide a new builtin GREL function,  called editDistance(), that takes 2 strings and return the the minimum number of single-character edits (insertions, deletions or substitutions) required to change one string into the other

For example:
`editDistance("New York", "NewYork")` -> `1`
`editDistance(“M. Makeba”, “Miriam Makeba”)` -> `5`

### Alternatives considered
Using Jython

thadguidry:
1. Also, do we code our own, and optimize and use only a single 1-D array?
https://www.geeksforgeeks.org/edit-distance-dp-5/

2. or use Apache Commons Text and then which algorithm for edit distance?
   - A. Do we use a memoization implementation with a particular algorithm?
   - B. Or use an algorithm that uses less memory like with a single 1-D array ([`unlimitedCompare`](https://github.com/apache/commons-text/blob/e7479e23b44a2257d61759ef897d20bc7cdbff2b/src/main/java/org/apache/commons/text/similarity/LevenshteinDistance.java#L259))
https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/similarity/EditDistance.html

Which algorithm would be best if folks want to compute edit distances between two VERY VERY large strings (one version of a books' contents to another version of a books' content) ?

The reason there are many algorithms is that there is no 1 size fits all.
We might have to offer 2 or 3 `editDistance` functions, or name the edit distance functions appropriately for the underlying algorithm and document the functions well.

dino2580:
https://github.com/OpenRefine/OpenRefine/pull/6627#issue-2313463452 Can you suggest some improvements

thadguidry:
@dino2580 Yes.  Don't roll our own edit distance implementation.  Instead, use the standard highly optimized Apache Commons Text library that already has intelligence for using 2D or 1D cost tables for edit distance using the Levenshtein distance metric.  Code details in https://github.com/apache/commons-text/blob/master/src/main/java/org/apache/commons/text/similarity/LevenshteinDistance.java

[...]

Issue #10, opened 2012-10-15
OAuth fails on sign in


tfmorris:
_Original author: iainsproat (May 10, 2010 16:14:55)_

Using Windows XP, Java Runtime 1.6.0 rev 20, Gridworks r667

Steps to reproduce: #  1 Using a project with columns mapped to Freebase schema. #  2 clicked through to Load into Freebase. #  3 clicked on 'Sign into Freebase'.

The below error is produced (gridworks logging set to verbose), and json 
dumped as a file in the browser.

19:11:19.031 [                  servlet]  &gt; GET check-authorization (16ms)
19:11:19.078 [                  servlet]  &lt; GET check-authorization (47ms)
19:11:30.781 [                  servlet]  &gt; GET authorize (11703ms)
19:11:32.859 [     org.apache.http.wire]   &gt;&gt; &quot;POST 
/api/oauth/request_token HTTP/1.1[EOL]&quot; (2078ms)
19:11:32.859 [     org.apache.http.wire]   &gt;&gt; &quot;Authorization: OAuth 
realm=&quot;freebase.com&quot;, 
oauth_callback=&quot;http&#37;3A&#37;2F&#37;2F127.0.0.1&#37;3A3333&#37;2Fcommand&#37;2Fauthorize&#37;2Fwww.f
reebase.com&quot;, oauth_consumer_key=&quot;&#37;239202a8c04000641f80000000keykey&quot;, 
oauth_version=&quot;1.0&quot;, oauth_signature_method=&quot;HMAC-SHA1&quot;, 
oauth_timestamp=&quot;1273504291&quot;, oauth_nonce=&quot;-6157124724385079801&quot;, 
oauth_signature=&quot;S872Vk0YA7IaMnz31IA88signature&quot;[EOL]&quot; (0ms)
19:11:32.859 [     org.apache.http.wire]   &gt;&gt; &quot;Content-Length: 0[EOL]&quot;(0ms)
19:11:32.859 [     org.apache.http.wire]   &gt;&gt; &quot;Host: 
www.freebase.com[EOL]&quot;(0ms)
19:11:32.859 [     org.apache.http.wire]   &gt;&gt; &quot;Connection: Keep-
Alive[EOL]&quot;(0ms)
19:11:32.859 [     org.apache.http.wire]   &gt;&gt; &quot;User-Agent: Apache-
HttpClient/4.0.1 (java 1.5)[EOL]&quot; (0ms)
[...]

PR #6674, opened 2024-06-17
UI: Clustering with Custom Distances and Keyers

zyadtaha:
The second pr in the GSoC project #4301 

- Replace `<select>` with custom dropdown
- Update test setup to include dropdown menu functionality

Here's what the dropdown menu looks like now:
Screenshot from 2024-06-17 08-03-24

### DIFF
--- a/main/webapp/modules/core/styles/dialogs/clustering-dialog.css
+++ b/main/webapp/modules/core/styles/dialogs/clustering-dialog.css
@@ -172,18 +172,18 @@ table.clustering-dialog-entry-table a:hover {
   display: inline-block;
 }
 
-.keyingDropdown {
+.keying-dropdown {
   display: inline-block;
 }
 
-#keyingDropdownBtn {
+#keying-dropdown-btn {
   min-width: 130px;
   text-align: left;
   background-color: #ffffff;
   border: 1px solid #727272;
 }
 
-.keyingFunctionSelector {
+.keying-function-selector {
   display: none;
   position: absolute;
   background-color: #f9f9f9;
@@ -192,14 +192,14 @@ table.clustering-dialog-entry-table a:hover {
   z-index: 1;
 }
 
-.keyingFunctionSelector a {
+.keying-function-selector a {
   color: black;
   padding: 4px 8px;
   text-decoration: none;
   display: block;
 }
 
-.keyingFunctionSelector a:hover {
+.keying-function-selector a:hover {
   background-color: #f1f1f1;
 }
 
@@ -207,18 +207,18 @@ table.clustering-dialog-entry-table a:hover {
   position: relative;
 }
 
-.knnDropdown {
+.knn-dropdown {
   display: inline-block;
 }
 
-#knnDropdownBtn {
+#knn-dropdown-btn {
   min-width: 130px;
   text-align: left;
   background-color: #ffffff;
   border: 1px solid #727272;
 }
 
-.distanceFunctionSelector {
+.distance-function-selector {
   display: none;
   position: absolute;
   background-color: #f9f9f9;
@@ -227,14 +227,14 @@ table.clustering-dialog-entry-table a:hover {
   z-index: 1;
 }
 
-.distanceFunctionSelector a {
+.distance-function-selector a {
   color: black;
   padding: 4px 8px;
   text-decoration: none;
   display: block;
 }
 
-.distanceFunctionSelector a:hover {
+.distance-function-selector a:hover {
   background-color: #f1f1f1;
 }
 
--- a/main/webapp/modules/core/styles/dialogs/clustering-dialog.css
+++ b/main/webapp/modules/core/styles/dialogs/clustering-dialog.css
@@ -166,3 +166,78 @@ table.clustering-dialog-entry-table a:hover {
   text-decoration: none;
   visibility: hidden;
 }
+
+.binning-controls {
+  position: relative;
+  display: inline-block;
+}
+
+.keyingDropdown {
+  display: inline-block;
+}
+
+#keyingDropdownBtn {
+  min-width: 130px;
+  text-align: left;
+  background-color: #ffffff;
+  border: 1px solid #727272;
+}
+
+.keyingFunctionSelector {
+  display: none;
+  position: absolute;
+  background-color: #f9f9f9;
+  min-width: 130px;
+  box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.2);
+  z-index: 1;
+}
+
+.keyingFunctionSelector a {
+  color: black;
+  padding: 4px 8px;
+  text-decoration: none;
+  display: block;
+}
+
+.keyingFunctionSelector a:hover {
+  background-color: #f1f1f1;
+}
+
+.knn-controls {
+  position: relative;
+}
+
+.knnDropdown {
+  display: inline-block;
+}
+
+#knnDropdownBtn {
+  min-width: 130px;
+  text-align: left;
+  background-color: #ffffff;
+  border: 1px solid #727272;
+}
+
+.distanceFunctionSelector {
+  display: none;
+  position: absolute;
+  background-color: #f9f9f9;
+  min-width: 130px;
+  box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.2);
+  z-index: 1;
+}
+
+.distanceFunctionSelector a {
+  color: black;
+  padding: 4px 8px;
+  text-decoration: none;
+  display: block;
+}
+
+.distanceFunctionSelector a:hover {
+  background-color: #f1f1f1;
+}
+
+.show {
+  display: block;
+}
--- a/main/tests/cypress/cypress/e2e/project/grid/column/edit-cells/cluster-and-edit.cy.js
+++ b/main/tests/cypress/cypress/e2e/project/grid/column/edit-cells/cluster-and-edit.cy.js
@@ -56,7 +56,7 @@ describe(__filename, function () {
             {
                 methodName: 'Key collision',
                 selector: 'keyingFunctionSelector',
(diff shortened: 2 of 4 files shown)


zyadtaha:
Hi @wetneb,

I updated the functions selector to a dropdown menu. Now, I'm considering changing the method selector too. What do you think about that?

Also, should I complete all the frontend development in one pull request (this one), or would you prefer multiple pull requests for easier reviewing?

I know you're busy with the barcamp, so take your time to reply. I'm not in a rush.

Thank you!

wetneb:
Congrats for getting this done so quickly! Indeed the BarCamp is keeping me quite busy but I will try to have a closer look before our meeting on Thursday.

wetneb:
To give a bit more context to the rest of the community, it would be worth pointing to your proposed design, which will help understand why you are converting a native drop-down select to something else (because it would not make sense to do that in isolation, without wanting to make further changes)

> I updated the functions selector to a dropdown menu. Now, I'm considering changing the method selector too. What do you think about that?

I think we should stick to native widgets as much as possible. If you foresee needing to make similar changes to the method selector it would make sense, but I am not sure why we would need that? What sort of changes do you have in mind?

Before reviewing the code in details, I'd like to list the broad sorts of problems I can think of in this case:
* accessibility might not be as good as the native widget. For instance, I noticed that it's not possible (or I didn't figure out how) to select an option using the keyboard only.
* visual style: it needs to look natural. I think you've done a pretty good job at that (at least on my platform and to my taste) but I wonder if it would be worth re-introducing some sort of downwards arrow like the ones native widgets have (at least on my platform):
image

But I think it doesn't need to look exactly like a native drop-down menu. It's likely not possible to do anyway, given that the rendering likely depends on the operating system / browser in many ways. You could also "own it" and go for a fairly different design, I guess.

main/webapp/modules/core/styles/dialogs/clustering-dialog.css
```diff
@@ -166,3 +166,78 @@ table.clustering-dialog-entry-table a:hover {
...
+}
+
+#keyingDropdownBtn {
+  min-width: 130px;
+  text-align: left;
+  background-color: #ffffff;
+  border: 1px solid #727272;
+}
+
+.keyingFunctionSelector {
```
wetneb, line 186: Nitpick: for consistency with the rest of the class names, you could stick to this syntax:
```suggestion
.keying-function-selector {
```
and similarly for other CSS rules in this file

ostephens:
> To give a bit more context to the rest of the community, it would be worth pointing to your proposed design, which will help understand why you are converting a native drop-down select to something else (because it would not make sense to do that in isolation, without wanting to make further changes)

Just to give a +1 to this - I'm very interested in understanding the approach being taken overall here

zyadtaha:
> > To give a bit more context to the rest of the community, it would be worth pointing to your proposed design, which will help understand why you are converting a native drop-down select to something else (because it would not make sense to do that in isolation, without wanting to make further changes)
> 
> Just to give a +1 to this - I'm very interested in understanding the approach being taken overall here

The main reason is that we can't dynamically add a new option to the native dropdown (created with the <select> element).

Also, to move the "Add new function" feature into the dropdown menu which will save a lot of space in the already crowded clustering dialog window.

Here's an example.
Screenshot from 2024-06-21 21-15-10

zyadtaha:
I'm not sure why, but it seems the picture didn't show up in my last comment.image

tfmorris:
> The main reason is that we can't dynamically add a new option to the native dropdown (created with the element).

Can you expand on this? The current drop down menus are constructed dynamically from Javascript, so, on the surface, it seems like the same technique could possibly be used to change them at run time.

zyadtaha:
> > The main reason is that we can't dynamically add a new option to the native dropdown (created with the element).
> 
> Can you expand on this? The current drop down menus are constructed dynamically from Javascript, so, on the surface, it seems like the same technique could possibly be used to change them at run time.

Yes, you are correct. I apologize for the confusion. 

The reason is that a `<select>` element is restricted to containing only `<option>` elements.

Here are the implications:
- It's invalid to add an `<hr>` element inside a `<select>` element to separate options.
- It's not possible to add a delete icon to each option using the native dropdown menu.
- As the number of custom expressions grows, we will require dropdown menus with search bars like this. 
Screenshot from 2024-06-22 02-24-57
However, we cannot place the search bar inside the `<select>` element for the same reason as before. This arrangement would appear odd, with the search bar outside the dropdown menu.
Screenshot from 2024-06-22 02-29-02

Abbe98:
Are you sure you want a dropdown menu at all? Seems like it will be very hard to make it accessible and in line with the current UI given your requirements. Why not a modal with a filterable list instead? Or simply a `<datalist>`?

antoine2711:
> * It's invalid to add an `<hr>` element inside a `<select>` element to separate options.

I don’t understand, @zyadtaha.

This works in Safari and in Chrome:
```
  <select name="papersize">
    <option>Select a paper size</option>
    <hr>
    <option>5.5 × 8.5 in</option>
    <option>8.5 × 11.0 in</option>
    <option>8.5 × 14.0 in</option>
    <option>11.0 × 17.0 in</option>
    <hr>
    <option>A3</option>
    <option>A4</option>
    <option>A5</option>
    <option>A6</option>
    <hr>
    <option>Envelope #10</option>
    <option>Envelope B5</option>
    <option>Envelope C5</option>
    <option>Envelope Monarch</option>
</select>
```

What do you mean?

Regards, Antoine

zyadtaha:
> > * It's invalid to add an `<hr>` element inside a `<select>` element to separate options.
> 
> I don’t understand, @zyadtaha.
> 
> This works in Safari and in Chrome:
> 
> ```
>   <select name="papersize">
>     <option>Select a paper size</option>
>     <hr>
>     <option>5.5 × 8.5 in</option>
>     <option>8.5 × 11.0 in</option>
>     <option>8.5 × 14.0 in</option>
>     <option>11.0 × 17.0 in</option>
>     <hr>
>     <option>A3</option>
>     <option>A4</option>
>     <option>A5</option>
>     <option>A6</option>
>     <hr>
>     <option>Envelope #10</option>
>     <option>Envelope B5</option>
>     <option>Envelope C5</option>
>     <option>Envelope Monarch</option>
> </select>
> ```
> 
> See: https://webkit.org/blog/14933/bringing-back-horizontal-rules-in-select-elements/
> 
> What do you mean?
> 
> Regards, Antoine

Thank you for correcting me. It seems I made a mistake in my search. But we still can't add icons or input elements, can we?

zyadtaha:
> Are you sure you want a dropdown menu at all? Seems like it will be very hard to make it accessible and in line with the current UI given your requirements. Why not a modal with a filterable list instead? Or simply a `<datalist>` element?

What are your thoughts regarding the current options? @wetneb @ostephens @tfmorris @antoine2711

wetneb:
I think it's worth experimenting with other designs! I don't know what a "modal with a filterable list" or a "`<datalist>`" would look like. Perhaps you could try re-designing this selection, with the understanding that people are ready to see bigger UI changes there and you don't need to squeeze in your feature in the exact current layout.

ostephens:
@zyadtaha thanks for the extra information. Can I check I understand the expected workflow? Is the idea:

1. User goes to cluster screen
2. User goes to select keying or distance function in dropdown
3. User finds no suitable function, so clicks "Add new"
4. Some screen or dialog presented where user can define their function (have you given any thought to what this would look like yet?)
5. User creates/saves function
6. User returned to select list with new option added

Have I got that correct?

zyadtaha:
> 4. Some screen or dialog presented where user can define their function (have you given any thought to what this would look like yet?)
>
- We've decided to use the current expression dialog, which has tabs for Preview, History, and Favorites. The History tab lets users see past expressions with option to star. This dialog also allows users to choose their preferred expression language, not limited to GREL.
- When adding a custom distance function, it's important to explain which variables users can use, like value1 and value2. To address this, we've included a default expression in the editor that uses these values (for example, editDistance(value1, value2) as shown in #6613).
> 7. User selects new function (or automatically selected?)
>
Yes, I would go for automatic selection. 

> Have I got that correct?
>
Yes, absolutely.

ostephens:
Thanks @zyadtaha. I guess the question then becomes how would the user generally manage the expressions they had written? It seems like there should be some option to Edit and Delete the expression (possibly duplicate as well?). While I can see the create option working OK as you describe above, I think adding in any additional functions could make the UI very crowded.

This makes me think that at least worth considering having the options for managing the available functions somewhere outside the dropdown menu. Potentially the two approaches could be combined of course, but splitting this out could avoid having this custom dropdown and offer a bit more space. I'd be interested to know how others (both on this thread @antoine2711 @Abbe98 @tfmorris and other users who are maybe on the forum but not on GitHub) might feel about this.

If you've already had these discussions somewhere and decided not to go this way then of course I respect that!

wetneb:
The original discussion is indeed on the forum:
https://forum.openrefine.org/t/user-defined-clustering-project/1528/3?u=antonin_d
In this discussion there was support from Benjamin (not sure about GitHub username) for integrating the delete buttons inside the drop-down.

ostephens:
Thanks @wetneb  - that's very helpful. I don't want to re-hash discussions from the forum here. I do think those designs are missing an option to edit the user defined function which I think could be important - I can add a comment to the forum.

tfmorris:
I think @Abbe98 's point about accessibility is a good one. Also, it's a red flag to me that no one has created a filterable pulldown menu with deletable items that you can reuse. Using a dialog instead will give you a lot more real estate and navigational flexibility to work with when designing the UX.

zyadtaha:
Closing this to make the PR simpler.

wetneb:
There is a new version of this PR at #6705: you are all encouraged to chime in on the new design there.

PR #6696, opened 2024-06-25
Fix preservation of pagination when matching a single cell

### LINKED ISSUE #6695: Regression: matching a single cell to a reconciliation candidate resets the grid to the first page
In 3.8.1 (and the current master branch), matching a single cell to a reconciliation candidate resets the grid to the first page.
Before, matching a single cell to a candidate (with the button with a single check-mark) did not do that - only matching a cell to multiple candidates (with the double check-mark) did.

### To Reproduce
1. First, reconcile a column
2. Then, move to a different page in the grid
3. Finally, match a single cell to a reconciliation candidate

### Current Results

The cell is matched and the grid view is reset to the first page.

### Expected Behavior

The position in the grid should be preserved.

### Versions
 - OpenRefine: 3.8.1, master

### Additional context

Reported by @arthurpsmith on Mastodon: https://fediscience.org/@apsmith/112672042744786060.

This will be fixed by #6546, but because this is a pretty bad bug, it might be worth investigating if we can include a smaller fix to restore the 3.7 behaviour in 3.8. Given the scale of the changes in #6546 I'd be reluctant to port it as such to 3.8.

wetneb:
Fixes #6695.

- revert #6239 which forces the re-fetching and re-rendering of the entire grid
- add missing `cellUI` parameter to enable in-place rendering of the updated cell after a single match action
- add a Cypress regression test

I would like to backport this to 3.8.

### DIFF
--- a/main/tests/cypress/cypress/e2e/project/grid/column/reconcile/actions/cell-buttons.cy.js
+++ b/main/tests/cypress/cypress/e2e/project/grid/column/reconcile/actions/cell-buttons.cy.js
@@ -51,6 +51,30 @@ describe('In-cell reconciliation buttons', () => {
         //confirming the no. of candidates after we click on see less
         cy.getCell(0, 'species').find('.data-table-recon-candidate:visible').should('have.length', 4);
     });
+
+    it('Matching a single cell preserves pagination', () => {
+        cy.visitOpenRefine();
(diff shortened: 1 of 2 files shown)


tfmorris (approved):

tfmorris:
I agree with backporting this to 3.8

PR #6645, opened 2024-05-28
tests: Clean up unnecessary try-catch blocks

wetneb:
Closes #6644.

PR #6701, opened 2024-07-02
Fix type of error message returned by the Filter GREL control

### LINKED ISSUE #6694: Filter() control returns text string not error object when first argument is not an array
Compare the code for Filter
https://github.com/OpenRefine/OpenRefine/blob/05978a07fa8cdca0dc356233b178ef85d76c1676/main/src/com/google/refine/grel/controls/Filter.java#L70

to equivalent code in ForEach
https://github.com/OpenRefine/OpenRefine/blob/05978a07fa8cdca0dc356233b178ef85d76c1676/main/src/com/google/refine/grel/controls/ForEach.java#L74

Filter should return an eval error not a string at this point so it can be caught appropriately in GREL / UI

sunil-atheer:
Filter class argument type mismatch

Fixes #6694 

- Filter class now returns proper error if the argument is invalid
- Updated Filter test for params validation

### DIFF
--- a/main/resources/com/google/refine/grel/ControlEvalError.properties
+++ b/main/resources/com/google/refine/grel/ControlEvalError.properties
@@ -13,4 +13,4 @@ expects_second_arg_var_name={0} expects second argument to be a variable name
 expects_second_arg_index_var_name={0} expects second argument to be the index's variable name
 expects_third_arg_element_var_name={0} expects second argument to be the element's variable name
 expects_second_third_args_different={0} expects the second and third arguments to be different variable names
-expects_array_arg={0} is not an array
+expects_array_arg={0} expects an array as its first argument
--- a/main/resources/com/google/refine/grel/ControlEvalError.properties
+++ b/main/resources/com/google/refine/grel/ControlEvalError.properties
@@ -13,3 +13,4 @@ expects_second_arg_var_name={0} expects second argument to be a variable name
 expects_second_arg_index_var_name={0} expects second argument to be the index's variable name
 expects_third_arg_element_var_name={0} expects second argument to be the element's variable name
 expects_second_third_args_different={0} expects the second and third arguments to be different variable names
+expects_array_arg={0} is not an array
--- a/main/resources/com/google/refine/grel/ControlEvalError.properties
+++ b/main/resources/com/google/refine/grel/ControlEvalError.properties
@@ -13,4 +13,4 @@ expects_second_arg_var_name={0} expects second argument to be a variable name
 expects_second_arg_index_var_name={0} expects second argument to be the index's variable name
 expects_third_arg_element_var_name={0} expects second argument to be the element's variable name
 expects_second_third_args_different={0} expects the second and third arguments to be different variable names
-expects_array_arg={0} expects an array as its first argument
+expects_first_arg_array={0} expects an array as its first argument
--- a/main/resources/com/google/refine/grel/ControlEvalError.properties
+++ b/main/resources/com/google/refine/grel/ControlEvalError.properties
@@ -1,4 +1,3 @@
-filter=First argument is not an array
 foreach_index=First argument to forEachIndex is not an array or JSON object
 foreach=First argument to forEach is not an array or JSON object
 for_range=First, second, and third arguments of forRange must all be numbers
--- a/main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
+++ b/main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
@@ -27,14 +27,13 @@
 
 package com.google.refine.grel.controls;
 
-import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.assertThrows;
 import static org.testng.Assert.fail;
 
 import java.util.Properties;
 
 import org.testng.annotations.Test;
 
-import com.google.refine.expr.EvalError;
 import com.google.refine.expr.Evaluable;
 import com.google.refine.expr.MetaParser;
 import com.google.refine.expr.ParsingException;
@@ -49,31 +48,29 @@ public class FilterTests extends GrelTestBase {
         TestUtils.isSerializedTo(new Filter(), json);
     }
 
-    private void assertParseError(String expression) throws ParsingException {
-        Evaluable eval = MetaParser.parse("grel:" + expression);
-        Object result = eval.evaluate(bindings);
-        assertTrue(result instanceof EvalError, "Expression didn't return error : " + expression);
+    private void assertEvaluatesToSuccess(String expression)  {
+        try {
+            Evaluable eval = MetaParser.parse("grel:" + expression);
+            Object result = eval.evaluate(bindings);
+        }
+        catch (ParsingException e) {
+            fail("Expression returned error : " + expression);
+        }
     }
 
     @Test
     public void testInvalidParams() throws ParsingException {
         bindings = new Properties();
         bindings.put("v", "");
-        assertParseError("filter('test', v, v)");
-        try {
-            assertParseError("filter([], 1, 1)");
-            fail("Didn't throw a ParsingException for wrong argument type");
-        } catch (ParsingException e) {
-        }
-        try {
-            assertParseError("filter([], v)");
-            fail("Didn't throw a ParsingException for 2 arguments");
-        } catch (ParsingException e) {
-        }
-        try {
-            assertParseError("filter([])");
-            fail("Didn't throw a ParsingException for 1 argument");
-        } catch (ParsingException e) {
-        }
+        assertEvaluatesToSuccess("filter('test', v, v)");
+
+        assertThrows("Didn't throw a ParsingException for wrong argument type", ParsingException.class,
+                () -> MetaParser.parse("grel:filter([], 1, 1)"));
+
+        assertThrows("Didn't throw a ParsingException for 2 arguments", ParsingException.class,
+                () -> MetaParser.parse("grel:filter([], v)"));
+
+        assertThrows("Didn't throw a ParsingException for 1 argument", ParsingException.class,
+                () -> MetaParser.parse("grel:\"filter([])"));
     }
 }
--- a/main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
+++ b/main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
@@ -27,15 +27,48 @@
 
 package com.google.refine.grel.controls;
 
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
+
+import java.util.Properties;
+
 import org.testng.annotations.Test;
 
+import com.google.refine.expr.EvalError;
+import com.google.refine.expr.Evaluable;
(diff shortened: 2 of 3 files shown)


sunil-atheer:
@wetneb Can you review the fix

main/resources/com/google/refine/grel/ControlEvalError.properties
```diff
@@ -13,3 +13,4 @@ expects_second_arg_var_name={0} expects second argument to be a variable name
 expects_second_arg_index_var_name={0} expects second argument to be the index's variable name
 expects_third_arg_element_var_name={0} expects second argument to be the element's variable name
 expects_second_third_args_different={0} expects the second and third arguments to be different variable names
+expects_array_arg={0} is not an array
```
wetneb, line 16: I wouldn't introduce a new translation string: why not keep the existing one? It does not seem to be used anywhere so you can modify it as you wish.
Looking at the pattern of the error messages around it, I would suggest something like `{0} expects an array as its first argument`, where `{0}` gets replaced by the function name, not the value.

wetneb (changes requested):
Thank you for this contribution!

ostephens:
> Filter class argument type mismatch
> 
> Fixes #{issue number here}
> 
> ## Changes proposed in this pull request:

@sunil-atheer when you do a PR it helps if you fill out the details in the template as should here. Including the issue number here (rather than just in the title) will mean that GitHub automatically links together the issue and the PR, and summarising the changes is helpful (although very brief for a small request like this) so those reviewing understand what you are doing before looking at the detail in the code

ostephens:
I'd suggest that this PR could be a good opportunity to expand test coverage for the `filter` function to include tests for invalid parameter behaviour.

Filter tests are in https://github.com/OpenRefine/OpenRefine/blob/master/main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
See forEach tests for example of how to test for behaviour on invalid parameters https://github.com/OpenRefine/OpenRefine/blob/c4b0847e1f61952c1ce55dd37036aa3a56f08c51/main/tests/server/src/com/google/refine/grel/controls/ForEachTests.java#L59

main/resources/com/google/refine/grel/ControlEvalError.properties
ostephens: @wetneb That was my initial thought too. However I think it might be the case that `filter` actually accepts the same types for the first argument as `forEach` and `forEachIndex`. If this could be verified, then having a new translation string like proposed here and using it for all cases could make sense?

wetneb:
@sunil-atheer I encourage you to edit the description of your pull request, following @ostephens' advice. In the meantime I am changing the PR title already, so that it is more informative. That helps people understand what the pull request does, for instance when looking at [the list of our open pull requests](https://github.com/OpenRefine/OpenRefine/pulls).

main/resources/com/google/refine/grel/ControlEvalError.properties
```diff
@@ -13,4 +13,4 @@ expects_second_arg_var_name={0} expects second argument to be a variable name
 expects_second_arg_index_var_name={0} expects second argument to be the index's variable name
 expects_third_arg_element_var_name={0} expects second argument to be the element's variable name
 expects_second_third_args_different={0} expects the second and third arguments to be different variable names
-expects_array_arg={0} is not an array
\ No newline at end of file
+expects_array_arg={0} expects an array as its first argument
```
ostephens, line 16: For consistency with other properties here I'd suggest naming this `expects_first_arg_array` (in case later need to have a similar constraint on a second or third argument etc.)

main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
```diff
@@ -27,15 +27,48 @@
...
+        Object result = eval.evaluate(bindings);
+        assertTrue(result instanceof EvalError, "Expression didn't return error : " + expression);
+    }
+
+    @Test
+    public void testInvalidParams() throws ParsingException {
+        bindings = new Properties();
+        bindings.put("v", "");
+        assertParseError("filter('test', v, v)");
+        try {
```
ostephens, line 63: I think you could include a test for the use of the variable here as well (can be seen in foreach tests like `assertParseError("forEach([], 1, 1)");)`

main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
```diff
@@ -27,15 +27,48 @@
...
+        Evaluable eval = MetaParser.parse("grel:" + expression);
+        Object result = eval.evaluate(bindings);
+        assertTrue(result instanceof EvalError, "Expression didn't return error : " + expression);
+    }
+
+    @Test
+    public void testInvalidParams() throws ParsingException {
+        bindings = new Properties();
+        bindings.put("v", "");
+        assertParseError("filter('test', v, v)");
```
ostephens, line 62: I think this should also be wrapped in a try/throw in order for us to check both that the first argument is rejected because it's not an array, and the catch so we check that the result of that is that we throw an EvalError rather than a string?

Note this doesn't seem to have been done in forEach but I think it's the right thing to do
@wetneb am I right?

ostephens (changes requested):

main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
sunil-atheer: I did not include this as the 3rd arg is used if its a boolean

main/resources/com/google/refine/grel/ControlEvalError.properties
sunil-atheer: Updated key

main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
wetneb: Intuitively, rejecting the first argument because it's not array should not throw an exception at parsing stage, it should return an EvalError at evaluation stage, so there shouldn't be a need for a try/catch here.
Besides, to check that something throws an exception it's better to use `assertThrows()` as it makes the assertion clearer and the code simpler.

wetneb: No need to clean up the existing ones as I have a pull request open to address that: https://github.com/OpenRefine/OpenRefine/pull/6645. But it's helpful if you can use `assertThrows` for any new assertions you are making.

main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
ostephens: But we can still test that the second argument is a variable here?

sunilnatraj: Added test case for this.

ostephens (approved):

sunilnatraj (approved):

main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
```diff
@@ -27,15 +27,53 @@
...
+        try {
+            assertParseError("filter([], 1, 1)");
+            fail("Didn't throw a ParsingException for wrong argument type");
+        } catch (ParsingException e) {
+        }
```
wetneb, line 67: As mentioned in my earlier comment, it is better if you can use `assertThrows` to assert that a particular piece of code throws an exception. Here is an example to help you understand what it would look like:
```suggestion
        assertThrows("Didn't throw a ParsingException for wrong argument type", ParsingException.class,
                () -> MetaParser.parse("grel:filter([], 1, 1)"));
```
Because the error is coming from the parsing stage, not the evaluation stage, we can make the assertion narrower by calling the parser only and checking that it does throw.

Would you be able to adapt the rest of the test following this pattern?

wetneb (changes requested):

main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
sunilnatraj: @wetneb I updated all the cases to use assertThrows

main/tests/server/src/com/google/refine/grel/controls/FilterTests.java
```diff
@@ -27,15 +27,53 @@
...
-public class FilterTests {
+public class FilterTests extends GrelTestBase {
 
     @Test
     public void serializeFilter() {
         String json = "{\"description\":\"Evaluates expression a to an array. Then for each array element, binds its value to variable name v, evaluates expression test which should return a boolean. If the boolean is true, pushes v onto the result array.\",\"params\":\"expression a, variable v, expression test\",\"returns\":\"array\"}";
         TestUtils.isSerializedTo(new Filter(), json);
     }
+
+    private void assertParseError(String expression) throws ParsingException {
```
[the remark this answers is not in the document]
sunilnatraj: @wetneb As the 1st case is a positive case; i updated the name of the function and its working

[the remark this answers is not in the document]
wetneb: Well, now this first case does not assert that the returned value is an `EvalError` anymoe, which I think would be worth doing. So I would add back this assertion and change the name of this method to `assertEvaluatesToError`, again.

You seem to be unclear on the distinction between a parse error and an evaluation error. Let me try to explain this distinction with an analogy: instead of talking about GREL expressions, let's talk about sentences in English.
Consider the following sentences:
1. "The moon satellite a is of Elvis Presley." - > this sentence is syntactically incorrect: it has a parse error, because the words are in an incorrect order
2. "The moon is a satellite of Elvis Presley." -> this sentence is syntactically correct, but it does not make sense: humans don't have satellites.
3. "The moon is a satellite of the Earth." -> this sentence is syntactically correct, and its meaning is true.

As you can see in those examples, sentences in English can have (at least) two sorts of problems:
1. being syntactically incorrect: that means that their structure does not respect the grammar of the language
2. being semantically wrong: that means that although their structure respects the rules of the language, they don't make sense

Similarly, in GREL, expressions can have two sorts of problems:
* a syntax error, meaning that they do not conform to the grammar of GREL. In this case, `ParsingException` is thrown
* an evaluation error: meaning that although they could be parsed successfully (without throwing a `ParsingException`), evaluating them returns an `EvalError`

When writing tests for this, it is important that this distinction is clear so that you can correctly assert whether a `ParsingException` is thrown at the parsing stage or an `EvalError` is returned at the evaluation stage.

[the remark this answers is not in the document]
sunilnatraj: @wetneb Thanks for the explanation; i missed the Evaluation error case. I have updated the function. Please review

wetneb (approved):

sunilnatraj:
@wetneb  All approvals are in place and tests have passed. Who will merge the PR

sunilnatraj:
@wetneb  All approvals are in place and tests have passed. Who will merge the PR

sunilnatraj:
Thank you @wetneb @ostephens

PR #6702, opened 2024-07-02
Ignore updates to impsort-maven-plugin

wetneb:
Since it requires higher Java versions.


tfmorris (changes requested):
I have a proposed alternative in #6704 which I think will fix this.

tfmorris (approved):
OK, I'm fine with freezing these updates.

PR #6704, opened 2024-07-02
Don't run impsort:sort by default during builds. Fixes #6703

### LINKED ISSUE #6703: Snapshot build shouldn't try to sort imports
Our current setup is to do code formatting in separate lint goal and then explicitly check that formatting as part of our CI builds, but we also have a default setup to sort imports as part of a build. This is unnecessary and the results of it are ignored, so we should remove it, particularly since it restricts the Java version we can use in the build environment.

### To Reproduce

See
PR updating impsort version https://github.com/OpenRefine/OpenRefine/pull/6670
failed build https://github.com/OpenRefine/OpenRefine/actions/runs/9460457438/job/26059310906

### Current Results
Java 11 build fails

### Expected Behavior
Java 11 build succeeds

### Additional context
We may want to think through whether there are any other implications to using different Java versions in our snapshot builds than we do in our CI.

tfmorris:
Fixes #6703 This is a proposed alternative to #6702.

- don't run `impsort:sort` by default during builds
- update impsort plugin to 1.11.0

We run `impsort:sort` explicitly with `./refine lint` and `impsort:check` as part of CI, so we don't need the default execution setup.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -282,7 +282,7 @@
       <plugin>
         <groupId>net.revelc.code</groupId>
         <artifactId>impsort-maven-plugin</artifactId>
-        <version>1.9.0</version>
+        <version>1.10.0</version>
         <configuration>
           <groups>java.,javax.,*,com.google.refine,org.openrefine</groups>
           <staticGroups>java,*</staticGroups>
--- a/pom.xml
+++ b/pom.xml
@@ -288,14 +288,6 @@
           <staticGroups>java,*</staticGroups>
           <removeUnused>true</removeUnused>
         </configuration>
-        <executions>
-          <execution>
-            <id>sort-imports</id>
-            <goals>
-              <goal>sort</goal><!-- runs at process-sources phase by default -->
-            </goals>
-          </execution>
-        </executions>
       </plugin>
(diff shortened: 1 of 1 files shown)


wetneb (changes requested):
The intention to have this plugin run during compilation was to lint code without requiring an explicit `./refine lint` from the developer, but I am not opposed to removing it as the execution of this plugin is actually quite slow and can increase build times significantly.

However, updating the plugin to the newest version would mean that we require Java 17 for development. I don't think this update justifies that. I think dropping support for Java versions should be motivated than more compelling reasons, such as an important update of a runtime dependency. This is just a linter so I would say there is even value in not updating it frequently, to avoid unnecessary churn.

tfmorris:
> The intention to have this plugin run during compilation was to lint code without requiring an explicit ./refine lint from the developer, but I am not opposed to removing it as the execution of this plugin is actually quite slow and can increase build times significantly.

Right now we run import sorting by default, but not formatting (`formatter:format`). We should probably do both or neither.

wetneb:
Yes, it would make sense to do both or neither. Happy either way.

PR #6707, opened 2024-07-05
Add escape() function options to escape URL path and fragment

### LINKED ISSUE #6139: Add escape() function options to escape URL path and fragment
The current `escape("url")` option is only appropriate for encoding the query string portion of a URL, but it would be useful for the user to be able to  correctly encode both the path and fragment portions of the URL, each of which has different encoding rules.

### Proposed solution
Add the following options to the `escape()` function:
* `"urlform"` which uses Guava's `UrlEscapers.urlFormParameterEscaper().escape(s)` (needs to be double checked, but is probably the same as the current `"url"` escaping)
* ~`"urlpath"`~ EDIT: `"urlpathsegment"` which uses Guava's `UrlEscapers.urlPathSegmentEscaper().escape(s)`
* `"urlfragment"` which uses Guava's `UrlEscapers.urlFragmentEscaper().escape(s)`

### Alternatives considered
* tell users to use `escape("url").replace("+","%20")` as a workaround for URL path escaping
* introduce a general `translate()` function which does character translation

### Additional context
If we correctly escape URLs by default for the "Add column by fetching URLs..." operation, the general need for this will be greatly reduced and users will only need these escapes if they're creating escaped URLs to use in a different context.

sunilnatraj:
Escape function updated to handle path and fragment element escaping

Fixes #6139 

- URL elements parsed and each element is separately handled using Apache UrlEscapers library
- If input is not a valid URL the input is returned

### DIFF
--- a/main/src/com/google/refine/expr/functions/strings/Escape.java
+++ b/main/src/com/google/refine/expr/functions/strings/Escape.java
@@ -33,8 +33,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package com.google.refine.expr.functions.strings;
 
-import java.io.UnsupportedEncodingException;
-import java.net.URLEncoder;
+import java.net.URL;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.MalformedURLException;
 import java.util.Properties;
 
 import org.apache.commons.text.StringEscapeUtils;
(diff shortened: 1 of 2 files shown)


tfmorris (commented):
Thanks for the contribution! The current implementation in this PR doesn't match the specification outlined in the related issue. I'm open to discussing a change to the specification, but let's do that on the issue so that it doesn't get lost. Specifically, it calls for three new escape types, but you've changed the implementation of the existing "url" escape type.

PR #6708, opened 2024-07-06
Fix issue #6381 a grel issue

### LINKED ISSUE #6381: Integer divide by zero returns null instead of error or infinity
Our divide by zero behavior is currently inconsistent and should be cleaned up

### Current Results
```
0 / 0 == NaN
0.0 / 0.0 = NaN
1 / 0 == null
1.1 / 0 == Infinity
-1.1 / 0 == -Infinity
```

### Expected Behavior
There are a couple of reasonable solutions (`null` isn't one of them):
1. return an error
2. return Infinity or -Infinity 

I'd argue for 2, since we're already returning NaN for 0/0, effectively promoting the integer operands to a floating point result, and this would be the equivalent.

### Screenshots

Screenshot 2024-02-21 at 1 14 25 PM

### Additional context

As an aside, when testing this I (re)discovered that there's no way to get a floating point 1.0. It will always get demoted back to an integer which is kind of taking the whole polymorphism thing to an extreme.

Kurocifer:
Fixes #6381 

- Add a return case for the division of integers, an if statement for when the divisor is zero, which will return "-Infinity" for a negative dividend and "Infinity" for a positive dividend.
-

### DIFF
--- a/main/src/com/google/refine/grel/ast/OperatorCallExpr.java
+++ b/main/src/com/google/refine/grel/ast/OperatorCallExpr.java
@@ -78,6 +78,8 @@ public class OperatorCallExpr implements Evaluable {
                     } else if ("/".equals(_op)) {
                         if (n2 == 0 && n1 == 0) {
                             return Double.NaN;
+                        } else if (n2 == 0) {
+                            return n1 > 0 ? Double.POSITIVE_INFINITY : Double.NEGATIVE_INFINITY;
                         }
                         // TODO: This will throw on divide by zero - return Double.Infinity / -Infinity instead?
                         return n1 / n2;


Kurocifer:
Hey @tfmorris I don't know if this is the kind of fix that was intended for this issue.

main/src/com/google/refine/grel/ast/OperatorCallExpr.java
```diff
@@ -78,6 +78,8 @@ public Object evaluate(Properties bindings) {
                     } else if ("/".equals(_op)) {
                         if (n2 == 0 && n1 == 0) {
                             return Double.NaN;
+                        } else if (n2 == 0) {
```
tfmorris, line 81: This could be simplified so that you're only testing for `n2 == 0` once and then handling all the `n1` cases inside that branch.

tfmorris (changes requested):
Thanks for the contribution! Can you please add a test to confirm that the behavior doesn't break in the future?

main/src/com/google/refine/grel/ast/OperatorCallExpr.java
Kurocifer: Okay let me fix that and add the test for the division.

Kurocifer:
Just made the changes you requested @tfmorris and added the test for division.

tfmorris (approved):
Apologies for the delay. I was away sailing last week.

This looks great. Thanks very much!

Kurocifer:
Pretty cool ⛴️. Thanks

PR #6724, opened 2024-07-17
Make sure that the file page exists before uploading SDC

sebastian-berlin-wmse:
Checks if the file page has been created after a file has been uploaded. If not it will wait for a bit and retry. This will continue until the page exists or the user interrupts, with the delay between checks increasing up to a maximum. At the moment the wait time starts at one second and increases by one second per wait. The max wait time is one minute.

Adds logging for:
* When a file will be uploaded in chunks
* When a chunk is uploaded
* When all chunks have been uploaded
* When the commit request has been made
* Waiting for file page to exist

### DIFF
--- a/extensions/wikibase/src/org/openrefine/wikibase/updates/MediaInfoEdit.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/updates/MediaInfoEdit.java
@@ -30,6 +30,7 @@ import org.wikidata.wdtk.wikibaseapi.WikibaseDataEditor;
 import org.wikidata.wdtk.wikibaseapi.apierrors.MediaWikiApiErrorException;
 
 import org.openrefine.wikibase.editing.MediaFileUtils;
+import org.openrefine.wikibase.editing.MediaFileUtils.MediaUploadResponse;
 import org.openrefine.wikibase.editing.NewEntityLibrary;
 import org.openrefine.wikibase.editing.ReconEntityRewriter;
 import org.openrefine.wikibase.schema.entityvalues.ReconEntityIdValue;
@@ -49,6 +50,8 @@ public class MediaInfoEdit extends LabeledStatementEntityEdit {
 
     static final Logger logger = LoggerFactory.getLogger(MediaInfoEdit.class);
     protected final boolean overrideWikitext;
+    protected int startWait;
+    protected int maxWait;
 
     /**
      * Constructor.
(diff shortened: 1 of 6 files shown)


extensions/wikibase/src/org/openrefine/wikibase/updates/MediaInfoEditBuilder.java
```diff
@@ -139,13 +143,27 @@ public MediaInfoEditBuilder setOverrideWikitext(boolean overrideWikitext) {
         return this;
     }
 
+    /**
+     * Sets the wait times when waiting for a page to exist after a file ha been uploaded
```
wetneb, line 147:
```suggestion
     * Sets the wait times when waiting for a page to exist after a file has been uploaded
```

extensions/wikibase/src/org/openrefine/wikibase/updates/MediaInfoEdit.java
```diff
@@ -247,7 +255,18 @@ public MediaInfoIdValue uploadNewFile(WikibaseDataEditor editor, MediaFileUtils
 
         response.checkForErrors();
 
+        List<String> filenames = new ArrayList<String>();
+        filenames.add(fileName);
```
wetneb, line 259: You could use `Collections.singletonList(fileName)` instead, as you're building a list that isn't intended to be modified and only ever contain a single element. Not a big deal, but makes the code a bit simpler, clearer and marginally more efficient.

extensions/wikibase/src/org/openrefine/wikibase/updates/MediaInfoEdit.java
```diff
@@ -49,6 +50,8 @@ public class MediaInfoEdit extends LabeledStatementEntityEdit {
 
     static final Logger logger = LoggerFactory.getLogger(MediaInfoEdit.class);
     protected final boolean overrideWikitext;
+    protected int startWait;
+    protected int maxWait;
```
wetneb, line 54: To me it feels unnatural to have those parameters be stored as fields of this class. The `MediaInfoEdit` class stores the representation of the edit as generated by the schema from project data. The "runtime" settings of how this edit is performed (credentials, maxlag parameter, editing tags, timeouts…) are stored in the `EditBatchProcessor` class, so I would rather put those new parameters there as they hold the same sort of data.

The parameters stored in `EditBatchProcessor` could (in the future) be exposed as a "editing settings" panel in the UI (currently, only maxlag is displayed, and probably too prominently). That would also let the user edit those waiting parameters in such a screen, which would feel consistent to me.

sebastian-berlin-wmse: Makes sense. I added the wait times to EditBatchProcessor. Not sure if they should also be defined there as I did now. Some variables come from constructor parameters, but these new variables are hard coded for now, so I don't know if that makes sense here.

extensions/wikibase/src/org/openrefine/wikibase/updates/MediaInfoEdit.java
sebastian-berlin-wmse: Done.

wetneb (approved):
It looks great like this, thanks a lot!

PR #6736, opened 2024-07-25
Improves NGramFingerprintKeyer and adds tests

thadguidry:
- change NGramFingerprintKeyer to use a character set (aka character class) in regex which handles the OR condition intrinsically.  As an example `[$\u20AC]` would match a dollar or euro sign.  See https://rules.sonarsource.com/javascript/RSPEC-6035/
- adds tests for NGramFingerprintKeyer


tfmorris:
Thanks for the contribution. The CI failure is because the new module needs formatting to pass our linting, but the current  NGramFingerprintKeyer tests are in [KeyerTests.java](https://github.com/OpenRefine/OpenRefine/blob/master/main/tests/server/src/com/google/refine/clustering/binning/KeyerTests.java), so we should either split that up or add your new tests there so that we don't have the tests in two different places.

tfmorris (changes requested):
@thadguidry Please revise per the above (sorry I didn't make it an official review)

thadguidry:
> @thadguidry Please revise per the above (sorry I didn't make it an official review)

You want me to move my tests into KeyerTests.java ?

tfmorris:
I don't have a strong opinion on which way to organize things, although perhaps @wetneb does.

Either:
1. Put your new tests in the existing KeyerTests.java and delete the new or NGramFingerprintKeyerTest.java, or
2. Extract the N-gram keyer tests from KeyerTests and put in your new module, then rename KeyerTests

Actually, now that I look at it, KeyerTests also has UserDefinedKeyer tests in it too, so I guess I'd recommend 1.

Then run `./refine lint` on everything so that it passes the CI formatting checks.

wetneb:
I think we should generally strive to have one test class corresponding to its tested class. So I think we should have:
* one `FingerprintKeyerTests`, which tests the behaviour of `FingerprintKeyer`
* one `UserDefinedKeyerTests`, which tests the behaviour of `UserDefinedKeyer`

and so on

tfmorris:
> I think we should generally strive to have one test class corresponding to its tested class. 

I'm fine with that. We should try to remember this during code reviews since the `UserDefinedKeyer` tests were just added in June.

@thadguidry  let's go with option 3:

3. Extract the N-gram fingerprint keyer tests from `KeyerTests` and add them to `NGramFingerprintKeyerTest` and run `/refine lint` before committing.

We can do the rest of the `KeyerTests` refactoring separately unless you're dying to do it.

tfmorris (approved):
Looks great! Thanks.

PR #6739, opened 2024-07-29
GREL Zip function

### LINKED ISSUE #5440: Ability to zip columns together when joining
It would be easier if OpenRefine had a zip-like function when joining columns with multi-valued cells. 
Imagine the following table :
| Col1 | Col2 | Col3 |
|--|--|--|
| 1,2,3 | A,B,C | 1A,2B,3C |

Col3 is the desired outcome as then I can split the cell on `,` to achieve my final wish:

| Col1 |
|--|
| 1A |
| 2B |
| 3C |

The idea is to have this in the UI in the join columns menu as I have encountered this issue in several datasets

### Proposed solution

To achieve this I made a GREL expression:
```
forEachIndex(
	cells["Col 1"].value.split(","),
	i,
	v,
	v + cells["Col 2"].value.split(",").get(i)
).join(",")
```
However, this is error-prone and I want the python-like behavior where the zipping stops when one of the arrays runs out of elements. If I do a check things become much more ugly and increasing the amount of columns I want to zip becomes very complex.

```
if(
	length(cells["Col 1"].value.split(","))>length(cells["Col 2"].value.split(",")),
	forEachIndex(
		cells["Col 1"].value.split(",").slice(0,length(cells["Col 2"].value.split(","))),
		i,
		v,
		v + cells["Col 2"].value.split(",").get(i)
	).join(","),
	forEachIndex(
		cells["Col 1"].value.split(","),
		i,
		v,
		v + cells["Col 2"].value.split(",").get(i)
	).join(",")
)
```
### Alternatives considered

I assume that using jython for the purpose would be much easier as python has a zip function but it would be nice to have it in the UI.

sunilnatraj:
Zip function addition to GREL

Fixes #5440 

- Zip function takes in a 2 or more Array parameters and generates output which is an Array of arrays 
- The length of the output will be that of the shortest input parameter
- Supports text, number, date
- Usage - zip(array 1, array 2, ...)
- Examples
1. zip( [1, 2], ["A", "B"] ) produces [ [ 1, "A" ], [ 2, "B" ] ]
2. zip( [1, 2], ["A", "B"] , [ now(), now() ] ) produces  [ 1, "A", "2024-08-01T09:24:02.063008Z" ], [ 2, "B", "2024-08-01T09:24:02.063017Z" ] ]

### DIFF
--- a/main/resources/com/google/refine/grel/FunctionDescription.properties
+++ b/main/resources/com/google/refine/grel/FunctionDescription.properties
@@ -6,6 +6,7 @@ arr_join=Joins the items in the array with sep, and returns it all as a string.
 arr_reverse=Reverses array a.
 arr_sort=Sorts the array in ascending order. Sorting is case-sensitive, uppercase first and lowercase second.
 arr_uniques=Returns the array with duplicates removed. Case-sensitive.
+arr_zip=Combine two or more arrays into a single array of arrays, where the elements at the same index from the input arrays are paired together to form a sub-array in the output array.
 
 # 2. Booleans
 bool_and=Uses the logical operator AND on two or more booleans to output a boolean. Evaluates multiple statements into booleans, then returns true if all the statements are true. For example, (1 < 3).and(1 < 0) returns false because one condition is true and one is false.
--- a/main/resources/com/google/refine/grel/FunctionDescription.properties
+++ b/main/resources/com/google/refine/grel/FunctionDescription.properties
@@ -6,7 +6,7 @@ arr_join=Joins the items in the array with sep, and returns it all as a string.
 arr_reverse=Reverses array a.
 arr_sort=Sorts the array in ascending order. Sorting is case-sensitive, uppercase first and lowercase second.
 arr_uniques=Returns the array with duplicates removed. Case-sensitive.
-arr_zip=Combine two or more arrays into a single array of arrays, where the elements at the same index from the input arrays are paired together to form a sub-array in the output array.
+arr_zip=Combines two or more arrays into a single array of arrays, where the elements at the same index from the input arrays are grouped together to form a sub-array in the output array.
 
 # 2. Booleans
 bool_and=Uses the logical operator AND on two or more booleans to output a boolean. Evaluates multiple statements into booleans, then returns true if all the statements are true. For example, (1 < 3).and(1 < 0) returns false because one condition is true and one is false.
--- a/main/src/com/google/refine/expr/functions/arrays/Zip.java
+++ b/main/src/com/google/refine/expr/functions/arrays/Zip.java
@@ -61,14 +61,7 @@ public class Zip implements Function {
                 } else if (v.getClass().isArray()) {
                     iterators.add(Arrays.stream((Object[]) v).spliterator());
                 } else if (v instanceof ArrayNode) {
-                    ArrayNode a = (ArrayNode) v;
-                    List<Object> ziparg = new ArrayList<>();
-                    if (a.isArray()) {
-                        for (JsonNode objNode : a) {
-                            ziparg.add(objNode);
-                        }
-                    }
-                    iterators.add(ziparg.stream().spliterator());
+                    iterators.add(((ArrayNode) v).spliterator());
                 } else {
                     iterators.add(((List<Object>) v).spliterator());
                 }
--- a/main/src/com/google/refine/expr/functions/arrays/Zip.java
+++ b/main/src/com/google/refine/expr/functions/arrays/Zip.java
@@ -1,0 +1,153 @@
+/*
+
+Copyright 2010, Google Inc.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+    * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+    * Neither the name of Google Inc. nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+*/
+
+package com.google.refine.expr.functions.arrays;
+
+import static java.util.stream.Collectors.toCollection;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Properties;
+import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.google.common.collect.Streams;
+
+import com.google.refine.expr.EvalError;
+import com.google.refine.expr.ExpressionUtils;
+import com.google.refine.grel.ControlFunctionRegistry;
+import com.google.refine.grel.EvalErrorMessage;
+import com.google.refine.grel.Function;
+import com.google.refine.grel.FunctionDescription;
+
+public class Zip implements Function {
+
+    @Override
+    public Object call(Properties bindings, Object[] args) {
+        if (args.length == 1) {
+            Object[] v = (Object[]) args[0];
+            List<ArrayList> output = new ArrayList<>();
+            ArrayList arg1 = new ArrayList<>();
+            ArrayList arg2 = new ArrayList<>();
+
+            boolean isFirst = true;
+            if (v != null) {
+                for (Object o : v) {
+                    if (o.getClass().isArray() || o instanceof List<?> || o instanceof ArrayNode) {
+                        if (o.getClass().isArray()) {
+                            arg2.addAll(Arrays.asList((Object[]) o));
+                        } else if (o instanceof ArrayNode) {
+                            ArrayNode a = (ArrayNode) o;
+                            if (a.isArray()) {
+                                for (JsonNode objNode : a) {
+                                    arg2.add(objNode);
+                                }
+                            }
+                        } else {
+                            for (Object z : ExpressionUtils.toObjectList(o)) {
+                                if (z != null) {
+                                    arg2.add(z.toString());
+                                } else {
+                                    arg2.add(z);
+                                }
+                            }
+                        }
+                        if (isFirst) {
+                            for (Object x : arg2) {
+                                arg1 = new ArrayList() {
+
+                                    {
+                                        add(x);
+                                    }
+                                };
+                                output.add(arg1);
+                            }
+                        } else {
+                            output = (List<ArrayList>) Streams.zip(output.stream(),
+                                    arg2.stream(),
+                                    (a, b) -> collectElements(a, b))
+                                    .collect(Collectors.toList());
+                        }
+                        isFirst = false;
+                    }
+                    arg2.clear();
+                }
+            }
+            return output;
+        }
+        return new EvalError(EvalErrorMessage.expects_one_array(ControlFunctionRegistry.getFunctionName(this)));
+    }
+
+    @Override
+    public String getDescription() {
+        return FunctionDescription.arr_join();
+    }
+
+    @Override
+    public String getParams() {
+        return "array a, string sep";
+    }
+
+    @Override
+    public String getReturns() {
+        return "string";
+    }
+
+    private static ArrayList collectElements(Object a, Object b) {
+        ArrayList returnList = new ArrayList<>();
+
+        if (a == null) {
+            returnList.add(a);
+        } else if (a instanceof ArrayList) {
+            ArrayList x = (ArrayList) a;
+            returnList = (ArrayList) x.stream().collect(toCollection(ArrayList::new));
+        } else {
+            returnList.add(a.toString());
+        }
+
+        if (b == null) {
+            returnList.add(b);
+        } else if (b instanceof ArrayList) {
+            ArrayList y = (ArrayList) b;
+            int i;
+            for (i = 0; i < y.size(); i++) {
+                returnList.add(y.get(i).toString());
+            }
+        } else {
+            returnList.add(b.toString());
+        }
+        return returnList;
+    }
+}
--- a/main/src/com/google/refine/expr/functions/arrays/Zip.java
+++ b/main/src/com/google/refine/expr/functions/arrays/Zip.java
@@ -84,12 +84,13 @@ public class Zip implements Function {
             while (!done)  {
                 List<Object> currentElements = new ArrayList<>();
                 for (Spliterator<?> iterator : iterators) {
-                    Object[] elementHolder = new Object[1];
-                    if (iterator.tryAdvance(e -> elementHolder[0] = e)) {
-                        currentElements.add(elementHolder[0]);
+                    if (!iterator.tryAdvance(e -> currentElements.add(e)) {
+                        done = true;
                     }
                 }
-                output.add(new ArrayList<>(currentElements));
+                if (!done) {
+                    output.add(currentElements);
+                }
             }
             return output;
         }
--- a/main/src/com/google/refine/expr/functions/arrays/Zip.java
+++ b/main/src/com/google/refine/expr/functions/arrays/Zip.java
@@ -57,59 +57,45 @@ public class Zip implements Function {
     @Override
     public Object call(Properties bindings, Object[] args) {
         if (args.length >= 2) {
-            List<List> output = new ArrayList<>();
-            List ziparg1 = new ArrayList<>();
-            List ziparg2 = new ArrayList<>();
-            boolean isFirst = true;
+            List<List> input = new ArrayList<>();
+            List<List> output;
 
-            for (int i = 0; i < args.length; i++) {
-                Object v = args[i];
-                if (v == null) {
-                    return new EvalError(
-                            EvalErrorMessage.expects_at_least_two_or_more_array_args(ControlFunctionRegistry.getFunctionName(this)));
-                }
-
-                if (!(v.getClass().isArray() || v instanceof List<?> || v instanceof ArrayNode)) {
+            for (Object v : args) {
+                List ziparg = new ArrayList<>();
+                if (v == null || !(v.getClass().isArray() || v instanceof List<?> || v instanceof ArrayNode)) {
                     return new EvalError(
                             EvalErrorMessage.expects_at_least_two_or_more_array_args(ControlFunctionRegistry.getFunctionName(this)));
                 }
 
                 if (v.getClass().isArray()) {
-                    ziparg2.addAll(Arrays.asList((Object[]) v));
+                    ziparg.addAll(Arrays.asList((Object[]) v));
                 } else if (v instanceof ArrayNode) {
                     ArrayNode a = (ArrayNode) v;
                     if (a.isArray()) {
                         for (JsonNode objNode : a) {
-                            ziparg2.add(objNode);
+                            ziparg.add(objNode);
                         }
                     }
                 } else {
                     for (Object z : ExpressionUtils.toObjectList(v)) {
-                        if (z != null) {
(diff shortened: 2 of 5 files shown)


sunilnatraj:
@wetneb @ostephens @jquartel Please review the PR for zip function

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,153 @@
...
+                }
+            }
+            return output;
+        }
+        return new EvalError(EvalErrorMessage.expects_one_array(ControlFunctionRegistry.getFunctionName(this)));
+    }
+
+    @Override
+    public String getDescription() {
+        return FunctionDescription.arr_join();
```
tfmorris, line 115: This looks like it's going to pull in the wrong description

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @Override
+    public String getParams() {
+        return "array a, string sep";
```
tfmorris, line 120: This parameter description appear left over from the Join function and needs to be updated.

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @Override
+    public String getReturns() {
+        return "string";
```
tfmorris, line 125: This return also needs to be updated to match the new behavior.

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    private static ArrayList collectElements(Object a, Object b) {
+        ArrayList returnList = new ArrayList<>();
```
tfmorris, line 129: It's good practice to use interfaces rather than concrete classes where possible.

```suggestion
    private static List collectElements(Object a, Object b) {
        List returnList = new ArrayList<>();
```

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,153 @@
...
+                                    .collect(Collectors.toList());
+                        }
+                        isFirst = false;
+                    }
+                    arg2.clear();
+                }
+            }
+            return output;
+        }
+        return new EvalError(EvalErrorMessage.expects_one_array(ControlFunctionRegistry.getFunctionName(this)));
```
tfmorris, line 110: This error message doesn't look correct since the function was spec'd to accept multiple arrays.

main/tests/server/src/com/google/refine/expr/functions/arrays/ZipTests.java
```diff
@@ -0,0 +1,48 @@
...
+import org.testng.annotations.Test;
+
+import com.google.refine.expr.ParsingException;
+import com.google.refine.grel.GrelTestBase;
+
+public class ZipTests extends GrelTestBase {
+
+    @Test
+    public void zipArray() throws ParsingException {
+        String[] test = { "[ [1,2,3], ['A','B','C'] ].zip()", "[[1, A], [2, B], [3, C]]" };
```
tfmorris, line 13: The tests need to be updated to the following form which matches the original specification
```suggestion
        String[] test = { "zip([1,2,3],['A','B','C'])", "[[1, A], [2, B], [3, C]]" };
```

tfmorris (changes requested):
Thanks for the contribution! I didn't realize it until I was well into the review, so some of the earlier comments may not make sense, but the implementation doesn't match the specification.

What you wrote:

> Zip function takes in a 2 or more Array parameters and generates output which is an Array of arrays

is correct, but then the usage example later shows just a single array parameter:

> `[a, b, c, d]`.zip()

when I would expect `zip(a, b, c, d)` (or `a.zip(b, c, d)` in the postfix notation version).

main/src/com/google/refine/expr/functions/arrays/Zip.java
sunilnatraj: Updated description

main/src/com/google/refine/expr/functions/arrays/Zip.java
sunilnatraj: Updated param description

main/src/com/google/refine/expr/functions/arrays/Zip.java
sunilnatraj: Updated return description

main/tests/server/src/com/google/refine/expr/functions/arrays/ZipTests.java
sunilnatraj: Updated all test cases

main/src/com/google/refine/expr/functions/arrays/Zip.java
sunilnatraj: Updated error message and handled all scenarios

sunilnatraj:
@tfmorris Thank you for reviewing the code and feedback. I have addressed all the points. 

1. zip( [1, 2], ["A", "B"] ) produces [ [ 1, "A" ], [ 2, "B" ] ]
2. zip( [1, 2], ["A", "B"] , [ now(), now() ] ) produces [ 1, "A", "2024-08-01T09:24:02.063008Z" ], [ 2, "B", "2024-08-01T09:24:02.063017Z" ] ]

ostephens:
I don't know the answer to this, but just asking the question... is it correct that this should be implemented as a function, or should it be a control? Most code we have that iterate over an array are done as controls, although I've never really fully understood the reasons. I asked this question a long time ago on the (then) forum: https://groups.google.com/g/openrefine-dev/c/QvMSiHu-8u0/m/HcZt7PioAQAJ

@tfmorris can you comment?

tfmorris:
@sunilnatraj Thanks for the update. I'll have another look.

@ostephens David's comment on the Control interface says

> * Interface of GREL controls such as if, forEach, forNonBlank, with. A control can decide which part of the code to
> * execute and can affect the environment bindings. Functions, on the other hand, can't do either.

which I think summarizes the main difference. For a function, all the arguments get evaluated and if they generate errors, those will propagate, preventing the execution of the function (which wouldn't work for controls like `ifError`).

Since this is a straightforward transformation without any conditional execution or evaluation, I think implementing it as a Function instead of a Control is the right thing.

main/tests/server/src/com/google/refine/expr/functions/arrays/ZipTests.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.google.refine.expr.ParsingException;
+import com.google.refine.grel.GrelTestBase;
+
+public class ZipTests extends GrelTestBase {
+
+    @Test
+    public void zipArray() throws ParsingException {
+        String[] test = { "zip([1,2,3], ['A','B','C'])", "[[1, A], [2, B], [3, C]]" };
+        parseEval(bindings, test);
```
tfmorris, line 14: `parseEval` uses `toString` internally, so I don't really trust it for testing this kind of operation. I think it would be better to have your own help which knows how to compare two arrays of arrays to make sure that you're getting the results that you expect.

tfmorris (changes requested):
I'll do some more testing and review, but one thing I noticed with my interactive testing is that `zip([1,2,3],[4,5,6])` results in 
    `[ [ 1, "4" ], [ 2, "5" ], [ 3, "6" ] ]` 
instead of the correct 
    `[ [ 1, 4 ], [ 2, 5 ], [ 3, 6 ] ]`

This is being masked in the tests because of the operation of the `parseEval` helper.

I think you need a helper which knows how to compare the actual values, not their string equivalents.

main/tests/server/src/com/google/refine/expr/functions/arrays/ZipTests.java
sunilnatraj: @tfmorris I have fixed the incorrect type conversion and validated it via the application. 

I will check on updating the unit test to do array comparison.

sunilnatraj: @tfmorris Added test cases with support to compare the output array against expected result.

ostephens:
Looks like this is getting close to being merged.  Note that the official documentation will need updating with an explanation of this function - if this PR gets merged please create an Issue in https://github.com/OpenRefine/openrefine.org to add the documentation

tfmorris (changes requested):
This is greatly improved and produces the expected results. Thanks for all your work.

The main thing left to improve is to try and simplify the code as much as possible. I left some specific suggestions for various pieces, but I think the main body of the function needs to be restructured. 

I would suggest breaking the processing into two phases:
1. Validate each argument and collect some type of Stream/iterator/spliterator for it (I used spliterators for my experiments)
2. Generate output tuples (lists) from the next elements of each input until one of the inputs is empty

I think this will allow simple direct construction of the output without touching the inputs multiple times and will be easier to understand.

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,159 @@
...
+                Object v = args[i];
+                if (v == null) {
+                    return new EvalError(
+                            EvalErrorMessage.expects_at_least_two_or_more_array_args(ControlFunctionRegistry.getFunctionName(this)));
+                }
+
+                if (!(v.getClass().isArray() || v instanceof List<?> || v instanceof ArrayNode)) {
+                    return new EvalError(
+                            EvalErrorMessage.expects_at_least_two_or_more_array_args(ControlFunctionRegistry.getFunctionName(this)));
+                }
```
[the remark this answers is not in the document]
sunilnatraj: merged the 2 checks

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,159 @@
...
+                } else {
+                    output = (List<List>) Streams.zip(output.stream(),
+                            ziparg2.stream(),
+                            (a, b) -> collectElements(a, b))
+                            .collect(Collectors.toList());
+                }
+                isFirst = false;
+                ziparg2.clear();
+            }
+            return output;
```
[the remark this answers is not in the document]
sunilnatraj: @tfmorris I have refactored the flow  - process the input arguments and then generate output
I did not replace the zip function from google streams as internally it is also using streams to merge the 2 inputs.  Please review the updated flow and if you believe that the use to google streams zip needs to be avoided.

sunilnatraj:
@tfmorris Thank you the review and feedback. 

I have refactored the flow - process the input arguments and then generate output
I did not replace the zip function from google streams as internally it is also using streams to merge the 2 inputs. Please review the updated flow and if you believe that the use to google streams zip needs to be avoided.

sunilnatraj:
@tfmorris Hi Can you review the PR when you get a chance

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,134 @@
...
+
+            if (input.size() > 2) {
+                for (int i = 2; i < input.size(); i++) {
+                    output = (List<List>) Streams.zip(output.stream(),
+                            input.get(i).stream(),
+                            (a, b) -> collectElements(a, b))
+                            .collect(Collectors.toList());
+                }
+            }
+            return output;
```
tfmorris, line 100: Then you can change the order of iteration so that you first iterate by element and then by input argument similar to the following:

```suggestion
            List<List> result = new ArrayList<>();
            boolean done = false;
            while (!done) {
                List element = new ArrayList<>();
                
                // Construct an output element by getting 
                // input element from each iterator
                // if any iterator is empty, we are done

                if (!done) {
                    result.add(element);
                }
            }
           return result;
```

Does that make sense? Let me know if you have any questions.

main/resources/com/google/refine/grel/FunctionDescription.properties
```diff
@@ -6,6 +6,7 @@ arr_join=Joins the items in the array with sep, and returns it all as a string.
 arr_reverse=Reverses array a.
 arr_sort=Sorts the array in ascending order. Sorting is case-sensitive, uppercase first and lowercase second.
 arr_uniques=Returns the array with duplicates removed. Case-sensitive.
+arr_zip=Combine two or more arrays into a single array of arrays, where the elements at the same index from the input arrays are paired together to form a sub-array in the output array.
```
tfmorris, line 9:
```suggestion
arr_zip=Combines two or more arrays into a single array of arrays, where the elements at the same index from the input arrays are grouped together to form a sub-array in the output array.
```

tfmorris (changes requested):
Thanks. That's definitely improved, but I think we can simplify it even more to get rid of the N>2 vs N=2 distinction so that all inputs are processed uniformly. See my suggestions inline and let me know if you have any questions.

Thanks for the sticking with this and apologies for the delay in reviewing your updates.

main/src/com/google/refine/expr/functions/arrays/Zip.java
sunilnatraj: @tfmorris Thanks for the feedback. Can you also confirm if the output of the zip is not limited to the shortest input. 
Is the below acceptable output of zip function?

-  zip([1,2], ['A','B','C'], ['X','Y','Z']) -> [[[1, A, X], [2, B, Y], [null, C, Z]]]

tfmorris: Is there a reason that you'd prefer your alternative? If there's a user needs based reason, we should discuss it on the original issue where folks will see the discussion, but if it's just an implementation based preference, I think it would be better to make the implementation conform to the specification as currently given, i.e. the output for your example should be

`zip([1,2], ['A','B','C'], ['X','Y','Z']) -> [[[1, A, X], [2, B, Y]]]`

sunilnatraj: @tfmorris The query came up as part of implementation. No change required in the zip working definition

sunilnatraj:
@tfmorris Thanks for your inputs, I have removed the Streams zip dependency and updated the working as per your feedback. Please review the PR.

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,115 @@
...
+            int minSize = Integer.MAX_VALUE;
+            for (Spliterator<?> iterator : iterators) {
+                minSize = Math.min(minSize, (int) iterator.estimateSize());
+            }
+
+            List<List> output = new ArrayList<>();
+            List<Object> currentElements = new ArrayList<>(minSize);
+
+            for (int i = 0; i < minSize; i++) {
+                currentElements.clear();
```
tfmorris, line 86:
```suggestion
            boolean done = false;
            while (!done)  {
                List<Object> currentElements = new ArrayList<>();
```

I think it's simpler to just create your output list inside the loop. You are going to use it every time except the last iteration, so there's really no appreciable extra object creation overhead.

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,115 @@
...
+                } else {
+                    iterators.add(((List<Object>) v).spliterator());
+                }
+            }
+
+            int minSize = Integer.MAX_VALUE;
+            for (Spliterator<?> iterator : iterators) {
+                minSize = Math.min(minSize, (int) iterator.estimateSize());
+            }
+
```
tfmorris, line 81:
```suggestion
```

The fact that the method is called `estimateSize()` is a red flag that it probably shouldn't be relied upon for correct operation of the algorithm.

I would suggest iterating until one of the iterators is empty.

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,115 @@
...
+                    iterators.add(Arrays.stream((Object[]) v).spliterator());
+                } else if (v instanceof ArrayNode) {
+                    ArrayNode a = (ArrayNode) v;
+                    List<Object> ziparg = new ArrayList<>();
+                    if (a.isArray()) {
+                        for (JsonNode objNode : a) {
+                            ziparg.add(objNode);
+                        }
+                    }
+                    iterators.add(ziparg.stream().spliterator());
```
tfmorris, line 71:
```suggestion
                    iterators.add(((ArrayNode) v).spliterator());
```

This can be simplified by using the `spliterator` method of ArrayNode.

main/src/com/google/refine/expr/functions/arrays/Zip.java
```diff
@@ -0,0 +1,115 @@
...
+
+            for (int i = 0; i < minSize; i++) {
+                currentElements.clear();
+                for (Spliterator<?> iterator : iterators) {
+                    Object[] elementHolder = new Object[1];
+                    if (iterator.tryAdvance(e -> elementHolder[0] = e)) {
+                        currentElements.add(elementHolder[0]);
+                    }
+                }
+                output.add(new ArrayList<>(currentElements));
```
tfmorris, line 93:
```suggestion
                    if (!iterator.tryAdvance(e -> currentElements.add(e)) {
                        done = true;
                    }
                }
                if (!done) {
                    output.add(currentElements);
                }
```

You can add the element to your output list directly inside the tryAdvance method and by inverting the if test, you can set a `done` flag when one of your iterators is empty and it's time to leave.

tfmorris (changes requested):
This is looking really good! I think it's almost ready to merge. Hopefully all my individual comments/suggestions make sense when all stitched together, but basically:

- don't depend on Spliterator.estimateSize. I don't think it's guaranteed to give you a true size (thus the "estimate" in the name).
- a few simplifications to small sections of the code.

I think it'll be ready to merge after this round. Thanks for all your work!

main/src/com/google/refine/expr/functions/arrays/Zip.java
sunilnatraj: Removed the size estimation logic

main/src/com/google/refine/expr/functions/arrays/Zip.java
sunilnatraj: I added the loop as the iterator fails when processing Array node.

sunilnatraj:
@tfmorris Thank you for your inputs; I have further simplified the logic. I hope it is now good to be committed

main/src/com/google/refine/expr/functions/arrays/Zip.java
tfmorris: It doesn't fail for me using your tests. Do you have a separate test which demonstrates the failure? Could you add it to the test suite please so that someone doesn't try to optimize this in the future. Thanks!

main/tests/server/src/com/google/refine/expr/functions/arrays/ZipTests.java
```diff
@@ -0,0 +1,109 @@
...
+        List<List> expected = List.of(
+                Lists.newArrayList(1, "A", "Ben"),
+                Lists.newArrayList(null, "B", null),
+                Lists.newArrayList(3.142, "C", "Hen"));
+
+        assertEquals(invoke("zip", arg1, arg2, arg3), expected);
+    }
+
+    @Test
+    public void testZipforParams() {
```
tfmorris, line 95:
```suggestion
    public void testZipForParams() {
```

main/tests/server/src/com/google/refine/expr/functions/arrays/ZipTests.java
```diff
@@ -0,0 +1,109 @@
...
+        List<List> expected = List.of(
+                List.of(1, 7.89),
+                List.of(2, 8.90),
+                List.of(3, 9.01));
+
+        assertEquals(invoke("zip", arg1, arg2), expected);
+    }
+
+    @Test
+    public void testZipwithStrArrays() {
```
tfmorris, line 68:
```suggestion
    public void testZipWithStrArrays() {
```

main/tests/server/src/com/google/refine/expr/functions/arrays/ZipTests.java
```diff
@@ -0,0 +1,109 @@
...
+        String[] test8 = {
+                "zip([\"Bob has a cat & dog\", \"Cat's name is Dotty\", \"Dog's name is Dots\"], [11, 22, 33, 44, 55], '[\"Doe\", \"Foe\", 456, 789]'.parseJson() )",
+                "[[Bob has a cat & dog, 11, \"Doe\"], [Cat's name is Dotty, 22, \"Foe\"], [Dog's name is Dots, 33, 456]]" };
+        parseEval(bindings, test8);
+
+        // TODO: Add tests for List<Object> returned from ExpressionUtils.toObjectList()
+    }
+
+    @Test
+    public void testZipwithIntArrays() {
```
tfmorris, line 55:
```suggestion
    public void testZipWithIntArrays() {
```

main/tests/server/src/com/google/refine/expr/functions/arrays/ZipTests.java
```diff
@@ -0,0 +1,109 @@
...
+    @Test
+    public void testZipforParams() {
+        List arg1 = List.of(1, 2, 3);
+        List arg2 = List.of(7.89, 8.90, 9.01);
+
+        List<List> expected = List.of(
+                List.of(1, 7.89),
+                List.of(2, 8.90),
+                List.of(3, 9.01));
+
```
tfmorris, line 103:
```suggestion
```

This is unused

tfmorris (commented):
Looks great! It's ready for merge except for the question I have below on the rejected suggestion for the `ArrayNode.spliterator`. I can't get it to fail for me, so a test would be good to make sure someone doesn't shoot themselves in the foot in the future.

Thanks for sticking with it and sorry the iterations have taken so long with me being away.

sunilnatraj:
@tfmorris The unit test that was failing, it was case for testing JSON. I updated the code and re-ran all the tests and no failure noted. Thank you for all your guidance on this.


`// Tests for JSON array
        String[] test8 = {
                "zip([\"Bob has a cat & dog\", \"Cat's name is Dotty\", \"Dog's name is Dots\"], [11, 22, 33, 44, 55], '[\"Doe\", \"Foe\", 456, 789]'.parseJson() )",
                "[[Bob has a cat & dog, 11, \"Doe\"], [Cat's name is Dotty, 22, \"Foe\"], [Dog's name is Dots, 33, 456]]" };
        parseEval(bindings, test8);`

tfmorris (approved):
Looks great. Thank you! And congratulations on another PR approved.

sunilnatraj:
Thanks @tfmorris

PR #6740, opened 2024-07-29
Extract column dependencies of GREL expressions

wetneb:
This adds support for extracting the columns a GREL expression depends on.

This is a building block towards the support for analyzing the column dependencies of an operation, for instance to check that those columns exist in the project before applying the operation, or to visualize the dependencies of operations in a graph-based view.

As many static analysis tasks, the full problem is undecidable so this is meant to provide an approximation of it which is safe, meaning that if we succeed in extracting a set of columns for an expression, the expression is guaranteed not to rely on other columns. When the heuristics fail to return a set of columns, the expression may rely on any column.

### DIFF
--- a/modules/core/src/main/java/com/google/refine/expr/Evaluable.java
+++ b/modules/core/src/main/java/com/google/refine/expr/Evaluable.java
@@ -33,7 +33,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package com.google.refine.expr;
 
+import java.util.Optional;
 import java.util.Properties;
+import java.util.Set;
 
 /**
  * Interface for evaluable expressions in any arbitrary language.
@@ -47,4 +49,19 @@ public interface Evaluable {
      * @return
      */
     public Object evaluate(Properties bindings);
+
+    /**
+     * Returns the names of the columns this expression depends on.
+     *
+     * @param baseColumn
+     *            the name of the column this expression is based on (none if the expression is not evaluated on a
+     *            particular column)
+     * @return none if the columns could not be isolated: in this case, the expression might depend on all columns in
+     *         the project. Note that this is different from returning an empty set, which means that the expression is
+     *         constant.
+     */
+    public default Optional<Set<String>> getColumnDependencies(Optional<String> baseColumn) {
+        return Optional.empty();
+    }
+
 }
--- a/modules/core/src/main/java/com/google/refine/expr/Evaluable.java
+++ b/modules/core/src/main/java/com/google/refine/expr/Evaluable.java
@@ -51,14 +51,16 @@ public interface Evaluable {
     public Object evaluate(Properties bindings);
 
     /**
-     * Returns the names of the columns this expression depends on.
+     * Returns an approximation of the names of the columns this expression depends on. This approximation is designed
+     * to be safe: if a set of column names is returned, then the expression does not read any other column than the
+     * ones mentioned, regardless of the data it is executed on.
      *
      * @param baseColumn
      *            the name of the column this expression is based on (none if the expression is not evaluated on a
      *            particular column)
-     * @return none if the columns could not be isolated: in this case, the expression might depend on all columns in
-     *         the project. Note that this is different from returning an empty set, which means that the expression is
-     *         constant.
+     * @return {@link Optional#empty()} if the columns could not be isolated: in this case, the expression might depend
+     *         on all columns in the project. Note that this is different from returning an empty set, which means that
+     *         the expression is constant.
      */
     public default Optional<Set<String>> getColumnDependencies(Optional<String> baseColumn) {
         return Optional.empty();
--- a/modules/grel/pom.xml
+++ b/modules/grel/pom.xml
@@ -139,6 +139,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>${mockito.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.openrefine</groupId>
       <artifactId>core</artifactId>
       <version>${project.version}</version>
--- a/modules/grel/src/main/java/com/google/refine/grel/ast/ControlCallExpr.java
+++ b/modules/grel/src/main/java/com/google/refine/grel/ast/ControlCallExpr.java
@@ -33,7 +33,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package com.google.refine.grel.ast;
 
+import java.util.HashSet;
+import java.util.Optional;
 import java.util.Properties;
+import java.util.Set;
 
 import com.google.refine.expr.EvalError;
 import com.google.refine.expr.Evaluable;
@@ -59,6 +62,19 @@ public class ControlCallExpr implements Evaluable {
         } catch (Exception e) {
             return new EvalError(e.toString());
         }
+    }
+
+    @Override
+    public final Optional<Set<String>> getColumnDependencies(Optional<String> baseColumn) {
+        Set<String> dependencies = new HashSet<>();
+        for (Evaluable ev : _args) {
+            Optional<Set<String>> deps = ev.getColumnDependencies(baseColumn);
+            if (deps.isEmpty()) {
+                return Optional.empty();
+            }
+            dependencies.addAll(deps.get());
+        }
+        return Optional.of(dependencies);
     }
 
     @Override
--- a/modules/grel/src/main/java/com/google/refine/grel/ast/FieldAccessorExpr.java
+++ b/modules/grel/src/main/java/com/google/refine/grel/ast/FieldAccessorExpr.java
@@ -33,7 +33,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package com.google.refine.grel.ast;
 
+import java.util.Collections;
+import java.util.Optional;
 import java.util.Properties;
+import java.util.Set;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
@@ -71,6 +74,22 @@ public class FieldAccessorExpr implements Evaluable {
             return JsonValueConverter.convert(value);
         } else {
             return null;
+        }
+    }
+
+    @Override
+    public Optional<Set<String>> getColumnDependencies(Optional<String> baseColumn) {
+        Optional<Set<String>> innerDeps = _inner.getColumnDependencies(baseColumn);
+        if (innerDeps.isPresent()) {
+            return innerDeps;
+        } else {
+            String innerStr = _inner.toString();
+            if ("cells".equals(innerStr) || "row.cells".equals(innerStr)) {
+                return Optional.of(Collections.singleton(_fieldName));
+            }
+            // TODO add support for starred, flagged, rowIndex, which are not real columns
+            // but whose dependency could also be analyzed.
+            return Optional.empty();
         }
     }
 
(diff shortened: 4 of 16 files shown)


main/src/com/google/refine/grel/ast/FunctionCallExpr.java
```diff
@@ -72,6 +77,27 @@ public Object evaluate(Properties bindings) {
...
+    @Override
+    public final Optional<Set<String>> getColumnDependencies(Optional<String> baseColumn) {
+        // special case to handle "get(cells, "foo")" which only depends on the "foo" column
+        // even though the cells variable has a greater reach
+        if (_function instanceof Get && _args.length == 2 && (new VariableExpr("cells")).equals(_args[0]) &&
+                _args[1] != null && _args[1] instanceof LiteralExpr) {
+            String columnName = ((LiteralExpr) _args[1])._value.toString();
+            return Optional.of(Collections.singleton(columnName));
+        }
+        // TODO distinguish functions which are "pure" and those which access external data, like cross or facetCount
```
wetneb, line 89: This will be handled in a follow-up PR.

main/tests/server/src/com/google/refine/grel/GrelTests.java
```diff
@@ -226,6 +230,37 @@ public void testCrossFunctionEval() throws ParsingException {
...
 
+    @Test
+    public void testColumnDependencies() throws ParsingException {
+        // integration test for column dependency extraction
+
+        String baseColumn = "base";
+        String tests[][] = {
+                { "value", "base" },
+                { "cell.recon.match.id", "base" },
+                { "value + 'a'", "base" },
```
thadguidry, line 241: Do we also have a tests somewhere for:
1. boolean true/false based on condition if cell value matches value returned from a cross() with another column in same project, as well as different project?  Something like:
```
if ( value.match(  cross(project, colX, colY) ) == true,
    cross(project, colA, colB),
    value
   )
```

main/tests/server/src/com/google/refine/grel/GrelTests.java
```diff
@@ -226,6 +230,37 @@ public void testCrossFunctionEval() throws ParsingException {
...
+                { "\"constant\"", "" },
+                { "1", "" },
+                { "cells.foo", "foo" },
+                { "value + ' ' + cells.foo.value", "base,foo" },
+                { "cells[\"foo\"].value+'_'+value", "base,foo" },
```
tfmorris, line 246:
```suggestion
                { "cells[\"foo\"].value+'_'+value", "base,foo" },
                { "cells.get(\"foo\").value+'_'+value", "base,foo" },
```

main/tests/server/src/com/google/refine/grel/GrelTests.java
```diff
@@ -226,6 +230,37 @@ public void testCrossFunctionEval() throws ParsingException {
...
+                { "parseHtml(value.trim())", "base" },
+                { "cells", null },
+                // this could be analyzed too, but we will never reach completeness anyway!
+                { "get(cells, 'foo'+'bar')", null },
```
tfmorris, line 250: I don't think I understand the comment here. Isn't the column dependency whatever the second argument evaluates to? ie "foobar" in this case.

tfmorris (commented):
I took a pass through and made some comments, but I don't think I have a feel for whether static (lexical only?) analysis is the right way to do this. Would it be simpler to just have operations report what columns they read/wrote?

I think it's more important that operations reliably report errors when they fail, which I don't think is currently the case for all operations, rather than knowing beforehand whether they're likely to fail.

wetneb:
Thanks for the feedback! Here are some responses to your high-level comments.

> I took a pass through and made some comments, but I don't think I have a feel for whether static (lexical only?) analysis is the right way to do this. Would it be simpler to just have operations report what columns they read/wrote?

The goal is precisely to enable operations to report what columns they read and write. Perhaps you have another approach in mind to make this possible? For instance, consider the operation "add column based on this column". How would you change this operation so that it announces which columns it reads? If you do not do so by analyzing the expression which underpins the operation, then I guess you would do so by somehow storing the columns read as a separate metadata field in the operation. How would this metadata field be populated? Would you ask the user to provide this information in the dialog which triggers this operation? As a user, I would find it rather cumbersome to have to provide this information, especially since the benefit isn't directly obvious.

With my approach, the experience of applying operations is preserved. The user does not need to take any extra steps to specify which columns a particular operation depends on or affects.

> I think it's more important that operations reliably report errors when they fail, which I don't think is currently the case for all operations, rather than knowing beforehand whether they're likely to fail.

This sort of analysis isn't really about preventing failures: the main user-facing benefit it would enable is letting the user align the columns mentioned in a recipe to the columns actually present in their project.
I agree error handling in operations needs improving, but that's a distinct subject.

main/src/com/google/refine/grel/ast/ControlCallExpr.java
```diff
@@ -61,6 +64,19 @@ public Object evaluate(Properties bindings) {
         }
     }
 
+    @Override
+    public final Optional<Set<String>> getColumnDependencies(Optional<String> baseColumn) {
+        Set<String> dependencies = new HashSet<>();
+        for (Evaluable ev : _args) {
+            Optional<Set<String>> deps = ev.getColumnDependencies(baseColumn);
+            if (deps.isEmpty()) {
+                return Optional.empty();
```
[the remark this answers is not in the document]
tfmorris: On reflection, I realize that OR is the correct semantics. If ANY argument is indeterminate, the whole expression is.

tfmorris:
> How would this metadata field be populated? Would you ask the user to provide this information in the dialog which triggers this operation?

I'm not quite THAT stupid! I was assuming that it could be gathered during execution and reported after the fact.  Did you decide that tracing all the column accesses would be too slow (or invasive)?

It seems to me that it could be implemented in a backward compatible fashion, might not be terribly slow, and would give 100% accuracy. A sketch of the idea that I have in mind is to use something like:

```java
public interface EvaluableInstrumented extends Evaluable {
    Object evaluate(Properties bindings, EvaluableMetrics metrics);

    class EvaluableMetrics {
       long[] columnAccessCount = new long[0];
    }
}
```

and do similar extensions for `Control`, `Function`, `HasFields` and implement the new interfaces in `Row` and `CellTuple` (as well as the AST classes you modified).

Callers which wanted instrumentation would do something like:

```java
                if (metrics != null && eval instanceof EvaluableInstrumented) {
                    o = ((EvaluableInstrumented) eval).evaluate(bindings, metrics);
                } else {
                    o = eval.evaluate(bindings);
                }
```

We'd have to measure the performance impact, of course, but I think it should be manageable and there are things which could be done (sampling, etc) to mitigate the impact if necessary. This would move the instrumentation to the point at which the cells are accessed, which should make it 100% accurate.

I sketched out the beginnings of this, but don't have a full implementation, so there could be dragons lurking, but I suspect it's doable.

wetneb:
That's an interesting approach! It could potentially work for Jython too, which could be nice (although I suspect the static analysis approach could also work for Jython, with a bit more effort). But I also see some problems with it:
* the fact that such metadata would only be available after executing the operation would make it less useful. Consider the case where you are applying a recipe consisting of a sequence of operations, and the columns mentioned in that recipe are not present in the project you are applying it to. We want to give you the opportunity to map the columns mentioned in your recipe to the ones in your project. This is obviously something that needs to happen before we run the operations in the recipe. If we are not able to extract the column dependencies from expressions, that means that those column dependencies must be included in the JSON serialization of the operations. That means:
   * we don't offer this feature for recipes made with older versions of OpenRefine
   * operations must be able to modify their own metadata fields after their own execution (to store which columns were read), in a way which affects their own JSON serialization. That's a bit of an anti-pattern in my opinion, because operation metadata should just be an abstract description of a transformation to carry out on the grid. Any data that is obtained after execution should rather go to the Change class.
* You write that it would give "100% accuracy", but I don't see it like that. It is true that it would return a set of columns in all cases, but that set of columns would not always accurately represent the set of columns that the operation depends on in general, because changes in project data could affect which columns end up being accessed.
   * for instance, if using the expression `cells[value].value`, which dynamically fetches the value in a column designated by the current value, you would obtain a set of columns which would depend on the values present in the base column for the operation. When re-applying this operation on a new grid, it would not make sense to prompt the user to align those columns to the new project, because the data in the base column will likely reference other columns
   * another instance would be an expression which just concatenates values of all columns into a new column: the expression would always depend on all columns in the project. If we are simply analyzing at runtime which columns have been accessed, we'd return the list of all columns in the project where the operation was originally applied, but again, asking the user to map each such column to a column in the current project would not be useful
* As mentioned above, the goal of this feature is to give the user the opportunity to correct the columns mentioned in their recipe so that it becomes applicable to the current project. Once the user has provided the column mapping, one needs to adapt the operations with the updated column names. It's something that can be done quite naturally with the same sort of syntactic analysis, so to me, it makes sense to use the same technique both for the extraction of column dependencies and for the rewrite. There are alternatives for sure, but they don't seem like a reliable foundation to invest in:
   * we could simply do a text replace on the expressions, mapping the old column names to the new ones. Doing that, we run into the problem of replacing other parts of the expression which aren't actually used to access columns, generating syntactically invalid expressions, and other things like that.
   * we could insert explicit column-renaming operations before running the recipe. Those would then be visible in the undo/redo history, which wouldn't be very clean in my opinion
   * we could run the operations on top of a sort of "virtual" project where columns would have been temporarily renamed. That'd be pretty messy in terms of implementation and it would also keep the mentions of the old columns in the metadata of the applied operations, which would be confusing to users

thadguidry:
> * As mentioned above, the goal of this feature is to give the user the opportunity to correct the columns mentioned in their recipe so that it becomes applicable to the current project. Once the user has provided the column mapping, one needs to adapt the operations with the updated column names. It's something that can be done quite naturally with the same sort of syntactic analysis, so to me, it makes sense to use the same technique both for the extraction of column dependencies and for the rewrite.

This is very much what other ETL (workflow) programs do, allowing the user to remap columns in their workflow steps so that they can be aligned to the newly imported dataset.
Is it worthwhile looking at how Apache Hop or other ETL programs do this in Java (usually through -Meta classes referenced like in https://github.com/apache/hop/blob/main/plugins/transforms/concatfields/src/main/java/org/apache/hop/pipeline/transforms/concatfields/ConcatFields.java)?

tfmorris:
OK, sounds like that idea doesn't resonate. I'll save it for other types of metrics.

> > I think it's more important that operations reliably report errors when they fail, which I don't think is currently the case for all operations, rather than knowing beforehand whether they're likely to fail.
>
> This sort of analysis isn't really about preventing failures: the main user-facing benefit it would enable is letting the user align the columns mentioned in a recipe to the columns actually present in their project.
I agree error handling in operations needs improving, but that's a distinct subject.

My point in mentioning it was that I believe ensuring correct operation is much higher priority than adding additional features and should be addressed first. Not only are there known problems with the current Apply History operation, but there's virtually no test coverage. Why don't we make it reliable first? Buggy == not reproducible.

wetneb:
> My point in mentioning it was that I believe ensuring correct operation is much higher priority than adding additional features and should be addressed first. Not only are there known problems with the current Apply History operation, but there's virtually no test coverage. Why don't we make it reliable first? Buggy == not reproducible.

Yes, I plan to work on this also. I have heard you.

I have rebased the branch and would be keen to merge it unless you have other thoughts.

modules/core/src/main/java/com/google/refine/expr/Evaluable.java
```diff
@@ -47,4 +49,19 @@ public interface Evaluable {
...
      */
     public Object evaluate(Properties bindings);
+
+    /**
+     * Returns the names of the columns this expression depends on.
+     *
+     * @param baseColumn
+     *            the name of the column this expression is based on (none if the expression is not evaluated on a
+     *            particular column)
+     * @return none if the columns could not be isolated: in this case, the expression might depend on all columns in
```
tfmorris, line 59:
```suggestion
     * @return none (ie Optional.empty()) if the columns could not be isolated: in this case, the expression might depend on all columns in
```

modules/core/src/main/java/com/google/refine/expr/Evaluable.java
```diff
@@ -47,4 +49,19 @@ public interface Evaluable {
      * @return
      */
     public Object evaluate(Properties bindings);
+
+    /**
+     * Returns the names of the columns this expression depends on.
```
tfmorris, line 54:
```suggestion
     * Returns a best guess of the names of the columns this expression depends on. NOTE: This has the following limitations: ...
```

I think it would be worth expanding this with the known limitations, so that potential callers have their expectations set correctly.

tfmorris (approved):
I've suggested a couple of improvements to the Javadoc and am OK with it being merged if you believe it will provide value to something down the road.

PR #6749, opened 2024-07-30
Clean up file export. Fixes #6549

### LINKED ISSUE #6549: Text files are saving wrong on macOS Sonoma
### To Reproduce
1. First, I am downloading data from OpenRefine on my Mac (macOS Sonoma Version 14.4.1) in any text file format (CSV, TXT or TSV); selecting the option in the drop-down list or using the custom tabular option.
Screenshot 2024-04-19 at 1 26 09 PM

2. Then, I get the files as plain text but always with an apparently encoding error ("%2E") and always with the extension ".txt"
Screenshot 2024-04-19 at 1 21 09 PM

### Expected Behavior

Despite you can change the name and the format extension afterwards (to CSV for example) I would like to get my file with the original name and the correct extension (TSV, CSV...)

I have changed the language region settings of my MAC, the app to open text files by default and other possibilities but nothing works. 

### Versions
 - Operating System:  macOS Sonoma Version 14.4.1) 
 - Browser Version: Safari
 - OpenRefine:  I have tried with the three last versions and I am getting the same result

### Additional notes
I have now this problem with this software version but with previous macOS versions it worked well.

tfmorris:
Fixes #6549

- don't percent encode period (.) or any other safe characters
- use upper case UTF-8 character set designator per spec
- construct a safe USASCII fallback filename
- for Safari browsers, send the filename as native UTF-8 without rfc5987 encoding since it doesn't support it

### DIFF
--- a/main/src/com/google/refine/commands/project/ExportRowsCommand.java
+++ b/main/src/com/google/refine/commands/project/ExportRowsCommand.java
@@ -45,6 +45,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import com.google.common.net.PercentEscaper;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.hc.core5.http.HttpStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


wetneb (approved):
I assume you considered adding tests for that and it's hard for some reason?

tfmorris:
Cypress doesn't support Safari and the issue was a browser-specific bug. You'd also need a way to figure out what filename the browser used to save the local file which is not something I know how to do in Cypress (but then I don't know Cypress very well).

One testing side note - my test project of "Çłíp_bôârd-" gets saved by Safari as "Ç íp_bôârd-.csv", for reasons I don't understand, while Chrome saves it as "Çłíp_bôârd-.csv", but even though it dropped the ł (l with bar), I think the resulting filename is better than today's "%C3%87%C5%82%C3%ADp%5Fb%C3%B4%C3%A2rd%2D%2Ecsv.txt" (and today's "Clipboard%2Ecsv.txt" becomes "Clipboard.csv", which fixes the actual bug which was reported).

PR #6751, opened 2024-07-31
refactor GetRowsCommand.internalRespond method

Abhiram98:
split up method, rename variables

- extract lines 194-202 to a new method "getRowsForImportJob" to handle the retrieval of rows from an importing job
- rename variable "rwv" -> "rowWriter"
- extract lines 233-259 into "visitRowsOrRecords"
- extract lines 263-267 that pools the reconstructions in a new method called "poolReconstructions" for clarity and reusability

### DIFF
--- a/main/src/com/google/refine/commands/row/GetRowsCommand.java
+++ b/main/src/com/google/refine/commands/row/GetRowsCommand.java
@@ -191,15 +191,7 @@ public class GetRowsCommand extends Command {
         try {
             Project project = null;
 
-            // This command also supports retrieving rows for an importing job.
-            String importingJobID = request.getParameter("importingJobID");
-            if (importingJobID != null) {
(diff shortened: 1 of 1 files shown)


wetneb (commented):
Thank you for the contribution!
It is difficult for me to appreciate your motivation here. Splitting a method into more methods can make sense in certain cases, but it is also possible to do too much of that: by splitting a method you are creating more indirections which can make the code flow harder to follow.

I would suggest that this is not the best task for a new contributor. Did you consider tackling an open issue instead? We have a guide for first code contributions that I encourage you to follow:
https://openrefine.org/docs/technical-reference/code-contributions

Abhiram98:
Thank you for your feedback!

Do any of the refactoring changes in this PR interest you? I am happy to revert others as necessary.

PR #6752, opened 2024-07-31
Refactoring RowAddition Classes

Abhiram98:
- rename "rows" -> "rowList"
- extract method to perform project operation
- rename "data" (line 76) -> "indexString"
- rename "data" (line 85) -> "rowDataArray"
- extract the logic of writing index and count field along with their values into "writeFieldAndValue"
- extract "processLine" method

### DIFF
--- a/main/src/com/google/refine/commands/row/AddRowsCommand.java
+++ b/main/src/com/google/refine/commands/row/AddRowsCommand.java
@@ -73,7 +73,8 @@ public class AddRowsCommand extends Command {
         }
     }
 
-    private static void performProjectOperation(HttpServletRequest request, HttpServletResponse response, Project project, List<Row> rows, int insertionIndex) throws Exception {
+    private static void performProjectOperation(HttpServletRequest request, HttpServletResponse response, Project project, List<Row> rows,
+            int insertionIndex) throws Exception {
         AbstractOperation op = new RowAdditionOperation(rows, insertionIndex);
         Process process = op.createProcess(project, new Properties());
 
--- a/main/src/com/google/refine/commands/row/AddRowsCommand.java
+++ b/main/src/com/google/refine/commands/row/AddRowsCommand.java
@@ -58,23 +58,31 @@ public class AddRowsCommand extends Command {
             return;
         }
 
+        processRequest(request, response);
+    }
+
+    private void processRequest(HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException {
         try {
             Project project = getProject(request);
-            List<Row> rows = getRowData(request);
+            List<Row> rowList = getRowData(request);
             int insertionIndex = getInsertionIndex(request, project);
 
-            AbstractOperation op = new RowAdditionOperation(rows, insertionIndex);
-            Process process = op.createProcess(project, new Properties());
-
-            performProcessAndRespond(request, response, project, process);
+            performProjectOperation(request, response, project, rowList, insertionIndex);
         } catch (Exception e) {
             respondException(response, e);
         }
     }
 
+    private static void performProjectOperation(HttpServletRequest request, HttpServletResponse response, Project project, List<Row> rows, int insertionIndex) throws Exception {
+        AbstractOperation op = new RowAdditionOperation(rows, insertionIndex);
+        Process process = op.createProcess(project, new Properties());
+
+        performProcessAndRespond(request, response, project, process);
+    }
+
     public int getInsertionIndex(HttpServletRequest request, Project project) {
-        String data = request.getParameter(INDEX_PARAMETER);
-        int index = Integer.parseInt(data);
+        String indexString = request.getParameter(INDEX_PARAMETER);
+        int index = Integer.parseInt(indexString);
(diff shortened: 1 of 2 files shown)


tfmorris:
Welcome to the OpenRefine project! Thanks for the contribution. I'm not sure I understand the the purpose of this refactoring. Can you provide some background? Is it associated with an open issue?

Abhiram98:
Thanks for the welcome! 

These changes were inspired by looking at PR #6461 , and are not associated with an open issue. They make the code more readable and easier to follow. 

Some variables were renamed(ex data -> indexString) and larger methods(ex 'RowAdditionChange.load') were split up into smaller ones.

tfmorris:
Thanks for the contribution and apologies for the delay in reviewing. Unfortunately, the project is so short-handed that we don't really have the bandwidth to review refactorings which are essentially just cosmetic. I'll go through and provide feedback on your suggestions to help inform future contributions.

main/src/com/google/refine/commands/row/AddRowsCommand.java
```diff
@@ -58,37 +58,46 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
             return;
         }
 
+        processRequest(request, response);
```
tfmorris, line 61: Splitting 6 lines of code into two extracted methods which call each other actually makes this code harder to read, in my opinion, because of the nesting, the separation of code which is essentially a linear flow, and the interruption of the method signatures.

main/src/com/google/refine/commands/row/AddRowsCommand.java
```diff
@@ -58,37 +58,46 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+    }
+
+    private void processRequest(HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException {
         try {
             Project project = getProject(request);
-            List<Row> rows = getRowData(request);
+            List<Row> rowList = getRowData(request);
```
tfmorris, line 67: I think `rows` is arguably just as good a name as `rowList`

main/src/com/google/refine/model/changes/RowAdditionChange.java
```diff
@@ -81,18 +81,22 @@ public void revert(Project project) {
...
 
+        writeFieldAndValue(writer);
+        for (Row row : _additionalRows) {
+            row.save(writer, options);
+            writer.write('\n');
+        }
+        writer.write(EOC + "\n");
+    }
+
+    private void writeFieldAndValue(Writer writer) throws IOException {
```
tfmorris, line 92: Since this method isn't reused anywhere, I actually prefer a linear flow. If there were multiple copies of the code which could be extracted into a single method and replaced with calls to it, that would be a different story, but since it's only used once, it's on the small side to extract into a method.

main/src/com/google/refine/model/changes/RowAdditionChange.java
```diff
@@ -102,21 +106,27 @@ static public Change load(LineNumberReader reader, Pool pool) throws Exception {
...
-                }
-            }
+            index = processLine(reader, pool, rows, index, line);
         }
 
         return new RowAdditionChange(rows, index);
     }
+
+    private static int processLine(LineNumberReader reader, Pool pool, List<Row> rows, int index, String line) throws IOException {
+        int count;
```
tfmorris, line 116: Declaring `count` where it's first used would probably be better (and the declaration in the `load()` method above looks like it's no longer used and can probably be removed).

main/src/com/google/refine/commands/row/AddRowsCommand.java
```diff
@@ -58,37 +58,46 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+        AbstractOperation op = new RowAdditionOperation(rows, insertionIndex);
+        Process process = op.createProcess(project, new Properties());
+
+        performProcessAndRespond(request, response, project, process);
+    }
+
     public int getInsertionIndex(HttpServletRequest request, Project project) {
-        String data = request.getParameter(INDEX_PARAMETER);
-        int index = Integer.parseInt(data);
+        String indexString = request.getParameter(INDEX_PARAMETER);
```
tfmorris, line 85: While this is arguably a better variable name, the real issue with this code is that if the parameter is empty or badly formatted, `Integer.parseInt` is going to throw a NumberFormatException.

tfmorris (commented):
I've provided feedback on the suggested changes inline below. I hope you find the feedback useful and not too harsh, but I also want to be realistic about the types of refactorings that we would find useful and be able to invest the time in reviewing.

Improvements to variable names would be much easier to get accepted during the initial code review process, but once the code is committed, the bar becomes higher because any improvement needs to be weighed against the cost of the code churn.

If you'd like work specifically on improving this class, I'd suggest writing tests that exercise missing and poorly formatted parameters and then improving the code to handle that.

There are also a large number of open issues, some labelled with "Good First Issue," that you could investigate.

Abhiram98:
Thank you very much for your detailed feedback! This is very valuable to me

PR #6755, opened 2024-08-02
Add new GREL function to calculate the edit distance #6613

zyadtaha:
(cherry picked from commit ba224938c5b7d7828033604e6b005177efa51f1e)

Fixes #{issue number here}

- 
- 
-

### DIFF
--- a/main/src/com/google/refine/util/EditDistance.java
+++ b/main/src/com/google/refine/util/EditDistance.java
@@ -1,0 +1,23 @@
+import org.apache.commons.text.similarity.LevenshteinDistance;
+
+public class EditDistance {
+
+    /**
+     * Calculates the edit distance between two strings using the Levenshtein distance algorithm.
+     * 
+     * @param s1        First string.
+     * @param s2        Second string.
+     * @param threshold Optional threshold for early termination.
+     * @return The Levenshtein distance between the two strings.
+     */
+    public static Integer editDistance(String s1, String s2, Integer... threshold) {
+        LevenshteinDistance levenshteinDistance;
+        if (threshold.length > 0 && threshold[0] != null) {
+            levenshteinDistance = new LevenshteinDistance(threshold[0]);
+        } else {
+            levenshteinDistance = new LevenshteinDistance();
+        }
+
+        return levenshteinDistance.apply(s1, s2);
+    }
+}


zyadtaha:
I think you were interested in this function, so if you want to review this PR, feel free to do so. @thadguidry

main/src/com/google/refine/expr/functions/strings/EditDistance.java
```diff
@@ -0,0 +1,60 @@
...
+            LevenshteinDistance levenshteinDistance;
+            if (threshold != null) {
+                levenshteinDistance = new LevenshteinDistance(threshold);
+            } else {
+                levenshteinDistance = new LevenshteinDistance();
+            }
+
+            return levenshteinDistance.apply(s1, s2);
+        }
+        return new EvalError(ControlFunctionRegistry.getFunctionName(this) + " expects 2 arguments: string 1, string 2");
```
ostephens, line 43: Use EvalErrorMessage expects_two_strings so this error is translatable. Example implementation: https://github.com/OpenRefine/OpenRefine/blob/d723b0680bbacd1ada4f31d3f518a299f3d2d9de/main/src/com/google/refine/expr/functions/strings/IndexOf.java#L55

main/src/com/google/refine/expr/functions/strings/EditDistance.java
```diff
@@ -0,0 +1,60 @@
+
+package com.google.refine.expr.functions.strings;
+
+import java.util.Properties;
+
+import org.apache.commons.text.similarity.LevenshteinDistance;
```
ostephens, line 6: Why are we using a different Levenshtein distance calculator here to the one we use in our clustering? For consistency shouldn't we use the same algorithm in both places?

main/src/com/google/refine/expr/functions/strings/EditDistance.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Calculate the number of edits required to make one value perfectly match another.";
+    }
+
+    @Override
+    public String getParams() {
+        return "string s1, string s2";
```
ostephens, line 53: getParams here should include optional parameters so:
`string s1, string s2, integer threshold (optional)`

ostephens (commented):
There are different ways of calculating editDistance (see https://en.wikipedia.org/wiki/Edit_distance#Types_of_edit_distance) so I think this function probably needs to be named more specifically "levenshteinDistance" 

I think being more specific in naming is probably a better idea than offering multiple algorithms in a single function as we know that different functions could have different additional parameters (e.g. 'threshold' in this case)

thadguidry:
Agree with @ostephens  that the function should be named `LevenshteinDistance()`

main/src/com/google/refine/expr/functions/strings/EditDistance.java
thadguidry: Hi @ostephens !

The Simile Vicino library is no longer maintained (well, @tfmorris technically is our maintainer now for it I think?).
But indeed not everything old is worthless.  However, in the case of the Simile Vicino library, it uses basically William Cohen's naive implementation of NeedlemanWunsch with unit costs and is not highly optimized, and can use quite a bit of memory keeping unused parts of the matrix already discounted, but still held in memory (I guess Cohen thought memory was cheap 15 years ago, but didn't consider that maybe 1 million strings would be compared in batches? dunno).
https://github.com/TeamCohen/secondstring/blob/master/src/com/wcohen/ss/MemoMatrix.java

- https://github.com/tfmorris/simile-vicino-original/blob/master/src/edu/mit/simile/vicino/distances/LevenshteinDistance.java ->
  - https://github.com/TeamCohen/secondstring/blob/master/src/com/wcohen/ss/Levenstein.java ->
    - https://github.com/TeamCohen/secondstring/blob/master/src/com/wcohen/ss/NeedlemanWunsch.java

In contrast to the Apache Commons Text LevenshteinDistance implementation which can use less memory and actually avoid OutOfMemoryError and also uses faster twin single-dimensional arrays, and is highly performant with large strings as well as small and the JVM optimizes it well through intrinsic operations to boot.  Apache also has the `unlimitedCompare()` option without a threshold which allows distance calculations that are not limited to a maximum length. https://github.com/apache/commons-text/blob/master/src/main/java/org/apache/commons/text/similarity/LevenshteinDistance.java

ostephens: @thadguidry I'm not arguing for one option or the other, but consistency in OpenRefine. I think we should use the same Levenshtein implementation in the clustering as in this function otherwise clusters found via this option are going to differ to clusters found in the clustering tool

thadguidry: Levenshtein distance calculation is the same and thus the same result (number of signle-character edits needed to change one word to another).  However, the number of calculations needed to arrive at a result can be less with Apache's implementation.  Does that make more sense?

ostephens: Again I'm not arguing about which one is used, just use the same one in both places!

ostephens: >Levenshtein distance calculation is the same and thus the same result (number of single-character edits needed to change one word to another)

I think I've seen some variation in the past in terms of what counts as a "single character edit" in different implementations, specifically in terms of the substitution type, but I'm not completely sure tbf. I see the argument that "Levenshtein should be Levenshtein so they ought to be identical in outcome" but I'm not sure this is guaranteed

thadguidry: Ah, you mean picking one and using it in both the new GREL function and also in Clustering.  Agree.  Sorry, I misunderstood your original concern.

ostephens: And just to follow up on my concerns, I think this description of Levenshtein and related algorithms shows that they can actually vary in terms of the 'distance' they calculate:

From https://www.baeldung.com/cs/levenshtein-distance-computation, with my bold+italic emphasis on key points:
> Levenshtein distance is the most popular metric among the family of distance metrics known as edit distance. These sibling distance metrics differ in the set of elementary operations allowed to execute the transformation. For example, [Hamming distance](https://www.baeldung.com/cs/string-similarity-edit-distance#levenshtein-distance) permits substitutions only. On the other hand, the [Damerau-Levenshtein distance](https://www.baeldung.com/cs/string-similarity-edit-distance#damerau-levenshtein-distance) allows character transpositions in addition to the set defined by the Levenshtein distance. Thus, it’s commonly used instead of classical Levenshtein distance _**under the same name.**_

>In classical Levenshtein distance, every operation has a unit cost. Additionally, _**specialized versions may assign different costs to operations**_. Furthermore, the cost may also depend on the character. Here, we want to define a complex operation cost function to represent character visual or phonetic similarity.

zyadtaha: The clustering results from Apache are different from those from Simile Vicino. Is that alright with you? @thadguidry @ostephens

Here are the exported clusters if you want to review them:
- [Clustering results with Apache LevenshteinDistance](https://github.com/user-attachments/files/16478108/clusters_lieux.de.tournage.a.paris.csv_Producteur_2024-08-03T04_53_55.662Z.json)
- [Clustering results with Simile Vicino LevenshteinDistance](https://github.com/user-attachments/files/16478104/clusters_lieux.de.tournage.a.paris.csv_Producteur_2024-08-03T04_52_44.944Z.json)

zyadtaha: P.S. Antonin and I planned that if we finished the clustering project early, we would work on finishing [the PR for the passjoin algorithm](https://github.com/OpenRefine/OpenRefine/pull/1007). According to the original contributor, this algorithm is very fast:

> (1) The new algorithm is about 60 times faster than the current Levenshtein-distance KNN algorithm used in OpenRefine. (Detail: https://gist.github.com/lispc/d4ee8de81e7bfa6bc352)

thadguidry: @zyadtaha Please provide me the original dataset as well, in order for me to analyze the cluster sets properly.

zyadtaha: @thadguidry here it is: [lieux-de-tournage-a-paris.csv](https://github.com/user-attachments/files/16478842/lieux-de-tournage-a-paris.csv)

thadguidry: Well, to start with, and maybe if we answer why, we'll see the root cause of issues.
If I compare just 1 cluster set (a number of clusters with minor additions/substitutions/deletes) such as

image

When I look at the APACHE LEVENSHTEIN JSON file in OpenRefine and facet on the choices, I don't see `Pathé ProdExé` as I do in the original data.

When looking at SIMILE VICINO JSON

image

then I do see that one of the choices is `Pathé ProdExé`, which makes sense to me.

This leads me to suspect that:
- A. Apache Levenshtein or some of our other code or your changes is inserting a whitespace when lowercase/uppercase is within the same word boundary?  I.E.  ``Pathé ProdExé` becomes `Pathé Prod Exé` maybe?
- B. Your export for Apache Levenshtein clusters isn't complete or missing something?
- C. Something else I'm missing?

@ostephens Maybe we just focus on this string and compare notes and @zyadtaha perhaps debug on your end to see what is happening?
or we just punt for now and stick to Simile Vicino's Levenshtein implementation for both.  And worry about debugging Apache Levenshtein later?  Up to you folks.

zyadtaha: @thadguidry, @ostephens thanks for your review.

First, why choose between Apache and Simile Vicino? Isn't the PassJoin algorithm faster than both?

Second, I accidentally sent the wrong file out of the two I sent earlier. Sorry about that.
Here are the correct data:
- [Original dataset](https://github.com/user-attachments/files/16503465/Original.xls)
- [Clustering results with Simile Vicino LevenshteinDistance](https://github.com/user-attachments/files/16503457/Simile.xls)
- [Clustering results with Apache LevenshteinDistance](https://github.com/user-attachments/files/16503458/Apache.xls)

They differ only in this value:
```
Original
Pathé ProdExé 3
----------------------
Simile
PATHÉ PROD EXÉ 3
----------------------
Apache
Pathé ProdExé 3
```

> or we just punt for now and stick to Simile Vicino's Levenshtein implementation for both. And worry about debugging Apache Levenshtein later? Up to you folks.

Yes, let's proceed with the Simile Vicino implementation in this PR, and then open an issue to decide which algorithm—Simile Vicino, Apache, or PassJoin—works best with OpenRefine. 
Do you all agree? @wetneb @ostephens @thadguidry

thadguidry: Yes, Simile Vicino for now as our chosen Levenshtein implementation is fine.  It's slow as molasses because of only using 1 index, but we can offer faster alternatives in-situ later on.

> First, why choose between Apache and Simile Vicino? Isn't the PassJoin algorithm faster than both?
Yes, it's faster, primarily because it uses [Vectors and your CPU's SSE extensions](https://github.com/thadguidry/Passjoin15/blob/master/passjoin.cc#L2357).  My understanding was that PassJoin was opinionated (their paper describes "efficient techniques"):
>> for each string, Pass-Join selects some of its substrings and uses the selected substrings to find candidate pairs using the inverted indices.

So my worry would be where we see large differences and lose out on for certain formats of long or short strings.  Most of the implementations of various edit distance algorithms call them "terms" but sometimes "substrings".  Either way, there is some complexity in what is considered a "word", "term", "substring", "part" of a string.  Certainly, whitespace is a word boundary, but sometimes there are other boundaries, like maybe 2 doublequotes "" together might signify an error condition in the string of `Samuel "" Clemens` where `Langhorne` wasn't inserted as a middle name from some system, and so the string produced was `Samuel "" Clemens` instead of `Samuel Langhorne Clemens`.  The edit distance is quite large between `Samuel Langhorne Clemens` and `Samuel "" Clemens` as it should be, but depending on the algorithm, it might be less/more small than you first realize depending on how those 2 doublequotes are categorized and treated as.  This is just a simple example, but it gets more involved in some algorithms in how they treat Unicode Connector Punctuation or Dash Punctuation (where Dashes are often indeed used as connectors, example, Pass-Join versus PassJoin), in Levenshtein the edit distance is always 1, but some implementation of Levenshtein (like @ostephens hinted at) might assume the edit distance is 0, because connectors or dashes were implemented to fallout.

So any algorithm for edit distance has to make assumptions on what is or isn't a word boundary, where should a word be segmented or have it's boundaries considered.  Some algorithms are aggressive in segmentation, and some less so.  I don't know what category PassJoin (original) falls into.  I do know that Apache only uses the [Unicode Boundary categories](https://www.unicode.org/reports/tr29/) and Simile mostly as well (except for a few new ones).

My worry is much less subdued when we will offer users correctly: multiple algorithms to pick and choose from, instead of any 1 "Edit Distance" implementation.  Because there isn't just 1, but like 25+.  And also provide to the user in our UI some help context links to our docs for each algorithm to let them know about quirks or segmentation rules in the implementations.

zyadtaha: @thadguidry Thanks for your great explanation :) 
Would it be helpful if I opened a new issue to continue the discussion and gather opinions from other contributors?

thadguidry: I'd suggest you do that kind of discussion in our Forums under [Development](https://forum.openrefine.org/c/dev/8)

wetneb (approved):

main/src/com/google/refine/expr/functions/strings/LevenshteinDistance.java
```diff
@@ -0,0 +1,50 @@
+
```
tfmorris, line 1: Please include our standard module header.

main/tests/server/src/com/google/refine/expr/functions/strings/LevenshteinDistanceTest.java
```diff
@@ -0,0 +1,39 @@
+
```
tfmorris, line 1: Please include our standard module header.

tfmorris (changes requested):

main/src/com/google/refine/expr/functions/strings/LevenshteinDistance.java
wetneb: If we care about this I'd rather add automation to take care of it rather than asking contributors to do it manually…

tfmorris: I'm fine with automation. Most IDEs can be set up to use a standard module template. As a matter of fact, we used to have this in the Eclipse project settings. We could restore that or add external tooling (or even change the code style if that's what the team wants).

wetneb: There is the `mvn license:update-file-header` command which could be run from time to time (with the plugin appropriately configured), or integrated in `./refine lint`.
Also maybe the shorter license header could be used:
```
//  SPDX-FileCopyrightText: OpenRefine contributors
//  SPDX-License-Identifier: BSD-3-Clause
```

thadguidry: I do like the shorter license header.

ostephens (approved):

main/src/com/google/refine/expr/functions/strings/EditDistance.java
zyadtaha: [Here's the discussion](https://forum.openrefine.org/t/best-levenshtein-distance-algorithm-for-openrefine-simile-vicino-apache-or-passjoin/1715)

tfmorris (changes requested):
Thanks for adding the copyright headers while we sort out what our approach will be long term. 

Two tiny tweaks and this is ready to merge. Thanks!

main/src/com/google/refine/expr/functions/strings/LevenshteinDistance.java
```diff
@@ -0,0 +1,82 @@
+/*
+
+Copyright 2010, Google Inc.
```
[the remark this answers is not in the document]
zyadtaha: Done.

main/src/com/google/refine/commands/project/ExportRowsCommand.java
```diff
@@ -107,9 +108,23 @@
...
                 String filename = path.substring(path.lastIndexOf('/') + 1);
-                PercentEscaper escaper = new PercentEscaper("", false);
-                filename = escaper.escape(filename);
-                response.setHeader("Content-Disposition", "attachment; filename=" + filename + "; filename*=utf-8' '" + filename);
+                String userAgent = request.getHeader("User-Agent");
+                if (userAgent != null && userAgent.contains("Safari/") && !userAgent.contains("Chrome/")
+                        && !userAgent.contains("Chromium/")) {
+                    // Safari doesn't support rfc5897 and just wants straight UTF-8, but strip any controls to avoid
+                    // complaints about potential request/response splitting attacks
+                    response.setHeader("Content-Disposition", "attachment; filename=" + filename.replaceAll("\\p{Cntrl}", ""));
```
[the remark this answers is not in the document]
zyadtaha: I think this something unrelated to this PR. Should I dismiss it? @tfmorris

zyadtaha:
Sorry if I made things complex. It is ready to merge now.

tfmorris (approved):
Looks great! Thanks for your persistence.

main/tests/server/src/com/google/refine/expr/functions/strings/LevenshteinDistanceTest.java
```diff
@@ -0,0 +1,71 @@
...
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+*/
+
+package com.google.refine.expr.functions.strings;
+
+import org.junit.Assert;
```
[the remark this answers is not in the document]
zyadtaha: Thanks for catching that! I'll use TestNG in upcoming PRs. Let me know if there are any specific areas to watch out for.

zyadtaha:
I've submitted a pull request for the documentation, which you can find [here](https://github.com/OpenRefine/openrefine.org/pull/363).

PR #6758, opened 2024-08-03
add missing parameter for JavaDocs in MediaInfoEdit

thadguidry:
@wetneb do you think I worded the parameter description well enough?


wetneb (changes requested):
I would not say that an edit contributes to rows, but rather the other way around. Since edits are generated by rows, the meaning of "contributing" in this context is "which rows contributed to generating this edit"

wetneb (approved):

PR #6769, opened 2024-08-09
CR char added in import flow for CSV/TSV/Clipboard when last record has no value set in the last cell/column

### LINKED ISSUE #6691: Carriage Return added to cell value incorrectly during import
### Issue
During clipboard import using a minimal dataset to reproduce the issue found in a larger datasets with sometimes voided sensor data from oceanographic buoys.  When cells are `null`, not all cells are detected as `null` and if `null` found at the end of a row are getting incorrectly replaced as a carriage return (CR) for the cell value, rather than being treated as a row delineation and the cells' value storing a `null`.

### To Reproduce
1. Add minimal dataset to the Clipboard importer and Create Project with default settings checked.
2. Then, click Edit on Column C, row 4 and inspect its value which has an embedded char(13) carriage return.

### Current Results
char(13) is hidden but selectable and can be copied.  Also shown in between a `<span> </span>` of HTML.

### Expected Behavior
I expected that the 4th row currently being imported it seems as:

```
null,null,CR
````

would be parsed and stored as

```
null,null,null
```

### Screenshots
image
image
image

### Versions
 - Operating System: Windows 11 Pro 64GB
 - OpenRefine: 3.8.1

### Datasets

```
"A","B","C"
"1","2","3"
"a","a","a"
"","",""
,,
'','',''
```

Optional - The last row 6 in the CSV are wrapped single quotes and were added just to help me verify during Java debugging and allow comparison to CSV row 4 and 5 for the CsvParser and the TabularImportingParserBase debugging.

### Additional context

When sensor row data lacks values then it is voided (connection issues, low battery, etc.) with `null` values.
For instance, there indeed would sometimes be rows looking like the minimal dataset where no value exists between comma `,` separators.  I.E. `null` values (not empty strings wrapped in double quotes).
When sensor row data has values, then they are typically wrapped in double quoted strings, sometimes with certain buoy model versions the sensors produce empty strings because the value is that of something akin to "nothing detected", so that is represented as `""`.

sunilnatraj:
CR char added in import flow for CSV/TSV/Clipboard when last record has no value set in the last cell/column

Fixes #6691 

- Update the isNonBlankData method in ExpressionUtils to do empty check post control character removal 
- This will ignore strings with only control characters
- Tested with CSV/TSV file and Clipboard import flows

### DIFF
--- a/main/src/com/google/refine/expr/ExpressionUtils.java
+++ b/main/src/com/google/refine/expr/ExpressionUtils.java
@@ -43,6 +43,7 @@ import java.util.Set;
 
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.base.CharMatcher;
 
 import com.google.refine.model.Cell;
 import com.google.refine.model.Project;
@@ -111,7 +112,7 @@ public class ExpressionUtils {
     static public boolean isNonBlankData(Object o) {
         return o != null &&
                 !(o instanceof EvalError) &&
-                (!(o instanceof String) || ((String) o).length() > 0);
+                (!(o instanceof String) || !CharMatcher.javaIsoControl().removeFrom((String) o).isEmpty());
     }
 
     static public boolean isTrue(Object o) {


sunilnatraj:
@wetneb Can you please review this

PR #6771, opened 2024-08-10
The README states a future NPM version, so this must be about node.js

Ash-Crow:
- The README file states that it requires NPM 16 or newer, but npm is currently at version 10.8. It is however distributed as part of node.js which itself is at version 22.6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # OpenRefine
 
 [![DOI](https://zenodo.org/badge/6220644.svg)](https://zenodo.org/badge/latestdoi/6220644)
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thanks a lot, the text was indeed misleading.
I wonder if we should still mention that we depend on NPM, in cases people have Node.js installed but not NPM? But then I wouldn't know which minimal version to require.

tfmorris:
@wetneb is this requirement needed for anything other than Cypress? Cypress docs currently say Node 18,20, or 22 and our [Cypress installation instructions](https://openrefine.org/docs/technical-reference/functional-tests#1-install-cypress) call for yarn.

tfmorris (approved):
Thanks for the contribution! I bumped the Node version to 18 to match the Cypress requirement

PR #6778, opened 2024-08-11
fix: Positioning of drop placeholder for file import

### LINKED ISSUE #6735: Incorrect placement of drag-and-drop placeholder on start screen
The placeholder of the drag-and-drop area ("Drag files here") in the start screen is incorrectly placed outside of the area that accepts drops:

### To Reproduce
Steps to reproduce the behavior: just start OpenRefine

### Current Results

image

### Expected Behavior

The text should be centered in the area that accepts drops.

### Versions
 - Operating System: Debian
 - Browser Version:  Firefox 115.12.0esr
 - OpenRefine: master branch

wetneb:
Closes #6735.

The fix avoids the `65%` which was chosen to fix the text approximately as this value can depend on font size and ensures that the width of the containing element matches that of the file input.


tfmorris (approved):
Doesn't look any different on my system, so if it improves rendering on other systems, I'm fine with it.

PR #6788, opened 2024-08-13
File type detection fails when multiple files are imported at once

### LINKED ISSUE #6772: File type detection fails when multiple files are imported at once
The heuristic to detect file types seems to have a weird behavior when working on multiple files, even though those files have a consistent file type which OpenRefine is able to detect correctly when they are imported individually

### To Reproduce

Create two text files `test.md` and `test2.md`, both containing the following content:
```
## Markdown test

This is a [test](https://gnu.org) test
```

2. Import both of those files at the same time (by selecting them both in the same folder)

### Current Results

The files are incorrectly detected as Excel files, which obviously fails.

### Expected Behavior

They should be detected as text files (which indeed happens when selecting only one file)

### Screenshots

image

### Versions
 - Operating System: Debian testing
 - JRE or JDK Version: 19
 - OpenRefine: 3.9-SNAPSHOT
 
### Additional context
Observed at Wikimania 2024

sunilnatraj:
File format info was not getting set when multiple files are uploaded.

Fixes #6772 

- Set format in the import flow for multiple files
- Identify format based on the file, added guessBetterFormat() method
- Tested with multiple file types the format is getting set correctly

### DIFF
--- a/main/src/com/google/refine/importing/ImportingUtilities.java
+++ b/main/src/com/google/refine/importing/ImportingUtilities.java
@@ -439,6 +439,7 @@ public class ImportingUtilities {
                             calculateProgressPercent(update.totalExpectedSize, update.totalRetrievedSize));
 
                     JSONUtilities.safePut(fileRecord, "size", saveStreamToFile(stream, file, null));
+                    JSONUtilities.safePut(fileRecord, "format", guessBetterFormat(file, request.getCharacterEncoding(), "text"));
                     // TODO: This needs to be refactored to be able to test import from archives
                     if (postProcessRetrievedFile(rawDataDir, file, fileRecord, fileRecords, progress)) {
                         archiveCount++;
@@ -1036,6 +1037,32 @@ public class ImportingUtilities {
                     }
                 }
             }
+        }
+        return bestFormat;
+    }
+
+    static String guessBetterFormat(File file, String fileEncoding, String bestFormat) {
+        if (bestFormat != null && file != null) {
+            while (true) {
+                String betterFormat = null;
+
+                List<FormatGuesser> guessers = ImportingManager.formatToGuessers.get(bestFormat);
+                if (guessers != null) {
+                    for (FormatGuesser guesser : guessers) {
+                        betterFormat = guesser.guess(file, fileEncoding, bestFormat);
+                        if (betterFormat != null) {
+                            break;
+                        }
+                    }
+                }
+
+                if (betterFormat != null && !betterFormat.equals(bestFormat)) {
+                    bestFormat = betterFormat;
+                } else {
+                    break;
+                }
+            }
+
         }
         return bestFormat;
     }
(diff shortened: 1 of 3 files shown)


sunilnatraj:
@wetneb Please review

wetneb (commented):
Thank you for tackling this! I haven't had a close look at the code but for a change like this I think it would be worth having some tests, would you be up for adding some?

sunilnatraj:
@wetneb I will check and update the test cases

main/src/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -1040,6 +1041,32 @@ static String guessBetterFormat(ImportingJob job, ArrayNode fileRecords, String
         return bestFormat;
     }
 
+    static String guessBetterFormat(File file, String fileEncoding, String bestFormat) {
```
tfmorris, line 1044: It looks like the bulk of the code in this method is duplicated from the method immediately above. 
Can things be refactored to reduce/eliminate the duplication?

tfmorris (changes requested):
Thanks for the contribution! In addition to @wetneb 's suggestion for test coverage, it looks like you might be able to reduce them amount of duplicated code in line with the DRY (Don't Repeat Yourself) principle.

main/src/com/google/refine/importing/ImportingUtilities.java
sunilnatraj: @tfmorris I have addressed this.

sunilnatraj:
@tfmorris @wetneb Added test cases and removed redundant code. Please review

tfmorris:
Thanks for the updates! The code looks good and my functional testing was good. I'm not sure the file selection code is working as intended, but I think that's a separate problem.

The test failure, however, appears real. I've re-run the tests a couple of times and the failure has persisted, so I don't think it's just a flaky test.

tfmorris:
Let's have another look at the test failure here

tfmorris:
I had a quick look at the failing tests and it looks like they're all failing for the same reason - it's guessing that a CSV file is just a Line oriented text file instead of the more specific CSV file type.

You should be able to reproduce the problem by importing the attached file from the test. It should be guessed as a CSV file.

[food.mini.csv](https://github.com/user-attachments/files/16723762/food.mini.csv)

sunilnatraj:
@tfmorris I have addressed the issue; can you please review the PR

tfmorris (approved):
Looks great! Thanks for writing the tests too.

It looks like perhaps you merged `master` into your branch and things got messed up, so I rebased things onto master. Generally I think `rebase` is better for syncing things up than `merge` but it's somewhat a matter of preference.

Two other minor things I fixed up:
- camel case test names
- removed try/except block and just let the test framework fail on any errors.

tfmorris:
And congratulations on your first pull request merged!

sunilnatraj:
Thank you @tfmorris

PR #6792, opened 2024-08-15
Remove uses of JUnit. Fixes #6791.

### LINKED ISSUE #6791: Migrate JUnit tests to TestNG
The bulk of our code uses TestNG, but there are a few cases where JUnit's asserts is used, including some which were added recently. 

We should:
1. Migrate the remaining JUnit asserts to TestNG
2. If possible, figure out a way to prevent/warn them from being added in the future.

tfmorris:
Fixes #6791

- migrates usages of org.junit.Assert to org.testng.Assert
- migrates usages of org.junit.assertThrows to TestNG annotations
- migrates some uses of org.testng.AssertJUnit.assertEquals to org.testng.Assert.assertEquals, swapping the order of parameters (unless they were in the wrong order to start with)
- adds TODOs for remaining uses of org.testng.AssertJUnit (which uses the JUnit parameter order)


wetneb (approved):
Thanks for taking care of that!

PR #6796, opened 2024-08-16
Extend Exporter interfaces to phase out use of Properties

tfmorris:
Refs #6419.

This PR includes just the interfaces changes with no changes to the bundled exporters or tests to demonstrate 100% API compatibility:
- replace Properties with Map<String, String> in a new `export` method
- deprecate the old method
- add default implementations for both new and old signature
- add helper utilities

A followup PR based on the work by @karindev in #6460 will migrate the bundled exporters to use the new API. Until the deprecation period expires, exporters which support the new or old interface will continue to work unchanged.

### DIFF
--- a/main/src/com/google/refine/commands/project/ExportRowsCommand.java
+++ b/main/src/com/google/refine/commands/project/ExportRowsCommand.java
@@ -38,6 +38,7 @@ import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.util.Enumeration;
+import java.util.Map;
 import java.util.Properties;
 
 import javax.servlet.ServletException;
@@ -69,6 +70,7 @@ public class ExportRowsCommand extends Command {
      * This command uses POST but is left CSRF-unprotected as it does not incur a state change.
      */
 
+    @Deprecated(since = "3.9")
     @SuppressWarnings("unchecked")
     static public Properties getRequestParameters(HttpServletRequest request) {
         Properties options = new Properties();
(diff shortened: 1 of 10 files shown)


wetneb (commented):
I'm not sure about this. I have doubts that using `String optionsString` is much better than using `Properties`, and so whether the round of deprecation and migration is worth it.

main/src/com/google/refine/exporters/CustomizableTabularExporterUtilities.java
```diff
@@ -74,13 +74,35 @@ abstract public class CustomizableTabularExporterUtilities {
...
+     *            a serializer which implements the TabularSerializer interface
+     * @since 3.9
+     */
+    static public void exportRows(
+            final Project project,
+            final Engine engine,
+            String optionsString,
+            final TabularSerializer serializer) {
 
-        String optionsString = (params != null) ? params.getProperty("options") : null;
```
tfmorris, line 83: @wetneb when you say 

> I have doubts that using `String optionsString` is much better than using `Properties`, and so whether the round of deprecation and migration is worth it.

Is this what you're referring to? (It would help me follow along if Github's contextual commenting were used, perhaps with specific suggestions.) 

To clarify, I didn't introduce `optionString` - it's preexisting.  The old API used a `Properties` with a serialized JSON string in the "options" key. Since that was the only thing used, I simplified the API to just pass the string instead of a Map<String, String> with the serialized JSON keyed by "options" like was done for the Properties.

If you feel strongly that a 1:1 replacement with Map<String, String> is better, I'm happy to make the change, but I want to make sure I understand what your objection is.

wetneb: Ah yes, that's fine, thanks for the clarification.

wetneb (approved):

PR #6799, opened 2024-08-19
File type detection fails when multiple files are imported at once

### LINKED ISSUE #6772: File type detection fails when multiple files are imported at once

sunilnatraj:
Added definition for MD format files

Fixes #6772 

- Added definition for MD format files 
- Performed test with multiple file formats
-

### DIFF
--- a/main/webapp/modules/core/MOD-INF/controller.js
+++ b/main/webapp/modules/core/MOD-INF/controller.js
@@ -235,6 +235,7 @@ function registerImporting() {
   IM.registerExtension(".txt", "text");
   IM.registerExtension(".csv", "text/line-based/*sv");
   IM.registerExtension(".tsv", "text/line-based/*sv");
+  IM.registerExtension(".md", "text");
 
   IM.registerExtension(".xml", "text/xml");
   IM.registerExtension(".atom", "text/xml");


tfmorris:
Thanks for sticking with this. I think your previous PR was closer to what we need. Perhaps we should focus on fixing the test failure in that one.

Adding a definition for Markdown files is really just a workaround, because there will always be other types of text files with extensions that we don't recognize, and we would still like things to work in that case.

sunilnatraj:
@tfmorris I will look into the logic for identification of file contents. Do i open a new issue to review and address the FormatGuesser issue?

tfmorris:
> Do i open a new issue to review and address the FormatGuesser issue?

If you think there is a problem with it, yes, please open an issue with as much detail as possible. Thanks!

tfmorris:
I'm going to close this since it's not really the direction we want to go. I've reopened #6788 and provided some feedback there on the failing tests.

PR #6811, opened 2024-08-20
fix #6764

### LINKED ISSUE #6764: ui-tabs-active appears as a text input area rather than a pointer or default
There are places throughout the UI where the ui-tabs apply `cursor: text` rather than `cursor: default` or `cursor: pointer`.  The `cursor: text` allows highlighting the ui-tab text (not beneficial I think) and confusingly changes the cursor shape when hovered over ui-tabs to that of a text input bar making the `ui-tab-active` look like input text areas rather than clickable elements like buttons and tabs typically are.

### To Reproduce
1. First, click on Undo/Redo to make it an active tab, then move mouse over Facet/Filter and Undo/Redo.
2. Notice that the cursor shape changes to a text input bar over Undo/Redo.

### Current Results
ui-tabs in many places of OpenRefine such as Undo/Redo tab and some of the Export dialogs are confusingly showing the `cursor: text` and producing a text input bar upon hovering over any tabs that are in the state of `ui-tabs-active` it seems.

### Expected Behavior
We have various dialogs throughout OpenRefine that use Jquery-ui and it's handling of `ui-tabs` and it's other subclasses such as `ui-tabs-anchor` and `ui-tabs-nav`.  It would be optimal that users see the same cursor shape for ui-tabs since throughout they are indeed clickable elements and not for text input or even beneficial to users to capture the ui-tab text phrase for copy/paste operations.  So I would expect the ui-tabs throughout OpenRefine to show with `cursor: pointer` to bring a consistent UI Tab experience and make it more clear that they are clickable elements.  For `ui-tabs-active` which are not considered clickable since the user is on that active tab, then I would expect we just make those `cursor: default` for the regular arrow cursor rather than `cursor: pointer` for the hand pointer?  Whichever one feels right I think.

### Screenshots

UndoRedoInput

### Versions
 - OpenRefine: 3.8.2

### Additional context
The current default jquery-ui rule seems to be:
```
.ui-tabs .ui-tabs-nav li.ui-tabs-active .ui-tabs-anchor, .ui-tabs .ui-tabs-nav li.ui-state-disabled .ui-tabs-anchor, .ui-tabs .ui-tabs-nav li.ui-tabs-loading .ui-tabs-anchor {
    cursor: text;
}
```
which we probably just need to override the CSS for `.ui-tabs` or maybe specifically just `.ui-tabs-active` , I haven't investigated all the `ui-tabs` areas of OpenRefine but only Undo/Redo and a few Export dialogs.

GittyHarsha:
Fixes #6764 
Before:
openrefine_before

After:
openrefine_after

Changes proposed in this pull request: Add the below in `main/webapp/modules/core/styles/jquery-ui-overrides.css`
```
.ui-tabs .ui-tabs-nav li.ui-tabs-active .ui-tabs-anchor,
.ui-tabs .ui-tabs-nav li.ui-state-disabled .ui-tabs-anchor,
.ui-tabs .ui-tabs-nav li.ui-tabs-loading .ui-tabs-anchor {
	cursor: default;
}
```

### DIFF
--- a/main/webapp/modules/core/externals/jquery-ui/css/ui-lightness/jquery-ui.css
+++ b/main/webapp/modules/core/externals/jquery-ui/css/ui-lightness/jquery-ui.css
@@ -860,7 +860,7 @@ button.ui-button::-moz-focus-inner {
 .ui-tabs .ui-tabs-nav li.ui-tabs-active .ui-tabs-anchor,
 .ui-tabs .ui-tabs-nav li.ui-state-disabled .ui-tabs-anchor,
 .ui-tabs .ui-tabs-nav li.ui-tabs-loading .ui-tabs-anchor {
-	cursor: text;
+	cursor: default;
 }
 .ui-tabs-collapsible .ui-tabs-nav li.ui-tabs-active .ui-tabs-anchor {
 	cursor: pointer;
--- a/main/webapp/modules/core/externals/jquery-ui/css/ui-lightness/jquery-ui.css
+++ b/main/webapp/modules/core/externals/jquery-ui/css/ui-lightness/jquery-ui.css
@@ -860,7 +860,7 @@ button.ui-button::-moz-focus-inner {
 .ui-tabs .ui-tabs-nav li.ui-tabs-active .ui-tabs-anchor,
 .ui-tabs .ui-tabs-nav li.ui-state-disabled .ui-tabs-anchor,
 .ui-tabs .ui-tabs-nav li.ui-tabs-loading .ui-tabs-anchor {
-	cursor: default;
+	cursor: text;
 }
 .ui-tabs-collapsible .ui-tabs-nav li.ui-tabs-active .ui-tabs-anchor {
 	cursor: pointer;
--- a/main/webapp/modules/core/styles/jquery-ui-overrides.css
+++ b/main/webapp/modules/core/styles/jquery-ui-overrides.css
@@ -35,6 +35,12 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
   padding: 0;
 }
 
+.ui-tabs .ui-tabs-nav li.ui-tabs-active .ui-tabs-anchor,
+.ui-tabs .ui-tabs-nav li.ui-state-disabled .ui-tabs-anchor,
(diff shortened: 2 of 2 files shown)


thadguidry:
Hi @GittyHarsha you will instead need to put the default setting in our jquery overrides file under this path:
https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/styles/jquery-ui-overrides.css

thadguidry:
Hi @GittyHarsha you will instead need to put the default setting in our jquery overrides file under this path:
https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/styles/jquery-ui-overrides.css

Do you think or have you tested if only using the setting within `.ui-tabs` scope is enough to take care of it globally?

GittyHarsha:
@thadguidry I investigated the jqueryUI `.ui-tabs` subclasses and found out that only `.ui-tabs-anchor` `has cursor: text` when it is nested under elements  of classes as mentioned in the CSS rule above.


> Do you think or have you tested if only using the setting within `.ui-tabs` scope is enough to take care of it globally?

Sorry I couldn't understand what you mean by taking care globally?
I think the .ui-tabs subclasses like `.ui-tabs-anchor` are generally used nested under elements of `.ui-tabs-active` and other similar subclasses.

So this code change should be sufficient to resolve?

GittyHarsha:
> you will instead need to put the default setting in our jquery overrides file under this path

Sure! Will make the changes

thadguidry (approved):
Thanks so much @GittyHarsha.  Looks good and thanks for the screenshots.

GittyHarsha:
Thanks a lot! 
This is my first open source contribution 
Looking forward to contribute more!

thadguidry:
@GittyHarsha Perhaps you would want to tackle one of our other issues?

GittyHarsha:
> @GittyHarsha Perhaps you would want to tackle one of our other issues?

Sure!

thadguidry:
Go to Issues, then select Labels... then choose Module: Frontend , or perhaps any that intrigue you or align with your skills.  If you know Java as well, then that would be Backend.  You can also combine Labels in search.

GittyHarsha:
> hen choose Module: Frontend , or perhaps any that intrigue you or align with your skills. If you know Java as well, then that would be Backend. You can also combine Labels in search.

Thanks ! I'll look into it.

PR #6823, opened 2024-08-27
Fix Uncaught RangeError when using Cell Edit button to change String to Date data type.

### LINKED ISSUE #6534: Using Cell Edit button to change String to Date data type gives Uncaught RangeError
Console error when changing a cells data type from Text to Date via Cell Edit button.

### To Reproduce
1. First, click Edit Cell and type in a String such "2023-12-01" and click Apply button
2. Then, .click on Edit Cell and choose from it's dropdown to change to `Data type: date`
3. Finally, click Apply button

### Current Results

1. The console immediately shows the following error (No Java stacktrace shows in Terminal log however):

```
project-bundle.js:46913  Uncaught RangeError: toString() radix argument must be between 2 and 36
    at Number.toString (<anonymous>)
    at HTMLButtonElement.commit (project-bundle.js:46913:21)
    at HTMLButtonElement.dispatch (project-bundle.js:5147:27)
    at elemData.handle (project-bundle.js:4951:28)
```
as shown:
image    

2. If I try to use a partial format in the cell value, then I get a popup alert for "not a valid date." as shown:

image

3. If I try to use the full format as is recommended above the Cell Edit dialog, then I get the previous Console error as shown:

image

### Expected Behavior
The cells' String value is converted to Date data type when the correct format is used.
I assume that when following the Steps to Reproduce, that a cell text value of "2023-12-01" would be good enough to convert to the Date data type?
But no format seems to be working.

If I use Edit Cell -> Common Transform -> to Date , then that method does seem to be working however.
So the issue is just likely somewhere in the Edit button on a Cell and perhaps the Javascript function or callbacks? 

### Versions
 - Browser Version: Edge Latest
 - JRE or JDK Version: JDK 21
 - OpenRefine: 4.0 Snapshot

GittyHarsha:
Fixes #6534 

## Issue
The date entered in the cell is parsed using the `Date.parse()` method which returns a timestamp. This timestamp is converted to a formatted date string using `toString("YYYY-MM-DDTHH:mm:ss.sssZ")`.

but concerning the below link
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Bad_radix

The `toString` method has an optional parameter radix, which should be an integer between 2 and 36. But here the radix is given a string as an argument.

## Changes proposed in this pull request:
- use `toISOString` method on a new Date object constructed using the parsed date value.

image

### DIFF
--- a/main/tests/cypress/cypress/e2e/project/grid/row/edit_cells.cy.js
+++ b/main/tests/cypress/cypress/e2e/project/grid/row/edit_cells.cy.js
@@ -117,4 +117,21 @@ describe(__filename, function () {
     cy.assertCellEquals(0, 'Water', '42');
     cy.assertCellEquals(1, 'Water', '42');
   });
+
+  it('Test edit a cell to change String to Date data type', function () {
+    cy.loadAndVisitProject('food.mini');
+    cy.getCell(1, 'Water')
+      .trigger('mouseover')
+      .find('.data-table-cell-edit')
+      .click();
+    cy.get('.menu-container.data-table-cell-editor #typeSelectId').select('date');
+    cy.get('.menu-container.data-table-cell-editor').should('exist');
+    cy.get('.menu-container.data-table-cell-editor textarea').type(
+      '2024-12-12'
+    );
+    cy.get('.menu-container button[bind="okButton"]').click();
+    // ensure value has been changed in the grid
+    cy.get('.menu-container.data-table-cell-editor').should('not.exist');
+    cy.assertCellEquals(1, 'Water', '2024-12-12T00:00:00Z');
+  });
 });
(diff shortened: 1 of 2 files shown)


tfmorris (commented):
Thanks for the contribution! Any chance you could add a Cypress test to make sure that we don't break this in the future?

Also, it's helpful if you can make the pull request title and commit message a little more descriptive so that they stand alone without having to refer to github.

tfmorris:
I think here is where the test could go: https://github.com/OpenRefine/OpenRefine/blob/master/main/tests/cypress/cypress/e2e/project/grid/row/edit_cells.cy.js

GittyHarsha:
Sure! I'll look into it

GittyHarsha:
@tfmorris I added a test case. Should I add more to test any other scenarios?

GittyHarsha:
@tfmorris I added a test case that simulates editing a cell to "2024-12-12" and changing the data type to date.
Should I add test cases for any other scenarios?

tfmorris (approved):
Awesome. Thanks for the test coverage!

GittyHarsha:
Thanks!

PR #6827, opened 2024-09-02
Fix long-running test in StandardReconConfigTests

### LINKED ISSUE #6733: batchReconTestConnectionError stalls tests for 60 seconds
This recently added test: https://github.com/OpenRefine/OpenRefine/blob/d29a6e125d4b938766f2de9e1ca83d0e66fcc888/main/tests/server/src/com/google/refine/model/recon/StandardReconConfigTests.java#L393
adds 60 seconds of dead time to the unit tests.

### To Reproduce
1. Run the unit tests

### Current Results
That one test takes 60 of the 85 seconds for the main unit tests, dwarfing the time used by the other 936 tests.

### Expected Behavior
The test should complete within a few seconds, at the very most, ideally less than a second or two.

wetneb:
Closes #6733. The problem was that the connection error was not that of the mocked response, but instead of the attempted retry made by our HTTP client, which was left stalling because no corresponding mocked response had been set up.

<s>Let's see if my assertion is generic enough because I suspect the error message might be platform-dependent.</s> It is indeed platform dependent now that the test is fixed, so I have removed the assertion on the specific error message and just check that some error message is returned.

### DIFF
--- a/main/tests/server/src/com/google/refine/model/recon/StandardReconConfigTests.java
+++ b/main/tests/server/src/com/google/refine/model/recon/StandardReconConfigTests.java
@@ -464,10 +464,8 @@ public class StandardReconConfigTests extends RefineTest {
             RecordedRequest request1 = server.takeRequest();
             assertNotNull(request1);
 
-            // assertions
+            // the error message is unstable and system-dependent, so we are not asserting for its exact contents.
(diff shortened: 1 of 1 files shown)


tfmorris (approved):
LGTM Main tests run in under 25 seconds on my system.

Issue #6824, opened 2024-08-30
Four Cypress e2e tests are flaky


tfmorris:
The first one is flaky 100% of the time and the other three about 50%:

1. Test the history behavior, ensure expressions are stored
  project/grid/column/edit-column/add_columns_based_on_this_column.cy.js
I suspect perhaps this fails the first time through every time and succeeds on retry
1. Ensure new column is added based on previous column in Python
project/grid/column/edit-column/add_columns_based_on_this_column.cy.js
this has a 10 second timeout which is longer than normal, so perhaps it's known to run long?
1. Ensures the working of Start-Over Button
create-project/preview_project.cy.js
This might be a bug in the application code as claimed by Cypress which says the error is   "Cannot read properties of undefined (reading 'length')"
1. Test a valid Python expression 
project/grid/misc/expressions.cy.js

https://cloud.cypress.io/projects/s5du3k/analytics/flaky-tests/ddc1a2a4-3b8d-d38c-392f-261e0738a123-d77ab44a-cb4f-d134-2324-2a5d78ece7db?branches=%5B%7B%22label%22%3A%22master%22%2C%22suggested%22%3Afalse%2C%22value%22%3A%22master%22%7D%5D&browsers=%5B%5D&chartRangeMostCommonErrors=%5B%5D&chartRangeSlowestTests=%5B%5D&chartRangeTopFailures=%5B%5D&committers=%5B%5D&cypressVersions=%5B%5D&flaky=%5B%5D&operatingSystems=%5B%5D&runGroups=%5B%5D&specFiles=%5B%5D&status=%5B%7B%22label%22%3A%22Passed%22%2C%22value%22%3A%22PASSED%22%7D%2C%7B%22label%22%3A%22Failed%22%2C%22value%22%3A%22FAILED%22%7D%5D&tags=%5B%5D&tagsMatch=ANY&timeInterval=WEEK&timeRange=%7B%22startDate%22%3A%222024-08-23%22%2C%22endDate%22%3A%222024-08-30%22%7D&viewBy=TEST_CASE

### Current Results
Several tests are consistently flaky.

### Expected Behavior
No consistently flaky tests

tfmorris:
The Cypress dashboard doesn't capture the Javascript stack trace from item 3 above. Here it is:

```
      at Refine.PreviewTable._render (http://127.0.0.1:3333/index-bundle.js:47981:28)
[...]

PR #6834, opened 2024-09-02
Increase timeout for Clojure initialization in test

tfmorris:
Followup to #6825 (fix for #6824) which is triggering snapshot build failures even though it ran without issue in CI. :(

Looking at the (successful) CI logs, it appears that Clojure initialization takes pretty close to the default 4 sec timeout, which might cause the test to be flaky, depending on host machine & load, so adding 100% buffer to the timeout.


tfmorris:
Automerged to see if we can unbreak the snapshot build

PR #6770, opened 2024-08-09
WIP - Adds new JavaFX launcher

thadguidry:
Fixes #3221

DRAFT - Work In Progress
- initial module that can be run independently.

TODO:
- [X] make start button launch main app
- [X] stop
- [X] close X should confirm with verify message "ARE YOU SURE?  Yes/No"
- [ ] views:
  - [X] logging (works minimally - but no Copy or Save button yet) - not even sure if we want logs written to disk as check box option?
  - [ ] JVM details?
  - [ ] JVM memory?

PR #6841, opened 2024-09-04
Restore line ending normalization

tfmorris:
For reasons I don't understand, source file line ending normalization was disabled in 9c7cbcb6 (without updating the associated comment). This PR reenables it so that we don't get PRs like the draft #6770 which have every single file listed as modified with whitespace (ie line ending) changes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 # Auto detect text files and perform LF normalization
-* text=off
+* text=auto
 
 # Custom for Visual Studio
 *.cs     diff=csharp


.gitattributes
```diff
@@ -1,5 +1,5 @@
 # Auto detect text files and perform LF normalization
-* text=off
+* text=auto
```
thadguidry, line 2:
```suggestion
* text=auto
* text eol=lf
```
Looks like ideally, we should be further explicit.  This seems to help especially, and when `eol` is unspecified, it [ends up converting to CRLF on Windows by default](https://git-scm.com/docs/gitattributes#_eol), and just not really ideal for our cross platform development, so ends up causing more issues overall.  It mitigate that somewhat for Windows users especially, let's also set `eol=lf` and that will help since it's highly likely that folks typically will NOT have set their `core.autocrlf=true` or `core.autocrlf=input`.

Poking around, this seems to be the settings that other open source projects do that are cross platform, they explicitly normalize the repo to LF endings.

thadguidry (changes requested):

wetneb (approved):

tfmorris:
> it [ends up converting to CRLF on Windows by default](https://git-scm.com/docs/gitattributes#_eol), and just not really ideal for our cross platform development, so ends up causing more issues overall.

@thadguidry Can you expand on the types of problems that this causes? The intention of the setting is that git uses the platform's native line ending locally to simplify use of native tools like text editors, etc, but converts to/from the native LF format in git.

I'm happy to discuss expanding/changing this separately, but the intent here was simplify to reverse earlier damage that was done and put things back the way they were.

thadguidry:
I'm saying (suggesting) that converting to/from LF format is not necessary anymore and can be set to LF all the time.  The only exception is that of `.bat` and `.cmd` files where there's a small chance that a 512 byte boundary might land right at a LF character, and that causes label jumps in batch files to miss and thus batch files labeled functions don't fire or work as intended.

The Git on Windows installation defaults to this below checked setting which configures git to checkout as-is and convert to LF when committing.
image

The IDE's on Windows also set this same preference if git hasn't been configured for the user yet.  Thus, the option of reverting back to how things were before with using native EOL on the platform is also fine.

wetneb:
Let's merge this as is and discuss any further improvements in a separate PR.

thadguidry:
I found the issue I was having.  Evidentially, if you ALREADY have git config `core.autocrlf` set with something, it doesn't seem that the Git for Windows installer will overwrite that previous setting.  Seems to only work for new installations.  After I manually changed my setting to `core.autocrlf=input` then magically the line endings on Windows when doing checkout and pushes is now back to LF.  (regardless of this .gitattributes setting, which is secondary priority, whereas user config's have first priority) Hooray.

Now we know @tfmorris and dunno if we might mention in our contributor docs?  Seems a bit important though for contributors using Windows?

wetneb:
If it's about removing a particular config option that you have set yourself before I don't think this is worth documenting. When people set such a config option themselves they are generally aware of the consequences.

thadguidry:
@wetneb I was quite aware, and had it originally set to `core.autocrlf=false` which should NOT change line endings, and haven't had a problem in 10 or so years? But something caused CRLF to be introduced.  My suspicion lies with my switching to use IntelliJ IDEA for about a week (for its JavaFX capabilities), but agree this is certainly a local config issue in my case.  Anyways, moving on.

PR #6844, opened 2024-09-06
Add alphabetical/chronological sort for multi-file import

### LINKED ISSUE #6773: Sort file names when importing multiple files
When selecting multiple files to import in OpenRefine (with the "Browse…" button), the files don't seem to be imported in any particular order when concatenating them into one project.

Consider the case where you are importing a dump which consists of one file per month for a given year:
```
2023-01-01.txt
2023-02-01.txt
2023-03-01.txt
2023-04-01.txt
...
2023-12-01.txt
```

In this case, it would be helpful if the files could be imported in alphabetical order, which would correspond to chronological order.

### Proposed solution

It is fairly common to design file names so that alphabetically sorting them makes sense. Instead of using an arbitrary and potentially non-deterministic order, we could just sort files by file names before import.

### Alternatives considered

We could also give the user the option to sort files using various criteria (file name, date of last change…)

### Additional context

Observed at Wikimania 2024

elebitzero:
Fixes #6773 

- Sort files alphabetically/chronologically by default for multi-file import.
- Add the ability to sort files in ascending or descending order and sort by size/name with dropdowns.
- Add 'Sort files' section in the multi-file import panel.

AFTER:
OR Import Files Sort Demo


wetneb:
Welcome back @elebitzero, very nice to see you here again :)

I tried it out and while it does sort the files at the first stage where we ask the user to select which files to import, that order does not seem to carry over to the next stages:
image

You can try the following archive: 
[archive.zip](https://github.com/user-attachments/files/16963798/archive.zip)
(you need to disable parsing column headers from the file to get the sort of output I got above)

wetneb:
Very nice, making it configurable is obviously much better.

I wonder if it wouldn't be more practical to have the sorting options in the panel where files can be filtered. For instance in the left-hand side panel in the screenshot below, underneath the "Select by extension" and "Select by regex on file names" sections, as an additional "Sort files" option.
image

With such a move, we would expose those options only when importing multiple files at once, and keep the UI less cluttered for use cases where a single file is imported (which is the norm, I would say). For reference, this is what the start page looks like after this PR:
image


Moving those new sorting options to the next stage would also have the benefit of making the sorting options available to the "Web Addresses (URLs)" import workflow, where it is also possible to import multiple files at once.

wetneb (approved):
Thank you so much @elebitzero! As far as I am concerned, this looks really good.

PR #6854, opened 2024-09-12
Fix message key for description too long issue

sebastian-berlin-wmse:


wetneb (approved):
It does match what we have in the translation.json files, thank you!

PR #6861, opened 2024-09-21
fixes #6164 remove waitForOrOperation() usage and the helper itself

### LINKED ISSUE #6164: Cypress tests: remove calls to waitForImportUpdate() and waitForOrOperation()
Our Cypress test suite relies in various places on two helpers, `waitForImportUpdate()` and `waitForOrOperation()`.

Quoting @elebitzero from #6162:

> `waitForImportUpdate()` has the same problem as `waitForOrOperation()` which I have been trying to remove.  Both of those methods are waiting for a temporary loading/in-progress element to be visible and then not visible.
> 
> The problem is that the element may have been shown and hidden again before this code executes, causing it to wait forever until it times out.
> 
> The solution is to not focus on the transient loading indicators, but to just do `cy.get()` and assertions on the expected state after the operation completes. For long running operations,  the timeout on the `cy.get()` command may need to be increased if the operation could exceed the default timeout of 4 seconds.

### Proposed solution

Remove all uses of those helpers and remove the helpers themselves. This means that the Cypress tests need to be adapted to wait on the actual update, for instance by looking for a particular cell content. See #6162 for an example of how this can be done.

### Alternatives considered

None

### Additional context

Follow-up to #6162.
Marking this as a good first issue because it may not require much familiarity with OpenRefine itself. See https://openrefine.org/docs/technical-reference/functional-tests for documentation about this test suite.

akashinde:
Fixes #6164 

- Usage of `waitForOrOperation()` helper function has been removed from cypress test.
- The helper function itself as well has been removed.
- In-between checks have been added to achieve waiting and to satisfy test conditions.

Steps taken before commit:
- [x]  run lint, clean, build
- [x]  run cypress tests (PASSED)
- [x]  run java tests (PASSED)

Final result:
image

### DIFF
--- a/main/tests/cypress/cypress/e2e/project/grid/column/facet/facets.numeric.cy.js
+++ b/main/tests/cypress/cypress/e2e/project/grid/column/facet/facets.numeric.cy.js
@@ -38,16 +38,19 @@ describe(__filename, function () {
     cy.get('.slider-widget-bracket').eq(1)
     .trigger('mousedown',{ force: true })
     .trigger('mousemove',-130,0,{ force: true })
-    .trigger('mouseup',{ force: true })
+    .trigger('mouseup',{ force: true });
+    cy.get('.facet-range-status').contains('0 — 45');
     cy.get('#summary-bar').contains('72 matching rows');
 
     // wait for the numeric facet to be refreshed before next mouse events
     cy.get('.browsing-panel-indicator').should("not.be.visible");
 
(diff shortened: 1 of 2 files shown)


tfmorris (approved):
Yay! Glad to see that bit of code disappear. Looks great! Thank you very much.

I added extra tests for the actual facet values as a double check just to make the tests a little bit more bulletproof.

PR #6865, opened 2024-09-22
Avoid creating identical facets in reconciliation and Wikibase editing operations

### LINKED ISSUE #6863: Avoid creation of identical facets
Certain operations, like the reconciliation or the Wikibase upload operations, create facets after having run.
Sometimes, those operations can be run multiple times in a row. For instance, when reconciling different parts of a dataset using different settings (via facets), or making a new Wikibase upload after having changed the schema.

In those situations, running the operation again will create the same facets again, even if they are still open from the previous operation.
This is annoying because those identical facets are taking up space and are confusing.

### Proposed solution

Only create the facets if they don't already exist. This should apply automated creation of facets via operations only, because it can sometimes make sense to create the same facet twice and use it with different settings (such as creating two text filters to combine their filters).

### Alternatives considered

Status quo

### Additional context

Follow-up to #6555.

wetneb:
Fixes #6863 following the approach defined there.

A small refactoring was required, to separate out facet UI initialization from facet constructors, so that facets can be checked for duplicates before they appear onscsreen.

Not including a Cypress test because the corresponding operations are not covered in Cypress tests yet, because of their reliance on external services (there used to be a test of reconciliation which spun `reconcile-csv` locally but it was removed because it was flaky and required running the service separately for the test suite to pass).

The diffs shown by GitHub seem off, probably because of CRLF/LF differences (see #6841).

### DIFF
--- a/main/webapp/modules/core/scripts/project/process-panel.js
+++ b/main/webapp/modules/core/scripts/project/process-panel.js
@@ -230,7 +230,8 @@ ProcessPanel.prototype._perform = function(jobs) {
         ui.browsingEngine.addFacet(
             job.facetType,
             job.facetConfig,
-            job.facetOptions
+            job.facetOptions,
+            true
         );
       } catch (e) {
         //
--- a/main/webapp/modules/core/scripts/facets/facet.js
+++ b/main/webapp/modules/core/scripts/facets/facet.js
@@ -32,14 +32,21 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
 class Facet {
+  // Facet constructors shouldn't make any changes to the DOM yet.
+  // Any such effects should go in prepareUI(), which is called
+  // after adding the facet by the BrowsingEngine.
   constructor(div, config, options) {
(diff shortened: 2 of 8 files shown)


main/webapp/modules/core/scripts/project/process-panel.js
```diff
@@ -230,7 +230,8 @@ ProcessPanel.prototype._perform = function(jobs) {
         ui.browsingEngine.addFacet(
             job.facetType,
             job.facetConfig,
-            job.facetOptions
+            job.facetOptions,
+            true
```
tfmorris, line 234: The connection of this to reconciliation is totally non-obvious to me, but I'm going to assume that this is correct.

tfmorris (approved):

main/webapp/modules/core/scripts/project/process-panel.js
wetneb: There are only two operations (in core and bundled extensions) which make use of the `createFacet` action to be run after the operation completes: reconciliation and Wikibase upload. This part of the code handles such an action, and by adding this `true` boolean we are making sure that facets are deduplicated before being added.

PR #6873, opened 2024-09-24
Revert "Bump eslint in /main/tests/cypress in the cypress group (#6866)"

tfmorris:
No issue. Attempt to fix build failure.

This reverts commit b96ae694b66209417c23abf4fa471892f3781ce4.

It doesn't seem like it should be related, but see if his fixes the snapshot build failure during Javadoc generation.

### DIFF
--- a/main/tests/cypress/package.json
+++ b/main/tests/cypress/package.json
@@ -20,7 +20,7 @@
         "cypress-file-upload": "^5.0.8"
     },
     "devDependencies": {
-        "eslint": "^9.11.0",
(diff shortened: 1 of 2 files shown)


tfmorris:
@wetneb Any idea why Javadoc generation errors (which I would have expected to be warnings) are suddenly causing snapshot build failures?

wetneb:
I think that's indeed a discrepancy between the CI job that runs on PRs and the snapshot release. It would be worth unifying indeed (but I think the discrepancy is not new, it just doesn't come up so often).

Issue #6874, opened 2024-09-24
Github snapshot build action failing on Javadoc generation


tfmorris:
I don't understand what changed, but it looks like what used to be Javadoc warnings are now being treated as errors, failing the build. The Maven javadoc plugin version is unchanged, so it must be a config setting (perhaps a Github default?) which has changed.

Last successful build (2 days ago): https://github.com/OpenRefine/OpenRefine/actions/runs/10983260272/job/30492432424
Recent failing build: https://github.com/OpenRefine/OpenRefine/actions/runs/11006936008/job/30562016644

@wetneb Any idea what might have changed?

tfmorris:
@wetneb any idea what specifically might have changed with the snapshot build? It worked two days ago, but doesn't work now. I haven't changed anything that I'm aware of.

wetneb:
What has changed is commit https://github.com/OpenRefine/OpenRefine/commit/0998923c818f5480acf0dcee57d1828aa70cf789, which introduced a Javadoc comment which fails to compile. So I have opened #6875 to fix the build.

tfmorris:
Thanks! An extra pair of eyes definitely helps. I missed that the build for this merge was the first one that was failing and was looking at the dependency updates.

wetneb:
Fixed by #6875. To make it easier to debug those sorts of cases, I have opened #7018.

PR #6875, opened 2024-09-24
Fix Javadoc error

wetneb:
For #6874.

This only fixes the linting error, not the broader fact that the PR workflow doesn't test javadoc generation and the snapshot release pipeline relies on it working.


wetneb:
Merging to fix the snapshot releases

PR #6880, opened 2024-09-28
ODS file incorrectly recognized as XLS #6877

### LINKED ISSUE #6877: ODS file incorrectly recognized as XLS
When I upload an ODS file to OpenRefine, it tries to parse it as an Excel file first.

### To Reproduce
1. First, upload an ODS file (with a filename such as `test.ods`)

### Current Results

The importer selected automatically is "Excel files" and I get the error:
```
Error during initialization - org.apache.poi.openxml4j.exceptions.ODFNotOfficeXmlFileException: The supplied data appears to be in ODF (Open Document) Format. Formats like these (eg ODS, ODP) are not supported, try Apache ODFToolkit
```
I need to manually select "Open Document Format spreadsheets (.ods)" for it to be parsed correctly

### Expected Behavior

The file is automatically detected as an ODS file and parsed correctly.

### Screenshots

image

### Versions
 - Operating System: Debian
 - Browser Version: Firefox
 - JRE or JDK Version: 21
 - OpenRefine: 3.9-SNAPSHOT (master branch)

### Datasets
[test.ods](https://github.com/user-attachments/files/17142008/test.ods)

sunilnatraj:
Identifying file format from registered file types as format guessers support text based files.

Fixes #6877 

- Set seed file format guessed from file using ImportingManger.getFormatFromFileName
- Added implementation in BinaryFormatGuesser
- Registered BinaryFormatGuesser 
- Added test case for ODS file format
-

### DIFF
--- a/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
+++ b/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
@@ -426,6 +426,8 @@ public class ImportingUtilities {
                     long fileSize = fileItem.getSize();
 
                     File file = allocateFile(rawDataDir, fileName);
+                    String bestFormat = ImportingManager.extensionToFormat.get(file.getName().substring(file.getName().indexOf('.')));
+                    bestFormat = bestFormat == null ? "text" : bestFormat;
 
                     ObjectNode fileRecord = ParsingUtilities.mapper.createObjectNode();
                     JSONUtilities.safePut(fileRecord, "origin", "upload");
@@ -439,7 +441,7 @@ public class ImportingUtilities {
                             calculateProgressPercent(update.totalExpectedSize, update.totalRetrievedSize));
 
                     JSONUtilities.safePut(fileRecord, "size", saveStreamToFile(stream, file, null));
-                    JSONUtilities.safePut(fileRecord, "format", guessBetterFormat(file, request.getCharacterEncoding(), "text"));
+                    JSONUtilities.safePut(fileRecord, "format", guessBetterFormat(file, request.getCharacterEncoding(), bestFormat));
                     // TODO: This needs to be refactored to be able to test import from archives
                     if (postProcessRetrievedFile(rawDataDir, file, fileRecord, fileRecords, progress)) {
                         archiveCount++;
--- a/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
+++ b/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
@@ -426,8 +426,6 @@ public class ImportingUtilities {
                     long fileSize = fileItem.getSize();
 
                     File file = allocateFile(rawDataDir, fileName);
-                    String bestFormat = ImportingManager.getFormatFromFileName(file.getName());
-                    bestFormat = bestFormat == null ? "text" : bestFormat;
 
                     ObjectNode fileRecord = ParsingUtilities.mapper.createObjectNode();
                     JSONUtilities.safePut(fileRecord, "origin", "upload");
@@ -441,7 +439,6 @@ public class ImportingUtilities {
                             calculateProgressPercent(update.totalExpectedSize, update.totalRetrievedSize));
 
                     JSONUtilities.safePut(fileRecord, "size", saveStreamToFile(stream, file, null));
-                    JSONUtilities.safePut(fileRecord, "format", guessBetterFormat(file, request.getCharacterEncoding(), bestFormat));
                     // TODO: This needs to be refactored to be able to test import from archives
                     if (postProcessRetrievedFile(rawDataDir, file, fileRecord, fileRecords, progress)) {
                         archiveCount++;
@@ -702,10 +699,11 @@ public class ImportingUtilities {
 
     static public void postProcessSingleRetrievedFile(File file, ObjectNode fileRecord) {
         if (!fileRecord.has("format")) {
-            JSONUtilities.safePut(fileRecord, "format",
-                    ImportingManager.getFormat(
-                            file.getName(),
-                            JSONUtilities.getString(fileRecord, "declaredMimeType", null)));
+            String encoding = JSONUtilities.getString(fileRecord, "declaredEncoding", null);
+            String bestFormat = ImportingManager.getFormat(file.getName(), JSONUtilities.getString(fileRecord, "declaredMimeType", null));
+            bestFormat = bestFormat == null ? "text" : bestFormat;
+            String format = guessBetterFormat(file, encoding, bestFormat);
+            JSONUtilities.safePut(fileRecord, "format", format);
         }
     }
 
--- a/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
+++ b/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
@@ -308,7 +308,7 @@ public class ImportingUtilities {
                             calculateProgressPercent(update.totalExpectedSize, update.totalRetrievedSize));
 
                     JSONUtilities.safePut(fileRecord, "size", saveStreamToFile(stream, file, null));
-                    JSONUtilities.safePut(fileRecord, "format", guessBetterFprmat(file, fileRecord));
+                    JSONUtilities.safePut(fileRecord, "format", guessBetterFormat(file, fileRecord));
                     JSONUtilities.append(fileRecords, fileRecord);
 
                     clipboardCount++;
@@ -699,7 +699,7 @@ public class ImportingUtilities {
 
     static public void postProcessSingleRetrievedFile(File file, ObjectNode fileRecord) {
         if (!fileRecord.has("format")) {
-            JSONUtilities.safePut(fileRecord, "format", guessBetterFprmat(file, fileRecord));
+            JSONUtilities.safePut(fileRecord, "format", guessBetterFormat(file, fileRecord));
         }
     }
 
@@ -1044,7 +1044,7 @@ public class ImportingUtilities {
         return bestFormat;
     }
 
-    static String guessBetterFprmat(File file, ObjectNode fileRecord) {
+    static String guessBetterFormat(File file, ObjectNode fileRecord) {
         String encoding = JSONUtilities.getString(fileRecord, "declaredEncoding", null);
         String bestFormat = ImportingManager.getFormat(file.getName(), JSONUtilities.getString(fileRecord, "declaredMimeType", null));
         bestFormat = bestFormat == null ? "text" : bestFormat;
--- a/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
+++ b/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
@@ -426,7 +426,7 @@ public class ImportingUtilities {
                     long fileSize = fileItem.getSize();
 
                     File file = allocateFile(rawDataDir, fileName);
-                    String bestFormat = ImportingManager.extensionToFormat.get(file.getName().substring(file.getName().indexOf('.')));
+                    String bestFormat = ImportingManager.getFormatFromFileName(file.getName());
                     bestFormat = bestFormat == null ? "text" : bestFormat;
 
                     ObjectNode fileRecord = ParsingUtilities.mapper.createObjectNode();
--- a/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
+++ b/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
@@ -301,7 +301,6 @@ public class ImportingUtilities {
                     JSONUtilities.safePut(fileRecord, "origin", "clipboard");
                     JSONUtilities.safePut(fileRecord, "declaredEncoding", encoding);
                     JSONUtilities.safePut(fileRecord, "declaredMimeType", (String) null);
-                    JSONUtilities.safePut(fileRecord, "format", "text");
                     JSONUtilities.safePut(fileRecord, "fileName", "(clipboard)");
                     JSONUtilities.safePut(fileRecord, "location", getRelativePath(file, rawDataDir));
 
@@ -309,6 +308,7 @@ public class ImportingUtilities {
                             calculateProgressPercent(update.totalExpectedSize, update.totalRetrievedSize));
 
                     JSONUtilities.safePut(fileRecord, "size", saveStreamToFile(stream, file, null));
+                    JSONUtilities.safePut(fileRecord, "format", guessBetterFprmat(file, fileRecord));
                     JSONUtilities.append(fileRecords, fileRecord);
 
                     clipboardCount++;
@@ -699,11 +699,7 @@ public class ImportingUtilities {
 
     static public void postProcessSingleRetrievedFile(File file, ObjectNode fileRecord) {
         if (!fileRecord.has("format")) {
-            String encoding = JSONUtilities.getString(fileRecord, "declaredEncoding", null);
-            String bestFormat = ImportingManager.getFormat(file.getName(), JSONUtilities.getString(fileRecord, "declaredMimeType", null));
-            bestFormat = bestFormat == null ? "text" : bestFormat;
-            String format = guessBetterFormat(file, encoding, bestFormat);
-            JSONUtilities.safePut(fileRecord, "format", format);
+            JSONUtilities.safePut(fileRecord, "format", guessBetterFprmat(file, fileRecord));
         }
     }
 
@@ -1046,6 +1042,13 @@ public class ImportingUtilities {
 
         }
         return bestFormat;
+    }
+
+    static String guessBetterFprmat(File file, ObjectNode fileRecord) {
+        String encoding = JSONUtilities.getString(fileRecord, "declaredEncoding", null);
+        String bestFormat = ImportingManager.getFormat(file.getName(), JSONUtilities.getString(fileRecord, "declaredMimeType", null));
+        bestFormat = bestFormat == null ? "text" : bestFormat;
+        return guessBetterFormat(file, encoding, bestFormat);
     }
 
     static void rankFormats(ImportingJob job, final String bestFormat, ArrayNode rankedFormats) {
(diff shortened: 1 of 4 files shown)


sunilnatraj:
@wetneb @tfmorris Can you review this PR

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -426,6 +426,8 @@ public String handleResponse(final ClassicHttpResponse response) throws IOExcept
                     long fileSize = fileItem.getSize();
 
                     File file = allocateFile(rawDataDir, fileName);
+                    String bestFormat = ImportingManager.extensionToFormat.get(file.getName().substring(file.getName().indexOf('.')));
```
wetneb, line 429: This will fail for filenames which don't have any `.` in them. Instead of re-implementing this, we should rather go through `ImportingManager.getFormat` which already contains the required logic. See my comment above.

wetneb (changes requested):

It looks like we already have some logic to auto-detect the format from the extension, as `ImportingManager.getFormat(String fileName, String mimeType)`. This seems to be called only later on in `postProcessSingleRetrievedFile` and only if no format is defined yet. Since a blanket format ("binary") is defined earlier, this logic does not get executed. Your PR adds detection by extension earlier, meaning that it now gets done in two places. Intuitively, it should only be done in one place which works.

This could be done by either fixing the logic in `postProcessSingleRetrievedFile` so that it gets executed even if there is a pre-defined blanket format. Or implementing extension-based format guessing in a `FormatGuesser` instance, which would look architecturally cleaner to me. It should let us delete the `postProcessSingleRetrievedFile` method altogether. What do you think?

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
sunilnatraj: Updated to use ImportingManger.getFormatFromFileName method

sunilnatraj:
@wetneb Thanks for the feedback. I have added implementation in the BinaryFormatGuesser.

wetneb (changes requested):
Thanks for your changes! It looks like my comments were not clear enough. Let me try to clarify.

By suggesting to implement a `FormatGuesser`, I was not asking that you implement magic number sniffing. It would be useful to add such a feature, but it's unrelated to the issue at hand. In the problem we are trying to solve, detecting the file type from the extension should be enough. So, let's try to keep changes made in this PR focused on that.

On top of this, I brought up another problem with your implementation, which you haven't solved: the fact that after your changes, format guessing based on file extension is done in two places:
* the one you have introduced, which works
* in `postProcessSingleRetrievedFile`, which does not work

I would rather have this logic only in one place, which works. By suggesting a `FormatGuesser`, my intention was that a `FormatGuesser` would be an appropriate location for the *unique* implementation of this detection logic, based on the file extension.

Is this any clearer?

sunilnatraj:
@wetneb Regarding the point that "in postProcessSingleRetrievedFile, which does not work", I did check the flow of file(s) import flow and as the format is set in retrieveContentFromPostRequest flow, the flow in postProcessSingleRetrievedFile is skipped. Am i missing anything on this point?

wetneb:
> the flow in postProcessSingleRetrievedFile is skipped

That's correct, and I would say that's the root cause of #6877.

sunilnatraj:
@wetneb Per my review the root cause for issue 6877 

The file format identification relies on the client setting the format based on the file extension, when a file whose extension has not been defined is processed the assumption is it is a text based file, there are 2 file format guessers available that attempt to identify the format, as part of the format identification flow the ODS file is identified as a binary format because the file signature is the same as archive file.

wetneb:
@sunilnatraj did you have a chance to have a look at [my suggestions above](https://github.com/OpenRefine/OpenRefine/pull/6880#pullrequestreview-2344735033)? It does not look like you have addressed them so far.

sunilnatraj:
@wetneb I have made changes as suggested. moved the file format identification in the postProcessSingleRetrievedFile method. Please review.

sunilnatraj:
@wetneb Can you please review

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -439,7 +439,6 @@ public String handleResponse(final ClassicHttpResponse response) throws IOExcept
                             calculateProgressPercent(update.totalExpectedSize, update.totalRetrievedSize));
 
                     JSONUtilities.safePut(fileRecord, "size", saveStreamToFile(stream, file, null));
-                    JSONUtilities.safePut(fileRecord, "format", guessBetterFormat(file, request.getCharacterEncoding(), "text"));
```
wetneb, line 442: This is not the only place where a `"format"` is set in this method (see the cases above for `"clipboard"` and `"download"` cases). So by removing it only in this particular case we are introducing a discrepancy.

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -700,10 +699,11 @@ static public boolean postProcessRetrievedFile(
...
         if (!fileRecord.has("format")) {
-            JSONUtilities.safePut(fileRecord, "format",
-                    ImportingManager.getFormat(
-                            file.getName(),
-                            JSONUtilities.getString(fileRecord, "declaredMimeType", null)));
+            String encoding = JSONUtilities.getString(fileRecord, "declaredEncoding", null);
+            String bestFormat = ImportingManager.getFormat(file.getName(), JSONUtilities.getString(fileRecord, "declaredMimeType", null));
+            bestFormat = bestFormat == null ? "text" : bestFormat;
+            String format = guessBetterFormat(file, encoding, bestFormat);
+            JSONUtilities.safePut(fileRecord, "format", format);
```
wetneb, line 706: I would rather move this logic to a `FormatGuesser` which would be called consistently alongside other format-guessing heuristics.

wetneb (changes requested):
Thanks, that is better.

I think it would be cleaner to move this detection to a `FormatGuesser` instead, see the reasons below.

tfmorris:
Hi @sunilnatraj. Sorry for the delayed review, but I've been traveling. Something which I think is confounding your testing is that the various formats, extensions, guessers, etc which are initialized in the Javascript `controller.js` file aren't being setup in your test setup. You can probably just set up the ones you need, but here's the complete set (currently):

```java

+        // Initialization cloned from controller.js
+        ImportingManager.registerFormat("text", "core-import-formats/text" ); // generic format, no parser to handle it
+        ImportingManager.registerFormat("text/line-based", "core-import-formats/text/line-based", "LineBasedParserUI",
+                new com.google.refine.importers.LineBasedImporter());
+        ImportingManager.registerFormat("text/line-based/*sv", "core-import-formats/text/line-based/*sv", "SeparatorBasedParserUI",
+                new com.google.refine.importers.SeparatorBasedImporter());
+        ImportingManager.registerFormat("text/line-based/fixed-width", "core-import-formats/text/line-based/fixed-width", "FixedWidthParserUI",
+                new com.google.refine.importers.FixedWidthImporter());
+
+        ImportingManager.registerFormat("text/rdf/nt", "core-import-formats/text/rdf/nt", "RdfTriplesParserUI",
+                new com.google.refine.importers.RdfTripleImporter(com.google.refine.importers.RdfTripleImporter.Mode.NT));
+        ImportingManager.registerFormat("text/rdf/n3", "core-import-formats/text/rdf/n3", "RdfTriplesParserUI",
+                new com.google.refine.importers.RdfTripleImporter(com.google.refine.importers.RdfTripleImporter.Mode.N3));
+        ImportingManager.registerFormat("text/rdf/ttl", "core-import-formats/text/rdf/ttl", "RdfTriplesParserUI",
+                new com.google.refine.importers.RdfTripleImporter(com.google.refine.importers.RdfTripleImporter.Mode.TTL));
+        ImportingManager.registerFormat("text/rdf/xml", "core-import-formats/text/rdf/xml", "RdfTriplesParserUI", new com.google.refine.importers.RdfXmlTripleImporter());
+        ImportingManager.registerFormat("text/rdf/ld+json", "core-import-formats/text/rdf/ld+json", "RdfTriplesParserUI", new com.google.refine.importers.RdfJsonldTripleImporter());
+
+        ImportingManager.registerFormat("text/xml", "core-import-formats/text/xml", "XmlParserUI", new com.google.refine.importers.XmlImporter());
+        ImportingManager.registerFormat("binary/text/xml/xls/xlsx", "core-import-formats/binary/text/xml/xls/xlsx", "ExcelParserUI", new com.google.refine.importers.ExcelImporter());
+        ImportingManager.registerFormat("text/xml/ods", "core-import-formats/text/xml/ods", "ExcelParserUI", new com.google.refine.importers.OdsImporter());
+        ImportingManager.registerFormat("text/json", "core-import-formats/text/json", "JsonParserUI", new com.google.refine.importers.JsonImporter());
+        ImportingManager.registerFormat("text/marc", "core-import-formats/text/marc", "XmlParserUI", new com.google.refine.importers.MarcImporter());
+        ImportingManager.registerFormat("text/wiki", "core-import-formats/text/wiki", "WikitextParserUI", new com.google.refine.importers.WikitextImporter());
+
+        ImportingManager.registerFormat("binary", "core-import-formats/binary"); // generic format, no parser to handle it
+
+        ImportingManager.registerFormat("service", "core-import-formats/service"); // generic format, no parser to handle it
+
+        /*
+         *  Extension to format mappings
+         */
+        ImportingManager.registerExtension(".txt", "text");
+        ImportingManager.registerExtension(".csv", "text/line-based/*sv");
+        ImportingManager.registerExtension(".tsv", "text/line-based/*sv");
+
+        ImportingManager.registerExtension(".xml", "text/xml");
+        ImportingManager.registerExtension(".atom", "text/xml");
+
+        ImportingManager.registerExtension(".json", "text/json");
+        ImportingManager.registerExtension(".js", "text/json");
+
+        ImportingManager.registerExtension(".xls", "binary/text/xml/xls/xlsx");
+        ImportingManager.registerExtension(".xlsx", "binary/text/xml/xls/xlsx");
+
+        ImportingManager.registerExtension(".ods", "text/xml/ods");
+
+        ImportingManager.registerExtension(".nt", "text/rdf/nt");
+        ImportingManager.registerExtension(".ntriples", "text/rdf/nt");
+
+        ImportingManager.registerExtension(".n3", "text/rdf/n3");
+        ImportingManager.registerExtension(".ttl", "text/rdf/ttl");
+        ImportingManager.registerExtension(".jsonld", "text/rdf/ld+json");
+        ImportingManager.registerExtension(".rdf", "text/rdf/xml");
+
+        ImportingManager.registerExtension(".marc", "text/marc");
+        ImportingManager.registerExtension(".mrc", "text/marc");
+
+        ImportingManager.registerExtension(".wiki", "text/wiki");

+
+        /*
+         *  Mime type to format mappings
+         */
+        ImportingManager.registerMimeType("text/plain", "text");
+        ImportingManager.registerMimeType("text/csv", "text/line-based/*sv");
+        ImportingManager.registerMimeType("text/x-csv", "text/line-based/*sv");
+        ImportingManager.registerMimeType("text/tab-separated-value", "text/line-based/*sv");
+        ImportingManager.registerMimeType("text/tab-separated-values", "text/line-based/*sv");
+
+        ImportingManager.registerMimeType("text/fixed-width", "text/line-based/fixed-width");
+
+        ImportingManager.registerMimeType("application/n-triples", "text/rdf/nt");
+        ImportingManager.registerMimeType("text/n3", "text/rdf/n3");
+        ImportingManager.registerMimeType("text/rdf+n3", "text/rdf/n3");
+        ImportingManager.registerMimeType("text/turtle", "text/rdf/ttl");
+        ImportingManager.registerMimeType("application/xml", "text/xml");
+        ImportingManager.registerMimeType("text/xml", "text/xml");
+        ImportingManager.registerMimeType("+xml", "text/xml"); // suffix will be tried only as fallback
+        ImportingManager.registerMimeType("application/rdf+xml", "text/rdf/xml");
+        ImportingManager.registerMimeType("application/ld+json", "text/rdf/ld+json");
+        ImportingManager.registerMimeType("application/atom+xml", "text/xml");
+
+        ImportingManager.registerMimeType("application/msexcel", "binary/text/xml/xls/xlsx");
+        ImportingManager.registerMimeType("application/x-msexcel", "binary/text/xml/xls/xlsx");
+        ImportingManager.registerMimeType("application/x-ms-excel", "binary/text/xml/xls/xlsx");
+        ImportingManager.registerMimeType("application/vnd.ms-excel", "binary/text/xml/xls/xlsx");
+        ImportingManager.registerMimeType("application/x-excel", "binary/text/xml/xls/xlsx");
+        ImportingManager.registerMimeType("application/xls", "binary/text/xml/xls/xlsx");
+        ImportingManager.registerMimeType("application/x-xls", "binary/text/xml/xls/xlsx");
+        ImportingManager.registerMimeType("application/vnd.openxmlformats-officedocument.spreadsheetml.sheet", "binary/text/xml/xls/xlsx");
+        ImportingManager.registerMimeType("application/vnd.openxmlformats-officedocument.spreadsheetml.template", "binary/text/xml/xls/xlsx");
+
+        ImportingManager.registerMimeType("application/vnd.oasis.opendocument.spreadsheet","text/xml/ods");
+
+        ImportingManager.registerMimeType("application/json", "text/json");
+        ImportingManager.registerMimeType("application/javascript", "text/json");
+        ImportingManager.registerMimeType("text/json", "text/json");
+        ImportingManager.registerMimeType("+json", "text/json"); // suffix will be tried only as fallback
+
+        ImportingManager.registerMimeType("application/marc", "text/marc");
```

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
sunilnatraj: @wetneb Are you suggesting moving the logic to an overloaded guessBetterFormat method?

wetneb: Yes, that would be intuitively cleaner to me

sunilnatraj: Code updated; please review

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
sunilnatraj: Flow updated - format identification using the newly added method is used for clipboard, url/download and files in the import flow.

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
wetneb: Sorry, it looks like there was a misunderstanding again. I wrote above "Ideally in a new FormatGuesser implementation, to keep it separate from the existing ones which try to guess types via other means."

By this I expected that you would create a new class (say `com.google.refino.importers.ExtensionFormatGuesser`), in a new file. This class would `implement FormatGuesser`, and overload the only method offered by this interface: `public String guess(File file, String encoding, String seedFormat);`

It could look like this:
```java

public class ExtensionFormatGuesser implements FormatGuesser {
     public String guess(File file, String encoding, String seedFormat) {
          // ...
    }
}
```

sunilnatraj: @wetneb FormatGuessers are defined for a format class, each formatter attenpt to guess a more correct format by sampling data. In the issue being addressed the file format is not available as the the client is not passing this info, the proposed solution addresses this by identifying the format from the pre-defined dataset of file extension to format, so it is not tied to a file format, which is why adding this in a FormatGuesser did not seem the right approach. 

Adding a format guesser for this would be something like this; does this seem right?

`public class ExtensionFormatGuesser implements FormatGuesser {

    @Override
    public String guess(File file, String encoding, String seedFormat) {
        String betterFormat = ImportingManager.getFormat(file.getName(), null);
        return betterFormat != null ? betterFormat : seedFormat ;
    }
}`

`    static String guessBetterFprmat(File file, ObjectNode fileRecord) {
        String encoding = JSONUtilities.getString(fileRecord, "declaredEncoding", null);
        FormatGuesser guesser = new ExtensionFormatGuesser();
        String betterFormat = guesser.guess(file, encoding, "text");
        return guessBetterFormat(file, encoding, betterFormat);
    }`

sunilnatraj: @wetneb Can you review my previous note and let know next steps.

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -1048,6 +1044,13 @@ static String guessBetterFormat(File file, String fileEncoding, String bestForma
         return bestFormat;
     }
 
+    static String guessBetterFprmat(File file, ObjectNode fileRecord) {
```
wetneb, line 1047: Typo: `Fprmat` should be `Format`

wetneb (commented):
Ok, that makes sense, let's keep it as it is now. There is just one typo to fix before merging.

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
sunilnatraj: Fixed

wetneb (approved):
Thanks a lot for your perseverance!

wetneb:
@sunilnatraj it's your fifth PR and you've clearly put a lot of effort into them - thanks again!
I'd be interested to know what motivates you to work on OpenRefine. Are you using it yourself for some projects perhaps?
Is there anything we could do to make your work easier?

sunilnatraj:
@wetneb Firstly thanks to you for the support & guidance. I wanted to work on open source projects, I was browsing thru goodfirstissue.dev that is how i got to open refine. I am beginning to like the work. Should i join a forum that will help interact / contribute with you & team.

wetneb:
Great! yes, you're welcome to join https://forum.openrefine.org/. We have an [introductions thread](https://forum.openrefine.org/t/introductions-thread/69) if you want to present yourself there. There are also some threads which should you give you a sense of the most acute user needs:
* https://forum.openrefine.org/t/results-from-the-feature-prioritization-survey-2024/1847
* https://forum.openrefine.org/t/prioritisation-of-wikimedia-sveriges-development-work/1803

Many of the issues mentioned there are probably quite hard to implement but maybe you find things to your taste.

PR #6881, opened 2024-09-28
Refetch reconcile service manifest for new changes (#3645)

### LINKED ISSUE #3645: Double check the reconciliation manifest each time reconciliation is fired.

akashinde:
Fixes #3645 

- `ReconciliationManager.refetchServiceManifest` function has been created which can be used to refetch reconcile service  manifest provided service object as input.
- Then the function has been used in `ReconStandardServicePanel` to fetch the latest manifest of selected service.

Steps taken before commit:
- [x] run lint, clean, build
- [x] run cypress tests (PASSED)
- [x] run java tests (PASSED)

Final result:
[Before]
Screenshot_2024-09-28_20-44-12
[After manifest is modified (adding preview) and restarted, then on type-guessing panel]
Screenshot_2024-09-28_20-42-12

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
+++ b/main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
@@ -222,12 +222,20 @@ ReconDialog.prototype._onAddStandardService = function() {
  */
 ReconDialog.prototype._refetchServiceManifest = function(service, f) {
   var dismissBusy = DialogSystem.showBusy($.i18n('core-recon/contact-service')+"...");
-  $.ajax({ url: service.url })
+  // First, try with CORS (default "json" dataType)
+  $.ajax({ url: service.url, dataType: 'json', timeout: 5000 })
     .done(function() {
       ReconciliationManager.reRegisterService(service, f, true);
     })
     .fail(function(jqXHR, textStatus, errorThrown) {
-      alert($.i18n('core-recon/error-contact')+': ' + textStatus + ' : ' + errorThrown + ' - ' + service.url);
+      // If it fails, try with JSONP
+      $.ajax({ url: service.url, dataType: 'jsonp', timeout: 5000 })
+        .done(function() {
+          ReconciliationManager.reRegisterService(service, f, true);
+        })
+        .fail(function(jqXHR, textStatus, errorThrown) {
+          alert($.i18n('core-recon/error-contact')+': ' + textStatus + ' : ' + errorThrown + ' - ' + service.url);
+        })
     })
     .always(function() { dismissBusy(); });
 };
--- a/main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
+++ b/main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
@@ -60,7 +60,10 @@ ReconDialog.prototype._createDialog = function() {
   this._elmts.cancelButton.on('click',function() { self._dismiss(); });
   this._elmts.nextButton.on('click',function() { 
     if(self._record){
-    self._nextDialog(self._column, self._selectedServiceRecordIndex, self._serviceRecords, self._record, self.previousRecord);
+      // refetch selected service manifest then navigate to next dialog on success
+      self._refetchServiceManifest(self._record.service, function() {
+        self._nextDialog(self._column, self._selectedServiceRecordIndex, self._serviceRecords, self._record, self.previousRecord);
+      });
     }
     else{
       var message = document.getElementById('popup-message');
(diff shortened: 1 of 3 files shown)


wetneb (changes requested):
Thank you for working on this!

I am a bit worried by the control flow of your implementation. We need to anticipate that re-fetching the manifest of the service will take some time. Given that the code directly continues with guessing types, without waiting for this re-fetching to complete, I imagine this could cause some problems:
* the guessing of types will be done without taking into account any new information from the re-fetching. At the moment this operation doesn't rely on anything from the manifest (I think) beyond the URL of the endpoint, but in the future that could change and the URL of the endpoint itself could ideally be updated by the re-fetching (following redirections, for instance).
* if the service is unreachable, we'll probably get two different errors popping up as `alert()`: one from the type-guessing, one from the re-fetching. Also, because you are unregistering the service first, the service will disappear from the list of services, even if it was only temporarily unavailable, which is not ideal

So I would be more inclined to adopt a different control-flow, where we wait for the manifest-refetching to complete before we proceed to the type guessing.

akashinde:
Apologies for late response! Thank you for the review.

Yes, that makes sense. For this scenario my proposal would be as follow:

- Okay, For the waiting part, I was thinking we can make a manifest call (with async:false) on the service list page itself. 
Then 
-1. on fail, throw alert if service is unreachable and stay on the same page
-2. on pass, we can re-register the service, and then move to next page.

- For the second case, I think with this approach we will now have separate alerts. And we can prevent removal of service by
-1. while calling the manifest (on service list page), if success then only re-register
-2. if failed then do not re-register and throw an alert

Please feel free to let me know your approach or suggestions on this! 

Also one question, for the loading part, is this component  `DialogSystem.showBusy($.i18n('core-recon/contact-service')+"...");` responsible and could be used for this scenario as well? or there's any other common loading component I can use?

wetneb:
Thanks! That sounds like a good approach to me. I'd generally try to avoid `async: false` calls so I wound tend to formulate the new control flow with callbacks still, but arguably that's more of a matter of taste.

> Also one question, for the loading part, is this component `DialogSystem.showBusy($.i18n('core-recon/contact-service')+"...");` responsible and could be used for this scenario as well? or there's any other common loading component I can use?

That looks appropriate to me. We can try with this and see if it gives a satisfactory UX.

tfmorris:
>  I'd generally try to avoid async: false calls so I wound tend to formulate the new control flow with callbacks still, but arguably that's more of a matter of taste.

I think it's a best practice, not just a matter of taste. Many browsers and performance tools will warn about synchronous calls blocking other activities.

akashinde:
Thank you @wetneb for your feedback! I understand your point about preferring callbacks over async: false calls. I'll make sure to refactor the control flow using callbacks for better flexibility. Regarding the loading component, I'll proceed with using `DialogSystem.showBusy($.i18n('core-recon/contact-service')+"...")` for now, as you suggested, and we can assess how it impacts the overall UX.

Thank you @tfmorris for pointing that out! I appreciate the clarification. I agree that avoiding synchronous calls is indeed best practice, especially with the performance implications you've highlighted. I'll ensure the control flow is updated accordingly.

Thanks again for your guidance!

akashinde:
Hello,

I have pushed new changes. It's much better now thanks to all suggestions. Also please feel free to remove docs, if not relevant, that I have added above function definitions.

Here's result screenshot for reference:
image

main/webapp/modules/core/scripts/reconciliation/recon-dialog.js
```diff
@@ -211,4 +214,20 @@ ReconDialog.prototype._onAddStandardService = function() {
...
-
+/**
+ * Refetches the manifest for a reconciliation service and re-registers it.
+ * 
+ * @param {Object} service - The reconciliation service object
+ * @param {Function} [f] - Optional callback function to execute after re-registering
+ */
+ReconDialog.prototype._refetchServiceManifest = function(service, f) {
+  var dismissBusy = DialogSystem.showBusy($.i18n('core-recon/contact-service')+"...");
+  $.ajax({ url: service.url })
```
wetneb, line 225: We currently let users add reconciliation services via two means:
* normal requests, like the one you are doing here, which succeed if the service has set the appropriate CORS headers
* via the legacy JSONP method (by adding a `callback` parameter to the GET URL)

I think this refetching will fail if the service was added via JSONP, so perhaps this needs refactoring so this request is done by the ReconciliationManager already (which already handles those two cases).

Eventually we should drop JSONP support, but there are still a few publicly available services which are only accessible via JSONP: https://reconciliation-api.github.io/testbench/

akashinde:
Thanks @wetneb  for the review. You are correct, I missed this use case. If I understand correctly, then we just have to make two calls (one with cors and one with legacy JSONP) like we do while registering the service. I have pushed the changes and below attached some result that I tested.

Result:
image

I added two logs to check the flow which shows that it's calling inner ajax on fail of outer ajax. Let me know if this approach is good enough!

wetneb (commented):
Sorry about the delay in reviewing this. I tested it and it works great as a user.
But we are now re-fetching the manifest twice: once in the body of `_refetchServiceManifest`, and then a second time in `reRegisterService`, which doesn't sound ideal. Intuitively, once we have re-fetched the manifest once, we should be able to add it to the `ReconciliationManager` directly without re-fetching it. And ideally, we should avoid duplicating the logic of `registerStandardService` in `_refetchServiceManifest`: those two subsequent calls (first CORS, second JSONP) should be defined only once and reused where needed, be it for the initial registration of a service or its re-fetching. Does this make sense to you?

wetneb:
Hi @akashinde, do you still intend to work on this?

wetneb:
@akashinde I'm closing this for now per inactivity, but happy to reopen when you're working on it again.

akashinde:
Thanks for waiting @wetneb, I have opened new PR as per your review. https://github.com/OpenRefine/OpenRefine/pull/7317

PR #6895, opened 2024-10-09
wikibase: Render property names in issue reports #6768

### LINKED ISSUE #6768: wikibase: Render property names in issue reports
Sometimes OpenRefine generates Wikibase issues of this form:

> **[male population (P1540)](http://www.wikidata.org/entity/P1540) is missing a [P585](http://www.wikidata.org/entity/P585) qualifier.**
> Statements using [male population (P1540)](http://www.wikidata.org/entity/P1540) such as the one on [Katowice (Q588)](http://www.wikidata.org/entity/Q588) are missing a mandatory [P585](http://www.wikidata.org/entity/P585) qualifier.

Note the fact that "P585" is written without the corresponding name ("point in time" as of today, in English).
This happens for properties which are mentioned in constraints but are not present in the candidate edits or project data.
For instance, the "male population (P1540)" property was input in the schema, so OpenRefine knows its label without having to fetch it from Wikidata, which is why it is already included in the generated issue.

### Proposed solution

It would be much more user-friendly if those properties could also be rendered with the label.
This could be done either in the backend (by fetching the names for the missing properties when generating the issues) or on the frontend-side (by making an API call to the Wikibase instance to fetch them when rendering). In both cases, it would be good to cache this information for the duration of the work session (such as, the lifetime of the OpenRefine web interface) to avoid re-fetching those too often from the Wikibase instance.

### Alternatives considered

Status quo

### Additional context

Follow-up from [the OpenRefine-Wikidata training at Wikimania](https://wikimania.eventyay.com/2024/talk/PBMMEJ/)

sunilnatraj:
Fixes #6768 

- Scrutinizers to provide information of warning keys that hold property keys that need label augmentation
- Updated EditInspector to inspect and update Warnings generated, load property details into cache and update Warning detail with property label
- Property label is set according to the locale setting


sunilnatraj:
@wetneb Can you please review this PR

extensions/wikibase/tests/src/org/openrefine/wikibase/commands/PreviewWikibaseWarningTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import com.google.refine.model.Project;
+import com.google.refine.util.ParsingUtilities;
+
+import org.openrefine.wikibase.testing.TestingData;
+
+public class PreviewWikibaseWarningTest extends SchemaCommandTest {
```
wetneb, line 45: Why not add your test case to the existing `PreviewWikibaseSchemaCommandTest` test class?

extensions/wikibase/tests/data/manifest/wikidata-manifest-warning-data-test.json
```diff
@@ -0,0 +1,92 @@
+{
+    "version": "1.0",
+    "mediawiki": {
+        "name": "Wikidata",
+        "root": "https://www.wikidata.org/wiki/",
```
wetneb, line 5: Do we need a new copy of the Wikidata manifest, can't we reuse the existing ones we have in this folder?

wetneb (commented):
Thank you very much!

I wonder if it's really necessary to introduce the `getWarningsRelatedPropertyKeys` method. Wouldn't it be simpler if the `EditInspector` could directly look for any entity missing a label in the generated issues, collect the list of such entity ids, fetch their labels, and add the labels to the issues? It would remove the need to keep track of which properties need fetching when writing a scrutinizer, which would be helpful I would say.

Also, there could potentially be cases where other types of entities (mostly, items) would need fetching, so it would probably be worth handling such cases too.

extensions/wikibase/tests/src/org/openrefine/wikibase/commands/PreviewWikibaseWarningTest.java
sunilnatraj: The reason to create a separate file was because the cache initialisation in PreviewWikibaseSchemaCommandTest results in failure to do property fetch.  I could not identify a solution, let me know if there is a solution.

extensions/wikibase/tests/data/manifest/wikidata-manifest-warning-data-test.json
sunilnatraj: Yes we can do that; removing the added manifest file. Thanks.

sunilnatraj:
@wetneb Thanks for the review and feedback. I have updated the code in EditInspector to review each Warning and update all values in properties that are of type PropertyIdValue. 
You also made another point "_Also, there could potentially be cases where other types of entities (mostly, items) would need fetching, so it would probably be worth handling such cases too._"  Can you share some more details on this.

extensions/wikibase/tests/src/org/openrefine/wikibase/commands/PreviewWikibaseWarningTest.java
wetneb: fine by me

wetneb:
Thank you, it looks a lot simpler like this!

> You also made another point "Also, there could potentially be cases where other types of entities (mostly, items) would need fetching, so it would probably be worth handling such cases too." Can you share some more details on this.

See for instance the ["item-requires-statement constraint"](https://www.wikidata.org/wiki/Help:Property_constraints_portal/Item#Item_2), where the constraint definition can be used to specify an expected value for the required statement. OpenRefine doesn't currently display this value, but it could:
image

In this example, it could be worth improving the error message so that it specifies that [Canada (Q16)](https://www.wikidata.org/wiki/Q16) is the expected value of the required statement. For this to work, we'd need to be able to fetch the label of Q16 (just like we are fetching the label of P17 thanks to your PR).

sunilnatraj:
@wetneb Can you review below screenshot if this suffices the requirement of including the value if defined in the constraint. I have not attempted to resolve the label of the value, the value has a hyperlink that can be used to get the actual label value. 
Screenshot 2024-10-14 at 3 04 34 PM

wetneb:
That looks beautiful! I'd expect there shouldn't be very hard to change your code so that not just `PropertyIdValue`s get their labels fetched, but also `ItemIdValue`s, no?

sunilnatraj:
@wetneb Thank you. Yes we can also fetch Item details.

sunilnatraj:
@wetneb Item label included in warning detail

Screenshot 2024-10-14 at 4 25 23 PM

sunilnatraj:
@wetneb I have committed code to include the suggested value from constraint. The suggested value check is only done for 4 warning types; if there are other warning types to include / exclude do let me know.

extensions/wikibase/module/langs/translation-en.json
```diff
@@ -109,6 +109,7 @@
     "wikibase-schema/incomplete-schema-could-not-be-saved": "Your schema is incomplete so it cannot be saved yet.",
     "wikibase-schema/unsaved-warning": "You have made unsaved changes to your Wikibase schema. Close anyway?",
     "wikibase-schema/incompatible-site-iri-reconciled-column": "This column is reconciled to the identifier space $1, but this input requires a column reconciled to $2.",
+    "wikibase-schema/expected-value": " with value set as ",
```
wetneb, line 112: It would be better to define a second version of the full message. Constructing localized messages by concatenating them can cause problems because you are assuming a specific word order which might not work in all languages.

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
```diff
@@ -215,6 +217,11 @@ public void inspect(List<EntityEdit> editBatch, WikibaseSchema schema) throws Ex
     }
 
     private void resolveWarningPropertyLabels() throws ExecutionException {
+        List<String> warningTypeswithSuggestedData = List.of("new-item-requires-property-to-have-certain-values",
+                "new-item-requires-certain-other-statement",
+                "existing-item-requires-property-to-have-certain-values",
+                "existing-item-requires-certain-other-statement");
```
wetneb, line 223: I would expect that we don't need to list the constraint types here: as discussed earlier, the later code could directly inspect the entities that are returned as part of the warnings, and add the labels if they are missing.

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
```diff
@@ -226,26 +233,50 @@ private void resolveWarningPropertyLabels() throws ExecutionException {
...
                                 PropertyDocument propertyDocument = (PropertyDocument) entityCache.get(property);
                                 if (propertyDocument != null && propertyDocument.getLabels() != null) {
                                     label = propertyDocument.getLabels().get(Locale.getDefault().getLanguage()) != null
                                             ? propertyDocument.getLabels().get(Locale.getDefault().getLanguage()).getText()
                                             : "";
                                 }
-
-                                // Update the property with a new SuggestedPropertyIdValue
                                 warning.setProperty(key, new SuggestedPropertyIdValue(property.getId(), property.getSiteIri(), label));
                             }
```
wetneb, line 248: Here would be a good place to add handling for `ItemIdValue` in addition to `PropertyIdValue`

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
```diff
@@ -226,26 +233,50 @@ private void resolveWarningPropertyLabels() throws ExecutionException {
...
+                                SuggestedItemIdValue itemIdValue = new SuggestedItemIdValue(((EntityIdValue) statement.getValue()).getId(),
+                                        current.getSiteIri(), itemLabel);
+                                SuggestedPropertyIdValue updated = new SuggestedPropertyIdValue(current.getId(), current.getSiteIri(),
+                                        current.getLabel(), itemIdValue);
+                                warning.setProperty("added_property_entity", updated);
+                            }
+                        });
+                    }
+                }
+            });
```
wetneb, line 279: I wouldn't have expected this new block of code to be necessary: I would just have added support for `ItemIdValue` alongside `PropertyIdValue` above.

extensions/wikibase/src/org/openrefine/wikibase/schema/entityvalues/SuggestedPropertyIdValue.java
```diff
@@ -26,20 +26,33 @@
...
 
 public class SuggestedPropertyIdValue extends SuggestedEntityIdValue implements PropertyIdValue {
 
+    private Value suggestedValue;
+
     public SuggestedPropertyIdValue(String id, String siteIRI, String label) {
         super(id, siteIRI, label);
     }
 
+    public SuggestedPropertyIdValue(String id, String siteIRI, String label, Value suggestedValue) {
```
wetneb, line 39: I don't think this is necessary: ItemIdValues can already be returned independently from PropertyIdValue, which is more flexible. There could be warnings where we are including an ItemIdValue without an associated PropertyIdValue, so I would prefer not to couple them here.

wetneb (changes requested):
Thank you for your hard work! Your changes took an unexpected turn. It looks like the communication between the two of us is not working very well, sorry if my comments were unclear. I am not sure how to improve though.

extensions/wikibase/module/langs/translation-en.json
sunilnatraj: @wetneb The reason a separate label was added is because the value suggestion is optional. Any suggestion on how to handle this case if we were to include the message in the full message. Will it be easier if instead of verbose string we use something like  "Quebec ... (P633) requires statement with property country (P17) **[ Expected value - Canada ]"**

sunilnatraj:
@wetneb You have been very clear in the communication, thank you. I added the reason behind the approach. LMK if i missed something obvious.

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
wetneb: Instead of adding an item id value inside a property id value, I think the scrutinizer should add two separate parameters to the warning it creates: one for the property and one for the item.

extensions/wikibase/module/langs/translation-en.json
wetneb: Define two messages for the warning text:
* one for the full text of the warning description, without suggested value
* one for the full text of the warning description, with suggested value

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
sunilnatraj: @wetneb Can you also confirm if the logic for identifying the expected value is correct?

wetneb: no, please just read the expected value from the constraints definition and pass it as a separate warning parameter (independently of the property id). This probably requires changing the warning id as well.

sunilnatraj: @wetneb  thanks for the clarification, i had ended up re-inventing the constraint flow.. 
One query, the value definition in ItemRequiresConstraint is a List, what needs to be done if there are multiple values.

extensions/wikibase/module/langs/translation-en.json
sunilnatraj: Added separate label entries for warnings where suggested value is available

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
sunilnatraj: Done

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
sunilnatraj: Removed block

extensions/wikibase/src/org/openrefine/wikibase/schema/entityvalues/SuggestedPropertyIdValue.java
sunilnatraj: Reverted changes to SuggestedIdValue

extensions/wikibase/module/scripts/warnings-renderer.js
```diff
@@ -85,8 +82,12 @@ WarningsRenderer._createFacetForWarning = function (warning) {
 };
 
 WarningsRenderer._renderWarning = function (warning, onLocateRows) {
-  var title = WarningsRenderer._replaceIssueProperties($.i18n('warnings-messages/' + warning.type + '/title'), warning.properties);
-  var body = WarningsRenderer._replaceIssueProperties($.i18n('warnings-messages/' + warning.type + '/body'), warning.properties);
+  var warningKey = warning.type;
+  if ( warning.properties && warning.properties.hasOwnProperty("item_entity") ) {
+       warningKey += "-with-suggested-value";
+       }
```
wetneb, line 88: This change should not be necessary. Passing an item inside a warning could be used for all sorts of purposes, not just for this particular example, so it's better to keep this generic. Instead, change the Scrutinizer class so that it emits a different warning id whenever it adds an `item_entity`.

extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/ItemRequiresScrutinizer.java
```diff
@@ -106,6 +107,9 @@ public void scrutinizeStatementEdit(StatementEntityEdit update) {
                     issue.setProperty("property_entity", propertyId);
                     issue.setProperty("added_property_entity", itemRequiresPid);
                     issue.setProperty("example_entity", update.getEntityId());
+                    if (constraint.itemList.size() == 1) {
+                        issue.setProperty("item_entity", (ItemIdValue) constraint.itemList.get(0));
+                    }
```
wetneb, line 112: This is the place where you can also change the warning id that you are emitting.

extensions/wikibase/module/langs/translation-en.json
```diff
@@ -205,12 +204,20 @@
...
     "warnings-messages/existing-item-requires-property-to-have-certain-values/title": "{property_entity} requires property {added_property_entity} to have certain values.",
     "warnings-messages/existing-item-requires-property-to-have-certain-values/body": "Property {added_property_entity} doesn't have appropriate values such as on item {example_entity}. If the item already has statements with property {added_property_entity} and with appropriate values in {wikibase_name}, then this warning can be ignored.",
+    "warnings-messages/existing-item-requires-property-to-have-certain-values-with-suggested-value/title": "{property_entity} requires property {added_property_entity} with value set as {item_entity}.",
+    "warnings-messages/existing-item-requires-property-to-have-certain-values-with-suggested-value/body": "Property {added_property_entity} must have value set as {item_entity} such as on item {example_entity}. If the item already has statements with property {added_property_entity} and with appropriate values in {wikibase_name}, then this warning can be ignored.",
```
wetneb, line 216: I think the current wording is a bit ambiguous because it makes me think that as a user I should imitate whatever there is on `example_entity`, but it's the other way around: the `example_entity` is an example of where the statement is missing. So I would propose to improve the message as follows:
```suggestion
    "warnings-messages/existing-item-requires-property-to-have-certain-values-with-suggested-value/body": "On entities such as {example_entity}, property {added_property_entity} must have a value set to {item_entity}. If the entity already has statements with property {added_property_entity} and with appropriate values in {wikibase_name}, then this warning can be ignored.",
```

extensions/wikibase/module/langs/translation-en.json
```diff
@@ -205,12 +204,20 @@
...
+    "warnings-messages/new-item-requires-certain-other-statement-with-suggested-value/title": "{property_entity} requires statement with property {added_property_entity} with value set as {item_entity}.",
+    "warnings-messages/new-item-requires-certain-other-statement-with-suggested-value/body": "This property is expected to have another statement with property {added_property_entity} with value set as {item_entity} such as on item {example_entity}.",
     "warnings-messages/existing-item-requires-property-to-have-certain-values/title": "{property_entity} requires property {added_property_entity} to have certain values.",
     "warnings-messages/existing-item-requires-property-to-have-certain-values/body": "Property {added_property_entity} doesn't have appropriate values such as on item {example_entity}. If the item already has statements with property {added_property_entity} and with appropriate values in {wikibase_name}, then this warning can be ignored.",
+    "warnings-messages/existing-item-requires-property-to-have-certain-values-with-suggested-value/title": "{property_entity} requires property {added_property_entity} with value set as {item_entity}.",
```
wetneb, line 215:
```suggestion
    "warnings-messages/existing-item-requires-property-to-have-certain-values-with-suggested-value/title": "{property_entity} requires property {added_property_entity} with value set to {item_entity}.",
```

extensions/wikibase/module/langs/translation-en.json
```diff
@@ -205,12 +204,20 @@
...
+    "warnings-messages/existing-item-requires-property-to-have-certain-values-with-suggested-value/body": "Property {added_property_entity} must have value set as {item_entity} such as on item {example_entity}. If the item already has statements with property {added_property_entity} and with appropriate values in {wikibase_name}, then this warning can be ignored.",
     "warnings-messages/existing-item-requires-certain-other-statement/title": "{property_entity} requires statement with property {added_property_entity}.",
     "warnings-messages/existing-item-requires-certain-other-statement/body": "This property is expected to have another statement with property {added_property_entity} such as on item {example_entity}. If the item already has statements with property {added_property_entity} in {wikibase_name}, then this warning can be ignored.",
+    "warnings-messages/existing-item-requires-certain-other-statement-with-suggested-value/title": "{property_entity} requires statement with property {added_property_entity} with value set as {item_entity}.",
```
wetneb, line 219:
```suggestion
    "warnings-messages/existing-item-requires-certain-other-statement-with-suggested-value/title": "{property_entity} requires statement with property {added_property_entity} with value set to {item_entity}.",
```

extensions/wikibase/module/langs/translation-en.json
```diff
@@ -205,12 +204,20 @@
...
+    "warnings-messages/existing-item-requires-certain-other-statement-with-suggested-value/body": "This property is expected to have another statement with property {added_property_entity} with value set as {item_entity} such as on item {example_entity}. If the item already has statements with property {added_property_entity} in {wikibase_name}, then this warning can be ignored.",
```
wetneb, line 220: Same as above:
```suggestion
    "warnings-messages/existing-item-requires-certain-other-statement-with-suggested-value/body": "Entities such as {example_entity} are expected to have another statement with property {added_property_entity} with a value set to {item_entity}. If the item already has statements with property {added_property_entity} in {wikibase_name}, then this warning can be ignored.",
```

extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/ItemRequiresScrutinizer.java
sunilnatraj: QAWarning did not have setter for type so extended logic in the existing QA warning constructor to set correct warning type

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
```diff
@@ -206,6 +211,50 @@ public void inspect(List<EntityEdit> editBatch, WikibaseSchema schema) throws Ex
...
+                                String itemLabel = "";
+                                ItemDocument itemDocument = (ItemDocument) entityCache.get(itemIdValue);
+
+                                if (itemDocument != null && itemDocument.getLabels() != null) {
+                                    itemLabel = itemDocument.getLabels().get(Locale.getDefault().getLanguage()) != null
+                                            ? itemDocument.getLabels().get(Locale.getDefault().getLanguage()).getText()
+                                            : "";
+                                }
+                                warning.setProperty(key,
+                                        new SuggestedPropertyIdValue(itemIdValue.getId(), itemIdValue.getSiteIri(), itemLabel));
```
wetneb, line 254: Very nice, we're getting there! To reduce the amount of duplicated code here, you could try to exploit the fact that both `ItemDocument` and `PropertyDocument` are specializations of `LabeledDocument`.
So you could merge the two `if` blocks by having a condition `if (value instanceof ItemIdValue || value instanceof PropertyIdValue)`, and then using `LabeledDocument` in place of `ItemDocument` and `PropertyDocument`. You'll still need to make a distinction at the end to create either a `SuggestedPropertyIdValue` or an `ItemIdValue` but I think the deduplication should still be worth it.

sunilnatraj: Updated code ; duplication removed. Please review. @wetneb

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
```diff
@@ -206,6 +211,44 @@ public void inspect(List<EntityEdit> editBatch, WikibaseSchema schema) throws Ex
...
+            resolveWarningPropertyLabels();
+        }
+    }
+
+    private void resolveWarningPropertyLabels() throws ExecutionException {
+        if (entityCache != null) {
+            List<EntityIdValue> propertyIdValues = warningStore.getWarnings().stream()
+                    .flatMap(warning -> warning.getProperties().entrySet().stream()
+                            .filter(entry -> entry.getValue() instanceof PropertyIdValue)
+                            .map(entry -> (PropertyIdValue) entry.getValue()))
```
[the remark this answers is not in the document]
sunilnatraj: Updated flow to filter out entities with label present

sunilnatraj:
@wetneb I have updated the test case, using EntityCacheStub and also removed the newly added Test class, merged the test case in the existing PreviewWikibaseSchemaCommandTest

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
```diff
@@ -220,8 +221,15 @@ private void resolveWarningPropertyLabels() throws ExecutionException {
         if (entityCache != null) {
             List<EntityIdValue> propertyIdValues = warningStore.getWarnings().stream()
                     .flatMap(warning -> warning.getProperties().entrySet().stream()
-                            .filter(entry -> entry.getValue() instanceof PropertyIdValue)
-                            .map(entry -> (PropertyIdValue) entry.getValue()))
+                            .filter(entry -> entry.getValue() instanceof EntityIdValue)
+                            .filter(entry -> {
+                                if (entry instanceof SuggestedEntityIdValue) {
+                                    String label = ((SuggestedEntityIdValue) entry.getValue()).getLabel();
+                                    return label != null && !label.isEmpty();
```
wetneb, line 228: Nice! But I am confused by this `return` statement. Shouldn't it be the negation instead? Because we only want to fetch the label if it is not present yet, right? Looking at the new files in the `data/entitycache` directory, your code is likely fetching more entities than it needs to.
```suggestion
                                    return label == null || label.isEmpty();
```

extensions/wikibase/tests/src/org/openrefine/wikibase/commands/PreviewWikibaseSchemaCommandTest.java
```diff
@@ -110,4 +120,33 @@ public void testInvalidManifest() throws IOException, ServletException {
         assertEquals(writer.toString(),
                 "{\"code\":\"error\",\"message\":\"Wikibase manifest could not be parsed. Error message: invalid manifest format\"}");
     }
+
+    @Test
+    public void testWarningData() throws Exception {
```
wetneb, line 125: Very cool that you were able to merge this back into the existing test class!

sunilnatraj: Thanks

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
sunilnatraj: @wetneb Thanks for catching this; i missed updating this.

wetneb:
Thank you! Would you mind removing the unnecessary files you added to the entity cache, just as a way of validating that we are indeed not requesting more entities than necessary?

sunilnatraj:
@wetneb Removed P131 & P1435 property data file

extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/ItemRequiresScrutinizer.java
```diff
@@ -100,20 +106,32 @@ public void scrutinizeStatementEdit(StatementEntityEdit update) {
                 PropertyIdValue itemRequiresPid = constraint.itemRequiresPid;
                 List<Value> itemList = constraint.itemList;
                 if (!propertyIdValueValueMap.containsKey(itemRequiresPid)) {
-                    QAWarning issue = new QAWarning(update.isNew() ? newItemRequirePropertyType : existingItemRequirePropertyType,
+                    QAWarning issue = new QAWarning(update.isNew()
+                            ? (constraint.itemList.size() == 1 ? newItemRequirePropertywithSuggestedValueType : newItemRequirePropertyType)
+                            : (constraint.itemList.size() == 1 ? existingItemRequirePropertywithSuggestedValueType
+                                    : existingItemRequirePropertyType),
```
wetneb, line 112: I wonder if `constraint.itemList.size() == 1` can be true at all in this case, because we are in an `if (!propertyIdValueValueMap.containsKey(itemRequiresPid)) { … }`
It looks like in this case, the constraint definition does not include any items as expected values, so we should know that `constraint.itemList.size() == 0`, no? Or did I miss anything?

If that's indeed true, then that means there are two less warning ids to introduce (and we can remove the associated translations).

sunilnatraj: @wetneb In my testing using the test data, I see a warning of type "existing-item-requires-certain-other-statement" P633 in relation to P1435 has 16 suggested values.

wetneb: This one is expected indeed, but I don't think you can generate `new-item-requires-certain-other-statement-with-suggested-value` or `existing-item-requires-certain-other-statement-with-suggested-value`.

sunilnatraj: "ew-item-requires-certain-other-statement-with-suggested-value" - comes up in the unit test "testNewItemTrigger"

extensions/wikibase/tests/src/org/openrefine/wikibase/commands/PreviewWikibaseSchemaCommandTest.java
```diff
@@ -110,4 +120,33 @@ public void testInvalidManifest() throws IOException, ServletException {
...
+        ObjectNode response = ParsingUtilities.evaluateJsonStringToObjectNode(writer.toString());
+        ArrayNode issues = (ArrayNode) response.get("warnings");
+        for (JsonNode node : issues) {
+            // Read aggregationId
+            String aggregationId = node.get("aggregationId").asText();
+            // Check for nested property properties/missing_property_entity/label
+            JsonNode addedPropertyLabel = node.path("properties").path("added_property_entity").path("label");
+            JsonNode itemEntityLabel = node.path("properties").path("item_entity").path("label");
+
+            if (aggregationId.equals("existing-item-requires-certain-other-statement-with-suggested-value_P633P17")) {
```
wetneb, line 144: That's an `if` condition - note that you are not asserting that this warning id is emitted at all.

extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/ItemRequiresScrutinizer.java
wetneb: It comes up in the condition of an `if` block, but nothing tells me that this condition is ever met ;)

extensions/wikibase/tests/src/org/openrefine/wikibase/commands/PreviewWikibaseSchemaCommandTest.java
sunilnatraj: Fixed

extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/ItemRequiresScrutinizer.java
sunilnatraj: Fixed the miss in the unit test.

wetneb (changes requested):
Ok, sorry the code path is indeed reachable, but I still think we don't need to declare 4 new warning types: only two should be sufficient.

I don't see a meaningful difference (beyond slight formulation details) between:
* `existing-item-requires-property-to-have-certain-values-with-suggested-value` and `existing-item-requires-certain-other-statement-with-suggested-value`
* `new-item-requires-property-to-have-certain-values-with-suggested-value` and `new-item-requires-certain-other-statement-with-suggested-value`

I would propose that we keep:
* `existing-item-requires-property-to-have-certain-values-with-suggested-value` 
* `new-item-requires-property-to-have-certain-values-with-suggested-value`

and that we delete the two other ones, replacing their uses by those two that we keep.

This would make translators' life easier, otherwise they are translating essentially the same content twice.

What do you think?

sunilnatraj:
@wetneb I have made the updates

wetneb (approved):
Amazing! Let's merge it. Thanks a lot for your patience and perseverance on this!

sunilnatraj:
Thank you @wetneb for your inputs that made this possible

PR #6896, opened 2024-10-11
Improve SECURITY.md to avoid having to update it after each release

wetneb:
This avoids mentioning a specific version and attempts to describe the policy we have followed so far, when patching vulnerabilities.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,10 +2,8 @@
 
 ## Supported Versions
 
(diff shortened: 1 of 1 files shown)


tfmorris (changes requested):
That's good future proofing. I tweaked the wording slightly to try to avoid leaving the impression that all vulnerabilities get fixed, sinc e it's, at best, only the reported ones.

PR #6898, opened 2024-10-11
Fix Google Data extension to use JSON instead of JSONP

### LINKED ISSUE #6897: Google Data extension doesn't work
The removal of JSONP support in #6546 broke the Google Data extension which was using JSONP. This can be easily fixed by updating the extension, but it would be more friendly and make debugging easier if attempts to use JSONP triggered an error because otherwise Javascript clients (and perhaps others) will fail in obscure ways.

### To Reproduce
- Attempt to create a project from either a public or private Google Sheet 

### Current Results

Spinner is shown and operation never completes.

### Expected Behavior

Preview rows are displayed.

### Versions

 - OpenRefine: master (not present in 3.8.x)

### Additional context
 #4082  is related.

tfmorris:
Fixes #6897 

- Switch Google extension to use JSON since JSONP has been removed from the GetRows command
- Attempt to provide useful error messages to callers who use JSONP with the GetRows command
-

### DIFF
--- a/extensions/gdata/module/scripts/index/importing-controller.js
+++ b/extensions/gdata/module/scripts/index/importing-controller.js
@@ -70,7 +70,7 @@ Refine.GDataImportingController.prototype.startImportingDocument = function(doc)
             function(data2) {
             dismiss();
 
-            if (data2.status == 'ok') {
+            if (data2.status === 'ok') {
                 self._doc = doc;
                 self._jobID = data.jobID;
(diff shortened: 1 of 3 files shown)


tfmorris:
I agree that those checks were overly aggressive, so I've removed them and just put them in the GetRowsCommand (as well as leaving the supporting methods available in Command for other commands to use if they want/need).

wetneb (approved):

Issue #6401, opened 2024-02-26
Replace use of java.util.Properties in Bindings


tfmorris:
Currently our variable storage, called "bindings", uses the `java.util.Properties` class for its implementation and its interface. It also pre-populates a variety of standard variables with newly instantiated objects `WrappedCell`, `WrappedRow`, and `CellTuple` which are rarely ever used, causing the object creation overhead to be wasted

Although it's strongly discouraged, the current code bypasses the `setProperty` and `getProperty` methods to use `Hashtable`'s `get()` and `put()` directly, allowing it to support an effective contract of `Map<String, Object>` (but without support for `null` values since `Hashtable` doesn't support them).

### Proposed solution

- Introduce a Bindings class which of `HashMap<String, Object>` which does lazy creation of all wrapped objects only when they are needed.
- Extend the interfaces which currently use `Properties` like `Evaluable`, `HasFields`, `HasFieldsList`, `Function`,  and `Control` to use the new Bindings class with default implementations for backward compatibility.
- Add a new `Binder2` interface which implements the new protocol variable lookup protocol
- Change code which does direct creation of bindings using `new Properties()` to use `ExpressionUtils.getBindings(Project)` like the main code does. Most occurrences of this pattern are scattered throughout the test code. 
- Log warnings on the console (or notify user at startup?) if any legacy Function, Control, or Binder is registered.

### Alternatives considered

- Not attempt compatibility and force all extension writers to update their code immediately.
- Compatibility could be implemented on a per-call basis or the entire system could fall back to legacy mode if any legacy Function, Control, or Binder is registered - to be investigated if this simplifies things

wetneb:
The `Binder` extension point is currently used by at least one extension (RDF Transform). If we want to get rid of it, it probably makes sense to offer an alternative.

tfmorris:
> The `Binder` extension point is currently used by at least one extension (RDF Transform). 

Thanks for the heads up! I guess that illustrates the danger of assuming that any public interface is unused, even if it's not documented. I'll reintroduce the Binder2 interface that I had and figure out what needs to be done to preserve backward compatibility. I'm pretty sure there's no test coverage of this feature.

I've updated the description with more information on API compatibility.

tfmorris:
[...]

PR #6899, opened 2024-10-12
refactoring: Introduce base interface for GREL expressions

wetneb:
So far, AST nodes of GREL expressions are `Evaluable` themselves. This refactoring introduces a dedicated base interface for such AST nodes, `GrelExpr`, so that there is no confusion with `Evaluable`, which is meant to be used as interface with the rest of the code (facets, operations and other such components).

This makes it possible to add methods common to all GREL expressions on the `GrelExpr` interface, without polluting the `Evaluable` interface itself.

### DIFF
--- a/main/src/com/google/refine/templating/Parser.java
+++ b/main/src/com/google/refine/templating/Parser.java
@@ -38,6 +38,7 @@ import java.util.List;
 
 import com.google.refine.expr.MetaParser;
 import com.google.refine.expr.ParsingException;
+import com.google.refine.grel.GrelEvaluable;
 import com.google.refine.grel.ast.FieldAccessorExpr;
 import com.google.refine.grel.ast.VariableExpr;
 
@@ -96,11 +97,12 @@ public class Parser {
 
                     fragments.add(
                             new DynamicFragment(
-                                    new FieldAccessorExpr(
+                                    new GrelEvaluable(
                                             new FieldAccessorExpr(
-                                                    new VariableExpr("cells"),
-                                                    columnName),
-                                            "value")));
+                                                    new FieldAccessorExpr(
+                                                            new VariableExpr("cells"),
+                                                            columnName),
+                                                    "value"))));
 
                     continue;
                 }
--- a/modules/grel/src/main/java/com/google/refine/grel/Control.java
+++ b/modules/grel/src/main/java/com/google/refine/grel/Control.java
@@ -39,7 +39,7 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-import com.google.refine.expr.Evaluable;
+import com.google.refine.grel.ast.GrelExpr;
 
 /**
  * Interface of GREL controls such as if, forEach, forNonBlank, with. A control can decide which part of the code to
@@ -47,9 +47,9 @@ import com.google.refine.expr.Evaluable;
  */
 public interface Control {
 
-    public Object call(Properties bindings, Evaluable[] args);
+    public Object call(Properties bindings, GrelExpr[] args);
 
-    public String checkArguments(Evaluable[] args);
+    public String checkArguments(GrelExpr[] args);
 
     @JsonProperty("description")
     public String getDescription();
--- a/modules/grel/src/main/java/com/google/refine/grel/GrelEvaluable.java
+++ b/modules/grel/src/main/java/com/google/refine/grel/GrelEvaluable.java
@@ -1,0 +1,32 @@
+
+package com.google.refine.grel;
+
+import java.util.Optional;
+import java.util.Properties;
+import java.util.Set;
+
+import com.google.refine.expr.Evaluable;
+import com.google.refine.grel.ast.GrelExpr;
+
+/**
+ * An evaluable made out of a GREL expression.
+ */
+public class GrelEvaluable implements Evaluable {
+
+    private GrelExpr _expr;
+
+    public GrelEvaluable(GrelExpr expr) {
+        _expr = expr;
+    }
+
+    @Override
+    public Object evaluate(Properties bindings) {
+        return _expr.evaluate(bindings);
+    }
+
+    @Override
+    public Optional<Set<String>> getColumnDependencies(Optional<String> baseColumn) {
+        return _expr.getColumnDependencies(baseColumn);
+    }
+
+}
--- a/modules/grel/src/main/java/com/google/refine/grel/Parser.java
+++ b/modules/grel/src/main/java/com/google/refine/grel/Parser.java
@@ -48,6 +48,7 @@ import com.google.refine.grel.Scanner.TokenType;
 import com.google.refine.grel.ast.ControlCallExpr;
 import com.google.refine.grel.ast.FieldAccessorExpr;
 import com.google.refine.grel.ast.FunctionCallExpr;
+import com.google.refine.grel.ast.GrelExpr;
 import com.google.refine.grel.ast.LiteralExpr;
 import com.google.refine.grel.ast.OperatorCallExpr;
 import com.google.refine.grel.ast.VariableExpr;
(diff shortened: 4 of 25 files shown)


tfmorris (commented):
> This makes it possible to add methods common to all GREL expressions on the GrelExpr interface, without polluting the Evaluable interface itself.

I don't see any additional methods. Is this something that you plan to introduce in the future?

Assuming that there will be additional methods at some point, is there a reason that GrepExpr doesn't exend Evaluable to preserve backward compatibility?

Some of the changes, as currently proposed, break compatibility. Is there a demarcation where you're considering some of these interfaces/classes "internal" in some way and others public?

If there are other goals that you're trying to achieve here, it would be helpful to understand what they are.

Sorry for all the questions, but I feel like I don't understand the whole context.

wetneb:
> I don't see any additional methods. Is this something that you plan to introduce in the future?

Yes, I thought it's worth doing the refactoring and the introduction of such methods separately, to keep things clean

> Assuming that there will be additional methods at some point, is there a reason that GrepExpr doesn't exend Evaluable to preserve backward compatibility?

For instance I plan to make it so that `Evaluable::toString()` returns the entire expression (such as `grel:value.toNumber()`) whereas `GrelExpr::toString()` returns the string representation of the expression without prefix (as it currently does), but I guess I could also introduce different methods.

> Some of the changes, as currently proposed, break compatibility. Is there a demarcation where you're considering some of these interfaces/classes "internal" in some way and others public?

As far as I can see, this refactor does not affect the intended extension points:
-  the definition of new functions stays the same
- the definition of new expression languages stays compatible, with a deprecation
- the registration of binders also stays the same

That covers all relevant extension points listed at https://openrefine.org/docs/technical-reference/writing-extensions

If you think we should also maintain compatibility for GREL controls I can try to preserve the signature, but to me I don't see that as a public extension point.
The rest of the changes affect classes that define the GREL AST, which I don't see as a part of our extension interface (for instance because introducing new GREL AST nodes would generally imply making changes to the parser itself).

> If there are other goals that you're trying to achieve here, it would be helpful to understand what they are.

It's a preparation for introducing further methods to allow rewriting an expression by renaming columns, and felt to me like a sensible clean-up. If you oppose it I can try to work without.

tfmorris:
> If you oppose it I can try to work without.

Not opposed, just trying to understand it. I am, however, very reluctant to introduce new APIs using the Properties class, so perhaps we should try to make progress on #6401 if we're going to be mucking with the evaluation framework.

> That covers all relevant extension points listed at https://openrefine.org/docs/technical-reference/writing-extensions

Can we assume that these are the only things that we need to worry about compatibility for? I generally assume anything declared as public needs to be preserved if at all possible.

wetneb:
This is definitely a good occasion to discuss what `public` classes in OpenRefine should mean, because we seem to have very different expectations indeed (following up on the [last discussion](https://github.com/OpenRefine/OpenRefine/pull/6555)).

Here's how I see it. Our Java classes are sorted into various packages, roughly organized by feature area and topic. For a class in package `com.google.refine.foo` to be visible in `com.google.refine.bar`, we need that class to be declared as `public`. As a consequence, virtually all of our top-level classes are marked as `public` (by this I mean, classes which aren't inner classes, which is the overwhelming majority of all classes).

Take for instance `com.google.refine.browsing.util.ExpressionTimeValueBinner`. In my opinion, that's something that ought to be considered internal, meaning that extensions shouldn't see this class at all and only interact with it through a `TimeRangeFacet` (which is the user-facing feature it supports). However, `ExpressionTimeValueBinner` is `public`. `TimeRangeFacet` is in a different package `com.google.refine.browsing.facets` so it wouldn't be able to use `ExpressionTimeValueBinner` if it were any less than `public`.

The bottom line is: unless we change package structure, basically all of classes need to be `public` so that we can use them internally (sorry if I'm stating the obvious - I needed to look up the exact rules of Java visibility to write this myself…)

Ideally, we'd need a way to say "this class should only be visible in `com.google.refine.*`", but that's not something that Java encapsulation lets us express. There might be ways to do something like that with the [JPMS](https://www.oracle.com/corporate/features/understanding-java-9-modules.html) but adopting this technology [looks pretty hard](https://www.reddit.com/r/java/comments/1dxbag2/comment/lc0nnh5/).

The creation of independent Maven modules (`core` and `grel`) is one way to introduce a form of encapsulation: if extension authors take care of only adding `core` as a dependency to their build system, then they won't see other classes. But of course `core` still contains a lot of classes that ought to be internal. And the extension point to define new GREL functions is in `grel` obviously, so it's also legitimate to add a dependency to that (although we could reconsider this, moving this extension point to `core`, which would perhaps be a good step towards [making GREL functions available in other languages](https://github.com/OpenRefine/OpenRefine/issues/#2251), but I digress).

Another approach would be to reorganize our packages so that internal classes can be grouped together with the public classes they support in the same package. This way we'd be able to use the default visibility ("package-private") for internal classes and methods. We'd probably need a much flatter directory structure. Of course this is a big breaking change…

We could also try to add Java annotations or Javadoc specifying the intended visibility of the classes (it's something I have seen in many other projects). Of course with no enforcement guarantees, and it would have been better to have it from the start.

Anyway, as things stand, if we try to maintain every class marked as `public` as if it's part of our extension API and any breaking change to it means new major version, I would say we are shooting ourselves in the foot. The code would become quite quickly riddled with deprecated methods and it would make it quite a bit harder to introduce any new functionality.

> I am, however, very reluctant to introduce new APIs using the Properties class, so perhaps we should try to make progress on https://github.com/OpenRefine/OpenRefine/issues/6401 if we're going to be mucking with the evaluation framework.

Well, from my point of view the `GrelExpr` interface is an internal one, so it's not defining a new API. It's marked as public for the reasons above, but it's not intended for external use.

Again, I don't want to cling on to this specific PR more than necessary - I could drop it and hack my way around the problem. But it would be very useful to reach some sort of common ground on this visibility topic because it's going to keep popping up all the time otherwise.

Also, it's worth discussing who we are maintaining compatibility for. Personally, I want to maintain compatibility for extensions. There can be other uses of our Java classes:
* treating our Maven artifacts as a library for a project that's not an OpenRefine extension,
* maintaining an OpenRefine fork,
* any other funky variation of the above…

I am happy with people using our classes in these other ways (the license doesn't forbid it) but I am not interested in considering such use cases when deciding whether to maintain a particular API. Do you see it in a different way?

tfmorris:
Sorry for the delay in replying. This is probably a conversation that's important enough to be worth having outside of the context of a single PR review.

I think that the default starting point, years ago, was more or less "Everything is part of the API" just because we didn't publish any guidelines saying otherwise. Whether or not this has changed because there is no a little bit for formal documentation on the various formal extension points is something worth considering. I don't know how strongly developers have been warned not to use anything else or what their expectations are.

As a point of comparison, here are the guidelines that Eclipse offers developers: https://www.eclipse.org/articles/Article-API-Use/index.html

I'm not sure where to draw the line, but I think we need to have a clear definition. While it's definitely more work to maintain compatibility, I think it provides significant benefits as well.

wetneb:
Thanks for the reply.

I agree it would be useful to have more clarity on what should be stable or not, so I made #6928 to help with that. It's not a silver bullet, but it could hopefully help us document this gradually and at least make it possible to introduce new internal elements identified as such from the start.

I would still argue that "Everything is part of the API" isn't a viable status-quo for the project. I don't remember reading that anywhere and I would find it surprising that the absence of guidelines is understood like that by extension developers. If that were the case, then the same should apply to the front-end since we have no guidelines for it, so our DOM structure and Javascript method signatures would need to stay stable… I think we're very far from providing that!
Even in the back-end, I don't think the migrations OffsetDateTime or to Jackson have conveyed a particular faith in an "Everything is part of the API" principle (I'm not saying I would do them again either).

wetneb:
Let's drop this given @tfmorris' reluctance - I hope we can make progress in #6928 instead

tfmorris:
I didn't look in detail, but I thought this could be done in a compatible way, which was why I asked:

> Assuming that there will be additional methods at some point, is there a reason that GrepExpr doesn't exend Evaluable to preserve backward compatibility?

Is that a possibility or did I miss something?

PR #6904, opened 2024-10-15
Move error reporting in apply-operations dialog away from alert into dialog

wetneb:
When trying to apply operations with JSON syntax errors in the "Apply Operation History" dialog, the error currently gets reported as a JS alert. This PR:
* moves the error message into the dialog to make it less agressive
* improves the error message (the content hasn't necessarily been "pasted", so remove reference to that, and include the specifics of why the JSON is invalid)

image

### DIFF
--- a/main/webapp/modules/core/styles/project/sidebar.css
+++ b/main/webapp/modules/core/styles/project/sidebar.css
@@ -251,3 +251,7 @@ textarea.history-operation-json {
   min-height: 450px;
   min-width: 374px;
 }
+
+.history-operation-json-error {
+  color: var(--error-red);
+}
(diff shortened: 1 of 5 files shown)


main/webapp/modules/core/styles/project/sidebar.css
```diff
@@ -251,3 +251,7 @@ textarea.history-operation-json {
   min-height: 450px;
   min-width: 374px;
 }
+
+.history-operation-json-error {
+  color: var(--error-red);
```
wetneb, line 256: The error currently renders in black because of a dependency of this PR on #6903.

tfmorris (approved):

PR #6907, opened 2024-10-17
apply-operations: Report invalid JSON structure from the backend to the frontend

wetneb:
This adds error reporting to the apply-operations command, for things that are valid JSON but when that JSON does not correspond to a valid operation history (for instance `[{}]`).

This will still accept operations with missing parameters, such as:
```
[
    { "op": "core/mass-edit" }
]
```
Rejecting those sorts of inputs will be done in further PRs.

image

### DIFF
--- a/main/src/com/google/refine/commands/history/ApplyOperationsCommand.java
+++ b/main/src/com/google/refine/commands/history/ApplyOperationsCommand.java
@@ -42,6 +42,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import com.fasterxml.jackson.core.type.TypeReference;
+import org.apache.commons.lang.Validate;
 
 import com.google.refine.commands.Command;
 import com.google.refine.model.AbstractOperation;
@@ -66,8 +67,18 @@ public class ApplyOperationsCommand extends Command {
             List<AbstractOperation> operations = ParsingUtilities.mapper.readValue(jsonString,
                     new TypeReference<List<AbstractOperation>>() {
                     });
+
+            // Validate all operations first
             for (AbstractOperation operation : operations) {
-                runOperation(project, operation);
+                Validate.notNull(operation);
+                Validate.isTrue(!(operation instanceof UnknownOperation), "Unknown operation type: " + operation.getOperationId());
+                operation.validate();
+            }
+
+            // Run all operations in sequence
+            for (AbstractOperation operation : operations) {
+                Process process = operation.createProcess(project, new Properties());
+                project.processManager.queueProcess(process);
             }
 
             if (project.processManager.hasPending()) {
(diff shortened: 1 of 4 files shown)


main/src/com/google/refine/commands/history/ApplyOperationsCommand.java
```diff
@@ -59,26 +59,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
-            for (int i = 0; i < count; i++) {
-                if (a.get(i) instanceof ObjectNode) {
-                    ObjectNode obj = (ObjectNode) a.get(i);
+            Project project = getProject(request);
+            String jsonString = request.getParameter("operations");
 
-                    reconstructOperation(project, obj);
+            try {
+                ArrayNode a = ParsingUtilities.evaluateJsonStringToArrayNode(jsonString);
+                int count = a.size();
```
tfmorris, line 68:
```suggestion
                ArrayNode a = ParsingUtilities.evaluateJsonStringToArrayNode(jsonString);
                if (a != null) {
                    int count = a.size();
```

It looks like this method can return null (and doesn't report errors :( ), so call to `.size()` needs to be protected

main/src/com/google/refine/commands/history/ApplyOperationsCommand.java
```diff
@@ -59,26 +59,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+                return;
             }
 
             if (project.processManager.hasPending()) {
                 respond(response, "{ \"code\" : \"pending\" }");
             } else {
                 respond(response, "{ \"code\" : \"ok\" }");
             }
-        } catch (IOException e) {
+        } catch (Exception e) {
```
tfmorris, line 86: Why was this catch widened? I think we should try to avoid catching `Exception` except in extreme last-chance cases

tfmorris (changes requested):
This looks like a nice addition. One small change request and one question below...

main/src/com/google/refine/commands/history/ApplyOperationsCommand.java
wetneb: Because `reconstructOperation` doesn't throw `IOException` when an invalid structure is encountered, but `IllegalArgumentException`. Catching all exceptions at the root of the command is the only way I can think of to make sure the backend responds with an appropriate JSON response to the frontend, as it isn't handled by the framework for us. Do you have examples of exceptions you would let through?

main/src/com/google/refine/commands/history/ApplyOperationsCommand.java
wetneb: I've simplified the whole structure, to stop using this utility method and improve exception handling.

wetneb:
With #6908 being merged, I have resolved the conflicts with this PR. Now that operation validation is available, I have added another small change on top (c7312dc508beb7814d8129a8817ae948700b078a), to completely separate the validation of the operations and their execution. The goal there is to avoid leaving the user in a half-baked state where only some of the operations ran. This will be improved later with the validation of column dependencies ahead of time.

main/webapp/modules/core/styles/project/sidebar.css
```diff
@@ -254,4 +254,5 @@ textarea.history-operation-json {
 
 .history-operation-json-error {
   color: var(--error-red);
+  max-width: 750px;
```
Abbe98, line 257: `max-width: calc(800px - var(--padding-looser));` in case you want to make it pixel perfect.

Abbe98 (approved):
Great addition, makes me want to work on syntax highlighting and and a consistent error reporting UI :-)

sunilnatraj (approved):

PR #6908, opened 2024-10-17
Introduce AbstractOperation::validate() to check operation parameters

wetneb:
Most operations don't check their parameters before execution, so when they are deserialized from invalid JSON payloads, this can lead to execution errors which ought to be prevented earlier.

This introduces a validate() method on operations giving them the opportunity to check the validity of their parameters. It would intuitively be more natural to do those checks in the operation constructors, as it would guarantee that if you have an instance of an operation, it has suitable parameters, but if we were to do this, we would run into the risk that deserializing operations fail at the loading of a project (for instance: say I use an extension which provides a custom expression language, which I then use in an operation. If I then remove the extension, I wouldn't be able to open my project anymore)

For now this PR just contains the basic scaffold without implementing any checks in any operation, just to validate the general approach. I would then migrate the error-checking done in `Command` classes into the `validate()` methods of the corresponding operations (so it's not duplicated).


wetneb:
Fishing for a reviewer for this one, so I can make progress on the follow-up PRs too.

sunilnatraj (approved):

PR #6912, opened 2024-10-18
Add error reporting to JSON preview parse. Fixes #6454

### LINKED ISSUE #6454: JSON parse errors not reported to the user
As reported in #3243, JSON parsing errors, at least during the original preview parse, are being silently ignored. They should be reported to the user.

### To Reproduce
1. First, do ...
2. Then, ...
3. Finally, ...

### Current Results
Invalid JSON data is silently ignored.

### Expected Behavior
Errors returned by the JSON parser should be reported to the user.

tfmorris:
Fixes #6454

- Add error reporting to JSON preview parse.


wetneb (approved):
Looks great, thank you!

PR #6914, opened 2024-10-21
wikibase: Test failure when run on non-English locales

### LINKED ISSUE #6911: wikibase: Test failure when run on non-English locales
When running the test suite with a French locale, I get:

```
[ERROR] Failures: 
[ERROR]   PreviewWikibaseSchemaCommandTest.testWarningData:147 expected [country] but found [pays]
```

This is a symptom of the fact that https://github.com/OpenRefine/OpenRefine/pull/6895 (cc @sunilnatraj) picks the language in which to fetch the labels based on the system locale.

Instead, we should change the command to take this language from the HTTP request so that it can match the language used in the UI (which is often not the same thing as the system locale).

As a temporary workaround, one can [change the locale used by Maven](https://stackoverflow.com/questions/8901880/set-surefire-plugin-locale-in-maven-pom-file) to get the test suite to pass.

sunilnatraj:
Fixes #6911 

- Set language before test as the test validates label which is language specific
- 
-


sunilnatraj:
@wetneb Raised PR to address the unit test issue

wetneb (approved):
Thank you very much for taking care of that!

PR #6915, opened 2024-10-21
ui: Fix dialog warning of history overwriting for some operations

### LINKED ISSUE #6746: Transformations are silently ignored if Undo is pending
### To Reproduce
1. Open a project with an operation history
2. Undo the last operation by clicking on the 2nd to last operation in the history
3. Use the Edit Cells -> Transform operation to make any change to the grid

### Current Results
Nothing happens.

### Expected Behavior
Confirmation dialog asking if you really want to discard the undone operation

### Versions
 - Operating System: macOS Sonoma 14.5 (23F79)
 - Browser Version: Chrome  Version 126.0.6478.127 (Official Build) (arm64)
 - OpenRefine: 3.9 SNAPSHOT

wetneb:
Closes #6746.

Somehow I missed a lot of places where the callbacks to dismiss the operation dialog had to be moved to the `Refine.postCoreProcess` call. I am not sure how this went unnoticed, because I remember testing this manually…
I reviewed all calls of `Refine.postCoreProcess` and made the necessary changes.

When running an operation which is configured in a dialog, the new dialog warning against history overwriting is shown when the user validates the operation dialog. I realized that one could instead choose to display the warning when the operation dialog is first opened. This could be a follow-up change if we prefer this design.

### DIFF
--- a/main/tests/cypress/cypress/e2e/project/undo_redo/warn_for_history_deletion.cy.js
+++ b/main/tests/cypress/cypress/e2e/project/undo_redo/warn_for_history_deletion.cy.js
@@ -83,4 +83,50 @@ describe(__filename, function () {
     // There are no future changes anymore, since they have been deleted
     cy.get('.history-panel-body .history-future a').should('not.exist');
   });
+
+  it('Apply a text transform after undoing some changes', function () {
+    cy.loadAndVisitProject('food.mini');
+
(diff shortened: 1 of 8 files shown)


tfmorris (approved):
Thanks! I only manually tested the one case that I found before, but assuming the rest are good too.

PR #6924, opened 2024-10-22
Fix database import SQLite not closing connections

### LINKED ISSUE #6922: SQLite connections remain open after use
While testing a few things for this [reported SQLite issue on the forum](https://forum.openrefine.org/t/unauthorized-to-open-sqlite-database/1835), I did find an issue with our database access code. SQLite connections remain open and are not properly closed after use (so a SQLite file will remain locked on the filesystem and inaccessible; i.e. you can't move it or compress it, etc.).

### To Reproduce
1. First, in OpenRefine, create a new connection to a SQLite file on disk
2. Then, click on Test button on Database import
3. Finally, try to compress the SQLite file on disk 

### Current Results
filesystem ERROR that SQLite file is open and in use.

### Expected Behavior
SQLite file on disk should not remain open and in use after OpenRefine has finished using a database connection to it.

### Versions
 - Browser Version: Edge Version 130.0.2849.46
 - JRE or JDK Version: openjdk version "21" 2023-09-19 LTS
 - OpenRefine: Version 3.9-SNAPSHOT

### Datasets

SQLite .db test file:
[test.zip](https://github.com/user-attachments/files/17468595/test.zip)

thadguidry:
Fixes #6922

- Closes SQLite connections remaining open (and db file in use)

### DIFF
--- a/extensions/database/src/com/google/refine/extension/database/sqlite/SQLiteDatabaseService.java
+++ b/extensions/database/src/com/google/refine/extension/database/sqlite/SQLiteDatabaseService.java
@@ -116,8 +116,8 @@ public class SQLiteDatabaseService extends DatabaseService {
     @Override
     public DatabaseInfo executeQuery(DatabaseConfiguration dbConfig, String query) throws DatabaseServiceException {
         try (Connection connection = SQLiteConnectionManager.getInstance().getConnection(dbConfig);
-             Statement statement = connection.createStatement();
-             ResultSet queryResult = statement.executeQuery(query)) {
+                Statement statement = connection.createStatement();
+                ResultSet queryResult = statement.executeQuery(query)) {
             ResultSetMetaData metadata = queryResult.getMetaData();
             int columnCount = metadata.getColumnCount();
             ArrayList<DatabaseColumn> columns = new ArrayList<>(columnCount);
@@ -152,8 +152,8 @@ public class SQLiteDatabaseService extends DatabaseService {
     @Override
     public DatabaseInfo testQuery(DatabaseConfiguration dbConfig, String query) throws DatabaseServiceException {
         try (Connection connection = SQLiteConnectionManager.getInstance().getConnection(dbConfig);
-             Statement statement = connection.createStatement();
-             ResultSet queryResult = statement.executeQuery(query)) {
+                Statement statement = connection.createStatement();
+                ResultSet queryResult = statement.executeQuery(query)) {
             return new DatabaseInfo();
         } catch (SQLException e) {
             throw new DatabaseServiceException(true, e.getSQLState(), e.getErrorCode(), e.getMessage());
@@ -164,8 +164,8 @@ public class SQLiteDatabaseService extends DatabaseService {
     public List<DatabaseColumn> getColumns(DatabaseConfiguration dbConfig, String query)
             throws DatabaseServiceException {
         try (Connection connection = SQLiteConnectionManager.getInstance().getConnection(dbConfig);
-             Statement statement = connection.createStatement();
-             ResultSet queryResult = statement.executeQuery(query)) {
+                Statement statement = connection.createStatement();
+                ResultSet queryResult = statement.executeQuery(query)) {
             ResultSetMetaData metadata = queryResult.getMetaData();
             int columnCount = metadata.getColumnCount();
             ArrayList<DatabaseColumn> columns = new ArrayList<>(columnCount);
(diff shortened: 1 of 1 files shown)


wetneb (commented):
That looks good to me, thanks for catching that!
Did you consider using the try-with-resources syntax instead?

wetneb:
Do you mind me changing it to this syntax in this PR? I think it would simplify the code a bit.

thadguidry:
@wetneb sure thing.

extensions/database/src/com/google/refine/extension/database/sqlite/SQLiteDatabaseService.java
```diff
@@ -199,20 +184,17 @@
...
-        Connection connection = SQLiteConnectionManager.getInstance().getConnection(dbConfig);
-        Statement statement = null;
-        ResultSet queryResult = null;
-        try {
-            statement = connection.createStatement();
-            statement.setFetchSize(10);
-            queryResult = statement.executeQuery(query);
+        try (Connection connection = SQLiteConnectionManager.getInstance().getConnection(dbConfig);
+             Statement statement = connection.createStatement();
+             ResultSet queryResult = statement.executeQuery(query)) {
```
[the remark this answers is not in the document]
wetneb: I agree we don't need to log those SQL errors with `logger.error`, but I wonder if adding the `SQLException:: ` prefix to the error message is really useful? If it's meant to be reported to the user, we probably don't need to make it look more like a Java exception, do we? As a user, I would prefer to have just the SQL error without any Java things around.

extensions/database/src/com/google/refine/extension/database/sqlite/SQLiteDatabaseService.java
```diff
@@ -147,47 +148,31 @@ public DatabaseInfo executeQuery(DatabaseConfiguration dbConfig, String query) t
             dbInfo.setRows(rows);
             return dbInfo;
         } catch (SQLException e) {
-            logger.error("SQLException::", e);
-            throw new DatabaseServiceException(true, e.getSQLState(), e.getErrorCode(), e.getMessage());
-        } finally {
-            SQLiteConnectionManager.getInstance().shutdown();
```
wetneb, line 153: Is this `SQLiteConnectionManager.getInstance().shutdown()` somehow implied already by the try-with-resources statement? I would have left it in the `finally` block intuitively

thadguidry: Yes, SQLiteConnectionManager already implements AutoCloseable, inherited through java.sql.Connection.
https://docs.oracle.com/en/java/javase/21/docs/api/java.sql/java/sql/Connection.html

thadguidry: In regards to the error logging, there's already user friendly error messaging for Connections in SQLiteConnectionManager.java

So that leaves Query or or formally JDBC calls to the DatabaseService and so in the SQLiteDatabaseService.java that's where java.sql.SQLException (and through JDBC basically) takes care of those more user friendly messages - see the docs on java.sql.SQLException -->
>java.sql.SQLException
>An exception that provides information on a database access error or other errors.
>
>Each SQLException provides several kinds of information:
>
>a string describing the error. This is used as the Java Exception message, available via the method getMessage.
>
>a "SQLstate" string, which follows either the XOPEN SQLstate conventions or the SQL:2003 conventions. The values of the SQLState string are described in the appropriate spec. The DatabaseMetaData method getSQLStateType can be used to discover whether the driver returns the XOPEN type or the SQL:2003 type.
>
>an integer error code that is specific to each vendor. Normally this will be the actual error code returned by the underlying database.
>
>a chain to a next Exception. This can be used to provide additional error information.
>
>the causal relationship, if any for this SQLException.
>
>Since:
>1.1

We just need to ensure we pass the SQLException message, the SQLstate, and Error Code down for easy visibility into our logs.  Which I have done.

I'm not sure what you mean by "Java things" ?  lolol, help me, seriously though, I'm not as good at some of this as you are, and maybe some of the DB stuff I have more acquaintance with than you have, so we're a great team!

thadguidry: Also, Statement and ResultSet are also AutoCloseable interfaces.

wetneb: I would just leave the `DatabaseServiceException` thrown by this method like they were before.

thadguidry:
@wetneb good now?

wetneb:
Looks great, thank you :) one round of linting and it'll be good to go

wetneb (approved):

PR #6926, opened 2024-10-23
Fixing flaky test - Improving comparator in BinningCluster.java

mihirgune:
### Description

This pull request fixes flakiness in the following test - [`testSerializeBinningClusterer()`] (https://github.com/OpenRefine/OpenRefine/blob/master/modules/core/src/test/java/com/google/refine/clustering/binning/BinningClustererTests.java#L85) in the `OpenRefine - core` module.

When this test is executed using the [nondex](https://github.com/TestingResearchIllinois/NonDex) plugin with the following comand - 

`mvn -pl modules/core edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest="com.google.refine.clustering.binning.BinningClustererTests#testSerializeBinningClusterer"`

 it occasionally throws this error - 

`[ERROR]   BinningClustererTests.testSerializeBinningClusterer:97 Objects above are not equal as JSON strings.
`
with this json-diff

```
[ [ {        [ [ {
  "c" : 1,     "c" : 1,
  "v" : "a"    "v" : "c"
}, {         }, {
  "c" : 1,     "c" : 1,
  "v" : "à"    "v" : "?"
} ], [ {     } ], [ {
  "c" : 1,     "c" : 1,
  "v" : "c"    "v" : "a"
}, {         }, {
  "c" : 1,     "c" : 1,
  "v" : "?"    "v" : "à"
} ] ]        } ] ]
```

This error is thrown when the test is comparing the expected JSON with the target JSON in the [com.google.refine.util.TestUtils#assertEqualsAsJson](https://github.com/OpenRefine/OpenRefine/blob/7f0be3d6ac61323f2258c10536d6332ce01761be/modules/core/src/test/java/com/google/refine/util/TestUtils.java#L82) method, as the ordering of the arrays containing clusters in the JSON is incorrect.

The sample clusters being used in the test are a List<Map<String, Integer>, and can be in the order `[{c=1, ?=1}, {a=1, à=1}]` or `[{a=1, à=1}, {c=1, ?=1}]` non-deterministically. Since the ordering of elements within a list is deterministic in a JSON, the JSON equality check was failing.

### Fix

The fix for this test involved fixing the comparator used to sort the clusters in BinningCluster.java - [SizeComparator](https://github.com/OpenRefine/OpenRefine/blob/7f0be3d6ac61323f2258c10536d6332ce01761be/modules/core/src/main/java/com/google/refine/clustering/binning/BinningClusterer.java#L188). In the context of the sample JSONs utilized in the test, the comparator was returning `0`, indicating that the clusters were considered equal, which led to inconsistent test outcomes.

Firstly, the condition on [Line 196](https://github.com/OpenRefine/OpenRefine/blob/7f0be3d6ac61323f2258c10536d6332ce01761be/modules/core/src/main/java/com/google/refine/clustering/binning/BinningClusterer.java#L196) was incorrect, where it compared the maps, directly, instead of comparing the sizes of the maps.

Secondly, there was no tie-breaker in the Comparator for maps that have equal sizes and equal values. In this case, my fix is doing a lexicographical comparison of the keys of said maps.

Please let me know if you'd like to see some changes to this fix!


wetneb (approved):
Thanks for catching this! I agree with your analysis.

wetneb:
The end to end test failure looks like it's not spurious - I ran the test suite twice and the same test fails again. Would you be interested in having a look?

tfmorris:
I suspect that we're going to have to choose between:
- using the default JVM order that we've always used, which has, in practice, been stable, or
- using a new order which is guaranteed to be stable, but is incompatible with the historical order

It's not clear to me that the second is clearly better, but it's also not clear to me that this is very visible to users since it is only used in clustering

wetneb:
Given the failing E2E test and the lack of recent activity, I am tentatively closing this, but feel free to reopen (or comment) if you think this is a mistake.

mihirgune:
Hi, I apologize for the lack of activity, this somehow dropped off my radar. I'll take a look at the end-to-end test right away and get back to you.

wetneb:
@mihirgune do you still intend to work on this?

mihirgune:
Hi @wetneb, I do intend to. Since I'm a student, I was swamped with finals over the last week. You can expect an update to this pull request over the next two weeks.

Thanks for your patience.

PR #6928, opened 2024-10-23
core: Introduce @Internal annotation

wetneb:
To mark things that are technically public but shouldn't be used outside of OpenRefine, in particular not by extensions, this PR proposes to use an annotation and adds it on a few classes to demonstrate how I would expect we use it. It's by no means exhaustive, and I expect marking a particular class as internal will often spark a debate.

It's meant to be sort of band-aid for the problem that Java encapsulation doesn't let us align class and method visibilities to match their expected stability, at least not without changing the package structure (as discussed in #6555 and #6899). It's meant to be similar to `@Deprecated`, except that it's only deprecated outside our code base. After an element has been marked as internal for a while, we should feel empowered to change it, remove it, or move it to a different package so that it can properly be package-private.

It would ideally be better to use a standard annotation for this (as it would make some sort of IDE support more likely), but I am not aware of any.

### DIFF
--- a/modules/core/src/main/java/com/google/refine/annotations/Internal.java
+++ b/modules/core/src/main/java/com/google/refine/annotations/Internal.java
@@ -1,0 +1,19 @@
+
+package com.google.refine.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation signifying that the given class or method is only provided for internal use (inside OpenRefine itself) and
+ * should not be used by extensions, as it could change or be removed without notice.
+ */
+@Retention(RetentionPolicy.SOURCE)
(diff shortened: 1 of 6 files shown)


wetneb:
Today I learned that [Kotlin offers the `internal` visibility modifier](https://kotlinlang.org/docs/visibility-modifiers.html#modules) (alongside `public`, `private` and the like), which does just that.
Not saying we should migrate to Kotlin, but at least it's a sign that it's a known issue with the Java encapsulation system.

thadguidry:
We could update **parts** of OpenRefine to Kotlin, or even Groovy, or gasp Rust. ;-)

tfmorris:
If our intent is to make most things internal only, would it be better to have the annotation be Public or API or something and have Internal be the default?

wetneb:
That could also work for sure.

wetneb:
I've opened #7017 as an alternative to demonstrate that approach too.

wetneb:
Closing in favour of #7017

PR #6929, opened 2024-10-23
core: Introduce getSource() and getLanguagePrefix() on Evaluable

wetneb:
As a preparation for introducing the ability to rewrite an `Evaluable` after a renaming of the columns it depends on, this introduces getters to enable representing the expression as text again after this rewrite.

This will in turn enable the rewrite of facet configurations and operations and their serialization in JSON (for instance for #133).

### DIFF
--- a/extensions/jython/src/com/google/refine/jython/JythonEvaluable.java
+++ b/extensions/jython/src/com/google/refine/jython/JythonEvaluable.java
@@ -68,13 +68,16 @@ public class JythonEvaluable implements Evaluable {
         return new LanguageSpecificParser() {
 
             @Override
-            public Evaluable parse(String s) throws ParsingException {
-                return new JythonEvaluable(s);
+            public Evaluable parse(String source, String languagePrefix) throws ParsingException {
+                return new JythonEvaluable(source, languagePrefix);
             }
+
         };
     }
 
     private final String s_functionName;
+    private final String s_originalSource;
+    private final String s_languagePrefix;
 
     private static PythonInterpreter _engine;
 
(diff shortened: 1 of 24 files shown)


extensions/jython/src/com/google/refine/jython/JythonEvaluable.java
```diff
@@ -103,17 +106,24 @@ public Evaluable parse(String s) throws ParsingException {
         logger.debug("Done with static block in Jython initialization");
     }
 
-    public JythonEvaluable(String s) {
+    // Convenience constructor for tests
+    protected JythonEvaluable(String source) {
```
wetneb, line 110: That's a change of public interface, but in an extension, so from my perspective this is fine.

wetneb:
I have added new AST nodes for bracketed expressions and array literals, so that those can be represented more faithfully when pretty-printing a parsed GREL expression. Without this, an expression such as `(1 + 2) * 3` would be pretty-printed back to `1 + 2 * 3`, which is not equivalent to the original. Or `[ 1, 2 ]` would be pretty-printed back to `argsToArray(1, 2)` which is as nice.

modules/grel/src/test/java/com/google/refine/grel/GrelTests.java
```diff
@@ -261,6 +261,34 @@ public void testColumnDependencies() throws ParsingException {
...
+                { "\"foo\"", "\"foo\"" },
+                { "'\"'", "\"\\\"\"" }, // TODO we could print the string with the original quotes to avoid the escaping
+                { "1", "1" },
+                { "4 * (5 + 6)", "4 * (5 + 6)" },
+                { "cells.foo", "cells.foo" },
+                { "value + ' ' + cells.foo.value", "value + \" \" + cells.foo.value" },
+                { "cells[\"foo\"].value", "cells.get(\"foo\").value" }, // TODO this could be more faithful
+                { "toDate( value+4 )", "toDate(value + 4)" },
+                { "(value + 4).toDate()", "(value + 4).toDate()" }, // TODO this could potentially be improved
+                { "forEach([3, 4,8,7], v, mod(v,2))", "forEach([3, 4, 8, 7], v, mod(v, 2))" },
```
wetneb, line 280: This test gives an overview of the sorts of differences introduced by the pretty-printing logic.

PR #6945, opened 2024-10-26
ui: Fix CSRF token fetching utility (and Wikibase login)

### LINKED ISSUE #6941: Wikidata login failed
We were unable to login to production wikidata from the latest version of OoenRefine. 

### To Reproduce
1. Download OpenRefine 3.8.4, prepackaged exe for Windows and install on Windows 10
2. Prepare a Wikidata change and try to apply it
3. When asked for login, use the normal wikidata login (reproduced for both new and old accounts, 5 in total, using different internet connections) 

### Current Results

Login fails with a message indicating wrong credentials were introduced 

### Expected Behavior

Login should be OK

### Versions
 - Browser Version: Edge, Chrome and Firefox latest versions
 - JRE or JDK Version: 
 - OpenRefine: 3.8.4 prepackaged installer. 

### Datasets
Probably not relevant, but can be provides on request

### Additional context
Found during a training session and reproduced reliably by both participants and trainers.

Not sure if relevant, but we had reconciled the data using the Romanian endpoint, not English

wetneb:
Fixes #6941

When adding CSRF protection to various new commands (for [GHSA-3jm4-c6qf-jrh3](https://github.com/OpenRefine/OpenRefine/security/advisories/GHSA-3jm4-c6qf-jrh3)), I set up the token passing by GET or POST parameter depending on the situation. Somehow this has led me to believe that the `postCSRF` utility should be adapted to append the CSRF token at the end of a URL with a `?` separator if it does not have any GET parameters yet. But in fact this method is exclusively used with POST and the string in question is never a full URL but the contents of the POST body, where no `?` is expected, so this was a mistake. The passing of CSRF tokens by GET is done by using `wrapCSRF` directly.

### DIFF
--- a/main/webapp/modules/core/scripts/util/csrf.js
+++ b/main/webapp/modules/core/scripts/util/csrf.js
@@ -20,11 +20,7 @@ CSRFUtil.postCSRF = function(url, data, success, dataType, failCallback) {
     return CSRFUtil.wrapCSRF(function(token) {
         var fullData = data || {};
         if (typeof fullData == 'string') {
-            if (fullData.includes('?')) {
-              fullData = fullData + "&" + $.param({csrf_token: token});
-            } else {
-              fullData = fullData + "?" + $.param({csrf_token: token});
-            }
+            fullData = fullData + "&" + $.param({csrf_token: token});
         } else {
             fullData['csrf_token'] = token;
         }


tfmorris (approved):
This caught my eye in the review, but I failed to follow through and investigate. Sorry about that!

wetneb:
I think that should go in a 3.8.5 release. Should I initiate that or are you interested in doing this release, @tfmorris?

tfmorris:
I can handle it as practice for 3.9. Is there information recorded anywhere about what went wrong with 3.8.3 so I know what to avoid?

wetneb:
The problem with 3.8.3 was an incorrect cherry-picking of a commit, where I accidentally committed a file which had merge conflicts in it. I had to fix that with https://github.com/OpenRefine/OpenRefine/commit/8d4c1628138ab9df5ddc389de4ea2a17fd44a67c.

On top of that, the update of Butterfly broke the entire app, because I hadn't realized that some JSON-related utility that I had removed was actually in use in OpenRefine (while the contrary had been stated before). So I had to adapt that here:
https://github.com/OpenRefine/OpenRefine/commit/5ea2e6c4c65f0840383393275c328ea880ab0edd

Beyond my own sloppiness, I would blame those issues on:
* the fact that I had to work without CI as those were all vulnerability fixes
* the difficulty to run the entire Cypress test suite locally, which means that I don't run it systematically after my changes
* the split between the Butterfly repository and the OpenRefine one, making updates like this one more complicated

PR #6947, opened 2024-10-28
csrf: Fix callback wiring-up for templating exporter

wetneb:
Follow-up to 090924c.

When writing 090924c, I thought it could be useful to return the jQuery request object created in the `wrapCSRF` utility so that it could be used for further chaining, such as adding error handling with the `.fail(…)` method.
That was a bad idea since this would bind the failure callback to the request that fetches a new CSRF token, not the one that actually does the POST request for which the token is fetched. Instead, the failure callback should be passed to the `postCSRF` / `wrapCSRF` method as last argument. I have reviewed all the places where those are used and made sure we don't rely on the return value.


tfmorris (approved):
Rubber stamp LGTM

wetneb:
I think it's worth it indeed.

PR #6961, opened 2024-10-29
ci: Move Cypress record key to a GH Actions secret

### LINKED ISSUE #6959: Move Cypress Cloud secret to Github Secrets
We are currently publishing the secret for Cypress cloud in our workflow file, so anyone who clones the repo can push to our dashboard, making for confusing debugging and wasted time.  We should invalidate that secret, mint a new one, move it to Github secrets, and adjust the workflow to reference it from the secrets store

https://github.com/OpenRefine/OpenRefine/issues/6942#issuecomment-2442785321_

wetneb:
Closes #6959.

I haven't invalidated the old token yet, it should be done after merging this PR. I believe existing PRs will then start using the new workflow, if not they can be rebased.

### DIFF
--- a/.github/workflows/pull_request_e2e.yml
+++ b/.github/workflows/pull_request_e2e.yml
@@ -80,7 +80,7 @@ jobs:
           ./refine e2e_tests
         env:
           CYPRESS_BROWSER: ${{ matrix.browser }}
-          CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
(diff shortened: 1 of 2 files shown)


tfmorris (approved):

tfmorris (approved):
I'm not exactly sure what fixed this since it started working when I made an unrelated change to move the Cypress project ID to a variable, but I also refreshed the secret along the way (with the same value), so perhaps that triggered something? Or perhaps it was just a stale cache?

In any case, it appears to be working now and it's slightly simplified.

wetneb:
Curious, let's hope for the best that it's stable…

PR #6965, opened 2024-10-31
fixes #6672 : Set a delay between opening submenu and hover. 

### LINKED ISSUE #6672: Improve hover submenu navigation/selection
Citing Hao Ye from the carpentries:

> 2. Navigating the submenus can require moving horizontally to the submenu, and then vertically to the option (blue path). Instinctively, I might want to move from where my cursor is to the selection directly (pink path), but this exits out of the selection for the submenu and doesn't work. Not sure of an obvious fix, but this is an area of improvement for the UX.
> image

### Proposed solution

* I checked what other tools do (for instance Google Docs, LibreOffice Calc…) and they all seem to have a little delay between hovering a menu entry and opening the corresponding submenu (maybe 0.5 s?). This is sufficient to ensure that when moving the mouse diagonally, any submenus met in passing are not opened.
* Alternatively we could require users to click on menu entries to open the corresponding submenus (that's what my Firefox does, for instance)

### Alternatives considered

We are aware that those submenus are cluttered and inconvenient and there is interest in moving to something else. But that is a more involved change and I think it shouldn't block a small tweak like this one.

### Additional context

Forwarded via Jennifer Stubbs of the Carpentries.

Shubham-M-Rathod:
Fixes #6672

-  Set a delay between opening submenu and hover. 
-  Diagonal path probem solved by checking if currenlty on submenu-item, added _hoverTimeout in MenuSystem.

https://github.com/user-attachments/assets/5d0a8a47-3ee7-4b46-bb58-dee1a14baa12


Shubham-M-Rathod (commented):
Am new to open-source and am open to any feedback to help me learn. Thanks

wetneb (changes requested):
Hi, welcome to OpenRefine and thanks a lot for tackling this!
I have tried it out and it works! I have a few small suggestions though:
* I would decrease the delay between hover and menu opening a little bit, it feels a bit unresponsive to me at the moment. Maybe 250-300ms would be sufficient.
* It looks weird to me that two different menu items can be highlighted in light blue at the same time. I would rather only highlight the item that is currently being <s>highlighted</s> hovered. That means, if another menu item still has its submenu open but is no longer hovered, I would remove its highlighting.

Here is a quick video to illustrate the problem of two menu items being highlighted at the same time:
[Capture vidéo du 2024-11-02 17-09-24.webm](https://github.com/user-attachments/assets/d5c9a908-95fd-4e10-871b-717137c98835)

Also, it looks like you have included changes to `package-lock.json`, `package.json` and `yarn.lock` (which you can see in the [Files tab of this pull request](https://github.com/OpenRefine/OpenRefine/pull/6965/files)). We try to keep those untouched, as dependency upgrades are made separately. You could restore those files to their original contents with `git checkout master main/webapp/package-lock.json` for instance.

Shubham-M-Rathod:
https://github.com/user-attachments/assets/aee873a6-eb1a-45f5-b4e1-c2d2d992bd4f

Thanks for the feedback and support. I tried to make changes accordingly. Set the delay to 300ms. It seems smooth to me now, as well as took care not to highlight multiple menu items.

wetneb:
That looks much better, thank you!

I have noticed another problem.
* When the mouse temporarily hovers another menu item that opens a submenu (such as "Edit column…"), your code works great: it will not immediately close the existing submenu
* However, if the menu item that is being temporarily hovered does not correspond to a submenu (such as "Sort…") then the existing submenu is closed immediately

You can see this behaviour in this video:
[Capture vidéo du 2024-11-03 07-21-20.webm](https://github.com/user-attachments/assets/e4112de0-5d46-490d-b8fa-a07782b1c198)

I think it would be better to have a consistent behaviour for all types of menu items.

Another small thing I thought about is that it could be simpler to apply the light blue color using a `:hover` CSS selector rather than by manually adding and removing a class via jQuery. I know that's not something you introduced, but it might make the code a bit simpler to move to CSS for that. Up to you!

Shubham-M-Rathod:
@wetneb 
Hey, Thanks for having patience with me.

https://github.com/user-attachments/assets/0ad2abe6-de85-4c28-9b7a-7191c15804cd

I have also removed unnecessary addition and removal of class via jquery.

wetneb (approved):
I love it. It makes such a difference!
Leaving open for a bit if others want to chime in, but merging soon by default.

PR #6966, opened 2024-11-01
Allow uploading new file versions

### LINKED ISSUE #5959: Support uploading new version of media files
In the Wikibase  extension, the media file upload feature does not make it possible to upload new versions of files.

### Proposed solution

If the user generates edits on matched cells where a file path is also provided, treat those edits as an upload of a new version of the file.
It is likely useful to generate a QA warning about this to make sure this is not done unintentionally.

### Alternatives considered

Decide to forbid this and make it clearer from the UI (via a QA warning for instance) that this cannot be done.

### Additional context

Reported on the forum: https://forum.openrefine.org/t/upload-new-image-version-to-existing-file-on-wikimedia-commons/852

sebastian-berlin-wmse:
If there's a match for a file an upload request will be made. The response should contain warnings since the file would already exist. If the only warnings are "exists" and "duplicateversions" another request will be made with `ignorewarnings=1`. This will upload the file as a new version. Allowing these warnings means that you can upload a file with the same name as one that already exists even if the new file is that same as and older version.

An issue is shown when a new version is about to be uploaded to prevent this happening by mistake.

Fixes #5959


sebastian-berlin-wmse:
Looks like the rebase didn't work as I thought it would😩

PR #6967, opened 2024-11-01
Allow uploading new file versions

### LINKED ISSUE #5959: Support uploading new version of media files

sebastian-berlin-wmse:
If there's a match for a file an upload request will be made. The response should contain warnings since the file would already exist. If the only warnings are "exists" and "duplicateversions" another request will be made with `ignorewarnings=1`. This will upload the file as a new version. Allowing these warnings means that you can upload a file with the same name as one that already exists even if the new file is that same as and older version.

An issue is shown when a new version is about to be uploaded to prevent this happening by mistake.

Fixes #5959

### DIFF
--- a/extensions/wikibase/module/langs/translation-en.json
+++ b/extensions/wikibase/module/langs/translation-en.json
@@ -288,6 +288,8 @@
     "warnings-messages/duplicate-file-names-in-batch/body": "The file names provided for the new media files should all be distinct. Some file names, such as <span class=\"wb-issue-preformat\">{example_filename}</span>, are used for more than one file.",
     "warnings-messages/file-names-already-exist-on-wiki/title": "Naming conflict between new and existing files on the wiki",
     "warnings-messages/file-names-already-exist-on-wiki/body": "The file names provided for the new media files should be distinct from any existing file names on the wiki. This is not the case for <span class=\"wb-issue-preformat\">{example_filename}</span>.",
+    "warnings-messages/upload-new-file-version/title": "Uploading new file version",
+    "warnings-messages/upload-new-file-version/body": "Some files, such as <span class=\"wb-issue-preformat\">{example_filename}</span>, have been matched to existing files. This will upload new versions of those files.",
     "warnings-messages/invalid-characters-in-file-name/title": "Invalid characters in file names",
     "warnings-messages/invalid-characters-in-file-name/body": "Filenames such as <span class=\"wb-issue-preformat\">{example_filename}</span> contain the character <span class=\"wb-issue-preformat\">{invalid_character}</span> which cannot be part of file names in a wiki.",
     "warnings-messages/file-name-too-long/title": "File names exceeding the maximum length",
--- a/extensions/wikibase/src/org/openrefine/wikibase/editing/EditBatchProcessor.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/editing/EditBatchProcessor.java
@@ -218,14 +218,19 @@ public class EditBatchProcessor {
                 // custom code for handling our custom updates to mediainfo, which cover editing more than Wikibase
                 if (entityUpdate instanceof FullMediaInfoUpdate) {
                     FullMediaInfoUpdate fullMediaInfoUpdate = (FullMediaInfoUpdate) entityUpdate;
+                    MediaFileUtils mediaFileUtils = new MediaFileUtils(connection);
+
+                    if (fullMediaInfoUpdate.getFilePath() != null) {
+                        ((MediaInfoEdit) update).uploadFile(mediaFileUtils, summary, tags, filePageWaitTime,
+                                filePageMaxWaitTime);
+                    }
+
                     if (fullMediaInfoUpdate.isOverridingWikitext() && fullMediaInfoUpdate.getWikitext() != null) {
-                        MediaFileUtils mediaFileUtils = new MediaFileUtils(connection);
                         long pageId = Long.parseLong(fullMediaInfoUpdate.getEntityId().getId().substring(1));
                         long revisionId = mediaFileUtils.editPage(pageId, fullMediaInfoUpdate.getWikitext(), summary, tags);
                         lastRevisionId = OptionalLong.of(revisionId);
                     } else {
                         // manually purge the wikitext page associated with this mediainfo
-                        MediaFileUtils mediaFileUtils = new MediaFileUtils(connection);
                         mediaFileUtils.purgePage(Long.parseLong(entityUpdate.getEntityId().getId().substring(1)));
                     }
                 }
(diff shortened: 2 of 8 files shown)


wetneb (commented):
Fantastic that you implemented this!

I haven't tried it out yet, but I have some initial questions on the overall design:
* from what I understand, when uploading a new version of a file, your code will do two uploads:
   * a first one without `ignorewarnings=1`, which will fail because of warnings
   * a second one with `ignorewarnings=1`, after having checked that all the warnings returned are indeed things we want to ignore because they are expected for a new file version
  
   The fact that we're uploading the same file twice is maybe not ideal. I imagine that you made it like this because you don't want to add `ignorewarnings=1` upfront, in case there are some other warnings that we don't want to ignore. Correct? Did you consider including `ignorewarnings=1` upfront but only when we know we are uploading a new version (with a matched reconciled cell)? Maybe that's an acceptable middle ground?
* the API action that uploads the file also takes the wikitext as parameter. If an empty wikitext is provided, will this erase the existing wikitext? As a user, if I want to upload new versions of files without changing the wikitext, it would be convenient if I could leave the wikitext field empty in the schema (otherwise I need to fetch the existing wikitext and add it again untouched, which is maybe a bit cumbersome). I haven't checked how this works, maybe that's already supported as is, so it's really a question :)

The CI failure is due to validation errors for the Javadocs:
```
 Error:  /home/runner/work/OpenRefine/OpenRefine/extensions/wikibase/src/org/openrefine/wikibase/updates/MediaInfoEdit.java:306: error: @param name not found
Error:       * @param filePageWaitTime
Error:                ^
Error:  /home/runner/work/OpenRefine/OpenRefine/extensions/wikibase/src/org/openrefine/wikibase/updates/MediaInfoEdit.java:308: error: @param name not found
Error:       * @param filePageMaxWaitTime
Error:                ^
```

sebastian-berlin-wmse:
> * from what I understand, when uploading a new version of a file, your code will do two uploads:
>    * a first one without `ignorewarnings=1`, which will fail because of warnings
>    * a second one with `ignorewarnings=1`, after having checked that all the warnings returned are indeed things we want to ignore because they are expected for a new file version
>   
>    The fact that we're uploading the same file twice is maybe not ideal. I imagine that you made it like this because you don't want to add `ignorewarnings=1` upfront, in case there are some other warnings that we don't want to ignore. Correct? Did you consider including `ignorewarnings=1` upfront but only when we know we are uploading a new version (with a matched reconciled cell)? Maybe that's an acceptable middle ground?

The reason I did it this way was to err on the side of caution. [The documentation on warnings](https://www.mediawiki.org/wiki/API:Upload#Upload_warnings) isn't very clear, even about what warnings there are. I would think that at least `duplicate-archive` you don't want to ignore since that means you'd upload the same file that has explicitly been deleted earlier.

sebastian-berlin-wmse:
> The CI failure is due to validation errors for the Javadocs:
> ```
>  Error:  /home/runner/work/OpenRefine/OpenRefine/extensions/wikibase/src/org/openrefine/wikibase/updates/MediaInfoEdit.java:306: error: @param name not found
> Error:       * @param filePageWaitTime
> Error:                ^
> Error:  /home/runner/work/OpenRefine/OpenRefine/extensions/wikibase/src/org/openrefine/wikibase/updates/MediaInfoEdit.java:308: error: @param name not found
> Error:       * @param filePageMaxWaitTime
> Error:                ^
> ```

Thanks. There were so many errors I wasn't sure where to start, but that seems to have fixed them all🤨

sebastian-berlin-wmse:
I did some more digging and apparently when a file upload generates warnings the file kept in the upload stash. This means we shouldn't have to upload it again, but instead give `filekey` in the second request.

wetneb:
@sebastian-berlin-wmse very nice! Looks like we just need to solve the conflicts and it should be good to go :)

sebastian-berlin-wmse:
Again I was having issues with Git and couldn't get it to apply commits in the order I wanted for rebasing. I ended up force pushing which may make it harder to follow the changes in the review. Sorry about that.

wetneb (commented):
Thank you, force pushing isn't a problem as far as I am concerned.

If I am reading your changes correctly, it seems that we will now overwrite any existing file regardless of whether the reconciled cell was marked as "new" or matched to an existing cell. Is that true?

I was intuitively assuming that we would ignore the warnings only when the cell was originally matched to an existing file. This would avoid the following scenario:
* the user bypasses reconciliation because they assume that none of the file names exist on the wiki. They directly mark all cells as reconciled to "new".
* one of the filenames actually corresponds to an existing file
* they don't get any warnings about uploading new versions of files, given that all their cells are reconciled to new
* in the upload process, warnings about files getting overwritten are ignored, so the user ends up overwriting an existing file without wanting to

sebastian-berlin-wmse:
> If I am reading your changes correctly, it seems that we will now overwrite any existing file regardless of whether the reconciled cell was marked as "new" or matched to an existing media entity. Is that true?

Yes, there's no extra check for whether the cell is "new" or not. However, if a new file version is uploaded also depends on scrutinisers and API warnings. If I read your scenarios below correctly they should still not lead to uploads.

> I was intuitively assuming that we would ignore the warnings only when the cell was originally matched to an existing file. This would avoid the following scenario:
> 
>    * the user bypasses reconciliation because they assume that none of the file names exist on the wiki. They directly mark all cells as reconciled to "new".
>    * one of the filenames actually corresponds to an existing file

If a file name already exists it will be caught by `FileNameScrutinizer`.

>    * they don't get any warnings about uploading new versions of files, given that all their cells are reconciled to new
>    * in the upload process, warnings about files getting overwritten are ignored, so the user ends up overwriting an existing file without wanting to

The only warnings that are ignored are `exists`, which is raised when there is a file with the same name, and `duplicateversions`, which is raised when a file with the same name exists and the same file (as in identical file content) is an old version of that file. Any other warnings will still lead to aborting the upload.

wetneb (approved):
Ok, I forgot FileNameScrutinizer did that, so it feels less important to me. I still think it would be worth doing, but I can do that as a separate PR.

mvolz:
So, I am getting this warning and I DON'T want to upload new versions - I want it to only upload the new files that don't exist. (I had a few files fail to upload due to bad filename and only want to upload these!)

Is there a setting or something to prevent this?

wetneb:
Using a judgment facet to keep only the cells reconciled to a new item, not the matched ones?

PR #6969, opened 2024-11-04
Inadequate rate-limiting when editing Wikimedia Commons

### LINKED ISSUE #6055: Inadequate rate-limiting when editing Wikimedia Commons
When editing files on Wikimedia Commons, it is possible to encounter rate-limits which imply that edits are not made on the wiki.

### To Reproduce
Steps to reproduce the behavior: see the [forum post](https://forum.openrefine.org/t/openrefine-pause-edit-for-wikidata-edits-but-not-for-wikimedia-edits/972).

### Current Results
Edits are skipped.

### Expected Behavior
The process waits before attempting edits, meaning that edits are successful.

### Versions
 - OpenRefine: 3.8-SNAPSHOT

### Additional context

Edits to Wikimedia Commons involve a bit more operations than on Wikidata, since separate API actions are required to upload the file (if any), edit the structured data and edit or update the wikitext. So it's not so surprising that we are running into a stricter rate-limiting than on Wikidata. The thresholds in use on both wikis might be different too. We need to investigate what those thresholds are and enforce them properly. Also, we should make sure that failed edits are detected properly and re-attempted.

sunilnatraj:
Added retry & backoff settings in Wiki data add / update / delete flows

Fixes #6055 

- Add / Update flows - added retry and exponential backoff settings on WikibaseDataEditor
- MediaFileUtils - added retry support in purgePage flow 
- Updated purge test case to reflect retry support

### DIFF
--- a/extensions/wikibase/src/org/openrefine/wikibase/editing/EditBatchProcessor.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/editing/EditBatchProcessor.java
@@ -119,6 +119,10 @@ public class EditBatchProcessor {
         editor.setAverageTimePerEdit(maxEditsPerMinute <= 0 ? 0 : (int) (1000 * (maxEditsPerMinute / 60.)));
         // set maxlag based on preference store
         editor.setMaxLag(maxLag);
+        // configure retry limit, backoff settings
+        editor.setMaxLagMaxRetries(3);
+        editor.setMaxLagFirstWaitTime(1000);
+        editor.setMaxLagBackOffFactor(2.0);
 
         this.library = library;
         this.summary = summary;
--- a/extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
@@ -78,7 +78,19 @@ public class MediaFileUtils {
         parameters.put("action", "purge");
         parameters.put("pageids", Long.toString(pageid));
 
-        apiConnection.sendJsonRequest("POST", parameters);
+        int retries = 3;
+        MediaWikiApiErrorException lastException = null;
+        while (retries > 0) {
+            try {
+                JsonNode response = apiConnection.sendJsonRequest("POST", parameters);
+            } catch (MediaWikiApiErrorException e) {
+                lastException = e;
+            }
+            retries--;
+        }
+        if (retries <= 0 && lastException != null) {
+            throw lastException;
+        }
     }
 
     /**
--- a/extensions/wikibase/tests/src/org/openrefine/wikibase/editing/MediaFileUtilsTest.java
+++ b/extensions/wikibase/tests/src/org/openrefine/wikibase/editing/MediaFileUtilsTest.java
@@ -83,7 +83,7 @@ public class MediaFileUtilsTest {
         Map<String, String> params = new HashMap<>();
         params.put("action", "purge");
         params.put("pageids", "12345");
-        verify(connection, times(1)).sendJsonRequest("POST", params);
+        verify(connection, times(3)).sendJsonRequest("POST", params);
     }
 
     @Test


sunilnatraj:
@wetneb Please review

sunilnatraj:
@wetneb  In the data update flow, the APIConnection log entries have the the context as Purge. is there a miss in the logging?
[..kibaseapi.ApiConnection] API warning [purge]: You've exceeded your rate limit. Please wait some time and try again.

wetneb:
Thanks!

The scenario mentioned [on the forum](https://forum.openrefine.org/t/openrefine-pause-edit-for-wikidata-edits-but-not-for-wikimedia-edits/972) is about editing the wikitext of Commons files, not their captions, so I think it would be worth adapting your test scenario to match that. The API actions used to edit captions and the wikitext are different.

> I ddi set the disableEditing for this test.

I'm not sure how we could validate the resolution of the problem if editing is disabled - I would do a live test in this case, making sure I don't leave any bad data behind. From your description of your testing, it's also not clear to me if you tested before your changes (to confirm you can reproduce the issue) and after your changes (to confirm that they make the expected difference)?

> is there a miss in the logging?

I don't understand the question.

wetneb:
I'm also adding @sebastian-berlin-wmse in the loop as WMSE has ranked this issue quite high in their roadmap if I recall correctly.

sunilnatraj:
> [..kibaseapi.ApiConnection] API warning [purge]: You've exceeded your rate limit. Please wait some time and try again.

@wetneb My observation is that the logging in Wikibase.ApiConnection is printing "Purge" even when the action is Edit.

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -78,7 +78,19 @@ public void purgePage(long pageid) throws IOException, MediaWikiApiErrorExceptio
...
-        apiConnection.sendJsonRequest("POST", parameters);
+        int retries = 3;
+        MediaWikiApiErrorException lastException = null;
+        while (retries > 0) {
+            try {
+                JsonNode response = apiConnection.sendJsonRequest("POST", parameters);
+            } catch (MediaWikiApiErrorException e) {
+                lastException = e;
+            }
+            retries--;
```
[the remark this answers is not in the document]
sunilnatraj: Updated flow to handle rate limit failure.

extensions/wikibase/tests/src/org/openrefine/wikibase/editing/MediaFileUtilsTest.java
```diff
@@ -83,7 +83,7 @@ public void testPurge() throws IOException, MediaWikiApiErrorException {
         Map<String, String> params = new HashMap<>();
         params.put("action", "purge");
         params.put("pageids", "12345");
-        verify(connection, times(1)).sendJsonRequest("POST", params);
+        verify(connection, times(3)).sendJsonRequest("POST", params);
```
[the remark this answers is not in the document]
sunilnatraj: Reverted the earlier change as with the updated method this test works fine. Will check on how to add a test case for the rate limit case.

sunilnatraj:
@wetneb @sebastian-berlin-wmse I have updated the page edit & purge flows and added retry capability. My understanding is the rate limit is per minute the retry & exponential backoff is to try 5 time with increased wait time to ensure the request is successful. 
Tested the flows with updates to Wikitext, also tested with the "override existing wikitext" set and  unset. 
Attached console log pre fix and post fix.
[postFix-TestLog-1.log](https://github.com/user-attachments/files/17633000/postFix-TestLog-1.log)
[pre-Test-Run1.log](https://github.com/user-attachments/files/17633001/pre-Test-Run1.log)

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -77,8 +78,31 @@ public void purgePage(long pageid) throws IOException, MediaWikiApiErrorExceptio
...
+            try {
+                JsonNode response = apiConnection.sendJsonRequest("POST", parameters);
+                if (!(response != null && response.has("warnings"))) {
+                    return;
+                }
+            } catch (MediaWikiApiErrorException | IOException e) {
+            }
+            retries--;
+            if (retries > 0) {
+                backofTime = retries == 4 ? maxLagWaitTime : maxLagWaitTime * backoffFactor;
```
wetneb, line 93: This means that we are going to sleep for `maxLagWaitTime` the first time and then `maxLagWitTime * backoffFactor`. That's not quite what Wikidata-Toolkit does: an exponential backoff strategy, where the wait time will increase by a fixed factor at each new attempt.
The reason why this strategy is preferable is that it helps the web service reach a load equilibrium. You can read more about this here: https://en.wikipedia.org/wiki/Exponential_backoff#Rate_limiting
You can see how it is implemented in Wikidata-Toolkit:
https://github.com/Wikidata/Wikidata-Toolkit/blob/a9b1fb4d4cc36d7441d9a7e6cc5af02979bd3da1/wdtk-wikibaseapi/src/main/java/org/wikidata/wdtk/wikibaseapi/WbEditingAction.java#L743-L764

I'd like to use the same strategy here for consistency.

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -77,8 +78,31 @@ public void purgePage(long pageid) throws IOException, MediaWikiApiErrorExceptio
...
         parameters.put("pageids", Long.toString(pageid));
-
-        apiConnection.sendJsonRequest("POST", parameters);
+        int retries = 5;
+        int backofTime = maxLagWaitTime;
+        while (retries > 0) {
+            try {
+                JsonNode response = apiConnection.sendJsonRequest("POST", parameters);
+                if (!(response != null && response.has("warnings"))) {
+                    return;
```
wetneb, line 87: Not every warning is necessarily a rate-limiting one. Is there a warning code we could rely on to check if it's a rate-limiting one? There might be warnings that are not worth retrying (such as insufficient editing permissions, lack of authentication…).

wetneb (commented):
Thanks a lot, that looks a lot better.  Did you consider writing tests for it?
You should be able to enqueue multiple mocked responses to a mock ApiConnection, so that the first one contains a warning, and the seconds succeeds for instance.

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
sunilnatraj: I will check this

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
sunilnatraj: There is no code; i did not include the check for the rate limit as I was not sure if the warning text is internationalised. Any guidance on this?

sunilnatraj:
> Thanks a lot, that looks a lot better. Did you consider writing tests for it? You should be able to enqueue multiple mocked responses to a mock ApiConnection, so that the first one contains a warning, and the seconds succeeds for instance.

@wetneb  Yes i am working on the unit test

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
wetneb: I agree, if it's just text it feels too brittle to rely on it indeed.

extensions/wikibase/src/org/openrefine/wikibase/editing/EditBatchProcessor.java
```diff
@@ -119,6 +119,10 @@ public EditBatchProcessor(WikibaseDataFetcher fetcher, WikibaseDataEditor editor
         editor.setAverageTimePerEdit(maxEditsPerMinute <= 0 ? 0 : (int) (1000 * (maxEditsPerMinute / 60.)));
         // set maxlag based on preference store
         editor.setMaxLag(maxLag);
+        // configure retry limit, backoff settings
+        editor.setMaxLagMaxRetries(5);
+        editor.setMaxLagFirstWaitTime(5000);
+        editor.setMaxLagBackOffFactor(2.0);
```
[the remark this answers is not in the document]
sunilnatraj: @wetneb 
- Removed the WikibaseEditor exponential backoff setting from EditBatchProcessor
- In MediaFileUtils - setting the retry limit as 14 and BackoffFactor as 1.5
- Updated the logic to do exponential backoff 
- Added exponential backoff in the uploadFile method. 
- The retry limit here is 3 - should we make this also consistent with page Edit & purge calls?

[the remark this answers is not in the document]
sunilnatraj: Also formatted the log entry
03:15:20.256 [...editing.MediaFileUtils] -- purgePage:API error. Attempts left 13. Pausing 5 secs before retry. (357ms)
03:15:25.735 [...editing.MediaFileUtils] -- purgePage:API error. Attempts left 12. Pausing 7 secs before retry. (5479ms)

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -77,8 +79,32 @@
...
+            if (retries > 0) {
+                logger.info(String.format("-- purgePage:API error. Attempts left %d. Pausing %d secs before retry.", retries,
+                        backofTime / 1000));
+                try {
+                    Thread.sleep(backofTime);
+                } catch (InterruptedException e1) {
+                    Thread.currentThread().interrupt();
+                    retries = 0;
+                }
+                backofTime *= backoffFactor;
```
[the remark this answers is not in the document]
sunilnatraj: Fixed

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -311,11 +350,12 @@
             } catch (MaxlagErrorException e) { // wait for 5 seconds
                 lastException = e;
                 try {
-                    Thread.sleep(maxLagWaitTime);
+                    Thread.sleep(backofTime);
                 } catch (InterruptedException e1) {
                     Thread.currentThread().interrupt();
                     retries = 0;
                 }
+                backofTime *= backoffFactor;
```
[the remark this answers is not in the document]
sunilnatraj: Fixed

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -77,8 +79,32 @@ public void purgePage(long pageid) throws IOException, MediaWikiApiErrorExceptio
         Map<String, String> parameters = new HashMap<>();
         parameters.put("action", "purge");
         parameters.put("pageids", Long.toString(pageid));
-
-        apiConnection.sendJsonRequest("POST", parameters);
+        int retries = maxRetries;
+        long backofTime = maxLagWaitTime;
```
wetneb, line 83: `backof` should be `backOff` (note the two "f")

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -77,8 +79,32 @@ public void purgePage(long pageid) throws IOException, MediaWikiApiErrorExceptio
...
+            try {
+                JsonNode response = apiConnection.sendJsonRequest("POST", parameters);
+                if (!(response != null && response.has("warnings"))) {
+                    return;
+                }
+            } catch (MediaWikiApiErrorException | IOException e) {
+            }
+            retries--;
+            if (retries > 0) {
+                logger.info(String.format("-- purgePage:API error. Attempts left %d. Pausing %d secs before retry.", retries,
```
wetneb, line 94:
```suggestion
                logger.info(String.format("-- purgePage:API error. %d attempts left. Pausing %d secs before retry.", retries,
```

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -258,6 +285,17 @@ public long editPage(long pageId, String wikitext, String summary, List<String>
                 lastException = e;
             }
             retries--;
+            if (retries > 0) {
+                try {
+                    logger.info(String.format("-- editPage:API error. Attempts left %d. Pausing %d secs before retry.", retries,
```
wetneb, line 290:
```suggestion
                    logger.info(String.format("-- editPage:API error. %d attempts left. Pausing %d secs before retry.", retries,
```

wetneb (commented):
Very nice, thank you!

To make sure that the tests don't take too long to execute, I would propose to make the maxLagWaitTime in MediaFileUtils configurable. This way, you could set it to a low value (say 100ms) in the tests, so that they execute faster.

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
sunilnatraj: Corrected typo

sunilnatraj:
> Very nice, thank you!
> 
> To make sure that the tests don't take too long to execute, I would propose to make the maxLagWaitTime in MediaFileUtils configurable. This way, you could set it to a low value (say 100ms) in the tests, so that they execute faster.

@wetneb Updated the unit test to set maxLagTime to 100ms.

wetneb (approved):

wetneb:
Brilliant! That looks good to me. Leaving it open for a bit longer to give a chance to @sebastian-berlin-wmse to chime in :)

wetneb:
Also pinging @Vesihiisi in case they are interested in testing this out :)

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
sebastian-berlin-wmse: According to https://www.mediawiki.org/wiki/API:Etiquette#Request_limit the response should contain an error (not a warning) with the code `ratelimited`.

sunilnatraj: @sebastian-berlin-wmse  As mentioned in the link  API returns an error only when it is a edit call --> _Requests which make edits, modify state or otherwise are not read-only requests, are subject to rate limiting._

In case of Delete / Purge call the rate limit is returned as a warning.

sebastian-berlin-wmse: Looks like it. I didn't think `purge` was counted as read-only, but apparently it does.

You can still get a code for the warning if you add parameter `errorformat=raw`. That'll give you a response like:
```
{
  "batchcomplete": "",
  "warnings": [
    {
      "code": "ratelimited",
      "key": "apierror-ratelimited",
      "params": [],
      "module": "purge"
    }
  ],
  "purge": [
    {
      "ns": 0,
      "title": "Main Page"
    }
  ]
}
```

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
```diff
@@ -49,6 +49,8 @@ public class MediaFileUtils {
     protected String csrfToken = null;
     protected int maxLagWaitTime = 5000; // ms
     protected String filePrefix = "File:"; // configurable?
+    protected double backoffFactor = 1.5;
```
sebastian-berlin-wmse, line 52: What's the maximum total wait time using these values? I don't know how to do the math quickly, but staring at 5 seconds and increasing by half up to 14 time sounds like you'd end up with a pretty big number.

sunilnatraj: Max wait time 1598 seconds

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
sunilnatraj: Updated code to pass param errorformat and validating code in response to identify if it is a rate limit error.

sunilnatraj:
@wetneb @sebastian-berlin-wmse

extensions/wikibase/src/org/openrefine/wikibase/editing/MediaFileUtils.java
wetneb: I wouldn't mark this conversation as resolved, as there could still be the decision to change some of those parameters to get a shorter maximum wait.

wetneb:
I think it's to be expected that @sebastian-berlin-wmse takes a bit longer to respond, as they also work on other projects. Don't worry, I'm confident that this PR will be merged.

wetneb (approved):
Two weeks have passed, let's merge it as is. We can always make adjustments to the waiting time later on.

PR #6974, opened 2024-11-05
grel: Translate expressions after a column rename

wetneb:
This adds support for transforming GREL expressions to adapt them after a column rename.
This interface is also made available to other expression languages, in a backwards compatible way thanks to a default implementation which treats all expressions as un-analyzable by default.

The goal is to make it possible to reliably adapt an history recipe to a new project when column names are different. Similar translation methods would be introduced on operation classes. It could also be useful to solve #133.

### DIFF
--- a/modules/core/src/main/java/com/google/refine/expr/Evaluable.java
+++ b/modules/core/src/main/java/com/google/refine/expr/Evaluable.java
@@ -91,16 +91,19 @@ public interface Evaluable {
     }
 
     /**
-     * Translates this expression by simultaneously substituting column names, as specified by the supplied map.
+     * Translates this expression by simultaneously substituting column names, as specified by the supplied map. This
+     * transformation is done on a best-effort basis. For instance, in cases where column dependencies cannot be
+     * isolated, the resulting expression could still rely on columns via their old names. The goal of this method is to
+     * rename as many references as possible. As such, it is always appropriate to return the original expression as a
+     * fall-back.
      *
      * @param substitutions
      *            a map specifying new names for some columns. Keys of the map are old column names, values are the new
      *            names for those columns. If a column name is not present in the map, the column is not renamed.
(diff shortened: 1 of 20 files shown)


modules/core/src/main/java/com/google/refine/expr/Evaluable.java
```diff
@@ -66,4 +67,17 @@ public default Optional<Set<String>> getColumnDependencies(Optional<String> base
...
+    /**
+     * Translates this expression by simultaneously substituting column names, as specified by the supplied map.
+     *
+     * @param substitutions
+     *            a map specifying new names for some columns. Keys of the map are old column names, values are the new
+     *            names for those columns. If a column name is not present in the map, the column is not renamed.
+     * @return a new expression with updated column names. If this renaming isn't supported, {@link Optional#empty()}
+     *         should be returned.
+     */
+    public default Optional<Evaluable> renameColumnDependencies(Map<String, String> substitutions) {
```
wetneb, line 79: After solving #133 based on this PR, I realized that the current approach of either renaming fully an operation or bailing out by returning `Optional.none()` isn't ideal. It feels more useful to me to have this function work on a best-effort basis, renaming as many column references as possible, but always producing an output.

In the case of renaming facets, you generally want to keep all your facets and have them renamed as fully as possible. Any remaining references to the former column should show up as facets with evaluation errors, which can then be fixed or deleted manually.

So I have a follow-up commit which changes that (changing the return type to `Evaluable` instead of `Optional<Evaluable>`), which has the benefit of simplifying the logic in a lot of implementations of this method. It depends on #6929 so I'll add this commit here once that one is merged.

wetneb: I've added the follow-up commit mentioned above: 0dd612286337e0125f973337a8667d1e4beb75fe

wetneb:
I plan to merge this soon.

PR #6958, opened 2024-10-28
gdata: Remove extension, migrating it to a standalone repository

wetneb:
As mentioned in #6931, I think it would be worth un-bundling the gdata extension, because the setup required to use it (getting Google API credentials) is unfortunately way too involved for most of our users (including myself).

A draft of the standalone repository is now at:
https://github.com/wetneb/refine-gdata-extension
I have extracted it from this repository, preserving commit history and attribution. I have added a few changes on top to make it build as a standalone Maven project.

I would propose to migrate the repository to the OpenRefine organization if this PR is merged.

I have locked the Weblate translations for this module, so that we don't lose translations in the migration process.

List of follow-up actions:
* [ ] update the README of the extension to reflect its current state
* [ ] migrate issues from https://github.com/OpenRefine/OpenRefine/issues?q=is%3Aopen+is%3Aissue+label%3Agoogle to the new repository
* [ ] migrate [the wiki page with instructions to get the Google API credentials](https://github.com/OpenRefine/OpenRefine/wiki/Google-Extension) to the README of the extension (for instance)
* [ ] make a first release of the extension
* [ ] re-enable Weblate integration, pointing to the new repository
* [ ] enable dependabot too
* [ ] add the extension to https://openrefine.org/extensions

Issue #5282, opened 2022-09-22
Move Wikibase extension to its own repository


wetneb:
Our integration with Wikibase is developed as an extension, whose source code lives in this repository and which is shipped by default with OpenRefine.

This extension sees quite some activity and is quite popular in the Wikimedia and Wikibase communities. There are expectations about its maintenance and development and @trnstlntk has been active in encouraging various organizations in this scene to take responsibility for this integration.

On my side, I have planned to reduce my activity in this domain, because I notice it is keeping me away from important improvements to OpenRefine's core. In a context where I am trying to get more developers onboard the project, it feels like it is easier to have them take responsibility for a part of the project that matters to a strong and organized community of users (typically if those developers are to be hired by a partner). Organizations are less likely to foster work on generic parts of OpenRefine, because they want their energy to be focused on their own community.

In this context, moving the Wikibase extension to its own repository would have the benefit of making it easier for other developers to take ownership of this integration. I can also see other advantages:
* we would be reducing the size of this repository, letting us concentrate on a smaller set of features
* the CI would run faster, the issue and PR backlog would get smaller

By default, this move would mean that users would have to install the extension manually. I think this would be a good motivation for us to improve the user experience in installing extensions. There are various proposals around that:
* https://github.com/OpenRefine/OpenRefine/issues/3223
* https://github.com/OpenRefine/OpenRefine/issues/3326

### Proposed solution

Move the extension to its own repository and let people install it manually.

### Alternatives considered

As a temporary measure, we could keep shipping it by default in our releases (by dynamically downloading the built extension at packaging time and unpacking it in the extensions directory).

This is a tempting workaround given how brittle our extension mechanism is, but it would have some downsides:
* we need to make sure the extension we ship is actually compatible with the core software, which requires some coordination with the maintainers of the extension. That could potentially block a release (we would either have to wait for the Wikibase extension to be fixed after introducing a breaking change or make a new release without it)
* it would postpone the improvements to the extension install UX, which feel very necessary in my opinion

Abbe98:
Just moving it to its own repository and not serve users are not my intent nor a resonable way forward in my opinion, bundling the Wikimedia extensions together in a dedicated distribution serves users better than we do today. Separering the repositories would also make it easier not only to show the state of things but it also becomes easier to highlight eventual work and contributors.

Abbe98:
I'm curious what thought @thadguidry and @tfmorris might have.

thadguidry:
[...]

Issue #3223, opened 2020-09-26
install extensions from GUI


msaby:
Extensions are not hard to install, but it would be a little easier for newcomers if they could be installed and uninstalled from GUI.

tfmorris:
Presumably this should include the full life cycle similar to browser extensions with the ability to enable/disable, configure, and remove extensions as well.

antoine2711:
What would be the source for the integrated extensions installer? Could we have a repo on OpenRefine that would contain a copy of all the installable extensions? That way, we could provide some kind of control/quality of the extensions, and it would be easier to control versions installed, and their compatibility.

I also think we should have the same kind of installation commodity for services.

wetneb:
> What would be the source for the integrated extensions installer? Could we have a repo on OpenRefine that would contain a copy of all the installable extensions? That way, we could provide some kind of control/quality of the extensions, and it would be easier to control versions installed, and their compatibility.

I think that would go against the spirit of extensions: anyone can develop them and install them without having to go through vetting by a central authority.
We could certainly have an online catalogue of known extensions, but people should not be forced to go through that, I think.

antoine2711:
> I think that would go against the spirit of extensions: anyone can develop them and install them without having to go through vetting by a central authority. We could certainly have an online catalogue of known extensions, but people should not be forced to go through that, I think.

@wetneb: I'm totally like you for not locking people in this hypothetical extension repo.

I think we should have a file, like a JSON, with a list of all known extensions. These extensions could be anywhere. 

But, for the extensions that are very common, or the one that no longer have maintenance, having them in a repo could gives us better quality control. Also, since a change in the core might trigger changes in extensions, and those change would be the sames across extensions, then working them in batch might be efficient.

And, since you can put many .git repos in a single repo, we wouldn't loose the ability to send back these changes to thier respective repos.

And that way, it would also be easier to do a CD/CI that would test them automatically with newer version of OR.

Regards, Antoine

lozanaross:
> Presumably this should include the full life cycle similar to browser extensions with the ability to enable/disable, configure, and remove extensions as well.

Reading this after some time and just wondering if my hypothetical GUI needs to include some "configure" button that opens up more options via a dialog - wouldn't that be quite hard because we would need to force extension developers to expose somewhere possible configs that we would then render in the GUI? Maybe it's easy and I'm just finding it hard to imagine, so wanted to inquire further.

ostephens:
[...]

PR #6975, opened 2024-11-06
Add an extension mangement page

Abbe98:
This is a proof-of-concept for a poor man's extension manager. It allows users to see installed(loaded) extensions(modules), quickly open the extension directory, quickly discover extensions, and quickly access the documentation. My only intent atm is to test this with users and see if this can mitigate user concerns about the barriers of the existing extension installation process.

Screenshot 2024-11-06 at 16-25-32 OpenRefine

Related but not more than that:  #3223
In the light of #6958, #5282 and similar issues raised in the past years.

It's a PoC, i18n is missing, it does not use jQuery, one would need to pair something like this with a documentation update, etc. Not much extension metadata is displayed(requires Butterfly changes)...


magdmartin:
@Abbe98 Thanks. This is a great first step to improve extension management and address discoverability issues.

wetneb:
Very nice! I think this would be a very welcome addition.

For the left-hand side menu entry, I would propose "Extensions" instead of "Manage extensions".

wetneb:
Looks good to me!

One small thing: I noticed that all three buttons at the top are marked as "primary" (meaning that their text is in bold). To me, the purpose of marking a button as primary is to make it stand out in comparison to its neighbours as the sensible default thing to click, so if all buttons in an area are marked as primary, it feels a bit like they are all fiercely competing and shouting to be clicked.

So I'd rather remove the primary class on all of them, or leave one as primary if you want to highlight it more than others. The "documentation" and "discover extensions" ones could perhaps be shown as links instead, but then I understand laying them out nicely is perhaps more tricky (although with a bit of padding it shouldn't be too bad I think).

Abbe98:
I never intended for the `primary` style to be there, I'm just used to the primary/secondary pattern from Bootstrap and other frameworks... Fixed now.

wetneb (approved):
Thank you again!

magdmartin:
thank you @Abbe98 

@tfmorris will it be part of the 3.9 release?

sunilnatraj:
@Abbe98 Application start fails as extensionsDir is null. the exists check fails. extensionsDir is being initialized for Unix OS only.


Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.io.File.exists()" because "extensionsDir" is null
	at com.google.refine.RefineServer.getDataDir(Refine.java:446)
	at com.google.refine.RefineServer.configure(Refine.java:337)

wetneb:
Sorry about failing to catch that!

tfmorris:
> @tfmorris will it be part of the 3.9 release?

It certainly can be since I haven't cut it yet, but I need to get back on top of release planning.

PR #6981, opened 2024-11-08
Assign different cascade layers to CSS from different modules

Abbe98:
This assigns CSS from each module into a different cascade layer while also ensuring that the "core" layer takes priority.

This makes it so that CSS from core got a higher priority than that of extensions. It does not eliminate extensions ability to write over core styles(they might use `!important`) and they can still change rules and defaults provided by the browser, but it makes it slightly harder accidentally break core styles.

### DIFF
--- a/main/webapp/modules/core/MOD-INF/controller.js
+++ b/main/webapp/modules/core/MOD-INF/controller.js
@@ -693,11 +693,18 @@ function process(path, request, response) {
         for (var key in styles) {
           if (styles.hasOwnProperty(key)) {
             var qualifiedPath = styles[key];
-            styleInjection.push(
-                '<link type="text/css" rel="stylesheet" href="' + qualifiedPath.fullPath.substring(1) + '" />');
+            var styleSource;
+            if (qualifiedPath.fullPath.substring(1).startsWith("extension/")) {
+                styleSource = qualifiedPath.fullPath.split("/")[2];
+            } else {
+                styleSource = "core";
+            }
+
+
+            styleInjection.push('@import url(' + qualifiedPath.fullPath.substring(1) + ') layer(' + styleSource + ');');
           }
         }
-        context.styleInjection = styleInjection.join("\n");
+        context.styleInjection = '<style>@layer core;\n' + styleInjection.join("\n") + '</style>';
 
         if (bundle) {
           context.scriptInjection = '<script type="text/javascript" src="' + path + '-bundle.js"></script>';


Abbe98:
Someone more familiar than me with the bundled extension should check if there is any breakage from this(which probably means we have styles which should move from an extension to core).

wetneb:
I had no idea this was possible - looks like a very good fit.

For ease of reviewing, this is what the generated CSS looks like:
```css
@layer core;
@import url(externals/jquery-ui/css/ui-lightness/jquery-ui.css) layer(core);
@import url(3rdparty/select2/select2.css) layer(core);
@import url(3rdparty/tablesorter/theme.blue.css) layer(core);
@import url(styles/theme.css) layer(core);
@import url(styles/jquery-ui-overrides.css) layer(core);
...
@import url(styles/index/json-parser-ui.css) layer(core);
@import url(styles/index/wikitext-parser-ui.css) layer(core);
@import url(extension/database/styles/jquery.contextMenu.css) layer(database);
@import url(extension/database/styles/database-import.css) layer(database);
@import url(extension/gdata/styles/importing-controller.css) layer(gdata);
```

I had a quick look at how various parts of the UI defined by the extensions look like, didn't notice anything odd.
I'd be keen to merge this as is, because if anything, it's more likely to bring unexpected fixes than unexpected bugs (see for instance the CSS leaks from the database extension we had earlier).

Abbe98:
> I had no idea this was possible - looks like a very good fit.

It reached baseline ~two months ago :-) 

> I'd be keen to merge this as is, because if anything, it's more likely to bring unexpected fixes than unexpected bugs (see for instance the CSS leaks from the database extension we had earlier).

I'm aware of one issue over at the Commons repository as it unintentionally styles all import pages, but this would solve more bugs than it causes and I would personally be happy to follow up with a proper fix over in that repository.

wetneb:
> It reached baseline quite recently and the HTML specification has yet to catch up. :-)

ah yes, https://caniuse.com/css-cascade-layers says it works on 95% of the population. I don't know if it's good enough, maybe we should have a clear policy about it. I wonder what happens when it doesn't - do you know we just lose the prioritization or do we fail to include the CSS entirely?

Abbe98:
> > It reached baseline quite recently and the HTML specification has yet to catch up. :-)
> 
> ah yes, https://caniuse.com/css-cascade-layers says it works on 95% of the population. I don't know if it's good enough, maybe we should have a clear policy about it. I wonder what happens when it doesn't - do you know we just lose the prioritization or do we fail to include the CSS entirely?

Those 5% is mostly IE and Safari on iOS, non of which are supported by OR right now. Honestly, being more conservative than Baseline would do more harm than good to the project.

wetneb (approved):
I think we can go ahead - worst case, it can be reverted without big implications.

PR #6987, opened 2024-11-12
Add jquery-ui@1.12.1 to be loaded via npm

### LINKED ISSUE #6909: jquery-ui: switch to fetching it via NPM instead of using a vendored version?

mathdebate09:
Fixes #6909

- Add `jquery-ui` to be downloaded via `npm`
- Remove `jquery-ui` mention for `postinstall`
- [NITS] Add text for https://github.com/odyniec/imgareaselect as [@odyniec/imgareaselect](https://github.com/odyniec/imgareaselect) (no-bare-urls practice)

### DIFF
--- a/main/webapp/README.md
+++ b/main/webapp/README.md
@@ -8,8 +8,7 @@ Dependencies available in the npm registry are added to package.json.
 
 A few of the dependencies are not available in npm.
 
-* jquery-ui: packages are generated from jquery-ui website (https://jqueryui.com/download/#!version=1.12.1).
-* imgAreaSelect: cannot find 0.9.2 version in npm, found 0.9.11-rc.1 on GitHub (https://github.com/odyniec/imgareaselect).
+* imgAreaSelect: cannot find 0.9.2 version in npm, found 0.9.11-rc.1 on GitHub [@odyniec/imgareaselect](https://github.com/odyniec/imgareaselect).
 * suggest (4.3a): not in npm, contains fixes from Google Code SVN.
 
 These dependencies are located under `webapp/modules/core/externals`.
--- a/main/webapp/package-lock.json
+++ b/main/webapp/package-lock.json
@@ -10,6 +10,7 @@
         "@wikimedia/jquery.i18n": "1.0.9",
         "jquery": "3.7.1",
         "jquery-migrate": "3.5.2",
+        "jquery-ui": "^1.12.1",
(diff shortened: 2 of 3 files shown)


mathdebate09:
I believe more stuff is redundant but I'm not sure if it'll break things
- [ ] remove the folder [main/webapp/modules/core/externals/jquery-ui](https://github.com/OpenRefine/OpenRefine/tree/master/main/webapp/modules/core/externals/jquery-ui)
- [ ] remove the license file [main/webapp/licenses/jquery_ui.LICENSE.txt](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/licenses/jquery_ui.LICENSE.txt)
- [ ] remove all mentions of `external/jquery-ui/**/*` from [main/webapp/modules/core/MOD-INF/controller.js](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/MOD-INF/controller.js)

wetneb:
To me, the goal is precisely to remove `[main/webapp/modules/core/externals/jquery-ui](https://github.com/OpenRefine/OpenRefine/tree/master/main/webapp/modules/core/externals/jquery-ui)`, so I would like to do this in the same PR.

wetneb:
And now the goal would be to make sure that the application still works :) The end-to-end tests suggest it does not.

wetneb:
@mathdebate09 are you still working on this? Let me know if anything about the task is unclear.

To summarize, the goal would be to keep OpenRefine's functionality intact, while no longer storing a copy of jquery-ui in this repository. Instead, the jquery-ui library would be downloaded at build time from NPM, similarly to other libraries listed in [`main/webapp/package.json`](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/package.json).

mathdebate09:
Yes I'm trying to figure out the setup still 😅, it's been a while since I've done something with java, I'm sorry if this is troublesome

wetneb:
Great! For this particular task I don't think you'd need to do any Java work, apart from installing Java and Maven on your machine to be able to start OpenRefine.

wetneb:
@mathdebate09 are you still working on this? I would close the PR otherwise.

mathdebate09:
Hello actually my finals came up so it's a bit hectic
I'll get working on this as soon as I'm done with them, but if the issue has high priority, you can re-assign it to someone else

PR #6988, opened 2024-11-13
facets: Add column dependency extraction

wetneb:
This adds a method of `FacetConfig` to extract column dependencies from a facet, and to `EngineConfig` to get the set of all columns referenced by facets.

This will then be used to expose a similar method on operation classes.

For #133. Follow-up to #6740.


wetneb:
Merging this soon.

PR #6993, opened 2024-11-18
Update CI action names

thadguidry:
We have 2 Actions named the same thing: "CI"
Let's make it easier to distinguish between them.

### DIFF
--- a/.github/workflows/pull_request_e2e.yml
+++ b/.github/workflows/pull_request_e2e.yml
@@ -1,4 +1,4 @@
-name: CI E2E
+name: E2E CI
 
 on:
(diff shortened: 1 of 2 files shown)


wetneb:
I would rather say "server CI" than "CI SERVER" (otherwise it reads like "the server that provides CI to us", and maybe the all caps are not necessary). Similarly "E2E CI" instead of "CI E2E".

thadguidry:
What?!? But everyone loves "THAD CAPS". Thanks for reminding me of my bad habits. ;-)

wetneb (approved):

PR #6999, opened 2024-11-19
fix: Remove usage of org.json in ComputeClustersCommand

wetneb:
@tfmorris pointed out that we have this use of `org.json` creeping again in the codebase. We migrated out of org.json years ago to Jackson, so it makes sense to keep relying on that only.

The dependency to org.json was made possible because a new version of odftoolkit depends on it, and there isn't any isolation of dependencies. We originally migrated out of org.json for licensing reasons, but the author has in the meantime re-licensed it as public domain dedication, so a priori it's fine to have it as a transitive dependency again.


wetneb:
Merging in the coming days unless there are concerns.

PR #7003, opened 2024-11-22
Mark tables used for layout as presentational

Abbe98:
In the best of world we would move away from using tables for layout, in the meanwhile lets mark them as presentational.


wetneb (approved):
Sounds good!

PR #7004, opened 2024-11-25
Issue5048 Start a Wikimedia Commons upload project with just a folder…

### LINKED ISSUE #5048: Start a Wikimedia Commons upload project with just a folder of files
As an uploader of files to Wikimedia Commons, it would save me time if I could just provide a folder that contains all the files that I would like to work on that could be used to create a new project.

### Proposed solution
Add another option on the `Create project` menu (possibly as a second option under `This computer`) that asks the user to specify a directory on their computer. This should create a new project that has (at least) two columns, one with the filenames and one with the full file path.

### Additional context

This is dependent on the Wikimedia Commons upload feature that's currently available in the 3.7 snapshot.

sunilnatraj:
Fixes #5048 

- UI - Added option in create project flow to specify a directory
- Files from the specified directory are read and a CSV file is created with details -fileName, fileSize(KB), fileExtension, lastModifiedTime, creationTime, author, filePath


Abbe98 (changes requested):
Doing this server side looks rather dangerous and would break all compatibility with remote hosts(including Wikimedias PAWS). You can however, do the same client side using the [File System API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API).

In the broader sense I like that this is in core and not tailored for Wikimedia Commons only as I easily think of many use cases beyond that of #5048.

wetneb:
> Doing this server side looks rather dangerous and would break all compatibility with remote hosts(including Wikimedias PAWS). You can however, do the same client side using the [File System API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API).

I don't see it like that. OpenRefine is currently designed to run locally, so I don't think we can qualify this of of "breaking all compatibility", since it's not a use case we have ever claimed to support. In the case of PAWS and uploading files to Commons, doing the file listing in the browser would actually make this new feature useless for PAWS users, because the actual file uploading to Commons interprets the file paths in the backend-side. The current way to use PAWS for a Commons upload is indeed to upload the files to the PAWS workspace first, so that's where they need to be listed.

So, funnily enough, if we want to help PAWS users here, I think we need to keep listing files in the backend as this PR does. Unless we redesign the Commons support in the Wikibase extension to fetch files to upload via the browser, but that sounds like a rather complicated move and it's not clear to me the UX would be improved.

Abbe98:
>> Doing this server side looks rather dangerous and would break all compatibility with remote hosts(including Wikimedias PAWS). You can however, do the same client side using the [File System API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API).
>
>I don't see it like that. OpenRefine is currently designed to run locally, so I don't think we can qualify this of of "breaking all compatibility", since it's not a use case we have ever claimed to support. In the case of PAWS and uploading files to Commons, doing the file listing in the browser would actually make this new feature useless for PAWS users, because the actual file uploading to Commons interprets the file paths in the backend-side. The current way to use PAWS for a Commons upload is indeed to upload the files to the PAWS workspace first, so that's where they need to be listed.
>
>So, funnily enough, if we want to help PAWS users here, I think we need to keep listing files in the backend as this PR does. Unless we redesign the Commons support in the Wikibase extension to fetch files to upload via the browser, but that sounds like a rather complicated move and it's not clear to me the UX would be improved.

My point is not about the usefulness of the functionality, it's about exposing the remote file system.  If we want to do this server-side(as you say there is a use case) let's put in an extension(just like with Jython).

Abbe98:
>Note: I am also supportive of designing this feature so that it can be useful beyond the Commons import use case. Perhaps there are use cases for which listing the files from the browser would be better, but at the moment I can't think of any. 

Native file picking/accessibility and support for remote hosts? Support for containerized distributions(docker, snap, flatpak, etc)?

Abbe98:
> When I wrote "use cases", I was more thinking about user stories. In which context would you use this feature apart from Commons uploads? Meaning: what sort of data are you dealing with, what do you want to do with this data, what do you want to export it to, and so on. I think it's generally worth stepping back and looking at the whole user journey, as the case of PAWS shows.

Mostly preparation of other types of uploads, plenty of upload tools have two steps with files+tabular metadata. So I imagine that most would use it to bootstrap such metadata files. Kinda like a combination of `ls` + `printf` (but way easier for most users).

sunilnatraj:
@wetneb @Abbe98 Here is a suggestion to support file listing feature with ability to add support for more sources going forward. As in the database connection, provide a set of sources from which file list can be obtained.  

Screenshot 2024-11-28 at 5 16 03 PM

Abbe98:
> @wetneb @Abbe98 Here is a suggestion to support file listing feature with ability to add support for more sources going forward. As in the database connection, provide a set of sources from which file list can be obtained. Mockup - made using image editor
> Screenshot 2024-11-28 at 5 16 03 PM

I would suggest that you go with a regular `select` element(or something similar) inside the panel to pick a source. The existing pattern with nested sidebars is not the best as each time a sidebar is opened/closed the content will "jump" to to either side.

sunilnatraj:
@wetneb @Abbe98  Updated the file list generation flow; for showcase purpose i added options of Google Drive, Flickr. Let me know if the flow is as expected. This now gives the ability to add support to get file list from multiple sources.

https://github.com/user-attachments/assets/0eaa7260-1dc6-4daa-937c-7bd385541f7c

wetneb:
I would like propose to step back a bit. I am a bit worried that if we already get on to UI details and enlarging the scope of this effort to other file sources, it might result in a lot of work from @sunilnatraj which could then be blocked by higher-level considerations, don't you think? It seems to me that we currently don't have an agreement on rather important points:

* Where should this feature be implemented? It could be in this repository, in the Commons extension, or in another extension. It's a question on which it would be worth hearing opinions from more people, I think.
* Should file listing be done on front-end side or back-end side. It's something worth getting an agreement early on since the implementation will be very different depending on the answer.

Perhaps those things are worth discussing in #5048 ? I think this PR is a great way to show what it could look like and prompt us to identify design questions. Sorry about that @sunilnatraj, I probably should have flagged that earlier.

antoine2711:
@sunilnatraj : I have a suggestion for you: would you be willing to implement that in a totally new extention, let’s call it here the File Extension? I think what you did is great, and that the discussion in the community will be more of a break than helping you code. I’m not saying to not listen to the proposition of the community, but rather that it will give you more flexibility to develop a working product, while letting the community contribute at their own pace, while not stoping you.

I would be very happy to test your extension. And this extension could be useful not only for WikiCommons, but also as a standalone.

Features suggestions:
- [ ] Add an option to import the text of each file in a new column
- [ ] It would be great to have BOTH server side import (from a path of a folder on the server) and import thru multiple folder or file download (for PAWS).
- [ ] Have a column with a checksum, like MD5, SHA, SHA256m etc…
- [ ] Have a column with permissions (RWX etc) on the server side import

Thank you very much for this contribution, it’s awesome!

Regards, Antoine

sunilnatraj:
Closing this PR as the decision is to provide this support in an extension

Issue #7010, opened 2024-11-26
Browse Workspace Directory and Open Directory Extension does not work on ubuntu / chrome


magdmartin:
when testing extension management page #6975 I found that  the button `Open extensions directory` does not open Nautilus the file manager in Ubuntu.

logs
```
03:24:30.704 [                   refine] GET /command/core/get-csrf-token (4407ms)
03:24:30.708 [                   refine] POST /command/core/open-extensions-dir (4ms)
```

 Following @Abbe98 comment below, I also confirm that the `Browse Workspace Directory` does nothing 

### Screenshots
![image](https://github.com/user-attachments/assets/28a05024-1f5a-4aab-9c9e-02ef3a582046)
### Versions
 - Operating System: Ubuntu 22.04.5
 - Browser Version: chrome Version 131.0.6778.85 (Official Build) (64-bit)
 - JRE or JDK Version: openjdk version "19.0.2" 2023-01-17
 - OpenRefine: built from master 30d8223ab59259

Abbe98:
The links open in a new tab as of #7012. Regarding the extensions directory, do the regular "open workspace directory" button work for you? They should work exactly the same.

Abbe98:
It works on my Fedora/Gnome system(although I use Nemo and not Nautilus), maybe @Ainali can reproduce this?

Ainali:
It works on vanilla Ubuntu 24.04.1, opening Nautilus for me.

SoryRawyer:
Hey folks, I was able to reproduce this and I think I tracked down why this is happening. I believe this has to do with headless vs non-headless installations of Java. As I understand it, a "headless" version of Java comes without graphics support. This version of Java would fail to open Nautilus (and, as it turns out, doesn't open the browser either) because [this check](https://github.com/OpenRefine/OpenRefine/blob/master/modules/core/src/main/java/com/google/refine/util/OpenDirectoryUtilities.java#L11) returns `false`.

I'm running Ubuntu 24.04.02 and my `apt`-installed Java is headless, but a Java installation from Adoptium (the location recommended in the [OpenRefine install docs](https://openrefine.org/docs/manual/installing#java)) is not. I was unknowingly using the system-installed Java (I added the Adoptium JDK location to the _end_ of my `PATH` variable, not before 😅 ). Using the Adoptium JDK resolved this issue for me.

Here is the code I ran to determine if Java was headless or not (I couldn't find a way to do this purely through the command line):
```java
[...]

Issue #7009, opened 2024-11-26
OpenRefine application fails to start - https://github.com/OpenRefine/OpenRefine/pull/6975


sunilnatraj:
Changes introduced in https://github.com/OpenRefine/OpenRefine/pull/6975
@Abbe98 Application start fails as extensionsDir is null. 
xtensionsDir is being initialized for Unix OS only. On other OS it is set as null and that is resulting in failure.

Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.io.File.exists()" because "extensionsDir" is null
at com.google.refine.RefineServer.getDataDir(Refine.java:446)
at com.google.refine.RefineServer.configure(Refine.java:337)

wetneb:
The Windows workflow only runs the unit tests on Windows, not the end to end tests.

wetneb:
@Abbe98 I've temporarily reverted the change so as to ease the pressure on your side to fix it in a timely manner, but looking forward to including it again once the problem is fixed. Happy to have a look at it myself too.

wetneb:
Closed by #7011.

PR #7012, opened 2024-11-27
Add an extension mangement page

Abbe98:
Follow up to #6975, #7009, and #7010.


main/src/com/google/refine/commands/OpenExtensionsDirCommand.java
```diff
@@ -0,0 +1,49 @@
...
+
+            if (Desktop.isDesktopSupported()) {
+                Desktop desktop = Desktop.getDesktop();
+                desktop.open(dir);
+            } else /* if Mac */ {
+                Runtime.getRuntime().exec(
+                        "open .",
+                        new String[] {},
+                        dir);
+            }
```
wetneb, line 41: Following up on @magdmartin's comment in #7010, perhaps it would be worth having this logic in a separate function, which would be used both by the "open workspace directory" and the "open extensions directory" features. This way we could refine the support for each platform in a single place. (For instance a utility class with a static method on it.) I'm not suggesting that refinement should be done in this PR, it can be done separately.

Abbe98:
> Following up on @magdmartin's comment in https://github.com/OpenRefine/OpenRefine/issues/7010, perhaps it would be worth having this logic in a separate function, which would be used both by the "open workspace directory" and the "open extensions directory" features. This way we could refine the support for each platform in a single place. (For instance a utility class with a static method on it.) I'm not suggesting that refinement should be done in this PR, it can be done separately.

I see a lot of potential follow ups, beyond small things like what you propose I see two medium ones as rather interesting:

 - Exposing more module metadata through Butterfly and render it in the table it (I think @tfmorris and I had a brief exchange about this somewhere in the past)
 - Making the two open directories feature stoggable in configuration(rather than exposing them and just throw errors).

wetneb:
Great, looking forward to those! Don't you think it would make sense to introduce this common function in this PR already?

Abbe98:
> Great, looking forward to those! Don't you think it would make sense to introduce this common function in this PR already?

Personally I would much rather get this on to master ASAP as it exposes it to more users/feedback.

wetneb (changes requested):
I would prefer to do this simplification before merging, to avoid knowingly introducing code duplication. I am happy to make the corresponding changes in this PR if you would like me to.

Abbe98:
@wetneb `OpenDirectoryUtilities.OpenDirectory()` might not be the best way to name things, feel free to change it during review.

I decided not to touch the `if (!"127.0.0.1".equals(serverName) && !"localhost".equals(serverName))` logic as it seems to be broken, given it doesn't actually say anything about from where the content is served, looking at the `Host` header might be better but it's likely not preserved in most remote setups. It should be solved wither by a startup flag or something like #7014, I do however, consider both approaches out of scope for this PR.

main/src/com/google/refine/commands/OpenExtensionsDirCommand.java
```diff
@@ -0,0 +1,41 @@
...
+
+        String serverName = request.getServerName();
+
+        if (!"127.0.0.1".equals(serverName) && !"localhost".equals(serverName)) {
+            respond(response,
+                    "{ \"code\" : \"error\", \"message\" : \"Extension directory can only be opened on the local machine where OpenRefine is run.\" }");
+        } else if (ProjectManager.singleton instanceof FileProjectManager) {
+            File dir = new File(((FileProjectManager) ProjectManager.singleton).getWorkspaceDir(), "extensions");
+
+            OpenDirectoryUtilities.OpenDirectory(dir);
```
wetneb, line 33: Java methods generally start with a lower case letter:
```suggestion
            OpenDirectoryUtilities.openDirectory(dir);
```

main/src/com/google/refine/commands/OpenWorkspaceDirCommand.java
```diff
@@ -62,15 +62,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
-            if (Desktop.isDesktopSupported()) {
-                Desktop desktop = Desktop.getDesktop();
-                desktop.open(dir);
-            } else /* if Mac */ {
-                Runtime.getRuntime().exec(
-                        "open .",
-                        new String[] {},
-                        dir);
-            }
+            OpenDirectoryUtilities.OpenDirectory(dir);
```
wetneb, line 65:
```suggestion
            OpenDirectoryUtilities.openDirectory(dir);
```

modules/core/src/main/java/com/google/refine/util/OpenDirectoryUtilities.java
```diff
@@ -0,0 +1,21 @@
+
+package com.google.refine.util;
+
+import java.awt.Desktop;
+import java.io.File;
+import java.io.IOException;
+
+public class OpenDirectoryUtilities {
+
+    public static void OpenDirectory(File dir) throws IOException {
```
wetneb, line 10:
```suggestion
    public static void openDirectory(File dir) throws IOException {
```

wetneb (commented):
@sunilnatraj would you be able to confirm that this works as expected under Windows?

sunilnatraj:
> @sunilnatraj would you be able to confirm that this works as expected under Windows?

@wetneb I am able to launch the application without any issues. I am using a Mac system. I do not have a Windows system to check this out.

wetneb:
My bad, I had misunderstood #7009.

wetneb (approved):

Issue #6646, opened 2024-05-28
Allow user to automatically report their OpenRefine installation configuration


magdmartin:
Following discussions on the forum when [drafting the 2024 user survey](https://forum.openrefine.org/t/drafting-the-2024-user-survey/1431), this issue proposes adding an option for a user to visualize and share OpenRefine's configuration settings. 

The user can copy-paste this information and provide it when creating a bug or requesting support on the forum. 

### Proposed solution
Add an option on the right menu to `view OpenRefine configuration`. This will open a new tab with the following information (the list is non-definitive; feel free to add or remove)
* version of OpenRefine
* OS and its version
* current browser and its version
* list of extensions installed and their version
* list of reconciliation services installed 
* variables value for -p -i -H -w -d - -v -x -c from `refine` file or passed when starting OpenRefine
* variable values from `refine.ini` including 
   *  REFINE_MAX_FORM_CONTENT_SIZE
   * REFINE_MEMORY
   * REFINE_MIN_MEMORY
   * JAVA_HOME
   * JAVA_OPTIONS
   * REFINE_AUTOSAVE_PERIOD
* language set 
*  from the preferences 
    * ui.browsing.listFacet.limit
    * notification.status		
* Number of projects (I don't know if at this stage we want to collect project-specific statistics)

We can add a button to send the information to the development team so we can compile statistics.

tfmorris:
I'm not sure what value the number of projects adds here. Does it help with debugging?

@Abbe98 Can you expand on your thumbs down?

magdmartin:
I've realized that thanks to #6421, the About page now includes some of that information.
* version of OpenRefine
* list of extensions installed
* Java version

I agree that from a troubleshooting perspective, the number of projects does not provide meaningful information.

Aric262:
I would like to work on this issue. Please assign it to me

PR #7019, opened 2024-11-28
Make it possibly for users to quickly copy system and version information

Abbe98:
This change makes it possibly for users to copy detailed version and system information by clicking the version on the index page.

Example clipboard contents:

```
3.9-SNAPSHOT [TRUNK]
OpenJDK Runtime Environment 21.0.4+7
OpenJDK 64-Bit Server VM 21.0.4+7
Modules: command-palette, core, database, jython, module, pc-axis, wikidata
Client user-agent: Mozilla/5.0 (X11; Linux x86_64; rv:131.0) Gecko/20100101 Firefox/131.0
```

### DIFF
--- a/main/webapp/modules/core/index.vt
+++ b/main/webapp/modules/core/index.vt
@@ -56,7 +56,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     <div id="project-links">
       <div id="logo-container">
         <img alt="OpenRefine" src="images/logo-gem-126.svg" height="65" />
-        <div id="openrefine-version"></div>
+        <div id="openrefine-version" role="button" tabindex="0" title="">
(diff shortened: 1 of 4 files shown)


wetneb:
Nice! But it's not very discoverable, is it? To me the version number looks the same as before:
image
Or is it a pattern that is established in other tools, such that you'd expect users to know they can click on the version number and get those details copied in the clipboard?

In general, I would open the About page of a tool to find this information. We do have a bit of it already:
image

Why not add the missing bits there?

Abbe98:
I would imagine that this would be documented in our issue template once this has shipped in a release.

The about page is broken in various ways and do not follow the pattern of other pages, it should be fixed but it's not on my radar atm.

magdmartin:
I support adding the process to share their system configuration in the issue template and maybe in the forum too. 
This relates to https://github.com/OpenRefine/OpenRefine/issues/6646

wetneb:
I agree it's a useful feature but I would prefer to make it more discoverable from the tool itself - at the moment it feels a bit like an easter egg. Even for a user who reads about it from the issue template, it might be confusing that the version number doesn't indicate any clickability.
My preference would be to follow the UX pattern established in other tools, of retrieving the full version information from the about page.

magdmartin:
I think we should merge it as is and address the discoverability issue in a later phase.

tfmorris:
How about if we add a copy icon affordance next to version string to help guide the user? 
(Similar to what Github uses next to the code example above in the PR description) 
copy-svgrepo-com

Abbe98:
Updated with @tfmorris suggestion:

Screenshot 2025-05-06 at 07-08-11 OpenRefine

SoryRawyer (approved):
I think this looks great, thanks for adding it! Since there's been some discussion around this, I'll wait until Monday to merge in case anyone has any final comments.

PR #7024, opened 2024-11-30
operations: Implement validate() method for a first batch of classes

wetneb:
This adds validate() methods for operations in the `com.google.refine.operations.cell` package.
It also introduces the required methods on `EngineConfig` and `FacetConfig` to do the validation there too.


wetneb:
Merging in the coming days unless there are any concerns.

Abbe98 (approved):
Looks good to me, and it does not touch existing interfaces so I think you should merge this if it unblocks you @wetneb!

wetneb:
Thanks a lot! It's never "blocking" in the sense that I can always pile up more commits on my branch but it's indeed more reassuring to get it merged.

PR #7031, opened 2024-12-03
wikibase: Only ignore file upload warnings for matched cells

wetneb:
This can be useful in scenarios where a file was uploaded just before us, meaning that the file name clash was not caught by FileNameScrutinizer.

Follow-up to #6967.


sebastian-berlin-wmse (approved):
I wasn't able to try actually making an upload (due to unrelated issues locally), but the changes look good to me.

PR #7033, opened 2024-12-04
chore: Deprecate EngineConfig::reconstruct

wetneb:
This method returns null when supplied with an invalid argument, which is bad practice. Instead of swallowing the exception and printing its stack trace, it should rather let the exception flow to the caller. Because fixing this behaviour would be a breaking change, I have deprecated the method and introduced a new one with the better behaviour. I have also introduced EngineConfig::defaultRowBased, as the reconstruct was overwhelmingly being used to build a default row-based engine configuration.


Abbe98 (approved):
Looks good and the naming is IMO much better.

PR #7040, opened 2024-12-11
Issue 3218 - Add "remove duplicate rows" menu item

### LINKED ISSUE #3218: Add "remove duplicate rows" menu item
After fixing inconsistencies with the clustering and merging functionality, I assume it is a common to want to remove potentially duplicate rows.

### Proposed solution

It would be great to add an item to the `All -> Edit rows` menu to be able to quickly remove duplicated rows, named for example "Remove duplicate rows".

This functionality could be either:

* Simple version: check the data in _all_ the columns and only remove rows that have exactly the same values in all columns.
* More complex, with more options: add an intermediate dialogue that allows the user to choose which columns are checked for exact duplication (with all columns checked by default).

### Alternatives considered

Currently, a few tutorials and videos [are](https://programminghistorian.org/en/lessons/cleaning-data-with-openrefine) [floating](https://kb.refinepro.com/2011/08/remove-duplicate.html) [around](https://guides.library.illinois.edu/openrefine/duplicates) with instructions like the following:

1. Sort by columns that will guarantee that duplicates will be grouped together
1. Make the sort permanent with `Sort -> Reorder rows permanently`
1. Blank columns down with `All -> Edit columns -> Blank down`
1. Facet by blank on a column that is expected to have data for all unique items
1. Select the blank rows (i.e. the `true` value)
1. Remove all matches with `All -> Edit rows -> Remove matching rows`

This is hardly intuitive nor straight-forward.

sunilnatraj:
Fixes #3218

- Added new row level command RemoveDuplicateRows
- Updated menu to add new sub-option under rows menu
- Duplicate row identification uses the columns as specified by the user to identify duplicate rows in the data
- First row of duplicate rows is preserved


main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
```diff
@@ -0,0 +1,108 @@
...
+                "Duplicate rows removal completed. " + rowIndices.size() + " rows removed.",
+                this,
+                new RowRemovalChange(rowIndices));
+    }
+
+    private void findDuplicateRows(Project project, List<Column> criteriaColumns, List<Integer> rowIndices) {
+        Set<String> uniqueKeys = new HashSet<>();
+
+        int c = project.recordModel.getRecordCount();
+        for (int r = 0; r < c; r++) {
```
wetneb, line 76: Instead of looping on all rows, did you consider using the `RowVisitor` interface with `engine.getAllFilteredRows()`? This would improve this operation to respect facets, which could likely be useful (to only remove duplicate rows in a sub-part of the project). See `RowRemovalOperation` for an example of how it's done.

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
```diff
@@ -0,0 +1,108 @@
...
+                criteriaColumns.add(toColumn);
+            }
+        }
+
+        findDuplicateRows(project, criteriaColumns, rowIndices);
+
+        return new HistoryEntry(
+                historyEntryID,
+                project,
+                "Duplicate rows removal completed. " + rowIndices.size() + " rows removed.",
```
wetneb, line 67: The `completed.` part strikes me as inconsistent with the style of the other history entry descriptions. Perhaps `Remove 123 duplicate rows.` would be simpler and more fitting (see for instance `RowRemovalOperation`).

wetneb (commented):
Very nice, impressive work!

The UI currently looks like this:
image

I would suggest the following changes:
* instead of having "Select All" as a checkbox, I would rather provide two buttons, "Select all" and "Deselect all" (note the capitalization: without capitalizing "all") similarly to what is done in the "Extract" dialog of the undo/redo history panel. It would make the UX more consistent and in my opinion would also be clearer.
* I think the verb "analyze" isn't so fitting for describing which columns are used to group rows together. You could instead write "Group rows by columns" (which would perhaps remind SQL-savvy users of the `GROUP BY` syntax). Maybe others have better suggestions.
* There seems to be currently two scrollable areas inside each other (which I failed to show on the screenshot because my screenshot tool doesn't let me hover the area while making the screenshot), which roughly correspond to both divs with outlines that we can see in the screenshot. I'd suggest only keeping the inner one.
* Maybe the title "Remove duplicates" could be "Remove duplicate rows" to make it clear that the operation isn't relying on record structures (so far?)

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private void findDuplicateRows(Project project, List<Column> criteriaColumns, List<Integer> rowIndices) {
+        Set<String> uniqueKeys = new HashSet<>();
+
+        int c = project.recordModel.getRecordCount();
+        for (int r = 0; r < c; r++) {
+            Row row = project.rows.get(r);
+            String key = criteriaColumns.stream()
+                    .map(col -> normalizeValue(row.getCell(col.getCellIndex())))
+                    .collect(Collectors.joining("|"));
```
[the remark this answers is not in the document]
sunilnatraj: @wetneb Would it make sense to use a UUID as the separator. With List will there be performance overhead ?

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private static String normalizeValue(Object value) {
+        SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
+        if (value == null) {
+            return "";
+        } else if (value instanceof Date) {
+            return dateFormat.format((Date) value);
+        } else if (value instanceof String && ((String) value).trim().startsWith("{")) {
+            return normalizeJson((String) value);
```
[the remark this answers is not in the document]
sunilnatraj: @wetneb The reason to support JSON came into consideration based on the recent work on commons media upload where Wikitext is in JSON format.

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
sunilnatraj: Updated

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
sunilnatraj: @wetneb I did check this option first, my understanding is that it is tightly coupled to facet definition and the visitor evaluates each row against the facet criteria. As duplicate row removal cannot use the facet criteria model i used the raw iteration. Let me know if i missed something here.

sunilnatraj:
@wetneb Addressed UI feedback. Also put in an updated label - please review.
Screenshot 2024-12-12 at 10 56 34 AM
Screenshot 2024-12-12 at 11 15 18 AM

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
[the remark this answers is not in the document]
wetneb: Adding a long separator has a performance overhead too, as you'll be allocating a lot of long strings, whereas the `List` would just wrap around existing ones. You could try to measure both approaches if you want, but given that `List` is more readable and simpler, I would go for that.

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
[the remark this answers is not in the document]
wetneb: On top of my head I can't think of cases where JSON is used inside wikitext for commons files. Do you have examples?

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
sunilnatraj: Updated to use RowVisitor

sunilnatraj:
If facet filter has been applied and user requests for duplicate rows removal, only the rows that satisfy the facet criteria are considered for duplicate row removal.

main/webapp/modules/core/styles/dialogs/remove-duplicate-rows-dialog.css
```diff
@@ -0,0 +1,59 @@
...
+.remove-duplicate-rows-dialog-instruction {
+  padding: var(--padding-loose);
+  font-weight: bold;
+  font-size: 1.1em;
+  color: var(--text-primary);
+  margin-top: 3px;
+  margin-bottom: 3px;
+}
+
+.column-reordering-dialog-column {
```
wetneb, line 43: The class name suggests this corresponds to a different dialog, so it could probably be removed from this file, no?

main/webapp/modules/core/scripts/dialogs/remove-duplicate-rows-dialog.js
```diff
@@ -0,0 +1,104 @@
...
+    this._elmts.selectAllButton.html($.i18n('core-buttons/select-all'));
+    this._elmts.deselectAllButton.html($.i18n('core-buttons/deselect-all'));
+    
+    this._level = DialogSystem.showDialog(dialog);
+    var columnList = this._elmts.columnContainer;
+
+    for (var i = 0; i < theProject.columnModel.columns.length; i++) {
+        var column = theProject.columnModel.columns[i];
+        var name = column.name;
+        var columnCheckbox = $('<div><input type="checkbox" class="column-checkbox" id="column-' + i + '" value="' + name + '" checked=true /> <label for="column-' + i + '">' + name + '</label></div>');
```
wetneb, line 61: Instead of concatenating the name into the HTML, it would be safer to add it with `.attr()` and `.text()` to make sure it is properly escaped (in case the column names contain some special characters).

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -371,6 +371,9 @@
     "core-dialogs/name": "Name",
     "core-dialogs/action": "Action",
     "core-dialogs/should-return-number": "your expression should return a number",
+    "core-dialogs/remove-duplicate-rows": "Remove duplicate rows",
+    "core-dialogs/remove-duplicate-rows-col-select": "Select columns to identify duplicates",
+    "core-buttons/no-criteria-selected": "Row duplicate identification criteria not set. No column(s) selected.",
```
wetneb, line 376: Maybe we can simplify this message a bit, I don't think it needs to be two sentences. How about:
```suggestion
    "core-buttons/no-criteria-selected": "At least one column needs to be selected to identify duplicates.",
```

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
```diff
@@ -0,0 +1,130 @@
...
+import com.google.refine.model.Row;
+import com.google.refine.model.changes.RowRemovalChange;
+import com.google.refine.operations.EngineDependentOperation;
+import com.google.refine.operations.OperationDescription;
+
+public class RowDuplicatesRemovalOperation extends EngineDependentOperation {
+
+    final protected List<String> _criteria;
+    final List<Column> criteriaColumns = new ArrayList<Column>();
+    HashSet<Object> rowUniqueKeys = new HashSet<>();
```
wetneb, line 35: The fields in an `Operation` class should only contain the operation metadata, not the data obtained from its execution. I would move this field into the `RowVisitor` since it's only needed there.

main/webapp/modules/core/langs/translation-en.json
```diff
@@ -371,6 +371,9 @@
     "core-dialogs/name": "Name",
     "core-dialogs/action": "Action",
     "core-dialogs/should-return-number": "your expression should return a number",
+    "core-dialogs/remove-duplicate-rows": "Remove duplicate rows",
+    "core-dialogs/remove-duplicate-rows-col-select": "Select columns to identify duplicates",
```
wetneb, line 375: Is "identify duplicates" clear enough? I think it could be clearer with "group", "index" or "key", but I might be biased. Thoughts, @ostephens and @thadguidry?

wetneb (commented):
Thanks a lot, it's looking close :)

main/webapp/modules/core/styles/dialogs/remove-duplicate-rows-dialog.css
sunilnatraj: Removed.

main/webapp/modules/core/scripts/dialogs/remove-duplicate-rows-dialog.js
sunilnatraj: Addressed

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
sunilnatraj: Moved rowUniqueKeys to RowVisitor

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
```diff
@@ -0,0 +1,130 @@
...
+            public boolean visit(Project project, int rowIndex, Row row) {
+                if (isDuplicate(row, rowUniqueKeys)) {
+                    rowIndices.add(rowIndex);
+                }
+                return false;
+            }
+        }.init(rowIndices);
+    }
+
+    private boolean isDuplicate(Row row, HashSet<Object> rowUniqueKeys) {
```
wetneb, line 108: You could even move that method inside the visitor, which would mean you wouldn't even need to pass the `rowUniqueKeys` as parameter as it would already be accessible as an attribute.

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
```diff
@@ -0,0 +1,130 @@
+/*
+
+Copyright 2024 OpenRefine
+All rights reserved.
```
wetneb, line 4: This copyright header seems different from the other ones we use, with the usual BSD Clause 3 license. Was it on purpose?

main/src/com/google/refine/operations/row/RowDuplicatesRemovalOperation.java
sunilnatraj: Addressed

main/webapp/modules/core/scripts/dialogs/remove-duplicate-rows-dialog.html
```diff
@@ -0,0 +1,25 @@
...
+                <div bind="or_dialog_columnsList"></div>
+            </div>
+            <table role="presentation">
+                <tr>
+                    <td>
+                        <div class="remove-duplicate-rows-dialog-column-container" bind="columnContainer"></div>
+                    </td>
+                </tr>
+            </table>
+            <td>
```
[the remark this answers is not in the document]
sunilnatraj: Addressed

main/tests/server/src/com/google/refine/commands/row/RemoveDuplicateRowsCommandTest.java
```diff
@@ -0,0 +1,222 @@
...
+                                        "    \"\"LATITUDE\"\": 41.8402,\n" +
+                                        "    \"\"LONGITUDE\"\": -72.01,\n" +
+                                        "    \"\"ELEVATION\"\": 209,\n" +
+                                        "    \"\"MAPID\"\": \"\"Hampton, CT\"\",\n" +
+                                        "    \"\"LAND_USE\"\": \"\"Urban/Agric\"\",\n" +
+                                        "    \"\"TERRAIN\"\": \"\"Rolling\"\",\n" +
+                                        "    \"\"MLM\"\": \"\"Mixed\"\",\n" +
+                                        "    \"\"NADP_ID\"\": \"\"CT15\"\",\n" +
+                                        "    \"\"NADP_DISTANCE\"\": \"\".0236\"\",\n" +
+                                        "    \"\"UPDATE_DATE\"\": \"\"2004-11-03 08:41:33\"\"}",
```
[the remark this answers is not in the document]
sunilnatraj: Fixed

wetneb (approved):
I ran out of things to complain about!

I'd wait a bit more to give more people a chance to chime in as it's a rather important feature, that has been asked for a long time. For instance about the wording question above, but not only.

main/webapp/modules/core/langs/translation-en.json
sunilnatraj: @ostephens @thadguidry Can you share your inputs on this

thadguidry: "Pick columns used to identify duplicate rows" is clear but:
1. it states what picking columns is used for.
2. it clarifies that duplicate rows will be identified and located.
3. but does not say anything about removal built-in, but the modal title does.

I'm proposing:
"Pick columns used to identify and remove duplicate rows" would handle all 3.

I wouldn't worry about longer phrases that bring clarity since text wrapping or smaller font sizes help when horizontally constrained, which we are not in this modal.

thadguidry:
Built and tested.  Seems to work as I expected, even within records mode.  Thanks for this @sunilnatraj !

main/webapp/modules/core/langs/translation-en.json
thadguidry:
```suggestion
    "core-dialogs/remove-duplicate-rows-col-select": "Select columns used to identify duplicate rows",
```

thadguidry (changes requested):

sunilnatraj:
Thank you @thadguidry

sunilnatraj:
@thadguidry Can you review the PR; the label update has been committed.

thadguidry:
works and looks good now!
Thanks @sunilnatraj !

PR #7043, opened 2024-12-12
operations: Implement validate() method for c.g.r.o.column classes

wetneb:
Following up on #7024, continuing the implementation for other classes in the `com.google.refine.operations.column` package.


wetneb:
Merging this soon.

PR #7053, opened 2024-12-30
wikibase: Improve recon name for created entities

### LINKED ISSUE #6996: Commons upload: newly uploaded cells should contain `File:` prefix in cell.recon.name
According to @sunilnatraj's analysis in https://github.com/OpenRefine/CommonsExtension/pull/153, newly-uploaded files get represented by matched reconciled cells whose `.name` field does not contain a `File:` prefix, contrary to matches returned by the reconciliation service (for existing files). It would be worth unifying those two.
If that's indeed the case, the simplest would be (I think) to change our upload code so that matched cells do contain the `File:` prefix. It could be that this breaks some workflows, if folks explicitly rely on those filenames not having the `File:` prefix, but I would expect downstream uses of uploaded files generally go through the Mids instead, so I think it would be fine to introduce such a change.

sunilnatraj:
Fixes #6996 
- Extended NewEntityLibrary to also hold label for new entities 
- Updated EditBatchProcessor performEdit to set label from the value set in NewEntityLibrary
- Updated test cases


extensions/wikibase/tests/src/org/openrefine/wikibase/operations/PerformWikibaseEditsOperationTest.java
```diff
@@ -77,7 +77,7 @@ public void testGetTagCandidates() {
     @Test
     public void testLoadChange()
             throws Exception {
-        String changeString = "newItems={\"qidMap\":{\"1234\":\"Q789\"}}\n" + "/ec/\n";
+        String changeString = "newItems={\"qidMap\":{\"1234\":{\"returnedId\":\"Q789\",\"label\":\"Q789\"}}}\n" + "/ec/\n";
```
wetneb, line 80: We'll need to make sure the old format of such Change objects can still be deserialized correctly, so that operations run with previous versions of OpenRefine can still be done and undone correctly. So there should be two tests: the current one (unchanged) and another one for the new format.

To make this backwards compatibility simpler, I would recommend using two separate maps (one for the entity ids, one for the names) instead of storing pairs of values in maps. It's probably worth updating the Java types accordingly.

extensions/wikibase/src/org/openrefine/wikibase/editing/EditBatchProcessor.java
```diff
@@ -183,14 +185,22 @@ public EditResult performEdit()
                 // New entities
                 ReconEntityIdValue newCell = (ReconEntityIdValue) update.getEntityId();
                 EntityIdValue createdDocId;
+                String label = "";
```
wetneb, line 188: Maybe this variable could be called `name` instead, because it's going to be stored as the reconciliation candidate's name. It's not always a label (since for MediaInfo entities it is a filename).

extensions/wikibase/src/org/openrefine/wikibase/editing/EditBatchProcessor.java
```diff
@@ -183,14 +185,22 @@ public EditResult performEdit()
...
                     createdDocId = ((MediaInfoEdit) update).uploadNewFile(editor, mediaFileUtils, summary, tags, filePageWaitTime,
                             filePageMaxWaitTime);
+                    label = "File:".concat(((MediaInfoEdit) update).getFileName());
                 } else {
-                    createdDocId = editor.createEntityDocument(update.toNewEntity(), summary, tags).getEntityId();
+                    LabeledDocument labeledDocument = (LabeledDocument) editor.createEntityDocument(update.toNewEntity(), summary, tags);
+                    createdDocId = labeledDocument.getEntityId();
+                    if (labeledDocument != null && labeledDocument.getLabels() != null) {
+                        label = labeledDocument.getLabels().get(Locale.getDefault().getLanguage()) != null
+                                ? labeledDocument.getLabels().get(Locale.getDefault().getLanguage()).getText()
```
wetneb, line 199: I think it would be worth having a fall back on taking *any* label if we can't find one for the current locale. And then fall back on the entity id if there isn't any label, because using the empty string as a name for a reconciliation candidate is likely to be not so pleasant for the user.

wetneb (commented):
Thank you, that's much better!

sunilnatraj:
@wetneb Need some guidance - I keep getting this error "badtoken] Invalid CSRF token." when connecting to Wikimedia commons api.

extensions/wikibase/tests/src/org/openrefine/wikibase/operations/PerformWikibaseEditsOperationTest.java
wetneb: That would work, but I think that to ensure backwards compatibility, it would be simpler to use **two separate maps**, as I wrote above. This means:
```java
Map<Long, String> reconToEntityId;
Map<Long, String> reconToName;
```

sunilnatraj: Perfect

wetneb:
@sunilnatraj I've renamed this PR from "Issue 6996" to "wikibase: Improve recon name for created entities" so that the title of the pull request is more informative. When looking at the list of open pull requests, it helps people understand what the PRs are about, without having to learn issue numbers by heart.

extensions/wikibase/src/org/openrefine/wikibase/editing/EditBatchProcessor.java
sunilnatraj: Added support for defaulting

extensions/wikibase/tests/src/org/openrefine/wikibase/operations/PerformWikibaseEditsOperationTest.java
```diff
@@ -77,7 +77,7 @@ public void testGetTagCandidates() {
     @Test
     public void testLoadChange()
             throws Exception {
-        String changeString = "newItems={\"qidMap\":{\"1234\":\"Q789\"}}\n" + "/ec/\n";
+        String changeString = "newItems={\"qidMap\":{\"1234\":\"Q789\"},\"qnameMap\":{}}\n" + "/ec/\n";
```
wetneb, line 80: Let's keep this test exactly like it was (no name map) to demonstrate that the previous change format is still supported.

extensions/wikibase/src/org/openrefine/wikibase/editing/NewEntityLibrary.java
```diff
@@ -155,7 +168,12 @@ public void updateReconciledCells(Project project, boolean reset) {
      */
     @JsonProperty("qidMap")
     public Map<Long, String> getIdMap() {
-        return map;
+        return reconToEntityId;
+    }
+
+    @JsonProperty("qnameMap")
```
wetneb, line 174: I would call it `nameMap` instead of `qnameMap`. The term "qid" refers to item identifiers in Wikidata slang.

extensions/wikibase/tests/src/org/openrefine/wikibase/operations/PerformWikibaseEditsOperationTest.java
```diff
@@ -95,4 +95,24 @@ public void testLoadChange()
...
+        String changeString = "newItems={\"qidMap\":{\"1234\":\"Q789\"},\"qnameMap\":{\"1234\":\"Q789\"}}\n" + "/ec/\n";
+        LineNumberReader reader = makeReader(changeString);
+        Change change = PerformWikibaseEditsOperation.PerformWikibaseEditsChange.load(reader, pool);
+
+        project.rows.get(0).cells.set(0, TestingData.makeNewItemCell(1234L, "my new item"));
+
+        change.apply(project);
+
+        assertEquals(Recon.Judgment.Matched, project.rows.get(0).cells.get(0).recon.judgment);
+        assertEquals("Q789", project.rows.get(0).cells.get(0).recon.match.id);
```
wetneb, line 110: It would be good to also add an assertion about the `recon.match.name`, since changing this field is the goal of this PR.

sunilnatraj: Added

extensions/wikibase/src/org/openrefine/wikibase/editing/NewEntityLibrary.java
sunilnatraj: Updated

extensions/wikibase/tests/src/org/openrefine/wikibase/operations/PerformWikibaseEditsOperationTest.java
sunilnatraj: Fixed

sunilnatraj:
@wetneb Only the Linux test for Database extension is failing. It appears to be an environment / db issue. 

Error:  com.google.refine.extension.database.DatabaseImportControllerTest.testDoPostParsePreview -- Time elapsed: 0.290 s <<< FAILURE!
Error: ng.AssertionError: expected [ok] but found [error]

**>> json::{"status":"error","message":"java.sql.SQLException: stmt pointer is closed\n"}**

wetneb:
@sunilnatraj yes, don't worry about it, I'll remove this test as it's flaky.

wetneb:
@sunilnatraj you should be able to trigger a new build now (for instance by reverting your last commit, to make sure this PR doesn't touch the database extension). Sorry for the trouble!

sunilnatraj:
@wetneb I have fixed the lint issue and committed the change. Thank you

extensions/wikibase/tests/src/org/openrefine/wikibase/operations/PerformWikibaseEditsOperationTest.java
```diff
@@ -92,7 +93,28 @@ public void testLoadChange()
 
         assertEquals(Recon.Judgment.New, project.rows.get(0).cells.get(0).recon.judgment);
 
-        assertEquals(changeString, saveChange(change));
+        assertEquals(validateChangeString, saveChange(change));
```
wetneb, line 96: We should keep the test as it was before, to demonstrate that the previous format is still supported.

sunilnatraj: @wetneb The input format is the previous one; the output format is the new one. Is the expectation that the output format is also as per the previous version?

sunilnatraj: @wetneb I have added support to ensure older version is not changed.

wetneb (approved):

wetneb:
Thanks a lot for tackling this @sunilnatraj!

sunilnatraj:
Thank you @wetneb

PR #7055, opened 2025-01-01
operations: Let core and extensions register icons

wetneb:
This PR makes it possible to associate icons to operations.

Those icons are meant to be added to menu items, dialog headers, history entries and more. They are meant to help users recognize operations visually, instead of relying on text. See [the forum topic about their design](https://forum.openrefine.org/t/operation-logos/1307).

Properly registering the logos and including them in the UI will follow.

There are surely many other ways to associate operations to icons (for instance, supplying the icon at the same time as providing the Java class for it), I am not particularly attached to this approach.

### DIFF
--- a/main/webapp/modules/core/scripts/project/operation-icons.js
+++ b/main/webapp/modules/core/scripts/project/operation-icons.js
@@ -4,6 +4,13 @@ var OperationIconRegistry = {
 };
 
 OperationIconRegistry.setIcon = function(operationId, iconPath) {
+  if (!iconPath.endsWith(".svg")) {
+    throw new Error('Operation icon must be a path to a SVG file.');
+  }
+  // this will cause the image to load so that once it is added to the DOM it will already be in the browser cache
+  var image = new Image();
+  image.src = iconPath;
+
   this.map.set(operationId, iconPath);
 }
 
--- a/main/webapp/modules/core/scripts/project/operation-icons.js
+++ b/main/webapp/modules/core/scripts/project/operation-icons.js
@@ -1,0 +1,13 @@
+
+var OperationIconRegistry = {
+  map: new Map()
+};
+
+OperationIconRegistry.setIcon = function(operationId, iconPath) {
+  this.map.set(operationId, iconPath);
+}
+
+OperationIconRegistry.getIcon = function(operationId) {
+  return this.map.get(operationId);
+}
+
--- a/main/webapp/modules/core/MOD-INF/controller.js
+++ b/main/webapp/modules/core/MOD-INF/controller.js
@@ -503,6 +503,7 @@ function init() {
       "scripts/project/summary-bar.js",
       "scripts/project/exporters.js",
       "scripts/project/scripting.js",
+      "scripts/project/operation-icons.js",
 
       "scripts/facets/facet.js",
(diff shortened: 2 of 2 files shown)


Abbe98:
Given that this affects the look and feel of OpenRefine/the client rather than the backend, wouldn't it be better to do the operations name->icon on the client-side?

wetneb:
To me it made sense to store this in the same place where operations are registered, and to make that available to the frontend via a command. Would you just store that in a hard-coded map in a JS file then?

Abbe98:
I wouldn’t hard code it but rather make the map an extension point(which OR
would also use to register the defaults). It would also allow extensions to
control the look and feel for operations registered by other
extensions(which I think is a very good thing, especially for those
retheming/white labeling OpenRefine).

wetneb:
Fine by me! Something like that? Intuitively, it seems more future-proof to offer dedicated functions to register and retrieve icons, so that we don't have to stick to this `Map` object in the future.

I am wondering about performance though, after observing in https://github.com/OpenRefine/OpenRefine/pull/7060 how images load only when the corresponding DOM element is created. I thought about defining the icons in CSS, so registering CSS class names instead of paths to files, but that would make it harder to include those icons in a generated SVG image (or canvas), which I need for the recipe generation.

Perhaps it's good enough if we pre-load all icons after the page has loaded, so that they are cached for later? It will still generate quite some requests I guess.

main/webapp/modules/core/scripts/project/operation-icons.js
```diff
@@ -0,0 +1,13 @@
+
+var OperationIconRegistry = {
+  map: new Map()
+};
+
+OperationIconRegistry.setIcon = function(operationId, iconPath) {
+  this.map.set(operationId, iconPath);
```
Abbe98, line 7:
```suggestion
  // this will cause the image to load so that once it is added to the DOM it will already be in the browser cache
  var image = new Image();
  image.src = iconPath;

  this.map.set(operationId, iconPath);
```

main/webapp/modules/core/scripts/project/operation-icons.js
```diff
@@ -0,0 +1,13 @@
+
+var OperationIconRegistry = {
+  map: new Map()
+};
+
+OperationIconRegistry.setIcon = function(operationId, iconPath) {
```
Abbe98, line 6:
```suggestion
OperationIconRegistry.setIcon = function(operationId, iconPath) {
  if (!iconPath.endsWith(".svg")) {
    throw new Error('Operation icon must be a path to a SVG file.');
  }
```

Just so that in case we want to move to a model in which the SVG markup is injected into the DOM we don't need a breaking change.

Abbe98 (changes requested):
> Perhaps it's good enough if we pre-load all icons after the page has loaded, so that they are cached for later? It will still generate quite some requests I guess.

If this becomes an issue we could either set `loading="lazy"` in the DOM or let this extension point load the actual SVG markup and inject that. The later would open up for client side theming anyway.

wetneb:
Makes sense, thanks!

Abbe98 (approved):
Looking forward to see this in use!

PR #7056, opened 2025-01-01
operations: Expose column dependencies and diffs

wetneb:
This pull request adds metadata fields to operations:
* the _column dependencies_, which is a set of column names that are required to be present in the project for this operation to be applicable
* the _columns diff_, which is an object indicating which columns are added, modified or deleted by this operation.

Those two metadata fields are currently not added to the JSON serialization of operations, in the interest of keeping that representation unchanged and because their contents can be derived from other operation parameters. The fields are also optional, to ensure backwards compatibility.

Those fields will enable two things:
* to predict the set of columns present at each stage of the execution of a list of operations, so that we can anticipate missing columns before running any operation. For instance, consider a case where the user wants to apply a recipe which starts by fetching URLs from a column `foo`, storing the responses in column `bar`, and then running a text transform operation on the column `bar`. Thanks to the operation metadata introduced by this PR, we will be able to detect that the recipe requires column `foo` to exist in the project. We can validate that the column `bar` required by the second operation in the recipe is created by the first operation. We can also infer that the recipe requires column `bar` not to exist in the initial project (since it would conflict with the newly created column).
* to offer a graphical representation of a recipe, to help the user understand its structure. I would like this graphical representation to eventually replace the JSON representation that is currently exposed to users.

I imagine that it could be useful to still tweak the public API introduced by this PR, so I would rather not include it in the upcoming 3.9 release, but only merge it afterwards, so that it gives a bit more time to do adjustments if needed.

### DIFF
--- a/modules/core/src/main/java/com/google/refine/model/AbstractOperation.java
+++ b/modules/core/src/main/java/com/google/refine/model/AbstractOperation.java
@@ -33,7 +33,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package com.google.refine.model;
 
+import java.util.Optional;
 import java.util.Properties;
+import java.util.Set;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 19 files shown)


modules/core/src/main/java/com/google/refine/model/AbstractOperation.java
```diff
@@ -99,4 +101,23 @@ public String getOperationId() {
...
+    }
+
+    /**
+     * If the effect of the operation on the set of columns in the project is predictable, this effect can be exposed in
+     * this method. Otherwise, {@link Optional#empty()} can be returned.
+     */
+    @JsonIgnore
+    public Optional<ColumnsDiff> getColumnsDiff() {
+        return Optional.empty();
+    }
```
wetneb, line 122: Those are the two methods introduced by this PR (with their implementations and tests in a bunch of classes).

wetneb:
I plan to merge this soon unless there are any concerns.

PR #7059, opened 2025-01-03
ui: Remove usage of table for layout in menu system

wetneb:
This shouldn't make any difference to users, just simplifies/cleans up the DOM. (The only minor difference I am aware of on Firefox is that the right triangle gets downloaded only once and not requested every time a menu is opened, which seems to be a rather good thing!)

I also changed the introduction of the label to use `$.text()` instead of `$.html()`, which could break some extensions which rely on inserting custom HTML in menus, but makes us safe from special characters in translations of menu entries.


Abbe98 (approved):
Nice, also ensures that the image do not pollute the DOM/accessibility tree. :-) 

I think that the support for HTML-injection here is unintended and we are probably better of without it.

PR #7060, opened 2025-01-04
Add icons to operations in column menu

wetneb:
This adds icons to the column menu, with the aim:
* to help users identify menu items visually, rather than by reading text
* to help users better relate the menu entries to the history entries (once the icons will be made available there, which requires #7055), and then in the graphical representations of recipes
* open up the possibility to move out of the column menus, by integrating operations in a toolbar for instance

The icons are all made by me based on [the design proposed on the forum](https://forum.openrefine.org/t/operation-logos/1307). I have made a few tweaks, mostly to simplify the icons so that they are recognizable at the small scale needed to fit them in operation menus.

The icons can surely be improved further. I don't know what is the best place to discuss changes to them. We can try doing that in this PR. This doesn't cover all operations or menu entries, we would need to add more later on, but I think this is already a rather big batch, which I anticipate you'll have many suggestions of improvements.

Once there is consensus on the icons, I would work on optimizing them for size and making them load ahead of time to avoid the small delay when opening a menu for the first time.

## Screenshots

image

image

image

image

### DIFF
--- a/main/webapp/modules/core/images/operations/add-column.svg
+++ b/main/webapp/modules/core/images/operations/add-column.svg
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="128"
+   height="128"
+   viewBox="0 0 33.866666 33.866667"
+   version="1.1"
+   id="svg5"
+   inkscape:version="1.2.2 (b0a8486541, 2022-12-01)"
+   sodipodi:docname="add-column.svg"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <sodipodi:namedview
+     id="namedview7"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:showpageshadow="2"
+     inkscape:pageopacity="0.0"
+     inkscape:pagecheckerboard="0"
+     inkscape:deskcolor="#d1d1d1"
+     inkscape:document-units="mm"
+     showgrid="true"
+     inkscape:zoom="3.1963805"
+     inkscape:cx="48.49235"
+     inkscape:cy="97.141127"
+     inkscape:window-width="1918"
+     inkscape:window-height="1054"
+     inkscape:window-x="0"
+     inkscape:window-y="24"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="layer1">
+    <inkscape:grid
+       type="xygrid"
+       id="grid1359" />
+  </sodipodi:namedview>
+  <defs
+     id="defs2" />
+  <g
+     inkscape:label="Calque 1"
+     inkscape:groupmode="layer"
+     id="layer1">
+    <path
+       style="fill:none;fill-rule:evenodd;stroke:#000000;stroke-width:3.175;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1;stroke-dasharray:none"
+       d="M 3.9995882,15.937216 H 29.624054"
+       id="path1550" />
+    <path
+       style="fill:none;fill-rule:evenodd;stroke:#000000;stroke-width:3.175;stroke-linecap:butt;stroke-linejoin:miter;stroke-dasharray:none;stroke-opacity:1"
+       d="M 16.754895,3.1514624 V 28.775928"
+       id="path1550-5" />
+  </g>
+</svg>
--- a/main/webapp/modules/core/images/operations/blank-down.svg
+++ b/main/webapp/modules/core/images/operations/blank-down.svg
@@ -1,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="128"
+   height="128"
+   viewBox="0 0 33.866666 33.866667"
+   version="1.1"
+   id="svg5"
+   inkscape:version="1.2.2 (b0a8486541, 2022-12-01)"
+   sodipodi:docname="blank-down.svg"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <sodipodi:namedview
+     id="namedview7"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:showpageshadow="2"
+     inkscape:pageopacity="0.0"
+     inkscape:pagecheckerboard="0"
+     inkscape:deskcolor="#d1d1d1"
+     inkscape:document-units="mm"
+     showgrid="true"
+     inkscape:zoom="1.1300911"
+     inkscape:cx="-76.10006"
+     inkscape:cy="-35.395377"
+     inkscape:window-width="1366"
+     inkscape:window-height="699"
+     inkscape:window-x="0"
+     inkscape:window-y="32"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="layer1"
+     showguides="true">
+    <inkscape:grid
+       type="xygrid"
+       id="grid1359" />
+  </sodipodi:namedview>
+  <defs
+     id="defs2" />
+  <g
+     inkscape:label="Calque 1"
+     inkscape:groupmode="layer"
+     id="layer1">
+    <path
+       style="fill:none;fill-rule:evenodd;stroke:#000000;stroke-width:2.11667;stroke-linecap:butt;stroke-linejoin:miter;stroke-dasharray:none;stroke-opacity:1"
+       d="M 21.087154,14.052839 17.069113,18.68474 13.45242,13.990023"
+       id="path1588-5"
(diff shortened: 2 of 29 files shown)


thadguidry:
Hmm, the icon for **remove this column** leans to much toward a temple or arch?  There's just something about it, compared to other trashcan icons.  I dunno.  Maybe if the lid edges ticked down or were not so extended over the rim?  I'd also probably use only 2 lines within (that would convert better in dark mode, if we ever have that)
image

Also, for reconciliation, I think the 2 overlapping circles might have their intersection darkened?
That's all from me.  Rest looks nice and meaningful.

IIITM-Jay:
Hi @wetneb and @thadguidry , For rename, how about the following icons:

image

image
<br>

image

Thanks!

IIITM-Jay:
Additionally, when I was going through the concerns as stated by @wetneb , as below
> I am wondering about performance though, after observing in https://github.com/OpenRefine/OpenRefine/pull/7060 how images load only when the corresponding DOM element is created. I thought about defining the icons in CSS, so registering CSS class names instead of paths to files, but that would make it harder to include those icons in a generated SVG image (or canvas), which I need for the recipe generation.

> Perhaps it's good enough if we pre-load all icons after the page has loaded, so that they are cached for later? It will still generate quite some requests I guess.

We can have **Image Sprite Sheet** , which will combine all smaller icons/ images into a single sprite. By using this, we can reduce the number of HTTP requests when rendering numerous icons on a page. 

Thanks!
-Jay

wetneb:
@thadguidry thanks a lot, I have updated the icons you mentioned. Happy to do further tweaks!

@IIITM-Jay concerning adding a rectangle (input field) in the rename icon, why not. Would it make it more recognizable to you @thadguidry?

About using a single sprite, yes that's one possibility. I still need to embed those SVG images into larger, machine-generated ones for the recipe visualization widget, so let's see if it's still possible with such a technique. We also want to keep the registration of icons extensible (we shouldn't require all operation icons to come from the same SVG file).
Anyway I think it's the sort of optimization that can be done as a follow-up later.

thadguidry:
I think the 2nd or 3rd one from @IIITM-Jay would be preferable.  I myself had thought of his 2nd suggestion, but didn't want to push on it, but glad he did.  FYI, many Asian language interfaces actually use something very close to that second one.  A B C as the 3 first letters in the Latin alphabet are used in science, medicine throughout the world and taught in schools at the age of 3, so the 2nd one would be quite universally known and easily recognized rename icon.

ostephens:
These look great @wetneb 

The only icon I'd question is the use of the capital Sigma for "Transform". This seems out of step with the other icons (as it is a greek letter rather than any attempt to illustrate the action to be taken) and also could be confusing as uppercase sigma is used to mean "sum" in mathematics. 

I appreciate that Google Sheets uses Σ for its `function` menu option. It makes slightly more sense in this context because summing is such a commonly used function in spreadsheets - but much less commonly in OR. Even in this context I don't think it's great.

Excel uses `fx` and I think this, or `f(x)` is better as it more generally says "some sort of function" but I still don't love it

Looking for some alternative icons, these options seemed more illustrative in a similar way to the other icons you have proposed:

- https://www.flaticon.com/free-icon/data-transformation_7440452?term=transformation&page=1&position=1&origin=tag&related_id=7440452
- https://www.flaticon.com/free-icon/transition_5405813?term=transform&page=1&position=68&origin=search&related_id=5405813
- https://thenounproject.com/icon/transformation-410836/
- https://thenounproject.com/icon/transformation-3767718/

thadguidry:
If I look to inspiration how the original designers thought about things...who tried to keep things aligned to our mission/vision "easy to use powertool for working with messy data : for non-programmers"

We then might instead look towards real world items... those that folks are most familiar with and universally recognize.
We might think of instead of a function how non-programmers might view it as -**building** or **making** something from parts, which in a realistic way (not math or programming) is what a function does; returns a value from parts after molding, shaping, amening, working other values or their parts.

To that end, we might look at real world tools such as:
- hammer
- hammer / wrench
image
- hammer / anvil
image
- toolbox? hmm, not as universal I think.
- pickaxe
image
- magic wand we already have for Common Transforms so... maybe we make that as the Transform, and Common Transforms becomes one of above like hammer / wrench or something else?
image

thadguidry:
For me with @ostephens suggestions, I only feel that his 4th example icon might stick in people's minds long term, as it shows a morph or transform from a circle to square.

thadguidry:
We'll have to be extra careful with "Common Transforms" because common translated in different languages sometimes subtly means "everyday thing", "often", "universal", etc.  So whichever icon we choose for Common Transforms needs to stick easily in peoples minds for that across languages. 😅

wetneb:
Thanks for the feedback!
I have updated the rename and transform icons as suggested. To me, they look less readable in this form though. What do you think?

image

image

It's worth keeping in mind that our column menu entries are tiny, so icons need to be extremely simple to remain understandable at such a small scale. An icon that might look attractive and understandable in an icon repository at a 256x256 px scale will look very different once it's inserted in the UI.

ostephens:
@wetneb I thought the magic wand for 'Common transformations' was good - and reading @thadguidry above he also seems to feel it is appropriate - so think that can be re-instated

ostephens:
@wetneb I definitely appreciate that once the icons are small they are much less readable (the 'rename' icon suffers from this IMO). What do you think about using `fx` or `f(x)` as clearer and more familiar to users?

thadguidry:
@wetneb Well, I've always said that the UI space is underutilized in many places of OpenRefine.  We can use some of that space.
They are less readable because they are a bit too small.  You need to enlarge them a bit.  Look at the amount of wasted space that could be used if you enlarged them and allow to extend more into the div margins of each menu item without sacrificing the readability of each menu item.

My hacky 5 second MS Paint example... I enlarged the icons only (ignore blurry, it's not Photoshop) and nothing else just to show the amount of space available without encroaching over div margins of each menu item.  Plenty of space.  Enlarge them would be my suggestion.

image

thadguidry:
In later OpenRefine releases, I'd also go as far as to say that we don't even have the Transform item in the menu... I'd move it into a right click option on any cell in any column that provides quick commons options including Transform.  It's something we almost did in the early days but lacked time and Jquery was fiddly to handle that, but not anymore.  Anyways, I digress.

wetneb:
Sorry about the disappearance of the magic wand, it was unintentional (I need wizardry training to handle such items!)

I have scaled icons up, which does help a bit, but it's unclear to me if the rename icon is still clear enough (the "a" is still tiny).

image

image

Here it is with `f(x)`:
image

thadguidry:
@wetneb much better!  I dunno, the f(x) is sticking to my head now.  Thanks @ostephens ! grrrrr!  lololol

Can you change it so that it's using the Unicode math symbols themselves (and then make the SVG from those)? -->
ƒ(𝑥)

thadguidry:
@wetneb Curious, How are you generating the SVG's?

ostephens:
Thanks @wetneb. I think ƒ(𝑥) works as the icon (happy if it's tweaked to use symbol suggested by @thadguidry but as a general principle happy with it's use here)

magdmartin:
Thanks for the integration @wetneb   I am also happy with the direction it is going. 

I want to confirm that all the operations under  `Facet`,`Commons transforms`, and `View` will be icon-less, and we are ok with having menu options with and without icons throughout the application. 

When I first looked at the fill-down icon, I thought it was an hourglass. On the top of @thadguidry comment, maybe we can make the top and bottom bar wider to match the size of other icons representing a row  (like the transpose or split/join icon)

The replace icon makes me think of the operation `add a column`. The replace function is often represented with this type of icon 
https://thenounproject.com/icon/replace-451579/ or even https://thenounproject.com/icon/transformation-3767718/

wetneb:
> I want to confirm that all the operations under Facet,Commons transforms, and View will be icon-less, and we are ok with having menu options with and without icons throughout the application.

I plan to add icons for all operations, this was just a first batch of them.
I would be happy to add icons for other things (such as facets) if people see the need for them (and have ideas how to represent them), but at the moment the priority is to cover operations because those icons will be used in menus, history and recipe visualization.

> The replace icon makes me think of the operation add a column. The replace function is often represented with this type of icon
> https://thenounproject.com/icon/replace-451579/ or even https://thenounproject.com/icon/transformation-3767718/

So, I could reuse the icon made for the "transform" operation and use it for replace, since we're going for "f(x)" for the transform? But I am not sure about having arrows in two directions for a replace operation, because it's really just going in one.

thadguidry:
@wetneb Agree, and that's the problem, the bi-direction of arrows (cycling or syncing) which throws off the context quite a bit.

I think synonymous with the term "replace" are the terms, substitute, switch, etc.?  And when searching around for icons under those terms, I find much stronger symbology, I think.  Such as this example:
image

https://www.flaticon.com/free-icon/shuffle_724979?term=substitute&page=1&position=3&origin=tag&related_id=724979

But that above substitute icon seems to also lose some context in the thought of a replace function (a string becomes a different string).  Which replace is somewhat similar to rename, somewhat.  Since we have a small letter `a` in the rename iconography, perhaps we can go with that same logic and make an icon like so:

(quick hack with unicode arrows - but I'd probably make the arrow a bit bigger and in the style of the other arrows used in the @wetneb icons thus far) 
`ab → a1`
`ab → xy`
`a ↦ b`

magdmartin:
> > The replace icon makes me think of the operation add a column. The replace function is often represented with this type of icon
> > https://thenounproject.com/icon/replace-451579/ or even https://thenounproject.com/icon/transformation-3767718/
> 
> So, I could reuse the icon made for the "transform" operation and use it for replace, since we're going for "f(x)" for the transform? But I am not sure about having arrows in two directions for a replace operation, because it's really just going in one.

I agree and I am OK going with a uni-directional arrow like this example: https://thenounproject.com/icon/transformation-410836/

wetneb:
The feedback on the icon design has stalled, so I am thinking it would be worth merging this. I am not super happy with the icons, I think they could be a lot more polished, but I hope to get more feedback on them during user testing for the recipe visualization.
Since this will not be part of 3.9, there should be plenty of time to refine the icons before 3.10.
Also, it would probably be useful to reduce the number of requests by bundling them (sprite sheet, CSS).

PR #7058, opened 2025-01-02
ci: Switch locale to French for Java tests

wetneb:
Fixes #7054.
I expect the CI to fail first, and require adapting the tests.

PR #7061, opened 2025-01-05
fixes #7025: Keep only matching rows

### LINKED ISSUE #7025: Keep only matching rows
We have a "Remove matching rows" menu entry in the "All" menu, which is helpful, but not completely ideal in terms of usability.

Say I just created a project and want to trim the project to a subset of rows, defined by multiple facets A, B and C.
The current way to do it would be to:
* add facet A, invert it to select the rows which don't match the criterion, remove matching rows, remove facet A
* add facet B, invert it, remove matching rows, remove facet B
* add facet C, invert it, remove matching rows, remove facet C

That's a lot of steps. It's primarily due to the fact that I cannot negate a list of facets, but only each facet independently.
But arguably it's also a bit weird to have to select the rows you want to remove. It has the downside that once I run the "Remove matching rows", I am left with an empty grid until I remove or reset at least one facet.

### Proposed solution

In addition to "Remove matching rows", offer a "Keep only matching rows" option, which deletes all rows/records not selected by the current facets.

Running the operation could potentially remove the active facets too, as they should generally not be needed afterwards (since all remaining rows match them).

### Alternatives considered

Offer a way to configure whether facets are combined using the "AND" or "OR" boolean operators would improve the workflow above slightly and would be useful in other contexts.

kalindafab:
Fixes #7025 

- Added New Operation: RowKeepMatchedOperation to allow users to retain only rows that match specific criteria.
- Added a corresponding sub-option under the Rows menu
- Only rows meeting all conditions specified in the facets are preserved.

I also explored adding a toggle for the AND/OR operator in facets filtering, but I faced challenges where the second facet disappeared when using the OR operator. This is something that might need additional discussion or a separate PR.

### DIFF
--- a/main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
+++ b/main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
@@ -2,6 +2,7 @@
 package com.google.refine.operations.row;
 
 import static com.google.refine.operations.OperationDescription.row_keep_matching_brief;
+import static com.google.refine.operations.OperationDescription.row_keep_matching_rows_desc;
 
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -33,6 +34,9 @@ public class RowKeepMatchedOperation extends EngineDependentOperation {
     protected String getBriefDescription(Project project) {
         return row_keep_matching_brief();
 
+    }
+    protected String createDescription(Project project, int rowCount) {
+        return row_keep_matching_rows_desc(rowCount);
     }
 
     @Override
@@ -67,7 +71,7 @@ public class RowKeepMatchedOperation extends EngineDependentOperation {
         return new HistoryEntry(
                 historyEntryID,
                 project,
-                row_keep_matching_brief() + " - " + rowsToRemove.size() + " rows removed",
+                row_keep_matching_brief() + " - "+createDescription(project, rowsToRemove.size()),
                 this,
                 new RowRemovalChange(rowsToRemove));
     }
--- a/main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
+++ b/main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
@@ -1,0 +1,72 @@
+
+package com.google.refine.operations.row;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import com.google.refine.browsing.Engine;
+import com.google.refine.browsing.EngineConfig;
+import com.google.refine.browsing.FilteredRows;
+import com.google.refine.browsing.RowVisitor;
+import com.google.refine.history.HistoryEntry;
+import com.google.refine.model.Project;
+import com.google.refine.model.Row;
+import com.google.refine.model.changes.RowRemovalChange;
+import com.google.refine.operations.EngineDependentOperation;
+
+public class RowKeepMatchedOperation extends EngineDependentOperation {
+
+    @JsonCreator
+    public RowKeepMatchedOperation(
+            @JsonProperty("engineConfig") EngineConfig engineConfig) {
+        super(engineConfig);
+    }
+
+    @Override
+    protected String getBriefDescription(Project project) {
+        return "Keep only matching rows";
+
+    }
+
+    @Override
+    protected HistoryEntry createHistoryEntry(Project project, long historyEntryID) throws Exception {
+        Engine engine = createEngine(project);
+        FilteredRows filteredRows = engine.getAllFilteredRows();
+
+        Set<Integer> matchedRows = new HashSet<>();
+
+        filteredRows.accept(project, new RowVisitor() {
+
+            @Override
+            public void start(Project project) {
+            }
+
+            @Override
+            public void end(Project project) {
+            }
+
+            @Override
+            public boolean visit(Project project, int rowIndex, Row row) {
+                matchedRows.add(rowIndex);
+                return false;
+            }
+        });
+        List<Integer> rowsToRemove = new ArrayList<>(project.rows.size() - matchedRows.size());
+        for (int i = 0; i < project.rows.size(); i++) {
+            if (!matchedRows.contains(i)) {
+                rowsToRemove.add(i);
+            }
+        }
+        return new HistoryEntry(
+                historyEntryID,
+                project,
+                "Keep only matching rows (" + rowsToRemove.size() + " rows removed)",
+                this,
+                new RowRemovalChange(rowsToRemove));
+    }
+}
--- a/main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
+++ b/main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
@@ -67,7 +67,7 @@ public class RowKeepMatchedOperation extends EngineDependentOperation {
         return new HistoryEntry(
                 historyEntryID,
                 project,
-                row_keep_matching_brief() + " (" + rowsToRemove.size() + " rows removed)",
+                row_keep_matching_brief() + " - " + rowsToRemove.size() + " rows removed",
                 this,
                 new RowRemovalChange(rowsToRemove));
     }
--- a/main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
+++ b/main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
@@ -35,8 +35,9 @@ public class RowKeepMatchedOperation extends EngineDependentOperation {
         return row_keep_matching_brief();
 
     }
+
     protected String createDescription(Project project, int rowCount) {
-           return row_keep_matching_rows_desc(rowCount);
+        return row_keep_matching_rows_desc(rowCount);
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


kalindafab:
@wetneb would you be able to review my PR when you have a moment? I appreciate your time and feedback!

main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @JsonCreator
+    public RowKeepMatchedOperation(
+            @JsonProperty("engineConfig") EngineConfig engineConfig) {
+        super(engineConfig);
+    }
+
+    @Override
+    protected String getBriefDescription(Project project) {
+        return "Keep only matching rows";
```
wetneb, line 32: This could be localized (especially since you have already introduced the `row_keep_matching_brief` message for it).

main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
```diff
@@ -0,0 +1,72 @@
...
+        List<Integer> rowsToRemove = new ArrayList<>(project.rows.size() - matchedRows.size());
+        for (int i = 0; i < project.rows.size(); i++) {
+            if (!matchedRows.contains(i)) {
+                rowsToRemove.add(i);
+            }
+        }
+        return new HistoryEntry(
+                historyEntryID,
+                project,
+                "Keep only matching rows (" + rowsToRemove.size() + " rows removed)",
```
wetneb, line 68: This could also be localized.

main/tests/server/src/com/google/refine/operations/row/RowKeepMatchedOperationTest.java
```diff
@@ -0,0 +1,204 @@
...
+
+    Project project;
+    ListFacetConfig facet;
+
+    @Test
+    public void serializeRowKeepMatchedOperation() throws IOException {
+        String json = "{"
+                + "\"op\":\"core/row-keep-matched\","
+                + "\"engineConfig\":{\"facets\":[],\"mode\":\"row-based\"},"
+                + "\"description\":\"Keep only matching rows\""
```
wetneb, line 67: Once you have localized the operation description, this will need adapting because of #7058 (let me merge it already).

main/tests/server/src/com/google/refine/operations/row/RowKeepMatchedOperationTest.java
```diff
@@ -0,0 +1,204 @@
...
+                + "}";
+
+        RowKeepMatchedOperation operation = ParsingUtilities.mapper.readValue(json, RowKeepMatchedOperation.class);
+        String serialized = ParsingUtilities.mapper.writeValueAsString(operation);
+
+//         System.out.println("Expected: " + json);
+//         System.out.println("Actual: " + serialized);
```
wetneb, line 74: Those comments can likely be removed.

main/webapp/modules/core/scripts/facets/list-facet.js
```diff
@@ -185,6 +187,12 @@ class ListFacet extends Facet {
             '<a href="javascript:{}" class="facet-choice-link" bind="changeButton">'+$.i18n('core-facets/change')+'</a>' +
             '<span bind="titleSpan"></span>' +
           '</td>' +
+         '<td width="1%">' +
+         '<div class="facet-mode-buttons">' +
+         '<a href="javascript:{}" title="AND mode" class="facet-combine-toggle' + (this._combineMode === "AND" ? ' active' : '') + '" data-mode="AND" bind="andModeButton">&amp;</a>' +
+         '<a href="javascript:{}" title="OR mode" class="facet-combine-toggle' + (this._combineMode === "OR" ? ' active' : '') + '" data-mode="OR" bind="orModeButton">∥</a>' +
+         '</div>' +
+        '</td>' +
```
wetneb, line 195: I would remove all the changes made to this file, since I agree this requires more design and should be left to another PR.

main/webapp/modules/core/styles/project/facets.css
```diff
@@ -357,3 +357,31 @@ img.facet-choice-link {
...
+  background: #4b82d3;
+  color: #fff;
+}
+.facet-combine-toggle:hover {
+  background: #eee;
+}
+
+.facet-combine-toggle.active:hover {
+  background: #3b71c2;
+}
```
wetneb, line 387: Same for this file, this seems to be only related to the AND/OR toggle - let's keep this file unchanged.

wetneb (commented):
Thanks a lot for tackling this! It looks really good to me, just a couple of comments which should hopefully be easy to address :)

main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
```diff
@@ -65,7 +67,7 @@ public boolean visit(Project project, int rowIndex, Row row) {
         return new HistoryEntry(
                 historyEntryID,
                 project,
-                "Keep only matching rows (" + rowsToRemove.size() + " rows removed)",
+                row_keep_matching_brief() + " (" + rowsToRemove.size() + " rows removed)",
```
wetneb, line 70: It would be better to have a separate message with a parameter, for the entire string. This will otherwise generate weird mixes of languages.

main/webapp/modules/core/scripts/facets/list-facet.js
kalindafab: you may remove all changes regarding AND/OR toggle.

main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
kalindafab: right! Let me fix it

main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
```diff
@@ -0,0 +1,74 @@
...
+        List<Integer> rowsToRemove = new ArrayList<>(project.rows.size() - matchedRows.size());
+        for (int i = 0; i < project.rows.size(); i++) {
+            if (!matchedRows.contains(i)) {
+                rowsToRemove.add(i);
+            }
+        }
+        return new HistoryEntry(
+                historyEntryID,
+                project,
+                row_keep_matching_brief() + " - " + rowsToRemove.size() + " rows removed",
```
wetneb, line 70: This is still a mix of languages: if OpenRefine is running in Italian and a translator added an Italian translation, then this will generate something like:
"Mantenere le righe corrispondenti - 34 rows removed"
This should be avoided, since it contains two different languages in the same phrase.

As I tried to explain before, we need to make the whole message translatable. This requires introducing a new message id, with a parameter containing the number of rows removed. To understand how to add parameters to a message, you could take inspiration from the `recon_clear_similar_cells_desc` message. Could you please do that?

kalindafab:
how about this one?

main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
magdmartin: For reference the section in our doc regarding [Adding a new string for localization](https://openrefine.org/docs/technical-reference/translating-ui#front-end-development)

main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
```diff
@@ -67,7 +71,7 @@ public boolean visit(Project project, int rowIndex, Row row) {
         return new HistoryEntry(
                 historyEntryID,
                 project,
-                row_keep_matching_brief() + " - " + rowsToRemove.size() + " rows removed",
+                row_keep_matching_brief() + " - "+createDescription(project, rowsToRemove.size()),
```
wetneb, line 74: That is sadly still not enough. The " - " you have inserted between the two messages might not make sense in all languages, so it is better to have a single message which spans over the whole description.

The message could be:
`Keep only matching rows ({0} rows removed)`

This makes it possible to give full control to the translators over how to translate the entire string in their language.

main/src/com/google/refine/operations/row/RowKeepMatchedOperation.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    protected String getBriefDescription(Project project) {
+        return row_keep_matching_brief();
+
+    }
+    protected String createDescription(Project project, int rowCount) {
+        return String.format("%s (%s)",
+                row_keep_matching_brief(),
+                row_keep_matching_rows_desc(rowCount)
```
wetneb, line 41: No, you are still concatenating two messages translated independently.
I would like you to use a single message for the entire string.

This means:
* `row_keep_matching_rows_desc=Keep only matching rows ({0} rows removed)`  in the properties file
* `return row_keep_matching_rows_desc(rowCount);` in the Java method.

I am confused as to what has been unclear in my messages above. Could you help me understand how I could have made them clearer, when I wrote "It would be better to have a separate message with a parameter, for the entire string"?

wetneb (changes requested):
Thank you, that's much better!
I tried it out interactively and it works great. I would only suggest to put the new menu entry directly underneath "Remove matching rows" because it feels to me that those two options belong together. You could perhaps even add a horizontal separator before the next option (Remove duplicate rows).
image

Also, you'll need to run `./refine lint` to apply our code style.

wetneb (approved):
Thanks a lot!

PR #7064, opened 2025-01-06
operations: Add validate() calls in commands

wetneb:
This adds calls to the `AbstractOperation::validate()` method in all operation commands.

This should generally not have an impact on user experience, because we expect the front-end to already validate operation parameters before submitting them to the backend. This is more of an additional safety net, in case that validation isn't done properly. The goal is to return an error message that's as helpful as possible instead of a random Java exception.

On the long term, I think we should deprecate all those operation commands and just use the generic `apply-operations` command with the operation represented in JSON. This would avoid having to devise how to represent operation metadata in two different ways (JSON and url-parameter encoding).


Abbe98 (approved):
This is great, especially for API users and third-party clients.

PR #7065, opened 2025-01-06
facets: Add a ability to translate after a column rename

wetneb:
For #133.

This doesn't just translate the columns to which the facets are applied, but also the expressions they rely on, in case other columns are mentioned there.

Beyond #133, this will then be used to offer similar functionality on operation classes, so that the column names referenced by a recipe (JSON representation of operation history) can be adapted when applying it. See the design discussions on the forum: https://forum.openrefine.org/t/column-selection-ui-for-applying-recipes/1998


wetneb:
Also merging this one soon unless anyone judges it not ripe enough.

Issue #95, opened 2012-10-15
Timeline facet


tfmorris:
_Original author: iainsproat (June 23, 2010 17:14:05)_

Feature suggestion for a timeline facet in Gridworks.

This works in all manners similar to a scatter facet, except that the x axis handles dates only.  Dates should be properly handled by Gridworks.

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=95_

PR #7068, opened 2025-01-07
Wikibase : New media upload support for validation of mandatory properties

sunilnatraj:
Fixes [#95](https://github.com/OpenRefine/CommonsExtension/issues/95)

- Support to define mandatory properties in New media upload
- Raise warnings if any of the mandatory properties is not set
- Updated manifest schema to validate new entries for the newly added mandatory properties list
- Updated Warning to not set example entity for new media case as the label resolver fails for non-existent entity
Screenshot 2025-01-07 at 2 41 10 PM


sunilnatraj:
@wetneb Can you review PR to update the manifest used in tests. [https://github.com/OpenRefine/wikibase-manifests/pull/14](url)

extensions/wikibase/tests/data/manifest/commons-manifest-v2.0.json
```diff
@@ -4,7 +4,8 @@
     "name": "Wikimedia Commons",
     "root": "https://commons.wikimedia.org/wiki/",
     "main_page": "https://commons.wikimedia.org/wiki/Main_Page",
-    "api": "https://commons.wikimedia.org/w/api.php"
+    "api": "https://commons.wikimedia.org/w/api.php",
+    "mandatoryMediaInfoPropertyIds": "P7482, P170, P6216"
```
wetneb, line 8: As [written earlier](https://github.com/OpenRefine/CommonsExtension/issues/95#issuecomment-2473135605):
> Instead of encoding lists as strings, I would use JSON's native ability to represent arrays: `["P7482", "P571", "P170", "P6216", "P275"]` instead of `"P7482, P571, P170, P6216, P275"`.

extensions/wikibase/module/langs/translation-en.json
```diff
@@ -273,13 +273,15 @@
     "warnings-messages/new-media-created/title": "New media uploaded",
     "warnings-messages/new-media-created/body": "This batch will upload new media files. Please make sure the files you upload are admissible on this wiki.",
     "warnings-messages/new-media-without-file-path/title": "New media without file path",
-    "warnings-messages/new-media-without-file-path/body": "Uploading new media files requires supplying a path (or URL) to the file to upload. No path was provided for {example_entity}, for instance.",
+    "warnings-messages/new-media-without-file-path/body": "Uploading new media files requires supplying a path (or URL) to the file to upload. No path was provided.",
```
wetneb, line 276: Why remove the example? Wouldn't it also be possible to offer an example entity for the new validation errors that you are introducing?

main/tests/cypress/cypress/e2e/extensions/wikibase/select_wikibase_instance.cy.js
```diff
@@ -133,6 +134,56 @@ describe(__filename, function () {
...
+            'val',
+            JSON.stringify(manifest)
+        );
+
+        cy.get('.add-wikibase-dialog button').contains('Add Wikibase').click();
+
+        // Validate alert for media wiki missing property
+        // manifest.mediawiki should have required property 'mandatoryMediaInfoPropertyIds'
+        cy.on('window:alert', (alertText) => {
+              expect(alertText).to.equal('manifest.mediawiki should have required property \'mandatoryMediaInfoPropertyIds\'');
```
wetneb, line 183: I don't think we should require manifests to contain the `mandatoryMediaInfoPropertyIds` field. To the contrary, we want to continue support manifests written for earlier versions of OpenRefine, so we should not fail if this property is absent.

wetneb (commented):
Thanks for tackling this!

extensions/wikibase/tests/data/manifest/commons-manifest-v2.0.json
sunilnatraj: Apologies I missed this. Fixed.

main/tests/cypress/cypress/e2e/extensions/wikibase/select_wikibase_instance.cy.js
sunilnatraj: Removed from mandatory keys list in schema definition. UI test case removed.

extensions/wikibase/module/langs/translation-en.json
sunilnatraj: The reason to remove the example is that in the case of new entity flow, the label resolution for warnings fail as in this case the example is an invalid entity reference. 
I am assuming we cannot use a static example.
Any suggestions @wetneb

sunilnatraj:
@wetneb Can you also review this [https://github.com/OpenRefine/wikibase-manifests/pull/14](url)

extensions/wikibase/module/langs/translation-en.json
wetneb: But the existing messages (`new-media-without-file-path` for instance) were already referring to new entities as examples, right? So that must be supported in some way, I assume. If this is broken, we should fix it (probably independently of this PR). To provide a name for a new entity, we can use either the text in the reconciled cell (which is more or less guaranteed to be non-empty I think) or the label added in the schema (if any), and we can add a fallback to a constant ("unnamed entity") if need be.

wetneb: Ah, so that's a bug introduced by #6895, because that new label fetching logic should not apply to new entities. Would you be open to:
* fixing that bug in a separate PR
* retaining the example.entities in this PR?

sunilnatraj: @wetneb I have addressed both in this PR; i hope that is ok.

extensions/wikibase/src/org/openrefine/wikibase/qa/EditInspector.java
```diff
@@ -93,6 +93,12 @@ public class EditInspector {
     private ApiConnection connection;
     private boolean slowMode;
 
+    // Add all warning types for new entity where the new entity is provided as example
+    private static List<String> warningTypesFilterListforLabelResolver = List.of(
+            NewEntityScrutinizer.newMediaWithoutFileNameType,
+            NewEntityScrutinizer.newMediaWithoutWikitextType,
+            NewEntityScrutinizer.newMediaWithoutFilePathType);
```
wetneb, line 100: I wouldn't hard code this here, since we might introduce other checks involving new entities later on.
Instead, we can make `EditInspector` to work with new entities in QA warning messages: see #7090.

sunilnatraj: @wetneb I reviewed your changes in 7090.  I will rollback my changes in this PR.

wetneb: Thanks a lot!

sunilnatraj: Reverted changes.

extensions/wikibase/tests/data/manifest/wikidata-manifest-v1.0.json
```diff
@@ -4,7 +4,8 @@
         "name": "Wikidata",
         "root": "https://www.wikidata.org/wiki/",
         "main_page": "https://www.wikidata.org/wiki/Wikidata:Main_Page",
-        "api": "https://www.wikidata.org/w/api.php"
+        "api": "https://www.wikidata.org/w/api.php",
+        "mandatoryMediaInfoPropertyIds": ["P7482", "P170", "P6216"]
```
wetneb, line 8: I would leave this field out, since it does not make sense for Wikidata.

extensions/wikibase/src/org/openrefine/wikibase/manifests/ManifestV1.java
```diff
@@ -22,15 +23,22 @@ public class ManifestV1 implements Manifest {
...
     public ManifestV1(JsonNode manifest) {
         version = manifest.path("version").textValue();
 
         JsonNode mediawiki = manifest.path("mediawiki");
         name = mediawiki.path("name").textValue();
         mediaWikiApiEndpoint = mediawiki.path("api").textValue();
+        mandatoryMediaInfoPropertyIds = new ArrayList<>();
+        for (JsonNode node : mediawiki.path("mandatoryMediaInfoPropertyIds")) {
+            mandatoryMediaInfoPropertyIds.add(node.asText());
+        }
```
wetneb, line 41: When `mandatoryMediaInfoPropertyIds` is not present and the `mediaWikiApiEndpoint` is that of Wikimedia Commons, we could pre-fill `mandatoryMediaInfoPropertyIds` with `["P7482", "P170", "P6216"]`.

wetneb (changes requested):
Very nice, I think we are nearly there.

I wonder if "critical" is the right severity level for those new checks. I would intuitively expect one level lower, because the user could still have legitimate reasons to upload without those properties being present. The "critical" level should be only used for problems that really cause technical issues during the upload, I would say.

When testing your pull request interactively, the QA warnings did not show up, because I already had the older Wikimedia Commons manifest installed in my OpenRefine instance. This is an issue that other users will likely encounter. I'm not really sure what the right solution to this problem looks like, but we could add a special case in the parsing code to add pre-defined property ids for Commons (see below).

sunilnatraj:
@wetneb Thanks for the review and feedback.

- Issue severity as Critical : I set it as critical as these properties have been identified as mandatory by community. Setting it as IMPORTANT
- Regarding manifest file, my understanding is that when end users upgrade they delete the older version and install the new version. This will ensure that the updated manifest is picked up. Let me know if this is not correct

extensions/wikibase/src/org/openrefine/wikibase/manifests/ManifestV1.java
sunilnatraj: @wetneb Per earlier discussion in the PR to update the manifest files, the decision was with end users to include the mandatory properties in the manifest definition. If we default it will apply to all the wikibase defintions. Is that expected?

extensions/wikibase/tests/data/manifest/wikidata-manifest-v1.0.json
sunilnatraj: Updated

wetneb:
> Regarding manifest file, my understanding is that when end users upgrade they delete the older version and install the new version. This will ensure that the updated manifest is picked up. Let me know if this is not correct

I think users typically upgrade by replacing OpenRefine by a new version, but keep their workspace where Wikibase manifests are stored. So I would find it unlikely that people spend time removing Wikimedia Commons from their installed Wikibases to add it again, especially given that adding a new Wikibase instance isn't exactly user-friendly…

extensions/wikibase/src/org/openrefine/wikibase/manifests/ManifestV1.java
```diff
@@ -22,15 +23,26 @@ public class ManifestV1 implements Manifest {
...
+        if (mandatoryMediaInfoPropertyIds.isEmpty()) {
+            mandatoryMediaInfoPropertyIds = new ArrayList<>(defaultMandatoryMediaInfoPropertyIds);
+        }
```
wetneb, line 45: This code would mean that it's impossible to declare a Wikibase instance without mandatory media info property ids. I think it's not ideal, because there might be instances where no particular properties are mandated.

Instead, I would restrict this fallback to:
* checking that the mediaWikiApiEndpoint is that of Wikimedia Commons, as I mentioned earlier. You could do this with `"https://commons.wikimedia.org/w/api.php".equals(mediaWikiApiEndpoint)`
* checking that the JSON field `"mandatoryMediaInfoPropertyId"` does not exist in the JSON manifest (which is different from being empty). This should likely work with something like `!mediawiki.has("mandatoryMediaInfoPropertyId")`.

sunilnatraj: Updated logic

main/tests/cypress/cypress/e2e/extensions/wikibase/select_wikibase_instance.cy.js
```diff
@@ -80,6 +80,7 @@ describe(__filename, function () {
                 root: 'https://or-wikibase-test.wiki.opencura.com/wiki/',
                 main_page: 'https://or-wikibase-test.wiki.opencura.com/wiki/Main_Page',
                 api: 'https://or-wikibase-test.wiki.opencura.com/w/api.php',
+                mandatoryMediaInfoPropertyIds: ["P7482", "P170", "P6216"]
```
wetneb, line 83: I think we can also remove this line, since it is specific to Wikimedia Commons.

extensions/wikibase/tests/data/manifest/wikidata-manifest-v1.0.json
sunilnatraj: @wetneb The wikidata manifest is used in the NewEntityScrutinzer which is why the mandatory fields key has been included.

wetneb: Ok, it's not super elegant but I guess it will do.

PR #7072, opened 2025-01-08
fixes #6878: Allow building and packaging Windows installer on Windows

### LINKED ISSUE #6878: Replace $VERSION by ${project.version} in WEB-INF/web.xml
##Context
Eventually, we'd like contributors to actually be able to run `mvn package` on Windows, instead of currently only Mac users.  This issue was opened to begin to investigate the `/packaging/pom.xml` and start to untangle things better for a profile for Windows, which currently fails to successfully produce an OpenRefine Windows package on Windows (sigh).  But first, we should take care of TODO's perhaps?  Previous maintainers left some TODO's without creating an issue to resolve it.  So this is that issue and my attempt to research and find a resolution.

We currently have a workaround for the `web.xml` version string replacement during packaging here:
https://github.com/OpenRefine/OpenRefine/blob/11168a425bfc829a8d81bc58c2cacb93cca19b52/packaging/pom.xml#L232-L23

I did some research, and it seems the Maven War Plugin could be used to automatically replace property placeholders (string interpolation) within the `web.xml` file?

Current TODO's are found under
- https://github.com/OpenRefine/OpenRefine/blob/11168a425bfc829a8d81bc58c2cacb93cca19b52/packaging/linux.xml#L66
- https://github.com/OpenRefine/OpenRefine/blob/11168a425bfc829a8d81bc58c2cacb93cca19b52/packaging/windows-with-jre.xml#L79

Example snippets that could be added to our main pom.xml :

### Example 1
```
      <!-- War plugin used to automatically replace property placeholders,
       such as ${project.version} within /WEB-INF when running 'mvn package'
       -->
      <plugin>
        <artifactId>maven-war-plugin</artifactId>
        <version>3.4</version>
        <configuration>
        <!-- this is weird but "filtering" here actually also does "replacing" of property placeholders -->
          <filteringDeploymentDescriptors>true</filteringDeploymentDescriptors>
        </configuration>
      </plugin>
```

Or more specifically it seems that we could also just apply "filtering" to resources that we want or need:
### Example 2
```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-war-plugin</artifactId>
    <configuration>
        <webResources>
            <resource>
                <filtering>true</filtering>
                <directory>src/main/webapp</directory>
                <includes>
                    <include>**/web.xml</include>
                </includes>
            </resource>
        </webResources>
    </configuration>
</plugin>
```

References:
1. https://maven.apache.org/plugins/maven-war-plugin/faq.html
2. https://stackoverflow.com/questions/15355708/how-to-replace-a-value-in-web-xml-with-a-maven-property
3. https://stackoverflow.com/questions/40430305/how-to-do-string-replacements-with-maven-war-plugin-2-6

kalindafab:
Fixes #6878 

- added dynamic version replacement 
- added filtering for specific resources
- there was chmod cmd which fails to run on windows  so i made executable to macos only.

@thadguidry what do you think about this?


thadguidry:
@kalindafab Did you get the package to be generated successfully?  I get testing errors.
When I checked out your PR, and run on Windows 11, for example, with this command `mvn '-Drevision=3.9.0' clean compile` I get test errors such as the following:

```
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  33.284 s
[INFO] Finished at: 2025-01-09T13:02:16+08:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:testCompile (default-testCompile) on project grel: Compilation failure: Compilation failure:
[ERROR] /F:/GitHubRepos/OpenRefine/modules/grel/src/test/java/com/google/refine/expr/functions/strings/SmartSplitTests.java:[40,30] cannot access com.google.refine.grel.ControlFunctionRegistry
[ERROR]   bad class file: F:\GitHubRepos\OpenRefine\modules\grel\target\classes\com\google\refine\grel\ControlFunctionRegistry.class
[ERROR]     unable to access file: java.nio.file.NoSuchFileException: F:\GitHubRepos\OpenRefine\modules\grel\target\classes\com\google\refine\grel\ControlFunctionRegistry.class
[ERROR]     Please remove or make sure it appears in the correct subdirectory of the classpath.
[ERROR] /F:/GitHubRepos/OpenRefine/modules/grel/src/test/java/com/google/refine/grel/ast/ControlCallExprTest.java:[13,30] cannot access com.google.refine.grel.Control
[ERROR]   bad class file: F:\GitHubRepos\OpenRefine\modules\grel\target\classes\com\google\refine\grel\Control.class
[ERROR]     unable to access file: java.nio.file.NoSuchFileException: F:\GitHubRepos\OpenRefine\modules\grel\target\classes\com\google\refine\grel\Control.class
[ERROR]     Please remove or make sure it appears in the correct subdirectory of the classpath.
[ERROR] /F:/GitHubRepos/OpenRefine/modules/grel/src/test/java/com/google/refine/grel/ast/ControlCallExprTest.java:[17,5] cannot find symbol    
[ERROR]   symbol:   class Control
[ERROR]   location: class com.google.refine.grel.ast.ControlCallExprTest
[ERROR] /F:/GitHubRepos/OpenRefine/modules/grel/src/test/java/com/google/refine/grel/ast/OperatorCallExprTest.java:[15,13] cannot access com.google.refine.grel.ast.OperatorCallExpr
[ERROR]   bad class file: F:\GitHubRepos\OpenRefine\modules\grel\target\classes\com\google\refine\grel\ast\OperatorCallExpr.class
[ERROR]     unable to access file: java.nio.file.NoSuchFileException: F:\GitHubRepos\OpenRefine\modules\grel\target\classes\com\google\refine\grel\ast\OperatorCallExpr.class
[ERROR]     Please remove or make sure it appears in the correct subdirectory of the classpath.
[ERROR] /F:/GitHubRepos/OpenRefine/modules/grel/src/test/java/com/google/refine/grel/GrelTestBase.java:[29,59] cannot access com.google.refine.grel.Parser
[ERROR]   bad class file: F:\GitHubRepos\OpenRefine\modules\grel\target\classes\com\google\refine\grel\Parser.class
[ERROR]     unable to access file: java.nio.file.NoSuchFileException: F:\GitHubRepos\OpenRefine\modules\grel\target\classes\com\google\refine\grel\Parser.class
[ERROR]     Please remove or make sure it appears in the correct subdirectory of the classpath.
[ERROR] /F:/GitHubRepos/OpenRefine/modules/grel/src/test/java/com/google/refine/expr/functions/LengthTests.java:[49,26] cannot access com.google.refine.expr.functions.Length
[ERROR]   bad class file: F:\GitHubRepos\OpenRefine\modules\grel\target\classes\com\google\refine\expr\functions\Length.class
[ERROR]     unable to access file: java.nio.file.NoSuchFileException: F:\GitHubRepos\OpenRefine\modules\grel\target\classes\com\google\refine\expr\functions\Length.class
[ERROR]     Please remove or make sure it appears in the correct subdirectory of the classpath.
[ERROR] /F:/GitHubRepos/OpenRefine/modules/grel/src/test/java/com/google/refine/expr/functions/strings/SmartSplitTests.java:[108,29] cannot find symbol
[ERROR]   symbol:   variable ControlFunctionRegistry
[ERROR]   location: class com.google.refine.expr.functions.strings.SmartSplitTests
[ERROR] /F:/GitHubRepos/OpenRefine/modules/grel/src/test/java/com/google/refine/grel/ast/ArrayExprTest.java:[15,24] cannot access Evaluable    
[ERROR]   class file for Evaluable not found
```

thadguidry:
OK, it seems that those errors are because I was still pointing to OpenJDK 21 and not OpenJDK 11.
Re-running now...

thadguidry:
It works!

```
[INFO] Using workdir F:\Maven_Local_Repo\net\sf\launch4j\launch4j\3.50\launch4j-3.50-workdir-win32
[WARNING] <bundledJreAsFallback/> has been removed! It has no effect!
[INFO] launch4j: Compiling resources
[INFO] launch4j: Linking
[INFO] launch4j: Successfully created F:\GitHubRepos\OpenRefine\packaging\target\openrefine_with_jre.exe
```

The zip file is created successfully as well

image

and when extracted looks like:

image

After double-clicking on `openrefine.exe` I do see a small issue where the version replacement didn't occur correct in the app?

image

thadguidry:
I also don't think that the `packaging\target` folder should have the following Mac or Linux resources?

- mac_jre
- OpenRefine.app
- mac_jre.tar.gz
- openrefine-linux-3.9-SNAPSHOT.tar.gz

thadguidry:
Where's the INNO installer?  Did it get skipped somehow erroneously?

thadguidry:
> I am glad its working somehow 😊. For the version not being passed to the package i think it the problem of filtering i need to make more research on this, but as i see the version on the name is 3.9 but when opening it shows you 3.8.7. i nedd to figure it out well
> 
No, the opening shows a notification banner we normally have when a new version exists.  In this case, the ${project.version} isn't a semantic version at all like `3.8.7` and so it doesn't match and OpenRefine "thinks" there's a new version and notifies with the yellow banner.  So ignore that.

> also about macos and linux resources being built the build process suggests that the build process is configured to package artifacts for multiple platforms by default, so i feel like we need to add plugins or profiles that bundle platform-specific dependencies and artifacts. using some Maven element like .

Correct, and @wetneb has some of this described in our docs.  Hmm, our CI builds and packages with a maven command, but then I see we have the `./refine dist <version>` command he documented here: https://openrefine.org/docs/technical-reference/build-test-run#building-distributions-kits
Maybe I need to use THAT command instead of `mvn ...` ?  What is the batch file `dist` section really performing?  Can we see?

> 
> is INNO setup alternative of dmg in windows? because i made generate-dmg for macos , because it was having chmod cmd which wasn't running on windows 10

DMG is only for Mac.  On Windows it's always .exe's.
I'm afraid here too, I'd need @wetneb help to answer you, or @tfmorris

wetneb:
The discussion here is going in a lot of directions and I am pinged multiple times, but I don't understand what is needed from me. Would someone be able to summarize one clear question for me to answer?

Looking at the changes in this PR, it seems that this is introducing a new plugin to generate a `.war` archive. This seems orthogonal to fixing the packaging workflow on Windows, no?

thadguidry:
> The discussion here is going in a lot of directions and I am pinged multiple times, but I don't understand what is needed from me. Would someone be able to summarize one clear question for me to answer?
>

Could the `./refine.bat` file be modified, in theory and in your opinion, so that it might be able to build the distribution for Windows on Windows?
Would that be the best approach?  Or would simply providing the long `mvn ...` command be enough? (to me it seems enough and now builds and packages with launch4j and bundles the zip, which prior to this PR, it never did for me on Windows)

> Looking at the changes in this PR, it seems that this is introducing a new plugin to generate a `.war` archive. This seems orthogonal to fixing the packaging workflow on Windows, no?

I suggested the war plugin in the original issue - it allows property placeholder replacement for the version string.  But I'm not convinced if it's really needed or not.

thadguidry:
@kalindafab disregard the INNO setup... our CI handles that, and it isn't a requirement to fix the issue of building and packaging the .exe and bundled .zip.  The INNO Installer packaging is a special version we offer for download, but it just makes it easy to install OpenRefine... dev's just need the bundled .zip which currently happens correctly on Windows with your changes from what I see.

The only issue that remains is that of picking up the custom version passed in as a mvn parameter... `mvn '-Drevision=3.9.8' package` and the version updated in the file names, and xml as appropriate.  Because currently, the files stay being created as `-3.9-SNAPSHOT` and in the UI, it shows as `${project.version}`.  Probably because we have hardcoded `3.9-SNAPSHOT` in `main\pom.xml` and `pom.xml` and all the pom's?

@wetneb The hardcoded version in pom's doesn't seem ideal.  This should always be passed in as a mvn parameter I think and our CI can use that method to set the version for snapshots and releases?

thadguidry:
> According to documentation ./refine dist only runs on mac os x. the issue here we need to make it work also on window right?
> 

I had split the Tasks in issue #6878 to fix the packaging build on Windows, and then also ensure that on Windows as well, a dev could build the Installer version.

This PR you are working on can be focused on only the `mvn ...` packaging build.  Improving the Windows `./refine.bat` to make it also handle `dist` that would create the Windows Installer packaging (and borrow from the [steps ](https://github.com/OpenRefine/OpenRefine/blob/15b5aae98997ff887598d7752e291fdc8b5ba7f7/.github/workflows/snapshot_release.yml#L314) in our `.github\workflows\snapshot_release.yml`) can be done in another PR by whomever wants to pick that task.

> also the hardcoding of 3.9 SNAPSHOT IN pom's can be replaced by dynamic version passed in mvn parameter.
If you can make that work, then great.  I think our CI `snapshot_release.yml` might then have to be modified a bit to account for that?

wetneb:
> Could the `./refine.bat` file be modified, in theory and in your opinion, so that it might be able to build the distribution for Windows on Windows?

I think the `mvn package -P windows` command looks like a good candidate for it.

thadguidry:
@kalindafab You'll need to run the right command, first of all:
`mvn -B package -DskipTests '-Drevision=3.9.8' -P windows`

The -P is for Profile where in the packaging pom.xml you'll see the profile id for `windows`.
Within that profile is where you can customize, as well as other individual pom's, but only where necessary of course.
I think you might want to get more familiar with Maven Profiles first. https://maven.apache.org/guides/introduction/introduction-to-profiles.html

I've just added a commit that enables the auto activation of the Windows profile when on Windows. (which eliminates the need to constantly pass the `-P windows` parameter for folks using Windows like me.

You'll then need to investigate how to filter the resources (or movement/copying of those resources folders) when the profile is set to Windows.  Consult the Maven docs (or individual plugins docs) or Stackoverflow for help in this.

Currently with the command `mvn package -DskipTests -Drevision=3.9.8` I get the following error, which is something else you will need to address likely in the assembly plugin or within the Windows Maven Profile?  Do we need those files?  Maybe something can be tweaked so that the assembly plugin doesn't need to see those files and can still assemble correctly for the Windows Profile and the resources needed and not needed? :-)  Dive in, have fun!  You're doing great and learning more Maven!

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:3.7.1:single (package-distributions) on project packaging: Error reading assemblies: Error locating assembly descriptor: windows-without-jre.xml
[ERROR]
[ERROR] [1] [INFO] Searching for file location: F:\GitHubRepos\OpenRefine\packaging\windows-without-jre.xml
[ERROR]
[ERROR] [2] [INFO] File: F:\GitHubRepos\OpenRefine\packaging\windows-without-jre.xml does not exist.
[ERROR]
[ERROR] [3] [INFO] File: F:\GitHubRepos\OpenRefine\windows-without-jre.xml does not exist.
[ERROR] -> [Help 1]
```

thadguidry:
@kalindafab I've helped you a bit more with another commit and added the extra `windows-without-jre.xml` file that comments out copying the jre folder.  Some developers will want to make the Windows distribution with the embedded JRE package... some without the embedded JRE.

thadguidry:
I then get the error:
```
[ERROR] Failed to execute goal on project packaging: Could not resolve dependencies for project org.openrefine:packaging:pom:3.9.8
[ERROR] dependency: org.openrefine:server:jar:3.9.8 (provided)
```

But if I look at the effective pom (without passing the revision parameter) this might be a clue:
```
PS F:\GitHubRepos\OpenRefine\packaging> mvn help:effective-pom -P windows
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for org.openrefine:server:jar must be a valid version but is '${revision}'. @ line 622, column 16
 @
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]
[ERROR]   The project org.openrefine:packaging:${revision} (F:\GitHubRepos\OpenRefine\packaging\pom.xml) has 1 error
[ERROR]     'dependencies.dependency.version' for org.openrefine:server:jar must be a valid version but is '${revision}'. @ line 622, column 16
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
```

If I build the effective pom for revision 3.9.8 and analyze it, I can see towards the beginning that indeed the Server jar is not getting complied but is defaulting in scope to be provided already, which is not what we want I think as a default for the Windows profile (since that environment is always assumed currently to be a local build?  But that might change in the future when we want to have signed .exe like we do for Mac distribution  Hmmm...):
```
  <dependencies>
    <dependency>
      <groupId>org.openrefine</groupId>
      <artifactId>server</artifactId>
      <version>3.9.8</version>
      <scope>provided</scope>
    </dependency>
    <dependency>
      <groupId>org.jvnet.localizer</groupId>
      <artifactId>localizer</artifactId>
      <version>1.31</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
  ```

thadguidry:
OK, I've sorted out the server versioning, which was being hardcoded and really should not be. (we can handle some of what we need to do for release versions in our CI workflows instead.)

@wetneb I've restructured our profiles to make it easier for local dev (and later CI build/release), but now assembly doesn't quite work out on Windows, and is skipped, so I need to include that in the default build section, but I'm unsure why we have assembly being geared for `mac.webapp.dir` ?  What's the history or reasoning on that?  The more I know the more I can make the local packaging work seamlessly no matter if done on Windows, Mac, Linux - the ideal state we want to be in on this PR.

wetneb:
> I'm unsure why we have assembly being geared for mac.webapp.dir ?

What do you mean by "geared for"? This variable should only be relevant when generatring the MacOS DMG file, I don't think it steers anything else.

thadguidry:
@wetneb OK, thanks for confirming.

wetneb:
This PR seems to contain various independent changes, some of which have influence on the build and packaging process in general (not just on Windows), so it looks unlikely to me that we can merge it. Perhaps #6878 could be made more specific, attempting to solve a particular error that happens when running the packaging process on Windows.

thadguidry:
@kalindafab Please revert to prior to my changes `git revert 3c0f472`.

thadguidry:
@wetneb it's not ready for review, this should be changed back to a Draft PR until we're really ready.

tfmorris:
Closing as stale.

PR #7077, opened 2025-01-10
Use all 4 cores during build test for server

thadguidry:
Ref: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

GitHub Runners for Ubuntu have 4 cores.
Maven 3+ has experimental support for parallel builds https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3
I've tested locally on my Windows 11 PC with `mvn -T 8 clean package` and saw no issues, but instead a vastly faster build about 1/4 the time, since we already did a pretty good job of our multi-module project and declaring all the dependencies correctly, it's able to figure out what can run in parallel over the CPU cores I have (8 cores).

The other alternative is to use Maven Daemon (built with GraalVM and fast like Gradle) https://github.com/apache/maven-mvnd But that's much more involved and requires building and setting up a custom Runner image that has Maven Daemon installed (since the normal Ubuntu runner image [does not include it](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md), but only Maven); and then we'd need to keep it up to date; more maintenance we don't need to take on.

Fixes #{issue number here}

- 
- 
-

### DIFF
--- a/.github/workflows/pull_request_server.yml
+++ b/.github/workflows/pull_request_server.yml
@@ -3,6 +3,7 @@ name: server CI
 on:
   pull_request:
     paths:
+      - '.github/workflows/pull_request_server.yml'
       - 'benchmark/**'
       - 'conf/**'
       - 'extensions/database/src/**'
--- a/.github/workflows/pull_request_server.yml
+++ b/.github/workflows/pull_request_server.yml
@@ -92,7 +92,7 @@ jobs:
       - name: Check Java linting
         id: java_linting
         run: |
-          mvn -B compile test-compile formatter:validate impsort:check javadoc:javadoc -Ddoclint=html,syntax,accessibility,reference
+          mvn -T 4 -B compile test-compile formatter:validate impsort:check javadoc:javadoc -Ddoclint=html,syntax,accessibility,reference
 
       - name: Configure connections to databases
         id: configure_db_connections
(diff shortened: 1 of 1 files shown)


wetneb:
Isn't it curious that the CI jobs that should indeed be sped up by the change are not getting run on this PR? Perhaps just because the paths don't include the workflows themselves.

thadguidry:
@wetneb  Looking deeper.  Wait, only the E2E tests ran on this, lolol?  Indeed, they are not even needed on this PR and the workflow(s) that should be run are not getting included.  Shall I handle updating those workflows?  Or did you want to address this yourself?  Happy either way.

wetneb:
Maybe you can just add `- '.github/workflows/pull_request_server.yml'` to the list of paths at the beginning of .github/workflows/pull_request_server.yml, in this PR?

wetneb (approved):

thadguidry:
Note to self, in case I forget in the future:
```
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked as thread-safe in OpenRefine - packaging:
[WARNING]   sh.tak.appbundler:appbundle-maven-plugin:1.2.0
[WARNING]   org.codehaus.mojo:license-maven-plugin:2.5.0
[WARNING]
[WARNING] Enable debug to see precisely which goals are not marked as thread-safe.
[WARNING] *****************************************************************
```

PR #7080, opened 2025-01-13
Change release manager to Antonin

wetneb:
See https://forum.openrefine.org/t/openrefine-3-9-release-schedule-contents/1796.

Given that I am in the process of leaving the project, I would like to pass on this role to someone else. If anyone is interested in taking up this role, I am available to help them get up to speed on the process.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -57,7 +57,7 @@ The Release Manager is responsible to
 * Coordinate with the community to select which issues and pull requests are part of new releases
 * Prepare and coordinate publishing new releases
 
-The current Release Manager is [Tom Morris](https://github.com/tfmorris).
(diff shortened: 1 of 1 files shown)


Abbe98 (approved):
Is there a case against having multiple release managers?

wetneb:
I think it likely would work too. The only downside I can imagine would be a dilution of responsibilities which could delay releases, but I guess we already have the problem of not doing releases often enough. Would you be interested in being part of a release team, @Abbe98?

Abbe98:
> Would you be interested in being part of a release team, @Abbe98?

Only as a backup so to say.

wetneb:
That's helpful, it's still unclear how the developer hire will go and it will surely take a bit of time before they have the necessary overview to make new releases.

PR #7091, opened 2025-01-18
ci: Run release workflow when a draft release is published

wetneb:
The GitHub Actions workflow that creates our release artifacts was not triggered when first creating a draft release and then publishing it. This fixes it.


Abbe98 (approved):

tfmorris:
This is causing the release workflow to get triggered twice. The second build fails to upload to OSSRH due to its idempotency requirements. I've opened #7692 to restructure the release flow to give us more flexibility, so we can address the duplicate runs as part of that.

PR #7096, opened 2025-01-21
operations: Expose column dependencies in c.g.r.o.cells

wetneb:
Following up on #7056, this adds the column dependencies and diffs to all operations in the `com.google.refine.operations.cells` package.


wetneb:

PR #7097, opened 2025-01-21
operations: Expose column dependencies in c.g.r.o.recon

wetneb:
Following up on #7056, this adds the column dependencies and diffs to operations in the `com.google.refine.operations.recon` package.

The data extension operation does not expose anything yet: this will be tackled in a separate PR, as it requires adaptations to the operation metadata.


wetneb:

PR #7098, opened 2025-01-21
operations: Expose column dependencies for c.g.r.o.rows

wetneb:
Following up on #7056, this adds the column dependencies and diffs to all operations in the `com.google.refine.operations.rows` package.


wetneb:

PR #7100, opened 2025-01-22
fix: Double opening of file selection in 'Apply' dialog

wetneb:
In Chrome-based browsers, the "Browse" button in the "Apply" dialog would open the file selection dialog twice when clicked.
This was because the file selection dialog was only set up after the first click. This PR simplifies that.

Discovered in a testing session with @thadguidry.


thadguidry:
I'm building this (having to re-download all the dependencies - HD crash as I said before)... give me a few hours to test, or in my AM.

thadguidry:
Fixed!  Good Job.  Works on Chrome and Edge for me.

wetneb:
Yes, it still works fine on Firefox.

thadguidry (approved):

PR #7101, opened 2025-01-22
ui: Use 'Run operations' instead of 'Perform operations' in Apply dialog

wetneb:
In the dialog to apply a recipe, the confirmation button currently reads "Perform operations". In a testing session with @thadguidry, we realized "perform" has somewhat abstract / convoluted tones. @thadguidry suggested "Apply operations" or "Replay operations". I think "Run operations" is even simpler and more straightforward. What do you think?

image

### DIFF
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -897,7 +897,7 @@
     "core-buttons/refresh": "Refresh",
     "core-buttons/reset-all": "Reset all",
     "core-buttons/remove-all": "Remove all",
-    "core-buttons/perform-op": "Perform operations",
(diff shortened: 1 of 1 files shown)


magdmartin:
This may be a small change in the UI, but it will render many tutorials and courses outdated. I am not convinced of the benefits of making this change versus the wider impact on contributors maintaining courses.

wetneb:
As you want, I am not attached to this, but the plan is to do much bigger changes to this dialog anyway, impacting any tutorials covering this area much more.

magdmartin:
Ah! If this is part of a larger change, then no worries; go ahead with the wording change. Thanks for the clarification.

PR #7102, opened 2025-01-26
fixes #7088 : Include version number in OpenRefine Windows installer path

### LINKED ISSUE #7088: Ensure that the OpenRefine Windows Installer version is created correctly (uses INNO Setup - .iss file, etc.) and verify the installer can install OpenRefine
- After #7076 which will allow anyone to create the Windows distribution files on a Windows system:
  - Run the command `./refine.bat dist` which will compile the source files and create a Windows installer (INNO) binary (and other files) in the `/target` folder.
  - Run the installer and ensure it displays the correct version in its title bar.
  - Ensure the destination path remains as `C:\Program Files (x86)\OpenRefine`.

shwetd19:
### This PR fixes issue #7088 by:

- Adding version number to installation path
- Enabling side-by-side installation of different versions
- Improving version handling with environment variable support
- Updating shortcuts to display version information

Tested with both environment variable and fallback version scenarios.


thadguidry:
Is `MyAppVersion` an environment variable we have somewhere already that will override the default in this `.iss` file somehow?

Can you show the Installer screenshot for the Select Destination Location step in the INNO Installer?

wetneb:
I'm not sure it's a good idea to encourage installing different versions of OpenRefine at the same time. If they use the same workspace (which I guess would be the case here), then this risks corrupting data. We try to ensure that a workspace created by an old version of OpenRefine can also be opened with a newer version, but in the reverse direction it is much harder to offer similar guarantees.

shwetd19:
Hello @thadguidry  , the `MyAppVersion` in this Inno Setup script isn't directly linked to an environment variable. It's defined at the top of the script with `#define MyAppVersion "3.8-SNAPSHOT"`. however here, we can modify the script to use an environment variable instead. 
Here's how:

```
#ifdef OPENREFINE_VERSION
  #define MyAppVersion GetEnv("OPENREFINE_VERSION")
#else
  #define MyAppVersion "3.8-SNAPSHOT"
#endif
```

should I proceed with it ?

thadguidry:
I agree with @wetneb and forgot about our backward compatibility issue for the workspace.
So, instead, I think for #7088 we only need to ensure that after building with the `dist` command, that the title in the installer shows the version.  I've updated the original issue comment https://github.com/OpenRefine/OpenRefine/issues/7088#issue-2796746917

wetneb:
Is this PR still being worked on? I still don't fully understand the issue it's meant to address, but if there is anything I can help with, let me know.

wetneb:
Closing per inactivity.

PR #7115, opened 2025-02-06
Fixes #7114 make refine.bat exit on errors from external programs

### LINKED ISSUE #7114: refine.bat script does not stop if pulling dependencies with NPM fails
It was reported to me that the build script on Windows (`refine.bat`) does not stop the build early if pulling the frontend dependencies with NPM fails, for instance because the version of NPM is not fitting.

Not having a Windows machine at hand, I am not able to reproduce this, but perhaps @thadguidry is interested in having a look?

thadguidry:
Fixes #7114

- exit properly if errors during calls to npm and mvn

### DIFF
--- a/refine.bat
+++ b/refine.bat
@@ -257,7 +257,7 @@ if ""%ACTION%"" == ""build"" goto :build-setup
 goto :endif
 :build-setup
 pushd main\webapp 
-call npm install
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Thanks a lot for taking care of that!

PR #7116, opened 2025-02-06
Introduce the Recipe class to hold a list of operations

wetneb:
By introducing a `Recipe` class, we group together functionality that relates to lists of operations into a topical class.
This is a first step towards handling recipes in other contexts than `ApplyOperationsCommand`.

### DIFF
--- a/modules/core/src/main/java/com/google/refine/operations/Recipe.java
+++ b/modules/core/src/main/java/com/google/refine/operations/Recipe.java
@@ -1,0 +1,135 @@
+
+package com.google.refine.operations;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonUnwrapped;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+import com.google.refine.model.AbstractOperation;
+import com.google.refine.model.ColumnsDiff;
+
+/**
+ * A list of operations to be applied in the specified order.
+ */
+public class Recipe {
+
+    private final List<AbstractOperation> operations;
+
+    @JsonCreator
+    public Recipe(
+            @JsonUnwrapped List<AbstractOperation> operations) {
+        this.operations = operations;
+    }
+
+    @JsonValue
+    public List<AbstractOperation> getOperations() {
+        return operations;
+    }
+
+    /**
+     * Checks that all operations in this recipe are valid (non null, known operations with valid parameters).
+     */
+    public void validate() {
+        for (AbstractOperation op : operations) {
+            if (op == null) {
+                throw new IllegalArgumentException("The list of operations contains 'null'");
+            }
+            if (op instanceof UnknownOperation) {
+                throw new IllegalArgumentException("Unknown operation id: " + op.getOperationId());
+            }
+            op.validate();
+        }
+    }
+
+    /**
+     * Computes which columns are required to be present in the project before applying the list of operations. The set
+     * that is returned is an under-approximation: if certain operations in the list fail to analyze their dependencies
+     * or their impact on the set of columns, then some required columns will be missed by this method, resulting in an
+     * error that will only be detected when the list of operations is applied.
+     * 
+     * @return a set of required column names
+     */
+    public Set<String> computeRequiredColumns() {
+        // columnNames represents the current set of column names in the project,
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/com/google/refine/operations/Recipe.java
```diff
@@ -0,0 +1,135 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonUnwrapped;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+import com.google.refine.model.AbstractOperation;
+import com.google.refine.model.ColumnsDiff;
+
+/**
+ * A list of operations to be applied in the specified order.
+ */
```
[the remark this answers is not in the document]
thadguidry: It's useful going forward to add `@since` and the OpenRefine version when the class or method is getting first introduced.  This helps extension authors and contributors track much more easily our evolution of API changes (no matter internal or external). 

FYI: https://www.baeldung.com/javadoc-version-since

modules/core/src/main/java/com/google/refine/operations/Recipe.java
```diff
@@ -0,0 +1,135 @@
...
+                if (columnsDiff.isPresent()) {
+                    allDependencies.addAll(columnsDiff.get().getImpliedDependencies());
+                }
+
+                for (String columnName : allDependencies) {
+                    if (!currentColumnNames.get().contains(columnName)) {
+                        if (requiredColumnNames.contains(columnName)) {
+                            // if this column has already been required before,
+                            // but is no longer part of the current columns,
+                            // that means it has been deleted since.
```
thadguidry, line 86:
```suggestion
                            // that means it has since been deleted.
```

modules/core/src/test/java/com/google/refine/operations/RecipeTests.java
```diff
@@ -0,0 +1,199 @@
...
+        Recipe recipe = ParsingUtilities.mapper.readValue(json, Recipe.class);
+
+        assertEquals(recipe.getOperations().size(), 1);
+        assertEquals(recipe.getOperations().get(0).getOperationId(), "unknown");
+
+        TestUtils.isSerializedTo(recipe, json);
+    }
+
+    @Test
+    public void testValidate() {
```
thadguidry, line 129: Just so I'm clear, this is intended for where we are testing to validate a set of operations in a Recipe? Looks like it.
Hmm, maybe we can rename it to `testValidateRecipeOperations` for clarity down the line?

modules/core/src/test/java/com/google/refine/operations/RecipeTests.java
```diff
@@ -0,0 +1,199 @@
...
+            return Optional.of(Set.of(columnName));
+        }
+
+        @Override
+        public Optional<ColumnsDiff> getColumnsDiff() {
+            return Optional.of(ColumnsDiff.modifySingleColumn(columnName));
+        }
+    }
+
+    class OpaqueOperation extends AbstractOperation {
```
thadguidry, line 102: We should probably briefly explain what an OpaqueOperation is in a comment here.  I already forgot what you said on our meeting.  Even better, if there's already a comment somewhere about this, we can just use a link tag within a comment also `* Please see {@link com.google.refine.operations.OpaqueOperation}` or whatever class better describes what an OpaqueOperation is.

thadguidry (changes requested):

modules/core/src/main/java/com/google/refine/operations/Recipe.java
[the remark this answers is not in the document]
wetneb: Sure, why not! I wouldn't backport this to 3.9, so it should rather be:
```suggestion
 * @since 3.10
 */
```

modules/core/src/test/java/com/google/refine/operations/RecipeTests.java
wetneb: Maybe just `testValidateMethod` to make it clear that we're testing `Recipe::validate()`.

modules/core/src/test/java/com/google/refine/operations/RecipeTests.java
thadguidry: Ah, then if it's only for this test class, then don't bother.

thadguidry: But much better now with these documented! Thanks it helps!

thadguidry (approved):

PR #7121, opened 2025-02-09
Update column-header-ui.js

DevangJagdale:
closes #6934

Fixes #{6934}

Changes proposed in this pull request:
- Highlight the border of a column when we want to resize it. It highlights only on mouse click and existing behavior is unchanged

Sample output:
image

PR #7129, opened 2025-02-13
Highlight column boundary while resizing

### LINKED ISSUE #6934: When clicking mouse down on a columns edge, change the whole column edge style to a thicker blue line
The only visual clue that you are dragging a columns' edge is that of a smaller drag handle cursor style.  It would be better to enhance the column's edge visually to allow users to much more quickly spot and see that they grabbed a column's edge.  The problem is exacerbated especially when there is not much data or no data in a cell.

### Proposed solution
The main purpose is to indicate to the user that they have a bigger indication that they have a good handle on a columns' edge.
It would be visually appealing and easier to see a columns edge whiling resizing it, if the entire column edge line could be highlighted with a somewhat different bolder style (a thicker blue line), like other apps do?  I would only want the style to change to a thicker blue line, once the mousedown event was fired (or rather when the dragging event started).  

### Alternatives considered
No enhancement and keep the thin blue column edge style while dragging and the only indiction of successful grab of a column edge with the small black cursor drag handle, but this makes it harder to quickly see that you are dragging a column's edge, especially depending on if there is text in the cell, or just an empty cell.

DevangJagdale:
Fixes #6934

- (https://github.com/OpenRefine/OpenRefine/pull/7121/)


wetneb:
Thank you! The solution you have come up with seems rather involved, since you're creating a dedicated element to display the column border.

Have you considered using CSS to highlight the column border? After a small investigation, it seems that setting `border-right: 4px solid blue` (for instance) on the `<col>` element representing the column being resized would work. It does require switching the overall table from `border-collapse: separate` to `border-collapse: collapse`, which seems to be causing some issues on Firefox, but works well in Chromium. Perhaps that's worth investigating? If it could be made to work on Firefox, it could potentially lead to a much simpler solution to this problem. If you're interested in exploring this path, you could break it down into two steps: first, one PR to switch to `border-collapse: collapse` while retaining the visual aspect on Firefox, and then another one to do the highlighting during resizing.

DevangJagdale:
> Thank you! The solution you have come up with seems rather involved, since you're creating a dedicated element to display the column border.
> 
> Have you considered using CSS to highlight the column border? After a small investigation, it seems that setting `border-right: 4px solid blue` (for instance) on the `<col>` element representing the column being resized would work. It does require switching the overall table from `border-collapse: separate` to `border-collapse: collapse`, which seems to be causing some issues on Firefox, but works well in Chromium. Perhaps that's worth investigating? If it could be made to work on Firefox, it could potentially lead to a much simpler solution to this problem. If you're interested in exploring this path, you could break it down into two steps: first, one PR to switch to `border-collapse: collapse` while retaining the visual aspect on Firefox, and then another one to do the highlighting during resizing.

Tried a different approach and it seems to work on different browsers

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -92,57 +92,127 @@ DataTableView.resizingState = {
...
+    return;
+  }
+
+  var thRect = thElement[0].getBoundingClientRect();
+
+  var themeColor = columnHeader._col.css('border-color');
+
+  $('#column-resize-highlight').remove();
+
+  var highlightLine = $('<div></div>')
```
wetneb, line 136: By proposing a CSS-based solution, I was expecting that there would not be any need for creating a dedicated element to represent the column border. See the details of my proposal in my earlier reply.

wetneb (changes requested):

DevangJagdale:
I added the highlight by only css property and it works on all browser except for firefox header. Now I think this is a different issue for firefox header only because if I see closely in below screenshot for firefox the header column separator lines are not present.
Firefox:
image
safari
image

wetneb:
Thank you for giving it a try, this looks much better!
I think the lack of borders in the table headers in Firefox is a problem, especially for tables created from tree-based importers (XML, JSON) because the headers will generally get bigger and the lack of borders will make it more difficult to understand the meaning of column groups in this context. This is why I suggested to first do a separate PR to switch from `border-collapse: separate` to `border-collapse: collapse`, doing the necessary changes for it to also work on Firefox.

SoryRawyer:
@DevangJagdale - given the lack of activity on this PR, I'm going to close it out. We can always reopen if and when this becomes active again.

PR #7131, opened 2025-02-16
Introduce GetColumnDependenciesCommand

wetneb:
This introduces a new command to compute the columns required by a sequence of operations. The command also returns the names of the new columns created by the recipe.

To that end, the `Recipe` class is improved to expose the new columns created. The processing of the operations list to derive such columns is moved to the validation stage of the recipe, so that the computation is done only once for both the required and created columns.


wetneb (commented):
Merging this soon to unblock the PR which will add the history visualization.

PR #7132, opened 2025-02-16
Add support for renaming columns in operation metadata

wetneb:
This PR introduces a new method `AbstractOperation::renameColumns(Map<String, String>)`, which can be used to obtain a new version of an operation's metadata after changing the name of some of the columns it refers to. The method does not modify the operation at hand, but returns an updated copy instead. It comes with a default implementation (which returns the same operation unchanged) to ensure backwards compatibility (which fulfills the contract as the method is expected to work on a best effort basis, renaming columns where possible).

I have added a few implementations of this method in some operations to give an example of the usage. Follow-up PRs will implement it in the rest of operations.


modules/core/src/main/java/com/google/refine/model/AbstractOperation.java
```diff
@@ -120,4 +121,17 @@ public Optional<Set<String>> getColumnDependencies() {
...
+     * some column references might fail to be updated, for instance if they are embedded into expressions that cannot
+     * be fully analyzed. As a fall-back solution, the same operation can be returned.
+     * 
+     * @param newColumnNames
+     *            a map from old to new column names
+     * @return the updated operation metadata
+     */
+    public AbstractOperation renameColumns(Map<String, String> newColumnNames) {
+        return this;
+    }
```
wetneb, line 136: This is the method being introduced.

wetneb:
Merging soon to enable the follow-up PRs that will implement this renaming in all operations.

PR #7134, opened 2025-02-16
Pure CSS alternative to highlighting history entries on hover

wetneb:
This changes the way history entries are highlighted when hovering one of them. So far, this was done with event listeners to maintain an absolutely positioned div which would add a special background color on the entries between the currently selected one and the hovered one. This simplifies it by removing that Javascript logic and defining the changes of background via CSS only. This relies on the [`:has()` pseudo-class](https://developer.mozilla.org/fr/docs/Web/CSS/:has), which is a recent CSS feature, but is available in "Baseline 2023".

This is a preparation for including operation icons in the history.

The behaviour remains unchanged as far as I can tell:
screencast

Tested on Firefox and Chromium.


Abbe98 (approved):
That's a lovely solution!

PR #7135, opened 2025-02-16
Show operation icons in history panel

wetneb:
- change the GetHistoryCommand to expose operation ids
- use those operation ids to show operation icons in the undo-redo panel

This requires #7134 to get the hovering behaviour to work correctly.

I'm not sure about the vertical alignment of the icons compared to the other descriptions. Currently they are centered, but perhaps it's better if their top boundaries are aligned.

Another question is whether step indices ("1. ", "2. "…) are really worth retaining. In my opinion they aren't very useful, I remember Zoe also proposed to remove them.

Another question is whether there should be an icon placeholder for steps which are either not associated to an operation (such as edits of single cells) or whose operation doesn't declare an icon. For now, the space is left blank.

### Screenshot

image

### DIFF
--- a/main/tests/server/src/com/google/refine/commands/history/GetHistoryCommandTests.java
+++ b/main/tests/server/src/com/google/refine/commands/history/GetHistoryCommandTests.java
@@ -1,0 +1,66 @@
+
+package com.google.refine.commands.history;
+
+import static org.mockito.Mockito.when;
+import static org.testng.Assert.assertEquals;
+
+import java.io.IOException;
+import java.io.Serializable;
+
+import javax.servlet.ServletException;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import com.google.refine.browsing.EngineConfig;
+import com.google.refine.commands.CommandTestBase;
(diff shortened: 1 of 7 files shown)


Abbe98:
> I'm not sure about the vertical alignment of the icons compared to the other descriptions. Currently they are centered, but perhaps it's better if their top boundaries are aligned.

I think it might become easier to read if they were top aligned, although you might need to play with the icon size as well.

> Another question is whether step indices ("1. ", "2. "…) are really worth retaining. In my opinion they aren't very useful, I remember Zoe also proposed to remove them.

I think these are needed as it's not otherwise obvious how the list is ordered until you start reading(especially if someone is used to other interfaces). It might be possible to use [counter()](https://developer.mozilla.org/en-US/docs/Web/CSS/counter) to render the numbers next to the icons, to avoid the current disconnect.

wetneb:
> It might be possible to use [counter()](https://developer.mozilla.org/en-US/docs/Web/CSS/counter) to render the numbers next to the icons, to avoid the current disconnect.

I'm not sure what you mean by "current disconnect", what sort of layout would you aim for?

wetneb:
I've switched to aligning the icon and text with `baseline` mode, which I think looks indeed better:
screenshot

Issue #7142, opened 2025-02-18
OpenRefine 3.9 will not launch on macOS when installed for all users of the machine (root:admin)


mikeorama:
When preparing OpenRefine for use in a managed classroom environment, where multiple students will log into the computer using their own accounts, and the OpenRefine.app itself is marked as owned by root:admin, the app will not launch when logged in as the student. Setting the user:group ownership to root:admin is a common practice in environments where multiple users may log into the computer and use the applications.

Research directed me to [issue 5160](https://github.com/OpenRefine/OpenRefine/issues/5160), specifically that the following two files may not have the correct permissions:

- [ ] Contents/Resources/webapp/WEB-INF/butterfly.properties
- [ ] Contents/Resources/webapp/WEB-INF/web.xml

Indeed this two files have mode 604. Changing file mode to 644 allows any user on the system to launch the app, no matter what user:group combination is applied to the entire app bundle - for instance, student1:staff, or root:staff, or root:wheel (as in the previous report).

wetneb:
Curious that this is a problem again, as we haven't changed anything to the release process since the fix for #5160 as far as I know. Do you know if the file permissions were set correctly in 3.8.7?

mikeorama:
I checked the 3.8.7, 3.8.2, 3.6.1 and 3.6.0 releases, and in all bundles except 3.6.0, the two files in question are also mode 604. The following is directly from the DMG for 3.8.7, downloaded this morning:

```
 % ls -l /Volumes/OpenRefine/OpenRefine.app/Contents/Resources/webapp/WEB-INF 
total 40
-rw----r--   1 user  staff  1071 Nov 21 15:05 butterfly.properties
-rw-r--r--   1 user  staff  1984 Nov 21 15:05 jetty-web.xml
drwxr-xr-x  91 user  staff  3094 Nov 21 15:05 lib
-rw-r--r--   1 user  staff   196 Nov 21 15:05 modules.properties
-rw-r--r--   1 user  staff   533 Nov 21 15:05 velocity.properties
-rw----r--   1 user  staff  1754 Nov 21 15:05 web.xml
```

In the case of 3.6.0, the files are mode 600.

[...]

PR #7144, opened 2025-02-19
MacOS: update file permissions in DMG package

wetneb:
For #7142, hopefully closes it, but I haven't tried yet. I would just check after the snapshot release workflow runs.


wetneb:
I would merge this soon and add it to the `3.9` branch, for it to be released in 3.9.1

PR #7145, opened 2025-02-19
move rename and remove column one menu up

### LINKED ISSUE #6280: Place 'edit' and ‘remove’ functionalities within the same menu tier
Currently, you have to navigate to 'edit column' first in order to remove a column. I found it a bit counterintuitive to have ‘remove’ functionality as a subset of the 'edit' functionality; to me, they are two separate kinds of actions on par with one another.

In other words: I’m either looking to edit a column or to delete it. I wouldn't think to edit a column I'm intending to get rid of, so navigating first to 'edit' in this instance feels a bit unclear. I think they should appear on the same level. 

Proposed solution: I'd like to propose adding the remove column functionality - or any kind of functionality that involves deleting information - to the first menu that pops up, not the second.

Here's a screenshot for reference:
Screenshot 2024-01-04 at 10 54 13 AM

smierz:
Fixes #6280 

**Status Quo**:
Currently the operations "Rename this column" and "Remove this column" are placed within the sub-menu "Edit column".

**Changes proposed in this pull request**:
Operations  "Rename this column" and "Remove this column" are placed within the main column menu.

Before             |  After
edit-column-before | edit-column-after

**Notes**:
* left out the corresponding icons of the actions to make the look more coherent 
(pattern is main menu -> no icon, sub-menu -> icon)
* shortened the operation names: "Remove ~this~ column" / "Rename ~this~ column"
While this was straight forward for rename operation as the text is only used once, I added another translation for remove column as the translation is also used in other places

### DIFF
--- a/main/tests/cypress/cypress/e2e/project/grid/column/edit-column/rename_remove_column.cy.js
+++ b/main/tests/cypress/cypress/e2e/project/grid/column/edit-column/rename_remove_column.cy.js
@@ -2,11 +2,11 @@ describe(__filename, function () {
   it('Ensures a column is removed from the data-table', function () {
     cy.loadAndVisitProject('food.mini');
 
-    cy.columnActionClick('Shrt_Desc', ['Edit column', 'Remove this column']);
+    cy.columnActionClick('Shrt_Desc', ['Remove column']);
 
     cy.assertNotificationContainingText('Remove column Shrt_Desc');
 
-    cy.columnActionClick('Water', ['Edit column', 'Remove this column']);
+    cy.columnActionClick('Water', ['Remove column']);
 
     cy.assertNotificationContainingText('Remove column Water');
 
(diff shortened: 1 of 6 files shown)


wetneb (approved):
Thanks a lot! I am not sure about removing the icons, I think it would probably not look too odd to still have them because those would still be "actions" that can be triggered directly, in contrast to the other entries in this menu (which all require further choices / steps). But also happy to do without.

smierz:
ok, I've put the open question (icons yes/no) up for discussion in the issue thread

smierz:
* re-added the icons in the menu entries for remove and rename 
* added some width to the menu-container, so that icon+text fits into one line (currently hard-coded in js) 

Preview:
 Image

wetneb (approved):
The point about the width of this menu being hard-coded is a good one, we need to handle longer menu entries in a more graceful way (for instance because this is already happening in translations).

smierz:
@wetneb thanks :)

PR #7152, opened 2025-02-21
Update FunctionDescription.properties

### LINKED ISSUE #6138: Help for escape("url") should document encoding generated
There is currently no information included in the help or documentation about what type of encoding is done for the `escape("url")` option, so a user could easily make the mistake that it correctly encodes all pieces of a URL when, in fact, the encoding that it generates is in `application/x-www-form-urlencoded` format.

### Expected Behavior
Product help and documentation is specific about what encoding is generated.

shreyas-koduvayur:
…e("url)

Fixes #6138

- Add specification for escape("url")
- url should be in application/x-www-form-urlencoded format
- documentation specific about the errors that could occur


wetneb (commented):
I don't think this embedded function description lends itself well to such long contents. I would intuitively expect that in the manual instead (openrefine.org/docs). Besides, the formulation could likely be improved, as the first sentence you added gives the impression (to me at least) that the function expects its input values to already be correctly encoded / escaped, which is confusing.

wetneb:
I'm closing this as you don't seem to be working on this PR anymore, @shreyas-koduvayur. Feel free to reopen if you want to work on it again.

PR #7153, opened 2025-02-24
operations: Renaming support for c.g.r.o.column

wetneb:
This adds support for updating the column names referenced in operations that are classified under the `com.google.refine.operations.column` package, following up on #7132.


wetneb:
Merging this soon to unblock #7158.

PR #7158, opened 2025-02-24
Map recipe columns to project columns in 'Apply' dialog

wetneb:
This adds support for mapping the columns mentioned in a recipe to the columns present in the project, when applying a recipe with the "Apply" dialog.

The first commit introduces the functionality in the backend. It relies on the relevant operations implementing the newly introduced `AbstractOperation::renameColumns()` method introduced in #7132 (see #7153, #7154, #7155, #7156 and #7157).

The second one exposes the functionality in the frontend, adding an additional step to give the opportunity to the user to map columns.

The last one adapts the Cypress test to take this new workflow into account. <s>It currently fails because of the dependency on #7153 and #7156.</s>

### Screenshot

image


wetneb:
One point that I would like to improve on (following testing with @magdmartin) is how to highlight the fields of the form that need filling in. In the screenshot above, the first two columns are pre-mapped (because columns with those exact names already exist in the project), and only the last one needs selecting. It would be good to highlight that fact better.
One possibility would be to remove the "Select a column" placeholder, keeping the select widget blank (which would make it stand out more visually).

magdmartin:
I see different options to highlight where the user needs to take action
* highlight with a different color when the drop-down needs a selection. However, for project with many columns, that information may be hidden behind a scroll
* Create a summary of the action needed at the top of the modal. For example, on the same line as the text `Required columns`. We can indicate a count of how many column needs to be mapped. 
* An error or warning message when the user select `Run Operations` if all columns are not mapped. Then

wetneb:
I have removed the "Select a column" placeholder, to make empty fields stand out more visually.

> highlight with a different color when the drop-down needs a selection. However, for project with many columns, that information may be hidden behind a scroll

It would probably not hurt, but I struggle to decide which color to use because I don't think this is a UI pattern we use anywhere else in the tool. It's also unclear to me what the shape of the background should be (how wide should it extend around the input field? Should corners be rounded? and so on)
So I'm a bit worried that this wouldn't look very natural in the UI, but if you have ideas… But perhaps the empty fields stand out well enough now that the placeholder is gone?

> An error or warning message when the user select Run Operations if all columns are not mapped.

That's already implemented: when submitting the form, any field left empty is highlighted to the user by the browser:

image

The exact appearance depends on the browser and platform, because it relies on the `required="true"` attribute.

> Create a summary of the action needed at the top of the modal. For example, on the same line as the text Required columns. We can indicate a count of how many column needs to be mapped.

This would perhaps be more natural if we hide the pre-mapped columns, because the number of fields would match the number announced in the text, no? We can also decide to switch to that.

Issue #7140, opened 2025-02-18
Ability to size column width to accommodate column name


jquartel:
Since 3.9.0 has introduced resizable columns, I'm finding many of my columns starting out too narrow to see the column name. When doing a string search (using the browser), a hit on one of these column names doesn't reveal the whole name, so you have to hand-resize each of the relevant columns in order to find the one you're looking for.

I think at the very least, clicking on a column's menu button should reveal the full column name somehow. But I think it would also be good to have something to click (or a preference somewhere) to set the column width minimum (across the whole project) to be wide enough for the column name.

<img width="752" alt="Image" src="https://github.com/user-attachments/assets/f4fa7928-6644-4507-8e33-7dac20dc6375" />

thadguidry:
I'd say this is definitely a preference kind-of-thing.  I personally like that the column sizing is congruent with the cell values (and NOT the column name - by default).  However, folks have different preferences, so we should indeed have a way to "quickly" maximize the column sizes to match the column name lengths. In Excel, LibreCalc, and Google Sheets, you double-click on any border between column headers to expand all the columns to the max length of cells.  I'm wondering how difficult that might be for us to mimic that common behavior?  But in your case, you're asking for resizing quickly to the length of the header row (column names).  I'd also be 👍 +1 for having a preference setting for this as well, but I think a more intuitive interface to just 1-click and it resizes by the length of the column names would be ideal.

How to 1-click and where and what that looks like is certainly up for debate.  But I can see us having a toggle button perhaps:

- Toggle button:  Values | Header or just V|H with some hover text ?

button hack:

![Image](https://github.com/user-attachments/assets/b1d2b686-b20a-4bb2-9749-f88874d45b55)

jquartel:
I was thinking maybe under the All > View have a new option like "Set column width to show full names".

jquartel:
> I'm considering just reverting back to showing all column names fully by default.

I think this would be good. Currently the resizing behavior in Safari is quite broken and it's pretty awkward trying to resize a lot of columns in order to see the names.

thadguidry:
Toggle.  Which leaves it up to the user to decide, instead of us.
With a global preference (since it's easily changed for those that need this functionality - like me sometimes)
And then default the preference out-of-the-box to showing full column names.
For those that want to tweak, resize, they still can, but at least the default experience is showing full column names as before, which is what most users expect.

jquartel:
[...]

PR #7168, opened 2025-02-26
Show all column headers in full by default

### LINKED ISSUE #7167: Impossible to resize the last column if the table is wider than the screen
See https://forum.openrefine.org/t/last-column-display-is-cut-off-since-update-to-3-9/2110:

> I recently upgraded to 3.9, and since the upgrade I've noticed that the last columns in my projects are cut off. It's like this for every project. I'm guessing it's a bug, but I'm not sure where to report it? If it's not a bug and there's a fix for it, please let me know!
> 
> EDIT: I forgot to mention that I'm on a Mac (running Sonoma).
> 
> EDIT2: I had a colleague check it on their windows machine and it's doing the same thing. It only happens though when you have many columns and have to scroll over. The behavior is the same in both Chrome and Firefox.
> 
> EDIT3: If you add a new column to the end, and then delete the new column, the last column is no longer cut off. This definitely seems like a bug, but I'm glad I found a workaround until it's fixed.

wetneb:
Closes #7167.

This PR lets the name of the last column to expand to its full width. This is meant to improve the UX in two situations:
 * if the grid is narrower than the screen, then it makes little sense to
    cut off the last column name because there is free space afterwards,
    so better let it expand
 * if the grid is wider than the screen, then it is impossible to drag
    the edge of the last column to the right, since there is no space on
    the right. So better have the column expanded to show the full column
    name too.

Also, this PR adds padding to the grid to make it possible to resize the last column, which still makes sense if we want to adapt the column width to better fit the contents of the column.

I will follow up with a toggle to expand / shrink all column headers, and improvements to the resizing itself.

#### Before

image

#### After

image

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
+++ b/main/webapp/modules/core/scripts/views/data-table/data-table-view.js
@@ -105,6 +105,9 @@ DataTableView.prototype._startResizing = function(columnIndex, clickEvent) {
   // for conversion from px to em
   state.emFactor = parseFloat(getComputedStyle($(".data-table-container colgroup")[0]).fontSize);
 
+  state.col.width(state.originalWidth);
+  columnHeader._td.classList.add('resized-column');
+
   $('body')
       .on('mousemove', DataTableView.mouseMoveListener)
       .on('mouseup', DataTableView.mouseReleaseListener);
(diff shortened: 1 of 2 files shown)


wetneb:
In view of #7140, I've added another commit which brings the table layout closer to what it used to be before the resizable columns feature: all column names are shown in full by default. See the discussion in #7140.

wetneb:
I would like to merge this and release it in 3.9.1 - any concerns?

PR #7171, opened 2025-03-02
removed  jquery-ui copy from externals

### LINKED ISSUE #6909: jquery-ui: switch to fetching it via NPM instead of using a vendored version?

DevangJagdale:
closes #6909

- copy of jquery-ui is removed from this repository
- added path to necessary places


DevangJagdale:
Hi @wetneb,
I see my below e2e test case for facet failed
image


the code for above testcase is
    //sliding the right slider
    cy.get('.slider-widget-bracket').eq(1)
    .trigger('mousedown',{ force: true })
    .trigger('mousemove',-130,0,{ force: true })
    .trigger('mouseup',{ force: true });
    cy.get('.facet-range-status').contains('0 — 45');
    cy.get('#summary-bar').contains('72 matching rows');
    
 When I run cypress locally I saw in that instead of selecting 0-45 it select 0-46 this causes it to fail.
 If I change the testcase to 0-46 and next line to 75 matching row it passes that part of test as 0-46 returns 75 rows
 
image


can you suggest me how should I proceed

DevangJagdale:
> Thanks a lot for working on this, the PR looks like it's going in the right direction!
> 
> It looks like the version of jQuery we have been bundling is 1.13.1. The one you added is 1.14.1. Maybe that explains the slight difference in behaviour caught by the Cypress test.
> 
> How about trying to set the version to 1.13.1 instead? We'd hopefully be able to merge this PR as is, and then deal with the upgrade separately.

In version 1.13.1 the full set of legacy themes (including ui‑lightness) is no longer bundled with the distribution. It has lots of missing files and images. To tackle this I added "jquery-ui-dist": "1.13.1", in package.json, but still the same testcases are failing for these versions. 

(The ui-lightness theme is available in jQuery UI 1.13.2 and 1.14.1)

wetneb:
Okay! And did you try 1.13.2 as well?
If what we have is a weird mixture of versions, then we can probably try to investigate manually, and perhaps update the test to match the new behaviour, it doesn't seem like a big deal…

DevangJagdale:
I tried 1.13.2 it fails at same places.
Should I proceed with 1.13.2? if you want I can check the 3 failing testcases

DevangJagdale:
1. Test facet by choice count
 It was failing because Facet by choice counts button was not in view so added cy.get('.facet-body-inner').scrollTo('bottom'); to fix this test
 
 2. Test the clustering using custom function
 It was failing because ok button was not pressed after adding custom function. Added this line cy.get('button[type="submit"][bind="okButton"]').click(); to solve it
 
 3. Test for change button and the sliding feature in numeric facet
 Failing because below lines were mismatched
 cy.get('.facet-range-status').contains('0 — 46');
    cy.get('#summary-bar').contains('75 matching rows');
    
    
    After changing them in my local cypress test all testcase are passing in chrome, but when I change to firefox some testcase fail.
    example of testcase 1
    in chrome after I scroll to bottom I can not scroll horizontally so the Facet by choice counts button is not covered by that scroll but in firefox we get scroll option which covers it so testcase fails because the center of this element is hidden from view.
    
    Chrome:
    
image

Firefox:
image

wetneb:
Ok, I thought we were just talking about one failing test for the numeric facet. Those other failing tests look like actual issues in the application, which shouldn't be worked around by adapting the tests, but fixing the application itself, since they involve some buttons not being visible when they should, or areas becoming scrollable when they should not, and so on.

By testing it interactively, I can already spot one issue in the project page, where the "Facet / Filter" and "Undo / Redo" tabs don't display properly:
image

So it looks like there is quite some manual checking work to do here. We can probably still rely on the Cypress test suite initially, by trying to address each failing test:
* observe the difference in the UI before and after the change
* decide if this change of appearance or behaviour is acceptable as a user
* if it is acceptable, then just adapt the Cypress test
* if it isn't acceptable, then the application itself needs fixing

SoryRawyer:
Hey @DevangJagdale, I just wanted to check in and see if there was anything I could help out with here. Do you have any questions about @wetneb's comment?

SoryRawyer:
Given this has been sitting open for a while, I'm going to close this for now. We can always reopen it later if needed.

PR #7173, opened 2025-03-03
Show Undo notification when applying recipes

### LINKED ISSUE #7172: Add undo notification for Apply dialog
When the user triggers an operation from the UI, they get a notification in the UI with a summary of the changes made, and the opportunity to undo it:

Image

When applying a recipe via the "Apply" dialog, they don't get this.

### Proposed solution

Offer the same sort of notifications for recipes run via the Apply dialog.
* If the recipe is made of a single operation, then the notification should be identical to the one obtained when running that operation manually.
* If the recipe contains multiple operations, then a summary showing the number of operations applied should be shown, with an undo button that would undo them all.

Image

### Alternatives considered

We could also attempt to show the descriptions of the operations themselves, but it seems difficult to make this fit into this yellow bubble.

### Additional context

This would also let us use the `apply-operations` command as a generic one to apply single operations, instead of defining one command per operation.

wetneb:
Fixes #7172.


Abbe98 (approved):
Works like a charm.

PR #7176, opened 2025-03-03
fix: weird layout for wikitext importer

### LINKED ISSUE #6498: Weird layout for Wikitext importer
The layout of the importer options for the Wikitext format is sub-optimal.

### To Reproduce
1. First, use the "Clipboard" panel to create a project from the following text:

```
{| class="wikitable"
|+ Caption text
|-
! Header text !! Header text !! Header text
|-
| Example || Example || Example
|-
| Example || Example || Example
|-
| Example || Example || Example
|}
```

### Current Results
On a sufficiently large screen, the checkboxes for the importing options are very far away from each other:

### Expected Behavior

There shouldn't be such a wide gap between the checkboxes and their associated labels.

### Screenshots

image

### Versions
 - Operating System: Debian
 - Browser Version: Firefox 115.8.0esr
 - OpenRefine: master branch

### Additional context

We probably shouldn't use HTML tables for the layout of importing options, as using HTML tables for layout is a long deprecated practice.

SoryRawyer:
Fixes #6498

- remove the `width` attribute in the wikidata importer settings HTML to move the labels closer to the check boxes

Although the issue mentions a preference for moving away from using `table` to format the settings, I decided to leave the existing HTML as it's more consistent with the other settings HTML snippets. I think it makes more sense to track the migration of these snippets to more modern styles in a follow-up task.

I've attached a screenshot of the new settings. Given that `Disable auto preview` is below the `Update preview` button, I felt it made sense to leave the indentation for the final checkbox as-is, but I'd be happy to align the button and checkbox with the rest of the settings if others feel that is more appropriate.

Screenshot from 2025-02-11 10-54-41

### DIFF
--- a/main/webapp/modules/core/scripts/index/parser-interfaces/wikitext-parser-ui.html
+++ b/main/webapp/modules/core/scripts/index/parser-interfaces/wikitext-parser-ui.html
@@ -1,46 +1,46 @@
 <div class="grid-layout layout-tight"><table role="presentation">
       <tr><td colspan="2" id="or-import-colsep"></td></tr>
-         
-      <tr><td width="1%"><input type="checkbox" bind="wikiCheckbox" id="$reconcileWiki" /></td><td><label for="$reconcileWiki" id="or-import-wiki-base-url"></label>
+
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Makes sense, thanks a lot!

PR #7187, opened 2025-03-04
Fix infinite loading spinner for invalid or unknown project IDs

### LINKED ISSUE #7175: Infinite loading for deleted projects
### To Reproduce
1. First, create a project
2. Then, copy the URL to that project (such as http://127.0.0.1:3333/project?project=2627873220862 )
3. Then, delete that project (in the list of projects at http://127.0.0.1:3333/ )
4. Visit the project page (at http://127.0.0.1:3333/project?project=2627873220862)

### Current Results

The loading spinner is shown indefinitely.

### Expected Behavior

A message should be shown to the user, explaining that the project cannot be found.

### Screenshots

Image

### Versions

 - OpenRefine: master branch

SoryRawyer:
Fixes #7175 

- This change catches a `ServletException` when looking up a project ID and returns an error. This leverages the existing client-side error-handling mechanisms to show users an error message if the project ID doesn't map to an existing project.

Attached is a screenshot showing the error message shown when loading an invalid or unknown project ID.
Screenshot from 2025-03-04 15-19-21

### DIFF
--- a/main/src/com/google/refine/commands/project/GetModelsCommand.java
+++ b/main/src/com/google/refine/commands/project/GetModelsCommand.java
@@ -114,7 +114,12 @@ public class GetModelsCommand extends Command {
             }
         }
         if (project == null) {
-            project = getProject(request);
+            try {
+                project = getProject(request);
+            } catch (ServletException e) {
+                respondStatusError(response, e.getMessage());
+                return;
+            }
         }
 
         response.setHeader("Cache-Control", "no-cache");


Abbe98 (approved):
Thank you!

wetneb (approved):
Thanks! I would argue that the `try { } catch { }` block could be broader, to encompass essentially all of the method's body, because OpenRefine doesn't provide this generic exception-catching mechanism so far sadly (https://github.com/OpenRefine/OpenRefine/issues/7190). But it also works like that.

PR #7193, opened 2025-03-06
closes #6281: renaming "view" options for columns in menu

### LINKED ISSUE #6281: Renaming 'View' in menu
If I want to hide or unhide a column, at the moment I'd need to first navigate to 'View,' in the first menu tier, then select 'Collapse' options or 'Expand' options in the second menu tier. 

I think it would be more intuitive to rename 'View' to 'Collapse/Expand.' The term 'View' is a bit vague and could mean many things (Zoom in / Zoom out, for example). For the sake of clarity, I'd change the language to match the functionality itself.

I'd also like to propose for discussion: what if we changed the language to Hide/Unhide instead of Collapse/Expand? I think that would be even clearer for new users of the tool.

Here's a screenshot for context:
Screenshot 2024-01-04 at 11 05 47 AM

PadyA0:
Fixes #6281

- I replaced the words "collapse", "expand" to clearer ones like "hide", "show" in the menu.


wetneb:
Have you tried running OpenRefine with your changes to check that your changes had the expected effect?

wetneb (changes requested):
Thank you, that is better! This is something you can keep in mind for your next open source contributions: always check that the changes you have made have the expected effect.

However, you have decided to rename the internationalization keys. I can see two problems with that:
* the updates you have done in the `translation-en.json` file are not consistent with the ones you have done in the javascript code. For instance, you have renamed `core-views/collapse-all` to `core-views/Hide-all` in the translation file, but not in the code, which will break the corresponding menu entry. Also, the capital `H` is inconsistent with the rest of the translation strings.
* by introducing new translation keys, you are discarding the existing translations in other languages. Translators will have to translate those strings from scratch again. Given that this is only a wording change and that the meaning is preserved, I think it would be better to keep the same translation strings so that the existing translations can continue to be used.

PadyA0:
Thank you for your detailed feedback. I appreciate you pointing out the inconsistencies in my changes.
Do you suggest reverting the translation file back to original?
I will go ahead and adjust that for a start.

wetneb:
>  Do you suggest reverting the translation file back to original?

No, I suggest keeping the original translation keys unchanged, but changing their associated values in English.

wetneb:
@PadyA0 have you tried to run OpenRefine with your changes and checked that they have the expected effect?

PadyA0:
I have verified that it works as expected. I intended to make just one final commit after verifying and squashing the previous ones because I did not realize each file I edited became a new commit. Let me know if you would like me to clean up the commit history now that I figured it out.

main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
```diff
@@ -182,7 +182,7 @@ DataTableColumnHeaderUI.prototype._createMenuForColumnHeader = function(elmt) {
     {
       id: "core/view",
       label: $.i18n('core-views/view'),
-      tooltip: $.i18n('core-views/collapse-expand'),
```
wetneb, line 185: I think this one should also be kept to its previous key.

wetneb (commented):
Thanks a lot!
We normally squash all commits together when merging PRs, so there is no need to do it on your side (but you're welcome to do it if you want to).

Another tip is that if you go to the [Files changed](https://github.com/OpenRefine/OpenRefine/pull/7193/files) tab of this pull request, you'll see the overall changes you are proposing, with all your commits. It's a good idea to check it out yourself, so that you can notice any issues. This would for instance let you notice the one below.

wetneb (approved):
Thanks a lot!

main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
PadyA0: right, I put it back. Currently checking the other lines

SoryRawyer:
Thanks for working on this, @PadyA0! And thanks for the reviews, @wetneb!

Issue #2253, opened 2019-12-16
Improve history visualization to show column dependencies


wetneb:
_This is work package 4 in our [2020 roadmap](http://openrefine.org/images/czi-eoss-proposal.pdf)._

The history tab in OpenRefine lists all the operations executed on the project. Operations are listed with their "description", a short text which summarizes what the operation does.
It is generally quite hard to understand a workflow by reading this, since:
* These descriptions mix human-readable text and (abstruse) expressions, which often overflow beyond the size of the menu. These are not localized as they are generated by the backend;
* The list of operations can grow a lot: many workflows contain hundreds of operations. The current view requires a lot of scrolling in these cases.
* It is hard to get a sense of the dependencies between the operations (the data flow)

## Goals

We should implement a graphical representation of workflows, which shows the dependencies between column-wise operations.

We should first implement simple visualization, without facets. Any applied facets are considered to be extra dependencies of the operations. A proposal for such diagrams is in [this paper](https://arxiv.org/pdf/1906.05937.pdf) (page 6).

- [ ] Draw mock-ups of how such graphs could concretely be integrated in the UI
- [ ] Implement the rendering in the UI
- [ ] Tweak as corner cases arise

thadguidry:
I personally prefer the idea of a horizontal timeline.  With a fast drag bar along the top that might categorize with color or other elements, as well as allowing to quickly move forward or reverse in time.
How dependencies (or just a counts of dependencies in a balloon circle) are shown and then possibly dragged around together can be evaluated.

A few Project management tools that I have used in work utilize graphs in various ways to represent some "connected-ness"...some with or without connecting lines.
https://asana.com/resources/visual-project-management-kanban-timeline-calendar
https://support.roadmunk.com/en/articles/2760490-visualizing-your-roadmap-with-dependencies

I am not convinced that edges as "connected lines" are the best presentation for "connected-ness". Lines get messy fast with lots of Objects/Containers and provide little value in feedback to a user (big fat Chord diagrams come to mind).
It depends on what we want to tell the user about the edges, or if we simply need them to know more about an entire "dependent cluster of operations that MUST be dragged around together".

What the user needs to know about the various Objects (a dependent cluster of operations, an independent operation, etc.) should probably be written down first.
Then with that knowledge, better UI's can be prototyped.

thadguidry:
Ports are one way to show a flow. https://gojs.net/latest/samples/dataFlow.html
[...]

PR #7196, opened 2025-03-09
Add recipe visualization to Extract dialog

wetneb:
For #2253.

This introduces an experimental visualization of recipes in the Extract dialog, helping the user grasp the columnar dependencies of the operations they have selected for extraction. See the [forum thread](https://forum.openrefine.org/t/recipe-visualization-prototype/2069) on this.

I'm proposing to introduce this in the Extract dialog, because it seems to be the place where this is the most useful: when extracting a recipe, it's useful to understand how it will generalize to other projects, and get a bit of help selecting the operations as required.

The visualization is rendered as an SVG element, with some event listeners to enable tooltips when hovering.
The rendering of the SVG is done in a few steps:
* first, in `computeColumnIds()`, the recipe is scanned to list all columns involved at any point in the recipe. Constraints on the placement of those columns are also gathered in this pass: for instance, if a column is created based on another column, we require that the created column appears to the right of its base column (similarly to what happens when running the operation). At the end of this, we obtain a list of columns and constraints on their relative placement
* then, we do a topological sort to assign positions to each column
* finally, for the rendering itself, we draw all columns at their computed positions and then add the operations on top.

The positions of the columns in the visualization might differ from their original position in the project, since recipes don't provide sufficient information to recover that. In most cases, recipes are meant to generalize well regardless of column positions (apart from the first column defining the record boundaries).

image

### DIFF
--- a/main/src/com/google/refine/commands/history/GetColumnDependenciesCommand.java
+++ b/main/src/com/google/refine/commands/history/GetColumnDependenciesCommand.java
@@ -17,6 +17,7 @@ import com.google.refine.commands.Command;
 import com.google.refine.model.AbstractOperation;
 import com.google.refine.model.ColumnsDiff;
 import com.google.refine.operations.Recipe;
+import com.google.refine.operations.Recipe.RecipeValidationException;
 import com.google.refine.util.ParsingUtilities;
 
 /**
@@ -77,6 +78,8 @@ public class GetColumnDependenciesCommand extends Command {
                     .collect(Collectors.toList());
 
             respondJSON(response, result);
+        } catch (RecipeValidationException e) {
+            respondJSON(response, e);
         } catch (Exception e) {
             respondException(response, e);
         }
--- a/main/tests/server/src/com/google/refine/commands/history/GetColumnDependenciesCommandTest.java
+++ b/main/tests/server/src/com/google/refine/commands/history/GetColumnDependenciesCommandTest.java
@@ -2,6 +2,8 @@
 package com.google.refine.commands.history;
 
 import static org.mockito.Mockito.when;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
 
 import java.io.IOException;
 import java.util.List;
@@ -9,6 +11,7 @@ import java.util.Optional;
 
 import javax.servlet.ServletException;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeMethod;
@@ -22,6 +25,8 @@ import com.google.refine.grel.Parser;
 import com.google.refine.operations.OperationDescription;
 import com.google.refine.operations.OperationRegistry;
 import com.google.refine.operations.cell.MassEditOperation;
+import com.google.refine.operations.column.ColumnAdditionOperation;
+import com.google.refine.util.ParsingUtilities;
 import com.google.refine.util.TestUtils;
 
 public class GetColumnDependenciesCommandTest extends CommandTestBase {
@@ -49,10 +54,32 @@ public class GetColumnDependenciesCommandTest extends CommandTestBase {
             + "  }\n"
             + "]";
 
+    // recipe which creates the same column twice, which is inconsistent
+    String inconsistentRecipe = "["
+            + "{"
+            + "  \"op\":\"core/column-addition\","
+            + "  \"engineConfig\":{\"facets\":[],\"mode\":\"row-based\"},"
+            + "  \"baseColumnName\":\"a\","
+            + "  \"expression\":\"grel:value\","
+            + "  \"onError\":\"set-to-blank\","
+            + "  \"newColumnName\":\"c\","
+            + "  \"columnInsertIndex\":1"
+            + "},{"
+            + "  \"op\":\"core/column-addition\","
+            + "  \"engineConfig\":{\"facets\":[],\"mode\":\"row-based\"},"
+            + "  \"baseColumnName\":\"a\","
+            + "  \"expression\":\"grel:value\","
+            + "  \"onError\":\"set-to-blank\","
+            + "  \"newColumnName\":\"c\","
+            + "  \"columnInsertIndex\":1"
+            + "}"
+            + "]";
+
     @BeforeMethod
     public void setUpDependencies() {
         command = new GetColumnDependenciesCommand();
         OperationRegistry.registerOperation(getCoreModule(), "mass-edit", MassEditOperation.class);
+        OperationRegistry.registerOperation(getCoreModule(), "column-addition", ColumnAdditionOperation.class);
     }
 
     @BeforeMethod
@@ -110,6 +137,20 @@ public class GetColumnDependenciesCommandTest extends CommandTestBase {
                 + "      } ]\n"
                 + "}";
         TestUtils.assertEqualsAsJson(response, json);
+    }
+
+    @Test
+    public void testInconsistentRecipe() throws Exception {
+        when(request.getParameter("csrf_token")).thenReturn(Command.csrfFactory.getFreshToken());
+        when(request.getParameter("operations")).thenReturn(inconsistentRecipe);
+
+        command.doPost(request, response);
+
+        String response = writer.toString();
+        JsonNode node = ParsingUtilities.mapper.readValue(response, JsonNode.class);
+        assertEquals(node.get("code").toString(), "\"error\"");
+        assertEquals(node.get("operationIndex").asInt(), 1);
+        assertTrue(node.get("message").asText().startsWith("Operation #2: Creation of column 'c' conflicts"));
     }
 
     @Test
--- a/main/webapp/modules/core/MOD-INF/controller.js
+++ b/main/webapp/modules/core/MOD-INF/controller.js
@@ -504,6 +504,7 @@ function init() {
 
       "scripts/widgets/histogram-widget.js",
       "scripts/widgets/slider-widget.js",
+      "scripts/widgets/recipe-widget.js",
 
       "scripts/project/browsing-engine.js",
       "scripts/project/history-panel.js",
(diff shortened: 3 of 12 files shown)


thadguidry:
This seems to be missing the column name above it labeled as `clusters`.

1. I'd also say that "Transpose cells across columns into rows" does 2 things, it removes a column(s), and it creates 1 or 2 new columns, depending on options chosen.  So it feels like it should have a + Plus circle for the 1 or 2 new columns made?

2. I created a simple project example, and used transpose across several columns to make 1 column named `clusters`.
Here, I only see the transpose icon in the visualization, but no label or a + Plus circle, as you have labels above your columns named `researcher` and `orchid`?

3. My other concern is that of supporting very long descriptive column names (diagonal labels would not be my first choice I'd pick - I'd prefer an ellipsis suffix to represent "more text" that allows hover over the column label to show full column name created in the tooltip).

Project file: [clusters.openrefine.tar.gz](https://github.com/user-attachments/files/19200400/clusters.openrefine.tar.gz)

image

thadguidry:
OH, I guess since I didn't make this obvious.  I'd vote to merge this and get feedback on Snapshots and make tweaks later on.  My points above are just that...feedback early :-)

wetneb:
Thanks both for the feedback! Let me try to follow up on all points you brought up.

> 1. I'd also say that "Transpose cells across columns into rows" does 2 things, it removes a column(s), and it creates 1 or 2 new columns, depending on options chosen. So it feels like it should have a + Plus circle for the 1 or 2 new columns made? or some merge-like viz showing the 8 columns merged into 1 named "clusters"?

This operation is currently treated as "opaque", but you're right that there is some potential to make it impacts on the columns clearer. I'm opening a specific issue about that: #7205.

> 2. I created a simple project example, and used transpose across several columns to make 1 column named clusters.
Here, I only see the transpose icon in the visualization, but no label or a + Plus circle, as you have labels above your columns named researcher and orchid?

Yes, because this operation is currently treated as opaque, it's rendered as a "box" (think "black box"). The rendering for a recipe with a single opaque operation in it isn't informative at all of course. But perhaps it could be improved, for instance by increasing the minimum width of such boxes, or making it clearer in some other way that it's "a black box".

> 3. My other concern is that of supporting very long descriptive column names (diagonal labels would not be my first choice I'd pick - I'd prefer an ellipsis suffix to represent "more text" that allows hover over the column label to show full column name created in the tooltip). But then again, it's not too bad and you do shift the viz pane to account for lengthy column names. Maybe this is a non-issue. Let's skip this and get feedback later and we can revisit this point.

Yes, the oblique isn't very convenient or pretty. I also considered showing the column labels that you currently get on hover by default, with a vertical offset to make sure they don't overlap:
image

It feels a bit more difficult to implement but perhaps it would be more readable?
About ellipsis, that's quite easy to introduce, but I do think that seeing full column names is quite important (especially given the feedback we got on 3.9.0).

> 4. Reorder rows permanently, isn't that really just a Sort permanently? The R icon made me chuckle.

That's just because there was no icon associated to the operation yet, so the code just picks the first letter of the description as make-shift icon. I've just merged #7159 which defined the logo for that operation and rebased this branch on top, so that the new icon appears.

And then @Abbe98's comments:

> There is a lot of great ideas in here but I think the issue would be better addressed by actually integrating this work within the operations list itself. There is no reason for it to be a simple list with checkboxes, the descriptions and code could be made distinct form each other, it could display icons, highlight dependencies and even integrate a dependency graph if we give more space.

I've been tempted by that actually. The two reasons why I went for this version for now are:
* while displaying the visualization of all operations in the project history can be useful, it seems more useful to me to visualize the selected ones only. By selecting all the operations (the default), the user is still able to get the visualization of the entire history. When selecting only the operations they want to retain, the visualization then acts as a preview of the recipe to be re-run elsewhere, which is useful to answer questions such as: what columns does it depend on? Does it contain any opaque operation, potentially leading to weird behaviours on grids of different shapes? Is there any obvious simplification of the recipe that one can notice visually? And so on. To me, it seems harder to convey those things visually if the entire history is shown, with checkboxes integrated in them. But one could also imagine some sort of highlighting to make the selected operations and the columns they depend on stand out from the rest. Does that sound like it would go in the right direction
* trying to err on the side of caution not to disturb people's existing workflows, by offering this new feature as something optional that people can switch to using the tabs.

By "dependency graph", do you mean the same sort of visualization or do you have something different in mind? It's likely that we can make very different sorts of visualizations, still relying on the same columnar metadata, and it can well be that those would be more useful or intuitive. If you have something else in mind, it would be super useful if roughly sketch what it would look like on some examples (without it needing to look pretty at all).

thadguidry:
### Column Name Labels
> It feels a bit more difficult to implement but perhaps it would be more readable?
About ellipsis, that's quite easy to introduce, but I do think that seeing full column names is quite important (especially given the feedback we got on 3.9.0).

OK, then 👍 for showing full column names.  And I do like your design for the non-overlapping column labels, so let's go with that design, it works much better.  As long as you can make the viz scroll around decently enough to show full column names, then that design suffices well to cover both cases.

### Row Reorder (Sort)
> That's just because there was no icon associated to the operation yet, so the code just picks the first letter of the description as make-shift icon. I've just merged https://github.com/OpenRefine/OpenRefine/pull/7159 which defined the logo for that operation and rebased this branch on top, so that the new icon appears.

OK, those column and row reorder icons look good enough in #7159 good job on those @wetneb !

magdmartin:
I like the concept @wetneb shared, and I support releasing it first under a "Visualize" tab as a beta feature. This integration will not disrupt users' current workflow. Instead, it will allow us to reach more users and explore how to visualize different types of projects, which will help us gather more feedback.

> By "dependency graph", do you mean the same sort of visualization or do you have something different in mind? It's likely that we can make very different sorts of visualizations, still relying on the same columnar metadata, and it can well be that those would be more useful or intuitive. If you have something else in mind, it would be super useful if roughly sketch what it would look like on some examples (without it needing to look pretty at all).

The idea of a dependency graph reminds me of a Gantt chart, where you can see the relationships between tasks. In the case of OpenRefine, this could involve visualizing columns and operations. Here are a couple of ideas:
- **From a specific operation, allow filtering to show only the predecessor and successor operations** to visually group operations that depend on each other. That can be done using a contextual menu in the [Operation History](https://forum.openrefine.org/t/four-operation-history-mockups/1362) that will open the visualization modal filtering only on the selected operations. 
- **From a column filter on operations that rely on that column or operations that modify that column**. That can be done by a new option in the column menu that will open the visualization modal filtering only on the selected operations. 

In such visualizations, I think it is important to take into account when a column is used in a facet or filter when performing an operation.

wetneb:
I've changed the column labels to be horizontal, as sketched above:
image

@magdmartin yes, linking this visualization to the contextual actions in the operation history designed earlier seems like a good move, as it would help the user understand why re-running or deleting an operation in the history affects further ones or not. I haven't worked on the implementation of this contextual menu, primarily by lack of time but also because I don't think it can be made very useful without the further architectural changes that would allow us to avoid re-running all future operations. At the moment, if we delete an operation deep in the history, we could be able to identify operations which should be removed, but all the operations which can be kept would need re-running, even though their results aren't affected by the removal.

SoryRawyer (approved):
I think this looks good, and while I think there's a bit too much code for me to review at once it seems like releasing this as part of the 3.10 beta would be a good opportunity to try it out and get feedback.

Since this has been open for a while, I'll approve but I'd like to wait to merge until Monday to allow for other people to provide any final feedback.

Abbe98:
Actually given the discussions here it might even be a good opportunity to add an official extension point here so that different extensions could provide different visualizations some might even want to infer external dependencies form recipes and visualize that.

wetneb:
Thanks for having a look again.

I'm not sure this is a great fit for an extension point, since visualizing operations in a significantly different way would likely require having the operations expose different metadata. Given that the operations themselves are extensible and that their metadata is shaped by the `AbstractOperation` interface defined in the core, it's not possible for an extension to have operations expose more than they currently do.

If you are thinking more about different presentations of the same metadata, then I'm also keen to hear more about them and adapt this implementation to go in that direction.

Abbe98:
I'm primarily thinking of an extension point to create new tabs*. Which metadata to visualize would be entirely up extension developers. I could imagine extensions extending the metadata by inspecting the GREL code or by looking at external calls being made, etc.

\* similar to what we have talked about doing for the main project area.

tfmorris:
> given that this is still under discussion, does it make more sense to release a 3.10 beta without this or wait another week or so to release a beta?

I guess I don't understand what would need to happen to make this available to users in 3.10 (or some other release). On the one hand, I share @Abbe98 's reservation about committing to such a big chunk of functionality without any real user feedback, but on the other I don't know how you go about collecting such feedback. Do we have a set of power users committed to running it through its paces during a beta test?

> If you are thinking more about different presentations of the same metadata, then I'm also keen to hear more about them and adapt this implementation to go in that direction.

I've pitched YAML a couple of times as a more user friendly, editable, alternative to JSON, but haven't got much traction with the idea.

wetneb:
I'm happy to close this PR as there just doesn't seem to be interest in the feature in the first place. I wish it could have been made clearer earlier, but that's life. I certainly wouldn't want to have it stand in the way of 3.10.

Abbe98:
> I share @Abbe98 's reservation about committing to such a big chunk of functionality without any real user feedback

My reservation is enforced given @wetneb commitment to leave the project, not only will this not get real user feedback the code will be left without a maintainer being invested in this. At the same time I do think that this would be perfect as an extension and I would be happy to help ironing out an extension point.

SoryRawyer:
We (@wetneb, myself, and the others at the Quebec City meetup) talked about this briefly during the open office hour call yesterday and I mentioned that I'd be happy to work on moving this functionality to an extension (and enabling extensions such as this if it's not clear how to do so currently). If anyone wants to collaborate on this, let me know!

Abbe98:
> If anyone wants to collaborate on this, let me know!

I'm happy to review given my current focus on extension points, we also have a very similar need for the main project area (#5335).

PR #7211, opened 2025-03-18
Send absolute row index in request when "last" pagination button is clicked

### LINKED ISSUE #7207: Incorrect behaviour of row view navigation buttons in facet-filtered view
I'm very glad that #7188 is fixed, but I'm seeing (in 3.9.1) some new incorrect behaviour of the "last>>" button when my table is filtered by a facet choice.

I have a dataset with 3674 rows and a facet which selects 96 rows. The first page (first 10 rows) has indices from 402 to 1035, though the row range indicator shows [1 - 1035]. I'm not too bothered about this since the row range indicator can't make much sense when the table is filtered. But now if I click "last>>" I get an empty table with row range indicator showing as [96 - 96]. Moreover, the "<<first" and "<previous" buttons are still disabled (they shouldn't be) and the "next>" and "last>>" buttons are not disabled (they should be). If I now click "last>>", nothing happens, and if I click "next>" I actually get the first page (i.e. that page with indices from 402 to 1035) though now "first>" is enabled (though it does nothing) as is "<<previous" (which takes me back to the empty [96 - 96] page).

In the same dataset I have another facet which has filtered 77 rows. Initially it comes up showing rows 56 to 65. Now if I click "next>" I get a page with row indeces 66 to 997 (which is 66-69 contiguous and 992-997 contiguous), which is correct. If, instead of next, I click "last>>" then I get a page showing rows 60 to 69, which is not even one page further ahead (since it includes 5 rows from the first page).

I am using Edge 134.0.3124.62 on MacOS 15.3.2 with OpenRefine 3.9.1.

SoryRawyer:
Fixes #7207

- This change addresses how the "last" pagination navigation button sends the ending point. The current behavior is that it sends the index of the last row relative to a faceted search, and this change is to send the index of the absolute last row in the project.
- This also includes a change to omit the `nextPageStart` variable if there are fewer than `limit` elements in the response. This will cause the `next` and `last` pagination buttons to be marked as inactive, as there are no more elements to paginate through.

### DIFF
--- a/main/src/com/google/refine/commands/row/GetRowsCommand.java
+++ b/main/src/com/google/refine/commands/row/GetRowsCommand.java
@@ -293,7 +293,7 @@ public class GetRowsCommand extends Command {
             JsonResult result = new JsonResult(engine.getMode(),
                     rwv.results, rwv.total,
                     engine.getMode() == Mode.RowBased ? project.rows.size() : project.recordModel.getRecordCount(),
-                    rwv.totalRows, start, end, limit, pool, previousPageEnd, nextPageStart);
+                    project.rows.size(), start, end, limit, pool, previousPageEnd, nextPageStart);
 
             respondJSON(response, result);
         } catch (IllegalJsonpException e2) {
--- a/main/src/com/google/refine/commands/row/GetRowsCommand.java
+++ b/main/src/com/google/refine/commands/row/GetRowsCommand.java
@@ -277,7 +277,7 @@ public class GetRowsCommand extends Command {
                 if (start > 0) {
                     previousPageEnd = start;
                 }
-                if (!wrappedRows.isEmpty()) {
+                if (!wrappedRows.isEmpty() && ((wrappedRows.size() >= limit))) {
                     nextPageStart = wrappedRows.get(wrappedRows.size() - 1).paginationIndex + 1;
                 }
             } else {
--- a/main/tests/server/src/com/google/refine/commands/row/GetRowsCommandTest.java
+++ b/main/tests/server/src/com/google/refine/commands/row/GetRowsCommandTest.java
@@ -44,6 +44,8 @@ import org.testng.annotations.Test;
 
 import com.google.refine.RefineTest;
 import com.google.refine.commands.Command;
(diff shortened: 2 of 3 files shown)


wetneb (changes requested):

I agree that clicking `Last` should use the total number of rows in the project, regardless of the filtering. Unfortunately, `rowModel.total` isn't quite that: in records mode, this will be the total number of records in the project. And when paginating through records, we still use the row ids to specify page boundaries, so we'd need to still use the total number of rows in the project in this case.

I can't see anywhere else where `rowModel.totalRows` is being used, so I think it would make sense to instead fix the backend to make sure this still contains the total number of rows in the project (regardless of the rows/record mode and of filtering options).

Also, it would be nice to have some automated tests covering your changes. Java unit tests are probably sufficient and faster.

SoryRawyer:
Ah I wasn't sure how this would affect records mode, thank you for clarifying! However, I can't find where `total` and `totalRows` are actually incremented separately. I can see in the response that they end up having different values, but in the code I just see them incremented one right after the other. My first impression was to move the increment logic for `totalRows` to be right above [`matchRow` in `ConjunctiveFilteredRows`](https://github.com/OpenRefine/OpenRefine/blob/master/modules/core/src/main/java/com/google/refine/browsing/util/ConjunctiveFilteredRows.java#L65) for row-oriented counts and above [`matchRecord` in `ConjunctiveFilteredRecords`](https://github.com/OpenRefine/OpenRefine/blob/master/modules/core/src/main/java/com/google/refine/browsing/util/ConjunctiveFilteredRecords.java#L64) for record-oriented counts. Does that make sense, or am I missing some other logic here?

In the meantime, I'll work on adding tests to check the response when a facet is passed to the command.

SoryRawyer:
Ah, I wasn't sure how this would affect records mode, thank you for clarifying that! I'm trying to add a test to make sure faceted search paginates correctly, but I'm having a hard time actually adding a facet within the test. I was thinking it could be possible if I copy some of the code from `TextSearchFacetTests`, but do you think there's a more convenient way to add a facet to a project in this case?

wetneb (approved):
Thanks a lot!

PR #7216, opened 2025-03-20
Use aria-labelledby over aria-labeledby

keithamus:
This PR changes `aria-labeledby` to `aria-labelledby`.

I'm making this change because the `aria-labeledby` (with a single `l` in the middle) is the alternate spelling to the correct `aria-labelledby` (with two `l`s in the middle). The two `l`s version is preferred over the single `l` version. Some browsers will only respect the version with two `l`s. 

I discovered this repository by performing a code search for uses of `aria-labeledby` and sending PRs to replace them. (I'm not a bot but someone who is working on fixing this issue). For more on this you could read https://github.com/w3c/aria/issues/2093.


Abbe98 (approved):
Thank you for taking the time to catch this!

Issue #3012, opened 2020-07-29
Python expression preview doesn't always report errors


tfmorris:
### To Reproduce
1. Select "Add Column based on this Column"
2. With the expression `cells.a.value.toNumber() - cells.b.value.toNumber()`, switch the expression language to Python

### Current Results

No error message is displayed, presumably because the response has an "ok" status:

```json
{"code":"ok","results":[{"message":"Traceback (most recent call last):\n  File \"<string>\", line 2, in __temp_594255022__\nAttributeError: 'unicode' object has no attribute 'toNumber'\n"}]}
```
### Expected Behavior

The error message should be displayed to the user

tfmorris:
I don't think this is the correct fix. Errors are normally reported in red in the top pane of the dialog (where both screenshots says "No syntax error" in the related PR #4991). This change removes the line number (which is useful information for helping to figure out where in the Python program the error is), but the returned string is still being considered a value for the preview, not an error.

I haven't looked at it since, but when I wrote up the bug report, I clearly thought that the problem was with the `"code": "ok"` part of the response. That could be incorrect, but it's definitely worth looking at.

thadguidry:
Could we also add a Cypress test for it?

yeungven:
Hi! I'm a student and my partner @hniswand and I would love to be assigned to this! We are tasked with contributing to an open source project for a class of ours.

Sudo-Aryan:
can I try with this?

antoine2711:
@Sudo-Aryan : I think this issue is actually dealt with already. See the corresponding PR.

I will close this.

Regards, Antoine

wetneb:
@antoine2711 see the discussion above: @tfmorris mentioned that the PR did something different and that this problem is still present.

antoine2711:
> @antoine2711 see the discussion above: @tfmorris mentioned that the PR did something different and that this problem is still present.

Oh. Sorry. Didn't see that.

Regard, A.

XxBrezxX:
Hey, I wanna work in this issue? Is it still available? Can you give me more details?

wetneb:
Hi @DiamondMyK, you have already been assigned #5440. Let's focus on that one and then once you have solved that one, we can assign you other issues. If you are no longer working on #5440, please unassign yourself from it first.

SymbolofMoon:
Hi @wetneb and @tfmorris. May I know if the issue is still there?
If yes can you clear me some points?
[...]

PR #7217, opened 2025-03-21
Identify syntax errors in Python expressions

SoryRawyer:
This addresses #3012, though I'm not sure if I can say it entirely fixes it.

- This change will catch a `PySyntaxError` when initializing a new `JythonEvaluable` object and will return the string "Syntax error" instead of the current "Internal error". It is a small change but I feel it is at least somewhat more helpful in guiding the user towards resolving the issue.

This is the existing error message:
Screenshot from 2025-03-21 10-24-13

And this is the error message after this change:
Screenshot from 2025-03-21 10-22-41

I realize this is a relatively small change, so please feel free to reject if you think it's unhelpful, but I feel it at least distinguishes between an internal error specific to OpenRefine and an error that a user could potentially resolve on their own.

### DIFF
--- a/extensions/jython/src/com/google/refine/jython/JythonEvaluable.java
+++ b/extensions/jython/src/com/google/refine/jython/JythonEvaluable.java
@@ -50,6 +50,7 @@ import org.python.core.PyLong;
 import org.python.core.PyNone;
 import org.python.core.PyObject;
 import org.python.core.PyString;
+import org.python.core.PySyntaxError;
 import org.python.util.PythonInterpreter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


wetneb (commented):
Very nice! I wonder if PySyntaxError contains any more information that could be worth reporting to the user, to make the message more specific than "Syntax error"?
Anyway, this is already a clear improvement.

SoryRawyer:
I looked into that, but I think the extra wrapper code we add to evaluate the expression ends up making the error message more confusing than helpful. I'm not sure it'd be possible to work backwards from the error message to provide more specific guidance.

wetneb (approved):

extensions/jython/src/com/google/refine/jython/JythonEvaluable.java
```diff
@@ -69,7 +70,11 @@ static public LanguageSpecificParser createParser() {
 
             @Override
             public Evaluable parse(String source, String languagePrefix) throws ParsingException {
-                return new JythonEvaluable(source, languagePrefix);
+                try {
+                    return new JythonEvaluable(source, languagePrefix);
+                } catch (PySyntaxError e) {
+                    throw new ParsingException("Syntax error");
```
tfmorris, line 76:
```suggestion
                    throw new ParsingException("Syntax error", e);
```

Thanks for tackling this! I know this has already been merged, but just wanted to note a suggestion to return original exception wrapped by ours, because it has useful information like the line and column number, the more specific Python parser message, etc which will be useful to the user in debugging their error.

Issue #7206, opened 2025-03-14
Customisation of max cell size for importing CSV file


tirthadg123:

docuracy:
**+1** I have the same problem using TsvParser.

magdmartin:
I’ve tested with this with 3.10-beta1 and I can confirm the new limit works as expected.

Using the following commands on Ubuntu:

* A cell with **32,769 chars** imports fine
* A cell with **150,000 chars** imports fine
* A cell with **~10.5M chars** fails, as expected

```bash
printf "id,bigcell\n1,%s\n" "$(printf 'A%.0s' {1..32769})" > too_long_cell.csv
printf "id,bigcell\n1,%s\n" "$(printf 'A%.0s' {1..150000})" > too_long_cell.csv
printf "id,bigcell\n1,%s\n" "$(printf 'A%.0s' {1..10585761})" > too_long_cell.csv
```

The failure happens cleanly once the cell exceeds **1024 × 1024 (1,048,576) characters**, with a clear and informative error message:

> `Maximum number of characters per column=1048576`

<img width="767" height="716" alt="Image" src="https://github.com/user-attachments/assets/1e18144f-2d20-4cfc-af0c-1d9feadf5702" />

Should we close it?

PR #7219, opened 2025-03-24
Increase limit for CSV cell size

wetneb:
Since [the current limit seems too low](https://forum.openrefine.org/t/errors-length-of-parsed-input-32769-exceeds-the-maximum-number-of-characters-defined-in-your-parser-settings-32768/2180/3), I think it's worth bumping this up. This makes it possible to store 1 million characters in each cell, which seems ample.


wetneb:
Related to #7206.

PR #7223, opened 2025-03-25
wikibase: Fix columnar analysis of upload operation

wetneb:
Fixes an oversight in my previous implementation of columnar dependency analysis for Wikibase operations.


wetneb:
Actually, I realized there is another issue with this new metadata: the "Wikibase editing results" column is only created if it doesn't exist, which doesn't match the semantics of "ColumnsDiff", so I need to adapt that.

PR #7225, opened 2025-03-26
fix: JSON deserialization of RowAdditionOperation

wetneb:
I realized that the serialization and deserialization code of RowAdditionOperation were inconsistent: the index at which the rows should be inserted was called `index` in serialization, but `insertionIndex` in deserialization.

This fixes deserialization. Since any serialized versions of this operation in workspaces will use the `index` field, there shouldn't be any need to keep supporting `insertionIndex`.

This PR conflicts with #7224: depending on which one gets merged first, the other one will need to be adapted to change the frontend to submit `index` instead of `insertionIndex`.


wetneb:
Oh I had completely forgot about this PR. Yes it would make sense to add it too. It is still necessary even after #7247, but the bug it fixes isn't as serious and would only play a role when trying to replay a row addition operation using the `Apply` dialog of the Undo/Redo tab.

I have rebased the PR to resolve the conflicts.

SoryRawyer (approved):
looks good, thanks for updating!

Issue #6538, opened 2024-04-15
Eliminate all uses of raster graphics in the UI


wetneb:
This issue is here to keep track of all the raster images that we include as part of OpenRefine's UI.
We would like to migrate them to SVGs instead, for better scalability and flexibility of styling (for instance to implement a dark mode).

As we work on migrating away from those raster images, we would like to preserve (or improve) the consistency of the visual style of OpenRefine, but we don't expect pixel-perfect replacements at all.
Any help to replace those PNG files would be most welcome, even without making the associated code changes: feel free to just upload the proposed SVG replacements as a comment on this issue and someone else can take care of integrating them in the code base.

**Note:** in the [graphics](https://github.com/OpenRefine/OpenRefine/tree/master/graphics) folder we have some PSD files which you could also work from (they are likely in higher resolution)

This is the current list of PNGs:
* [ ] [main/webapp/modules/core/images/close.png](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/images/close.png)
* [ ] [main/webapp/modules/core/images/slider-handle.png](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/images/slider-handle.png)
* [ ] [main/webapp/modules/core/images/menu-dropdown.png](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/images/menu-dropdown.png)
* [ ] [main/webapp/modules/core/images/scatterplot-icons.png](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/images/scatterplot-icons.png)
* [ ] [main/webapp/modules/core/images/leftpanel-showhide.png](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/images/leftpanel-showhide.png)
* [ ] [main/webapp/modules/core/images/minimize-map.png](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/images/minimize-map.png)
* [ ] [main/webapp/modules/core/images/star-flag-map.png](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/images/star-flag-map.png)
* [ ] [main/webapp/modules/core/images/right-arrow.png](https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/images/right-arrow.png)
[...]

PR #7226, opened 2025-03-26
replace png arrows with svg

smierz:
part of #6538

two of the png files within "main/webapp/modules/core/images/" are replaced with svg images

_**right-arrow.png**_ 
which is used within menu items to open sub-menus

Before|After
menu-png-right-arrow | menu-svg-right-arrow

<br>

_**down-arrow.png**_ 
which is used in Wikibase menu for example

Before|After
wikibase-png-down-arrow | wikibase-svg-down-arrow

and for opening the sorting sub-menu

Before|After
sort-png-down-arrow |sort-svg-down-arrow

and in a third use case, I didn't know how to reconstruct -> https://github.com/OpenRefine/OpenRefine/blob/master/main/webapp/modules/core/scripts/views/data-table/data-table-view.js#L376


wetneb (approved):
Looks great to me, thank you for the care in making the comparison so easy for reviewers.

smierz:
thanks :)

(btw I can't edit/update the checkboxes in the corresponding issue myself unfortunately)

wetneb:
Thanks for pointing that out, I have updated it. As far as I am concerned I think that's a good case for inviting you into the GitHub org already, but we're in the middle of governance transitions and the process for that is sadly still unclear, and perhaps others would prefer to see more contributions before such an invite.

By the way, I see you contributed to OpenAlex, that's really cool! I'm curious what brings you to OpenRefine and whether we can do anything to help :)

Issue #5576, opened 2023-01-19
Better generalizability for the reorder-columns operation


wetneb:
We offer an operation, `reorder-columns`, which is able to change the order of columns in an arbitrary way and remove any number of columns, in a single step.

![image](https://user-images.githubusercontent.com/309908/213440782-41467e32-c841-424a-bb17-6a049a61691e.png)

The operation which corresponds to this screenshot is represented internally as:
```json
  {
    "op": "core/column-reorder",
    "columnNames": [
      "Identifiant du lieu",
      "Année du tournage",
      "Type de tournage",
      "Titre",
      "director",
      "producer",
      "Réalisateur",
      "Producteur",
      "Code postal",
      "Coordonnée en X",
      "Coordonnée en Y",
      "geo_shape",
      "geo_point_2d"
    ],
    "description": "Reorder columns"
  }
```

In other words, the operation parameters simply remember what is the final order of the remaining columns after reorder.
As explained in #4055, this operation does not generalize well to other datasets, when it is used in the "Apply" dialog of the Undo/Redo tab, because any columns that were not in the original dataset but are present in the new one will be deleted by this operation.

### Proposed solution

We should find other ways to specify this operation so that the actual intent of the user is captured better.
It is difficult to come up with a precise specification, but as a litmus test, I would expect the following criteria to be satisfied:
* If I use the dialog above to delete columns only, without reordering the remaining columns, I would expect that the generalization of this operation only deletes the said columns, and leaves any other column (including those not present in the original dataset) untouched. This could be achieved by calling the `column-removal` operation instead, as it supports removing multiple columns at once after #5563.
* If I use the dialog to reorder columns only, without deleting any column, then I expect that the generalization of this operation will not delete any column in any other context. Additional columns not present in the original dataset should be placed in a *sensible* location in the resulting table. It is not a problem if this exact order is complicated to define, as the exact position of columns should generally not have a significant influence on operation workflows except in certain cases (such as the use of transpose operations).

### Alternatives considered

One could decide to break down the operation into multiple steps, applying the operations that remove or move a single column multiple times, to reach the desired state. Working on such a decomposition is likely useful to understand this issue better, but I would rather prefer that this dialog does not generate lengthy lists of steps in the project history.

### Additional context

Follow-up to #4055 and #5563.

thadguidry:
I tend to agree with the proposal.  So are you for/against splitting the dialog into two menu items?
[...]

PR #7227, opened 2025-03-26
Improve reproducibility of the column reordering operation

wetneb:
This implements the two criteria defined in #5576 concerning the behaviour of the operation when:
* only reordering columns, without deleting any
* only deleting columns, keeping the order of the remaining ones

Those changes can be made without changing the dialog's UI. Making the operation analyzable in general would likely require a different UI to let the user better specify their intent.

I am not sure if it should close the issue, maybe it's good enough, maybe not.

### DIFF
--- a/main/tests/server/src/com/google/refine/operations/column/ColumnReorderOperationTests.java
+++ b/main/tests/server/src/com/google/refine/operations/column/ColumnReorderOperationTests.java
@@ -37,6 +37,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.node.TextNode;
 import org.testng.Assert;
 import org.testng.annotations.BeforeMethod;
@@ -45,9 +47,11 @@ import org.testng.annotations.Test;
 
 import com.google.refine.RefineTest;
 import com.google.refine.model.AbstractOperation;
+import com.google.refine.model.ColumnsDiff;
 import com.google.refine.model.Project;
 import com.google.refine.operations.OperationDescription;
 import com.google.refine.operations.OperationRegistry;
+import com.google.refine.util.ParsingUtilities;
 import com.google.refine.util.TestUtils;
 
 public class ColumnReorderOperationTests extends RefineTest {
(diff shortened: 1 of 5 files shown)


SoryRawyer (commented):
@wetneb apologies for the delay reviewing this! Just to confirm, this is somewhat of a breaking change in that columns will no longer be removed in every case. Is that right? I know that's what the user feedback gathered by @magdmartin states, but if and when this is released I want to make sure to point this out in the release notes.

Also, it looks like there are `reorder-columns` commands in other frontend components as well: `menu-edit-column.js` and `data-table-view.js` also have reorder commands in them. Should those be updated as well (even if only to avoid having a `null` value for `isPureReorder`)?

main/tests/server/src/com/google/refine/operations/column/ColumnReorderOperationTests.java
```diff
@@ -71,15 +75,34 @@ public void createProject() {
...
+                + "\"isPureReorder\": false,"
                 + "\"columnNames\":[\"b\",\"c\",\"a\"]}");
     }
 
+    @Test
+    public void deserializeLegacyFormat() throws JsonMappingException, JsonProcessingException {
+        ColumnReorderOperation op = ParsingUtilities.mapper.readValue("{\"op\":\"core/column-reorder\","
+                + "\"columnNames\":[\"b\",\"c\",\"a\"]}", ColumnReorderOperation.class);
+        assertEquals(op._columnNames, List.of("b", "c", "a"));
+        assertEquals(op._isPureReorder, false);
```
wetneb, line 90: @SoryRawyer The test above shows the backwards-compatibility for this change. The old serialization format of this operation (without the `isPureReorder` field) will still be understood as before, as an operation that removes any columns that are not mentioned in the `columnNames` list.

This means that if users execute recipes generated from a previous version of OpenRefine with the new version, they should still get the same results.

wetneb:
> Also, it looks like there are reorder-columns commands in other frontend components as well: menu-edit-column.js and data-table-view.js also have reorder commands in them. Should those be updated as well (even if only to avoid having a null value for isPureReorder)?

That's a very good point! The column reordering operation that is used to implement the "join" action is one that is used to remove columns only, without doing any reordering. Thanks to the backwards-compatibility mentioned above, it doesn't need a `isPureReorder` field to continue working (but we could also add one for clarity if you want). In any case, I wanted to migrate this to use the `column-multi-removal` operation, as it would make it generalize better. Should I do this in this PR too?

SoryRawyer:
> Thanks to the backwards-compatibility mentioned above, it doesn't need a isPureReorder field to continue working (but we could also add one for clarity if you want). In any case, I wanted to migrate this to use the column-multi-removal operation, as it would make it generalize better. Should I do this in this PR too?

Thanks for the context! I don't think that's necessary to include in this PR.

I think this looks good! I'm going to approve but wait a day or two to merge in case anyone has a last-minute comment.

SoryRawyer (approved):

PR #7228, opened 2025-03-26
Update Cypress dependency files

wetneb:
Every time I run the Cypress test suite locally with `yarn install` and then `yarn run cypress open`, it makes changes to `package.json` and `yarn.lock` which I then need to avoid including in my commits. Perhaps this also happens to others and updating them once will make everyone's life easier.


SoryRawyer:
@wetneb, this would be really helpful. Would you mind updating the PR to use the latest master branch?

wetneb:
Done!

SoryRawyer (approved):
Thanks for taking care of this!

Issue #7074, opened 2025-01-09
Create Icons for  Facet,Commons transforms, and View


magdmartin:
Following #7060 we want to create icons for the operations under the `Facet`, `Commons transforms`, and `View` menu.
See also [the design proposed on the forum](https://forum.openrefine.org/t/operation-logos/1307)

> > I want to confirm that all the operations under Facet,Commons transforms, and View will be icon-less, and we are ok with having menu options with and without icons throughout the application.

> I plan to add icons for all operations, this was just a first batch of them.
>I would be happy to add icons for other things (such as facets) if people see the need for them (and have ideas how to represent them), but at the moment the priority is to cover operations because those icons will be used in menus, history and recipe visualization.

_Originally posted by @wetneb in https://github.com/OpenRefine/OpenRefine/issues/7060#issuecomment-2578589984_

DaxServer:
I've been using the 3.10 Snapshot for a few hours and have some feedback on the icons. Some context menus have a mix - with icons and without icons. It's getting a bit hard on the eyes and straining. Does anyone else also experience this?

<img width="132" height="278" alt="Image" src="https://github.com/user-attachments/assets/8530b4b5-80c4-4684-984b-68076347b141" />

SoryRawyer:
Thanks for the feedback @DaxServer! I think that's different enough from this issue so I created a new issue to track the design and implementation.

Aric262:
I would like to work on this issue. Please assign it to me

PR #7233, opened 2025-03-28
Introduce icons for facet types

wetneb:
For #7074.
I'm also thinking that the icons defined here could be used in the visualization, to make it clearer which filters are applied on each operation, but I don't have a precise idea of how that could work yet.

image


SoryRawyer (approved):
Looks great! I have a small question that shouldn't block this PR: did you create these icons based on the conversation in [the "Operation logos" thread](https://forum.openrefine.org/t/operation-logos/1307), or did these come from somewhere else?

PR #7236, opened 2025-03-31
Avoid name collisions of internal columns when applying recipes

### LINKED ISSUE #7235: Avoid name collisions of internal columns when applying recipes
Consider the following recipe:
1. Create a column `response` by fetching URLs from column `id` with expression `grel:"https://example.com/api/" + id + ".json"`
2. Create a column `species` from `response` using expression `grel:value.parseJson().species.join(', ')`
3. Delete column `response`

If I try to apply this recipe to a project where there is already a `response` column, the application of the recipe will fail, because the creation of the `response` column conflicts with the existing column. The column mapping feature I added in #7158 does not help because `response` isn't a created column of the recipe (since it's deleted by the last operation).

### Proposed solution

Let's call such columns "internal" columns: columns that are created by a recipe but are deleted before the end of the recipe. As a user, I don't actually care about the name of those columns when applying a recipe, since they won't be visible in the final grid. So I want OpenRefine to automatically rename those internal column names so that they don't conflict with existing column names. The recipe would then become:

1. Create a column `response_2` by fetching URLs from column `id` with expression `grel:"https://example.com/api/" + id + ".json"`
2. Create a column `species` from `response_2` using expression `grel:value.parseJson().species.join(', ')`
3. Delete column `response_2`

(assuming `response_2` doesn't exist in the project yet - the code would automatically look for a free column name of this form)

### Alternatives considered

We could also decide to expose such columns in the mapping dialog, letting the user deduplicate them manually, but I think it's adding more noise to the dialog.

wetneb:
Fixes #7235 (see the motivation there).

### DIFF
--- a/main/src/com/google/refine/commands/history/ApplyOperationsCommand.java
+++ b/main/src/com/google/refine/commands/history/ApplyOperationsCommand.java
@@ -39,6 +39,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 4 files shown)


SoryRawyer (approved):
I think this makes sense as an approach; if a user doesn't need to rely on the column name later in the project, then I think it's better for OpenRefine to manage the naming conflict instead of the user.

PR #7247, opened 2025-04-05
Fix deserialization of RowAdditionOperation.

### LINKED ISSUE #7245: Mass edit of new blank cells corrupts project in 3.9.2
Initially mentioned in [this forum thread](https://forum.openrefine.org/t/after-adding-rows-and-editing-a-cell-project-failing-to-load/2214)

### To Reproduce
1. Using OpenRefine 3.9.2
2. Open any project
3. Add multiple rows to the end of the project
4. Edit one of the cells and click "Apply to all identical cells"

### Current Results
The project cannot be loaded. After editing and upon re-opening the project, the browser is redirected to the loading screen.
### Expected Behavior
The result entered into the edit dialog box should be applied to all empty cells in the column.

### Versions
 - Operating System: Ubuntu 24.04
 - Browser Version: Firefox 136.0.3
 - JRE or JDK Version: 21.0.5
 - OpenRefine: 3.9.2

wetneb:
Fixes #7245.

See [my comment](https://github.com/OpenRefine/OpenRefine/issues/7245#issuecomment-2780552271) for an explanation of the root cause of the issue.

To fix it, I have:
* made sure that the operation doesn't add its own list of rows to the project, but a copy of it, so that its metadata remains unchanged when further changes happen to the grid
* introduced a new serialization format for the operation. The goal with this move is to make sure that we are able to deserialize all values that have already made it into people's workspaces. Using the fact that the row addition operation only lets the user add empty rows for now, I change the deserialization code so that the `rows` parameter of the operation is parsed as a generic array, and only its length is taken into account to generate a fresh list of empty rows (we really should have gone for just encoding the number of rows in the operation metadata, not the rows, but it's too late). A new parameter (`addedRows`) is introduced, which can be used in the future to retain actual row data (if the UI is ever made to support it), in which case the injection of the `Pool` object will need to be added on top.

I have also added a regression test to demonstrate the fix.

I would recommend to release this as 3.9.3 as soon as possible.

### DIFF
--- a/main/src/com/google/refine/operations/row/RowAdditionOperation.java
+++ b/main/src/com/google/refine/operations/row/RowAdditionOperation.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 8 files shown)


SoryRawyer (approved):
Thanks for sorting this out! I think this looks good, and I've been so far unable to reproduce the bug using this branch.

PR #7260, opened 2025-04-15
Resolve cypress test failures from language updates

SoryRawyer:
- This aligns the cypress tests with the language updates made in #7193. I don't think the cypress tests run automatically on pull requests that update the language files, which might explain why these failures weren't noticed before merging the PR.


wetneb (approved):
Good catch! The configuration of the GitHub Actions should ideally be tweaked so that changes to `translation-en.json` still trigger a build (but not for any other translation file).

tfmorris:
> That's what I was thinking too. I'll open an issue to include that specific file in the CI workflow.

#7261

PR #7268, opened 2025-04-18
feat: add mouse shortcuts to column headers

### LINKED ISSUE #6282: Improving renaming columns
When I go to rename a column, I thought it would be easier to click on there column name itself and edit there directly, instead of first going to 'Edit this column' then 'Rename' as pictured in the screenshot below.

I believe that change would be more intuitive from a usability standpoint as well, since I first think of 'Edit this column' as pertaining to the information in the column itself, not its descriptor above.

Screenshot 2024-01-04 at 11 14 38 AM

ntcho:
Resolves #6282

- Add double click mouse shortcut to open column renaming dialog
- Add right click mouse shortcut to open column menu
- Add middle click mouse shortcut to collapse column

Functionally it makes sense, but I think there should be a better visual clue that the shortcut exists.

Also the code for `dblclick` is duplicated from the [later part of the file](https://github.com/ntcho/OpenRefine/blob/209adfded01b10371652b37b06d23c9f3c7eebae/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js#L345-L396), I think there should be a better way to refactor it too.

### DIFF
--- a/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
@@ -121,6 +121,18 @@ DataTableColumnHeaderUI.prototype._render = function() {
     elmts.columnNameInput.trigger('focus').trigger('select');
   });
 
+  // shortcut: right click to show menu
+  this._td.addEventListener('contextmenu',function(e) {
+    e.preventDefault();
+    self._createMenuForColumnHeader(elmts.dropdownMenu);
+  });
+
+  // shortcut: middle click to collapse column
+  this._td.addEventListener('auxclick',function(e) {
+    self._dataTableView._collapsedColumnNames[self._column.name] = true;
+    self._dataTableView.render();
+  });
+
   elmts.dropdownMenu.on('click',function() {
     self._createMenuForColumnHeader(this);
   });
--- a/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
+++ b/main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
@@ -67,6 +67,60 @@ DataTableColumnHeaderUI.prototype._render = function() {
   var elmts = DOM.bind(td);
 
   elmts.nameContainer.text(this._column.name);
+
+  // shortcut: double click to rename column
+  this._td.addEventListener('dblclick',function() {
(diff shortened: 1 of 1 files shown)


magdmartin:
@ntcho thanks for the PR. I haven't compiled and reviewed the changes, but I'm curious about how you are addressing the discoverability concerns raised in https://github.com/OpenRefine/OpenRefine/issues/6282.

ntcho:
My current idea is having a muted text within the menu item describing the shortcut. For example:

`[ICON] Rename column          [ double click ]`
`[ICON] Collapse column        [ middle click ]`

I’ll look into some UI patterns for this and share a simple mockup here.

ntcho:
My current idea is having a muted text within the menu item describing the shortcut. For example:

`[ICON] Rename column          [ double click ]`
`[ICON] Collapse column        [ middle click ]`

I’ll look into some UI patterns for this and share a simple mockup here.

SoryRawyer:
Hey @ntcho, I just wanted to check in and see if you had any questions about the mock-ups or if you were waiting on anything from either me or @magdmartin. Aside from the discoverability component, I think this is ready to move out of draft status.

ntcho:
image

This is how I'm envisioning showing the shortcut (the `Alt + Click` shortcut is just an example for the hint, not an actuao shortcut). 

Alternatively, we can consider pure icon hints like this to reduce i18n work:

magdmartin:
Circling back on this PR since it greatly improves the user experience. 

After review, I think it is best **not** to introduce a new icon in the drop-down menu. We're currently moving toward a consistent use of icons to indicate the type of action a menu item will perform. Adding a new visual element with a different meaning could confuse users and reduce UI consistency.

Instead, as suggested in @tfmorris's [comment](https://github.com/OpenRefine/OpenRefine/issues/6282#issuecomment-1877918607), I propose highlighting the header in yellow on mouse hover, similar to the existing behavior for the project name. 

The tooltip in our case can be `double click to rename column` 

image



Additionally, we'll need to assess how this PR interacts with the column drag-and-drop functionality introduced in #7301 to ensure compatibility.

SoryRawyer:
Hi @ntcho and @magdmartin - I went ahead and added changes for the feedback described above. I opened a new pull request here: #7361. As a result, I'll close this pull request and we can discuss the changes on the new PR. We can always reopen this one later if we decide to go with this branch instead of the new one.

Issue #7214, opened 2025-03-20
Parsing Configuration: Disabling "Store blank columns" also removes file source column


b2m:
Disabling the checkbox on "Store blank columns" will also remove the file source column in the parsing configuration options screen.

### To Reproduce

1. First, create a project using multiple files.
2. Then, disable the checkbox "Store blank columns" in the parsing configuration options screen.

### Current Results

The column "File" with the file source also disappears, despite active checkbox on "Store file source".

### Expected Behavior

The column "File" stays when disabling the checkbox on "Store blank columns" and active checkbox on "Store file source".

### Screenshots

![Image](https://github.com/user-attachments/assets/a640d833-b301-4948-bd1d-051ebb17cafc)

![Image](https://github.com/user-attachments/assets/f1c653ef-39e4-4bd5-908e-9ea864f2c572)

### Versions

 - Operating System: Windows 10
 - Browser Version: Firefox 136
 - JRE or JDK Version: embedded
 - OpenRefine: 3.9.1

### Datasets

- [test-empty-cols1.csv](https://github.com/user-attachments/files/19369106/test-empty-cols1.csv)
- [test-empty-cols2.csv](https://github.com/user-attachments/files/19369107/test-empty-cols2.csv)

smierz:
Problem seems to be the order of operations:
* first the columns for filename/archive are added to the project in [ImportingParserBase](https://github.com/OpenRefine/OpenRefine/blob/master/modules/core/src/main/java/com/google/refine/importers/ImportingParserBase.java#L131-L137) at index 0 or 1
* then the rows get filled with the data from the csv files in `TabularImportingParserBase.readTable`
* then columns get dropped that have no data in them in `TabularImportingParserBase.deleteEmptyColumns`
--> since the columns for filename and archive have not been filled yet, they get dropped 📍 
* and in the end the [filename/archive should be added into the columns` rows](https://github.com/OpenRefine/OpenRefine/blob/master/modules/core/src/main/java/com/google/refine/importers/ImportingParserBase.java#L154-L164) from step 1, but since they were removed the data is not added

smierz:
Another problem with this processing order arises because the "store blank columns" property is calculated per file.

So if I extend the provided example to three files, with the "empty" column being
* not empty in file 0
* completely empty in file 1
* not empty in file 2
then the column gets added, removed, added resulting in data disappearing

[...]

PR #7272, opened 2025-04-23
store blank columns delete only after all files parsed

### LINKED ISSUE #7271: "store blank columns" is evaluated per file, might remove data prematurely
### To reproduce 

* First, create a project using multiple files, e.g. three csv files where one of the columns "empty" is
  * not empty in file 0
  * completely empty in file 1
  * not empty in file 2

*  Then, disable the checkbox "Store blank columns" in the parsing configuration options screen.

### Current Results

then the column gets added, removed, added resulting in data disappearing.

I hope it gets clearer with this image, where one of the entries in row "empty" disappears (before 2 entries, after 1 entry), blue lines marking the data per file.

Image

Image

### Expected Behavior

The column should only be removed if empty in all files.

smierz:
fixes #7271 

- instead of deleting empty columns per file, we store the boolean array "columnHasData" in the options
- we use it only after all files have been parsed to delete empty columns


smierz:
@wetneb don t know if [your offer](https://github.com/OpenRefine/OpenRefine/issues/7214#issuecomment-2763425560) still stands, but here is one PR hoping to improve the `storeBlankColumns` situation

SoryRawyer (commented):
Thanks for the PR! I just have a small question that I couldn't figure out looking at these changes. It looks like there's coverage that columns are removed after processing one file, but do any of the remaining tests confirm that blank columns are retained after `parseOneFile`?

smierz:
I'm not sure I understand the question, currently the deletion of columns happens within the processing of `parseOneFile`, this is what's causing the linked issue.

The changes proposed take the deletion out of `parseOneFile` so within the processing is no deletion of columns anymore, so columns are always retained after `parseOneFile`.

main/tests/server/src/com/google/refine/importers/ExcelImporterTests.java
```diff
@@ -420,37 +426,33 @@ public void readMultiSheetXlsx() throws IOException {
...
+        JSONUtilities.safePut(options, "storeBlankCellsAsNulls", false);
+        JSONUtilities.safePut(options, "storeBlankColumns", false); // delete empty columns
+
+        parse(SUT, fileRecords, options);
 
         // We should have one less than the start due to empty column being skipped
         assertEquals(project.columnModel.columns.size(), 12);
-        // NOTE: we need to redirect through the column model because the rows will still have empty cells
-        assertEquals(project.rows.get(1).getCellValue(project.columnModel.columns.get(1).getCellIndex()), true);
-        assertEquals(project.rows.get(1).getCellValue(project.columnModel.columns.get(2).getCellIndex()), EXPECTED_DATE_TIME);
```
wetneb, line 445: Wouldn't it be worth retaining those assertions above in the new version of the test too? I agree that the `verify` ones below can be deleted though.

main/tests/server/src/com/google/refine/importers/FixedWidthImporterTests.java
```diff
@@ -159,40 +165,41 @@ public void readColumnHeader() throws Exception {
...
+        JSONUtilities.safePut(options, "ignoreLines", 0);
+        JSONUtilities.safePut(options, "headerLines", 1);
 
-        ArrayNode columnNames = ParsingUtilities.mapper.createArrayNode();
-        columnNames.add("Col 1");
-        columnNames.add("Col 2");
-        columnNames.add("Col 3");
-        columnNames.add("Col 4");
-        columnNames.add("Col 5");
-        columnNames.add("Col 6");
```
wetneb, line 180: It looks like this test case supplied some custom column names. Wouldn't it be worth retaining them? Or is this functionality adequately tested elsewhere?

wetneb (dismissed):
Thanks a lot for working on this! It looks very good overall, but intuitively, I would rather keep the parsing options read-only and not use it to store the information of whether columns are empty or not. The `deleteEmptyColumns` method could do a loop over the project table to compute this information at that stage only. This would also let us remove the corresponding code from the tabular parsing methods and make it easier to offer this option for all importers in the future.

smierz:
Thank you for taking the time to review this PR.

**Re: extra loop for determining columns to delete**
Before starting the implementation, I explored different approaches, including the "brute force" loop over all rows and columns to determine whether a column should be deleted. Ultimately, I decided against that method because 
1) it would remove the already existing and approved logic for determining columns for deletion and
2) it would introduce an additional full loop over the project (after the parsing loop and the adding filename/archive loop) -> could degrade performance.

I do agree with you though that adding the additional loop that works on a `project` instead of a `line` would be beneficial for reusing the same functionality for all parsers (incl the tree-based ones).

**Re: addition of a temporary variable to the options object**
While options are treated as mostly stable, they are not strictly read-only — they are modified several times across the different parsers, e.g. `PCAxisImporter.parseOneFile` only exists to overwrite options.

Another reason for this approach: it allowed me to reuse the `columnHasData` logic, which had already been reviewed and approved in [a previous PR](https://github.com/OpenRefine/OpenRefine/pull/4757) . I hoped reusing an existing, approved method would help maintain consistency and simplify the review process.

wetneb:
Thanks for explaining your reasoning! If you ask me, I would still recommend not modifying options.

> it would remove the already existing and approved logic for determining columns for deletion and

I think it would be acceptable to move this logic somewhere else - it feels like an acceptable refactoring.

> While options are treated as mostly stable, they are not strictly read-only — they are modified several times across the different parsers, e.g. PCAxisImporter.parseOneFile only exists to overwrite options.

I think it's good to avoid introducing more of those writes, to make it easier to move to a read-only solution in the future.

smierz:
ok, let's backtrack to the point where we are still d'accord:
*  reintroduced the assertions and columnNames into the tests (make no difference)
* reverted the commit with the solution where the options get a temp variable (tests will fail because reverted changes to parse method)


next steps:
* also add tests for https://github.com/OpenRefine/OpenRefine/issues/7214 as an additional loop should solve the issue as well
* implement the loop to determine blank columns within `ImportingParserBase.parse`

does this make sense?

smierz:
fyi there was an additional error within the functionality it seems:
within the `ExcelImporterTests` Column 6 only consists of empty Strings (see `createDataRow`, line 662) but was previously not determined as empty and not deleted  
--> needed to additionally fix the tests and also renamed the tests to emphasize that they should delete "blank" columns

main/tests/server/src/com/google/refine/importers/SeparatorBasedImporterTests.java
```diff
@@ -706,13 +741,26 @@ protected void prepareOptions(
         whenGetBooleanOption("includeArchiveFileName", options, includeArchiveFileName);
     }
 
+    private ObjectNode createBasicOptions(String sep, int limit, int skip, int ignoreLines,
```
SoryRawyer, line 744: Do you think it'd be worth moving this up to `ImporterTest` so that we could use it in the other importer test classes? Or does each importer have enough difference between their basic options that it doesn't make sense to generalize this a bit more?

modules/core/src/main/java/com/google/refine/importers/ImporterUtilities.java
```diff
@@ -198,6 +198,26 @@ static public void setupColumns(Project project, List<String> columnNames) {
         project.columnModel.update();
     }
 
+    /**
+     * @param columnsHasData
+     *            Record if there is data in each column( false:null;true:has data)
+     */
```
SoryRawyer, line 204: I think this needs an updated docstring now that it's moved.

SoryRawyer (commented):
Overall I think this looks good, I'd just be interested in hearing your thoughts on whether it'd be useful to generalize the `createBasicOptions` method.

main/tests/server/src/com/google/refine/importers/SeparatorBasedImporterTests.java
smierz: hmm maybe not in their current form as there are some options that are not relevant to other importers (as you mentioned). also I don't like the listing of values as it becomes hard to follow, if you open the file in GitHub or a simple text editor when there are no IDE features that tell you what each variable is for (and yes, I coded it, but only to stay consistent with the already existing functions `prepareFunctions`).

but it is worth looking into it, I rewrote it a bit to only set the default values for the tabular parsers as they have a common set of options, and then when using the method in testing, one would only overwrite the values relevant to the testing, which I think highlights the logic under test.

--> see new commit ["mv creation of options up to ImporterTest"](https://github.com/OpenRefine/OpenRefine/pull/7272/commits/3259731c83de11a4427a3eb887feebf71b22bc5c) and let me know what you think

smierz: btw I noticed two errors in `createBasicOptions` when I reworked it:
* `guessCellTypes` needs to be `guessCellValueTypes`
* `ignoreQuotes` is not an option, it needs to be `JSONUtilities.safePut("processQuotes", options, !ignoreQuotes);`

modules/core/src/main/java/com/google/refine/importers/ImporterUtilities.java
smierz: tbh the method in its current form is not very reusable imo. I kept it as is because it was already reviewed and as a first step my only goal was to fix the bugs by moving around the existing functionality. but the method is quite specific with its List of Booleans that need to be translated to the column indices. It would probably be easier to  understand and reuse this method if the input was a project and a List of Integers with columnIndices to delete. wdyt?

main/tests/server/src/com/google/refine/importers/SeparatorBasedImporterTests.java
smierz: ok 5 minutes after uploading this commit and closing my IDE, I was like: hmm but if I set all values to their defaults/fallbacks anyway, then why do I need a creation method at all?
-> `getDefaultTabularImportingOptions`  may simply be replaced by empty options

modules/core/src/main/java/com/google/refine/importers/ImporterUtilities.java
```diff
@@ -198,6 +198,26 @@ static public void setupColumns(Project project, List<String> columnNames) {
...
+    static public void deleteEmptyColumns(Project project, List<Boolean> columnsHasData) throws ModelException {
```
tfmorris, line 205:
```suggestion
    static void deleteEmptyColumns(Project project, List<Boolean> columnsHasData) throws ModelException {
```

Since this is mainly an internal helper, let's not make it public. Given that, I don't think it makes a huge difference what it takes for parameters (although I'd suggest boolean[] for efficiency) instead of List<Boolean>), for reasons you'll see in my comments on the caller.

modules/core/src/main/java/com/google/refine/importers/ImporterUtilities.java
```diff
@@ -198,6 +198,26 @@ static public void setupColumns(Project project, List<String> columnNames) {
...
+        // skip, if no columns to delete
+        if (columnsHasData.isEmpty() || columnsHasData.stream().allMatch(Boolean.TRUE::equals)) {
+            return;
+        }
+
```
tfmorris, line 210:
```suggestion
```

This extra pass is unnecessary, I think, since it'll be handled naturally by your loop below.

modules/core/src/main/java/com/google/refine/importers/ImporterUtilities.java
```diff
@@ -198,6 +198,26 @@ static public void setupColumns(Project project, List<String> columnNames) {
...
+        project.columnModel.update(); // make sure all our cell indexes are up to date
+        for (int c = 0; c < columnsHasData.size(); c++) {
+            if (Boolean.FALSE.equals(columnsHasData.get(c))) {
+                // remove column from columns
```
tfmorris, line 214:
```suggestion
            if (!columnsHasData.get(c))) {
```

or `!columnsHasData[c]` with a `boolean[]`

modules/core/src/main/java/com/google/refine/importers/ImportingParserBase.java
```diff
@@ -103,6 +104,12 @@ public void parse(Project project, ProjectMetadata metadata,
                 break;
             }
         }
+
+        // if user doesn't choose 'storeBlankColumns', delete all empty columns.
```
tfmorris, line 108:
```suggestion
        // if user set 'storeBlankColumns' to false, delete all empty columns.
```
Just a small rewording to make it clearer what happens if they take the default and don't set anything

modules/core/src/main/java/com/google/refine/importers/ImportingParserBase.java
```diff
@@ -244,4 +251,31 @@ private static int addColumn(Project project, String columnName, int columnId) {
...
+            while (rowSize >= columnsHasData.size()) {
+                columnsHasData.add(Boolean.FALSE); // init every col as empty
+            }
+
+            for (int i = 0; i < rowSize; i++) {
+                if (!row.isCellBlank(i)) {
+                    columnsHasData.set(i, Boolean.TRUE); // only if an entry is found, set col to hasData
+                }
+            }
+        }
```
tfmorris, line 272: I'm concerned about the performance impact of this for large projects. I think there are a few things that we can do to mitigate this:
- use `boolean[]` instead of `List<Boolean>` for the flags
- check the flag before checking the data to short circuit the eval ie `if (!columnsHasData[i] && !row.isCellBlank(i))`
- move the starting and ending indices for the cells in the rows so as to not recheck columns where we've already found data

I have a sketch of the code for this if it would be useful.

modules/core/src/main/java/com/google/refine/importers/TabularImportingParserBase.java
```diff
@@ -226,30 +215,8 @@ static public void readTable(
...
-    static public void deleteEmptyColumns(List<Boolean> columnsHasData, Project project) throws ModelException {
-        project.columnModel.update(); // make sure all our cell indexes are up to date
-        for (int c = 0; c < columnsHasData.size(); c++) {
-            if (!columnsHasData.get(c)) {
-                // remove column from columns
-                project.columnModel.removeColumnByCellIndex(c);
-            }
         }
-        project.columnModel.update();
     }
```
tfmorris, line 221:
```suggestion
     /**
     * If "storeBlankColumns" == false, delete blank columns.
     *
     * @param columnsHasData
     *            Record if there is data in each column( false:null;true:has data)
     */
     @Deprecated
    static public void deleteEmptyColumns(List<Boolean> columnsHasData, Project project) throws ModelException {
    ImporterUtilities.deleteEmptyColumns(project, columnsHasData);
    }
```

Since this is technically a public method (even though it probably shouldn't have been and no one should really be using it), let's practice good hygiene and retain it with an `@Deprecated` annotation.

tfmorris (changes requested):
Thanks for tackling both these issues and a rather gnarly bit of code!

It took me a little while to figure out what was going on with the tests. After digging in to it, I now know why the change from the mocks to an actual `options` object is necessary, but it would be helpful to reviewers to minimize the changes to existing tests and make them easier to review since test modifications are usually a red flag. I eventually convinced myself that no tests had been removed, but it took me a while.

Given the past problems we've had with this code, I think we should try to make the tests as robust as possible and test all the edge conditions we can think of. I noticed test cases for the last column being empty and interior columns being empty. Do we have one for the first column being empty? That might be a good edge case to add, if we don't already have it covered.

I understand your comment from the earlier discussions about wanting to use the cell checking code unchanged, since it had already been reviewed, but I think the potential performance impact on large project creation justifies trying to optimize it some.

Thanks again for all your work on this and apologies for the lengthy review cycle.

modules/core/src/main/java/com/google/refine/importers/ImportingParserBase.java
tfmorris: Here's the commit with the performance optimization: https://github.com/tfmorris/OpenRefine/commit/594e5ff70127e861c36bc9b7d1063045c53cf2d7
There are a couple of other commits on that fork with some of my other feedback comments implemented.

modules/core/src/main/java/com/google/refine/importers/TabularImportingParserBase.java
smierz: re-added the original public method in `TabularImportingParserBase` with a deprecation notice (see commit  "[deprecate deleteEmptyColumns instead of removing it](https://github.com/OpenRefine/OpenRefine/pull/7272/commits/ae1ae5afc9226a24f20b4bcb4a6b3b07af7ae8bd)")

modules/core/src/main/java/com/google/refine/importers/ImporterUtilities.java
smierz: removed the public modifier

main/tests/server/src/com/google/refine/importers/ExcelImporterTests.java
```diff
@@ -420,37 +427,108 @@ public void readMultiSheetXlsx() throws IOException {
     }
 
     @Test
-    public void testDeleteEmptyColumns() throws Exception {
-        ArrayNode sheets = ParsingUtilities.mapper.createArrayNode();
-        sheets.add(ParsingUtilities.mapper
-                .readTree("{name: \"file-source#Test Sheet 0\", fileNameAndSheetIndex: \"file-source#0\", rows: 31, selected: true}"));
+    public void testDeleteBlankColumns() throws IOException {
```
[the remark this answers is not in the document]
smierz: I'm positive that the renaming of a test is manageable for reviewers, given that I also elaborated on it and its intent in a separate comment -> https://github.com/OpenRefine/OpenRefine/pull/7272#issuecomment-2862686255

modules/core/src/main/java/com/google/refine/importers/ImporterUtilities.java
smierz: thanks for the explanation, but I'm already familiar with the difference between boolean/Boolean and accessing elements in arrays and lists

modules/core/src/main/java/com/google/refine/importers/ImportingParserBase.java
smierz: thanks for the suggestions, but that is not what I had in mind. 
I was already in the process of rewriting the method to make it more reusable by using a list of Integers representing the column indices (see comment https://github.com/OpenRefine/OpenRefine/pull/7272#discussion_r2099955035).
--> see commit "[optimize determination of blank columns](https://github.com/OpenRefine/OpenRefine/pull/7272/commits/7005d8bfbb639042991195a3490108d30c034197)"

there are also some concerns I have with the suggested moving of indices:
* imo this is not very readable/maintainable which seems key as you already commented on having past problems with this part of the codebase
* the moving of the indices relies on a specific setup of the file, e.g. if i have lines like this (empty, data, data, empty) the complete row will continually be checked, even though the inner columns were determined to contain data
* once all indices have been determined to have data in them, there should be a break-out of the loop
* tbh I'm not sure, why an array of booleans would improve the processing, the list of Booleans contains only of references to the static members of the Boolean class, so there is no object creation overhead

let me know what you think of the improved implementation in the linked commit, I would say it improves readibility and performance quite a bit

modules/core/src/main/java/com/google/refine/importers/ImporterUtilities.java
smierz: update: i went ahead and implemented the suggested changes (https://github.com/OpenRefine/OpenRefine/pull/7272/commits/7005d8bfbb639042991195a3490108d30c034197) . I think they improve the code quite a bit, explanations in this review discussion https://github.com/OpenRefine/OpenRefine/pull/7272#discussion_r2104364044 

Let's continue the conversation in the linked thread, so reviewers don't need to jump between two threads, I will therefore mark this conversation as resolved, hope that is ok.

modules/core/src/main/java/com/google/refine/importers/ImporterUtilities.java
smierz: * a complete extra pass is only in worst case, as `allMatch` will short circuit
* added it, because in case there is no column to delete, the `ColumnModel.update` method would still be called twice adding unnecessary calculations

smierz:
hello reviewers,
it looks like this PR is getting quite lengthy, that's why I would suggest that we divide the concerns into different steps, so we move towards a common goal :)

from my point of view these concerns were addressed/handled (correct me, if you disgree):
* adding an extra loop instead of modifying the `options`
* the processing functionality is correct and placed in the right files
* method modifiers are correct 
* docs/comments are correct and sufficient
* existing tests address the bugs described in the issue


these points are currently in discussion/open (correct me, if you disgree):
* determining blank columns could be optimized for performance/reusability
* additional test removing first column 
* too many perceived changes in tests

I'd suggest we concentrate on the optimization for determining blank columns for now (->[here](https://github.com/OpenRefine/OpenRefine/pull/7272#discussion_r2103412220)) and once we agree on this, move down the list, wdyt?

SoryRawyer:
I think this looks good! I'll try to respond to the remaining points individually:
- I think the performance here is acceptable; if there are additional performance gains to be had, we can address them in a follow up pull request.
- I think it's worthwhile to include an additional test case to check when the first column is blank.
- Regarding the changes to tests: I have the same opinion that test cases should change as little as possible, but I think that's a point we can address in the contributor guidelines.

To summarize, I think only the middle point is resolvable in this PR, and the other two can be handled outside of this PR. Thank you again for sticking with this PR, @smierz! It'll be really nice to have this merged and resolve the two linked issues.

tfmorris:
Sorry for the delay in replying. 

- My comment about test changes was feedback for the future, not a request for changes, since I'd already invested the time to figure out that they were innocuous, so it can be ignored for the purposes of this PR.
- We don't have good statistics on the shape of user projects, so it's kind of pointless to argue over performance optimizations, since we can't do it from a data driven point of view, but I'm happy to go with your implementation.
- I'd still like to see a test case for the a blank first column.

modules/core/src/main/java/com/google/refine/importers/ImportingParserBase.java
tfmorris: > once all indices have been determined to have data in them, there should be a break-out of the loop

Of course. Good catch!

> let me know what you think of the implementation in the linked commit, I would say it improves readibility while addressing the performance concerns

I'm fine with using that.

smierz:
great, it seems we are all on the same page regarding the functionality, so I'm going to resolve the conversation and move on to the tests.

Thanks all for the valuable feedback so far!

smierz:
Update:  **_added test that deletes columns from different index positions_** (first, last, middle).
The easiest way to implement this was to use the `FixedWidthImporter` with a `columnWidth` of 0 at different index positions which results in an empty column at the position.
-> commit "[add test deleting blank cols from different positions](https://github.com/OpenRefine/OpenRefine/pull/7272/commits/c5151c5d07135662318fb3ca6ef771cd490057ad)"

smierz:
Update: _**restored the original test cases**_ ( for easier diff between this branch and master )
* to re-add them I renamed the "testDeleteBlankColumns" tests that I originally added by simply adding a suffix about the column position ->  [rename test to include column position](https://github.com/OpenRefine/OpenRefine/pull/7272/commits/2aa163ce9dba57853804a8876cf7fc1b3f54a8b9)
* all tests were re-added to make diff easier for reviewers -> [restore original tests for deleting blank cols](https://github.com/OpenRefine/OpenRefine/pull/7272/commits/13c2bcb5eb18b877c96ef90f7b3f983732848fb2)


Additionally I made some more options explicit rather than relying on their defaults and added some more comments -> [make more options explicit](https://github.com/OpenRefine/OpenRefine/pull/7272/commits/c589911fe72c61fa38eb04af8a09270dcdd70f40).

---

As far as I can see these were the last commits addressing open change requests (correct me, if I'm wrong).
Additionally we can discuss some code placement questions regarding the tests:
* Leave `options` creation within `SeparatorBasedImporterTests.createOptions` or move up to `ImporterTest` ([discussion](https://github.com/OpenRefine/OpenRefine/pull/7272#discussion_r2098747754))?

* Move `prepareFileRecords` from `SeparatorBasedImporterTests` / `FixedWidthImporterTests` / `ExcelImporterTests` to `ImporterTest` ? 


I would also be fine with moving these discussions into follow-up issues and discussing them there.

main/tests/server/src/com/google/refine/importers/FixedWidthImporterTests.java
```diff
@@ -159,40 +165,93 @@ public void readColumnHeader() throws Exception {
...
-    public void testDeleteEmptyColumns() throws Exception {
-        StringReader reader = new StringReader(SAMPLE_ROW + "\nTooShort");
-
-        ArrayNode columnWidths = ParsingUtilities.mapper.createArrayNode();
-        // Set up blank column in project
-        JSONUtilities.append(columnWidths, 6);
-        JSONUtilities.append(columnWidths, 0);
-        JSONUtilities.append(columnWidths, 5);
-        JSONUtilities.append(columnWidths, 0);
-        JSONUtilities.append(columnWidths, 3);
```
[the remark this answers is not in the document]
smierz: re-added the original tests, so comparison between this branch and master should be easier now
 -> [restore original tests for deleting blank cols](https://github.com/OpenRefine/OpenRefine/pull/7272/commits/13c2bcb5eb18b877c96ef90f7b3f983732848fb2)

main/tests/server/src/com/google/refine/importers/ExcelImporterTests.java
```diff
@@ -420,37 +427,108 @@ public void readMultiSheetXlsx() throws IOException {
...
+        sheets.add(ParsingUtilities.mapper.readTree(
+                String.format("{name: \"%s#Test Sheet 0\", "
+                        + "fileNameAndSheetIndex: \"%s#0\", "
+                        + "rows: 31, "
+                        + "selected: true}",
+                        filename, filename)));
+        JSONUtilities.safePut(options, "sheets", sheets);
+
+        JSONUtilities.safePut(options, "headerLines", 0);
+        JSONUtilities.safePut(options, "skipDataLines", 1);
```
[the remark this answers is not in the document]
smierz: yup, it followed this [discussion](https://github.com/OpenRefine/OpenRefine/pull/7272#discussion_r2098747754).

re-added the original tests and some more options to make the review process easier 
(see [comment](https://github.com/OpenRefine/OpenRefine/pull/7272#issuecomment-2970279224) for more info).

SoryRawyer:
@wetneb, @tfmorris, I think both of you need to approve this PR so that it can be merged (unless there's a way to override GitHub's merge requirements that I'm not aware of).

tfmorris:
Only one of us should be needed. I'll get to it in the next couple of hours. Sorry for the continued delays!

tfmorris (approved):
Looks great. Thanks for the perseverance!

PR #7296, opened 2025-05-07
Proposal: Add JSDoc annotations to JavaScript extension points

Abbe98:
This is a proposal to annotate our frontend extension points with JSDoc annotations. This PR currently illustrates this by annotating the `DialogSystem`.

**Why**

First and for most this would make it clearer to us which interfaces are public, we already have a naming convention and we are good at catching & avoiding breaking changes, however, we are less good at making sure we don't extend these more than intended. By making our interfaces explicit with JSDoc we would avoid this.

Secondly, we will be able to mark JS extension points as deprecated.

Thirdly, JSDoc is supported by a lot of language servers and IDEs which will provide both extension and core developers with inline documentation, hints, and autocomplete features.

Fourthly (but out-of-scope for my own commitment) we could maybe at one point use this to generate documentation for extension developers.

### DIFF
--- a/main/webapp/modules/core/scripts/util/dialog.js
+++ b/main/webapp/modules/core/scripts/util/dialog.js
@@ -42,6 +42,13 @@ var escapeKey = function(event) {
   }
 }
 
+/**
+ * Show a dialog.
+ * @public
+ * @param {jQuery} elmt The dialog element to show.
+ * @param {function} onCancel The function to call when the dialog is dismissed.
+ * @return {number} The layer level of the dialog.
+ */
 DialogSystem.showDialog = function(elmt, onCancel) {
   var overlay = $('<div>&nbsp;</div>')
   .addClass("dialog-overlay")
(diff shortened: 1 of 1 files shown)


main/webapp/modules/core/scripts/util/dialog.js
```diff
@@ -42,6 +42,13 @@ var escapeKey = function(event) {
   }
 }
 
+/**
+ * Show a dialog.
+ * @public
+ * @param {jQuery} elmt The dialog element to show.
```
thadguidry, line 48: This is neat but confusing me.  Does this mean that it takes a jQuery object?  Aren't those simply regular JS objects?  Why not `{object}` ?  Sorry but learning here and wondering if my VSCode could make sense of this better with the right extension...maybe that's my problem here?  Any tips @Abbe98 ?  How's it look for you in VSCode when you hover over that "jQuery"?

main/webapp/modules/core/scripts/util/dialog.js
```diff
@@ -42,6 +42,13 @@ var escapeKey = function(event) {
...
+ * @param {function} onCancel The function to call when the dialog is dismissed.
```
SoryRawyer, line 49: Since this callback is passed through to another function (and I can't find any uses in the project), I think it would be helpful to further describe this parameter using the [callback annotation](https://jsdoc.app/tags-callback). What do you think?

main/webapp/modules/core/scripts/util/dialog.js
Abbe98: This is a great question. The param types are not necessarily native JS types, they can be anything. These can be declared either with an `@external` annotation as mentioned by @SoryRawyer or with `@typedef`. Some very common "types" like jQuery's `$` and DOM "types" like HTMLElement tend to, however, be left alone(some editors/extensions also bundle common `@typedef` declarations).

I'm not very opinionated about how it's done as long as they are resonable named.

Abbe98:
> I think this is a great idea. I also think it would be helpful to have some guidelines over what to annotate and how deeply, but I don't think that needs to get in the way of starting to annotate functions and files.

I totally agree but I think it will be easier to define guidelines when we have used it for a while, especially if we have had some real world feedback from other extension developers.

main/webapp/modules/core/scripts/util/dialog.js
Abbe98: Either that or by writing it like this: `@param {function(int)}`, the callback annotation is a little bit verbose for my taste as it needs an additional block, but it's a matter of taste and I'm happy either way.

SoryRawyer (approved):
Thanks for adding this! I'll approve but wait a day to merge in case this approval spurs any final comments.

tfmorris:
Looks great! One odd thing is that the source branch seems to be in this repo. Was that an oversight? I think we usually how PR branches on our own forks.

tfmorris:
@Abbe98 Ok, thanks. I'll see if we've got this documented somewhere and, if not, start a discussion.

PR #7300, opened 2025-05-12
feat(api): add support for flat array response in get-all-project-tags

### LINKED ISSUE #7295: API: Simplify the response format of `get-all-project-tags`
The response of `get-all-project-tags` is unnecessarily complicated:

```json
{"tags":["test","test2","test3","test4"]}
```

### Proposed solution

It could be simplified to only returning array:

```json
["test","test2","test3","test4"]
```

### Additional context
This is a breaking change and it would impact clients in the wild.

RAD1AN1TE:
feat(api): add support for flat array response in `get-all-project-tags`

Fixes #7295 

This pull request introduces an optional `format=flat` query parameter to the `get-all-project-tags` API endpoint. When specified, the API returns a flat JSON array of tags instead of the default object with a "tags" key.

Example:
```
GET /...?format=flat
Response: ["tag1", "tag2", "tag3"]
```

This enhancement maintains backward compatibility and provides clients with a simplified response format when desired.

### DIFF
--- a/main/src/com/google/refine/commands/workspace/GetAllProjectTagsCommand.java
+++ b/main/src/com/google/refine/commands/workspace/GetAllProjectTagsCommand.java
@@ -57,6 +57,13 @@ public class GetAllProjectTagsCommand extends Command {
 
         Map<String, Integer> tagMap = ProjectManager.singleton.getAllProjectsTags();
         Set<String> tags = tagMap == null ? Collections.emptySet() : tagMap.keySet();
-        respondJSON(response, new AllProjectsTags(tags));
+
+        String format = request.getParameter("format"); // e.g., ?format=flat
+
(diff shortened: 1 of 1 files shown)


SoryRawyer:
Hi @RAD1AN1TE, thanks for your interest in contributing! While this definitely fulfills the proposal put forth in #7295, I'm not sure there's consensus for making this change. I'd like to wait on reviewing this until there's been more discussion on the accompanying issue.

Abbe98 (changes requested):
I don't think that we want to extend the API surface area for a small quality of life improvement.

There is a [current thread](https://forum.openrefine.org/t/proposal-to-add-versioning-to-the-rest-api) on the forum that's highly relevant for how we move #7295 forward.

Issue #7273, opened 2025-04-24
Drag and drop for columns to reorder


magdmartin:
## Feature Description 

Clicking on a column header allows you to drag it to a new position in the grid. This action will be registered as a column reordering in the history.

### Alternatives Considered
Currently, users must either manually move each column left or right or use the “Re-order / Remove Columns” modal. 

### Additional Context
I am creating this ticket based on the [2024 Feature Survey](https://forum.openrefine.org/t/results-from-the-feature-prioritization-survey-2024/1847) and the [Goal Post page](https://openrefine.org/docs/technical-reference/goal-posts#drag-and-drop-for-columns).

magdmartin:
Thanks for looking into it. 
Yes, I believe a design similar to Google Sheets would feel the most natural.

tfmorris:
A significant difference between Google Sheets and OpenRefine is that Google Sheets has fixed column column labels (the letters of the alphabet A, B, C, ...) which aren't selectable, while OpenRefine has user specified column labels which are (were) selectable by clicking on them, which probably means that the Google Sheets design needs to be modified to fit. 

Some other differences:
- Column resize in Google Sheets shows additional bars in addition to changing cursor shape in the resize zone
- Google Sheets keeps the drop down menu affordance/icon hidden until you hover over the header cell making the headers less cluttered
- Google Sheets column proxy/shadow for drag and drop is a full height grey rectangle rather than just the header cell as in the current OpenRefine implementation

I agree with @Abbe98 's [comment](https://github.com/OpenRefine/OpenRefine/issues/6282#issuecomment-1877867229) about the danger of incremental one-of additions from the point of view of consistency and discoverability. I think we should collect all the things that have been requested (menus, resize, rename, reorder, etc) and figure out how they fit into a single holistic design.

Related tickets: 
- #6282 /  #7361 proposes the addition of mouse shortcuts to the header
- column resizing was introduced in #6563. 
- #7510 new ticket for a different solution to drag & drop

SoryRawyer:
I'm okay with reverting the PR to add drag and drop; the feature hasn't been included in a release so the removal would only affect people who are building off of the master branch.
Re @magdmartin's suggestion (on #7345) of looking for a designer, I would appreciate input from someone with expertise in UI/UX design. In the meantime, is this something we could start mocking up in Figma?

Given all this discussion, I think these are reasonable next steps:
- Close #7345, #7510, and #7361
- Revert #7301 
[...]

PR #7301, opened 2025-05-12
feat(ui): enable drag‑and‑drop column header reordering in DataTable view

JyothiSwaroopReddy07:
Fixes https://github.com/OpenRefine/OpenRefine/issues/7273

1. **Adds jQuery UI Sortable to header row** : Users can now drag any data‑column header left or right to rearrange columns directly in the grid.
2. **Keeps special “All” column fixed** : Sortable targets only elements after the first; the star/flag/index header remains immovable.
3. **Protects menu & resize interactions** : Drag initiation is canceled on .column-header-menu buttons and both resize grips, so existing UI controls work unchanged.
4. **Persists order via existing backend operation** : On drop, the new column sequence is sent to /core/reorder-columns; the step is recorded in history for full undo/redo support.
5. **No new dependencies introduced** : Reuses the jQuery UI library already bundled with OpenRefine.
6. **Light‑touch implementation** : Adds a data-col-name attribute when rendering headers, a small helper (_enableHeaderDrag) to wire Sortable, and one call in render().
7. **Column‑width cache & resizing unaffected** : Existing column width caching, resizing logic, and the dialog‑based “Reorder / Remove Columns” remain fully functional.


main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -196,11 +196,11 @@ DataTableView.prototype.render = function() {
     '</div>' +
     '<div bind="dataTableContainer" class="data-table-container">' +
       '<table class="data-table">'+
-        '<colgroup bind="colGroup"></colgroup>'+
+        '<colgroup bind="colGroup"></colgroup>'+
```
SoryRawyer, line 199: Would you mind removing this extra space at the end?

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
```diff
@@ -606,6 +610,34 @@ DataTableView.prototype._renderTableHeader = function(tableHeader, colGroup) {
   }
 }
 
+// --- NEW: enable drag-and-drop using jQuery UI Sortable ---
```
SoryRawyer, line 613: I think the comment about enabling drag-and-drop is useful, but could you remove the `NEW: ` from the comment?

SoryRawyer (changes requested):
This looks great, thank you for contributing! I tried it out and everything worked as expected. I just have a few tiny nit-picks about the code style and then we're good to merge this.

JyothiSwaroopReddy07:
Sure sir, will remove the extra spacing. i also wanted to add testcases.

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
JyothiSwaroopReddy07: done

main/webapp/modules/core/scripts/views/data-table/data-table-view.js
JyothiSwaroopReddy07: done

SoryRawyer (approved):
I think this looks good! I'll approve and let the CI checks run and merge if everything succeeds.

PR #7305, opened 2025-05-17
docs: add jsdoc annotations to CSRF utilities

Abbe98:


main/webapp/modules/core/scripts/project.js
```diff
@@ -39,7 +39,23 @@ var Refine = {};
...
+ * @public
+ * @param {function(string)} onCSRF - The callback to call with the token.
+ */
 Refine.wrapCSRF = CSRFUtil.wrapCSRF;
+
+/**
+ * Performs a POST request where an additional CSRF token is automatically attached.
+ * @public
+ * @param {string} url - The URL to send the request to.
+ * @param {data} data - The data to send with the request.
```
SoryRawyer, line 53: Looking at the implementation of `CSRFUtil.wrapCSRF`, it looks like this could be narrowed down to `{(Object|string)}`.

SoryRawyer (commented):
I think it might make more sense to add these annotations to the definitions themselves in `csrf.js`. What do you think?

Abbe98:
> I think it might make more sense to add these annotations to the definitions themselves in csrf.js. What do you think?

 I don't think it's possible to describe `Refine.*` from `csrf.js` although it's there the implementation is. What we could maybe do is to move the `Refine.wrapCSRF = CSRFUtil.wrapCSRF;`, etc to the CSRF utilities?

tfmorris:
I'm not much of a front end person, but I think it'd be better to have the docs co-located with the implementation. It's too hard to verify with then in different places.

Abbe98:
> I'm not much of a front end person, but I think it'd be better to have the docs co-located with the implementation. It's too hard to verify with then in different places.

See my reply to @SoryRawyer, the alternative is to rename `CSRFUtil` so that the annotations can be added directly but I would like to avoid changes that could have side effects.

tfmorris:
OK. How about it we also add the jsdoc to `csrf.js` ? Then a copy of it will be close to the implementation and we'll hopefully notice if they diverge.

Moving the definition of `Refine.wrapCSRF` to `csrf.js` doesn't seem right to me.

tfmorris (approved):
Prospectively approving based on suggested change to add the jsdoc to csrf.js as well.

Abbe98:
> OK. How about it we also add the jsdoc to csrf.js ? Then a copy of it will be close to the implementation and we'll hopefully notice if they diverge.

I would then add it without the `@public` flag and include an additional pointer to the `Refine.*` variant.

PR #7310, opened 2025-05-19
test: add regression test for AddColumnCommand parameter parsing

amansingh-swe:
Add regression unit test for AddColumnCommand parameter parsing

Description:
This PR adds a unit test for the AddColumnCommand#createOperation() method to ensure correct parsing and mapping of HTTP request parameters when creating a ColumnAdditionOperation. The test verifies that:
1. baseColumnName, expression, newColumnName, and columnInsertIndex are accurately extracted.
2. onError mode string is properly interpreted and passed as an enum.

It safeguards against future changes to request handling logic that might break column addition functionality


main/tests/server/src/com/google/refine/commands/column/AddColumnCommandTests.java
```diff
@@ -0,0 +1,56 @@
+
+package com.google.refine.commands.column;
+
+import static org.mockito.Mockito.*;
+import static org.testng.Assert.*;
```
SoryRawyer, line 5: In general, it's best to avoid importing everything in a package like this. For example, it seems like `when` is the only item we need from mockito here, so we could replace this import statement with `import static org.mockito.Mockito.when;`.

main/tests/server/src/com/google/refine/commands/column/AddColumnCommandTests.java
```diff
@@ -0,0 +1,56 @@
...
+import org.testng.annotations.Test;
+
+import com.google.refine.browsing.Engine;
+import com.google.refine.browsing.EngineConfig;
+import com.google.refine.model.AbstractOperation;
+import com.google.refine.model.Project;
+import com.google.refine.operations.OnError;
+import com.google.refine.operations.column.ColumnAdditionOperation;
+
+public class AddColumnCommandTests {
```
SoryRawyer, line 21: There's a base class called `CommandTestBase` that the command test classes are expected to extend. This would also remove the need for line 31, where you initialize `request`.

SoryRawyer (changes requested):
Hi @amansingh-swe, thanks for taking the time to improve OpenRefine's test coverage and prevent future breaks to the API for this command. I have a couple small suggestions but overall I think this looks good.

amansingh-swe:
Hi @SoryRawyer, Thanks for the review. I have made the AddColumnCommandTests extend the base class CommandTestBase and also addressed the static imports.
Changes Pushed. Let me know if there is anything else you would like readjusted

amansingh-swe:
Hey @SoryRawyer ,

Sorry for the confusion for some reason Intellij did not push my commit to the branch. I have synched the fork and made a new commit. I have checked it twice and things should be fine now.

Thank you again for your time and sorry for the confusion.

Thanks,
Aman

main/tests/server/src/com/google/refine/commands/column/AddColumnCommandTests.java
```diff
@@ -0,0 +1,61 @@
...
+public class AddColumnCommandTests extends CommandTestBase {
+
+    private AddColumnCommand command;
+    private Project project;
+    private EngineConfig engineConfig;
+
+    @BeforeMethod
+    public void setUpCommand() {
+        try {
+            setUpRequestResponse();
```
SoryRawyer: I think you can get rid of this call to `setUpRequestResponse` since it's also annotated with `@BeforeMethod` in `CommandTestBase`. Can you try removing that and the try/catch block and see if the test still succeeds?

main/tests/server/src/com/google/refine/commands/column/AddColumnCommandTests.java
```diff
@@ -0,0 +1,61 @@
...
+            command = new AddColumnCommand();
+            project = new Project();
+            engineConfig = new EngineConfig(Collections.emptyList(), Engine.Mode.RowBased);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Test
+    public void testCreateOperationWithValidParameters() throws Exception {
```
SoryRawyer, line 36: Could you be more specific about which exception(s) might be thrown?

amansingh-swe:
Hi @SoryRawyer,

Appreciate the feedback. I have made the required changes in commit [68a7f95](https://github.com/OpenRefine/OpenRefine/pull/7310/commits/68a7f956f0fd07938f05997fe1f5eae71c427e24). 

I have removed the setUpRequestResponse call from AddColumnCommandTests. Intellij was acting up and refusing to build. I probably had it in to debug.

main/tests/server/src/com/google/refine/commands/column/AddColumnCommandTests.java
amansingh-swe: For this The Exception is thrown from AddColumnCommand.createOperation().

` protected AbstractOperation createOperation(Project project,
            HttpServletRequest request, EngineConfig engineConfig) throws Exception`

This exception is probably redundant and was put in to handle errors on

`TextTransformOperation.stringToOnError(onError)`

But the exceptions are already checked in the stringToOnError function.

`static public OnError stringToOnError(String s) {
        if ("set-to-blank".equalsIgnoreCase(s)) {
            return OnError.SetToBlank;
        } else if ("store-error".equalsIgnoreCase(s)) {
            return OnError.StoreError;
        } else {
            return OnError.KeepOriginal;
        }
    }`

SoryRawyer: Ah sorry, I didn't see that this was inherited from `EngineDependentCommand::createOperation`'s signature. That makes sense!

SoryRawyer (approved):
Looks great! Thanks for sticking with it even with my pedantic reviews 😅

PR #7311, opened 2025-05-19
Refine tests

JyothiSwaroopReddy07:
Added new tests for Refine Class

### DIFF
--- a/modules/core/src/test/java/com/google/refine/RefineTest.java
+++ b/modules/core/src/test/java/com/google/refine/RefineTest.java
@@ -393,6 +393,43 @@ public class RefineTest {
         return runOperation(operation, project, -1);
     }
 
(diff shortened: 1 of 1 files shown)


SoryRawyer:
Hi @JyothiSwaroopReddy07, I appreciate the initiative to try and improve test coverage of OpenRefine. However, I'm not sure these tests meaningfully improve the confidence in our tests. The methods you've added tests for are themselves helper methods for tests. I can't speak to the entirety of the OpenRefine codebase, but my personal feeling is that test code doesn't always need to be tested itself.

SoryRawyer:
Also, I noticed that your changes for the drag and drop feature are also included in this PR. In the future, please separate your changes such that each PR focuses on one self-contained improvement.

PR #7316, opened 2025-05-21
ExporterManager: add an explicite extension point

Abbe98:
Prior to this change one would extend the export menu by pushing directly to an array. This comes with several issues:

 - The interface isn't actually controlled by us and is incredibly large.
 - We can't document the interface in a structured way(with for example JSDoc).
 - Extensions ability to modify the menu isn't limited.

This change does not break existing interfaces as the underlying array  is untouched. Some extensions do more than pushing to the array, however, that could be considered a side effect of the interface rather than an intended feature. If there are legitime use-cases beyond pushing additional extension points could be added.

### DIFF
--- a/main/webapp/modules/core/scripts/project/exporters.js
+++ b/main/webapp/modules/core/scripts/project/exporters.js
@@ -105,6 +105,29 @@ ExporterManager.prototype._initializeUI = function() {
   });
 };
 
+/**
+ * Adds a menu item to the end of the ExporterManager menu.
+ * @public
+ * @param {string} id - The item ID.
+ * @param {string} label - The item Label.
+ * @param {function} click - The function which is called upon an item click.
+ */
+ExporterManager.addMenuItemAfter = function(id, label, click) {
+    ExporterManager.MenuItems.push({
+        'id': id,
+        'label': label,
+        'click': click,
+    });
+}
+
+/**
+ * Adds a separator item to the end of the menu.
+ * @public
+ */
+ExporterManager.addMenuSpacerAfter = function() {
+    ExporterManager.MenuItems.push({});
+}
+
 ExporterManager.stripNonFileChars = function(name) {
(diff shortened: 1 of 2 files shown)


Abbe98:
PS this should not be squash merged.

main/webapp/modules/core/scripts/project/exporters.js
```diff
@@ -105,6 +105,18 @@ ExporterManager.prototype._initializeUI = function() {
   });
 };
 
+/**
+ * Adds a menu item to the ExporterManager menu.
+ * @public
+ * @param {object} item The menu item to add.
+ */
+ExporterManager.prototype.addMenuItem = function(item) {
+    // NOTE: while taking id, label, and click induvidually would
```
tfmorris, line 114:
```suggestion
    // NOTE: while taking id, label, and click individually would
```

Typo, but more importantly, at the risk of over engineering something trivial, perhaps we could address the comment by having two separate functions to make things explicit? e.g. `addMenuItem(id, label, click)` & `addMenuSpacer()` or something similar

tfmorris (changes requested):
I didn't investigate, but I suspect that the id typo is the reason for the e2e test failure.

main/webapp/modules/core/scripts/project/exporters.js
Abbe98: Yeah, I wanted to avoid new things and focus on documentation, but given that this is new no matter what, maybe I should go ahead and go straight for two new? From an extensions point of view the difference is minimal anyway.

tfmorris: That would be my suggestion, unless @SoryRawyer or @wetneb have a different idea.

SoryRawyer:
@tfmorris, do you have any additional comments? I think your approval might be needed here before we're able to merge.

PR #7317, opened 2025-05-22
Refetch reconcile service manifest for new changes (#3645) (reopened)

### LINKED ISSUE #3645: Double check the reconciliation manifest each time reconciliation is fired.

akashinde:
Fixes https://github.com/OpenRefine/OpenRefine/issues/3645

This is reopened PR for previously created PR https://github.com/OpenRefine/OpenRefine/pull/6881

Hi @wetneb,

Thanks for waiting, I have made changes according to your review from last PR https://github.com/OpenRefine/OpenRefine/pull/6881#pullrequestreview-2411282925.

As requested, following changes have been done:
- Replaced the existing AJAX calls in `ReconDialog` with a new `fetchServiceManifest` method in `ReconciliationManager` to handle both CORS and JSONP requests.
- Updated the `registerStandardService` method to utilize the new fetching logic.

### DIFF
--- a/main/webapp/modules/core/scripts/reconciliation/recon-manager.js
+++ b/main/webapp/modules/core/scripts/reconciliation/recon-manager.js
@@ -62,6 +62,48 @@ ReconciliationManager.registerService = function(service) {
   return ReconciliationManager.customServices.length - 1;
 };
 
+/**
+ * Fetches a reconciliation service manifest from the given URL.
+ * 
+ * @param {string} url - The URL of the reconciliation service manifest to fetch
+ * @param {Function} successCallback - Callback function called on successful fetch
+ * @param {Function} errorCallback - Optional callback function called if both fetch attempts fail
+ * @param {Function} alwaysCallback - Optional callback function called regardless of success or fail
+ * @returns {void}
+ */
+ReconciliationManager.fetchServiceManifest = function(url, successCallback, errorCallback, alwaysCallback) {
+  // First, try with CORS (default "json" dataType)
+  $.ajax(
+    url,
+    { 
+      "dataType" : "json",
+      "timeout": 5000
+    }
+  )
+  .done(function(data, textStatus, jqXHR) {
+    successCallback(data, "json");
+  })
+  .fail(function(jqXHR, textStatus, errorThrown) {
+    // If it fails, try with JSONP
+    $.ajax(
+      url,
+      { 
+        "dataType" : "jsonp",
+        "timeout": 5000
+      }
+    )
+    .done(function(data, textStatus, jqXHR) {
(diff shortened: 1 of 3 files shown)


Abbe98:
Could you separate the formatting changes form the others?

main/webapp/modules/core/scripts/reconciliation/recon-manager.js
```diff
@@ -99,34 +141,16 @@ ReconciliationManager.registerStandardService = function(url, f, silent) {
...
-        }
-    });
-  });
+    registerService,
+    function(jqXHR, textStatus, errorThrown) {
+      if (!silent) {
+        dismissBusy(); 
+        alert($.i18n('core-recon/error-contact')+': ' + textStatus + ' : ' + errorThrown + ' - ' + url);
+      }
+    }
```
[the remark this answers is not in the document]
tfmorris: The suggestion addresses code which was untouched by the PR, so I would ignore it for now. A more general review of where `dismissBusy()` is called might be called for, but I don't think this is the place to do it.

main/webapp/modules/core/scripts/reconciliation/recon-manager.js
```diff
@@ -146,6 +170,33 @@ ReconciliationManager.unregisterService = function(service, f) {
...
 };
 
+/**
+ * Re-registers a service.
+ * 
+ * @param {Object} service - The service to re-register
+ * @param {Function} f - Optional callback function to execute after re-registering
+ * @param {boolean} silent - If true, suppresses any error alerts
+ */
+ReconciliationManager.reRegisterService = function(service, f, silent) {
```
Abbe98, line 180: Let's name `f` something other than `f`, `callback` is a common name for such a parameter.

main/webapp/modules/core/scripts/reconciliation/recon-manager.js
```diff
@@ -99,34 +141,16 @@ ReconciliationManager.registerStandardService = function(url, f, silent) {
...
-          dismissBusy(); 
-          alert($.i18n('core-recon/error-contact')+': ' + textStatus + ' : ' + errorThrown + ' - ' + url);
-        }
-    });
-  });
+    registerService,
+    function(jqXHR, textStatus, errorThrown) {
+      if (!silent) {
+        dismissBusy(); 
+        alert($.i18n('core-recon/error-contact')+': ' + textStatus + ' : ' + errorThrown + ' - ' + url);
```
Abbe98, line 150: Let's avoid `alert` and use our inbuilt error modal instead.

Abbe98 (changes requested):

PR #7332, opened 2025-06-07
multi language support

### LINKED ISSUE #7330: Improve stop word handling - customizable and/or multi-language

- visible
- user customizable

### Proposed solution

### Alternatives considered

### Additional context

JyothiSwaroopReddy07:
Fixes #7330

* Refactored `StandardReconConfig.java` to support language-specific stop word lists.
* Added a new mechanism to **load stop words from `conf/stopwords.json`**, making the list:

  *  **Language-specific** (based on ISO 639-1 codes, e.g., `en`, `fr`, `zh`)
  * **User-customizable** without changing the code
* Ensured fallback behavior:

  * If a language’s stop word list is missing, it defaults to `"en"`
  * If `"en"` is also missing, it falls back to a hardcoded English stop word list
* Improved the `breakWords()` method to apply stop word filtering dynamically based on system locale or available mappings
* Included support for 11 languages with expanded stop word sets:

  * English, French, Spanish, German, Italian, Portuguese, Dutch
  * Chinese, Japanese, Hindi, Korean
* Added robust error handling and debug logging for easier testing and maintenance

### DIFF
--- a/conf/stopwords.json
+++ b/conf/stopwords.json
@@ -1,0 +1,50 @@
+{
+  "en": [
+    "the", "a", "an", "and", "or", "of", "to", "in", "on", "for", "with",
+    "at", "by", "from", "about", "as", "into", "like", "through", "after",
+    "over", "between", "out", "against", "during", "without", "before", "under",
+    "around", "among"
+  ],
+  "fr": [
+    "le", "la", "les", "un", "une", "et", "ou", "de", "du", "des", "pour",
+    "par", "dans", "en", "sur", "avec", "sans", "entre", "avant", "après",
+    "contre", "chez", "sous", "vers"
+  ],
+  "es": [
+    "el", "la", "los", "las", "un", "una", "y", "o", "de", "del", "para",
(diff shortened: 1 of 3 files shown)


SoryRawyer:
Hi @JyothiSwaroopReddy07, thanks for taking a look at this issue! After reading through the issue and your pull request, I'm not sure that the list of stop words in this PR matches the "user-editable" expectation in the issue. My interpretation of that was that users should be able to edit this list through the UI rather than updating a config file. It could be that including a configuration file as a starting point is a good solution, but I think this will be inconvenient to update once the application is packaged and distributed. I asked for clarification on this in the issue, so my apologies if I've misinterpreted the original request.

Additionally, it would be helpful to resolve the lint errors so that there's less noise for reviewers.

SoryRawyer:
@JyothiSwaroopReddy07 given the lack of activity on this PR, I'm going to close it out. We can always reopen if and when it becomes active again. Thanks for your understanding!

PR #7337, opened 2025-06-09
ETA Functionality for Frontend

### LINKED ISSUE #7318: Add duration and estimated completion time for long running operations
When performing long long-running operation, it would be nice to know
* for how long the operation has been running
* Estimate how long is left before completion.

### Proposed solution
The information can be added in the yellow banner showing the percentage progress.  

Image

amansingh-swe:
Fixes #7318 .

Add a ETA for long running operations like reconciliation. I have done a rough draft of how it might look like. I have just added an additional eta property in LongRunningProcess which can be calculated based on "done" tasks in

### DIFF
--- a/main/src/com/google/refine/operations/recon/ReconOperation.java
+++ b/main/src/com/google/refine/operations/recon/ReconOperation.java
@@ -291,6 +291,7 @@ public class ReconOperation extends EngineDependentOperation {
 
             int batchSize = _reconConfig.getBatchSize(_project.rows.size());
             int done = 0;
+            long startTime = System.currentTimeMillis();
 
             List<CellChange> cellChanges = new ArrayList<CellChange>(_entries.size());
             List<JobGroup> groups = new ArrayList<JobGroup>(jobKeyToGroup.values());
@@ -343,6 +344,11 @@ public class ReconOperation extends EngineDependentOperation {
                 }
 
                 _progress = done * 100 / groups.size();
+                double elapsedTime = System.currentTimeMillis() - startTime;
+                double remaining = (elapsedTime * 100 / _progress) - elapsedTime;
+                _eta = (int)Math.floor(Math.round(remaining / 1000.00) / 60.0);
+                logger.info(String.valueOf(_eta));
+
                 try {
                     Thread.sleep(50);
                 } catch (InterruptedException e) {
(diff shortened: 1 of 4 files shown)


SoryRawyer:
Thanks for taking a look at this! A couple notes before I get into the ETA calculation itself:
1. It looks like you've got your regression test changes included in this PR, which I don't think is intended.
2. Since this really just relies on `_progress` and the start timestamp, maybe this could be moved into a method on `LongRunningProcess` itself that subclasses could call during their `run` function. What do you think?
3. It looks like there are some linting issues, you can find these locally by running `./refine lint` in the root project directory.

amansingh-swe:
Hi @SoryRawyer,

Sorry for the hodgepodge of a pull request but I had an assignment due.

1. Here's what I am thinking.
2 functions in the LongRunningProcess class.
startProcess(): to set the startTime in the LongRunningProcess class. This will also abstract declaring startTime in the child classes
updateProcess(int done, int total): To update the _progress and then calculate the ETA.

Currently, _progress is set in the child classes directly accessing the _progress property from the parent and I am not sure if that is good design. For example in ReconOperation the code looks something like this:

_progress = done * 100 / groups.size();

This could be converted to 

UpdateProcess(done, groups.size()) 

which can  calculate the percent and the eta internally. 

Let me know what you think.

2. Maybe this pull request could just be backend changes? And I can do a new Pull request to update the frontend?

Let me know what you think about this

SoryRawyer:
1. I think that sounds like a good idea.
2. My hope is that the frontend changes won't be too large, so I think it makes sense to include them as part of this PR as well. Would you mind posting a screenshot of what your frontend changes look like?

amansingh-swe:
Hi @SoryRawyer ,

Totally forgot about this.
PFA screenshot of expected output.

PS I might need to make a new pull request after learning how to use git but mostly the everything else should look the same

image

SoryRawyer:
Hi @amansingh-swe, apologies for such a delayed response, but I just wanted to follow up and say that I think the screenshot you attached looks good. Please feel free to reach out if you have any questions about the code, git, or anything else related to this PR. I hope you're doing well!

SoryRawyer:
Hi @amansingh-swe, I'm going to close this PR since it's been a few months since the last activity. We can always reopen this if and when you're ready to work on it again.

PR #7345, opened 2025-06-19
Re-enable text selection on column headers

### LINKED ISSUE #7338: It's no longer possible to select column header text
### To Reproduce
In the project view try selecting a column header with the mouse.

Alternatively double click the header to attempt to select it.  

### Current Results
You end up draggning the column header.

### Expected Behavior
Users should be able to easily select the header name so that it can be copied.

### Versions
 - master branch

SoryRawyer:
Fixes #7338

- Adds a class name to the jQuery sortable element which would allow users to select the text. Previously, this capability was overridden by the drag and drop functionality provided by the sortable call.
- Adds a Cypress test to ensure that drag and drop still works (the `dragAndDrop` Cypress function was updated to be able to specify where on the source element to click; the default is `center` which is now selectable text and is no longer a valid drag and drop click target)

I wanted to add a test to ensure that the text was selectable, but several different methods of simulating text highlighting were not properly testing this interaction so I decided to open the pull request without such a test.

### DIFF
--- a/main/tests/cypress/cypress/e2e/project/grid/all-column/edit-columns.cy.js
+++ b/main/tests/cypress/cypress/e2e/project/grid/all-column/edit-columns.cy.js
@@ -20,6 +20,20 @@ describe(__filename, function () {
       ['BUTTER,WHIPPED,WITH SALT', '01002', '717', '15.87'],
     ]);
   });
+
+  it('Ensure columns can be reordered from the project grid via drag and drop', function () {
+    cy.loadAndVisitProject('food.mini');
+
+    cy.dragAndDrop('th[data-col-name="Shrt_Desc"]', 'th[data-col-name="NDB_No"]', 'top', 'top');
+    cy.dragAndDrop('th[data-col-name="Energ_Kcal"]', 'th[data-col-name="Water"]', 'top', 'top');
+
+    cy.assertGridEquals([
+      ['Shrt_Desc', 'NDB_No', 'Energ_Kcal', 'Water'],
+      ['BUTTER,WITH SALT', '01001', '717', '15.87'],
+      ['BUTTER,WHIPPED,WITH SALT', '01002', '717', '15.87'],
+    ]);
+  });
+
   it('Ensure columns are removed using remove column', function () {
     cy.loadAndVisitProject('food.mini');
     cy.columnActionClick('All', [
--- a/main/tests/cypress/cypress/support/commands.js
+++ b/main/tests/cypress/cypress/support/commands.js
@@ -324,8 +324,10 @@ Cypress.Commands.add(
  * targetSelector - jquery selector for the element to be dropped on
  * position - position relative to the target element to perform the drop
  */
-Cypress.Commands.add('dragAndDrop', (sourceSelector, targetSelector, position = 'center') => {
-  cy.get(sourceSelector).trigger('mousedown', { which: 1 });
+Cypress.Commands.add(
+  'dragAndDrop',
+  (sourceSelector, targetSelector, position = 'center', sourcePosition = 'center') => {
+  cy.get(sourceSelector).trigger('mousedown', sourcePosition, { which: 1 });
 
   cy.get(targetSelector) // eslint-disable-line
     .trigger('mousemove',position)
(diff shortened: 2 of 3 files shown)


SoryRawyer:
@Abbe98 are you still able to take a look at this?

SoryRawyer:
Hey @DevangJagdale, would you be interested in reviewing this PR? Given your recent contributions I think your feedback would be valuable!

DevangJagdale:
Hi! I’m new to open source and also new to OpenRefine, so this is my first review 🙂

From my testing, everything looks good to me functionally, the fix works as expected and I can now select text in the column headers. 

For cypress test I was thinking can we do this to simulate human text selection behavior, it looks similar to human behavior
what do you think?

`
it('Ensure column header text can be selected', function () {
    cy.loadAndVisitProject('food.mini');

    cy.get('th[data-col-name="Shrt_Desc"]').then(($header) => {
      const el = $header[0];

      cy.window().then((win) => {
        const range = win.document.createRange();
        range.selectNodeContents(el);

        const sel = win.getSelection();
        sel.removeAllRanges();
        sel.addRange(range);

        const selectedText = sel.toString().trim();

        // Assert header text is selected
        expect(selectedText).to.equal('Shrt_Desc');
      });
    });
  });

` 

One small UX suggestion: it might be nice if the mouse cursor changed to a hand icon when selecting column header to reorder, and to a grab hand icon when dragging to reorder columns. That would make the interaction more intuitive.

Other than that, everything looks great

SoryRawyer:
Thanks for the review! I think the icon change might be out of scope for this change, which is just to restore functionality that was lost in a recent change. As for the test, I tried the `addRange` approach but never got it to reliably test that text is still selectable (I can't find exactly what the issue was in my notes, but my memory of it is that this test still passed in the broken state).

DevangJagdale:
Thanks for clarifying! That makes sense regarding the icon change being out of scope, I agree it’s better handled separately so this PR stays focused on restoring the original functionality.

On the Cypress test: I see what you mean about the addRange approach. If it was still passing when the behavior was broken, then it’s not a reliable regression test. That’s really helpful context for me as I’m still learning the best way to simulate real world text selection in Cypress.

we could use realMouseDown() / realMouseMove() from cypress-real-events to simulate a more human-like drag selection, but from what I’ve seen it can still be flaky in CI.

Either way, I learned a lot from this PR and your feedback, so thanks!

tfmorris (commented):
I can confirm that column header text selection works again, but I'm not loving the discoverability of this feature. It wasn't great in the original implementation and it's worse now that the targets are smaller (and potentially VERY small). I missed the fact that the icon references in the earlier discussion were talking about cursor shapes.

Currently if the column name text fills the column header, you have:
- resize handles
- pulldown column menu
- header text

all with their own cursor shapes in a very small area. The even smaller leftover area uses the default cursor and is the only place a user can grab for their drag and drop operation.

At a minimum, I think we need to introduce a `move` cursor shape for the draggable areas, but we may actually want to consider adding some type of drag handle affordance that the user will visually recognize without having to mouse over it.

I'm OK with leaving that cleanup to a separate PR if that works the best, but I don't think we should release the feature without improving the UX.

p.s. I have no clue on the Cypress questions

SoryRawyer:
I think that makes sense, but I feel that fits best as a follow-up task. I haven't looked into the existing SVGs we already have in the project, but it's possible we already have a suitable icon to indicate (and provide space for) the ability to reorder columns by dragging them.

SoryRawyer:
@tfmorris I created an issue to discuss the UX concerns here: https://github.com/OpenRefine/OpenRefine/issues/7510
I'd like to use this PR to restore the text selection behavior that was lost with the addition of drag and drop.

tfmorris:
I think it's arguable whether #7273 is actually complete. Perhaps rather than creating a new issue, we should just reopen the original. As a corollary, rather than patching the current implementation, we should consider reverting it until we can revisit the implementation of it, and all the other features requested for column headers, holistically.

magdmartin:
@tfmorris @SoryRawyer do you think it is worth looking for UI / UX designer to help with the design and different interactions we want to implement? I can help advertise this as a mini project with designer interested in OpenRefine or OSS.

SoryRawyer:
Closing following the discussion on #7273

Issue #2741, opened 2020-06-16
Support more compression & archive formats


tfmorris:
Now that we're using Apache Commons Compress we are able to cheaply support additional compression and packaging formats like 7Zip and LZMA. 

Using Apache Tika instead would also give us access to a RAR parser: https://tika.apache.org/3.2.3/api/org/apache/tika/parser/pkg/RarParser.html

PR #7353, opened 2025-06-25
7314 gzip import

### LINKED ISSUE #7314: Non-archive gzipped files can't be imported any more
Since #6593 was merged, it's no longer possible to import single gzipped files since OpenRefine assumes anything gzipped is an archive of some type.

### To Reproduce
1. Attempt to create a project from a gzipped CSV (or other supported file type)

### Current Results
Blank preview screen with no error or any other feedback

### Expected Behavior
gzipped file contents is show in preview

### Versions
 - Operating System: 
 - Browser Version: 
 - JRE or JDK Version: 
 - OpenRefine: 3.9

tfmorris:
Fixes #7314

- Fix inverted parameter ordering for tests
- Fix regression for single file archives and refactor to make better use of Apache Commons Compress
- Add support for 7zip and other xz related formats


SoryRawyer:
@tfmorris - is there anything I can help out with to get this ready for review? I'd still like to include this in the 3.10 beta.

tfmorris:
@SoryRawyer I won't get to it today, but I'll put it on my list for tomorrow. The new functionality is complete, but needs test coverage. I think I've collected all the example test data that I need, but have to integrate it into the tests.

tfmorris:
It looks like we won't actually get 7zip support for "free" after all, since the Apache Compress implementation doesn't support streaming, which we currently require (although we want to remove this limitation to support the Parquet importer as well). On the plus side, the delay has meant that we picked up support for Zstandard with the latest Apache Compress release. I'll add a test for that and call it done.

tfmorris:
Sorry for the delay! I've removed all the added functionality and just focused this on fixing the existing bug, along with the core refactoring to use more of Apache Compress.

The addition of 7z, xz, Zstd, etc has been moved to #7428 which was started to address #2741, the issue that introduced the requirement for additional formats.

This is ready for review

Issue #7354, opened 2025-06-25
Clean up build warnings - particularly for annotation processing


tfmorris:
Our build has accumulated a number of warnings about things which aren't fatal (yet), but should be fixed. The first of these finally bit us due to Log4J taking a somewhat aggressive stance concerning their annotation processor. 

1. Annotation Processing (blocks #7349)

> Annotation processing is enabled because one or more processors were found
  on the class path. A future release of javac may disable annotation processing
  unless at least one processor is specified by name (-processor), or a search
  path is specified (--processor-path, --processor-module-path), or annotation
  processing is enabled explicitly (-proc:only, -proc:full).
  Use -Xlint:-options to suppress this message.
  Use -proc:none to disable annotation processing.

2. Dynamic Java Agent loading

> Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3
WARNING: A Java agent has been loaded dynamically (/Users/tfmorris/.m2/repository/net/bytebuddy/byte-buddy-agent/1.17.5/byte-buddy-agent-1.17.5.jar)
WARNING: If a serviceability tool is in use, please run with -XX:+EnableDynamicAgentLoading to hide this warning
WARNING: If a serviceability tool is not in use, please run with -Djdk.instrument.traceUsage for more information
WARNING: Dynamic loading of agents will be disallowed by default in a future release

3. Missing SLF4J providers (occurs in several different modules)

> SLF4J(W): No SLF4J providers were found.
SLF4J(W): Defaulting to no-operation (NOP) logger implementation
SLF4J(W): See https://www.slf4j.org/codes.html#noProviders for further details.

4. Ignored SLF4J static binding (Jython, Database)

> SLF4J(W): Class path contains SLF4J bindings targeting slf4j-api versions 1.7.x or earlier.
SLF4J(W): Ignoring binding found at [jar:file:/Users/tfmorris/.m2/repository/org/slf4j/slf4j-simple/1.7.36/slf4j-simple-1.7.36.jar!/org/slf4j/impl/StaticLoggerBinder.class]
[...]

PR #7355, opened 2025-06-29
Fix build warnings related to annotation processing as per issue #7354

sumitshresht:
Fixes #7349

-  **Removed deprecated** usage of <compilerVersion> in benchmark/pom.xml.
-  **Added standard Maven properties** <maven.compiler.source> and <maven.compiler.target> to specify the Java version.
- These changes align the project with modern Maven practices and avoid potential build issues with future Java versions.

### Warnings Resolved
**Deprecated Compiler Property:**

- Removed: 'compilerVersion' (user property 'maven.compiler.compilerVersion') is deprecated...
- Solution: Switched to standard properties <maven.compiler.source> and <maven.compiler.target>.

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -51,6 +51,11 @@
     <jmh.version>1.37</jmh.version>
     <javac.target>${java.minversion}</javac.target>
     <uberjar.name>openrefine-benchmarks</uberjar.name>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <jmh.version>1.37</jmh.version>
+    <uberjar.name>openrefine-benchmarks</uberjar.name>
+    <maven.compiler.source>${java.minversion}</maven.compiler.source>
+    <maven.compiler.target>${java.minversion}</maven.compiler.target>
   </properties>
 
   <build>
@@ -60,11 +65,11 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.14.0</version>
         <configuration>
-          <compilerVersion>${javac.target}</compilerVersion>
-          <source>${javac.target}</source>
-          <target>${javac.target}</target>
+          <source>${maven.compiler.source}</source>
+          <target>${maven.compiler.target}</target>
         </configuration>
       </plugin>
+
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -52,7 +52,7 @@
       <url>https://github.com/OpenRefine/OpenRefine/graphs/contributors</url>
       <roles>
         <role>Contributor</role>
-      </roles>  
+      </roles>
     </developer>
   </developers>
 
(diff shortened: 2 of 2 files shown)


sumitshresht:
"Hi, I have fixed the warnings as mentioned in issue #7349. Please let me know if further changes are needed. Thank you!"

SoryRawyer:
Hi @sumitshresht , thanks for looking into this! Could you please fill out the pull request template with details on your changes? It would be especially helpful to understand which warnings you're trying to resolve and how the changes in this PR address them.

sumitshresht:
### Changes Made

-  **Removed deprecated** usage of <compilerVersion> in benchmark/pom.xml.
-  **Added standard Maven properties** <maven.compiler.source> and <maven.compiler.target> to specify the Java version.
- These changes align the project with modern Maven practices and avoid potential build issues with future Java versions.

### Warnings Resolved
**Deprecated Compiler Property:**

- Removed: 'compilerVersion' (user property 'maven.compiler.compilerVersion') is deprecated...
- Solution: Switched to standard properties <maven.compiler.source> and <maven.compiler.target>.

benchmark/pom.xml
```diff
@@ -51,6 +51,11 @@
     <jmh.version>1.37</jmh.version>
     <javac.target>${java.minversion}</javac.target>
     <uberjar.name>openrefine-benchmarks</uberjar.name>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <jmh.version>1.37</jmh.version>
+    <uberjar.name>openrefine-benchmarks</uberjar.name>
```
SoryRawyer, line 56: Could you please remove these duplicate lines?

pom.xml
```diff
@@ -129,6 +129,7 @@
     <maven-deploy-plugin.version>3.1.4</maven-deploy-plugin.version>
     <license-plugin.version>2.6.0</license-plugin.version>
     <org-jvnet-localizer-plugin.version>1.31</org-jvnet-localizer-plugin.version>
+    <skipTests>false</skipTests>
```
SoryRawyer, line 132: Is this related to the benchmark pom changes? This doesn't seem like it resolves any of the warnings in the related issue.

SoryRawyer (changes requested):
Thanks for tackling some of these warnings! There are just a few small changes to make and then I think we're all set.

sumitshresht:
Thanks for pointing that out! I had initially removed <jmh.version>, but it caused a build error because it's not defined in the parent POM. Since it's only used in the benchmark module, I’ve kept it here and removed the truly duplicated properties like project.build.sourceEncoding and uberjar.name.

benchmark/pom.xml
sumitshresht: Thanks for pointing that out! I had initially removed <jmh.version>, but it caused a build error because it's not defined in the parent POM. Since it's only used in the benchmark module, I’ve kept it here and removed the truly duplicated properties like project.build.sourceEncoding and uberjar.name

tfmorris:
Welcome to the project @sumitshresht ! It looks like this PR is intended to address issue #7354 (the referenced #7349 is a PR), but #7354 is already assigned and has a fix in process, so I would suggest choosing an unassigned issue to work on.

Also, if you need to update your original issue (for example, to include the template info that @SoryRawyer  requested), you should be able to edit the issue. If this doesn't work for some reason, let us know and we can look into how the permissions are set up.

SoryRawyer:
Given that the underlying issue is already assigned, I'm going to go ahead and close this pull request. We can always reopen if needed.

PR #7361, opened 2025-07-02
Add mouse shortcuts to grid header

### LINKED ISSUE #6282: Improving renaming columns

SoryRawyer:
Fixes #6282 

This pull request extends the changes in #7268 to include feedback on the PR. The original commit has been kept to include proper attribution to @ntcho, who did the initial work to add these shortcuts.

- Add double click mouse shortcut to open column renaming dialog
- Add right click mouse shortcut to open column menu
- Add middle click mouse shortcut to collapse column

In addition to the original pull request, this update adds:
- additional text in the column title to indicate on hover that columns are editable with a double click
  - corresponding updates to tests that were broken due to this change (the `^=` operator looks for strings that start with a given input rather than an exact match)
  - a text snippet in `translation-en.json` for the new title text
- a Cypress test to confirm that double click to rename column works as expected

I tried to add cypress tests for the middle click functionality, but I was not able to reliably simulate the auxiliary click. If folks have suggestions for how to do that, I'd really appreciate it.

**Update (2025-11-05)**
Here are screenshots of what it looks like to hover over the header and then to update the column name.
On hover:
Screenshot from 2025-11-05 17-18-02

Editing the column name:
Screenshot from 2025-11-05 17-20-07

For comparison, here is what the project rename interaction looks like.
On hover:
Screenshot from 2025-11-05 17-21-03

Editing the project name:
Screenshot from 2025-11-05 17-21-18

### DIFF
--- a/main/tests/cypress/cypress/e2e/project/grid/column/edit-column/split_into_several_columns.cy.js
+++ b/main/tests/cypress/cypress/e2e/project/grid/column/edit-column/split_into_several_columns.cy.js
@@ -17,9 +17,9 @@ describe(__filename, function () {
     );
 
     cy.get('.data-table-header').find('th').should('have.length', 7)
-    cy.get('th:nth-child(3) > div.column-header-title > span').should('to.contain', 'Shrt_Desc 1');
-    cy.get('th:nth-child(4) > div.column-header-title > span').should('to.contain', 'Shrt_Desc 2');
-    cy.get('th:nth-child(5) > div.column-header-title > span').should('to.contain', 'Shrt_Desc 3');
+    cy.get('th:nth-child(3) > div.column-header-title > .column-header-name').should('to.contain', 'Shrt_Desc 1');
+    cy.get('th:nth-child(4) > div.column-header-title > .column-header-name').should('to.contain', 'Shrt_Desc 2');
+    cy.get('th:nth-child(5) > div.column-header-title > .column-header-name').should('to.contain', 'Shrt_Desc 3');
 
     cy.assertCellEquals(0, 'Shrt_Desc 1', 'BUTTER');
     cy.assertCellEquals(0, 'Shrt_Desc 2', 'WITH SALT');
@@ -44,9 +44,9 @@ describe(__filename, function () {
     );
 
     cy.get('.data-table-header').find('th').should('have.length', 8);
-    cy.get('th:nth-child(3) > div.column-header-title > span').should('to.contain', 'Shrt_Desc');
-    cy.get('th:nth-child(5) > div.column-header-title > span').should('to.contain', 'Shrt_Desc 2');
-    cy.get('th:nth-child(6) > div.column-header-title > span').should('to.contain', 'Shrt_Desc 3');
+    cy.get('th:nth-child(3) > div.column-header-title > .column-header-name').should('to.contain', 'Shrt_Desc');
+    cy.get('th:nth-child(5) > div.column-header-title > .column-header-name').should('to.contain', 'Shrt_Desc 2');
+    cy.get('th:nth-child(6) > div.column-header-title > .column-header-name').should('to.contain', 'Shrt_Desc 3');
 
     cy.assertCellEquals(0, 'Shrt_Desc', 'BUTTER,WITH SALT');
     cy.assertCellEquals(0, 'Shrt_Desc 1', 'BUTTER');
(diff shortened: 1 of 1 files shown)


main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
```diff
@@ -67,6 +67,72 @@ DataTableColumnHeaderUI.prototype._render = function() {
...
+            }
+        );
+      }
+    });
+    elmts.columnNameInput.trigger('focus').trigger('select');
+  });
+
+  // shortcut: right click to show menu
+  this._td.addEventListener('contextmenu',function(e) {
+    e.preventDefault();
```
Abbe98: Let's not override native browser features.

main/webapp/modules/core/scripts/views/data-table/column-header-ui.js
```diff
@@ -67,6 +67,72 @@ DataTableColumnHeaderUI.prototype._render = function() {
   var elmts = DOM.bind(td);
 
   elmts.nameContainer.text(this._column.name);
+
+  // shortcut: double click to rename column
+  this._td.addEventListener('dblclick',function() {
```
Abbe98: This appears to be copy-pasted form the existing dialog, break it out from the menu according to the existing pattern instead.

main/webapp/modules/core/scripts/views/data-table/column-header.html
```diff
@@ -1,5 +1,5 @@
-<div class="column-header-title"><button class="column-header-menu" bind="dropdownMenu"></button><span class="column-header-name" bind="nameContainer"></span></div>
+<div class="column-header-title editable-column-header-title"><button class="column-header-menu" bind="dropdownMenu"></button><span class="column-header-name" bind="nameContainer"></span></div>
```
Abbe98: It's a little bit weird that we make the div interactive especially as it wasn't that long ago we refactored the project name area to be accessible.

SoryRawyer:
Thanks for the review and historical context! I'll take a look at the project name refactoring. I noticed things were a bit different between this and the project name code, but my initial goal here was to fill out the original PR. @magdmartin and I talked about trying to get this into 3.10 since it's a requested feature, but I agree that it makes sense to revisit the approach instead of rushing it into the next release.

SoryRawyer:
@Abbe98 my apologies for letting this linger for so long. I've tried to address the points you mentioned. I also removed the right-click shortcut, as it felt awkward after removing `preventDefault`. I think it would be easy enough to re-add once the native browser functionality interaction is considered in more depth.

SoryRawyer:
@smierz would you be interested in reviewing this PR? It seems like you're not part of the GitHub org but I'd like to nominate you on the forum if you're interested! If you join then you'll be able to formally approve/request changes on pull requests.

tfmorris:
In case anyone else missed it, screenshots were added to the ^original description^ up top a couple of weeks ago. (Thanks @magdmartin )

Initial thoughts:
- I think double click has accessibility issues. It also conflicts with word select and I think text select was one of the previous complaints.
- While I appreciate that the two implementations are aligned, I don't think they necessarily need to be aligned to the current project rename, which I don't love.
- Specifically a pop-up rename dialog seems very heavyweight and a color highlight + hover text seems kind of opaque for indicating edit-ability
- The orange/red is garish and doesn't match the rest of the color palette

Here's an overview of lots of different styles for indicating editability:
https://webapphuddle.com/inline-edit-design/

I think I'd lean towards something along the lines of:
- get rid of yellow background
- indicate editability with pencil icons - always on, light grey for project name, normally hidden, but visible on hover for column headings. Icons go to right of project name and at far right of column header.
- all name editing happens in-line rather than using a popup dialog. ESC reverts. Clicking somewhere else commits.

magdmartin:
@tfmorris  I'll invite you to  review from the latest built shared [on the forum](https://forum.openrefine.org/t/project-grid-header-design/2622/2) as it has a different design than the screenshot. 

From my perspective I am happy with the current design (see video) for resize, rename and header selection:
* The **resize** affordance in the header works well.
* **Rename interaction**: after a second review I see that there is two pattern
  * One click select the yellow text and then you need to double click on it to open the modal. In that scenario, I would prefer a single click to trigger rename (consistent with how the project-name is edited), rather than double-click. 
  *  A double click directly on the column name on the header open the rename modal directly.
* **Header text selection** is now available via the rename functionality. It is an extra click from the current behaviour but consistent with the project name selection. 
* **middle click to collapse** works for me (not shown in the video)
* **right click on the column header to open the menu** does not work. However two right click collapse the column (regarless of the speed of the click - not shown in the video)

I recorded a short video showing the three comments: [ctg studies csv - OpenRefine.webm](https://github.com/user-attachments/assets/5d6ede8b-7abe-4cf5-869c-b54dc253afcc)

I am commenting on #7273 regarding the colum reordering 

_edited on November 26 after further test_

SoryRawyer:
Since there hasn't been much activity on this for a while, I'm going to close this. We can revisit if and when this conversation restarts.

PR #7370, opened 2025-07-14
row sampling during project creation

### LINKED ISSUE #705: Support row sampling during project creation
The only type of row subsampling that we support is N contiguous rows (after optionally skipping M rows).

It would be valuable to be able to support additional methods including:
- every Kth row (to allow repeatable sub-sampling)
- a pseudorandom selection of X% of the rows
- exactly Y rows randomly selected (using reservoir sampling)

smierz:
closes #705 

#### Changes proposed in this pull request:
- added 3 sampling methods (reservoir, bernoulli, systematic) which were specified in the linked issue
-  the 3 sampling methods are used within the import functionality to sample rows of a project 
   -  the sampling is hooked into the import process _after_ the parsing of the project (not within parsing to keep single responsibility and allow for reuse in other parsers)
   -  and _before_ removing empty columns (as within the sample more columns might be empty than in the original population)
- Because there were no importers specified in the issue, the functionality is currently **_only available_** and tested in the separator based import
   -  backend tests for the sampling functionality in `SeparatorBasedImporterTests` 
   -  e2e tests for frontend additions in `preview_project.cy.js`
   - frontend additions to enable/disable sampling and selecting the sampling method and factor   image

#### Notes:
* I'd prefer to keep this PR to one importer (separator based) and focus on the components and their interactions as this is my first bigger PR in this repo. Once we agree on the sampling for one importer, adding the functionality into other importers will be much easier and can be addressed in smaller follow-up PRs.
* Within `ImportingParserBase.sampleRows` I added some questions in the form of "ToDos" as I am not familiar what the project precomputes and what is cached, maybe someone that  is more familiar with the project could take a look?
* I'm not too much of a frontend dev so if anyone has suggestions/improvements/comments, feel very free to comment


SoryRawyer:
Hi @smierz, thanks for opening this and my apologies for taking so long to review! I'm still going through everything, but while testing it out I think the sampling might only occur on the first 100 rows of the new file(s). I'm using a file with ~700k rows and the sampling doesn't seem to line up with what I'd expect. Would you mind taking a look to see if we might be sampling from too small of an initial set of rows?

smierz:
Hi @SoryRawyer 
i haven't checked yet, when you said only the first 100 rows it rang a bell, because for the preview only 100 rows (see value for 'limit') are read from the file initially, only if you actually continue to create the project the whole file is parsed and then the sample should contain rows from all over the file. It's just an assumption but maybe that is what's happening?

Note
As far as I understand the preview will not necessarily be "correct" in terms of the whole project as it is limited to the first 100 rows of the first file to make it update fast, see comments here:
https://github.com/OpenRefine/OpenRefine/issues/7214#issuecomment-2809293290

---

_Update_:
I re-checked with a file with 2million rows and the sample is created from all over the file when creating the project (preview only considers first 100 rows). Also remember to page through the rows in the project or set the page size to 1000 for example, so you see all the sampled rows.

Btw there is also a test covering that the rows are taken from the whole file -> `SeparatorbasedImporterTests.testImportWithSystematicSampling`

SoryRawyer:
Thanks for checking!

On another note, I was thinking it might be nice to hold off on making this an extension point right now. Since these are the first samplers, I think it would make sense to release them in the backend and add an extension point once we have a better understanding of the interface required for custom samplers as well as a clear use case for a custom sampling method. What do you think?

smierz:
just for clarification:
if I understand correctly, you would prefer to plug-in the sampling methods directly for now (basically kick-out registry + sampler interface)?

SoryRawyer:
I was thinking we could still keep the registry, but make it non-public and populate it outside of `controller.js`. Hopefully that would minimize the changes needed to this PR but would still leave the option of promoting the registry to an extension point later if we'd like. How do you feel about that?

SoryRawyer (commented):
After reading through the code related to the "todo" questions, it's not clear to me what the best answer would be. @tfmorris, is this something you could answer?

tfmorris:
I'll add this to my list to review, but to answer the questions:

1. Writable global variables aren't a great design choice, but it is what the current design is.
2. Precompute cache clearing and other stuff done by RemoveRows should be unnecessary if this is only being done as part of the import process. If we were to introduce as SubsampleRows command, the answer would be different. I'll review the data flow when I do my code review, but, at glance, it looks like it's not an issue.

smierz:
@SoryRawyer 
> I was thinking we could still keep the registry, but make it non-public and populate it outside of `controller.js`. Hopefully that would minimize the changes needed to this PR but would still leave the option of promoting the registry to an extension point later if we'd like. How do you feel about that?

I tried rewriting the PR based on your suggestions:
* removed the changes to `controller.js`
* made the `SamplerRegistry` package-private

ok, so now I have the individual samplers in the main package and `Sampler`, `SamplerRegistry` and `ImportingParserBase` in core package. 

--> Since main depends on core, I can't fill the `SamplerRegistry` in the class itself because the individual samplers are unknown there. I either need to move the individual samplers into core or the sampler interface + registry into main.

--> If I keep the `SamplerRegistry` and make it non-public (package private), it's not accessible from within `ImportingParserBase` / `SeparatorBasedImporter` where it is used, since it is in its own package "sampling". To fix it I would have to move sampling classes (`SamplerRegistry`, `Sampler`, individual samplers) into the "importers" package. So in the end all sampler classes would reside in package "importers" (in core or main).

Just wanted to check, if that is what you had in mind and which place for the registry you would prefer, before reworking the structure.

.

Alternatively/additionally the `sample` method could also be rewritten to take (List, Map) parameters instead of (List, int), as to widen its use (also suggested for exporters here https://github.com/OpenRefine/OpenRefine/issues/6419)

SoryRawyer:
Hi @smierz, apologies for the delayed response. I really appreciate your comments here, since I didn't fully grasp the awkward implementation that my comment implied. I think moving things into the importers package for now makes sense (even maybe as a sub-package if that's not too convoluted). If and when we decide to make these interfaces public, we can always move them into a separate package that's accessible across the project.
What do you think? Does that make sense?

smierz:
I will look into it

tfmorris:
This is very cool!

> One thing that could be added would be to generate a seed within the frontend, add it to the sampling object within request that gets passed down all the way to the individual sampling methods. 

I think that could be useful.

re: public extension point - I have no problem marking a new public extension point as "provisional" or "experimental" until we are confident that we want to open it up for use. This can be done in documentation without having to change visibility of things.

re: single importer only to cut down on work required - That may be a sign that we need to refactor/simplify things to make that extra level of work go away or be greatly reduced. We should be able to introduce importer independent options without a lot of duplicated work.

My biggest question is where the sampler should be incorporated into the input flow. Doing the sampling post hoc vs inline removes the possibility of some potential efficiencies.

I think one good success criterium for the new API would be whether or not it can be used to efficiently implement our current "skip N", "limit to M" import options, which are, effectively, very simple samplers.

As a general rule, I'd like to see us leverage more of the Java 8 Streams API, although there may be too many legacy APIs involved here to make that a possibility. Something worth keeping in mind when designing new APIs though. Conceptually this might be modeled as a stream of records/rows to which a sample filter is applied.

I recognize that our baroque importer architecture with tree vs table, stream vs reader, etc, is a hindrance to designing/extending APIs, so I'll dig into this a little more to see if I can make some concrete suggestions for how an inline solution might work.

tfmorris:
Here's a sketch of what I'm thinking might work for a Sampler interface. Basically it's a generator for row indices with some special flag values for skipping rows and end of stream (when we reach the row limit).

```java
public interface Sampler {

    public final int SKIP = -1;
    public final int END = -2;

    /**
     *
     * @return an integer representing the next row index for the given sampling strategy or a special flag
     */
    public int nextIndex();
}
```

```java
public abstract class AbstractSampler implements Sampler {

    protected int limitLines = -1;

    public AbstractSampler(Integer limitLines, Number parameter) {
        this.limitLines = limitLines;
    }
}
```

switching the registration to Classes instead of instances and instantiating samplers with their needed parameters so that the instances can maintain the necessary state and modifying `TabularImportingParserBase` to call the samplers and modify  its processing based on the stream of indices that it receives (obviously `TreeImportingParserBase` and, perhaps, the RDF importer) would need parallel work done.

I sketched out enough of this on the tabular importer side of things to convince myself that it would work and I'm happy to provide that as a starting point if you want or you can tackle it with a fresh set of eyes.

PR #7376, opened 2025-07-19
Started working on Apache Parquet import

### LINKED ISSUE #1929: Has anyone looked at loading parquet files?
Not too familiar with refine but could have a look at adding this if someone gives me a starting point.

jze:
Started working on the import of Apache Parquet files. The main class is ParquetImporter. I have used the OdsImporter as a reference. An initial unit test is also included.

I am still struggling with the integration into the web interface.

Closes 

Fixes #1929

- import for Apache Parquet files


tfmorris:
Welcome to the project! It would be great to have a Parquet importer, so thank you for contributing.

You can fix the lint issues using the command `mvn formatter:format`, the committing and pushing the modified files.

I started a basic sketch of a Parquet importer almost a year ago, but didn't take it very far and haven't looked at it since. You can find it here, in case it's any help:  https://github.com/OpenRefine/OpenRefine/compare/master...tfmorris:OpenRefine:parquet-importer

Let us know if we can help with any of the pieces that you are struggling with

main/webapp/modules/core/MOD-INF/controller.js
```diff
@@ -232,6 +232,7 @@ function registerImporting() {
   IM.registerFormat("text/xml", "core-import-formats/text/xml", "XmlParserUI", new Packages.com.google.refine.importers.XmlImporter());
   IM.registerFormat("binary/text/xml/xls/xlsx", "core-import-formats/binary/text/xml/xls/xlsx", "ExcelParserUI", new Packages.com.google.refine.importers.ExcelImporter());
   IM.registerFormat("text/xml/ods", "core-import-formats/text/xml/ods", "ExcelParserUI", new Packages.com.google.refine.importers.OdsImporter());
+  IM.registerFormat("binary/parquet", "core-import-formats/binary/parquet", "ExcelParserUI", new Packages.com.google.refine.importers.ParquetImporter());
```
tfmorris:
```suggestion
  IM.registerFormat("binary/parquet", "core-import-formats/binary/parquet", "ParquetParserUI", new Packages.com.google.refine.importers.ParquetImporter());
```

main/src/com/google/refine/importers/ParquetImporter.java
```diff
@@ -0,0 +1,117 @@
...
+    public ParquetImporter() {
+        super(false);
+    }
+
+    @Override
+    public ObjectNode createParserUIInitializationData(
+            ImportingJob job, List<ObjectNode> fileRecords, String format) {
+        ObjectNode options = ParsingUtilities.mapper.createObjectNode();
+        JSONUtilities.safePut(options, "includeFileSources", true);
+        JSONUtilities.safePut(options, "encoding", "utf-8");
```
tfmorris:
```suggestion
```
Encoding is probably not relevant to a binary format like parquet

main/src/com/google/refine/importers/ParquetImporter.java
```diff
@@ -0,0 +1,117 @@
...
+                }
+
+                @Override
+                public List<Object> finish(List<Object> target) {
+                    return target;
+                }
+            };
+
+            // add each row to the project
+            Iterator<List<Object>> it = ParquetReader.streamContent(new File(fileSource), HydratorSupplier.constantly(hydrator)).iterator();
```
tfmorris: This should read from inputStream, but it'll take some work to make it compatible with the ParquetReader's requirements.

jze: I had already suspected that the problems I encountered are related to my attempt to read from a File and and an InputStream. Is it possible to read the file twice or has it to be done in one go?

jze: As a workaround I stored the data in a temporary file.

main/webapp/modules/core/MOD-INF/controller.js
jze: Using the `ExcelParserUI` as a reference I have created a `ParquetParserUI`.

main/src/com/google/refine/importers/ParquetImporter.java
tfmorris: That's fine as a hack to get things working, but a temporary file has already been created, so I think what we need to do is refactor things to provide access to it.

The main (only) reason for using an InputStream here is to allow for progress monitoring as the data is read.

jze: The required File object is even present in the [parseOneFile method](https://github.com/OpenRefine/OpenRefine/blob/0fc75028d91acb8dc2bef7d0c2c93a6035e5a623/modules/core/src/main/java/com/google/refine/importers/ImportingParserBase.java#L129). It would only have to be passed as a parameter.

jze: I have added a third mode to import directly from a file. If that is too much change for this pull request I can also create a new pull request with just that specific change.

tfmorris: Thanks for continuing to push this forward. I think we definitely want the API changes in a separate PR, but as I mentioned above https://github.com/OpenRefine/OpenRefine/pull/7376#discussion_r2220240279 the reason for using an InputStream (or Reader built on it) is to provide the user with progress tracking and I don't think we want to lose that (which we would with a file based API).

The other thing that we need to be careful of is backward compatibility. Things like reducing the visibility of methods, removing protected fields, etc are technically breaking changes.

On the flip side, if we're going to extend this API, I think it makes sense to do it in a cleaner way and not continue the use of the kludgy runtime constructor argument technique that is currently used.

My current thinking is to add support for passing a progress tracked [SeekableByteChannel](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/channels/SeekableByteChannel.html) (which would also potentially support other future NIO based importers) in the Importer API and use that with an adapter class which implements that Parquet [InputFile](https://javadoc.io/doc/org.apache.parquet/parquet-common/latest/org/apache/parquet/io/InputFile.html) interface.

Since this is a change to a public API which we advertise as an extension point, the proposed changes should get reviewed by the team, so I'll write up a proposal for folks to look at.

If you would like to look into whether there are any blockers to using a SeekableByteChannel as the data source for what you want to do, that would be very useful.

main/pom.xml
```diff
@@ -415,6 +415,10 @@
       <version>${log4j.version}</version>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
+      <groupId>blue.strategic.parquet</groupId>
```
tfmorris, line 419: I just noticed that this isn't using the Apache Parquet libraries, which is a software supply chain concern. I've seen descriptions of Maven formulas to shave down the Hadoop dependencies. Is that anything you've investigated? What's the size difference between the two methods?

jze: I've been working with this library for so long and haven't checked to see if the situation with the dependencies has improved. In the past, a huge number of libraries were added when depending on the `parquet-hadoop` library. It seems that the dependencies of  `parquet-hadoop` have been reduced significantly. Therefore, I am going to take a look if it can be done without the `parquet-floor` library.

jze: Unfortunately, I was mistaken. There are runtime dependencies to Hadoop. Using only the `org.apache` libraries a simple Parquet reader needs 76 MB of dependencies. Using the `parquet-floor` library only 18 MB dependencies are required. That's quite a difference. The Apache Hadoop library seems to import everything within reach. :wink:

tfmorris: Thanks for checking. This SO [answer](https://stackoverflow.com/a/60067027/167425) mentions the kind of dependency exclusions that I've seen reference, but I haven't tested to see how big a difference it makes. This particular one uses Gradle syntax, but I'm pretty sure I've seen the equivalent in Maven syntax.

tfmorris: I found the post with the Maven dependency exclusions in it. 
https://blakesmith.me/2024/10/05/how-to-use-parquet-java-without-hadoop.html

That technique seems to add about 32MB of dependencies, but I haven't looked at them in detail.

PR #7391, opened 2025-08-14
Remove myself from the core dev group

wetneb:
See https://github.com/OpenRefine/openrefine.org/pull/476


magdmartin:
If there are no other comments, I am planning to merge this on August 20 along with https://github.com/OpenRefine/openrefine.org/pull/476

Issue #7390, opened 2025-08-14
Snapshot builds no longer publishing


SoryRawyer:
[OSSRH reached end-of-life on June 30th, 2025](https://central.sonatype.org/news/20250326_ossrh_sunset/). As a result, all releases should be uploaded instead to the [Maven Central Repository](https://central.sonatype.org/publish/publish-portal-guide/).

I'm still working on figuring out the exact steps to upgrade and will update this issue once I have more information.

**Impact:** it seems to me that we're unable to publish artifacts as a result of this change. Here's an example of a recent build failure: https://github.com/OpenRefine/OpenRefine/actions/runs/16967842327/job/48096447760

SoryRawyer:
@tfmorris do you have the login information for the publisher portal? I'd be happy to look into this but I don't have access to the secrets used in the CI workflow.

tfmorris:
Can you clarify which publisher portal you're talking about? Releases are being published to Maven Central, but I think the reason that the snapshots weren't migrated there also is because Maven Central doesn't support publishing snapshot builds (although I'd have to do some research to verify that).

thadguidry:
They support SNAPSHOT releases for pre-release  development.  https://central.sonatype.org/publish/publish-portal-snapshots/

tfmorris:
@thadguidry Thanks for the clarification.

I think we need to enable snapshots for the org.openrefine namespace, which is not something that I have access to (and I can't see what accounts DO have access). Their documentation is pretty poor, but it's not obvious to me that granting additional users access to a namespace is possible, which would be a problem.

@SoryRawyer can you please check with Martin to see if he or the CS&T IT folks have access? If not, we'll need to get in touch with Antonin.

magdmartin:
CS&S IT does not have access to it.

A quick search on the forum turned those two posts

[Reproducibility project: June 2023 report](https://forum.openrefine.org/t/reproducibility-project-june-report/855) by @wetneb 
> One issue that has been preventing more user testing is the [#5947 unavailability of snapshot releases due to a problem with Sonatype's repository](https://github.com/OpenRefine/OpenRefine/issues/5947). We need to migrate to something else to restore the snapshots, because it feels like it's something which really does make it easier for a lot of people to try out development versions.

In November 2024 @thadguidry [asked](https://forum.openrefine.org/t/release-process-workflow-discussion/1576/3) 
> Who maintains the GPG_PRIVATE_KEY and PASSPHRASE?
[...]

PR #7401, opened 2025-08-25
Switch to OSSRH Staging API. Refs #7390

tfmorris:
This addresses #7390, but may not be a complete solution, but it's a starting point.

Fixes #{issue number here}

-  Change the server ID from ossrh to ossrh-staging-api
-  Switch URLs from oss.sonatype.org to ossrh-staging-api.sonatype.com

I suspect this will fail due to the use of the OSSRH token, but we'll see ...


tfmorris:
I'm going to merge this without review so that I can test the flow. I suspect it will fail due to the use of the OSSRH token, but we'll see...

PR #7403, opened 2025-08-25
Clean up snapshot publishing

### LINKED ISSUE #7390: Snapshot builds no longer publishing

tfmorris:
Fixes #7390 

- Rename a few things from OSSRH to Central to clarify what they represent
- Fix a couple of build warnings that are adding noise to logs (no functional change since they're already ignored)
- Add a placeholder for a POST which is documented as being required, but doesn't, on the surface, appear to be. Leaving available for now in case it turns out to be non-deterministic

Note that I haven't yet created the new secrets since that will invalidate the old ones. I'll do that after this is approved and ready for merge.

### DIFF
--- a/.github/workflows/snapshot_release.yml
+++ b/.github/workflows/snapshot_release.yml
@@ -240,6 +240,8 @@ jobs:
     - name: Package and upload to OSSRH
       run: |
         mvn -B deploy -DskipTests=true -DrepositoryId=ossrh-staging-api -Dapple.notarization.username=$APPLE_USERNAME -Dapple.notarization.password=$APPLE_PASSWORD -Dapple.notarization.team.id=$APPLE_TEAM_ID -Dgpg.signer=bc
+        # The below POST is documented as being necessary, but it seems to work without it. Also not clear if it needs to be authenticated.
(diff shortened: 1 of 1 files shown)


SoryRawyer (approved):
This looks good, thanks for taking care of it! Most of my maven knowledge is a little old, but I don't think the copilot-suggested changes are needed.

tfmorris:
Thanks for the review. I agree about Copilot's comments. I'm experimenting with on PRs, but I think it got confused here.

tfmorris:
So it turns out that Copilot was correct and the IDs do need to match. Rather than untangle which IDs get used where, I just reverted to the previous state of using the same repo id everywhere. Not ideal, but not high enough priority for me to wrestle with.

PR #7411, opened 2025-08-29
Update wdtk.version to 0.17.0

### LINKED ISSUE #7383: Wikidata blocking OpenRefine because it doesn't set User-Agent string
When I try to upload some data to Wikidata from OpenRefine, my Wikidata log in details are not recognised. I message appears indicating 'invalid credentials', which is incorrect as I'm able to log in on a separate tab. 
### To Reproduce
1. First, do ... edits on my data, and reconciliations
2. Then, ... apply the schema, and click on upload to Wikidata
3. Finally, ... a pop up screen comes up asking me to enter the log in details, which doesn't recognise and indicates 'invalid credentials'

### Screenshots

Image

### Versions
 - Operating System:  WIndows and Mac
 - Browser Version:  newest versions of Chrome and Safari
 - JRE or JDK Version: 
 - OpenRefine:

Abbe98:
Fixes #7383, although we should implement or own UA and fix non WDTK requests.

### DIFF
--- a/extensions/wikibase/pom.xml
+++ b/extensions/wikibase/pom.xml
@@ -15,7 +15,7 @@
   </parent>
 
   <properties>
-    <wdtk.version>0.16.0</wdtk.version>
+    <wdtk.version>0.17.0</wdtk.version>
   </properties>
 
   <build>


Abbe98:
Adding the main `okhttp` to the dependencies resolves the build issues, but it's not obvious to me why such a would be needed.

SoryRawyer:
It looks like this might be related to the change from okhttp 4.12 to 5.1 in WDTK. Dependabot had opened a PR for that in this repo too, but the tests for the Wikibase extension were failing. I just tried updating both dependencies locally and the Java tests ran successfully, so maybe we could bump the okhttp version in our main pom as well?

tfmorris:
okhttp changed coordinates and package names with the major version bump: https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-500

It's unfortunate to have a dependency major version bump bundled with a critical fix, but I think the fix is as simple as @SoryRawyer suggested, but why don't we put together a complete fix with the other changes suggested by @Abbe98. 

@Abbe98 do you know what "non-WDTK requests" need to be addressed? Is this something you're already working on, or should I take a look?

tfmorris:
Thanks for getting this kicked off @Abbe98. I've put up a first cut at a more comprehensive solution in #7412 which incorporates this, @SoryRawyer 's suggested OkHttp update, and a custom user agent string for both the WDTK and raw OkHttp calls. 

@Abbe98 Please let me know if you know of any other API calls which need to be addressed.

Abbe98:
> @Abbe98 Please let me know if you know of any other API calls which need to be addressed.

At least the manifest proxy.

PR #7415, opened 2025-09-01
Closes #6311: Unified style of column headers in import preview and p…

### LINKED ISSUE #6311: Make column header separators (td border) easier to see in Import Previews
If a project only has 1 row, for example, only the column header row, then the columns themselves are nearly impossible to distinguish.
In fact, the column header row in preview, simply looks like a long line of text.  There is only the default 1 pixel border separator which is basically "invisible".  
## separating by commas, default border, browser zoomed 200% - and still no column separation visible

Screenshot 2024-01-17 105624

### Proposed solution
We should bring a new column header style in the import previews (and maybe export previews, haven't checked).
This style would bring better readability to the column separation, especially when there is no data table with cell bordering to help the eye see the different columns.

`border-width` and `border-color` could be applied.
Optionally, perhaps a bit of `border-radius` could also be used.

```css
border-radius: 5px;
border-width: medium;
border-right-color: royalblue;
```

## improvement of readability for individual columns in header
Screenshot 2024-01-17 105515

style and colors can be tweaked, but you get the general idea and can clearly see the distinct columns in the header row in preview.

zleonenc:
Changed table header creation in preview-table.js to use <thead> and <th>. Updated CSS to apply header styles to th.column-header for consistency.

Closes #6311

### DIFF
--- a/main/tests/cypress/cypress/e2e/create-project/preview_project.cy.js
+++ b/main/tests/cypress/cypress/e2e/create-project/preview_project.cy.js
@@ -202,22 +202,22 @@ describe(__filename, function () {
     cy.visitOpenRefine();
     cy.createProjectThroughUserInterface('food-blank-column.mini.csv');
     cy.get('.create-project-ui-panel').contains('Configure parsing options');
-    // FIXME: data-table in preview has no thead, so first tr in tbody is a header
-    cy.get('table.data-table > tbody > tr:nth-child(2) > td:nth-child(1)').should('to.contain', '1.');
-    cy.get('table.data-table > tbody > tr:nth-child(2) > td:nth-child(2)').should('to.contain', '01001');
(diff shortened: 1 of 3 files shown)


SoryRawyer:
@zleonenc, would you be able to take a look at the test that's failing? Also, could you upload before and after screenshots of your change? I'm having a hard time seeing much of a difference locally and want to make sure I'm not missing something.

zleonenc:
@SoryRawyer, Sure, I've fixed it. I add the screenshots below

Before:
Screenshot 2025-09-06 at 21 40 24

After:
Screenshot 2025-09-06 at 21 41 16

Seems to me that the issue was partially fixed since the separator was already implemented. However, headers were part of tbody (in fact the previous tests acknowledged that).

Now headers and rows are appropriately separated into thead and tbody.

zleonenc:
Hi @SoryRawyer , the PR now passes the checks. Please let me know if anything else is needed.

PR #7425, opened 2025-09-02
Introduce jackson-bom and update Jackson to 2.20.0. Fixes #7424

### LINKED ISSUE #7424: Switch to using jackson-bom for versionign
Jackson is no longer keeping version numbers synchronized for all their components (starting with 2.20/2.20.0). To work around this, we should switch to using their `jackson-bom` component for versioning.

tfmorris:
Fixes #7424

- Introduce `jackson-bom` in `Maven dependencyManagement`
- update Jackson from 2.19.2 to 2.20.0/2.20


tfmorris:
> It seems like all the poms in this repo reference (in one way or another) the root pom as a parent. Could we move the dependencyManagement section to the root-level pom file instead of having it in each pom individually?

Good point. Done.

> It looks like we're able to remove jackson.version with this change, is that something we'd want to do?

I left it intentionally, just so all the versions would be centralized, but failed to update/use it, which definitely wasn't my intention, so thanks for catching that. A previous iteration had that change, but it seems to have gotten lost in the shuffle. It's been updated now.

SoryRawyer (approved):
Looks great, thanks for taking this on! I didn't know about jackson-bom; this is a nice improvement.

PR #7439, opened 2025-09-16
fix db import progress updates

### LINKED ISSUE #6654: Import progress bar exceeds the intended box
### To Reproduce
It seems the issue exists for Database imports (so far what I saw). Have one or more large projects already setup. Create a new project from Database (Postgres in this case).

### Current Results
Screenshot 2024-06-03 at 14 02 39

### Expected Behavior
Progress in percent of the current import.

### Versions
 - Operating System: MacOS M1 14.5
 - Browser Version: Brave Version 1.66.118 Chromium: 125.0.6422.147 (Official Build) (arm64)
 - JRE or JDK Version: OpenJDK 17 Docker image based on https://github.com/EasyPi/docker-openrefine/blob/master/Dockerfile
 - OpenRefine: 3.8.1

### Datasets
I have the [Crossref Data](https://academictorrents.com/details/d47fbe29e5ef93a6695421f79a6efa4b801acff1) imported into Postgres.

### Additional context
Rows being loaded from Postgres: 100,000

Here is the `metadata` from the API call to `/command/core/get-importing-job-status?jobID=38`

```json
{
    "metadata":{
        "created":"2024-06-03T11:59:20.648983793Z",
        "modified":"2024-06-03T11:59:20.648983793Z",
        "name":"",
        "tags":[
            
        ],
        "creator":"",
        "contributors":"",
        "subject":"",
        "description":"",
        "rowCount":0,
        "title":"",
        "version":"",
        "license":"",
        "homepage":"",
        "image":"",
        "importOptionMetadata":[
            
        ],
        "customMetadata":{
            
        }
    },
    "lastTouched":1717416051079,
    "updating":false,
    "canceled":false,
    "rawDataDir":"/tmp/jetty-0_0_0_0-3333-webapp-_-any-15943075615229453671/import/38/raw-data",
    "config":{
        "state":"creating-project",
        "hasData":false,
        "progress":{
            "message":"Reading jdbc:postgresql://crossrefdb.geo.svc:15432/crossref?useSSL=false",
            "percent":269,
            "memory":12655,
            "maxmemory":24166
        }
    }
}
```

smierz:
Fixes #6654 

**Problem**:
DB import uses a counter instead of percentage to track its progress (see corresponding issue)

**Changes proposed in this pull request:**
- before starting the db import, we query the db for the `COUNT(*)` of the specified query, similar to the existing functionality for querying batches using `LIMIT` and `OFFSET`
- case count available: 
we use the total count and an additional counter variable to track progress of the db import and calculate the percentage for the UI
- case count not available:
 (e.g. query failed or an extension uses constructor of `DBQueryResultImportReader` that does not include count) then we switch to not showing the progress bar but a spinner instead.

In any case we additionally add the absolute progress "(currentcount / totalCount)" to the progress message so the user can still see progress happening.

**Screenshots**:
or-db-import-progress
or-db-import-progress-unknown

**Notes**:
- additionally removed one log statement, which is already printed 5 lines prior so code is less cluttered (first commit)
- the variable `processedRows` in `DBQueryResultImportReader` is not needed and can be replaced by `currentCount` but lets focus on the correct progress implementation in this PR (no change in import functionality)
- building the sql statement is basically adapted from the limit query and has so much potential to go wrong but this needs to be handled in a separate PR where both query builders / query validation should be addressed  (if the query fails, we switch to the spinner and absolute progress so for this PR should be fine)

### DIFF
--- a/extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
+++ b/extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
@@ -118,11 +118,6 @@ public class DBQueryResultImportReader implements TableDataReader {
                     if (processedRows % 100 == 0) {
                         setProgress(job, querySource, progress++);
                     }
-                    if (processedRows % 10000 == 0) {
-                        if (logger.isDebugEnabled()) {
-                            logger.debug("[[ {} rows processed... ]]", processedRows);
-                        }
-                    }
                 }
                 return result;
             } else {
--- a/extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
+++ b/extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
@@ -59,8 +59,16 @@ public class DBQueryResultImportReader implements TableDataReader {
     private DatabaseService databaseService;
     private DatabaseQueryInfo dbQueryInfo;
     private int processedRows = 0;
+    private final Integer totalCount;
+    private int currentCount = 0;
+    private final boolean useTotalForProgress;
+    @Deprecated
     private static int progress = 0;
 
+    /**
+     * @deprecated use constructor that includes the total count instead
+     */
+    @Deprecated
     public DBQueryResultImportReader(
             ImportingJob job,
             DatabaseService databaseService,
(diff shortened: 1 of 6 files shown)


SoryRawyer (approved):
Thanks, I think this looks good! I tested a little bit and everything looked as expected.

PR #7443, opened 2025-09-19
Show error message on export-rows failure

### LINKED ISSUE #7368: Export to XLS with too many rows results in 500 status instead of error message to user
If you attempt to export more than the supported number of rows (65535) to XLS, it fails in a way which doesn't inform the user.

### Current Results

You get a HTTP 500 status to the web client and the following message in the server log:

```
03:17:12.428 [        ExportRowsCommand] error:Invalid row number (65536) outside allowable range (0..65535) (4502ms)
```

The message displayed to the user by Chrome isn't at all informative
```
This site can’t be reached
The webpage at http://127.0.0.1:3333/command/core/export-rows/wikidata-all-writers-missing-olid-tsv.xls?csrf_token=QAKOMr0Ifgkh94YArbZYUNISQTa5MP2b might be temporarily down or it may have moved permanently to a new web address.
ERR_INVALID_RESPONSE
```

### Expected Behavior

At least two improvements:
1. Explicitly check for exceeding the maximum number of rows (or columns) for the format
2. Catch-all error handler that displays some kind of fallback error to the user

### Versions
 - Operating System: 
 - Browser Version: 
 - JRE or JDK Version: 
 - OpenRefine: 3.9 HEAD

SoryRawyer:
This change updates the client-side exporter code to handle unexpected failures and present an error message to the user instead of showing a generic 500 error.

fixes #7368

- Replace the export-rows form submission with a more manual POST request
- Show an error pop pup if necessary instead of opening a new tab with a 500 error message

Some notes about this solution:
1. Getting the filename looks a little hacky, but all of the suggestions I've seen online involve complicated regular expressions to parse the header. Since we control the response format, this seemed like a reasonable tradeoff between readability and robustness.
2. There are additional `form.submit()` calls in the frontend and I think this approach could be refactored to replace those uses, but I figured that could be handled in follow up pull requests.

**Draft pull request question:**
The error message is currently just "Bad request". There is an error page that is created and returned as part of the HTTP response, but I'm not aware of an existing general error pop-up mechanism that would be appropriate for showing the page. Is there an existing mechanism that could be re-used? Should this alert direct users to look at the service logs to find the appropriate error?

Here are before/after screenshots of the errors shown to users:
Before (opens in a separate tab):
Screenshot from 2025-09-19 09-45-32

After (pops up within the OpenRefine project tab):
Screenshot from 2025-09-19 09-44-15

### DIFF
--- a/main/webapp/modules/core/scripts/project/exporters.js
+++ b/main/webapp/modules/core/scripts/project/exporters.js
@@ -136,12 +136,40 @@ ExporterManager.stripNonFileChars = function(name) {
 
 ExporterManager.handlers.exportRows = function(format, ext) {
   Refine.wrapCSRF(function(csrfToken) {
-    let form = ExporterManager.prepareExportRowsForm(format, true, ext, csrfToken);
-    document.body.appendChild(form);
-    form.submit();
-    document.body.removeChild(form);
+    const name = encodeURI(ExporterManager.stripNonFileChars(theProject.metadata.name));
+    const url = "command/core/export-rows/" + name + ((ext) ? ("." + ext) : "") + "?" + $.param({csrf_token: csrfToken});
+    const jsonData = {
+      project: theProject.id,
+      format: format,
+      engine: JSON.stringify(ui.browsingEngine.getJSON()),
+    };
+    $.ajax({
+      url: url,
+      data: jsonData,
+      type: "POST",
+      success: function(data, _textStatus, xhr) {
+        const filename = getResponseFilename(xhr.getResponseHeader("Content-Disposition"));
+        const blob = new Blob([data], {type: xhr.getResponseHeader("Content-Type")});
+        const anchor = document.createElement("a");
+        anchor.href = window.URL.createObjectURL(blob);
+        anchor.download = filename;
+        anchor.click();
+        anchor.remove();
+      },
+      error: function(_xhr, _textStatus, errorThrown) {
+        alert(errorThrown);
+      }
+    });
   });
 };
+
+function getResponseFilename(contentDisposition) {
+  for (let part of contentDisposition.split(";")) {
+    if (part.includes("filename=")) {
+      return part.replace("filename=", "").trim();
+    }
+  }
+}
 
 ExporterManager.prepareExportRowsForm = function(format, includeEngine, ext, csrfToken) {
   let name = encodeURI(ExporterManager.stripNonFileChars(theProject.metadata.name));


main/webapp/modules/core/scripts/project/exporters.js
```diff
@@ -136,13 +136,41 @@ ExporterManager.stripNonFileChars = function(name) {
...
+        const filename = getResponseFilename(xhr.getResponseHeader("Content-Disposition"));
+        const blob = new Blob([data], {type: xhr.getResponseHeader("Content-Type")});
+        const anchor = document.createElement("a");
+        anchor.href = window.URL.createObjectURL(blob);
+        anchor.download = filename;
+        anchor.click();
+        anchor.remove();
+      },
+      error: function(_xhr, _textStatus, errorThrown) {
+        alert(errorThrown);
```
tfmorris, line 160: alerts are ugly and not tailorable, so we've been trying to move away from them.  Using `DialogSystem.alert()` is a step in this direction.

```suggestion
        DialogSystem.alert(errorThrown);
```

SoryRawyer: Ah thank you, I was looking for something like this but didn't find anything. I'll make that change.

SoryRawyer:
I think the Java error is only returned as part of a larger "error page" HTML blob, which made it difficult to render in the alert. I was also a little wary of not having a translation available for the error, but I think the most productive way forward would be to return the error message on its own, present that to the user with `DialogSystem.alert`, and defer the question of translating backend errors for another time.

tfmorris:
It's been a while since I've looked at the error handling, so I don't know off the top of my head if there's an easy way to deal with (ie replace) the HTML error page. Ideally, something like this, which the user can trigger and which can be anticipated, should generate a translatable error. The more general case of a random exception traceback I would see being hidden in a (future) error dialog as an expandable "more details" pane under a generic translated "something bad happened, please include the information below when filing a bug report." 

This whole area needs to be refactored. If we decide to clean up the client/server protocol, that would be a good time to do it (because the refactoring would necessarily change behavior in places it's currently inconsistent).

A couple of years ago I started this process as part of rescuing some useful error reporting functionality from Antonin's branch, but it needs more work. https://github.com/OpenRefine/OpenRefine/compare/master...tfmorris:OpenRefine:875-ajax-error-handling

tfmorris:
Please hold off on working on this. I shouldn't have tried to review it by just looking at the diff and not actually running the code. Sorry! While overhauling the error handling is a good future goal, for the short term what is needed is to a) make sure that the HTML error page gets rendered correctly and b) return something more specific than a 500 server error (since it's not a bug in the code and it's not going to change if they retry).

The HTML error page rendering got messed up by either our Content-Disposition header changes or browser behavior changes (or both). The second piece needs a change in the ExcelExporter to actually check for the maximum number of rows, rather than letting it get deep into the bowels of the Apache POI library before throwing an exception.

I've got a branch that fixes both of these things (entirely on the backend), but it still needs tests written, which I won't get to until tomorrow.

Sorry for the misleading initial review!

SoryRawyer:
No worries! I was thinking we could return an error code that mapped to an internationalization key and fallback to a generic message if there's no match, which would allow us to reuse some of what's in this PR. I like your approach though, so I'll go ahead and close this.

PR #7446, opened 2025-09-22
feat(db-ext): uniform batch size handling for preview and import

### LINKED ISSUE #7444: uniform batch size handling for preview import (in db extension)
The db extension has a [property file](https://github.com/OpenRefine/OpenRefine/blob/master/extensions/database/module/MOD-INF/dbextension.properties) where one can specify the batch sizes for import
```
# Batch size for import data
preview.batchSize = 100
create.batchSize = 1000
```

These values are read and accessible via `DatabaseModuleImpl` but the method to access `getImportPreviewBatchSize` is never used, so this property has no effect.

The thing is that the same functionality to set a batch size for the preview import is implemented another time  in `DatabaseImportController` where it is hard-coded and used directly, see https://github.com/OpenRefine/OpenRefine/blob/d440aaead8414c18279b2cd6d94edbcfbe7d66a2/extensions/database/src/com/google/refine/extension/database/DatabaseImportController.java#L65

### Proposed solution
* Use the values of the properties file consistently in `DatabaseImportController` fore preview and import
* Additionally it would be great to move the functionality of `DatabaseImportController.getCreateBatchSize` into `DatabaseModuleImpl` so it wraps the single responsibility of returning a valid and useful property value (int) for both batch sizes and it does not need to be parsed / validated before every use

DaxServer:
Fixes #7444


extensions/database/src/com/google/refine/extension/database/DatabaseModuleImpl.java
```diff
@@ -65,18 +65,36 @@ public void init(ServletConfig config)
...
+        int batchSize = Integer.parseInt(DEFAULT_CREATE_PROJ_BATCH_SIZE);
         if (extensionProperties == null) {
-            return DEFAULT_CREATE_PROJ_BATCH_SIZE;
+            return batchSize;
         }
-        return extensionProperties.getProperty("create.batchSize", DEFAULT_CREATE_PROJ_BATCH_SIZE);
+        String propBatchSize = extensionProperties.getProperty("create.batchSize", DEFAULT_CREATE_PROJ_BATCH_SIZE);
+        if (propBatchSize != null && !propBatchSize.isEmpty()) {
+            try {
+                batchSize = Integer.parseInt(propBatchSize);
```
smierz: additionally check for batch sizes > 0, otherwise db import won`t work

extensions/database/src/com/google/refine/extension/database/DatabaseModuleImpl.java
```diff
@@ -65,18 +65,36 @@ public void init(ServletConfig config)
         logger.trace("Database Extension module initialization completed");
     }
 
-    public static String getImportCreateBatchSize() {
+    public static int getImportCreateBatchSize() {
+        int batchSize = Integer.parseInt(DEFAULT_CREATE_PROJ_BATCH_SIZE);
```
smierz: nit: it would be nice if the properties in the property file, their getters and their defaults follow the same naming convention, e.g. `create.batchSize` + `getCreateBatchSize` + `DEFAULT_CREATE_BATCH_SIZE`

extensions/database/src/com/google/refine/extension/database/DatabaseModuleImpl.java
```diff
@@ -65,18 +65,36 @@ public void init(ServletConfig config)
...
         if (extensionProperties == null) {
-            return DEFAULT_PREVIEW_BATCH_SIZE;
+            return batchSize;
+        }
+        String propBatchSize = extensionProperties.getProperty("preview.batchSize", DEFAULT_PREVIEW_BATCH_SIZE);
+        if (propBatchSize != null && !propBatchSize.isEmpty()) {
+            try {
+                batchSize = Integer.parseInt(propBatchSize);
+            } catch (NumberFormatException nfe) {
+
```
smierz, line 80: a log statement within the catch clause (maybe also when switching to default) would help users identify potential problems

extensions/database/src/com/google/refine/extension/database/DatabaseModuleImpl.java
```diff
@@ -65,18 +65,36 @@ public void init(ServletConfig config)
...
+            }
+        }
+        return batchSize;
     }
 
-    public static String getImportPreviewBatchSize() {
+    public static int getImportPreviewBatchSize() {
+        int batchSize = Integer.parseInt(DEFAULT_PREVIEW_BATCH_SIZE);
```
smierz: instead of parsing to `int`, you can make the default variables into (final) integers

extensions/database/tests/src/com/google/refine/extension/database/DatabaseModuleImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+        Field createBatchSizeField = DatabaseModuleImpl.class.getDeclaredField("DEFAULT_CREATE_PROJ_BATCH_SIZE");
+        createBatchSizeField.setAccessible(true);
+        originalCreateBatchSize = (String) createBatchSizeField.get(null);
+
+        Field previewBatchSizeField = DatabaseModuleImpl.class.getDeclaredField("DEFAULT_PREVIEW_BATCH_SIZE");
+        previewBatchSizeField.setAccessible(true);
+        originalPreviewBatchSize = (String) previewBatchSizeField.get(null);
+
+        // Set mocked default values
+        createBatchSizeField.set(null, MOCKED_DEFAULT_CREATE_BATCH_SIZE);
```
SoryRawyer: I think I'm missing something here. Why would we want to replace the default batch sizes?

DaxServer: I was thinking of a scenario where the default is changed from 100 to something else and thus did not want to hardcode 100 in the tests. Forgot that I could've just used the constant from the main class 😅

SoryRawyer (approved):
Looks great, thanks!

smierz:
What about my comment about making the defaults integers instead of strings? It was not adressed

SoryRawyer:
Ah sorry! I saw that it said "outdated" on the code preview window and thought that was addressed.

extensions/database/src/com/google/refine/extension/database/DatabaseModuleImpl.java
```diff
@@ -65,18 +65,37 @@ public void init(ServletConfig config)
...
-    public static String getImportPreviewBatchSize() {
+    private static int getBatchSize(String propertyName, String defaultValue) {
+        int batchSize = Integer.parseInt(defaultValue);
+
         if (extensionProperties == null) {
-            return DEFAULT_PREVIEW_BATCH_SIZE;
+            return batchSize;
+        }
+
+        String propBatchSize = extensionProperties.getProperty(propertyName, defaultValue);
```
SoryRawyer: I think we should be able to remove the default value here. The if statement on the next line should still be fine since `getProperty` will still return null if there's no value.

SoryRawyer (changes requested):
Apologies for the original oversight, but I agree with @smierz that the default batch sizes could be final integers instead of strings.

smierz:
Thanks for taking this on, I especially appreciate that you took a previously untested piece of code and thoroughly covered it. 

So far the code is sound but I would like to point out one aspect that could be improved: the code of `DatabaseModuleImpl.getBatchSize` is a bit hard to read/follow and thus to maintain. We're taken on this long-ish road following the adventures of the variable `batchSize` which we have to read from start to finish to get to a result (not meant derogatory, just trying to illustrate the point).

Compare this to a more defensive approach where the focus is to first cover everything that could go wrong and return a value asap:
```java
  private static int getBatchSize(String propertyName, int defaultValue) {

      // case: no properties exist
      if (extensionProperties == null) {
          return defaultValue;
      }

      // case: no property of propertyName exists
      String propBatchSize = extensionProperties.getProperty(propertyName);
      if (propBatchSize == null || propBatchSize.isEmpty()) {
          return defaultValue;
      }

      try { // happy path
          int batchSize = Integer.parseInt(propBatchSize);
          if (batchSize > 0) {
              return batchSize;
          }
      } catch (NumberFormatException nfe) {
          logger.warn("Error parsing {} property as Integer ({})", propertyName, propBatchSize);
      }

      // catch all
      return defaultValue;
  }
```

imo this is easier to maintain and read because
* if you wanted to extend it you simply add one more condition, 
* if you wanted to find a bug or follow the code, you can only focus on one of the conditions while ignoring the rest of the code because they are kind of independent of each other

What do you think?

DaxServer:
> Thanks for taking this on, I especially appreciate that you took a previously untested piece of code and thoroughly covered it.

Thanks. It's the AI, not me though ;)

tfmorris:
@smierz As far as I can see all of your suggestions were addressed. Do you mind taking another quick look to see if you agree?

extensions/database/tests/src/com/google/refine/extension/database/DatabaseModuleImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Test(dataProvider = "batchSizeTestData")
+    public void testBatchSizes(String propertyName, String propertyValue, String methodType,
+            int expectedValue, String message) throws Exception {
+        Properties testProps = new Properties();
+        testProps.setProperty(propertyName, propertyValue);
+        DatabaseModuleImpl.extensionProperties = testProps;
+
+        int result;
+        if ("create".equals(methodType)) {
```
smierz: parameter `methodType` is not needed as it is always the same as the prefix of `propertyName`
--> simply use 
`if ("create.batchSize".equals(propertyName)) { ...` 
and remove parameter `methodType` from method

smierz:
functionality looks good to me, one small note about tests

PR #7448, opened 2025-09-22
escape special characters to deserialize changes for column renaming operation

### LINKED ISSUE #7341: Unable to deserialize changes for column renaming operation
This is an analysis of the change loading bug reported [on the forum](https://forum.openrefine.org/t/fail-to-load-change-file/2365) by @robertgarrigos.

The problem is in the serialization of column rename changes, which implicitly assumes that neither the old nor the new column names contain newline characters.
When they do, OpenRefine will not be able to reload the change after being restarted, because they look like this:

```
3.10-SNAPSHOT
com.google.refine.model.changes.ColumnRenameChange
oldColumnName=_ - a
b
newColumnName=new name
/ec/
```

The change format should be changed so that it can reliably store newline characters. Ideally the deserialization code could be changed not to fail in this example, but the current serialization format is ambiguous (for instance as the old column name could contain `\nnewColumnName=…`). Those changes should be done in a backwards-compatible way, meaning that changes made with older versions of OpenRefine should still load as well as possible in newer versions.
The fact that the first line of the change contains the OpenRefine version could be used to switch deserialization codes accordingly.

### To Reproduce
1. First, create a project from the following JSON (to be pasted in the clipboard import tool, and then select JSON as a format):
```
[ {"a\nb":1, "c": 2} ]
```
2. Then, rename the first column (for instance to "new name")
3. Shut OpenRefine down (not just by closing the browser: also shut the server down)
4. Restart OpenRefine
5. Reload the project you just created
6. Try to undo the renaming operation in the `Undo / Redo` tab

### Current Results

You get a JS alert:
```
java.lang.RuntimeException: Failed to load change file /home/user/.local/share/openrefine/2342171871129.project/history/1749905084011.change.zip
```

### Expected Behavior

The change should undo

### Versions
 - Operating System: any
 - Browser Version: any
 - JRE or JDK Version: any
 - OpenRefine: likely any (the bug seems to have been there from the start)

DevangJagdale:
Fixes #7341


tfmorris:
p.s. because this is likely a much bigger task than you anticipated, you are welcome to switch to a more manageable sized task, if that's your preference.

tfmorris:
Thanks for additional updates. I apologize for not noticing before that this PR was created from your master branch, but that's going to be very susceptible to inadvertent changes. Could you please move these to a feature branch and open a new PR based on that?

You can do that by something along the lines of 

```
git remote update
git checkout master
git checkout -b 7341-column-name-escape
```

to create the new branch and then commit your changes to that, perhaps by using `git cherry-pick` (although you might want to use `git rebase -i` to squash some of the commits together so that you have fewer commits to cherry pick.

I will start my review in parallel with that process.

DevangJagdale:
https://github.com/OpenRefine/OpenRefine/pull/7461
use this PR instead

PR #7461, opened 2025-09-25
Escape special characters when deserializing column renames

### LINKED ISSUE #7341: Unable to deserialize changes for column renaming operation

DevangJagdale:
Fixes #7341 

- Handle newlines, carriage returns, backslashes
- Preserve legacy and corrupted formats
- Add comprehensive tests

https://github.com/user-attachments/assets/eccdfda9-c0b4-424d-a98d-2cdc1f5de754

### DIFF
--- a/modules/core/src/main/java/com/google/refine/model/changes/ColumnRenameChange.java
+++ b/modules/core/src/main/java/com/google/refine/model/changes/ColumnRenameChange.java
@@ -74,10 +74,10 @@ public class ColumnRenameChange extends ColumnChange {
     @Override
     public void save(Writer writer, Properties options) throws IOException {
         writer.write("oldColumnName=");
-        writer.write(_oldColumnName);
+        writer.write(escape(_oldColumnName));
         writer.write('\n');
         writer.write("newColumnName=");
-        writer.write(_newColumnName);
+        writer.write(escape(_newColumnName));
         writer.write('\n');
         writer.write("/ec/\n"); // end of change marker
     }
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/com/google/refine/model/changes/ColumnRenameChange.java
```diff
@@ -74,10 +74,10 @@ public void revert(Project project) {
     @Override
     public void save(Writer writer, Properties options) throws IOException {
         writer.write("oldColumnName=");
-        writer.write(_oldColumnName);
+        writer.write(escape(_oldColumnName));
```
tfmorris, line 77:
```suggestion
        writer.write(StringEscapeUtils.escapeJson(_oldColumnName));
```

There are plenty of JSON escape/unescape packages around, so no need to write our own. This one is from

```
import org.apache.commons.text.StringEscapeUtils;
```

modules/core/src/main/java/com/google/refine/model/changes/ColumnRenameChange.java
```diff
@@ -74,10 +74,10 @@ public void revert(Project project) {
...
         writer.write('\n');
         writer.write("newColumnName=");
-        writer.write(_newColumnName);
+        writer.write(escape(_newColumnName));
```
tfmorris, line 80:
```suggestion
        writer.write(StringEscapeUtils.escapeJson(_newColumnName));
```

modules/core/src/main/java/com/google/refine/model/changes/ColumnRenameChange.java
```diff
@@ -89,18 +89,176 @@ static public Change load(LineNumberReader reader, Pool pool) throws Exception {
...
+                        i++;
+                        break;
+                }
+            } else {
+                sb.append(c);
+            }
+        }
+        return sb.toString();
+    }
+
```
tfmorris, line 255: This can be deleted
```suggestion
```

modules/core/src/main/java/com/google/refine/model/changes/ColumnRenameChange.java
```diff
@@ -89,18 +89,176 @@ static public Change load(LineNumberReader reader, Pool pool) throws Exception {
...
+                    if (c < 0x20 || c > 0x7E) {
+                        sb.append(String.format("\\u%04x", (int) c));
+                    } else {
+                        sb.append(c);
+                    }
+            }
+        }
+        return sb.toString();
+    }
+
```
tfmorris, line 201: This can be deleted.
```suggestion
```

modules/core/src/main/java/com/google/refine/model/changes/ColumnRenameChange.java
```diff
@@ -89,18 +89,176 @@ static public Change load(LineNumberReader reader, Pool pool) throws Exception {
...
+                }
+            } else {
+                sb.append(c);
+            }
+        }
+        return sb.toString();
+    }
+
+    // Getters for testing
+    public String getOldColumnName() {
```
tfmorris, line 257:
```suggestion
   String getOldColumnName() {
```

Since this is for tests only, lets reduce its visibility

modules/core/src/main/java/com/google/refine/model/changes/ColumnRenameChange.java
```diff
@@ -89,18 +89,176 @@ static public Change load(LineNumberReader reader, Pool pool) throws Exception {
...
+        return _oldColumnName;
+    }
 
-        return change;
+    public String getNewColumnName() {
```
tfmorris, line 261:
```suggestion
   String getNewColumnName() {
```

modules/core/src/main/java/com/google/refine/model/changes/ColumnRenameChange.java
```diff
@@ -89,18 +89,176 @@ static public Change load(LineNumberReader reader, Pool pool) throws Exception {
...
+     * Process a field value, handling both new escaped format and legacy format. For legacy format, handle potential
+     * corruption from newline characters.
+     */
+    private static String processFieldValue(String value, LineNumberReader reader) throws IOException {
+        if (containsEscapeSequences(value)) {
+            return unescape(value);
+        } else {
+            // Legacy format - might be corrupted if original value contained newlines
+            return handleLegacyValue(value, reader);
+        }
```
tfmorris, line 120: I need to think about this more. I think it would be better to use the version information written to the change file, but I'm not sure how reliable it is and how difficult it would be to make use of.

modules/core/src/main/java/com/google/refine/model/changes/ColumnRenameChange.java
```diff
@@ -89,18 +89,176 @@ static public Change load(LineNumberReader reader, Pool pool) throws Exception {
...
+     * Check if a value contains escape sequences that indicate new format
+     */
+    private static boolean containsEscapeSequences(String value) {
+        return value.contains("\\n")
+                || value.contains("\\r")
+                || value.contains("\\t")
+                || value.contains("\\b")
+                || value.contains("\\f")
+                || value.matches(".*\\\\u[0-9a-fA-F]{4}.*")
+                || value.contains("\\\\");
```
tfmorris, line 162: If we end up using this, it can be simplified into a single regex. (but wait until we decide whether we need it at all)

modules/core/src/test/java/com/google/refine/model/changes/ColumnRenameChangeTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+    @Test
+    public void testEscapeNewlinesRoundTrip() throws Exception {
+        String oldName = "col\nwith\nlines";
+        String newName = "new\ncol";
+
+        ColumnRenameChange change = new ColumnRenameChange(oldName, newName);
+
+        Writer writer = new StringWriter();
+        change.save(writer, new Properties());
```
tfmorris, line 82:
```suggestion
        History.writeOneChange(writer, change, new Pool());
```

Using the same method as the production code will make this more authentic

modules/core/src/test/java/com/google/refine/model/changes/ColumnRenameChangeTest.java
```diff
@@ -0,0 +1,218 @@
...
+        String serialized = writer.toString();
+        assertEquals(
+                serialized,
+                "oldColumnName=col\\nwith\\nlines\n" +
+                        "newColumnName=new\\ncol\n" +
+                        "/ec/\n");
+
+        LineNumberReader reader = new LineNumberReader(new StringReader(serialized));
+        ColumnRenameChange crc = (ColumnRenameChange) ColumnRenameChange.load(reader, new Pool());
```
tfmorris, line 91:
```suggestion
        ColumnRenameChange crc = (ColumnRenameChange) History.readOneChange(reader, new Pool());
```
Using the same method as the production code will make this more authentic and also potentially provide access to the version info.

modules/core/src/test/java/com/google/refine/model/changes/ColumnRenameChangeTest.java
```diff
@@ -0,0 +1,218 @@
...
+        String newName = "new\ncol";
+
+        ColumnRenameChange change = new ColumnRenameChange(oldName, newName);
+
+        Writer writer = new StringWriter();
+        change.save(writer, new Properties());
+        String serialized = writer.toString();
+        assertEquals(
+                serialized,
+                "oldColumnName=col\\nwith\\nlines\n" +
```
tfmorris, line 86:
```suggestion
                RefineServlet.VERSION + "\n" + change.getClass().getCanonicalName() + "\"
                        "oldColumnName=col\\nwith\\nlines\n" +
                "oldColumnName=col\\nwith\\nlines\n" +
```

The tests will need to be adjusted to mirror the on-disk format used in production with the version and the classname in the first two lines.

However, there's a problem with using `RefineServlet.VERSION` in a non-servlet context for the tests, because it doesn't get initialized correctly.  It can be fixed, but needs changes to RefineServlet as well.

modules/core/src/test/java/com/google/refine/model/changes/ColumnRenameChangeTest.java
```diff
@@ -0,0 +1,218 @@
...
+import com.google.refine.model.Project;
+import com.google.refine.util.Pool;
+
+public class ColumnRenameChangeTest extends RefineTest {
+
+    private Project project;
+
+    @BeforeTest
+    public void init() {
+        logger = org.slf4j.LoggerFactory.getLogger(this.getClass());
```
tfmorris, line 33:
```suggestion
        logger = org.slf4j.LoggerFactory.getLogger(this.getClass());
        // VERSION doesn't get set properly if we're not running in a servlet context
        RefineServlet.initializeVersions(this.getClass().getClassLoader());
```

This one part of the change need to get the VERSION initialized properly. The other piece is to refactor the `init()` method in RefineServlet to extract this method so that it can be used here. See my main comment above for what the body of this method should look like.

modules/core/src/test/java/com/google/refine/model/changes/ColumnRenameChangeTest.java
```diff
@@ -0,0 +1,218 @@
...
+        assertNotNull(reverted);
+        assertEquals(reverted.getName(), "original_column");
+    }
+
+    @Test
+    public void testLoadFromResource() throws Exception {
+        // Skip this test if resource doesn't exist
+        InputStream in = getClass().getClassLoader()
+                .getResourceAsStream("changes/column_rename_legacy.txt");
+        if (in != null) {
```
tfmorris, line 193:
```suggestion
        InputStream in = getClass().getClassLoader()
                .getResourceAsStream("changes/column_rename_legacy.txt");
        assertNotNull(in)
```

This should either include the test data file or be removed. Preferably the former.

modules/core/src/test/java/com/google/refine/model/changes/ColumnRenameChangeTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+        ColumnRenameChange change = new ColumnRenameChange(oldName, newName);
+
+        Writer writer = new StringWriter();
+        change.save(writer, new Properties());
+        String serialized = writer.toString();
+
+        // Print actual value for debugging
+        System.out.println("Actual serialized: " + serialized);
+
```
tfmorris, line 132: Remove the debug print statement

```suggestion
```

tfmorris:
p.s. the way that you had the tests written originally to call `Change.load()` directly was totally reasonable for a unit test and arguably better since it has less coupling to other classes, but I think we're going to need the additional context to for the ultimate solution.

SoryRawyer:
Hey @DevangJagdale, I just wanted to check in and see if you had any questions about the feedback so far. Is there anything we can help with to keep this moving along?

SoryRawyer:
I'm going to close this due to inactivity. We can always reopen if and when needed.

PR #7464, opened 2025-09-26
Apply scoped cursor override in project.css to fix header cursor (#6933)

### LINKED ISSUE #6933: Cursor style changes from drag handle to arrow handle between odd/even elements in header group
I noticed that the cursor style changes from the drag handle style back to an arrow style on any of the odd/even <th> elements in groups.

### To Reproduce
1. First, hover mouse over a columns' edge
2. Then, move mouse up or down following the columns' edge in a column group

### Current Results
cursor style changes while following column edge, depending on if edge is somehow considered NOT part of a resizeable group or element? Dunno, tried to narrow the context as much as possible, but seemed to be most annoying in column header groups in Records mode.

### Expected Behavior
cursor style stays as a drag handle while moving along any column edge.

### Screenshots
CursorChangesOnHeaderRows

### Versions
 - JRE or JDK Version: openjdk 21
 - OpenRefine: master (trunk)

### Datasets

You can use this OpenRefine Project during testing or import some simple JSON record hierarchy:
https://drive.google.com/file/d/1FP2UcCkOn0TscImADL7FZdrDqbSd_nlo/view?usp=drive_link

DevangJagdale:
Fixes #6933

### DIFF
--- a/main/webapp/modules/core/styles/project.css
+++ b/main/webapp/modules/core/styles/project.css
@@ -128,3 +128,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #facet-panel, #history-panel {
   overflow: auto;
 }
+
+.data-table-header .column-header-resizer-left,
+.data-table-header .column-header-resizer-right,
+.data-table-header .ui-sortable-handle {
+  cursor: col-resize;
+}


SoryRawyer:
Hey @DevangJagdale, thanks for the contribution! Would you mind filling out the PR template? I've found that authors describing why they made a specific change is really helpful for me when I review.

Also, if you're opening a pull request through a workflow that doesn't prompt you with the template, would you mind sharing how you open pull requests? I personally rely mainly on the GitHub web UI to open pull requests but I know there are other ways to handle this so I'm curious to see what they're like.

DevangJagdale:
Hi! Thanks for the feedback, I'll fill out the PR template right away!

Regarding my workflow: I primarily use GitHub Desktop and GitHub CLI for all my local development work (branching, commits, rebasing, etc.) up until the PR creation stage. You're absolutely right about the template issue, when using GitHub Desktop to create PRs, it sometimes auto populates with my commit message description instead of showing the proper PR template, which can bypass the structured format you're looking for.

tfmorris (changes requested):
Why did you choose to place these changes in project.css? I would have thought that data-table-vew.css was a much more logical place, particularly since it would allow for review of the new CSS in the context of all the existing table header markup.

Also, although I'm not a front end expert, I'm a little suspicious that a better solution might be to make some of the existing CSS selectors more specific rather than layering on more overrides. Did you investigate that at all?

Lastly, if a code generation LLM (ie "AI") was used for any piece of this, please mention which one and how it was used. Thanks.

DevangJagdale:
I started by inspecting the header elements in the browser’s DevTools to see exactly which CSS rules were applying (and missing) the resize cursor on odd versus even groups. From there, I chose a small, scoped override in `project.css` so I could quickly enforce the correct `col-resize` cursor without touching the tangled existing selectors or JavaScript rendering code, keeping the rest of the layout safe while fixing the issue everywhere it appears. It was a rapid way to isolate the cursor fix and confirm it worked everywhere before refactoring. 
No LLM used

tfmorris:
Thanks for the clarification. Let's still move the new CSS into the appropriate file to make it easier to review in the context of the related CSS. https://github.com/OpenRefine/OpenRefine/blob/54b7050d418e5cdebdfbe1a57c50055ba2f1bacd/main/webapp/modules/core/styles/views/data-table-view.css#L95-L113

https://github.com/OpenRefine/OpenRefine/blob/54b7050d418e5cdebdfbe1a57c50055ba2f1bacd/main/webapp/modules/core/styles/views/data-table-view.css#L154-L163

SoryRawyer:
Hi @DevangJagdale, were you able to test this on Windows using the Edge browser? If so, would you mind uploading a screenshot or short video to show the before and after? The CSS looks good to me but I don't notice any difference in behavior between this branch and the current master branch when testing with Firefox and Chrome on Ubuntu.

SoryRawyer:
I'm going to close this pull request due to inactivity. We can always reopen in the future if and when we want to revisit this.

PR #7477, opened 2025-10-02
switch to new mysql connector coordinates

### LINKED ISSUE #7469: Switch to new coordinates for MySQLConnector
The dependency to mysql-connector-java has new coordinates since release 8.0.31:

> Important Change: To comply with proper naming guidelines, the Maven groupId and artifactId for Connector/J have been changed to the following since release 8.0.31:
> 
>     groupId: com.mysql
> 
>     artifactId: mysql-connector-j
> 
> The old groupId and artifactId can no longer be used for linking the Connector/J library starting with this release. Please make sure you switch to the new coordinates. 

Source: https://dev.mysql.com/doc/relnotes/connector-j/en/news-8-1-0.html

### Current Results
Dependency remains on version 8.0.30

### Expected Behavior
Dependency gets picked up by dependabot.

smierz:
fixes #7469 

- switch to new maven coordinates for dependency `mysql-connector-j` 
(only went up 8.0.30 -> 8.0.33 to make sure there are no compatibility issues)
- additionally the mysql connector documentation notes that there is a possible exclusion of `protobuf-java` when xdevapi is not used (which is the case in db extension), so I excluded it 

Sources:
- announcement in mysql connector release notes:
https://dev.mysql.com/doc/relnotes/connector-j/en/news-8-0-31.html
- note about movement of artifact on maven repository website: 
https://mvnrepository.com/artifact/mysql/mysql-connector-java
- Installing Connector/J Using Maven (exclusion): 
https://dev.mysql.com/doc/connector-j/en/connector-j-installing-maven.html

### DIFF
--- a/extensions/database/pom.xml
+++ b/extensions/database/pom.xml
@@ -99,7 +99,7 @@
     <dependency>
       <groupId>com.mysql</groupId>
       <artifactId>mysql-connector-j</artifactId>
-      <version>8.0.33</version>
+      <version>8.2.0</version>
       <exclusions>
         <exclusion>
           <!-- Only needed for the X DevAPI features -->


extensions/database/pom.xml
```diff
@@ -97,9 +97,15 @@
...
     </dependency>
     <dependency>
-      <groupId>mysql</groupId>
-      <artifactId>mysql-connector-java</artifactId>
-      <version>8.0.30</version>
+      <groupId>com.mysql</groupId>
+      <artifactId>mysql-connector-j</artifactId>
+      <version>8.0.33</version>
+      <exclusions>
+        <exclusion>
```
tfmorris, line 104:
```suggestion
+        <!-- Only needed for the X DevAPI features -->
        <exclusion>
```

Let's add a comment here, so we remember why it's excluded.

tfmorris (changes requested):
Just one small request for a comment to help our future memory.

I agree with the conservative approach to version updates, but we're already 2 1/2 years behind the times, so it would be great if someone who is knowledgeable about MySQL and uses it regularly could follow up with a separate change to bring us current.

smierz:
I wouldn't mind going up as high as version 8.2.0 in this PR

* afaict the "removal of auto-deserialization for BLOBs" in [8.2.0](https://dev.mysql.com/doc/relnotes/connector-j/en/news-8-2-0.html) would only be relevant when using the connection property `autoDeserialize=true` (see [deprecation announcement](https://dev.mysql.com/doc/relnotes/connector-j/en/news-8-1-0.html) for 8.1.0) which is not the case in db extension 
  * was also the cause for a vulnerability fixed in 8.2.0 ([maven repo overview](https://mvnrepository.com/artifact/com.mysql/mysql-connector-j))
* in 8.3.0 support for mysql server 5.7. [dropped](https://dev.mysql.com/doc/relnotes/connector-j/en/news-8-3-0.html), this may need discussion

tfmorris:
> I wouldn't mind going up as high as version 8.2.0 in this PR

OK, let's do that then. We don't support the BLOB datatype, so I don't think the auto-deserialization deprecation is an issue.

PR #7478, opened 2025-10-02
fix mysql driver class

### LINKED ISSUE #5794: MySQL driver class is deprecated
```
Loading class `com.mysql.jdbc.Driver'. This is deprecated. The new driver class is `com.mysql.cj.jdbc.Driver'. The driver is automatically registered via the SPI and manual loading of the driver class is generally unnecessary.
```

Seen in build logs:  https://github.com/OpenRefine/OpenRefine/actions/runs/4733389973/jobs/8400853802?pr=5638#step:9:2814

smierz:
fixes #5794

- change class name of mysql driver from "com.mysql.jdbc.Driver" to "com.mysql.cj.jdbc.Driver"
- the warning is triggered by this line in `MySQLConnectionManager`:
https://github.com/OpenRefine/OpenRefine/blob/b3693e0e564f1f287c8907f076d396e3ef21e72e/extensions/database/src/com/google/refine/extension/database/mysql/MySQLConnectionManager.java#L133
To have a quick way to reproduce the warning and see that it disappears after the change I added the "test", but the commit "reproduce warning" can be dropped before merging.

Source: https://dev.mysql.com/doc/connectors/en/connector-j-upgrading-to-8.0.html

### DIFF
--- a/extensions/database/src/com/google/refine/extension/database/mysql/MySQLDatabaseService.java
+++ b/extensions/database/src/com/google/refine/extension/database/mysql/MySQLDatabaseService.java
@@ -52,7 +52,7 @@ public class MySQLDatabaseService extends DatabaseService {
 
     private static final Logger logger = LoggerFactory.getLogger("MySQLDatabaseService");
     public static final String DB_NAME = "mysql";
-    public static final String DB_DRIVER = "com.mysql.cj.jdbc.Driver";
+    public static final String DB_DRIVER = com.mysql.cj.jdbc.Driver.class.getCanonicalName();
 
     private static MySQLDatabaseService instance;
 


extensions/database/src/com/google/refine/extension/database/mysql/MySQLDatabaseService.java
```diff
@@ -52,7 +52,7 @@ public class MySQLDatabaseService extends DatabaseService {
 
     private static final Logger logger = LoggerFactory.getLogger("MySQLDatabaseService");
     public static final String DB_NAME = "mysql";
-    public static final String DB_DRIVER = "com.mysql.jdbc.Driver";
+    public static final String DB_DRIVER = "com.mysql.cj.jdbc.Driver";
```
tfmorris:
```suggestion
    public static final String DB_DRIVER = com.mysql.cj.jdbc.Driver.class.getCanonicalName();
```

I dislike hiding class names in strings unless there's no choice. Is there any reason for us to not just use the class reference so that our tooling can help us track dependencies?

(I understand that this problem was not introduced by your PR)

tfmorris (approved):
I went ahead and updated the PR based on my suggestion and dropped the test. 

I've created #7504 to address the driver class references in the rest of the database drivers.

PR #7494, opened 2025-10-14
replace png in db extension with svg

smierz:
part of #6538 

**Changes proposed in this pull request**
- unused file `more_option-vert-16.png` is removed (no need for transformation)
- additionally the glyphicon font "halflings" is removed as well, these files were previously used in `bootstrap.css` which was removed in https://github.com/OpenRefine/OpenRefine/pull/5550 and are now unused
- only remaining png `more-option-horiz-16.png` is transformed to svg and gets a shorter name as the vertical one was removed

Before:
more-options-before

After:
more-options-after


extensions/database/module/styles/database-import.css
```diff
@@ -161,7 +161,7 @@
 .sc-context-more-vert:after {
```
tfmorris, line 161:
```suggestion
.sc-context-more:after {
```

I'm not sure changing this is worth it, but it caught my eye because of the icon name change. If it gets changed here, it would also mean changing the four references in `database-source-ui.js` and `database-extension.js` (which probably means it's not worth the hassle, but I'll leave it up to you.)

PR #7497, opened 2025-10-16
Add generalized support for Apache Compress formats

### LINKED ISSUE #2741: Support more compression & archive formats

tfmorris:
Fixes #2741

- Add test for Unix compress .Z files (supported by base Apache Compress)
- add optional dependency XZ to enable support for XZ, LZMA, & 7zip
- SevenZip decompressor requires special handling since it uses
  a SeekableByteChannel rather than a Stream
- Add xz to the list of compressors that support concatenated streams
- Add ZStandard optional dependency for Apache Compress and zstd tests

### DIFF
--- a/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
+++ b/modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
@@ -820,20 +820,21 @@ public class ImportingUtilities {
 
     /*
      * Apache Compress docs recommend using ZipFile over ZipArchiveInputStream, which is what ArchiveStreamFactory
-     * returns so we handle it separately
+     * returns, so we handle it separately
      */
     private static void explodeZip(File rawDataDir, ObjectNode archiveFileRecord, ArrayNode fileRecords, Progress progress, FileChannel fc)
             throws IOException {
-        ZipFile zf = new ZipFile.Builder().setSeekableByteChannel(fc.position(0)).get();
-        for (Iterator<ZipArchiveEntry> it = zf.getEntries().asIterator(); it.hasNext();) {
-            ZipArchiveEntry entry = it.next();
-            if (progress.isCanceled()) {
-                break;
-            }
-            if (!entry.isDirectory()) {
-                ObjectNode fileRecord2 = processArchiveEntry(rawDataDir, archiveFileRecord, progress, entry.getName(),
-                        zf.getInputStream(entry));
-                JSONUtilities.append(fileRecords, fileRecord2);
+        try (ZipFile zf = new ZipFile.Builder().setSeekableByteChannel(fc.position(0)).get()) {
+            for (Iterator<ZipArchiveEntry> it = zf.getEntries().asIterator(); it.hasNext();) {
+                ZipArchiveEntry entry = it.next();
+                if (progress.isCanceled()) {
+                    break;
+                }
+                if (!entry.isDirectory()) {
+                    ObjectNode fileRecord2 = processArchiveEntry(rawDataDir, archiveFileRecord, progress, entry.getName(),
+                            zf.getInputStream(entry));
+                    JSONUtilities.append(fileRecords, fileRecord2);
+                }
             }
         }
     }


modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -108,6 +111,7 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
 
 public class ImportingUtilities {
 
+    public static final int BUFFER_SIZE = 128 * 1024;
```
SoryRawyer: Does this get used anywhere?

tfmorris: Thanks for the review. Good catch on the unused constant. I wonder why Copilot wanted me to comment it instead of just removing it. :(

Unfortunately, I'm not convinced that the code for compressed archives (e.g. .tar.gz) actually works and we don't seem to have a test for it either. So, I'm going to at least add a test and, probably, fix the code.

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -108,6 +111,7 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
 
 public class ImportingUtilities {
```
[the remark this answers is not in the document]
tfmorris: Constant is unused and has been removed, so no need to document.

main/tests/server/src/com/google/refine/importing/ImportingUtilitiesTests.java
```diff
@@ -107,9 +106,9 @@ public void createProjectMetadataTest()
         ObjectNode optionObj = ParsingUtilities.evaluateJsonStringToObjectNode(
                 "{\"projectName\":\"acme\",\"projectTags\":[],\"created\":\"2017-12-18T13:28:40.659\",\"modified\":\"2017-12-20T09:28:06.654\",\"creator\":\"\",\"contributors\":\"\",\"subject\":\"\",\"description\":\"\",\"rowCount\":50,\"customMetadata\":{}}");
         ProjectMetadata pm = ImportingUtilities.createProjectMetadata(optionObj);
-        Assert.assertEquals(pm.getName(), "acme");
-        Assert.assertEquals(pm.getEncoding(), "UTF-8");
-        Assert.assertTrue(pm.getTags().length == 0);
+        assertEquals(pm.getName(), "acme");
+        assertEquals(pm.getEncoding(), "UTF-8");
+        assertEquals(pm.getTags().length, 0);
```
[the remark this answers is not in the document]
tfmorris: That's backwards.

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -796,40 +798,75 @@ static private boolean explodeArchive(
...
+            Progress progress, ArchiveInputStream archiveInputStream) throws IOException {
+        ArchiveEntry te;
+        while (!progress.isCanceled() && (te = archiveInputStream.getNextEntry()) != null) {
+            if (!te.isDirectory()) {
+                ObjectNode fileRecord2 = processArchiveEntry(rawDataDir, archiveFileRecord, progress, te.getName(),
+                        archiveInputStream);
+                JSONUtilities.append(fileRecords, fileRecord2);
+            }
+        }
+        archiveInputStream.close();
```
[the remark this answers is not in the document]
tfmorris: A real problem, but not the right solution. Fixed by adding try with resources to the caller.

tfmorris:
@SoryRawyer Sorry to ask you to review this twice. I removed the constant mentioned in your previous review, but also, more importantly, fixed the import of compressed archives (e.g. .tar.gz) and added a test to cover that case.

modules/core/src/main/java/com/google/refine/importing/ImportingUtilities.java
```diff
@@ -796,40 +798,75 @@ static private boolean explodeArchive(
...
+            throws IOException {
+        ZipFile zf = new ZipFile.Builder().setSeekableByteChannel(fc.position(0)).get();
+        for (Iterator<ZipArchiveEntry> it = zf.getEntries().asIterator(); it.hasNext();) {
+            ZipArchiveEntry entry = it.next();
+            if (progress.isCanceled()) {
+                break;
+            }
+            if (!entry.isDirectory()) {
+                ObjectNode fileRecord2 = processArchiveEntry(rawDataDir, archiveFileRecord, progress, entry.getName(),
+                        zf.getInputStream(entry));
```
SoryRawyer: Should this be in a try-with-resources block like the 7zip case below, or does that not make much of a difference here?

tfmorris: Good catch. For some reason my IDE didn't flag either of these cases, which I don't really understand, but the second one is fixed now too.

PR #7507, opened 2025-10-22
fix loading 100 rows in db import preview

### LINKED ISSUE #7499: db-extension: Preview does not load 100 rows, if batch size is smaller than 100
preview in db extension does not load 100 rows, if the batch size for the preview is smaller than 100.

### To Reproduce
1. First, set `preview.batchSize = 19` in `dbextension.properties`
2. Then, start the application and connect to a db
3. Finally, preview the data and notice that you only get 19 rows for the preview.

### Expected Behavior
Preview contains 100 entries as specified by `DEFAULT_PREVIEW_LIMIT`
https://github.com/OpenRefine/OpenRefine/blob/83d59aaa354fbd78b9c70c0b8c50cd306d322cb8/extensions/database/src/com/google/refine/extension/database/DatabaseImportController.java#L201

### Additional context
This is probably the reason, why `preview.batchSize` was previously hard-coded to 100, see
https://github.com/OpenRefine/OpenRefine/issues/7444

smierz:
fixes #7499 

**Changes proposed in this pull request**
- Currently the `end` flag that should signal the retrieval of the last batch from the db is set to `true` after loading first batch (because `dbRows.size() < batchSize + 1` is always true). Fixed by setting it to `dbRows.size() < batchSize` so the `end` flag is only set for the last batch.

- Additionally `dbRows` is checked for null in the previous if-statement, but not before setting the `end` flag this way. Fixed by putting a guard statement in directly after loading `dbRows`.

- The `end` parameter is set within `getRowsOfCells` and every call to this method after setting the flag to true should not result in additional db queries. That's why it should be checked first thing within the method instead of outside of it. Moved the flag check.

### DIFF
--- a/extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
+++ b/extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
@@ -145,7 +145,7 @@ public class DBQueryResultImportReader implements TableDataReader {
      * @throws DatabaseServiceException
      */
     private List<List<Object>> getRowsOfCells(int startRow) throws DatabaseServiceException {
-        logger.debug("Query db for next batch: [{},{}]", startRow + 1, startRow + batchSize);
+        logger.debug("Query db for next batch: [{},{}]", startRow, startRow + batchSize - 1); // 0-based
 
         // if end was already reached, do not query db again
         if (lastBatch) {


smierz:
This PR is currently only focused on fixing the batch loading of the db import preview, but there are some important additional considerations:

Before I started with the task I noticed that `DBQueryResultPreviewReader` and `DBQueryResultImportReader` both do the same task and have very similar code, the code for the method `getRowsOfCells` where the bug is located is even the same. So the question gets broadened: why is the import reader reading all batches but not the preview reader?

I investigated this by covering both in the exact same tests ([cover loading multiple batches in preview and import](https://github.com/OpenRefine/OpenRefine/pull/7507/commits/88d0f5cc21afb0b9ccb94da42d031717937d3a43)) and by comparing the remaining code in the only other method `getNextRowOfCells`  ([divide code into commented sections](https://github.com/OpenRefine/OpenRefine/pull/7507/commits/670bdc716a60ea4825791b157e271ce0722f4b6a)).
==> Notice that the import reader has one additional block of code that loads more batches by ignoring the `end` flag, which is titled "pre-load next batch before returning last row of current batch".

Up to this point I only fixed the bug in preview, but imo it would be preferable if the code changes were also transferred to the import reader. 
==> By doing that both readers would have identical code (apart from progress tracking) and could be merged into one class.

**Question**: 
We could stop at this point only fixing the preview reader, so we can get some feedback that the code changes work without breaking anything else and open a follow-up issue that merges both classes for later or we could do all in this PR.
What do you prefer?

tfmorris:
Based on your description, it sounds like this could definitely be refactored to eliminate the code duplication. In doing this depth of investigation, I suspect you've already done 3/4 of the necessary work. Multiple copies of code is almost never a good idea.

On a more general note, the database extension is in kind of rough shape and doesn't get much attention, so any time that you are willing to devote to it is greatly appreciated.

extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
```diff
@@ -147,11 +150,10 @@ public List<Object> getNextRowOfCells() throws IOException {
             }
 
         } catch (DatabaseServiceException e) {
+            // rethrow exception as IOException
             logger.error("DatabaseServiceException::{}", e);
```
tfmorris: I noticed this recently as a persistently bad pattern in the database extension where a format string expecting one or more parameters is used without any associated parameters.
```suggestion
            logger.error("DatabaseServiceException", e);
```

extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
```diff
@@ -161,12 +163,10 @@ public List<Object> getNextRowOfCells() throws IOException {
      * @throws DatabaseServiceException
      */
     private List<List<Object>> getRowsOfCells(int startRow) throws IOException, DatabaseServiceException {
-        // logger.info("Entry getRowsOfCells::startRow:" + startRow);
```
tfmorris, line 164: I'm not familiar enough with the module to know if these have value, but another possible solution would be to uncomment them and make them debug (or trace) level logs.

extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
```diff
@@ -208,14 +208,12 @@ private List<List<Object>> getRowsOfCells(int startRow) throws IOException, Data
                 }
 
                 rowsOfCells.add(rowOfCells);
-
             }
 
         }
+
         end = dbRows.size() < batchSize + 1;
```
tfmorris: This looks suspicious. Should the `+ 1` be removed to align with the other change you made?

extensions/database/tests/src/com/google/refine/extension/database/DBQueryResultImportReaderTest.java
```diff
@@ -0,0 +1,159 @@
...
+        // when
+        reader.getNextRowOfCells(); // ignore first call (header)
+
+        // simulate calls in TabularImportingParserBase.readTable
+        int count = tableCount;
+        int sizeOfLastBatch = tableCount % batchSize;
+        while (reader.getNextRowOfCells() != null) {
+            count--;
+            // assertEquals(reader.isEnd(), count < sizeOfLastBatch,
+            // "end should only be set after querying last batch, but was set after " + (tableCount - count) + " rows");
```
tfmorris: Why is this assert commented out here?

extensions/database/tests/src/com/google/refine/extension/database/DBQueryResultPreviewReaderTest.java
```diff
@@ -0,0 +1,158 @@
...
+import org.mockito.stubbing.Answer;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import com.google.refine.extension.database.model.DatabaseColumn;
+import com.google.refine.extension.database.model.DatabaseQueryInfo;
+import com.google.refine.extension.database.model.DatabaseRow;
+import com.google.refine.importing.ImportingJob;
+
+public class DBQueryResultPreviewReaderTest {
```
tfmorris: It would be good to try to avoid introducing more duplicate code. Is there a way that we can combine this into a parameterized version of the non-preview test?

extensions/database/tests/src/com/google/refine/extension/database/DBQueryResultPreviewReaderTest.java
```diff
@@ -0,0 +1,158 @@
...
+                    rowValues.add("2025-10-22");
+                    break;
+                case DOUBLE:
+                case FLOAT:
+                    rowValues.add("1.0");
+                    break;
+            }
+        }
+        return rowValues;
+    }
```
tfmorris: If it's too hard to combine these entire classes, perhaps we can at least factor out some of the common methods. Given a choice, I'd bias things towards the non-preview case, since that's the one that we expect to survive when things are merged.

extensions/database/src/com/google/refine/extension/database/DBQueryResultPreviewReader.java
```diff
@@ -132,62 +126,66 @@ public List<Object> getNextRowOfCells() throws IOException {
      * @throws DatabaseServiceException
      */
     private List<List<Object>> getRowsOfCells(int startRow) throws IOException, DatabaseServiceException {
-        // logger.info("Entry getRowsOfCells::startRow:" + startRow);
-
-        List<List<Object>> rowsOfCells = new ArrayList<List<Object>>(batchSize);
+        // if end was already reached, do not query db again
+        if (end) {
+            return new ArrayList<>();
```
tfmorris:
```suggestion
            return Collections.emptyList();
```

tfmorris (commented):
I should have looked at the code before offering an opinion on refactoring. Now I'm torn. One the one hand, it's basically a one line fix, which I'm reluctant to delay, but on the other hand, you're having to generate additional test code due to the duplication and the duplication means that there's a risk additional bugs in the parallel code. Having said that, I suspect that the refactoring will be non-trivial because the current control flow is a little messy. I don't understand why it's pre-fetching rows when there doesn't seem to be an obvious performance advantage to the extra complexity.

If you decide to stick with the current approach, the most important comments concern the commented out `assert` and the potential off-by-one problem.

smierz:
ok, I finished up the complete refactoring by 
* syncing the preview and import reader (all changes in processing of `end` flag are transferred to import reader as well)
* additionally the preloading of batches is removed because it is not necessary anymore 
* after that the preview reader may be replaced by the import reader because they have the same code (except for progress tracking)

I hope the trajectory becomes clearer and it answers some of the review comments.
Again, if this is too much change in one PR, I can also move the commits after [add log statements for debugging](https://github.com/OpenRefine/OpenRefine/pull/7507/commits/ba50d34ff31dc656f9343da2363b1ce34d35f7e1) into a separate PR.

extensions/database/tests/src/com/google/refine/extension/database/DBQueryResultPreviewReaderTest.java
smierz: the duplicate code is only temporary (preview reader and its tests will be removed), that's why I did not invest time into refactoring

extensions/database/tests/src/com/google/refine/extension/database/DBQueryResultImportReaderTest.java
smierz: because at this point the changes from the preview reader have not been introduced to the import reader, so the assertion did not hold (yet). I synced the changes between both classes and re-introduced the assertion.

extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
smierz: yes, synced the changes from preview to import reader

extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
smierz: I re-introduced debug statements with a bit more context than the ones before.
The reason I removed them was to make the preview and import reader more readable/comparable, so it becomes clear they have nearly the same code.

smierz:
>  I don't understand why it's pre-fetching rows when there doesn't seem to be an obvious performance advantage to the extra complexity.


Unfortunately the history of the extension is not included in this repo, but my guess is that both classes had the same problem with the off-by-one error, so two fixes were applied:
* preview hard-coded the `batchSize` to 100, so the problem that only one batch is loaded gets covered because the one batch holds all the needed 100 rows (see https://github.com/OpenRefine/OpenRefine/issues/7444)
* import patched the problem by introducing an additional batch loading statement that ignores the wrongly set `end` flag

extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
```diff
@@ -161,71 +146,77 @@ public List<Object> getNextRowOfCells() throws IOException {
...
-        return rowsOfCells;
 
+        end = dbRows.size() < batchSize;
+        return rowsOfCells;
     }
 
     private int calculateProgress() {
         // only use `totalCount` if it is available; otherwise progress can not be tracked
-        return useTotalForProgress ? (int) (((double) currentCount / totalCount) * 100) : -1;
+        return useTotalForProgress ? (int) (((double) nextRow / totalCount) * 100) : -1;
```
tfmorris, line 213:
```suggestion
        return useTotalForProgress ? nextRow * 100 / totalCount : -1;
```

I think this can be simplified by reordering the operations.

extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
```diff
@@ -94,64 +94,49 @@ public DBQueryResultImportReader(
...
-
+            // on first call get column names for header
             if (!usedHeaders) {
                 List<Object> row = new ArrayList<Object>(dbColumns.size());
                 for (DatabaseColumn cd : dbColumns) {
                     row.add(cd.getName());
                 }
                 usedHeaders = true;
-                // logger.info("Exit::getNextRowOfCells return header::row:" + row);
+                logger.debug("Return header on first call: {}", row.stream().map(Object::toString).collect(Collectors.joining(",")));
```
tfmorris: This is probably heavyweight enough that it should be guarded 
```suggestion
                if (logger.isDebugEnabled()) {
                    logger.debug("Return header on first call: {}", row.stream().map(Object::toString).collect(Collectors.joining(",")));
                }
```

smierz: switched to lazy eval using slfj4`s  fluent API and a supplier as parameter (https://stackoverflow.com/a/78517288)

smierz:
Some more commits addressing the review comments and minor things I noticed/refactored:
* (review comment) only changed one occurrence of returning an empty list, but there was one more 
-> [replace with ref to empty list](https://github.com/OpenRefine/OpenRefine/pull/7507/commits/d6ac840a236fe2a888a14ea8ef1a5d61a4799914) 
* we can construct the header in a one-liner
-> [refactor construction of header into one-liner](https://github.com/OpenRefine/OpenRefine/pull/7507/commits/4798e6d3486d9a54673810488bdb03a4a74c1fc7)
* (review comment) switched to lazy eval of debugging parameter using supplier  
-> [switch to lazy eval of logging parameter](https://github.com/OpenRefine/OpenRefine/pull/7507/commits/6389600ebc176555218eeaf9edfaea5eb00cd89b)
* renamed the `end` parameter to adhere to Java convention and also to clarify what it flags 
-> [rename boolean end for clarity](https://github.com/OpenRefine/OpenRefine/pull/7507/commits/ac4484a7f9c76ada141b433e81268c11719bd750)
* switched to 0-based log statements for the batch intervals 
-> [switch to 0-based logging for batches](https://github.com/OpenRefine/OpenRefine/pull/7507/commits/d3c3b8e7c0545f03509cb1f651e505070097125e)

Additionally I fixed a bunch of things IntelliJ flagged:
* [make parameters final](https://github.com/OpenRefine/OpenRefine/pull/7507/commits/0df1fa45308e57512f3f9bed4c0cb97dbd1912ef)
* (review comment) [fix parameter in error log](https://github.com/OpenRefine/OpenRefine/pull/7507/commits/8fbc39f922bb1ebc13fc7525be84337af517ab35)
* [rm IOException from method signature as it is never thrown](https://github.com/OpenRefine/OpenRefine/pull/7507/commits/dec85a6ed8200a08bd06c2b85fff0264759a3436)

extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
smierz: there is a subtle difference in both statements: 
while in the current statement both `nextRow` and `totalCount` can go the whole range of an int (as high as `Integer.MAX_VALUE`)  in the simplified statement an overflow would occur

extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
smierz: fixed it

extensions/database/src/com/google/refine/extension/database/DBQueryResultImportReader.java
tfmorris: Technically true, but Integer.MAX_VALUE is over 2 billion, which is way bigger than the maximum number of rows in a project, even considering the extra factor of 100. I'm not going to push for it to be simplified however.

tfmorris (approved):
Looks great. Thanks for all the work!

PR #7520, opened 2025-10-30
feat: Make columnNames variable available in Jython

### LINKED ISSUE #4481: Variable columnNames is not available in Jython
The variable columnNames is not available in Jython, whereas it is available in GREL. It is useful to iterate over all cells in a row, for instance with this expression:
`forEach(row.columnNames,cn,cells[cn].value)`
(see https://groups.google.com/g/openrefine/c/gV0bLUz7p7o)

### Proposed solution

We could make this variable available in Jython too.
It would also be good to [document](https://docs.openrefine.org/manual/jythonclojure) which variable names (and fields) are available in Jython, if this is really meant to differ from what is available in GREL.

### Alternatives considered

Make sure all variables and fields offered in GREL are also available in Jython and Clojure.

saitejx08:
Fixes #4481

- Added `columnNames` to the variable bindings in `ExpressionUtils.java`.
- Updated the function definition in `JythonEvaluable.java` to accept `columnNames` as an argument.
- Passed the `columnNames` variable to the Jython function call in `JythonEvaluable.java` to make it available in the expression.

### DIFF
--- a/modules/core/src/main/java/com/google/refine/expr/ExpressionUtils.java
+++ b/modules/core/src/main/java/com/google/refine/expr/ExpressionUtils.java
@@ -79,6 +79,11 @@ public class ExpressionUtils {
         bindings.put("row", new WrappedRow(project, rowIndex, row));
         bindings.put("cells", new CellTuple(project, row));
 
+        // ----- CHANGE IS HERE -----
+        if (project != null && project.columnModel != null) {
+            bindings.put("columnNames", project.columnModel.getColumnNames());
+        }
+
         if (columnName != null) {
             bindings.put("columnName", columnName);
         }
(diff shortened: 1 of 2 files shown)


SoryRawyer:
Hi @saitejx08, thanks for contributing! Would you mind updating the Jython extension tests to account for this change?

tfmorris:
@saitejx08 Apologies for leading you astray, but the feature request was invalid and should have been closed last year. We appreciate your contribution and hope you'll choose another issue to work on. When/if you do, no need to include the "Change is here" comments. We'll be able to easily spot the changes by looking at the diff.

modules/core/src/main/java/com/google/refine/expr/ExpressionUtils.java
```diff
@@ -79,6 +79,11 @@ static public void bind(Properties bindings, Row row, int rowIndex, String colum
         bindings.put("row", new WrappedRow(project, rowIndex, row));
         bindings.put("cells", new CellTuple(project, row));
 
+        // ----- CHANGE IS HERE -----
+        if (project != null && project.columnModel != null) {
+            bindings.put("columnNames", project.columnModel.getColumnNames());
```
tfmorris, line 84: This is the clue that the original request was incorrect about `columnNames` being available for GREL and not other expression languages. Because it wasn't included in the global bindings, it wasn't available for ANY languages (but is accessible via `row.columnNames`)

saitejx08:
Ok I understand it @tfmorris , thanks for the review!

PR #7528, opened 2025-11-04
Deprecate before delete

### LINKED ISSUE #7519: db-extension: remove unused methods from `DatabaseUtils`
**Status Quo**:
Within the database extension is the class `DatabaseUtils` that has several unused methods:
* `getSavedConnectionsSize`
* `decryptAll`
* `deleteAllSavedConnections`
 

**ToDo**:
If the methods are unused, they should be deprecated/removed.

smierz:
Fixes #7519
and additionally reverts the removal of public methods/classes and instead deprecates them.

- deprecate unused methods in `DatabaseUtils` 
->#7519
- deprecate public methods `getImportCreateBatchSize` and `getImportPreviewBatchSize` in `DatabaseModuleImpl` instead of deleting them
->https://github.com/OpenRefine/OpenRefine/pull/7446
- deprecate class `DBQueryResultPreviewReader` instead of deleting it
Note I used the last version before its removal : https://github.com/OpenRefine/OpenRefine/pull/7507/commits/c14b133b04e4d436bebbed28e2ea01ab2a97ef12
->  https://github.com/OpenRefine/OpenRefine/pull/7507
- and one more deprecated annotation I missed a while back for `TabularImportingParserBase.deleteEmptyColumns`

All these instances were oversights of mine, sorry! Now that they accumulated I can pack them all together in one PR.
If you deem some of the commits unnecessary, feel free to drop them.

-- 
Note: None of the reversed code deletions have been released yet.

### DIFF
--- a/extensions/database/src/com/google/refine/extension/database/DatabaseModuleImpl.java
+++ b/extensions/database/src/com/google/refine/extension/database/DatabaseModuleImpl.java
@@ -99,7 +99,7 @@ public class DatabaseModuleImpl extends ButterflyModuleImpl {
     /**
      * @deprecated use {@link #getCreateBatchSize()} instead.
      */
-    @Deprecated(since = "3.9")
+    @Deprecated(since = "3.10")
     public static String getImportCreateBatchSize() {
         return String.valueOf(getCreateBatchSize());
     }
@@ -107,7 +107,7 @@ public class DatabaseModuleImpl extends ButterflyModuleImpl {
     /**
      * @deprecated use {@link #getPreviewBatchSize()} instead.
      */
-    @Deprecated(since = "3.9")
+    @Deprecated(since = "3.10")
     public static String getImportPreviewBatchSize() {
         return String.valueOf(getPreviewBatchSize());
     }
--- a/extensions/database/src/com/google/refine/extension/database/DBQueryResultPreviewReader.java
+++ b/extensions/database/src/com/google/refine/extension/database/DBQueryResultPreviewReader.java
@@ -47,7 +47,7 @@ import com.google.refine.importing.ImportingJob;
 /**
  * @deprecated use {@link DBQueryResultImportReader} instead.
  */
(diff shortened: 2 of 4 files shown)


tfmorris:
Apologies for the delay. I thought I'd already reviewed this. :-( I'll make sure it gets done Friday, if Rory doesn't get to it first.

extensions/database/src/com/google/refine/extension/database/DatabaseModuleImpl.java
```diff
@@ -96,6 +96,22 @@ private static int getBatchSize(String propertyName, int defaultValue) {
...
+    }
+
+    /**
+     * @deprecated use {@link #getPreviewBatchSize()} instead.
+     */
+    @Deprecated(since = "3.10")
+    public static String getImportPreviewBatchSize() {
+        return String.valueOf(getPreviewBatchSize());
+    }
+
```
tfmorris, line 114: It looks like these were just added in October in 8c769bdcad345f23aef92c7e09b041abc7308f6e, so I think we can just delete them without deprecation.
```suggestion
```

tfmorris (changes requested):
I think one of these sets of deprecations is unnecessary since the methods were just added recently, but everything else looks perfect.

extensions/database/src/com/google/refine/extension/database/DatabaseModuleImpl.java
smierz: Before the linked commit you had two methods with signatures
* `public static String getImportCreateBatchSize()`
* `public static String getImportPreviewBatchSize()`

-> these methods were in db extension since addition to this repo (8 years ago).


And after the linked commit from October the methods are replaced by
* `public static int getCreateBatchSize()`
* `public static int getPreviewBatchSize()`

=> name and return type changed, 
old methods point to the new methods in their deprecation message

tfmorris (approved):
By bad! Sorry about that.

PR #7530, opened 2025-11-05
replace several png with svg

smierz:
part of #6538 

**Changes proposed in this pull request**:
- replace `menu-dropdown.png` with svg
- replace `leftpanel-showhide.png` with svg
- replace `close.png` with svg
- replace `star-flag-map.png` with svg
- replace `facet-resize-handle.png` with svg
- replace `minimize-map.png` with svg
- replace `close-map.png` with svg
- replace `checks-map.png` with svg
- replace `slider-handle.png` with svg
- replace `scatterplot-icons.png` with svg

**Screenshots**:

_**menu-dropdown.png**_ 
is used on the column headers to open the menu.
First picture set is normal menu-dropdown, second is on hover.

Before|After
menu-dropdown-before | menu-dropdown-after
menu-dropdown-hover-before | menu-dropdown-hover-after

_**leftpanel-showhide.png**_ 
is used on the sidebar to make it dis/appear.
First picture set is normal leftpanel-showhide, second is on hover.

Before|After
leftpanel-showhide-before | leftpanel-showhide-after
leftpanel-showhide-hover-before | leftpanel-showhide-hover-after

_**close.png**_ 
is used on "Open Project", "Create project with web address" and in preview window of fixed width importer. Screenshot is of the one on "Open Project":

Before|After
close-on-open-project-before | close-on-open-project-after


smierz:
**Update**:
* added more svg  (star-flag-map, facet-resize-handle, minimize-map, close-map)
* took out the deletion of replaced pngs


_**star-flag-map.png**_ 
is used in data table:

Before|After
star-flag-map-before | star-flag-map-after

**Facet icons (facet-resize-handle, minimize-map and close-map)**
are used in facet window 


Before|After
facet-window-before | facet-window-after

smierz:
**Update2**: added more svg (checks-map, slider-handle, scatterplot-icons)


_**checks-map.png**_ 
used in data table for reconcialiation

Before|After
checks-map-before | checks-map-after


_**slider-handle.png**_ 
used in numeric facet

Before|After
slider-handle-before | slider-handle-after


_**scatterplot-icons.png**_ 
used in scatterplot facet

Before|After
scatterplot-before | scatterplot-after

smierz:
tbh I don't know. After looking through the other PRs linked to the same issue, it seems mine are the only ones where the replaced PNGs were directly deleted, so I'm trying to be more conservative in my approach. 😓

tfmorris:
Given that the PNGs have direct replacements, I think we can schedule them for speedy removal, but we should probably be conservative and give the extension writers at least a little bit of warning in case they've reused them.

It might also be a good opportunity to think about and codify what our rules are for such reuse. e.g. Should extension writers make a private copy of any resources that they want to reuse from the core to insulate them from changes like this?

tfmorris (approved):
Awesome! Did you draw all those SVGs yourself? That's a lot of work!

At some point in the future, we probably want to look at replacing as many of them as possible with icons from some standard set to support our long-term goal of making the UI more themeable, but this is a great next step.

smierz:
> Given that the PNGs have direct replacements, I think we can schedule them for speedy removal, but we should probably be conservative and give the extension writers at least a little bit of warning in case they've reused them.

Should I re-add the PNGs I previously removed in https://github.com/OpenRefine/OpenRefine/pull/7226 and https://github.com/OpenRefine/OpenRefine/pull/7494 ?

> Awesome! Did you draw all those SVGs yourself? That's a lot of work!

Fortunately most of them are boxes or geometric shapes, so it wasn't too complicated. 
The flag from the star-flag-map is not from me though, it is adapted from the existing file `main/webapp/modules/core/images/operations/row-flag.svg`

tfmorris:
> Should I re-add the PNGs I previously removed in https://github.com/OpenRefine/OpenRefine/pull/7226 and https://github.com/OpenRefine/OpenRefine/pull/7494 ?

If it's not too much trouble, that would be great. Then we can make an announcement that they're all going away together and anyone who's using them will have fair warning.

PR #7554, opened 2025-12-01
fix typo in path to operation icon

smierz:
minimal fix, no associated issue

### Status quo
- when opening a project, one gets the following failing http request (404)
typo-in-operations-path

In the failing GET request notice the typo in URL path "images/operation/recon-match-specific-topic-to-cells.svg" 
 -> missing "s" at the end of  "operations".

### Reproduce 
The error is thrown when pre-loading the images (happens in same file)
https://github.com/OpenRefine/OpenRefine/blob/873965f16f23ac45b36c7acc556527602b03948d/main/webapp/modules/core/scripts/project/operation-icons.js#L9-L11

Test/reproduce on master by modifying the code like this and opening a project
```js
  // this will cause the image to load so that once it is added to the DOM it will already be in the browser cache
  var image = new Image();
  image.onerror = function () {
    console.error("Failed to load image:", image.src);
  };
  image.src = iconPath;
```

- fix typo in path to operation icon


smierz:
Same error appears when using "_Match all filtered cells to…_" in column menu
broken-img-in-history

because then a history entry in "Undo/Redo" tab is created with a broken image:
broken-image-in-history

SoryRawyer (approved):

PR #7572, opened 2025-12-10
Fix: enable middle click to open project in new tab

### LINKED ISSUE #6697: Recon suggest widget: enable middle-click to open URL in new tab without selecting

### Proposed solution

### Alternatives considered

Not sure?

### Additional context

> 
> image

riyaseema80:
This PR adds a helpful improvement to the reconciliation UI.
Users can now open a reconciliation suggestion in a new browser tab without actually selecting it.

What’s new

Middle-click (Windows/Linux) opens the entity page in a new tab

Command + click (macOS) does the same

A normal click still selects the suggestion as before

The browser’s default behavior and context menu continue to work normally

This makes it easier for users to quickly preview Wikidata (or other service) pages before deciding which suggestion to pick.

Closes #6697

### DIFF
--- a/main/webapp/modules/core/scripts/index/open-project-ui.js
+++ b/main/webapp/modules/core/scripts/index/open-project-ui.js
@@ -331,7 +331,9 @@ Refine.OpenProjectUI.prototype._renderProjects = function(data) {
       .addClass("searchable")
       .text(project.name)
       .attr("href", "project?project=" + project.id)
+      .attr("target", "_blank")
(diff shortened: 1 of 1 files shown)


SoryRawyer:
Thanks for working on this! I was looking at the original issue, and I don't think this PR satisfies the acceptance criteria. My understanding is that this change would probably be made in the `ReconCellRenderer` class. Are you able to take a screenshot and add that to this pull request?

SoryRawyer:
I'm going to close this due to inactivity, but we can always reopen if needed.

PR #7578, opened 2025-12-11
Fix #6697: Enable middle-click and Cmd+Click on reconciliation candidates

### LINKED ISSUE #6697: Recon suggest widget: enable middle-click to open URL in new tab without selecting

riyaseema80:
Implements the requested behavior from issue #6697.

Users can now open the Wikibase entity page in a new browser tab without selecting it:
- Cmd + Click (macOS)
- Ctrl + Click (Windows/Linux)
- Middle click

Normal click continues to select the candidate as expected.
Screenshot 2025-12-12 at 12 08 19 AM

### DIFF
--- a/main/webapp/modules/core/scripts/index/open-project-ui.js
+++ b/main/webapp/modules/core/scripts/index/open-project-ui.js
@@ -331,8 +331,18 @@ Refine.OpenProjectUI.prototype._renderProjects = function(data) {
       .addClass("searchable")
       .text(project.name)
       .attr("href", "project?project=" + project.id)
-      .attr("target", "_blank")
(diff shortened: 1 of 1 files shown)


SoryRawyer:
Hi @riyaseema80, these changes seem very similar to those in #7572. Can you share the process you followed for checking that this fulfills the requirements in the associated issue?

SoryRawyer:
I'm going to close this and #7572 since they've both been inactive for a little bit. We can always reopen this if and when you're ready to pick this up!

Issue #7521, opened 2025-10-30
Column mapping modal can exceed window height


SoryRawyer:
When applying a history JSON file to a project with a lot of columns, it's possible for the modal window to exceed beyond the bottom of the window, making it impossible to continue.

### To Reproduce
1. First, download the example data and JSON history file from this forum post: https://forum.openrefine.org/t/apply-operation-history-from-json-file-not-working/1582/11 
2. Then, create a project using the CSV included in the zip archive
3. Finally, try to apply the history in the JSON file in the zip archive

### Current Results
The column mapping screen extends beyond the bottom of the window, cutting off the rest of the columns and the buttons to proceed or go back.

### Expected Behavior
Any overflow in the column mapping dialog should scroll such that users can review all columns and proceed with applying the history.

### Versions
 - Operating System: Ubuntu 24.04
 - Browser Version: Firefox 144.0.2
 - JRE or JDK Version: openjdk 21.0.7
 - OpenRefine: 3.10-SNAPSHOT (master branch as of commit `b15ff8985930323f3b615472ba368b4ad4336538`)

### Datasets
The zip archive in the aforementioned forum post (direct link [here](https://forum.openrefine.org/uploads/short-url/tEP06GuAGot2W8QAIaeSHQp8K9t.zip)).

PR #7579, opened 2025-12-15
Allow partially-applying project histories

### LINKED ISSUE #7500: Allow users to partially apply histories
https://github.com/OpenRefine/OpenRefine/pull/6907 changed the behavior of applying JSON edit histories (recipes) by requiring an "all of nothing" approach. Prior to OpenRefine 3.9.0, any failures from applying histories were ~silently ignored~ inconsistently reported due to implementation bugs. The comments on [this forum thread](https://forum.openrefine.org/t/apply-operation-history-from-json-file-not-working/1582) indicate that this change, while well-intentioned, is making existing workflows more difficult. This feature request is to allow users to apply JSON histories that may contain failing (or irrelevant) operations.

### Proposed solution
As a first step, I think presenting users with the errors and allowing them to continue anyway would be a reasonable approach. In this scenario, if there are any errors in the operations to be applied, the user will be notified of each error and given the option to continue and apply the recipe or cancel and return to the previous dialog.

SoryRawyer:
Fixes #7500

- Changes the column mapping dialog to allow an incomplete mapping of recipe columns to project columns. This allows users to apply recipes where optional columns are ignored.
- `ApplyOperationsCommand` removes any operations that rely on a missing column
- Currently includes the change from #7521 since I was testing with a recipe that triggered that bug

An example project/history can be found in the initial description of issue #7521

The basic approach is to drop the requirement that all of a recipe's columns are present in the project. Since we can determine which columns are used in each operation, any operations featuring a column that is not included in the project will be excluded from the final process.

### DIFF
--- a/modules/core/src/main/java/com/google/refine/model/AbstractOperation.java
+++ b/modules/core/src/main/java/com/google/refine/model/AbstractOperation.java
@@ -139,7 +139,8 @@ abstract public class AbstractOperation {
     /**
      * Determine whether this operation relies on any of the input column names
      *
-     * @param columnNames a list of column names
+     * @param columnNames
+     *            a list of column names
      * @return boolean
      */
     public boolean dependsOnAny(Set<String> columnNames) {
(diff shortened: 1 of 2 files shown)


SoryRawyer:
Since there hasn't been much activity on this for a while, I'm going to close this. We can revisit if and when this conversation restarts.

PR #7553, opened 2025-11-30
Improve safety of getIntElement() by adding null and bounds checks

Dhyan011:
Fixes #7556

Summary

This pull request improves the safety and robustness of the getIntElement() method in JSONUtilities.

What Changed
	•	Added a null check for the input ArrayNode.
	•	Added index bounds validation (i < 0 and i >= a.size()).
	•	Prevents potential IndexOutOfBoundsException.
	•	Ensures the method always returns the default value on invalid input instead of throwing runtime errors.

Why This Change Is Important

Previously, the method accessed a.get(i) without checking whether i was within valid bounds. This could lead to runtime crashes during array processing.
The updated logic makes the method safer and more predictable, especially when dealing with user-provided or external JSON arrays.

Testing
	•	Manually verified with valid index
	•	Verified behavior with invalid indices
	•	Verified behavior with null input
All cases now return the default value safely.

PR #7586, opened 2025-12-17
Move new test to correct directory. Fixes #7553

tfmorris:
Fixes #7553

- Move the new class to the correct directory to match project conventions


smierz:
Afaict the class that is covered by tests (JSONUtilities) is placed in
`modules/core/src/main/java/com/google/refine/util/JSONUtilities.java` 

-->  I would expect to find the corresponding test class under 
`modules/core/src/test/java/com/google/refine/util/JsonUtilitiesTest.java`
not in module `main`

tfmorris:
Good catch. I just moved it to the nearest place that my IDE didn't complain about without investigating more. My bad! It is hopefully correct now.

SoryRawyer:
Thanks for catching this! I checked that the package name aligned with the other tests but must have missed that this was in the wrong module.

PR #7588, opened 2025-12-18
get jquery-ui from npm

### LINKED ISSUE #6909: jquery-ui: switch to fetching it via NPM instead of using a vendored version?

smierz:
Fixes #6909 

- get jquery-ui incl themes from npm
- rm copy of jquery-ui from this repo
- also update the included license file (copy-pasted the content from [jquery-ui branch for 1.13.x](https://github.com/jquery/jquery-ui/blob/1-13-stable/LICENSE.txt))

I also refactored the file `copy-dependencies.js` slightly, these changes are _optional_:
- fixed typo, formatting, added comments
- used node module `path` for path construction

### Notes regarding version
The jquery-ui version currently included on master is 1.13.1. 
My plan was to start from that version and when everything is working bump it up to the current version 1.14.1.

Ultimately I started with version **1.13.2** because only from this version on built files + themes are included in the package, see [release notes](https://blog.jqueryui.com/2022/07/jquery-ui-1-13-2-released/) or https://github.com/jquery/jquery-ui/issues/2011.
--> In version 1.13.1 one would have to build the files +themes currently included in this repo themselves. From 1.13.2 on the [package includes a folder "dist"](https://www.npmjs.com/package/jquery-ui/v/1.13.2?activeTab=code) where the files we currently include are located. I copied the files exactly in the structure as they are currently included.

Next I tried to bump up to current version _**1.14.1**_ and though the build still works, the [release notes of 1.14.x](https://blog.jqueryui.com/2024/08/jquery-ui-1-14-0-released/) state that there are _**breaking changes**_. For example I noticed that the window for the scatterplot does not open, an error is printed in console of web developer tools, so the upgrade will need some additional work and for now I only bumped it up to the latest version of 1.13.x (**1.13.3**).

### DIFF
--- a/main/webapp/copy-dependencies.js
+++ b/main/webapp/copy-dependencies.js
@@ -1,7 +1,8 @@
 import fs from 'fs';
+import path from 'path';
 
-const FROM_DIR = 'node_modules';
-const TO_DIR = 'modules/core/3rdparty';
+const FROM_DIR = path.join('node_modules');
+const TO_DIR = path.join('modules', 'core', '3rdparty');
 
 if (!fs.existsSync(TO_DIR)) {
     fs.mkdirSync(TO_DIR);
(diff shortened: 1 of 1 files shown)


SoryRawyer (approved):
I think this looks good! Thank you very much for taking this on.

SoryRawyer:
The reason I took so long to review this was because I wasn't entirely sure how to actually test it. I looked at a few parts of the UI that rely on jquery-ui (like allowing the facet drag and drop) and made sure they still worked as expected. I also went over the changelogs for 1.13.2 and 1.13.3 and I don't think those versions include anything that would affect us. I also confirmed that all the images we rely on are included in the copy dependencies script.

Thank you again @smierz !

smierz:
@SoryRawyer no worries, thanks for the thorough review!

PR #7590, opened 2025-12-20
Add Automatic & manual refresh to Wikibase Preview and Issues tabs

### LINKED ISSUE #6303: Refresh button in Wikibase preview tab
While working in my schema for a wikidata upload, I was working between my rows, Schema, Issues, and Preview tabs. I noticed that as I carried out operations on the grid, my Issues and Preview were not being refreshed/updated as I intuitively expected, so that everything would be up to date as I worked.

Preview

hariprabhu571:
Fixes #6303

### What this does
- Automatically refreshes Wikibase **Preview** and **Issues** tabs when schema or data changes
- Adds a **manual refresh control** as a fallback for users
- Ensures Preview and Issues stay in sync during editing

### Why
Previously, Preview and Issues tabs did not update as expected while working on the schema or grid.
This change improves usability by:
- keeping results updated automatically
- still allowing users to manually trigger a refresh when needed

### Testing
- OpenRefine 3.10-SNAPSHOT
- Windows
- Verified automatic refresh and manual refresh behavior

### DIFF
--- a/main/webapp/package-lock.json
+++ b/main/webapp/package-lock.json
@@ -30,7 +30,8 @@
     "node_modules/jquery": {
       "version": "3.7.1",
       "resolved": "https://registry.npmjs.org/jquery/-/jquery-3.7.1.tgz",
-      "integrity": "sha512-m4avr8yL8kmFN8psrbFFFmB/If14iN5o9nw/NgnnM+kybDJpRsAynV2BsfpTYrTRysYUdADVD7CkUUizgkpLfg=="
+      "integrity": "sha512-m4avr8yL8kmFN8psrbFFFmB/If14iN5o9nw/NgnnM+kybDJpRsAynV2BsfpTYrTRysYUdADVD7CkUUizgkpLfg==",
+      "peer": true
     },
     "node_modules/jquery-migrate": {
       "version": "3.6.0",
@@ -77,7 +78,8 @@
     "jquery": {
       "version": "3.7.1",
       "resolved": "https://registry.npmjs.org/jquery/-/jquery-3.7.1.tgz",
-      "integrity": "sha512-m4avr8yL8kmFN8psrbFFFmB/If14iN5o9nw/NgnnM+kybDJpRsAynV2BsfpTYrTRysYUdADVD7CkUUizgkpLfg=="
+      "integrity": "sha512-m4avr8yL8kmFN8psrbFFFmB/If14iN5o9nw/NgnnM+kybDJpRsAynV2BsfpTYrTRysYUdADVD7CkUUizgkpLfg==",
+      "peer": true
     },
     "jquery-migrate": {
       "version": "3.6.0",


SoryRawyer:
@hariprabhu571 thanks for opening a PR! Could you please include a screenshot showing the new UI details?

hariprabhu571:
Hi @SoryRawyer,

Hereby i have added the video demo and the Screenshot of the solution. Kindly Let me know if there any details Required further.


https://github.com/user-attachments/assets/58dcf26a-897e-4b32-90f8-9236eeb8c84d

Screenshot 2025-12-23 015007

hariprabhu571:
Hi @SoryRawyer,
I have added the Screentshots and a Demo video, hope that works, could you check for the PR, and let me know if there any additional details Required

extensions/wikibase/module/scripts/schema-alignment.js
```diff
@@ -101,6 +101,17 @@ SchemaAlignment.setUpTabs = function() {
         .attr('src', 'images/large-spinner.gif')
         .attr('width', '16px')
         .appendTo(issuesButton);
+  // Add refresh icon to Issues tab
+  this.issuesRefreshIcon = $('<span></span>')
```
SoryRawyer: Could you change this and the preview refresh icon below to be buttons instead of spans?

extensions/wikibase/module/scripts/schema-alignment.js
```diff
@@ -101,6 +101,17 @@ SchemaAlignment.setUpTabs = function() {
...
+        .addClass('wbs-refresh-icon')
+        .addClass('wbs-icon')
+        .attr('title', $.i18n('wikibase-schema/refresh-issues'))
+        .appendTo(issuesButton)
+        .on('click', function(e) {
+          e.preventDefault();
+          e.stopPropagation();
```
SoryRawyer: Ideally we should avoid disabling standard browser features. What happens when these two lines are removed?

extensions/wikibase/module/styles/schema-alignment.css
```diff
@@ -171,6 +171,23 @@
...
+  margin-left: 5px;
+  cursor: pointer;
+  opacity: 0.6;
+}
+
+.wbs-refresh-icon:hover {
+  opacity: 1;
+}
+
+.wbs-refresh-icon.wbs-icon {
```
SoryRawyer, line 187: I know we have SVGs defined inline elsewhere, but I think it'd be best to stick with the pattern of putting all images under the `images` directory (in this case `extensions/wikibase/module/images`) and referencing that file when needed. You can see a similar pattern here: https://github.com/OpenRefine/OpenRefine/pull/7530. There are also examples of Wikibase-specific images in this edited file.

hariprabhu571:
Hi @SoryRawyer,
Thanks for the review and the detailed feedback!

I’ll update the refresh controls to use `<button>` elements instead of `<span>`s and adjust the CSS accordingly.

I’ll also test removing `preventDefault()` and `stopPropagation()` to see if any behavior breaks, and will either remove them or document why they’re needed.

I’ve tested this locally and the automatic + manual refresh behavior works as intended on my setup. Happy to add more details if helpful.

extensions/wikibase/module/scripts/schema-alignment.js
hariprabhu571: before changes:
Screenshot 2025-12-29 232639

After changes: (span to button)
Screenshot 2025-12-29 235437

After changes: (span to button & With change in CSS)
image


@SoryRawyer, is that okay to add the CSS to align with the Required UI

extensions/wikibase/module/scripts/schema-alignment.js
hariprabhu571: What happens if removed
e.stopPropagation() — critical: Without it, clicking the refresh button bubbles to the parent div, triggering the tab switch handler. Result: clicking refresh also switches tabs.
e.preventDefault() — less critical here: For type="button" there’s usually no default action, but it’s a safeguard if the button is inside a form.

wetneb:
@SoryRawyer I could try this out locally to check that it works as expected, or I could get on a call with you to help you get more familiar with this area of OpenRefine, so that you are able to evaluate this sort of pull request on your own. What do you prefer?

Looking at the screenshots alone, I'm grateful for @hariprabhu571 for following the mock-up given in the issue by @cooperzoe , but I must say I am not really convinced by that design in the first place: those refresh buttons are a bit odd to fit inside the tab headers (the user needs to be quite careful in clicking the right part of the tab header depending on what they want to do). Those buttons should only needed in rare cases, since the contents of those tabs normally refresh automatically when the schema is changed, so it feels weird to me to keep those buttons visible at all times instead of embedding them in the tab body instead. That's just my opinion, since it was asked for, I leave it up to you both to decide what you want to do with it :)

hariprabhu571:
@wetneb , Thanks for mentioned that.

@SoryRawyer , Please let me know if Manual refresh is not required, I will fix that with only the automatic refresh.

SoryRawyer:
> @SoryRawyer I could try this out locally to check that it works as expected, or I could get on a call with you to help you get more familiar with this area of OpenRefine, so that you are able to evaluate this sort of pull request on your own. What do you prefer?

If you're up for it, a call would be great!

@hariprabhu571 I like @wetneb's suggestion (and trust their understanding of the requirements better than mine :) ) so moving the manual refresh buttons to the tab body sounds good to me.

hariprabhu571:
Hi @SoryRawyer, Any updates

SoryRawyer:
@hariprabhu571 I'm still working on things on my end, but in the meantime do you think you could try moving the refresh icons into the tab bodies rather than on the tabs themselves? I imagine this might resolve the need for calling `preventDefault` too!

SoryRawyer:
Hey @hariprabhu571 , I just wanted to check in and see if you have any questions about the feedback so far.

As for functionality, I think `SchemaAlignment.preview` is the right approach.

hariprabhu571:
Hi @SoryRawyer ,

Just to confirm my understanding:
you’d prefer removing the manual refresh control entirely and keeping only automatic refresh, implemented via SchemaAlignment.preview, correct?

If that’s the case, I’ll update the PR accordingly and push the changes.

Please let me know if I’ve interpreted this correctly. Thanks!

SoryRawyer:
Ah sorry for the misunderstanding, @hariprabhu571. I think the manual refresh arrow is a good idea, but I think keeping them within the tab body would be a better fit for the UI. The issues and preview tab could both put the arrow in the top-right corner (just under wikibase extension dropdown menu). What do you think?

hariprabhu571:
Hi @SoryRawyer 



So near to the extension drowdown, need a manual refresh svg icon right?, one single common icon for to refresh both issue tab and preview tab

SoryRawyer:
That's what I had in mind, yes. Thanks!

hariprabhu571:
Hi @SoryRawyer, as discussed, the refresh icons from the individual tabs have been moved to a common refresh button near the extension dropdown.

For any further queries or changes, please let me know.

SoryRawyer:
Thanks for the update @hariprabhu571. I'm still reviewing this, but in the meantime if you're able to upload a screenshot of the new functionality that would help move the review forward. Thanks!

hariprabhu571:
Hi @SoryRawyer,
Screenshot 2026-02-06 195842

Keep the same functionality, but instead of having separate refresh icons in the Issue and Preview tabs, create a single common refresh icon and place it near the extension dropdown, as shown in the screenshot. Let me know if there any further details needed.

hariprabhu571:
Hi @SoryRawyer , any updates?

extensions/wikibase/module/langs/translation-en.json
```diff
@@ -41,6 +41,8 @@
     "wikibase-schema/schema-tab-header": "Schema",
     "wikibase-schema/warnings-tab-header": "Issues",
     "wikibase-schema/edits-preview-tab-header": "Preview",
+    "wikibase-schema/refresh-issues": "Refresh Issues",
+    "wikibase-schema/refresh-preview": "Refresh Preview",
```
SoryRawyer: Since we just have one icon now, could you replace these with a single "Refresh Schema" item?

SoryRawyer (changes requested):
I just have this one final request, but otherwise I think this looks good! I really appreciate you sticking with this one despite the review lag on my end.

extensions/wikibase/module/langs/translation-en.json
hariprabhu571: > Since we just have one icon now, could you replace these with a single "Refresh Schema" item?

Sure, I will update that

hariprabhu571:
Hi @SoryRawyer, the changes which you mentioned is resolved. Kindly check and let me know the status.

main/webapp/package-lock.json
SoryRawyer, line 1: Would you mind removing this file from the PR? I don't think it's necessary for the change and we can update the lockfile in a separate PR if needed.

hariprabhu571: Removed the package-lock.json from the PR. Thanks for pointing it out!

hariprabhu571:
Hi @SoryRawyer, Removed the package-lock.json from the PR. Thanks for pointing it out!

hariprabhu571:
Hi @SoryRawyer ,
All checks are passing and review is approved.
While waiting for the merge, I’d like to continue contributing.
Could you suggest another issue I can work on?

SoryRawyer:
Hi @hariprabhu571, I saw yesterday that you already have a new issue assigned to you. Feel free to reach out on that issue if you have any questions about getting started!

PR #7596, opened 2025-12-26
UI stays stable as the user types.

### LINKED ISSUE #7571: Transform dialog resizes continually as typing
This is a relatively recent regression, but I don't know, off the top of my head, what the likely cause is. 

### To Reproduce
1. Use Edit Cells... -> Transform to bring up the dialog
2. Start typing a GREL expression

### Current Results

The Transform popup dialog continually resizes as you're typing GREL expressions, perhaps driven by the appearance and disappearance of the syntax error messages as one types partially formed expressions. This is SUPER annoying and distracting.

### Expected Behavior

UI stays stable as the user types.

### Versions
 - Operating System: 
 - Browser Version: 
 - JRE or JDK Version: 
 - OpenRefine:

meskatiyar:
UI stays stable as the user types.

Fixes #7571

- 
- 
-

### DIFF
--- a/main/tests/cypress/cypress/e2e/project/grid/misc/expressions.cy.js
+++ b/main/tests/cypress/cypress/e2e/project/grid/misc/expressions.cy.js
@@ -120,6 +120,26 @@ describe(__filename, function () {
     ).should('to.contain', 'Error: No matching method');
   });
 
+  it('Dialog height should remain stable while typing', function () {
+    cy.loadAndVisitProject('food.mini');
+    loadExpressionPanel();
+    cy.get('.dialog-frame').then(($frame) => {
+      const initialHeight = $frame[0].getBoundingClientRect().height;
+      cy.get('textarea.expression-preview-code').clear().type('(');
+      cy.wait(500); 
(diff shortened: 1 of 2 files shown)


SoryRawyer:
@meskatiyar thanks for your contribution! Could you please fill out the template with more details on your change? How did you decide on the final CSS changes to resolve this issue?

tfmorris:
I've added the original issue number so that it's linked, but as Rory said, please use the PR template.

Thank you for including an E2E test. The test failure is, I believe, unrelated to this PR, so that can be ignored. I've resubmitted the CI job and it will hopefully pass.

I've verified that this fixes the immediate problem that I reported, but I think we may want to pursue a more holistic fix to the dialog box layout (although we could also do that separately). I've updated the original issue with links to the changes that I think got us to this point. I think undoing or doing differently some of those changes might be the path forward.

@meskatiyar If you are willing to help with that, please have a look.

tfmorris:
Closing as stale. Unfortunately, in the current state, it doesn't solve the problem.

PR #7597, opened 2025-12-26
Defensively reject null or blank Host headers

sanket-darunkar:
While reviewing host validation logic, I noticed that the Host header
can be null or blank in malformed or non-browser HTTP requests, which
could lead to unexpected behavior.

This change adds an early defensive check to reject such requests,
ensuring host validation remains robust without altering existing
behavior for valid requests.

### DIFF
--- a/server/src/com/google/refine/ValidateHostHandler.java
+++ b/server/src/com/google/refine/ValidateHostHandler.java
@@ -64,6 +64,13 @@ class ValidateHostHandler extends HandlerWrapper {
 
     public boolean isValidHost(String host) {
 
+        // Defensive check: the Host header can be null or blank in malformed
+        // or non-browser HTTP requests. Such requests must be rejected to
+        // avoid unexpected behavior or bypassing host validation.
+        if (host == null || host.isBlank()) {
+            return false;
+        }
+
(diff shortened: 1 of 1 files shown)


sanket-darunkar:
Thanks for the review!

This change is limited to defensive validation of the Host header on the server side.
It should not impact UI behavior.

Happy to adjust or investigate further if the E2E failure is related.

SoryRawyer:
Hi @sanket-darunkar, thank you for your interest in contributing to OpenRefine and making our code base more resilient to null values! However, could you please ask to be assigned an existing issue? This doesn't seem like a particularly urgent fix as OpenRefine is intended to be used from within browsers and supporting non-browser HTTP traffic will likely be a much bigger initiative that will cover attack surface areas like the one you mentioned.

sanket-darunkar:
Thanks for the clarification, that makes sense.

I understand that OpenRefine is primarily browser-oriented and that broader handling of non-browser HTTP traffic would require a more coordinated initiative.

I’ll avoid proposing such changes without prior discussion and will look for existing issues to get assigned to before working on them.

Thanks again for the guidance, and I appreciate the review.

PR #7598, opened 2025-12-28
Validate project name before renaming

sanket-darunkar:
This change adds defensive validation to prevent renaming a project
with a null or blank name.

The command now rejects invalid input early and trims the provided
project name before persisting it, without affecting existing
behavior for valid requests.

### DIFF
--- a/server/src/com/google/refine/ValidateHostHandler.java
+++ b/server/src/com/google/refine/ValidateHostHandler.java
@@ -64,6 +64,13 @@ class ValidateHostHandler extends HandlerWrapper {
 
     public boolean isValidHost(String host) {
 
+        // Defensive check: the Host header can be null or blank in malformed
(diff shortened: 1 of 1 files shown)


SoryRawyer:
Hi @sanket-darunkar, I don't think the pull request description matches the code change included. Also, as I mentioned in my other comment, please comment on an existing issue and ask to be assigned rather than opening a PR with no associated issue.

sanket-darunkar:
Hi @SoryRawyer,

Thank you for the clarification and for pointing this out. You’re right — the PR description does not accurately reflect the changes included, and I also understand now that I should have first commented on an existing issue and asked to be assigned before opening a PR.

I apologize for not following the correct process. I’ll update this PR accordingly and will comment on an appropriate issue to request assignment before submitting a revised PR.

Thanks again for the guidance — I appreciate it.

PR #7608, opened 2026-01-13
Add `reduce` GREL control method

### LINKED ISSUE #7565: Add a `reduce` GREL function
When working with arrays, it is useful to be able to be able to iterate over all elements and return a single value. However, I haven't been able to find anything in GREL what would let users do this generically. There are specific functions like `sum` and `join`, but nothing that would allow a user-specified expression to build the result.

### Proposed solution
GREL should provide a `reduce` (sometimes referred to as `fold`) function that operates over arrays.

This is the usage that I have in mind, which would return 6:
`reduce([1, 2, 3], v, acc, 0, v + acc)`

Arguments:
- `[1, 2, 3]` is the input array
- `v` declares the variable name of the current element being processed
- `acc` declares the variable name of the accumulated value so far
- `0` is the initial value for `acc`
- `v + acc` is the expression to be evaluated for each item in the array.

### Alternatives considered
One alternative would be to create more special-purpose functions similar to `sum` and `join`. Providing a more generic function would be more complex but can be applied in many different ways.

SoryRawyer:
Fixes #7565

- Adds a `reduce` GREL control flow operation to operate on lists of values

Here's a screenshot showing the function in the help tab:
Screenshot from 2026-01-13 15-55-55


modules/grel/src/main/resources/com/google/refine/grel/ControlDescription.properties
```diff
@@ -13,3 +13,4 @@ is_null_desc=Returns whether o is null
 is_numeric_desc=Returns whether o can represent a number
 is_test_desc=Returns whether o can represent a number
 with_desc=Evaluates expression o and binds its value to variable name v. Then evaluates expression e and returns that result
+reduce_desc=Processes a list by evaluating expression func over each element of expression e, using initial as the initial value to the input function, v as the current element, and acc as the accumulated value
```
tfmorris:
```suggestion
reduce_desc=Processes a list by evaluating expression func over each element of list generated by expression e, using initial as the initial value to the input function, v as the current element, and acc as the accumulated value
```

modules/grel/src/test/java/com/google/refine/grel/controls/ReduceTests.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Test
+    public void testReduce() throws ParsingException {
+        bindings = new Properties();
+        bindings.put("v", "");
+        Object result = getResult("reduce([1, 2, 3], v, acc, 0, v + acc)");
+        Assert.assertEquals(result, 6L);
+    }
+}
```
tfmorris, line 101: I think the tests could be a little more comprehensive, particularly for error conditions like wrong number or type of arguments, as well as boundary conditions like empty lists and the various branches for different data types (at least the JSON Array should be easy to add, not sure about the other).

modules/grel/src/main/java/com/google/refine/grel/controls/Reduce.java
```diff
@@ -0,0 +1,159 @@
...
+            return new EvalError(ControlEvalError.expects_first_arg_array(ControlFunctionRegistry.getControlName(this)));
+        }
+
+        String valueVariable = ((VariableExpr) args[1]).getName();
+        String accumulatorVariable = ((VariableExpr) args[2]).getName();
+        Object initialValue = args[3].evaluate(bindings);
+        if (ExpressionUtils.isError(initialValue)) {
+            return initialValue;
+        }
+        bindings.put(accumulatorVariable, initialValue);
```
tfmorris, line 73: This will fail for an initial value of `null`.

modules/grel/src/main/java/com/google/refine/grel/controls/Reduce.java
```diff
@@ -0,0 +1,159 @@
...
+                    bindings.put(valueVariable, v);
+                } else {
+                    bindings.remove(valueVariable);
+                }
+
+                Object result = expr.evaluate(bindings);
+                if (ExpressionUtils.isError(result)) {
+                    return result;
+                }
+                bindings.put(accumulatorVariable, result);
```
tfmorris, line 90: This will fail if any expression result is `null`

modules/grel/src/main/java/com/google/refine/grel/controls/Reduce.java
```diff
@@ -0,0 +1,159 @@
...
```
tfmorris, line 109: Simplifying the three copies of this code would be a useful improvement.

modules/grel/src/test/java/com/google/refine/grel/controls/ReduceTests.java
```diff
@@ -0,0 +1,101 @@
...
+    // an example of evaluating grel code within a test
+    private Object getResult(String expression) throws ParsingException {
+        Evaluable eval = MetaParser.parse("grel:" + expression);
+        return eval.evaluate(bindings);
+    }
+
+    @Test
+    public void testBaseReduce() throws ParsingException {
+        Object result = getResult("reduce([1, 2, 3], v, acc, 0, v + acc)");
+        Assert.assertEquals(result, 6L);
```
tfmorris, line 58: Testing additional data types would be a good idea.

```suggestion
        Assert.assertEquals(getResult("reduce([1, 2, 3], v, acc, 0, v + acc)"), 6L);

        Assert.assertEquals(getResult("reduce(parseJson('[1, 2, 3]'), v, acc, 0, v + acc)"), 6L);

        Assert.assertEquals(getResult("reduce(['a', 'b', 'c'], v, acc, 'z', acc + v)"), "zabc");
```

tfmorris (changes requested):
That's an improvement. Thanks for the updates! There are still a couple of things which need cleanup. I think handling null Properties values can be handled with a little helper like this:

```java
    private static void safePut(Properties bindings, String name, Object value) {
        if (value != null) {
            bindings.put(name, value);
        } else {
            bindings.remove(name);
        }
    }
```

and the branch logic could be simplified along the lines of 

```
        Iterable<Object> values;
        if (inputList.getClass().isArray()) {
            values = Arrays.asList((Object[]) inputList);
        } else if (inputList instanceof ArrayNode) {
            values = ((ArrayNode) inputList).valueStream().map(JsonValueConverter::convert).collect(Collectors.toList());
        } else {
            values = ExpressionUtils.toObjectCollection(inputList);
        }

        for (Object v : values) {
            safePut(bindings, valueVariable, v);

            Object result = expr.evaluate(bindings);
            if (ExpressionUtils.isError(result)) {
                return result;
            }
            safePut(bindings, accumulatorVariable, result);
        }
        return bindings.get(accumulatorVariable);
```

I think the above might copy the entire ArrayNode, rather than iterating through it, so if there's a simple solution to that, it would be more efficient, but I don't think it's a huge deal given the size of arrays we're typically working with.

Also, Filter seems to save and restore the value of the variable used. I don't know off the top of my head if this is part of the contract, but it seems best to be consistent and follow the same pattern, wrapping it in a try/finally block.

modules/grel/src/test/java/com/google/refine/grel/controls/ReduceTests.java
```diff
@@ -0,0 +1,101 @@
...
+            MetaParser.parse("grel:reduce([1, 2, 3], v, acc, 0, v - acc, 0)");
+        } catch (ParsingException p) {
+            Assert.assertTrue(p.getMessage().contains(ControlEvalError.expects_five_args("reduce")));
+        }
+    }
+
+    @Test
+    public void testArgumentTypes() throws ParsingException {
+        try {
+            MetaParser.parse("grel:reduce([1, 2, 3], 0, acc, 0, v - acc)");
```
tfmorris, line 90:
```suggestion
            MetaParser.parse("grel:reduce([1, 2, 3], 0, acc, 0, v - acc)");
            Assert.fail("Should have thrown an exception");
```

Tests which expect an exception to be thrown should fail if it isn't. If you don't need to check the text of the message, you could also use `assertThrows()`

PR #7641, opened 2026-02-06
Fix invalid wbformatvalue requests in Wikibase extension (#5658)

### LINKED ISSUE #5658: Invalid wbformatvalue requests to preview values in the Wikibase extension
Ticket [T330859](https://phabricator.wikimedia.org/T330859) on Wikimedia's side reports that OpenRefine is making invalid requests to the MediaWiki API responsible for formatting Wikibase statement values. This API endpoint is used in the Preview tab of the Wikibase extension.

To reproduce the bug, one should likely run OpenRefine in a language where the `core-recon/wd-recon-lang` string is not translated, open the network inspector of your browser, create a Wikibase schema which adds a statement to some item, and go to the preview tab. There should be some failing requests to the Wikibase API as described in the ticket.

The ticket guesses that this might be the offending code: https://github.com/OpenRefine/OpenRefine/blob/5b4436272f8e45a68c9173dac67f6b6fde09dd5a/extensions/wikibase/module/scripts/preview-renderer.js#L333

There are a total of [5 places](https://github.com/search?type=code&q=repo%3AOpenRefine%2FOpenRefine+wd-recon-lang+language%3AJavaScript&l=JavaScript) in the code that reference this. They should probably all be replaced with a call to a helper function that falls back to "en" if the lookup fails (ie returns a string containing the original key "core-recon/wd-recon-lang").

Akrittt:
## Summary
Fixes #5658 - OpenRefine was making invalid API requests to Wikibase's `wbformatvalue` 
endpoint when the UI language was set to one where `core-recon/wd-recon-lang` is not translated.

## Problem
When `$.i18n('core-recon/wd-recon-lang')` fails to find a translation, it returns the 
literal key string `"core-recon/wd-recon-lang"` instead of a valid language code. This 
caused API requests like `{"lang":"core-recon/wd-recon-lang"}` which the MediaWiki API 
rejected as invalid.

## Solution
- Added `getReconLanguageCode()` helper function that:
  - Attempts to retrieve the language code from i18n
  - Detects if the result is the key itself (failed lookup)
  - Falls back to `"en"` if translation is missing
  - Validates language code format
  
- Replaced all 5 occurrences of `$.i18n('core-recon/wd-recon-lang')` with the helper function

### DIFF
--- a/extensions/wikibase/module/scripts/schema-alignment.js
+++ b/extensions/wikibase/module/scripts/schema-alignment.js
@@ -31,6 +31,23 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
  */
 
+function getReconLanguageCode() {
+    var langCode = $.i18n('core-recon/wd-recon-lang');
+    
+    // If the i18n lookup failed, it returns the key itself
+    // Check if the result looks like a translation key rather than a language code
+    if (!langCode || langCode === 'core-recon/wd-recon-lang' || langCode.indexOf('/') !== -1) {
+        return 'en'; // fallback to English
+    }
+    
+    // Additional validation: language codes are typically 2-3 characters
+    // (e.g., 'en', 'fr', 'zh-CN') and don't contain special characters except hyphens
+    if (langCode.length > 10 || /[^a-zA-Z-]/.test(langCode)) {
+        return 'en';
+    }
+    
+    return langCode;
+}
 function serviceLogoFromReconConfig(reconConfig) {
   var serviceUrl = null;
   var service = null;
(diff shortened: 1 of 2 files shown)


extensions/wikibase/module/scripts/schema-alignment.js
```diff
@@ -31,6 +31,23 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
  */
 
+function getReconLanguageCode() {
```
Abbe98, line 34: This should probably be named something else, given that we are not looking for the Recon-language code anymore. It should also be deduplicated.

extensions/wikibase/module/scripts/schema-alignment.js
```diff
@@ -31,6 +31,23 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    
+    // If the i18n lookup failed, it returns the key itself
+    // Check if the result looks like a translation key rather than a language code
+    if (!langCode || langCode === 'core-recon/wd-recon-lang' || langCode.indexOf('/') !== -1) {
+        return 'en'; // fallback to English
+    }
+    
+    // Additional validation: language codes are typically 2-3 characters
+    // (e.g., 'en', 'fr', 'zh-CN') and don't contain special characters except hyphens
+    if (langCode.length > 10 || /[^a-zA-Z-]/.test(langCode)) {
```
Abbe98, line 45: I think the language codes are not your regular language codes but Wikimedia ones: https://www.wikidata.org/wiki/Help:Wikimedia_language_codes/lists/all

Abbe98 (changes requested):
This is on the right track, that said I don't think we should use `core-recon/wd-recon-lang` for this at all, the translations seems to be all over the place, in some cases even invalid.

tfmorris:
Closing as stale.

PR #7649, opened 2026-02-06
Fix flaky e2e arrow navigation test. Fixes #7648

### LINKED ISSUE #7648: Flaky Cypress e2e test - arrow key navigation of expression history
The test
```
cypress/e2e/project/grid/column/facet/facets.cy.js Test navigating expression history with arrow keys across dialog reopen
```

is failing a high percentage (50%?) of the time and when it succeeds it's with a duration of almost 5 seconds, which makes me suspect that we could be in a boundary region near a timeout.

My IDE warns/complains that "it is not safe to chain further commands after a 'focus' command" which might also be a contributing factor.

https://github.com/OpenRefine/OpenRefine/blob/0b73f953cba119fbd1e70a1863f33ac0400f3550/main/tests/cypress/cypress/e2e/project/grid/column/facet/facets.cy.js#L410-L419

tfmorris:
Fixes #7648

- Don't chain off cy.focus(), clear(), or type() since that's unsupported
- use test helpers to minimize flakiness
- - add comments to document what each key press is doing
  (including why we need 2 up arrows to start)
- clean up by closing dialog at end of test

### DIFF
--- a/main/tests/cypress/cypress/support/commands.js
+++ b/main/tests/cypress/cypress/support/commands.js
@@ -192,8 +192,8 @@ Cypress.Commands.add('navigateTo', (target) => {
 Cypress.Commands.add('typeExpression', (expression, options = {}) => {
   cy.get('textarea.expression-preview-code', options).clear();
   cy.get('textarea.expression-preview-code', options).type(expression);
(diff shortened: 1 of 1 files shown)


tfmorris:
@SoryRawyer can you please review this when you get a chance? It's getting tedious to have to keep resubmitting the CI jobs failing due to flakiness.

Issue #7513, opened 2025-10-27
db-extension: `ExecuteQueryCommand` is never used


smierz:
**Status quo:**
Within the database extension is a command `ExecuteQueryCommand` that gets registered but afaict is never used.

**ToDo**:
If the command is never used or referenced it would make sense to remove it.

tfmorris:
Since this is registered and visible to the front end as the command `command/database/execute-query` we probably should give folks some advance warning that it will be removed.  The method `executeQuery` is also specified as part of the abstract `DatabaseService` class.

The `test-connect` and `connect` commands seem like they could be ripe for deduplication as well.

Having said that, this module hasn't been a high priority historically, so there are probably lots of things which could be improved if someone has the energy/time.

iz-Manik:
Hi, thanks for the context. I’ll start with a small PR to deprecate ExecuteQueryCommand, add documentation/comments warning about future removal, and verify current usage. This should give extension users advance notice before any removal. Please let me know if that sounds reasonable.

SoryRawyer:
Closed as part of #7651

smierz:
please also deprecate the corresponding methods in abstract `DatabaseService` + its implementations  ({SQLite, PgSQL, MySQL, MariaDB}+DatabaseService), because they are only used by `ExecuteQueryCommand` and once the command will be removed they can hopefully be removed as well  (see tfmorris comment above).

iz-Manik:
Hi @smierz,

I checked the `executeQuery(...)` method in `DatabaseService` and found that it is only referenced by the deprecated `ExecuteQueryCommand`.

Based on this, I’m planning to remove `executeQuery(...)` from `DatabaseService` and all its implementations, since it appears to be unused elsewhere.

Please let me know if you’d prefer keeping it deprecated instead of removing it.

Thanks!

tfmorris:
@iz-Manik The next release is a major release with no compatibility guarantees, so this is OK to remove rather than just deprecate.

iz-Manik:
Thanks for the clarification @tfmorris.

I’ll proceed with removing `executeQuery(...)` from `DatabaseService` and its implementations, along with the remaining unused references tied to `ExecuteQueryCommand`.

I’ll open a PR once the changes are ready.

PR #7651, opened 2026-02-10
Deprecate ExecuteQueryCommand and document planned removal

iz-Manik:
This PR deprecates ExecuteQueryCommand, which is currently registered and exposed
to the frontend as command/database/execute-query but appears unused in practice.

Changes:
- Mark ExecuteQueryCommand as @Deprecated with documentation
- Add a frontend-facing comment warning about future removal

No functional behavior has been changed.
The command remains registered for backward compatibility.

Refs #7513

### DIFF
--- a/extensions/database/src/com/google/refine/extension/database/cmd/ExecuteQueryCommand.java
+++ b/extensions/database/src/com/google/refine/extension/database/cmd/ExecuteQueryCommand.java
@@ -52,7 +52,7 @@ import com.google.refine.util.ParsingUtilities;
  * @deprecated This command is currently unused and is planned for removal in a future release. It is still registered
  *             as command/database/execute-query for backward compatibility.
  */
-@Deprecated
+@Deprecated(since = "3.10")
 public class ExecuteQueryCommand extends DatabaseCommand {
 
     private static final Logger logger = LoggerFactory.getLogger("ExecuteQueryCommand");


SoryRawyer:
@iz-Manik would you mind undoing the whitespace changes (or separating them into their own commit)? While I think they make the codebase nicer I find they're making it harder for me to focus on the relevant code changes.

iz-Manik:
Thanks for the suggestion! I’ve undone the accidental whitespace changes and kept the PR focused on the deprecation and documentation updates only.

iz-Manik:
Thanks for your patience! I’ve applied the project formatter to address the CI validation error. The PR should now be ready for re-review.

extensions/database/src/com/google/refine/extension/database/cmd/ExecuteQueryCommand.java
```diff
@@ -48,6 +48,11 @@
 import com.google.refine.extension.database.model.DatabaseInfo;
 import com.google.refine.util.ParsingUtilities;
 
+/**
+ * @deprecated This command is currently unused and is planned for removal in a future release. It is still registered
+ *             as command/database/execute-query for backward compatibility.
+ */
+@Deprecated
```
SoryRawyer: Could you add `(since = "3.10")` to the annotation? We should be able to release this during the 3.10 series and it'd be nice to include this detail if possible.

iz-Manik: Thanks for the suggestion! I’ve updated the annotation to @Deprecated(since = "3.10"). Please let me know if any further changes are needed.

PR #7653, opened 2026-02-13
Fix #6814: Allow reference property suggestions in Wikibase schema al…

### LINKED ISSUE #6814: Automatic suggestion of reference tags
I want the tags in the References to be suggested automatically. Every time, I manually type the tags url, title, retrieved, language of work, archive url, archive date, publisher, author, etc. After I type them, I check if I have written the correct property number. Whereas, for example, OpenRefine could have suggested P854 (Reference URL) to me **before I started typing**. (P854: https://w.wiki/9YPi)

A form that comes with ready-made reference tags would be useful.
Thank you. Best regards.

image

Srishtig0711:
This PR fixes issue #6814 by enabling automatic suggestion of reference properties in the Wikibase schema alignment UI.

Previously, suggestions were not appearing when adding references. This update restores proper suggestion behavior.

@SoryRawyer thank you for this opportunity.

### DIFF
--- a/extensions/wikibase/module/scripts/schema-alignment.js
+++ b/extensions/wikibase/module/scripts/schema-alignment.js
@@ -1206,20 +1206,51 @@ SchemaAlignment._addReference = function(container, json) {
    }
   } else {
 
-   var defaultProperties = ["P854", "P1476", "P813", "P123"];
+    var endpoint = WikibaseManager.getSelectedWikibaseApi();
 
-   for (var j = 0; j < defaultProperties.length; j++) {
-      SchemaAlignment._addQualifier(qualifierContainer, {
-         prop: {
-            type: "wbpropconstant",
-            pid: defaultProperties[j],
-            label: defaultProperties[j],
-            datatype: "string"
-         },
-         value: null
-      });
-   }
-  }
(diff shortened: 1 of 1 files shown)


Abbe98:
Wouldn't this break non Wikidata Wikibases?

Srishtig0711:
> Wouldn't this break non Wikidata Wikibases?

you're right.
The current implementation introduces Wikidata-specific property IDs (P854, P1476, P813, P123), which would not necessarily exist in other Wikibase instances.
I will revise this to avoid hardcoded property IDs and instead rely on the reconciliation service / Wikibase configuration for suggestions.

Abbe98:
Did you use a language model to generate this code?

Srishtig0711:
Thank you for your question. I did not use any AI tools to create the code for this change. I spoke with my university lecturers to understand the code and how to implement and what to change. After I implemented the solution myself, I used a language model to clarify certain concepts and improve my understanding of the code structure. I wrote and tested all changes myself, and I understand the implementation and the reasoning behind it.

Abbe98:
Sorry, but your implementation does not do what's described in the issue, nor is the API call a valid one.

PR #7662, opened 2026-02-17
Fix/7659 separate term monolingual lang codes

### LINKED ISSUE #7659: Not all Wikidata monolingual text language codes are valid for terms
### To Reproduce
1. When in a Schema (on Wikidata or Wikimedia Commons), select a specific lang like `kmr`, `apc` or `bik` in the drop-down menu (and probably other languages too, it's just the one that I spotted).

Image

2. Look at the preview, it says _Invalid language code 'bik'_

### Expected Behavior
Proposed languages should be accepted ;)

### Versions
 - Browser Version: Firefox
 - OpenRefine: e.g. OpenRefine 3.9.5

Godson-Ozioma:
Fixes #7659

Summary
Wikidata allows more language codes for monolingual-text claim values (1406) than for terms — labels, descriptions, aliases (632). The schema UI and validation used a single list, so codes like 'bik', 'apc', and 'kmr' were wrongly rejected in monolingual-text contexts. This PR separates term vs monolingual-text lists everywhere and uses the correct list per context.

Backend - LanguageCodeStore.java: Two contexts (TERM, MONOLINGUALTEXT), cache by (endpoint, context), 'getLanguageCodes(endpoint, context)'. Two static default lists ('DEFAULT_TERM_LANGUAGE_CODES', 'DEFAULT_LANGUAGE_CODES') as fallbacks on API error or null endpoint (Wikidata-only). No hardcoded base URL; fetch uses the given API endpoint.

Backend - ValidationState.java: Optional 'languageContext' (term | monolingualtext, default monolingualtext) and 'mediaWikiApiEndpoint' (get/set) so validation can use the right list per context.

Backend - WbLanguageConstant.java: Validation uses endpoint and context from ValidationState; 'normalizeLanguageCode(lang, endpoint, context)' added; one/two-arg overloads keep defaulting to monolingualtext for backward compatibility.

Backend - WbNameDescExpr.java: Sets 'validation.setLanguageContext(TERM)' before validating the inner monolingual value so name/desc languages are validated against the term list.

Backend - WikibaseSchema.java: Sets 'validationContext.setMediaWikiApiEndpoint(this.getMediaWikiApiEndpoint())' at the start of 'validate(ValidationState)' so all validation callers get the schema endpoint without per-caller changes.

Frontend - lang-suggest.js: Two default arrays ('DEFAULT_LANGUAGES_TERM', 'DEFAULT_LANGUAGES_MONOLINGUALTEXT'), cache by (endpoint, context), 'getLanguageCodes(callback, context)' with 'wbclcontext=' in the API URL, and 'languageContext' option for the suggest (default 'monolingualtext'). Outdated comment removed; short fallback comment added.

Frontend - schema-alignment.js: Optional 5th parameter 'options' on 'initField'; name/desc passes 'languageContext: "term"', monolingual-text language field passes 'languageContext: "monolingualtext"' into 'input.langsuggest(...)'.

Static lists: Default term and monolingual lists refreshed from Wikidata API (term 632, monolingual 1406) in Java and JS.

Tests: LanguageCodeStoreTest (null endpoint, term vs monolingual differ, backward-compatible one-arg); WbLanguageConstantTest (TERM rejects monolingual-only code e.g. 'und', 'normalizeLanguageCode' with context); WbNameDescExprTest (name/desc with invalid-for-term language yields validation error).

What to avoid (followed): No hardcoded Wikidata URL; no single shared list; outdated comment removed; static fallback for terms added.


extensions/wikibase/tests/src/org/openrefine/wikibase/utils/LanguageCodeStoreTest.java
```diff
@@ -0,0 +1,62 @@
+/*******************************************************************************
+ * MIT License
+ *
+ * Copyright (c) 2018 Antonin Delpeuch
```
Abbe98: Is this code copied from somewhere? Or did a LLM pick this up?

Godson-Ozioma: It wasn’t copied from elsewhere or added by an LLM. I used the same MIT License header that’s already used in other test files in this package (e.g. WbLanguageConstantTest.java, SnakUtilsTests.java), I added it to follow that convention.

The rest of the file is new test code for the issue #7659 changes, written to match the plan and existing test style here, it wasn’t copied from elsewhere or generated by an LLM as well.

Godson-Ozioma: Hi @Abbe98 

Is there any part of the code you would like me to throw more light on or explain how I did it? 

Also, is there anything you would like me to change, add or delete?

magdmartin: The current standard is "Copyright (c) 2026, OpenRefine contributors" followed by the text of the license as the module header. 
See https://forum.openrefine.org/t/format-of-copyright-notice-at-beginning-of-source-files/2702

Godson-Ozioma: Hi @magdmartin Thanks for the pointer. I’ll update the header in LanguageCodeStoreTest.java to use the current standard: 'Copyright (c) 2026, OpenRefine contributors' followed by the license text, as in the forum post.

Is there anything else you would like me to do or fix?

Godson-Ozioma: Hi @magdmartin 

For the license text, I used the BSD license text because every existing file with “OpenRefine contributors” (e.g. ValidateHostHandler.java, ForRangeTests.java) uses that same block, and the forum mentions BSD-3-Clause. 

I searched the repo for “2026” and “SPDX” and didn’t find any file with the new format yet, so I followed the forum and those existing OpenRefine-contributors headers, only updating the year to 2026. 

Happy to change it if you prefer something different.

Godson-Ozioma: Hi @magdmartin  and @Abbe98 

Just a gentle follow-up in case this got buried.

All requested changes (including the copyright header update) have been addressed.  
There are currently 4 workflows awaiting approval, happy to do anything needed from my side.

Let me know if there’s anything else you'd like adjusted.

Thanks again for the review!

magdmartin: @Godson-Ozioma, I can only comment on the Copyright part and it looks good to me.

Godson-Ozioma: Thanks @magdmartin  

I appreciate you reviewing the copyright update 

If there’s someone specific who should take a look at the feature/validation changes, I’m happy to tag them.

Thanks again!

Godson-Ozioma: Hi @magdmartin and @Abbe98 

Just a quick follow-up in case this got buried.

All requested changes (including the copyright header update) have been addressed, and I’ve pushed the latest updates. The PR should now reflect the full fix for #7659 including backend validation, frontend updates, and tests.

Please let me know if there’s anything else I should adjust or improve. Happy to make any further changes.

Thanks again for taking the time to review!

tfmorris (changes requested):
Apologies for the delay. Thanks for the ping/reminder.

- please address Copilot's comment on the regression
- please run `./refine lint` to resolve the linting issues
- although either license is acceptable, let's use the MIT license just to be consistent with the rest of the extension (but keep the OpenRefine contributors copyright)

Godson-Ozioma:
> Apologies for the delay. Thanks for the ping/reminder.
> 
> * please address Copilot's comment on the regression
> * please run `./refine lint` to resolve the linting issues
> * although either license is acceptable, let's use the MIT license just to be consistent with the rest of the extension (but keep the OpenRefine contributors copyright)

Hi @tfmorris 

I’ve addressed the review feedback and pushed the changes. Summary:

WbLanguageConstant – _langId is now set in the constructor (and still updated in validate() with the normalized value) to fix the serialization/evaluation regression.

LanguageCodeStoreTest – The file now uses the MIT license text while keeping the “Copyright (c) 2026, OpenRefine contributors” line.

lang-suggest.js – DEFAULT_LANGUAGES_MONOLINGUALTEXT has been refreshed to match the Java DEFAULT_LANGUAGE_CODES list.

Formatting / CI – Ran ./refine lint and committed the formatted files so formatter:validate and impsort:check pass locally; the server CI lint step should pass now.

I ran mvn -B formatter:validate impsort:check and mvn test -pl extensions/wikibase locally; both completed successfully. Please let me know if you’d like any further changes.

extensions/wikibase/src/org/openrefine/wikibase/schema/WbLanguageConstant.java
```diff
@@ -46,40 +49,64 @@ public class WbLanguageConstant extends WbConstantExpr<String> {
 
     @JsonCreator
     public WbLanguageConstant(@JsonProperty("id") String langId, @JsonProperty("label") String langLabel) {
-        _langId = normalizeLanguageCode(langId);
-        _langLabel = langLabel;
         _origLangId = langId;
+        _langId = langId;
+        _langLabel = langLabel;
```
tfmorris, line 53: Let's restore the original assignment order to minimize churn and make it easier to review the diff, but I'm also wondering if the `normalizeLanguageCode(langId)` needs to be restored. Was it removed for a reason?

extensions/wikibase/src/org/openrefine/wikibase/schema/WbLanguageConstant.java
```diff
@@ -117,12 +144,13 @@ public boolean equals(Object other) {
             return false;
         }
         WbLanguageConstant otherConstant = (WbLanguageConstant) other;
-        return _langId.equals(otherConstant.getLang()) && _langLabel.equals(otherConstant.getLabel());
+        return Objects.equals(_origLangId, otherConstant._origLangId)
+                && Objects.equals(_langLabel, otherConstant._langLabel);
```
tfmorris: Is there a reason that this was switched from `getLang()` to `_origLangId` and the language label was added to the comparison? It seems like the language id should be sufficient.

extensions/wikibase/src/org/openrefine/wikibase/schema/WbLanguageConstant.java
```diff
@@ -117,12 +144,13 @@ public boolean equals(Object other) {
...
     }
 
     @Override
     public int hashCode() {
-        return _langId.hashCode();
+        return Objects.hash(_origLangId, _langLabel);
```
tfmorris: Same comment as for equals()

tfmorris (commented):
This is looking really good -- and it's a big task, so thanks for tackling it.

Just a few comments/questions remaining...

extensions/wikibase/src/org/openrefine/wikibase/schema/WbLanguageConstant.java
Godson-Ozioma: I kept normalization out of the constructor because normalization depends on the API endpoint and language context (term vs monolingualtext), which we only have in validate(). So the constructor only sets _langId = langId so that getLang() and JSON serialization aren’t null before validation; validate() then overwrites _langId with the normalized value. 

I’ll restore the original assignment order and move _langId = langId to the end to keep the diff minimal.

extensions/wikibase/src/org/openrefine/wikibase/schema/WbLanguageConstant.java
Godson-Ozioma: I used _origLangId so equality was based on the value as given (before normalization), and I included _langLabel so two constants with the same code but different labels would not be considered equal. Looking at it again, you’re right, language id should be enough for equality. 

I’ll change it back to compare only the normalized language id via getLang() and remove _langLabel from the comparison.

extensions/wikibase/src/org/openrefine/wikibase/schema/WbLanguageConstant.java
Godson-Ozioma: just like equals() I will replace _origLangId and _langLabel with getLang()

Godson-Ozioma:
hi @tfmorris 

I have implemented all three comments.
- Moved _langId = langId to the end to match the original order 
- equals(): Switched back to comparing only the normalized language id via getLang() and removed _langLabel from the comparison.
- hashCode(): Updated to use getLang() only so it stays consistent with equals().

Kindly let me know if there is anything else you need me to update

tfmorris (approved):
Looks great. Thanks for sticking with it and apologies again for the delay.

PR #7672, opened 2026-02-27
Fix: override upstream tablesorter styles

Abbe98:
This commit overrides several upstream styles to:

1. ensure that the project table list is aligned with the project bar and footer content
2. ensure that the bottom of the table has a border
3. ensure that the table does not overflow horizontally

Before:
Screenshot from 2026-02-27 19-07-17
Screenshot from 2026-02-27 19-07-28

After:
Screenshot from 2026-02-27 19-07-49
Screenshot from 2026-02-27 19-07-36

### DIFF
--- a/main/webapp/modules/core/styles/index/open-project-ui.css
+++ b/main/webapp/modules/core/styles/index/open-project-ui.css
@@ -246,3 +246,11 @@ ul#tagsUl li a {
   text-align: center;
   font-size: x-large;
 }
+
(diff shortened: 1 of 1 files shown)


tfmorris:
Do we have an issue for this? Is it a regression from a tablesorter update or just something that was never noticed before?

Abbe98:
It's a regression but an old one. I don't think there is an issue.

tfmorris (approved):

PR #7689, opened 2026-03-02
client: get encodings using API rather than Velocity injection

Abbe98:
Builds upon the work in #6186 leaving only style and script injections in place. 

Fixes (yeah this should probably get an issue).

### DIFF
--- a/main/src/com/google/refine/commands/GetEncodingsCommand.java
+++ b/main/src/com/google/refine/commands/GetEncodingsCommand.java
@@ -27,10 +27,9 @@ public class GetEncodingsCommand extends Command {
             Charset charset = entry.getValue();
 
             Map<String, Object> encoding = Map.of(
-                "code", code,
-                "name", charset.displayName(),
-                "aliases", new ArrayList<>(charset.aliases())
-            );
+                    "code", code,
+                    "name", charset.displayName(),
+                    "aliases", new ArrayList<>(charset.aliases()));
 
             encodings.add(encoding);
         }


Abbe98:
@copilot can you run the java linting here?

main/webapp/modules/core/index.vt
```diff
@@ -40,7 +40,19 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
   <script type="text/javascript" src="wirings.js"></script>
   $scriptInjection
   $styleInjection
-  <script>Refine.encodings = $encodingJson;</script>
+  <script type="text/javascript">
+    // Fetch encodings from API
+    (function() {
+      var xhr = new XMLHttpRequest();
+      xhr.open('GET', 'command/core/get-encodings', false); // synchronous
```
tfmorris, line 47: Do these calls need to be synchronous? We've got a bunch of synchronous calls already at startup and it would be good to not add more if we can help it.

Abbe98: Good point, I will rework it. Do you know if extensions use window.Refine.encodings? Otherwise we could simply move the call to `encoding.js` and not have it sitting around in these files at all.

tfmorris: I'm not aware of any that would have a need to use it and a quick search of Github shows mostly OpenRefine clones with a smattering of repackagings like LODRefine.

Since we've decided that the next release will be a breaking release, I'm happy to live with the (low) risk that changing it will break something somewhere in the universe.

PR #7698, opened 2026-03-11
Add Unix domain socket support to RefineServer

### LINKED ISSUE #7700: Listen on Unix Domain Sockets
It would be great if OpenRefine could be served on a [unix domain socket](https://en.wikipedia.org/wiki/Unix_domain_socket) instead of a TCP socket.

Unix domain sockets are slightly more efficient that TCP sockets, and have the advantage that they can be secured using normal (posix) file permissions. This means that on a system with multiple users (and where multiple users have shell access), running a webapplication on unix sockets is more secure than on TCP sockets. Any user on the system can connect to the TCP socket, thereby getting access to another user's server (and potentially bypassing security, if this is implemented in a reverse proxy, by connecting directly on localhost). With a unix socket this is impossible.

We especially see a usecase for this on HPC systems (which sometimes don't even allow opening TCP ports).

Requirements:
- Allow launching OpenRefine with a a  user-defined socket path.
- Create the socket and listen on it (this can be handled by Jetty)
- Perhaps allow setting file permissions and umask for the socket (although we could also leave this up to the user -- you can control access to socket e.g. by passing in a path that is in a setgid directory, and by changing the umask for the `refine command`).

dometto:
Closes #7700

This implements support for Unix domain sockets (see e.g. https://jetty.org/docs/jetty/10/programming-guide/server/io-arch.html), which is valuable when using OpenRefine on a shared machine (e.g. an HPC system) with multiple users -- in that case, running on a TCP socket is not secure, as any user on the system can connect to the socket.

Our use case is launching OpenRefine via JupyterHub and a reverse proxy, spawning an instance for each user, on a shared cloud VM.

This PR modifies RefineServer's `init` to take a `socket` parameter. This can be passed in via the `refine` script using the `-s` switch. If `socket` is `null`, normal TCP socket with options (port, host, interface) will be used.

This doesn't yet add support to `refine.bat`, as I'm not able to test on Windows. I'm also not sure in what cases UDS support on Windows makes sense.

### DIFF
--- a/refine
+++ b/refine
@@ -93,8 +93,24 @@ check_downloaders() {
 
 check_running() {
     check_downloaders
-    URL="http://${REFINE_HOST_INTERNAL}:${REFINE_PORT}/"
+
     CHECK_STR="<title>OpenRefine</title>"
+    URL="http://${REFINE_HOST_INTERNAL}:${REFINE_PORT}/"
+
+    if [ -n "$REFINE_SOCKET" ] ; then
+        if [ "$CURL" ] ; then
+            RUNNING=`curl --unix-socket $REFINE_SOCKET --noproxy 127.0.0.1 -s http://localhost | grep "$CHECK_STR"`
+        else
+            # wget does not support connecting to unix domain sockets
+            # instead use a simple filesystem check to determine whether a file exists at the socket path
+            if [ -e "$REFINE_SOCKET" ] ; then
+                RUNNING="0"
+            else
+                RUNNING=""
+            fi
+        fi
+        return
+    fi
 
     if [ "$CURL" ] ; then
         curl --noproxy 127.0.0.1 -s -S -f $URL > /dev/null 2>&1
@@ -107,7 +123,7 @@ check_running() {
         if [ "$?" = "4" ] ; then
             NOT_RUNNING="1"
         fi
-    fi    
+    fi
 
     if [ -z "${NOT_RUNNING}" ] ; then
         if [ "$CURL" ] ; then
@@ -361,6 +377,36 @@ test() {
     $MVN test
 }
 
+uds_tests() {
+    add_option "-Drefine.headless=true"
+    add_option "-Drefine.socket=/tmp/openrefine_test.sock"
+    REFINE_SOCKET="/tmp/openrefine_test.sock"
+    check_running
+
+    if [ "$RUNNING" ] ; then
+        echo "An OpenRefine server is already running. Please shut it down so that we can start a test server."
+        exit
+    fi
+    echo "Starting OpenRefine server"
+
+    run fork > /dev/null
+
+    echo "Waiting for OpenRefine to load..."
+    sleep 5
+    check_running
+
+    if [ ! "$RUNNING" ] ; then
+        fail "OpenRefine failed to start on Unix Domain Socket."
+    else
+        echo ""
+        echo "Killing OpenRefine"
+        /bin/kill $REFINE_PID
+        echo "Cleaning up"
+        rm -rf "$REFINE_DATA_DIR"
+        return 0
+    fi
+}
+
 
 e2e_tests() {
     get_revision
@@ -435,8 +481,6 @@ e2e_tests() {
     echo ""
     echo "Killing OpenRefine"
     /bin/kill -9 $REFINE_PID
-    echo "Cleaning up"
-    rm -rf "$REFINE_DATA_DIR"
 
     if [ "$UI_TEST_SUCCESS" = "0" ] ; then
         error "The UI test suite failed."
@@ -748,6 +792,7 @@ case "$ACTION" in
   test) test $1;;
   tests) test $1;;
   e2e_tests) e2e_tests;;
+  uds_tests) uds_tests;;
   server_test) server_test $1;;  
   server_tests) server_test $1;;  
   extensions_test) extensions_test $1;;
--- a/refine
+++ b/refine
@@ -116,9 +116,9 @@ check_running() {
             RUNNING=`no_proxy=127.0.0.1 wget -O - $URL| grep "$CHECK_STR"` 
         fi    
         
-        # if [ -z "${RUNNING}" ] ; then
-        #     error "OpenRefine isn't running on $URL. Maybe a proxy issue?"
-        # fi
+        if [ -z "${RUNNING}" ] ; then
+            error "OpenRefine isn't running on $URL. Maybe a proxy issue?"
+        fi
     else
         RUNNING=""
     fi
--- a/refine
+++ b/refine
@@ -35,7 +35,7 @@ Options
     -i <interface>                The network interface OpenRefine should bind to. Default: 127.0.0.1.
     -m <memory>                   JVM min and max memory heap size. Default: 1400M.
     -p <port>                     Port that OpenRefine will listen to. Default: 3333.
-    -s <path>                     Path to unix domain socket at which the server will listen (interface and port will be ignored).
+    -s <path>                     Path to Unix domain socket at which the server will listen
     -v <level>                    Verbosity level [error,warn,info,debug,trace]. Default: info.
     -w <path>                     Path to the webapp. Default: main/webapp.
     -x <name=value>               Additional configuration parameters to pass to OpenRefine.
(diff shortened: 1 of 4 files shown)


dometto:
This still needs some tweaking. When serving over a UDS I get:

```
HTTP ERROR 404 Invalid hostname
```

tfmorris:
Thanks for the contribution, but please fill out the PR template, which includes a link to the issue that this is proposed to fix. If there isn't an issue yet, please create one so that the requirements can be discussed (as opposed to the implementation).

From an implementation point of view, you mention Jetty 12, but we're currently on Jetty 10.

tfmorris:
p.s. Your error sounds a lot like something you'd get from this:
https://github.com/OpenRefine/OpenRefine/blob/5ea573d5be4c5be8ba186947d0118e83ef83e13a/server/src/com/google/refine/ValidateHostHandler.java#L97

dometto:
> Thanks for the contribution, but please fill out the PR template, which includes a link to the issue that this is proposed to fix. If there isn't an issue yet, please create one so that the requirements can be discussed (as opposed to the implementation).

Will do and apologies for not filling out the template!

I linked to jetty 12 just because I was looking for an explanation of unix domain sockets in a jetty context -- the implementation doesn't actually depend on v12. Anyway I'll first open an issue to discuss requirements!

[EDIT: linked to the same page in the jetty 10 docs]

dometto:
> p.s. Your error sounds a lot like something you'd get from this:

Ah, indeed, I was incorrectly setting the host header!

That means this implementation works in the basics. Curious what you and other maintainers think of the feature request at https://github.com/OpenRefine/OpenRefine/pull/7698

server/src/com/google/refine/Refine.java
```diff
@@ -173,22 +183,32 @@ private static ThreadPool createThreadPool() {
...
         httpConfig.setSendServerVersion(false);
         HttpConnectionFactory httpFactory = new HttpConnectionFactory(httpConfig);
-        ServerConnector connector = new ServerConnector(this, httpFactory);
-        connector.setPort(port);
-        connector.setHost(iface);
-        connector.setIdleTimeout(Configurations.getInteger("server.connection.max_idle_time", 60000));
-        this.addConnector(connector);
+
+        int maxIdleTime = Configurations.getInteger("server.connection.max_idle_time", 60000);
+        if (socket == null) {
```
tfmorris, line 200: Is there a reason to force this to be an alternative rather than allowing both?

server/src/com/google/refine/Refine.java
```diff
@@ -173,22 +183,32 @@ private static ThreadPool createThreadPool() {
 
     private ThreadPoolExecutor threadPool;
 
-    public void init(String iface, int port, String host) throws Exception {
+    public void init(String iface, int port, String host, String socket) throws Exception {
```
tfmorris, line 188: Please keep the 2-arg form for compatibility since this is a public method.

refine
```diff
@@ -35,6 +35,7 @@ Options
     -i <interface>                The network interface OpenRefine should bind to. Default: 127.0.0.1.
     -m <memory>                   JVM min and max memory heap size. Default: 1400M.
     -p <port>                     Port that OpenRefine will listen to. Default: 3333.
+    -s <path>                     Path to unix domain socket at which the server will listen (interface and port will be ignored).
```
tfmorris:
```suggestion
    -s <path>                     Path to Unix domain socket at which the server will listen
```

refine
```diff
@@ -115,9 +116,9 @@ check_running() {
             RUNNING=`no_proxy=127.0.0.1 wget -O - $URL| grep "$CHECK_STR"` 
         fi    
         
-        if [ -z "${RUNNING}" ] ; then
-            error "OpenRefine isn't running on $URL. Maybe a proxy issue?"
-        fi
+        # if [ -z "${RUNNING}" ] ; then
+        #     error "OpenRefine isn't running on $URL. Maybe a proxy issue?"
+        # fi
```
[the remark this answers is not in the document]
tfmorris: In addition to restoring this code, a running check for the socket version is also needed.

tfmorris (changes requested):
In addition to the other comments, this needs at least a basic test for CI so that we know it's working (and don't break it in the future).

server/src/com/google/refine/Refine.java
dometto: Have the application listen simultaneously on both a TCP and UDS you mean? To be honest it never occurred to me that someone might want to do this! I guess it's technically possible, but I've never seen a usecase for this, as the reason for using UDS is usually that we *don't* want to bind on a TCP port for security reasons!

On the implementation level, this would require checking that IP and port aren't implicitly assigned a default, so that when someone sets `refine.socket` in the expectation that nothing will be served on TCP, a default TCP port isn't opened anyway.

From my perspective, I'm not sure it's worth the trouble. The app currently also doesn't allow specifying multiple TCP ports (I think), so why add functionality to listen on multiple sockets? Also FWIW, I've not seen other apps with UDS support allow serving on both UDS and a TCP socket (but this may also be because it's more difficult to achieve this in other languages and frameworks).

tfmorris: I'm OK with leaving it as is to start and expanding it later if someone identifies a need. Multi-protocol support is on my mind because I've been looking at HTTP/2 & HTTP/3.

refine
```diff
@@ -664,29 +670,31 @@ if [ -z "$REFINE_MAX_FORM_CONTENT_SIZE" ] ; then
...
+        fi
+    fi
+    
+    if [ "$REFINE_HOST" = '*' ] ; then
+        echo No host specified while binding to interface 0.0.0.0, guessing localhost.
+        REFINE_HOST_INTERNAL="localhost"
     else
-        REFINE_HOST="$REFINE_INTERFACE"
+        REFINE_HOST_INTERNAL="$REFINE_HOST"
     fi
```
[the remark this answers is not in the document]
dometto: I've opted to refactor `check_running` to take into account `REFINE_SOCKET`. This only works with `curl`, as `wget` does not support connect to UDS. A workaround could be to check for existence and mode of the socket if CURL is not available. Should I implement that?

dometto:
@tfmorris what would be an appropriate CI test for UDS support -- should I add one to the `e2e` or `server` tests?

tfmorris:
A simple test that starts the server on a UDS and confirms that you can connect to it and execute simple command (e.g. `get-version`) should be sufficient. It doesn't have to be anything fancy since the rest of the functionality is tested via HTTP and I wouldn't expect any problems as long as the basic connection works.
EDIT: Actually, it looks like the `refine` script just checks for the HTML home page title and that would be fine too.

The CI failure can likely be fixed by using `./refine lint`

dometto:
>  EDIT: Actually, it looks like the refine script just checks for the HTML home page title and that would be fine too.

I've added a test to the e2e workflow [here](https://github.com/OpenRefine/OpenRefine/pull/7698/changes#diff-9c82dc32e27ac14dd71847fd7b9cac749344d63ba69c9cc5d1e0278748d266d6). It feels a little awkward: on the one hand the `check_running` function in `refine` perform exactly the basic test that we want, so it makes sense to place the test in the e2e workflow, which already utilises the `refine` script. On the other hand, the e2e workflow uses a matrix of npm versions and browsers, which is not relevant to UDS support.

If you'd rather have me move this to the `server` workflow, I can look into that.

dometto:
In the end I've opted to put the UDS tests in a new job in the 'pull_request_e2e' workflow, which I've tentatively named `misc_integration_test`. This removes the test from the NPM/Cypress test matrix. Since we want to test the `refine` script's functionality as well, I think it fits here better that in the `pull_request_server` workflow?

I think all comments have been addressed so looking forward to another review. Also happy to rebase commits if you're not going to squash them anyway :)

refine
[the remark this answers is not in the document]
tfmorris: I think having a dependency on curl is OK, given how rare the usage is likely to be.

tfmorris (approved):

dometto:
@tfmorris any chance there could be a new release any time soon-ish? A release containing this PR is a show-stopper for us deploying OpenRefine at the moment, and I'd really like to add it to our managed JupyterHub environments. :)

magdmartin:
@dometto  I can't commit to the next release date, but in the meantime, you have access to the [snapshot release](https://github.com/openrefine/openrefine#snapshot-releases) if that could help.

tfmorris:
I had targeted this for our next major release, but it's small enough that we can probably slip it into 3.10.2

dometto:
> I had targeted this for our next major release, but it's small enough that we can probably slip it into 3.10.2

thanks, that would be amazing!

PR #7701, opened 2026-03-12
Update feature_request template

magdmartin:
Adding the Acceptance Criteria as part of the proposed solution section, following the [2025 Barcamp discussion](https://forum.openrefine.org/t/2025-barcamp-session-proposal-discussing-openrefine-roadmap-and-goal-post/2539)

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -9,8 +9,8 @@ assignees: ''
 
 <!-- Please provide a clear and concise description of your problem or unsatisfied needs. Ex. I'm always frustrated when [...] or, It would be easier if OpenRefine did [...]. This comment can be deleted, if desired, but it will be hidden in your final submission. Please make sure that your new text is outside the enclosing angle brackets. -->
 
-### Proposed solution
-<!-- If you have a proposal for how this need could be met, please provide a clear and concise description of what you want to happen. -->
+### Acceptance criteria or proposed solution
+<!-- Please describe the requirement and acceptance criteria for this feature. If you have a proposal for how this need could be met, please provide a clear and concise description of what you want to happen. -->
 
 ### Alternatives considered
 <!-- If there alternative solutions that you have considered or think should be considered, please list them here -->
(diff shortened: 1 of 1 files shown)


Abbe98:
I think the old phrasing is much better as it frames it as a wish while this frames it as if the reporter can set requirements and acceptance criteria something maintainers are much better suited to do. 

Looking at the Barcamp notes, is it actually the "proposed solution" section which needs work or may it actually be the first section where one would expect a use case to be described?

Currently that instruction is a mix of how comments work and what to put in there. Maybe the instruction shoud be split into two parts and the one about how to write your use cases should be expanded with good examples?

tfmorris:
The notes say:

> One suggestion was to reconsider the “proposed solution” section in the [GitHub feature request template](https://github.com/OpenRefine/OpenRefine/issues/new?template=feature_request.md). Jan also mentioned that Wikimedia projects often include acceptance criteria to clarify when a feature is considered complete.

so I think that's the right section, but I'd split things into three parts:

1. Problem statement - what is broken or could be improved?
2. Acceptance criteria - how will we know when we've fixed it?
3. Proposed solution (optional) - if you know of a potential solution, please suggest it, but focus on the requirements first

We could also add "below the fold" sections to the template which could be used as part of the refinement process. I see these as living documents, not thing that gets authored once and then accepted or rejected.

Perhaps we can find a good template from another project to use as a basis?

magdmartin:
Thanks all for the feedback, this helped clarify the direction.

This update tries to reflect both the Barcamp discussions and the points raised here:

- Put more emphasis on the use case/need in the first section
- Separate the expected outcome from the proposed solution, keeping the latter optional
- Avoid “acceptance criteria” wording, which felt a bit too formal in this context
- Add a lightweight “Discussion notes” section to support the idea of issues evolving over time (as suggested with the “below the fold” sections)

The intent is to keep the template approachable for contributors, while giving enough structure for maintainers to refine issues into roadmap / goal-post items as the discussion progresses.

Happy to adjust further if this still feels off.

tfmorris (approved):

PR #7705, opened 2026-03-13
Don't escape backslashes on TSV export

### LINKED ISSUE #7704: TSV exporter escapes backslashes
Per https://forum.openrefine.org/t/extra-backslashes-in-marc-fields-after-openrefine-tsv-export/2748/1
behavior of the TSV exporter has changed from not escaping any characters to escape the default \t \n \r \\

### To Reproduce
1. Export a TSV file with cells containing backslashes
2. Examine the file

### Current Results
Backslashes are doubled.

### Expected Behavior
Backslashes are untouched.

### Versions
 - Operating System: 
 - Browser Version: 
 - JRE or JDK Version: 
 - OpenRefine: 3.10-SNAPSHOT

tfmorris:
Fixes #7704

Restores historical behavior of previous TSV exporter.

- Add a TSV exporter test with backslash data
- Fix exporter to not escape backslashes

Note: this really should be controllable by the user since it makes output data ambiguous in the presence of other escaping.

We have existing test coverage for exporting tabs and newlines, so those are confirmed to still be correct.

### DIFF
--- a/main/src/com/google/refine/exporters/CsvExporter.java
+++ b/main/src/com/google/refine/exporters/CsvExporter.java
@@ -58,8 +58,6 @@ import com.google.refine.util.ParsingUtilities;
 public class CsvExporter implements WriterExporter {
 
     static CsvFormat DEFAULT_FORMAT = new CsvWriterSettings().getFormat();
-    static TsvFormat TSV_FORMAT = new TsvWriterSettings().getFormat();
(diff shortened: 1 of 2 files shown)


tfmorris:
I could have sworn this worked on my local machine, but not only does it not work, but in looking at the implementation, there doesn't appear to be a way to do this, at least with TsvWriter. Perhaps it's possible using the CsvWriter with a delimiter of `\t`.

PR #7719, opened 2026-03-19
Fix Python expression preview not showing errors. Fixes #3012

### LINKED ISSUE #3012: Python expression preview doesn't always report errors

ishankguptaaa:
## Summary

Fixes #3012

The inner `catch (Exception e)` block in `PreviewExpressionCommand.doPost()` silently swallowed exceptions during expression evaluation (`// ignore`), causing errors to display as `null` in the preview instead of proper error messages.

This was particularly noticeable with Python/Jython expressions using methods that don't exist on Python objects (e.g., `value.toNumber()`), where the error traceback was generated but never surfaced to the user.

### Changes
- **PreviewExpressionCommand.java**: Convert caught exceptions to `EvalError` using the `EvalError(Throwable)` constructor (which uses `e.toString()` for null-safe messages) instead of silently ignoring them
- **PreviewExpressionCommandTests.java**: Added unit test verifying that evaluation errors return error objects with messages, not null values
- **expressions.cy.js**: Added Cypress E2E regression test reproducing the exact #3012 scenario (Python `.toNumber()` on a string) and asserting `Error:` is shown instead of `null`

## Test plan

- [x] `./refine lint` passes
- [x] `mvn test -pl main -Dtest="PreviewExpressionCommandTests"` — 4/4 tests pass
- [x] New unit test `testEvalError` validates error objects are returned
- [x] New Cypress test validates Python expression errors display correctly in preview

### DIFF
--- a/extensions/jython/src/com/google/refine/jython/JythonEvaluable.java
+++ b/extensions/jython/src/com/google/refine/jython/JythonEvaluable.java
@@ -73,7 +73,7 @@ public class JythonEvaluable implements Evaluable {
                 try {
                     return new JythonEvaluable(source, languagePrefix);
                 } catch (PySyntaxError e) {
-                    throw new ParsingException("Syntax error");
+                    throw new ParsingException(e.getMessage(), e);
                 }
             }
 
--- a/extensions/jython/tests/src/com/google/refine/jython/JythonEvaluableTest.java
+++ b/extensions/jython/tests/src/com/google/refine/jython/JythonEvaluableTest.java
@@ -147,7 +147,7 @@ public class JythonEvaluableTest {
         assertEquals(evaluable.getLanguagePrefix(), "jython");
     }
 
-    @Test(expectedExceptions = ParsingException.class)
+    @Test(expectedExceptions = ParsingException.class, expectedExceptionsMessageRegExp = ".*SyntaxError.*")
     public void testParseError() throws ParsingException {
         JythonEvaluable.createParser().parse("foo(", "jython");
     }
--- a/main/src/com/google/refine/commands/expr/PreviewExpressionCommand.java
+++ b/main/src/com/google/refine/commands/expr/PreviewExpressionCommand.java
@@ -190,7 +190,7 @@ public class PreviewExpressionCommand extends Command {
                                 }
                             }
                         } catch (Exception e) {
-                            // ignore
+                            result = new EvalError(e);
                         }
                     }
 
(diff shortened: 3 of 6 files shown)


ishankguptaaa:
## Local Verification

Tested locally by running OpenRefine and hitting the `preview-expression` endpoint directly.

### Before (on `master`): errors were silently swallowed, results showed `null`

The old code had `catch (Exception e) { // ignore }`, so when a Python expression like `value.toNumber()` threw an error, `result` stayed `null` and the preview displayed "null" instead of an error message.

### After (this PR): errors are properly reported

```
POST /command/core/preview-expression
expression=jython:return value.toNumber()
```

Response:
```json
{
    "code": "ok",
    "results": [
        { "message": "AttributeError: 'unicode' object has no attribute 'toNumber'" },
        { "message": "AttributeError: 'unicode' object has no attribute 'toNumber'" },
        { "message": "AttributeError: 'unicode' object has no attribute 'toNumber'" }
    ]
}
```

The frontend's `renderValue()` correctly detects these objects (via `$.isPlainObject(v)`) and renders them as red error text.

### Working expression (for comparison)

```
POST /command/core/preview-expression
expression=jython:return value.upper()
```

Response:
```json
{
    "code": "ok",
    "results": [ "HELLO", "FOO", "123" ]
}
```

### Tests
- `./refine lint` — **PASS**
- `mvn test -pl main -Dtest="PreviewExpressionCommandTests"` — **4/4 PASS**

tfmorris:
Thanks for the contribution! Can you please add before and after screenshots to show the change visually?

ishankguptaaa:
> Thanks for the contribution! Can you please add before and after screenshots to show the change visually?

Important point to note here is the fix is defensive because as per my analyse please read the following:
  - JythonEvaluable catches PyException internally and returns EvalError
  - GREL evaluator catches errors internally and returns EvalError
  - Clojure evaluator catches Exception internally and returns EvalError

All three evaluators handle their own errors. The outer catch block is only a defensive fallback for unexpected Java exceptions that slip through evaluator error handling. Hence the second fix.

Image:
image

for jython:
image

ishankguptaaa:
hey @tfmorris , might have missed the PR please do review

tfmorris (approved):
Looks great. Thanks for the contribution and apologies for the delay in getting it reviewed!

PR #7721, opened 2026-03-20
feat(clustering): Replace Vicino with Apache Commons Text Levenshtein

### LINKED ISSUE #7726: Migrate Levenshtein clustering to a modern implementation

thadguidry:
Fixes #7726 

Switch the kNN clustering Levenshtein distance from Vicino library to Apache Commons Text, which is already a transitive dependency of the project.

The new ApacheLevenshteinDistance wrapper implements SimilarityDistance and delegates to org.apache.commons.text.similarity.LevenshteinDistance. Benchmarks show ~8.5x faster distance computation and ~6.5x faster end-to-end kNN clustering compared to the Vicino implementation.

Changes:
- Add ApacheLevenshteinDistance wrapper class in `modules/core`
- Register ApacheLevenshteinDistance for 'levenshtein' in `controller.js`
- Add commons-text dependency to `modules/core/pom.xml`
- Add JMH benchmarks: ApacheLevenshteinBenchmark and KNNLevenshteinClusteringBenchmark with benchmark README
- Update tests to use the new Apache-backed implementation

Discussion here: https://forum.openrefine.org/t/faster-clustering/2753


modules/core/src/test/java/com/google/refine/clustering/knn/kNNClustererTests.java
```diff
@@ -56,10 +56,17 @@ public class kNNClustererTests extends RefineTest {
     public static String clustererJson = "["
             + "   [{\"v\":\"ab\",\"c\":1},{\"v\":\"abc\",\"c\":1}]"
             + "]";
+    public static String levenshteinConfigJson = "{"
+            + "\"type\":\"knn\","
+            + "\"function\":\"levenshtein\","
+            + "\"column\":\"values\","
+            + "\"params\":{\"radius\":1,\"blocking-ngram-size\":2}"
```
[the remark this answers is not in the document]
tfmorris: I prefer "values" as the column name, but it looks like this is a bug in the existing code that was copy/pasted, so let's get rid of the duplication by creating a little helper method so we can fix the bug in a single place.

benchmark/src/main/java/org/openrefine/benchmark/KNNLevenshteinClusteringBenchmark.java
```diff
@@ -0,0 +1,245 @@
...
+    @OutputTimeUnit(TimeUnit.MILLISECONDS)
+    @Warmup(iterations = 1, time = 200, timeUnit = TimeUnit.MILLISECONDS)
+    @Measurement(iterations = 2, time = 300, timeUnit = TimeUnit.MILLISECONDS)
+    @Fork(1)
+    public void apacheKNNClustering(ExecutionPlan plan, Blackhole blackhole) throws IOException {
+        blackhole.consume(runKnnClustering(plan.values, APACHE_NAME));
+    }
+
+    private int runKnnClustering(List<String> values, String distanceName) throws IOException {
+        Project project = createProject(values);
```
tfmorris: This `createProject()` should be moved up/out to the setup phase.

modules/core/src/main/java/com/google/refine/clustering/knn/ApacheLevenshteinDistance.java
```diff
@@ -0,0 +1,43 @@
...
+package com.google.refine.clustering.knn;
+
+import org.apache.commons.text.similarity.LevenshteinDistance;
+
+/**
+ * Wrapper around Apache Commons Text Levenshtein implementation.
+ */
+public class ApacheLevenshteinDistance implements SimilarityDistance {
+
+    private final LevenshteinDistance distance = new LevenshteinDistance();
```
tfmorris:
```suggestion
    private final LevenshteinDistance distance = LevenshteinDistance.getDefaultInstance();
```

This constructor is apparently deprecated (surprised Copilot didn't catch that).

benchmark/src/main/java/org/openrefine/benchmark/KNNLevenshteinClusteringBenchmark.java
```diff
@@ -0,0 +1,245 @@
...
+
+    private static final Path LARGE_DATASET_PATH = Path.of("main/tests/data/acm_large.txt");
+    private static final Path CONTRACTS_DATASET_PATH = Path.of("main/tests/data/government_contracts.csv");
+    private static final int MIN_DATASET_SIZE = 5000;
+    private static final String VALUE_COLUMN = "value";
+    private static final String VICINO_NAME = "benchmark-vicino-levenshtein";
+    private static final String APACHE_NAME = "benchmark-apache-levenshtein";
+
+    @State(Scope.Benchmark)
+    public static class ExecutionPlan {
```
tfmorris, line 76: JMH is great for micro benchmarks like an individual function (e.g. LevenshteinDistance or ToNumber), but is probably overkill for benchmarks who's single iteration duration is in the seconds.

It doesn't need to be changed, but something to keep in mind for future clustering benchmarks.

tfmorris (changes requested):
Thanks for implementing this and, especially, for including a benchmark.

The Copilot comments seem valid, so please address those (I did some of the mechanical ones like indentation).

Since we're changing the underlying implementation, we should probably strive for a little better test coverage with multiple N-gram blocking factors and distance thresholds so that we're confident that the behavior is the same between the new and the old implementation.

I actually saw an even bigger improvement, particularly without project creation in the loop:

```bash
mvn -pl benchmark -am -DskipTests package
java -jar benchmark/target/openrefine-benchmarks.jar KNNLevenshteinClusteringBenchmark

Benchmark                                              (rowCount)  Mode  Cnt     Score   Error  Units
KNNLevenshteinClusteringBenchmark.apacheKNNClustering         200  avgt    2    52.040          ms/op
KNNLevenshteinClusteringBenchmark.apacheKNNClustering         500  avgt    2   323.397          ms/op
KNNLevenshteinClusteringBenchmark.vicinoKNNClustering         200  avgt    2   411.101          ms/op  7.9x
KNNLevenshteinClusteringBenchmark.vicinoKNNClustering         500  avgt    2  3355.692          ms/op 10.3x

with createProject() removed from inner loop
Benchmark                                              (rowCount)  Mode  Cnt     Score   Error  Units
KNNLevenshteinClusteringBenchmark.apacheKNNClustering         200  avgt    2    41.118          ms/op
KNNLevenshteinClusteringBenchmark.apacheKNNClustering         500  avgt    2   301.346          ms/op
KNNLevenshteinClusteringBenchmark.vicinoKNNClustering         200  avgt    2   420.840          ms/op 10.2x
KNNLevenshteinClusteringBenchmark.vicinoKNNClustering         500  avgt    2  3550.328          ms/op 11.8x
```

This is without any loop optimizations or the use of the threshold. It also continues using the Vicino N-gram clusterer for blocking

NOTE: The benchmarks hang for 30 seconds after each iteration due to Vicino using a thread executor pool with an idel timeout of 60 seconds, which is annoying, but probably not worth fixing since the whole goal of the exercise it to get rid of Vicino

benchmark/src/main/java/org/openrefine/benchmark/KNNLevenshteinClusteringBenchmark.java
thadguidry: ah good catch

modules/core/src/test/java/com/google/refine/clustering/knn/kNNClustererTests.java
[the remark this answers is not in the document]
thadguidry: OK, I've updated this, take a look.

benchmark/src/main/java/org/openrefine/benchmark/KNNLevenshteinClusteringBenchmark.java
thadguidry: Right, so for longer duration interations, we'd likely choose something like `System.nanoTime()` loop?

tfmorris (approved):

PR #7728, opened 2026-03-24
Replace Vicino PPM distance with Deflate-based NCD implementation

thadguidry:
(also updates the existing `/benchmark/README.md` from the #7721 with a new section)

### Benchmark Results

Benchmark results confirm Deflate NCD is 3-8x faster than Vicino PPM:

| Benchmark | Score (µs/op) | Error (±) | Speedup |
|:---|---:|---:|---:|
| `deflateShortStrings` | 5.835 | 0.129 | — |
| `vicinoPPMShortStrings` | 14.471 | 8.818 | — |
| **Short string improvement** | | | **2.5x** |
| `deflateMediumStrings` | 6.924 | 0.294 | — |
| `vicinoPPMMediumStrings` | 34.851 | 19.990 | — |
| **Medium string improvement** | | | **5.0x** |
| `deflateLongStrings` | 7.862 | 0.069 | — |
| `vicinoPPMLongStrings` | 67.377 | 2.950 | — |
| **Long string improvement** | | | **8.6x** |

### Summary  (Generated by Claude Opus 4.6)

This replaces the PPM (Prediction by Partial Matching) arithmetic coding distance from the unmaintained Vicino library with a Normalized Compression Distance (NCD) implementation using Java's built-in `java.util.zip.Deflater`. This directly addresses the concern about finding "a quicker (and better supported) algorithm than our current arithmetic coder with the PPM model."

### Research: How Vicino's PPM Distance Works

I decompiled and analyzed Vicino's implementation to fully understand what we're replacing. The architecture has two layers:

**`PPMDistance` extends `PseudoMetricDistance`**, which is the base class shared by all of Vicino's compression-based distances (PPM, GZip, BZip2). The `d2(a, b)` method in `PPMDistance` concatenates strings `a + b`, creates a `ByteArrayOutputStream`, wraps it in an `ArithCodeOutputStream` with a `PPMModel(order=8)` from the `com.colloquial.arithcode` library, writes the concatenated string's bytes through the arithmetic coder, closes the stream, and returns the compressed byte array length as a double. This is essentially measuring "how many bytes does PPM need to encode the concatenation of a and b."

**`PseudoMetricDistance.d(a, b)`** then uses `d2()` to compute a Normalized Compression Distance with the formula: `10.0 * ((d2(a,b) + d2(b,a)) / (d2(a,a) + d2(b,b)) - 1.0)`. This is a symmetrized variant of the standard NCD formula from the Cilibrasi & Vitányi paper ("Clustering by Compression", IEEE Trans. Information Theory, 2005). The idea is that if two strings are similar, compressing their concatenation won't be much larger than compressing either one alone, because the compressor can exploit the redundancy. The factor of 10 is a scaling constant and the subtraction of 1 shifts the baseline so identical strings yield distance 0.

The arithmetic coder + PPM model from `com.colloquial.arithcode` is the core dependency that makes this problematic — it's an obscure, unmaintained library that provides no performance advantage over well-supported compressors for the string lengths typically seen in data cleaning.

### Why Deflate Over PPM

Several factors guided the choice of Java's built-in `Deflater` with `BEST_COMPRESSION`:

1. **Zero additional dependencies.** `java.util.zip.Deflater` is part of the JDK and delegates to native zlib, which is one of the most battle-tested compression libraries in existence. This eliminates the `com.colloquial.arithcode` transitive dependency entirely.

2. **Performance.** Deflate is implemented in native code (zlib) called via JNI, whereas Vicino's PPM arithmetic coder is pure Java. For the short strings typical in data cleaning (names, addresses, organizations), the JNI overhead is amortized and native compression is significantly faster. The included JMH micro-benchmark allows direct comparison.

3. **NCD quality.** The academic literature on NCD notes that the compressor choice affects quality — better compressors yield NCD values closer to the theoretical Normalized Information Distance. In practice, for the short strings in data cleaning, all reasonable compressors (PPM, Deflate, GZip, BZip2) produce similar clustering results. Vicino's own codebase had `GZipDistance` and `BZip2Distance` alongside `PPMDistance`, all using the same `PseudoMetricDistance` NCD formula, which confirms the compressor is interchangeable.

4. **Maintainability.** Using JDK standard library means no version conflicts, no CVE tracking for third-party compression libraries, and no risk of the dependency becoming unavailable.

I also considered the `java-string-similarity` library by tdebatty (2.7k GitHub stars), but it doesn't include any compression-based distance metrics — it focuses on edit distances, token-based distances, and n-gram distances. Apache Commons Text similarly doesn't offer NCD. Implementing NCD directly with `Deflater` was the most straightforward approach requiring no new dependencies.

### Implementation Details

**`DeflateNCDDistance`** implements `SimilarityDistance` and uses the exact same NCD formula as Vicino's `PseudoMetricDistance`: `10 * ((Z(a+b) + Z(b+a)) / (Z(a+a) + Z(b+b)) - 1)`. Each `Z()` call creates a fresh `Deflater(BEST_COMPRESSION)`, compresses the UTF-8 bytes, and returns the compressed size. The `Deflater` is properly cleaned up via `end()` in a finally block to release native resources.

The distance is registered under the same `"ppm"` key for backward compatibility with existing saved clustering configurations.

### Changes

- **New:** `DeflateNCDDistance.java` — NCD implementation using JDK `Deflater`
- **New:** `DeflateNCDBenchmark.java` — JMH micro-benchmark comparing Deflate NCD vs Vicino PPM across short/medium/long string pairs
- **Modified:** `controller.js` — Register `DeflateNCDDistance` for the `"ppm"` key instead of `VicinoDistance(PPMDistance())`
- **Modified:** `kNNClustererTests.java` — Updated to use `DeflateNCDDistance`, added direct NCD validation test, fixed column name constant

All 203 core module tests pass.

### Next Steps

This is a stepping stone toward fully removing the Vicino dependency. After this and the Levenshtein replacement in #7721, the remaining Vicino usages would be `VPTreeClusterer` and `NGramClusterer` (used internally by `kNNClusterer`), plus any distances registered by extensions.

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -19,7 +19,11 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>main</artifactId>
       <version>${project.version}</version>
-      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.openrefine.dependencies</groupId>
+      <artifactId>vicino</artifactId>
+      <version>${vicino.version}</version>
     </dependency>
     <dependency>
       <groupId>javax.servlet</groupId>


tfmorris:
> I decompiled and analyzed Vicino's implementation to fully understand what we're replacing

could have just looked at the source code! ;) 

Seriously though, let's hold off on doing more work in this space until I get the infrastructure adjusted to be more friendly to efficient implementations. I already have a ZStd based replacement for arithcode that we can compare to this when we've got the framework cleaned up.

I will re-review #7721 and get it merged, but let's wait on anything additional.

thadguidry:
When you say "infrastructure" or "framework", what do you mean?  A DB backed OpenRefine? (I can work on that).  Or caching for Clustering as you mentioned in [forum post](https://forum.openrefine.org/t/faster-clustering/2753/12?u=thadguidry) or something else?

tfmorris:
> When you say "infrastructure" or "framework", what do you mean? A DB backed OpenRefine?

Nothing as grand as that. Principally, kNNcluster with the stuff that I mentioned in the forum post, but it leaks into Engine, ConjunctiveRowFilter, a replacement for FilteredRows, the Clusterer interface, etc. I want to keep the scope limited to clustering, but also use it to think about implications (compatibility, performance, etc) in other parts of the system.

tfmorris:
I just noticed this branch is in the main project repo. Please don't do that. I'll look at fixing the permissions, but until then please consider it a social contract among the engineering team. Thanks.

thadguidry:
@tfmorris oh crap you're right!  I thought ya'll had master already protected, why not?  I'll update my workflows so it doesn't accidentally happen, but yes, please disallow those permissions I still might have.

PR #7729, opened 2026-03-24
Make clustering operations cancellable with progress feedback

### LINKED ISSUE #289: Clustering should be cancelable
_Original author: robin.be...@gmail.com (December 16, 2010 21:53:00)_

Steps to reproduce the problem:
1. Select a column with a large number of long and different values
2. Select &quot;cluster&quot; and &quot;nearest neighbor&quot;
3. Close the cluster dialog window without waiting for results

Expected behavior would be to shutdown the clustering job.
Instead, it continues running in background, consuming all the CPU.

Version: Google Refine 2.0 [r1836]
Operating system: Mac OS X

_Original issue: http://code.google.com/p/google-refine/issues/detail?id=289_

thadguidry:
## Summary

Clustering operations now run asynchronously as a `LongRunningProcess`, allowing users to cancel them and see real-time progress in the UI. This addresses a long-standing usability issue where large datasets could cause the clustering dialog to appear frozen with no way to abort.

Closes #289

## Backend changes

- **Clusterer base class**: Added `cancel()`, `isCanceled()`, and `getProgress()` with volatile fields for thread-safe access
- **ClusteringProcess**: New `LongRunningProcess` subclass that wraps a `Clusterer` + `Engine` for async execution via ProcessManager
- **ComputeClustersCommand**: Now returns `{"code":"pending"}` immediately and runs clustering in background. Tracks active processes per project in a `ConcurrentHashMap`
- **GetClusteringStatusCommand** (new GET): Returns status (`idle`, `pending`, `running`, `done`, `error`) with progress percentage and cluster results
- **CancelClusteringCommand** (new POST, CSRF-protected): Cancels the active clustering for a project
- **NGramBlockingClusterer** (new): Replaces Vicino's `NGramClusterer` with a native implementation using the `SimilarityDistance` interface and Union-Find based clustering with n-gram blocking
- **kNNClusterer**: Removed dead `VPTreeClusteringRowVisitor` and `DistanceWrapper` code; `BlockingClusteringRowVisitor` now uses `NGramBlockingClusterer`
- **BinningClusterer**: `visit()` now returns the cancel flag to abort row iteration early

## Frontend changes

- **Progress bar + Cancel button**: New panel in the clustering dialog shows real-time progress percentage with a smooth animated bar and a Cancel button
- **Async polling**: The dialog POSTs to start clustering, then polls `get-clustering-status` every 500ms for progress updates
- **Auto-cancel on close**: Dismissing the dialog automatically cancels any active clustering operation
- **i18n**: Added strings for progress message, cancelled state, and cancel button label

## Tests

- `NGramBlockingClustererTests`: 6 tests covering basic clustering, no lonely clusters, cancellation, progress callback, and tokenization
- `ClusteringProcessTests`: 2 tests covering run-to-completion and cancellation
- `ComputeClustersCommandTests`: Updated existing tests for the new async flow

## Notes

- The Vicino Maven dependency is retained for now since `VicinoDistance` (the adapter) is still used for distance metrics. Full Vicino removal is planned as a follow-up once the distance metric refactoring (#7721, #7728) merges.
- The `VPTreeClusterer` code path was already unreachable (no UI exposed it) and has been removed as dead code.


modules/core/src/main/java/com/google/refine/clustering/Clusterer.java
```diff
@@ -41,9 +41,28 @@ public abstract class Clusterer {
...
+
+    public boolean isCanceled() {
+        return _canceled;
+    }
+
+    public int getProgress() {
+        return _progress;
+    }
+
+    public String getStage() {
```
thadguidry, line 62: Note: the idea of a Clustering Stage was to break up the parts of:

1. Scanning rows - i.e., Filtering (which in the future might be provided by a different backend or database and hence incur a cost and might be useful for users to see this as "not yet" part of the real clustering - distance comparisons).  Currently, this stage "feels slow" and is apparent only when there are millions of rows to be filtered and given to populate the Clusterer of choice.
2. Finding clusters - i.e., calling getClusters() with distance comparisons

See in this PR: [kNNClusterer `computeClusters`](https://github.com/OpenRefine/OpenRefine/pull/7729/changes#diff-19f04e69b30a56612ee36ef1b0fd5f1cf605cef6b0e5f4f626caced80d647ec7L238)

There might be other stages perhaps depending on a Cluster implementation in the future, who knows, it might be allowed to call an AI Agent to perform the skill.
But for now, this is the 2 stages of "slowness" that I felt with kNN types and saw for myself when very large datasets are clustered.  Other clusterers might have different stages, for instance, an initial one to facilitate a Preview later on, and so this helps towards that.

The binning clusterer doesn't report progress at all and that's currently by design.  I didn't see the need yet.

PR #7771, opened 2026-05-06
fix(exporter): preserve leading/trailing whitespace in CSV and TSV export

### LINKED ISSUE #7770: Export to .csv trims whitespace
If a project has cells that contain a newline in the beginning, then "Export -> Comma-separated value" will trim these newlines in the exported .csv file. It might be a data structure like this, for example:

```
col_1,col_2
test_1," 
test_2"
```

### To Reproduce
1. Save the above example as .csv file and then create an OpenRefine project from it.
2. In OpenRefine, choose "Export -> Comma-separated value"
3. Open the exported .csv file

### Current Results

The exported .csv in the attached project will have lost the newline in "col_2".

### Expected Behavior

I would have expected the content of the fields to be exported as they are.

### Versions
 - Browser Version: Firefox 150.0.1
 - JRE or JDK Version: openjdk version "21.0.7" 2025-04-15 LTS 
 - OpenRefine: OpenRefine 3.10.0

### Additional context

This occurs with both OpenRefine 3.8.7 and OpenRefine 3.10.0.

This bug report was opened based on a suggestion from the forums: https://forum.openrefine.org/t/export-to-csv-without-trimming-whitespace/2777/3

I cannot seem to attach files to this issue, so I hope it is reproducible with the given example.

ShresthSamyak:
## Summary

Fixes #7770

- `CsvWriterSettings` and `TsvWriterSettings` in univocity-parsers 2.9.1 default to `ignoreLeadingWhitespaces = true` and `ignoreTrailingWhitespaces = true`, which silently strips all leading/trailing whitespace — including `\n`, `\r\n`, and spaces — from every cell value on write.
- Fix: disable both flags explicitly on both writer settings so cell values are written verbatim.
- Regression tests added to `CsvExporterTests` and `TsvExporterTests` covering a cell value with a leading newline (`\ntest_2`).

## Test plan

- [x] `CsvExporterTests#exportCsvWithLeadingNewline` — verifies leading `\n` is preserved and quoted in CSV output
- [x] `TsvExporterTests#exportTsvWithLeadingNewline` — verifies leading `\n` is preserved (TSV-escaped) in TSV output
- [x] All existing CSV/TSV exporter tests continue to pass (`Tests run: 16, Failures: 0, Errors: 0`)

To run locally:
```
mvn -pl main test -Dtest=CsvExporterTests,TsvExporterTests
```

### DIFF
--- a/main/src/com/google/refine/exporters/CsvExporter.java
+++ b/main/src/com/google/refine/exporters/CsvExporter.java
@@ -112,9 +112,14 @@ public class CsvExporter implements WriterExporter {
         AbstractWriter csvWriter;
         if ("\t".equals(separator)) {
             TsvWriterSettings tsvSettings = new TsvWriterSettings();
+            tsvSettings.setIgnoreLeadingWhitespaces(false);
+            tsvSettings.setIgnoreTrailingWhitespaces(false);
+            tsvSettings.getFormat().setLineSeparator(lineSeparator);
             csvWriter = new TsvWriter(writer, tsvSettings);
         } else {
             CsvWriterSettings settings = new CsvWriterSettings();
+            settings.setIgnoreLeadingWhitespaces(false);
+            settings.setIgnoreTrailingWhitespaces(false);
             settings.setQuoteAllFields(quoteAll); // CSV only
             settings.getFormat().setLineSeparator(lineSeparator);
             settings.getFormat().setDelimiter(separator);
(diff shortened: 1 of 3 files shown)


ShresthSamyak:
Good catch, thanks @Copilot. Fixed in the latest commit — `tsvSettings.getFormat().setLineSeparator(lineSeparator)` is now applied in the TSV branch, consistent with the CSV branch.

ShresthSamyak:
Done — both test methods updated to also set a trailing-newline cell value in row 2, so `ignoreTrailingWhitespaces(false)` is now exercised in both the CSV and TSV cases.

tfmorris (approved):
Looks great. Thanks very much!

PR #7779, opened 2026-05-15
Add strongly typed SaveOptions. Fixes #7523

### LINKED ISSUE #7523: Replace Properties with WriterOptions in Change, HistoryEntry, Cell, etc
Currently all of the save code for project saves including HistoryEntry, Change, etc uses an opaque Properties object to pass in additional parameters. In practice, there are only a couple of things saved there using "magic" keys:
- "mode" : "save" to control weather it's a lightweight or full save
- "pool" which contains a Pool object with all the pooled recon values

Switching to an explicit data class which contains these fields will allow us to more easily track what is used where.

tfmorris:
Fixes #7523

- Introduces a SaveOptions class to hold the save mode and the Pool of recon values and uses it everywhere instead of a raw Properties object with magic key values
- adds backward/forward compatibility default method implementations to the Change interface
- adds Claude 4.5 Haiku generated test cases for the legacy compatibility - CHECK this carefully!

NOTE: Many of the old methods were just removed rather than deprecated since they're part of the internal save machinery, with the exception being the Change interface which people might have implemented in extensions.


modules/core/src/main/java/com/google/refine/history/Change.java
```diff
@@ -50,5 +50,38 @@ public interface Change {
...
-    public void save(Writer writer, Properties options) throws IOException;
+    /**
+     * Legacy API for save.
+     * 
+     * @deprecated since 4.0. New implementations should override {@link #save(Writer, SaveOptions)}
+     */
+    @Deprecated(since = "4.0", forRemoval = true)
+    default void save(Writer writer, Properties options) throws IOException {
+        throw new UnsupportedOperationException("Calling old-style API for new Changes is not supported");
+    };
```
tfmorris:
```suggestion
    }
```

modules/core/src/main/java/com/google/refine/history/Change.java
```diff
@@ -50,5 +50,38 @@ public interface Change {
...
+        if (options != null) {
+            if (options.isSaveMode()) {
+                properties.put("mode", "save");
+            }
+            if (options.getPool() != null) {
+                properties.put("pool", options.getPool());
+            }
+        }
+        save(writer, properties);
+    };
```
tfmorris:
```suggestion
    }
```

modules/core/src/main/java/com/google/refine/model/ColumnModel.java
```diff
@@ -43,13 +43,15 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
...
-import java.util.Properties;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+// TODO: Should this class be defined somewhere higher up?
+import com.google.refine.history.SaveOptions;
+
```
tfmorris:
```suggestion
import com.google.refine.history.SaveOptions;

// TODO: Should this class be defined somewhere higher up?
```

modules/core/src/main/java/com/google/refine/model/ColumnModel.java
```diff
@@ -43,13 +43,15 @@ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
...
 import java.util.List;
 import java.util.Map;
-import java.util.Properties;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+// TODO: Should this class be defined somewhere higher up?
```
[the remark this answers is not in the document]
tfmorris: Comment removed and SaveOptions moved to com.google.refine.io

PR #7792, opened 2026-05-19
fix(wikibase): focus first input when statement-configuration dialog opens

### LINKED ISSUE #7609: A bug that occurs when creating the Wikidata schema
### To Reproduce
Hello, I would like to write to you about a bug that occurs when creating the Wikidata schema.
The first part is no problem: When I click on "add statement", a small "property" window pops up with the cursor inside it. This is okay.
But the second part is always a bit annoying. I type the value of "property", press enter, and another small window pops up. However, the cursor is nowhere to be found! I have to manually take the mouse and click on that little window.
So I would like to ask you to fix this so that this cursor appears automatically in every new small window.
Thank you in advance!
Maio

### Versions
 - Browser Version: Firefox 147.0
 - JRE or JDK Version: none
 - OpenRefine: OpenRefine 3.9.5

### Additional context

I'm a complete newbie here, so please don't be angry that I messed up this report a bit. Thanks.

eric-song-dev:
When the dialog opens, no element receives keyboard focus, so the user must click into the dialog before they can interact with it. Match the behavior of save-schema-dialog and perform-edits-dialog by focusing the first input (the mode select) at the end of launch().

Fixes #7609

### DIFF
--- a/extensions/wikibase/module/scripts/dialogs/statement-configuration-dialog.js
+++ b/extensions/wikibase/module/scripts/dialogs/statement-configuration-dialog.js
@@ -199,5 +199,8 @@ StatementConfigurationDialog.launch = function(statement) {
     }
   });
 
+  // Match the focus behavior of save-schema-dialog and perform-edits-dialog so
+  // the user can interact with the dialog immediately via the keyboard.
+  this._elmts.modeInput.focus();
 };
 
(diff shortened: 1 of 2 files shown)


tfmorris (changes requested):
Welcome to the project and thanks for the contribution! The Copilot review includes a few small suggestions, all of which seem correct. Could you please address them and let us know when this is ready for re-review? Thanks!

PR #7793, opened 2026-05-19
Enable middle-click for recon suggestion links

### LINKED ISSUE #6697: Recon suggest widget: enable middle-click to open URL in new tab without selecting

mohamad922005:
Fixes #6697


mohamad922005:
Fixes #6697

tfmorris:
Welcome to the project and thank you for the contribution. Could you please add a Cypress end-to-end test for your fix so we can make sure that it doesn't regress in the future?

tfmorris:
Closing due to lack of response to request for test.

PR #7797, opened 2026-05-25
test(browsing): cover TextSearchFacet filter modes

Karry1123:
Adds unit tests for the previously untested TextSearchFacet and establishes the browsing/facets/ test directory. 
  Covers the filter's orthogonal axes (plain vs. regex, case sensitivity, inversion, and input validity) using an in-memory project. Also documents edge-case behaviors like null cell handling, empty query short-circuiting, and early exception throwing for invalid regexes.


tfmorris:
Same comment as the other PR, but let's focus on a single PR at a time, so I'm going to close this one for now.

PR #7798, opened 2026-05-25
test(expr): cover JsonValueConverter numeric and container edge cases

Karry1123:
Adds unit tests for the JsonValueConverter utility. Closes coverage gaps for null JsonNode references, BigInteger and BigDecimal coercion branches, and the full primitive surface of the convertComparable method. 
  The suite validates convertComparable behavior for both JSON and Java nulls, and pins down the IllegalArgumentException paths for complex container types (objects and arrays). A reader can assess the coercion contract and exception handling just by scanning the test names.


tfmorris:
Same comment as the other PR, but let's focus on a single PR at a time, so I'm going to close this one for now.

PR #7807, opened 2026-05-31
Fix #7330: Make stop words editable via preferences

### LINKED ISSUE #7330: Improve stop word handling - customizable and/or multi-language

DeepakJaiswar25:
Fixes #7330

- Added a default stopwords preference with the value: the,a,and,of,on,in,at,by
- Updated StandardReconConfig to load stop words from preferences instead of a hardcoded set.
- Trims whitespace around configured stop words.
- The default stop word list is used when no preference is configured.
- Added Unit tests for for custom stop words, whitespace trimming and default stop word list is used when no preference is configured.

### DIFF
--- a/modules/core/src/main/java/com/google/refine/model/recon/StandardReconConfig.java
+++ b/modules/core/src/main/java/com/google/refine/model/recon/StandardReconConfig.java
@@ -755,16 +755,16 @@ public class StandardReconConfig extends ReconConfig {
 
     static protected Set<String> getStopWords() {
 
-            Object prefValue = (ProjectManager.singleton.getPreferenceStore().get("stopwords"));
-            String stopwordsPref = (prefValue != null) ? prefValue.toString() : null;
-            if(stopwordsPref == null) {
-                stopwordsPref = "the,a,and,of,on,in,at,by";
-            }
-            Set<String> stopWords= new HashSet<>();
-            for(String stopWord: stopwordsPref.split(",")) {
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/com/google/refine/model/recon/StandardReconConfig.java
```diff
@@ -739,26 +740,41 @@ static protected double wordDistance(Set<String> longWords, Set<String> shortWor
...
+//        s_stopWords.add("the");
+//        s_stopWords.add("a");
+//        s_stopWords.add("and");
+//        s_stopWords.add("of");
+//        s_stopWords.add("on");
+//        s_stopWords.add("in");
+//        s_stopWords.add("at");
+//        s_stopWords.add("by");
+//    }
+
```
tfmorris:
```suggestion
```

tfmorris (changes requested):
Thanks for the contribution and sorry for the delay in getting it reviewed. I'm not sure why the CI pipeline didn't run. 

We don't want dead (ie commented out) code, so I'll push a change to that to see if it triggers the testins.

tfmorris (approved):
Thanks for the contribution! I went ahead and swapped in Copilot's suggested more efficient implementation. Apologies again for the delay in getting this reviewed.

PR #7811, opened 2026-06-05
Add confirmation dialog before removing facets

mohamad922005:
## What

Adds a confirmation dialog before removing a facet.

## Why

Users can accidentally remove a facet with a single click. This change adds a confirmation step to prevent accidental removal.

## Testing

- Added Cypress test coverage.
- Verified Cancel keeps the facet.
- Verified OK removes the facet.

### DIFF
--- a/main/tests/cypress/cypress/e2e/project/grid/column/facet/facets.cy.js
+++ b/main/tests/cypress/cypress/e2e/project/grid/column/facet/facets.cy.js
@@ -109,12 +109,36 @@ describe(__filename, function () {
     cy.getFacetContainer('Water').contains('Cluster');
   });
 
-  it('Delete a facet', function () {
+  it('Delete a facet with confirmation dialog', function () {
+
     cy.loadAndVisitProject('food.small');
+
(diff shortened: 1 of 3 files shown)


magdmartin:
@mohamad922005 thanks for your interest in OpenRefine. Is your change related to an existing issue?  If so, can you link it? 

This introduces a significant change to the User Experience, and I think we should seek first consensus within the community before adding a new dialogue.

mohamad922005:
Hello,
Thanks for your time. No, it is not linked to an existing issue but I designed and implemented this feature which I think it is essential.

Best regards,
Mouhamed


________________________________

Abbe98:
Is there a particular use case or need for this? Removing a facet is not a destructive action, they should always be reproducible.

mohamad922005:
Thank you for the feedback. That makes sense. I originally thought the confirmation could help prevent accidental clicks, but I understand that removing a facet is not destructive and can be reproduced easily.

________________________________

tfmorris:
Thanks for your contribution. Based on the discussion, I'm going to close this, but please feel free to pick another issue to work on.

PR #7818, opened 2026-06-07
feat(exporter): add Markdown table export

eric-song-dev:

- Add a new `MarkdownExporter` for exporting the current project view as a GitHub-flavored Markdown table.
- Register the `md` export format in the exporter registry and add it to the Export menu as `Markdown table`.
- Add unit tests covering normal table output, empty cells, escaped pipe characters, and newline handling inside cells.

Tests run:

- `mvn -pl main -Dtest=MarkdownExporterTests test`

### DIFF
--- a/main/src/com/google/refine/exporters/MarkdownExporter.java
+++ b/main/src/com/google/refine/exporters/MarkdownExporter.java
@@ -1,0 +1,105 @@
+/*
+
+Copyright 2010, Google Inc.
+Copyright 2024, OpenRefine contributors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+    * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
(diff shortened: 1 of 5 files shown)


magdmartin:
@eric-song-dev  thanks for your interest in OpenRefine. Is your change related to an existing issue?  If so, can you link it? 

I am curious regarding the use case for adding a markdown table export.

main/src/com/google/refine/exporters/MarkdownExporter.java
```diff
@@ -0,0 +1,105 @@
...
+                    writer.write("\n");
+
+                    if (isHeader) {
+                        writer.write("|");
+                        for (int i = 0; i < cells.size(); i++) {
+                            writer.write(" --- |");
+                        }
+                        writer.write("\n");
+                    }
+                } catch (IOException e) {
```
durkagiri21-ship-it, line 88: Should this IOException be logged instead of silently ignored? Swallowing exceptions here could make it hard to debug export failures in production. Might be worth adding at least a logger.error() call.

tfmorris:
Thanks for the contribution, but closing since this doesn't address an identified need in an open, triaged issue. Unfortunately, we don't have the bandwidth to review PRs which don't address pressing issues. 

Please look for an issue which has been tagged with the "Good First Issue" tag or which looks easy to solve for you. Thanks!

PR #7819, opened 2026-06-11
fix(wikibase): do not show required statement warnings for existing items

### LINKED ISSUE #6767: wikibase: Don't show "required statement" constraints for edits on existing entities
When uploading information to existing Wikibase items, we can get "issues" such as:

> **[male population (P1540)](http://www.wikidata.org/entity/P1540) requires statement with property [P17](http://www.wikidata.org/entity/P17).**
> This property is expected to have another statement with property [P17](http://www.wikidata.org/entity/P17) such as on item [Katowice (Q588)](http://www.wikidata.org/entity/Q588). If the item already has statements with property [P17](http://www.wikidata.org/entity/P17) in Wikidata, then this warning can be ignored.

During [an OpenRefine-Wikidata training at Wikimania](https://wikimania.eventyay.com/2024/talk/PBMMEJ/), this confused the trainer and audience into thinking that the "male population" property was only applicable to countries, whereas the meaning of this constraint is just to say that if "male population" is present on an item, the country of that item should also be added as a statement.

This issue could actually be safely ignored in this case because the corresponding items already had such a statement (which OpenRefine didn't check, as the issue message indicates).

### Proposed solution

Such issues should simply not be generated for edits on existing items, because it's confusing.

### Alternatives considered

The issue could be generated only in "slow mode" (when showing issues in the upload dialog), after checking that the corresponding statements are indeed missing.

### Additional context

Follow-up to the training session mentioned above.

whitewooood:
Fixes #6767

- Stop generating Wikibase required statement constraint warnings for edits on existing entities.
- Keep required statement checks for new entities.
- Update unit and preview-command tests to reflect the new behavior.

Validation:
- RED: `mvn -pl extensions/wikibase -Dtest=ItemRequiresScrutinizerTest test` failed because an existing entity still produced `existing-item-requires-property-to-have-certain-values-with-suggested-value`.
- GREEN: `mvn -pl extensions/wikibase -Dtest=ItemRequiresScrutinizerTest test` passed.
- `mvn -pl extensions/wikibase test` passed: 575 tests, 0 failures.
- `git diff --check` passed.

### DIFF
--- a/.github/workflows/codeql_java.yml
+++ b/.github/workflows/codeql_java.yml
@@ -24,6 +24,13 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
 
+      - name: Set up Java 21
+        uses: actions/setup-java@v5
+        with:
+          distribution: 'temurin'
+          java-version: 21
+          cache: 'maven'
+
       - name: Initialize CodeQL
         uses: github/codeql-action/init@v4
         with:
--- a/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/ItemRequiresScrutinizer.java
+++ b/extensions/wikibase/src/org/openrefine/wikibase/qa/scrutinizers/ItemRequiresScrutinizer.java
@@ -80,6 +80,10 @@ public class ItemRequiresScrutinizer extends EditScrutinizer {
     }
 
     public void scrutinizeStatementEdit(StatementEntityEdit update) {
+        if (!update.isNew()) {
+            return;
+        }
+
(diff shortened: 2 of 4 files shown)


tfmorris:
Thanks for the contribution. Does this address an open issue? If so, could
you please fill in the PR template with the issue? Thanks!

whitewooood:
Yes, this addresses #6767. I updated the PR description to use the template and include the issue link. Thanks for pointing it out.

tfmorris (approved):

I added a commit to configure the Java compiler for the CodeQL job so that CI would run, but that was independent of this fix.

PR #7850, opened 2026-07-16
Restore Cypress lint checks

### LINKED ISSUE #4211: Restore Es-Lint & prettier checks for CI/CD
During our efforts to run cypress tests in parallel, I've removed the steps for ES-Lint & Prettier.
This needs to be restored, and the source for Cypress tests probably needs to be re-prettified, as some invalid formatting has probably been introduced since.

@kushthedude apologies ! I removed that by mistake when I refactored the CI/CD for parallelism

aouxwoux:
## What changed

- migrated the Cypress suite from the deprecated eslintrc format to an ESLint 10 flat config
- restored Prettier and ESLint validation in the E2E pull-request workflow
- reformatted the Cypress suite and resolved the lint errors exposed by re-enabling the checks
- removed an unused Cypress import helper whose undefined variables were exposed by linting

## Why

The existing `lint` script was not run in CI, and ESLint 10 no longer loads `.eslintrc.json`. As a result, formatting and lint regressions had accumulated in the Cypress suite.

The new CI job skips downloading the Cypress binary because formatting and ESLint do not require it.

## Validation

- `yarn lint` — passes with 0 errors (4 existing `cypress/no-force` warnings)
- `git diff --check`

Fixes #4211


tfmorris:
@aouxwoux Thanks for the contribution! If you'd like to leave a comment on the linked issue, I can assign you there (that's a requirement of the way Github works).

We've certainly accumulated a lot of technical debt in this area, so it'll take me a little while to wade through it all, but looks good so far.

tfmorris (approved):
Looks great! Thanks very much for your contribution.

I've created https://github.com/OpenRefine/OpenRefine/issues/7857 as a follow-up to this.

PR #7853, opened 2026-07-20
Enable Unicode character classes in GREL regex and text filter

### LINKED ISSUE #1768: Unicode support for regex
Hello,

As a user, I want to deal with regex facets including accent selection into words. It is a solution to apply the regex modifier 'u' that would simplify my life. Strangely, it is possible to bypass this limitation using jython into columns creation, but it remains difficult because you have to code each transformation rather using a simple regex modifier.

As a solution, I wonder if the default modifier u could be used on project based on UTF files... For the moment, very annoying...

Best regards,

arimu1:
## Summary

Fixes https://github.com/OpenRefine/OpenRefine/issues/1768

Java's default regex `\w` / `\d` / `\s` are ASCII-only unless `Pattern.UNICODE_CHARACTER_CLASS` is set. GREL regex literals, `match()` with a string pattern, and the text-search facet regex mode now compile with that flag (and `UNICODE_CASE` whenever case-insensitive matching is requested), matching the approach already used in `MultiValuedCellSplitOperation`.

## Test plan

- [x] `mvn -pl modules/grel -am test -Dtest='MatchTests,GrelTests'` — Unicode `\w`/`\d`/`\s` / case-fold coverage
- [x] `mvn -pl main -am test -Dtest='TextSearchFacetTests'` — Unicode regex facet regression

Made with [Cursor](https://cursor.com)

### DIFF
--- a/main/tests/server/src/com/google/refine/browsing/facets/TextSearchFacetTests.java
+++ b/main/tests/server/src/com/google/refine/browsing/facets/TextSearchFacetTests.java
@@ -186,6 +186,36 @@ public class TextSearchFacetTests extends RefineTest {
     }
 
     @Test
+    public void testUnicodeRegExFilter() throws Exception {
+        // Regression test for #1768: regex text filter must use Unicode-aware
+        // predefined character classes. \\w must match accented letters.
+        project = createProject("TextSearchFacetUnicode",
+                new String[] { "Value" },
+                new Serializable[][] {
+                        { "café" },
+                        { " Espresso" },
+                        { "123" },
+                        { "Ångström" }
+                });
+
+        // \w should match accented Unicode letters.
+        String filter = "{\"type\":\"text\","
+                + "\"name\":\"Value\","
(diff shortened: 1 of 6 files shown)


tfmorris:
Hi @arimu1. Are you still working on this? If so, could you please address the review feedback from Copilot? Thanks!

arimu1:
Addressed the Copilot feedback:

1. **`Match.java`**: string-form regex now compiles with `Pattern.UNICODE_CHARACTER_CLASS | Pattern.UNICODE_CASE` so `(?i)` uses Unicode case folding.
2. **`MatchTests`**: added `matchFunctionUnicodeCaseInsensitiveTest` asserting `match("École", "(?i)école")` matches.

Verified locally: `mvn -pl modules/grel -am test -Dtest=com.google.refine.expr.functions.strings.MatchTests -Dsurefire.failIfNoSpecifiedTests=false` — 7 tests, 0 failures.

Issue #7474, opened 2025-09-30
Warn user about missing extensions on project import


tfmorris:
Importing a project had an non-standard extension used in its creation doesn't generate any warnings for the user, but will fail when they attempt to Undo/Redo operations which are part of the extension

### To Reproduce
1. First, do ...
2. Then, ...
3. Finally, ...

### Current Results

```
06:21:49.339 [                   refine] POST /command/core/undo-redo (20895ms)
06:21:49.375 [                  command] Exception caught (35ms)
java.lang.RuntimeException: Failed to load change file /Users/tfmorris/tmp/refinex/1900787754295.project/history/1758910964174.change.zip
	at com.google.refine.io.FileHistoryEntryManager.loadChange(FileHistoryEntryManager.java:85)

Caused by: java.lang.ClassNotFoundException: org.freeyourmetadata.ner.operations.NERChange
...
	at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:511)
...
        History.getChangeClass()
```

### Expected Behavior

We should:
1. Warn the user when the project is imported (and refuse to import it?)
2. Generate a useful error when they attempt to undo the operation and
3. Not corrupt the operation history

### Versions
 - Operating System: 
 - Browser Version: 
 - JRE or JDK Version: 
 - OpenRefine:

PR #7855, opened 2026-07-21
Warn user about missing extensions on project import

arimu1:
## What

Addresses #7474

When a project's history references a `Change` class registered by an extension that isn't installed (e.g. a project created with `freeyourmetadata-ner` installed, then reopened without it), OpenRefine currently gives no indication of the problem at import time. The failure only surfaces later, as a raw `ClassNotFoundException` wrapped in a `RuntimeException`, when the user tries to undo/redo that particular history entry:

```
java.lang.RuntimeException: Failed to load change file .../history/1758910964174.change.zip
	at com.google.refine.io.FileHistoryEntryManager.loadChange(FileHistoryEntryManager.java:85)
Caused by: java.lang.ClassNotFoundException: org.freeyourmetadata.ner.operations.NERChange
```

This PR detects the problem at import time instead. `FileHistoryEntryManager.findMissingChangeClasses(File)` scans a project's `history/*.change.zip` files for the class name(s) recorded in each `change.txt` and returns the list of classes that can't be resolved on the current classpath. This includes classes nested inside a `MassChange` (as produced by "Apply operations", column reordering, and other batch operations bundling several sub-changes into one file) — not just top-level ones. Resolving a class this way does invoke normal JVM class initialization (1-arg `Class.forName`, the same path `History.getChangeClass` already uses for undo/redo), so an installed extension's static initializers do run at import time; it does not construct, apply, or revert the change itself.

`ImportProjectCommand` calls this right after a project import completes. If any classes are missing, it logs a warning server-side and appends the list to the redirect URL. The project page (`project.js`) reads that parameter on load and shows the user an alert naming the missing classes, so they know some operations may not be undoable/redoable until the relevant extension is installed.

## Scope

This addresses the "warn the user on import" part of #7474 — which is also what the issue is titled. The other two asks in that issue are **not** implemented here and are left open:

- Making undo/redo produce a nicer error message instead of the raw `ClassNotFoundException` / `RuntimeException`.
- Deciding whether import should be refused outright when classes are missing (the issue itself is unsure about this: "and refuse to import it?").

I've kept this PR to the warning-on-import behavior to keep it focused and reviewable; a maintainer may want separate issues/PRs for the other two.

## Testing

- `./refine lint` — BUILD SUCCESS
- `FileHistoryEntryManagerTests` (`testFindMissingChangeClasses`, `testFindMissingChangeClassesNoHistory`, `testFindMissingChangeClassesInsideMassChange`, `testWriteHistoryEntry`) — 4/4 pass
- `ImportProjectCommandTests` — 1/1 pass
- `mvn -pl modules/core,main -am test-compile` — compiles cleanly

### DIFF
--- a/main/src/com/google/refine/commands/project/ImportProjectCommand.java
+++ b/main/src/com/google/refine/commands/project/ImportProjectCommand.java
@@ -37,6 +37,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.net.URLConnection;
+import java.net.URLEncoder;
+import java.util.List;
 import java.util.Map;
 
 import javax.servlet.ServletException;
@@ -53,6 +55,8 @@ import org.slf4j.LoggerFactory;
 import com.google.refine.ProjectManager;
 import com.google.refine.ProjectMetadata;
 import com.google.refine.commands.Command;
+import com.google.refine.io.FileHistoryEntryManager;
+import com.google.refine.io.FileProjectManager;
 import com.google.refine.model.Project;
 import com.google.refine.util.ParsingUtilities;
 
@@ -88,7 +92,20 @@ public class ImportProjectCommand extends Command {
                     }
                 }
 
-                redirect(response, request.getContextPath() + "/project?project=" + projectID);
+                StringBuilder redirectUrl = new StringBuilder(request.getContextPath() + "/project?project=" + projectID);
+
+                List<String> missingClasses = FileHistoryEntryManager.findMissingChangeClasses(
+                        ((FileProjectManager) ProjectManager.singleton).getProjectDir(projectID));
+                if (!missingClasses.isEmpty()) {
+                    logger.warn(
+                            "Imported project {} depends on classes which are not available on this instance of OpenRefine "
+                                    + "(the project might have been created with extensions which are not currently installed): {}",
+                            projectID, missingClasses);
+                    redirectUrl.append("&missing-extension-classes=")
+                            .append(URLEncoder.encode(String.join(",", missingClasses), "UTF-8"));
+                }
+
+                redirect(response, redirectUrl.toString());
             } else {
                 respondWithErrorPage(request, response, "Failed to import project. Reason unknown.", null);
             }
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -477,6 +477,7 @@
     "core-project/confirm-erasure-of-project-history": "Confirm deletion of project history",
     "core-project/applying-change-erases-entries": "Applying this change will overwrite the following {{plural:$1|history entry|$1 history entries}}:",
     "core-project/do-not-warn": "Do not warn about overwriting history entries in the future",
+    "core-project/warning-missing-extensions": "This project was created with extensions that are not currently installed, so some of its operations may not be undoable or redoable. Missing classes: $1",
     "core-recon/add-std-srv": "Add standard service",
     "core-recon/service-documentation":"Service documentation",
     "core-recon/cell-type": "Reconcile each cell to an entity of one of these types:",
--- a/main/webapp/modules/core/scripts/project.js
+++ b/main/webapp/modules/core/scripts/project.js
@@ -632,6 +632,10 @@ function onLoad() {
       }
     }
 
+    if ("missing-extension-classes" in params) {
+      alert($.i18n('core-project/warning-missing-extensions', decodeURIComponent(params["missing-extension-classes"])));
+    }
+
     Refine.reinitializeProjectData(
       function() {
         initializeUI(uiState);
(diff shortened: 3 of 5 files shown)


arimu1:
Pushed 7a161a9b1 addressing review feedback:

1. **Nested `MassChange` sub-changes weren't scanned (blocking).** `readChangeClassName` only ever looked at line 2 of `change.txt`, so an extension `Change` bundled inside a core `MassChange` (as produced by "Apply operations", column reordering, etc. — see `MassChange.save`/`History.writeOneChange`) went completely undetected. `findMissingChangeClasses` now walks the recorded `VERSION`/className pairs recursively, skipping past each resolvable sub-change's own content via the same reflective `load()` call `History.readOneChange` makes, so scanning can continue to the next sibling. If a nested class can't be resolved, the reader position can no longer be trusted for what follows it in that list, so scanning stops there — documented as a comment at that point (`FileHistoryEntryManager.java`). Added `testFindMissingChangeClassesInsideMassChange`, which builds the exact on-disk envelope `MassChange#save` produces.
2. PR description changed from "Fixes #7474" to "Addresses #7474" and now explicitly lists which of the issue's three asks remain open (nicer undo error, refuse-to-import), since this PR only implements the "warn on import" one.
3. `project.js:636` — removed the redundant `decodeURIComponent` on `missing-extension-classes`; `URLUtil.getParameters()` already decodes it, and the extra call could throw `URIError` on a stray `%`.
4. `FileHistoryEntryManager.readChangeClassNames` (formerly `readChangeClassName`) now logs a warning instead of silently swallowing `IOException` for an unreadable/corrupt change file.
5. Reworded the "without invoking any extension code" claim in the PR description — `isClassAvailable` resolves classes via `RefineServlet.getClass`, which uses 1-arg `Class.forName`, so an installed extension's static initializers do run at import time (harmless, but the original wording overstated it).

Local verification (JDK 21, `mvn`/`./refine lint`):
- `./refine lint` — BUILD SUCCESS
- `FileHistoryEntryManagerTests` — 4/4 pass
- `ImportProjectCommandTests` — 1/1 pass
- `mvn -pl modules/core,main -am test-compile` — clean

arimu1:
Pushed `15208d98c` addressing the review feedback:

1. **`pool.load(...)` is now lazy** — it's only ever consumed inside the `MassChange` recursion (`readMassChangeClassNames` → `readNestedChangeClassName` → `skipChangeContent`), so it's now loaded only when a change file's top-level class is `MassChange`, instead of unconditionally for every change file. Verified there's no other use of `pool` on the plain-change path.
2. The pool's `InputStreamReader` is now in try-with-resources (extracted into a small `loadPool(ZipFile)` helper), matching the `change.txt` reader.
3. Dropped the dead `isClassAvailable(className)` check on the `MassChange` branch — `MassChange` is a core class reached via a class literal, so it's always available; the check could never be false.
4. Left `skipChangeContent`'s `catch (Exception e)` as-is rather than widening it to `Error`. If the reflective `load()` call OOMs, that means the change's actual content is too large to load at all — the same thing would happen for real the next time this history entry is undone/redone. Swallowing it here would just hide a genuine resource problem behind a best-effort import-time warning; letting it propagate surfaces it plainly instead. Documented this reasoning inline.
5. Added `testFindMissingChangeClassesStopsAfterFirstUnresolvable` — the existing `testFindMissingChangeClassesInsideMassChange` always put the unresolvable class *last*, so the documented "scanning stops after an unresolvable nested class" limitation was never actually exercised. The new test puts an unavailable class before another unavailable sibling and asserts only the first is reported.

**One caveat worth weighing:** `HistoryEntry._change` is `transient` and lazily loaded, so under normal operation these nested `MassChange` sub-changes are never deserialized until an undo/redo actually needs them. This scan is genuinely new work proportional to a project's total `MassChange` content — it only runs once, at import, and only for files whose top-level class is `MassChange`, but on a long history over a large dataset it's a real one-time import-time cost. Flagging this so it can be weighed against the value of the warning.

All local checks green:
- `./refine lint` — `BUILD SUCCESS`
- `FileHistoryEntryManagerTests` — 5/5 (`Tests run: 5, Failures: 0, Errors: 0, Skipped: 0`)
- `ImportProjectCommandTests` — 1/1 (`Tests run: 1, Failures: 0, Errors: 0, Skipped: 0`)

(Aside: the `linux_server_tests (25)` CI failure is a `jacoco-maven-plugin:merge` `EOFException` on a truncated `OpenRefine - Jython extension.exec` coverage file — unrelated to this PR, all suites report 0 failures/0 errors, Windows is green.)

tfmorris (commented):
Thanks for tackling this! It's a non-trivial amount of work. Your caveat is an important one, so thank you for making it explicit. 

It's been a while since I've thought about this, so I'll need to refresh my memory, but I think I was thinking more of saving a catalog of extensions used in the project history, which could be quickly checked when the project is opened. In addition to being heavyweight, scanning the entire undo history seems like it could provoke additional problems (but that's just a gut feeling).

It'll probably take me a few days to review this thoroughly, but wanted to give you some quick feedback.

PR #7863, opened 2026-07-22
fix: delete webapp/WEB-INF/classes on clean (#6745)

### LINKED ISSUE #6745: Stale i18n resource file when building/running from IntelliJ
There is a stale properties file at `main/webapp/WEB-INF/classes/com/google/refine/operations/OperationDescription.properties` which IntelliJ is using for runs from the IDE and none of the IntelliJ Maven incantations seem to update it, including "Reload project" and "Generate sources and update folders". It also doesn't get cleaned up by `./refine clean`, so the only way to force it to be regenerated is to delete it by hand.

      78 ./main/target/classes/com/google/refine/operations/OperationDescription.properties - Maven build
      78 ./main/resources/com/google/refine/operations/OperationDescription.properties - source file
      77 ./main/webapp/WEB-INF/classes/com/google/refine/operations/OperationDescription.properties - IntelliJ build

I suspect there's a mismatch between the Maven phase that we're using and what IntelliJ expects, but need to dig into it more.

arimu1:
Fixes #6745

- Configure `maven-clean-plugin` in `main/pom.xml` to include `webapp/WEB-INF/classes` in the filesets to be deleted on `mvn clean`.
- Ensures stale property files such as `OperationDescription.properties` output to `main/webapp/WEB-INF/classes` during IDE builds (e.g., IntelliJ) are properly cleaned up when executing `./refine clean` / `mvn clean`.

### DIFF
--- a/main/pom.xml
+++ b/main/pom.xml
@@ -124,6 +124,9 @@
             <fileset>
               <directory>webapp/WEB-INF/lib</directory>
             </fileset>
+            <fileset>
+              <directory>webapp/WEB-INF/classes</directory>
+            </fileset>
           </filesets>
         </configuration>
       </plugin>


tfmorris (approved):
Looks good. Thanks very much for the contribution!

PR #7884, opened 2026-08-01
Make reconciliation stop words configurable via preferences (fixes #<…

### LINKED ISSUE #7330: Improve stop word handling - customizable and/or multi-language

efe-khrmn:
Fixes #7330

- Added a "stopwords" preference (default: "the,a,and,of,on,in,at,by") so the reconciliation stop word list is no longer hardcoded and English-only
- Modified StandardReconConfig to read this preference, split on commas, trim whitespace, and lowercase each entry before building the stop word set
- Added unit tests in StandardReconConfigTests covering default, custom, case-insensitive, and empty preference scenarios


tfmorris:
@efe-khrmn Welcome to the project! Thanks for the contribution, but we already have a pending PR #7807 which addresses #7330.  Feel free to select a different issue to work on.

PR #7895, opened 2026-08-06
Wikibase: Clean up unused constants and dead code in ItemRequiresScrutinizer

### LINKED ISSUE #7892: Wikibase : Don't show "required statement" constraints for edits on existing entities
When editing existing Wikibase entities, OpenRefine displays false positive QA scrutinizer warnings (such as "required statement" constraints from `ItemRequiresScrutinizer`). Because OpenRefine's local schema preview does not query online entity statements, it assumes required properties are missing on the target entity, confusing users during schema building and upload preview.

### To Reproduce
1. Open an OpenRefine project with cells reconciled to existing Wikibase/Wikidata entities (e.g., Katowice `Q588`).
2. Go to **Extensions** > **Wikibase** > **Edit Wikibase schema**.
3. Add a statement to an existing entity using a property that has an "item requires statement" constraint (e.g., `male population (P1540)`, which requires `country (P17)`).
4. Click on the **Issues** tab in the schema preview dialog.

### Current Results
OpenRefine displays a Quality Assurance warning:
**male population (P1540) requires statement with property P17.**
 *This property is expected to have another statement with property P17 such as on item Katowice (Q588). If the item already has statements with property P17 in Wikidata, then this warning can be ignored.*

### Expected Behavior
OpenRefine should not generate "required statement" constraint warnings for edits on existing entities, as local schema preview cannot check existing online statements and the resulting warning is a false positive that misleads users.

### Versions
-Operating System: Windows 11 Version 25H2 (Build 26200.8894)
-Browser Version: 151.0.7922.71 (Official Build) (64-bit)
-JRE or JDK Version: JDK 21+
-OpenRefine: 3.10-SNAPSHOT (master)

### Additional context
Reported after an OpenRefine-Wikidata training session at Wikimania, where trainers and participants were misled by false constraint warnings when editing existing Wikidata items.

Devansh-18155:
Suppress "required statement" constraint warnings for edits on existing entities.

Fixes #7892

- Skip `ItemRequiresScrutinizer` checks when editing existing entities (`!update.isNew()`) to avoid false-positive QA warnings during schema preview.
- Remove redundant and unused existing-item warning type constants from `ItemRequiresScrutinizer.java`.
- Clean up ternary checks for QA warning creation and add descriptive test documentation in `ItemRequiresScrutinizerTest.java`.


Devansh-18155:
hi @tfmorris and @magdmartin Please take a review on this pr. Thanks!

extensions/wikibase/tests/src/org/openrefine/wikibase/qa/scrutinizers/ItemRequiresScrutinizerTest.java
```diff
@@ -41,6 +41,9 @@ public EditScrutinizer getScrutinizer() {
         return new ItemRequiresScrutinizer();
     }
 
+    /**
+     * Verifies that edits on existing entities do not raise required statement constraint warnings.
+     */
     @Test
     public void testExistingItemTrigger() {
```
tfmorris: It doesn't look like this test actually changed. I'd expect a new test that fails when the changes in the ItemRequesScrutinizer have not been made.

Devansh-18155:
@tfmorris , can you please take a review in this pr i have made the above changes by  updated `ItemRequiresScrutinizerTest.java` to rename the test method to `testExistingItemDoesNotTriggerWarning()` and added documentation clarifying the behavior. 

The test suite now explicitly asserts that existing entity edits (`TestingData.existingId`) do not raise required statement constraint warnings (`assertNoWarningRaised()`) whereas new item edits (`TestingData.newIdA`) in `testNewItemTrigger()` continue to raise warnings (`assertWarningsRaised(...)`).

tfmorris:
Thanks for the updates, but my question wasn't really about the name or the comment, but the fact that there was no new (or changed) test.

If the Info level messages are incorrect, we should have a test that fails if there's a regression in the future and the info messages return. Does that make sense? The ideal test would be one which fails without your change and passes with your change.

Devansh-18155:
Thanks for clarifying, @tfmorris!

I realized my PR title may have caused confusion: the functional suppression `(`if (!update.isNew()) return;`)` was actually merged previously in commit `b26e0550`. 

This PR is a cleanup of the dead code left behind from that commit specifically removing the unreachable ternary branches `(`update.isNew() ? ... : ...`)`and the unused `existingItemRequire...` constants.

The regression tests (`testExistingItemDoesNotTriggerWarning` and `testWrongValue`) already exist and assert `assertNoWarningRaised()`. If any warnings were re-introduced for existing entities, those tests would catch the regression and fail.

I have updated the PR title to reflect that this is a dead code cleanup.

tfmorris (approved):
OK, that makes sense. Sorry for being slow to understand.

PR #7896, opened 2026-08-06
Localize the Actions column header in the preferences table

abhijitdalal26:
This is part of the remaining work on #1858: some UI text still bypasses the i18n library and stays hardcoded in English regardless of the user's locale.

In the preferences table (`main/webapp/modules/core/scripts/preferences.js`), the "Key" and "Value" column headers go through `$.i18n`, but the "Actions" header is a plain English string literal. That means it never gets translated even when the rest of the page is localized.

Added a `core-index/actions` key to `translation-en.json` and used it in `preferences.js` for the third header, following the same pattern already used for the other two.

## Testing

This is a small frontend-only change to a static JS file. I verified `translation-en.json` still parses as valid JSON and checked the diff is limited to this one header. I didn't build the full Java backend for this, since the change doesn't touch any server-side code, but I'm happy to run a fuller build if a maintainer wants me to.

### DIFF
--- a/main/webapp/modules/core/langs/translation-en.json
+++ b/main/webapp/modules/core/langs/translation-en.json
@@ -21,6 +21,7 @@
     "core-index/preferences": "Preferences",
     "core-index/key": "Key",
     "core-index/value": "Value",
+    "core-index/actions": "Actions",
     "core-index/add-pref": "Add preference",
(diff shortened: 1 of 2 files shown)


tfmorris (approved):
Thanks very much for the contribution!

PR #7904, opened 2026-08-11
Fixes #6852: format integral numbers as floating point

### LINKED ISSUE #6852: toString() doesn't cast int to float
I had a column in my projects with numerical values (imported as strings) that I wanted to serialise into a standard format. I used value.toNumber().toString("%.4f") and found that this mostly worked but it fell over with an error when the source value was represented as an integer. So it seems passing an integer into the format "%f" gives an error:

```
toString(1,"%f)
Error: java.util.IllegalFormatConversionException: f != java.lang.Long

```
The workaround is to force a cast to float by multiplying the value by 1.0, but I think this feels counter-intuitive. Given that OR only gives a single function toNumber() (not toFloat() and toInteger() ), I think it's appropriate for the toString function to apply whatever type casting is required for the format.

OpenRefine version 3.8.2

kumarcodes:
Fixes #6852.

### Summary
- Retry formatting integral values as double when the requested Java format rejects the integral type.
- Add regression coverage for %.4f and %2.2f applied to integral values.

### Testing
- Not run locally: compiling the project dependencies exceeded the available command time limit.

### DIFF
--- a/modules/grel/src/main/java/com/google/refine/expr/functions/ToString.java
+++ b/modules/grel/src/main/java/com/google/refine/expr/functions/ToString.java
@@ -60,6 +60,13 @@ public class ToString implements Function {
                     try {
                         return String.format((String) o2, o1);
                     } catch (IllegalFormatException e) {
+                        if (o1 instanceof Long || o1 instanceof Integer || o1 instanceof Short || o1 instanceof Byte) {
+                            try {
+                                return String.format((String) o2, ((Number) o1).doubleValue());
+                            } catch (IllegalFormatException e2) {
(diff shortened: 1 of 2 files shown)


tfmorris:
Welcome to the project and thanks for the contribution!

> Not run locally: compiling the project dependencies exceeded the available command time limit.

Was this written using a coding agent? If so, please disclose that fact and tell us more about which one, how it was prompted, etc.

Please have a look at the test failures. It looks like the tests are not locale independent (we use the French locale for CI to catch any US locale assumption issues).

tfmorris:
No response to request for test fixes, so closing. Feel free to reopen if you start working on it again.

PR #7933, opened 2026-08-28
Inline Cypress linting & UDS test

### LINKED ISSUE #7941: Refactor snapshot build to require successful tests before build
As mentioned in #7692, although the snapshot build currently runs the end-to-end tests, they aren't a prerequisite for the build, so we can have a failed tests in parallel with published releases, as happened on the 3.9 branch for 3.9.3+. It also looks like the Windows installer is being run twice, only one of which is used.

We want to reorganize things so that the build is broken into the following phases with each phase requiring successful completion of all pieces of the previous phase:

1. lint all sources, install and cache prerequisites, build Java artefacts
2. run all unit tests and e2e tests in parallel (same a pull request workflow up to here and should probably be restructured to be reusable)
3. build kits for Linux, Mac (Intel only currently, but soon to include ARM), Windows w/JRE, Windows w/o JRE
4. ? optional for release builds only ? - install and test built kits on their native platforms?
5. Upload Github release and Maven artefacts

This work should be done before #7692 is tackled.

tfmorris:
Fixes #7941

- restructure workflow to make it more efficient.
- cache OpenRefine build artefacts

### DIFF
--- a/.github/workflows/pull_request_e2e.yml
+++ b/.github/workflows/pull_request_e2e.yml
@@ -14,31 +14,6 @@ on:
 permissions: read-all
 
 jobs:
-  lint_cypress:
-    name: Lint Cypress tests
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v7
-
-      - name: Enable corepack
-        run: corepack enable
-
-      - name: Setup Node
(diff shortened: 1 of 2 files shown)


tfmorris:
I'm going to roll this into the more general refactoring for #7941

PR #7935, opened 2026-08-31
Quote key value & tighten up key file handling

tfmorris:
No issue.

- quote key value when writing it to key file
- tighten permissions on key file & make sure it gets deleted even on error
- wrap long maven command line
- remove maven debug (optimistically) 

First two sourced from @magdmartin and Claude Opus 5


tfmorris:
Needs to be tested live, so merging without review.

magdmartin:
@tfmorris, checking the log https://github.com/OpenRefine/OpenRefine/actions/runs/33425139115/job/99596769365#step:12:22 this may be my mistake when copy-pasting the key, I think I have omitted `-----BEGIN PRIVATE KEY-----` and  `-----END PRIVATE KEY-----` 

I just updated `APPLE_BUILD_KEY` in GitHub action secret

tfmorris:
Yeah, it was definitely never going to work without those lines since they
delineate the key.

Issue #7921, opened 2026-08-28
Update Apple notarization credentials for kit build


tfmorris:
The credentials that we used to notarize builds were removing in March 2026. New credentials based on a team key have been created. We need to switch the build over to use them for both snapshot and release builds.

tfmorris:
Fixed in a series of branches/commits culminating with #7940. Also required the Github key value secret to be recreated correctly. Thanks to @magdmartin for the secrets management and interfacing with CS&S.

magdmartin:
@tfmorris small follow-up cleanup on the secret management. Should we delete the following secrets under `Settings → Secrets and variables → Actions` since they are no longer used
* APPLE_PASSWORD 
* APPLE_USERNAME
* APPLE_TEAM_ID 

That would avoid leaving misleading variables for next time.

PR #7938, opened 2026-08-31
Apple Notarization followups

magdmartin:
Follow-up to #7935 
Related to #7921

This PR was prepared by Claude (Opus-5) following a troubleshooting session to move the Apple credential account to use `APPLE_ISSUER` `APPLE_BUILD_KEY_ID` and `APPLE_BUILD_KEY`

Claude identified the following cleanup task; I asked it to prepare the PR since it already had the context and details. 

Below is verbatim from Claude 

----

**Validate the notarization ID secrets.** Whitespace pasted into APPLE_BUILD_KEY_ID or APPLE_ISSUER is invisible in the secrets UI but changes the signed JWT, and Apple rejects it as a bare "401 Unauthenticated" naming neither the value nor the reason. Both have fixed widths, so trim and check them next to the existing key parse check.

**Drop --team-id.** It belongs to notarytool's Apple ID authentication method; API key authentication implies the team from the issuer ID. Confirmed by authenticating against the notary API with only the key, key ID and issuer ID.

**Add check_notarization_credentials.py.** Credentials previously could only be tested by running the full macOS job. This builds the same ES256 JWT notarytool builds and validates the triple in a second from any machine with Python — no macOS, no Xcode, no CI run — distinguishing 401 (values do not correspond), 403 (valid but lacking access) and 200 (plus the team's submission history). That history is team-wide, so it also answers when notarization last succeeded, which build logs cannot: log retention is shorter than the five and a half months this went unnoticed.

The README now records what each secret holds, that only the Account Holder can renew the certificate, and that local decryption needs the private key rather than just the passphrase

----

I let the Core Dev Team decide what should get merged. 

I am also preparing an update under https://openrefine.org/docs/technical-reference/maintainer-guidelines to explain how the Apple notarization is configured.

### DIFF
--- a/.github/workflows/snapshot_release.yml
+++ b/.github/workflows/snapshot_release.yml
@@ -189,9 +189,28 @@ jobs:
         wc "$RUNNER_TEMP/notary_key.p8"
         openssl pkey -in "$RUNNER_TEMP/notary_key.p8" -noout \
           || { echo "::error::APPLE_BUILD_KEY is not a valid PKCS#8 private key"; exit 1; }
+
+        # Whitespace pasted into these two secrets is invisible in the web UI,
+        # but it changes the JWT and Apple rejects it as "401 Unauthenticated"
+        # with no indication of which value is at fault. Normalise and check
+        # them here, where the error names the culprit.
+        key_id="${APPLE_BUILD_KEY_ID//[[:space:]]/}"
+        issuer="${APPLE_ISSUER//[[:space:]]/}"
+        [ ${#key_id} -eq 10 ] \
+          || { echo "::error::APPLE_BUILD_KEY_ID is ${#key_id} characters after trimming, expected 10"; exit 1; }
+        [ ${#issuer} -eq 36 ] \
+          || { echo "::error::APPLE_ISSUER is ${#issuer} characters after trimming, expected a 36-character UUID"; exit 1; }
+        # Trimmed values are no longer identical to the stored secrets, so they
+        # are not covered by the automatic masking. Mask them explicitly.
+        echo "::add-mask::$key_id"
+        echo "::add-mask::$issuer"
+        echo "APPLE_BUILD_KEY_ID=$key_id" >> "$GITHUB_ENV"
+        echo "APPLE_ISSUER=$issuer" >> "$GITHUB_ENV"
       env:
         APPLE_BUILD_KEY: |
           ${{ secrets.APPLE_BUILD_KEY }}
+        APPLE_BUILD_KEY_ID: ${{ secrets.APPLE_BUILD_KEY_ID }}
+        APPLE_ISSUER: ${{ secrets.APPLE_ISSUER }}
 
     - name: Set up Java 21
       uses: actions/setup-java@v5.7.0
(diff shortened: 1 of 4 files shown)


packaging/apple_certs/README
magdmartin, line 1: I think this information is best in the documentation rather than the README.

.github/workflows/snapshot_release.yml
```diff
@@ -248,16 +267,13 @@ jobs:
...
+          -Dapple.notarization.issuer="$APPLE_ISSUER" -Dgpg.signer=bc
         rm -f $RUNNER_TEMP/notary_key.p8
       env:
         CENTRAL_USER: ${{ secrets.CENTRAL_USER }}
         CENTRAL_PASS: ${{ secrets.CENTRAL_PASS }}
         MAVEN_GPG_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
         MAVEN_GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
-        APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
-        APPLE_BUILD_KEY_ID: ${{ secrets.APPLE_BUILD_KEY_ID }}
-        APPLE_ISSUER: ${{ secrets.APPLE_ISSUER }}
```
tfmorris, line 260: This deletes two secrets which are still being used.

.github/workflows/snapshot_release.yml
```diff
@@ -189,9 +189,28 @@ jobs:
...
+        # are not covered by the automatic masking. Mask them explicitly.
+        echo "::add-mask::$key_id"
+        echo "::add-mask::$issuer"
+        echo "APPLE_BUILD_KEY_ID=$key_id" >> "$GITHUB_ENV"
+        echo "APPLE_ISSUER=$issuer" >> "$GITHUB_ENV"
       env:
         APPLE_BUILD_KEY: |
           ${{ secrets.APPLE_BUILD_KEY }}
+        APPLE_BUILD_KEY_ID: ${{ secrets.APPLE_BUILD_KEY_ID }}
+        APPLE_ISSUER: ${{ secrets.APPLE_ISSUER }}
```
tfmorris, line 213: This is overkill and only deals with whitespace, not punctuation or any other key changes. Verbatim, means verbatim. Nothing works if the key is changed.

packaging/apple_certs/check_notarization_credentials.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python3
+"""Validate an App Store Connect API key against Apple's notary service.
+
+The Linux equivalent of `xcrun notarytool history` - builds the same ES256 JWT
+notarytool builds, and calls the same endpoint. Useful for separating credential
+problems from build-pipeline problems without a Mac or a full CI run.
```
tfmorris, line 6: This wouldn't have caught the problem with the malformed Github secret and duplicates functionality which is built in to Apple's `notarytool`

packaging/apple_certs/README
tfmorris: Developer information belongs on Github where the developers live. We don't want to be forcing them to leave Github to go to a separate site. Having said that, we want to be careful about repeating information which is already contained in the workflows, because it'll just come stale, so this should be significantly trimmed down. Something which could be added, perhaps, is a link to the relevant section of the Apple developer's site (although letting them use web search might be less likely to get stale).

tfmorris (commented):
Thanks for the followup. I removed the team ID and confirmed that things still work, so that'll be one less source of confusion.

We could add a little more high level documentation to the README, but I'm not sure the rest of Claude's stuff is useful.

magdmartin:
Mostly irrelevant or already addressed. I will open a separate PR for the documentation.

PR #7942, opened 2026-09-01
Refactor build - refs #7941

tfmorris:
Refs #7941

- Move build into common pre-job and cache result
- Move lint & misc. UDS test into pre-job
- Run unit tests in parallel with e2e tests for both PR and snapshot
- Make snapshot kit build depend on successful unit & e2e tests

Still to do:
- share/reuse common pieces of workflow
- ?maybe? - test built kits on their native platforms


tfmorris:
The PR piece works as intended, so let's merge to see what the snapshot build looks like

Issue #6052, opened 2023-09-12
Native macOS support (arm64/Apple Silicon, without Rosetta).


elvey:
OpenRefine v3.7.5 won't run natively/without Rosetta on M1 / M2 / M3 -powered Macs.  It's compiled for Intel.  It seems like some subparts are being compiled into Universal apps, but not the main app.  I haven't installed Rosetta for security reasons. I'd like to use OpenRefine relatively safely and run it natively.

wetneb:
We probably need two MacOS packages: one for x64, one for aarch64.
The build times will be quite long because notarizing two DMGs will take a really long time.

At the moment, all packages are generated by the same job, sequentially. Perhaps it is time to reconsider that and generate them on different jobs. We'd be running two MacOS jobs generating one DMG each. The downside of that is that all jobs will probably have to download dependencies and compile the app independently, unless we figure out a sensible way to share the intermediate compilation state.

tfmorris:
This describes how to build a homegrown universal JRE: https://incenp.org/notes/2023/universal-java-app-on-macos.html
but this should really be solved by the Java platform folks, so I think it's blocked by https://bugs.openjdk.org/browse/JDK-8310589

I'm not sure it makes a ton of sense to invest lots of time/effort in workarounds like multiple builds (which the user then needs to choose correctly) rather than just waiting for the ecosystem to mature.

elvey:
Thanks. Re. avoiding duplicate effort:fair enough.  I do note that [JDK-8310589] has Priority: P4: minor loss of function where easy workaround is present, however. 
I see Oracle offers, for macOS, arm64:
JDK 20 binaries are free to use in production and free to redistribute, at no cost, under the [Oracle No-Fee Terms and Conditions](https://java.com/freeuselicense).
https://download.oracle.com/java/20/latest/jdk-20_macos-aarch64_bin.dmg 
Separately, https://jdk.java.net/20/ ALSO offers a macOS / AArch64 (AKA macOS / arm64) binary. 
Maybe I'm clueless, but it seems like one or both of these would be usable?   The incenp note seems out of date in this respect.

wetneb:
Yes JDKs are available for both architectures - the problem is that there is not a single JDK which works for both of them, so we'd have to publish two separate packages for both architectures.

mhkeller:
Any update on this. It would be nice to have this work natively on newer Macs.

tfmorris:
[...]

PR #7947, opened 2026-09-04
Update Mac appbundler. Fixes #7894

### LINKED ISSUE #7894: JRELoadError trying to load bundled Java 21 JRE for Mac OS X
We currently use 
```
  <groupId>sh.tak.appbundler</groupId>
  <artifactId>appbundle-maven-plugin</artifactId>
```
as our Java bundler to build for Mac OS X, but as pointed out in #2067 , it [doesn't support recent versions of Java](https://github.com/federkasten/appbundle-maven-plugin/issues/61), most likely due to this change for Java 12 that changed the path to libjli.dylib which is used by the loader. When you try to launch the app, you get an enigmatic JRELoadError with no additional detail.

It's not clear what the best replacement is. https://github.com/perdian/macosappbundler-maven-plugin started as a fork of our current plugin, but is no longer a drop-in replacement

```
        <groupId>de.perdian.maven.plugins</groupId>
        <artifactId>macosappbundler-maven-plugin</artifactId>
```

There may be other options, but more investigation is required.

tfmorris:
Fixes #7894

- Switches to e.perdian.maven.plugins:macosappbundler-maven-plugin which supports recent Javas
- Removes codesigning and a bunch of other stuff that's now built-in
- Preps for, but does not implement, aarch64 support (#6052)
- Detabifies and normalizes indentation (in a separate commit)

### DIFF
--- a/packaging/pom.xml
+++ b/packaging/pom.xml
@@ -18,7 +18,7 @@
     <webapp.target>../main/target</webapp.target>
     <embedded.jre.version>${java.minversion}</embedded.jre.version>
 
-      <!-- Adoptium JREs x64 (undocumented x86_64 works too), aarch64; mvn os.arch x86_64, aarch64 -->
+    <!-- Adoptium JREs x64 (undocumented x86_64 works too), aarch64; mvn os.arch x86_64, aarch64 -->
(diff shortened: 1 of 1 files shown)


tfmorris:
Merging to unbreak Mac snapshot build

